PR #16, opened 2023-09-05
inserting inline images v0

orm011:


pixeltable/exec.py
```diff
@@ -675,11 +675,24 @@ def _open(self) -> None:
...
             self.output_rows = DataRowBatch(self.tbl, self.evaluator, len(self.input_rows))
+            # assign row ids prior to needing them for file names
+            self.output_rows.set_row_ids([self.start_row_id + i for i in range(len(self.output_rows))])
+            
             for info in self.input_cols:
                 assert info.col.name in self.row_column_pos
                 col_idx = self.row_column_pos[info.col.name]
                 for row_idx, input_row in enumerate(self.input_rows):
-                    self.output_rows[row_idx, info.slot_idx] = input_row[col_idx]
+                    if info.col.col_type.type_enum == catalog.ColumnType.Type.IMAGE and isinstance(input_row[col_idx], bytes):
```
mkornacker: use ColumnType.is_image_type()

pixeltable/exec.py
```diff
@@ -675,11 +675,24 @@ def _open(self) -> None:
...
+                        # We're receiving an inline image binary. We will save it to a file here and store the path.
+                        valbytes = input_row[col_idx]
+                        row = self.output_rows[row_idx]
+                        col = self.tbl.cols_by_name[info.col.name]
```
mkornacker: unnecessary: col == info.col

pixeltable/type_system.py
```diff
@@ -738,15 +739,26 @@ def to_tf(self) -> Union['tf.TypeSpec', Dict[str, 'tf.TypeSpec']]:
...
+        # make sure file path points to a valid image file or binary is a valid image
+        if not isinstance(val, (str, bytes)):
+            raise TypeError(f'Expected file path or binary, got {val}')
+
+        if isinstance(val, bytes):
+            try:
+                _ = Image.open(io.BytesIO(val))
+            except PIL.UnidentifiedImageError:
+                raise TypeError(f'Binary is not a valid image: {val}')
+
```
mkornacker: remove blank lines

mkornacker (dismissed):

pixeltable/exec.py
```diff
@@ -675,11 +675,23 @@ def _open(self) -> None:
...
             for info in self.input_cols:
                 assert info.col.name in self.row_column_pos
                 col_idx = self.row_column_pos[info.col.name]
                 for row_idx, input_row in enumerate(self.input_rows):
-                    self.output_rows[row_idx, info.slot_idx] = input_row[col_idx]
+                    if info.col.col_type.is_image_type() and isinstance(input_row[col_idx], bytes):
+                        # We're receiving an inline image binary. We will save it to a file here and store the path.
+                        valbytes = input_row[col_idx]
+                        row = self.output_rows[row_idx]
+                        valpath = str(ImageStore.get_path(self.tbl.id, info.col.id, row.v_min, row.row_id, '.bin'))
```
mkornacker: Just a heads up: I just merged the initial implementation of views, which required changes to ImageStore.get_path().

The best way to do this conversion is to pre-process 'rows' prior to line 678, so that both if branches benefit from it.

orm011:
Working on fixing test after merge.

pixeltable/type_system.py
```diff
@@ -738,15 +739,24 @@ def to_tf(self) -> Union['tf.TypeSpec', Dict[str, 'tf.TypeSpec']]:
...
-            raise TypeError(f'Expected file path, got {val}')
-        try:
-            _ = Image.open(val)
-        except FileNotFoundError:
-            raise TypeError(f'File not found: {val}')
-        except PIL.UnidentifiedImageError:
-            raise TypeError(f'File is not a valid image: {val}')
+        # make sure file path points to a valid image file or binary is a valid image
+        if not isinstance(val, (str, bytes)):
+            raise TypeError(f'Expected file path or binary, got {val}')
```
mkornacker: "or bytes"

pixeltable/type_system.py
```diff
@@ -738,15 +739,24 @@ def to_tf(self) -> Union['tf.TypeSpec', Dict[str, 'tf.TypeSpec']]:
...
+
+        if isinstance(val, bytes):
+            try:
+                _ = Image.open(io.BytesIO(val))
+            except PIL.UnidentifiedImageError:
+                raise TypeError(f'Binary is not a valid image: {val}')
```
mkornacker: Don't print val, it's not going to look good.

"Not a valid" image is probably the best we can do.

Create a test case for this.

pixeltable/exec.py
```diff
@@ -727,11 +727,23 @@ def _open(self) -> None:
         """Create row batch and populate with self.data"""
         if not self.tbl.extracts_frames():
             self.output_rows = DataRowBatch(self.tbl, self.evaluator, len(self.input_rows))
+            # assign row ids prior to needing them for file names
```
mkornacker: Not true anymore. Revert the rowid assignment logic.

pixeltable/exec.py
```diff
@@ -727,11 +727,23 @@ def _open(self) -> None:
...
+            self.output_rows.set_row_ids([self.start_row_id + i for i in range(len(self.output_rows))])
+            
             for info in self.input_cols:
                 assert info.col.name in self.row_column_pos
                 col_idx = self.row_column_pos[info.col.name]
                 for row_idx, input_row in enumerate(self.input_rows):
-                    self.output_rows[row_idx, info.slot_idx] = input_row[col_idx]
+                    if info.col.col_type.is_image_type() and isinstance(input_row[col_idx], bytes):
```
mkornacker: Do this in a separate preprocessing loop over self.input_rows before line 728, so that both if branches benefit from it.

mkornacker (changes requested):

pixeltable/type_system.py
mkornacker: By 'create test case' I meant: add some "with pytest.raises(...):" logic to the existing test_create_images().

orm011: I ended up adding a new test for bad images, as it made sense to test both inline and file. let me know if that works. 

This is out of scope for this review, but, as an aside. 

The general idea of preventing errors downstream produced by invalid images is helpful from a user perspective, as it happens that some large jobs (eg indexing) may fail due to a small number of bad images which were not handled in the code. So, it makes sense to offer this to users.

On the other hand, erroring an insert upon finding a bad image may be undesirable from user perspective, at least from a bulk load situation. There always are bad images in a large collection, and at least for most datasets this is not a problem, as no specific image item is crucial.

From a user perspective,  finding bad images and excluding them from the loading job is itself an annoying chore, a job the table is better equipped to do becase it can do a parallel job of this and handle the errores well. keeping this error meta-data about the files around for the user is helpful.  (I'm thinking of images in S3 or lustre FS, or even remote http images)

pixeltable/tests/test_table.py
```diff
@@ -94,15 +95,57 @@ def test_create_images(self, test_client: pt.Client) -> None:
...
+        # add literal image data and column
+        for r in rows:
+            with open(r[0], 'rb') as f:
+                r.append(f.read())
+
+        col_names.append('img_literal')
+        tbl.insert(sample_rows, columns=col_names)
+
+        # compare img and img_literal
+        # TODO: currently tbl.select(tabl.img == tbl.img_literal) returns False. should not.
```
mkornacker: simplify: "TODO: make tbl.select(tbl.img == tbl.img_literal) work

pixeltable/tests/test_table.py
```diff
@@ -94,15 +95,57 @@ def test_create_images(self, test_client: pt.Client) -> None:
...
+        tdf = tbl.select(tbl.img, tbl.img_literal).show()
+        pdf = tdf.to_pandas()
+        for tup in pdf.itertuples():
+            assert tup.img == tup.img_literal
+
+        # check that literal and non-literal image cols are equal (ie no loss of fidelity)
```
mkornacker: didn't you already check that in line 118?

remove lines 120-123

pixeltable/type_system.py
```diff
@@ -738,15 +739,24 @@ def to_tf(self) -> Union['tf.TypeSpec', Dict[str, 'tf.TypeSpec']]:
...
-            raise TypeError(f'File not found: {val}')
-        except PIL.UnidentifiedImageError:
-            raise TypeError(f'File is not a valid image: {val}')
+        # make sure file path points to a valid image file or binary is a valid image
+        if not isinstance(val, (str, bytes)):
+            raise TypeError(f'Expected file path or bytes')
+
+        if isinstance(val, bytes):
+            try:
+                _ = Image.open(io.BytesIO(val))
```
mkornacker, line 748: would you be able to load this with cv2.imgdecode() as well?

pixeltable/type_system.py
```diff
@@ -738,15 +739,24 @@ def to_tf(self) -> Union['tf.TypeSpec', Dict[str, 'tf.TypeSpec']]:
...
+            except PIL.UnidentifiedImageError:
+                raise TypeError(f'Bytes are not a valid PIL image')
```
mkornacker: in which case this should just be "image", not "PIL image"

pixeltable/tests/test_table.py
```diff
@@ -94,15 +95,57 @@ def test_create_images(self, test_client: pt.Client) -> None:
...
+        cols = [
+            catalog.Column('img', ImageType(nullable=False)),
+            catalog.Column('category', StringType(nullable=False)),
+            catalog.Column('split', StringType(nullable=False)),
+        ]
+        tbl = test_client.create_table('test', cols)
+        rows, col_names = read_data_file('imagenette2-160', 'manifest_bad.csv', ['img'])
+
+        # check insert with bad image file fails
+        with pytest.raises(exc.Error):
```
mkornacker: also check that we're recognizing the error correctly with something like this:
with .. as exc_info:
...
assert 'not a valid image' in str(exc_info.value)

pixeltable/exec.py
```diff
@@ -725,10 +725,26 @@ def __init__(
...
+            for info in self.input_cols:
+                col_idx = self.row_column_pos[info.col.name]
+                val = input_row[col_idx]
+                if info.col.col_type.is_image_type() and isinstance(val, bytes):
+                    # we will save literal to a file here and use this path as the new value
+                    valpath = str(ImageStore.get_path(self.tbl.id, info.col.id, self.tbl.version))
+                    open(valpath, 'wb').write(val)
+                    new_row[col_idx] = valpath
+            _input_rows.append(new_row)
+        self.input_rows = _input_rows
```
mkornacker, line 746: follow with blank line

pixeltable/exec.py
```diff
@@ -725,10 +725,26 @@ def __init__(
 
     def _open(self) -> None:
         """Create row batch and populate with self.data"""
+
+        for info in self.input_cols:
+            assert info.col.name in self.row_column_pos
+        # before anything, convert any literal images within the input rows into references
+        _input_rows = [] # do not mutate self.input_rows, this may modify the argument also
```
mkornacker: you can create the copy here: ... = [row.copy() for row in self.input_rows]

pixeltable/exec.py
```diff
@@ -725,10 +725,26 @@ def __init__(
...
+        for input_row in self.input_rows:
```
mkornacker: ... and then loop over input_cols first (like in line 747), so you only need to do the row_column_pos lookup once per column

pixeltable/tests/test_table.py
```diff
@@ -94,15 +95,57 @@ def test_create_images(self, test_client: pt.Client) -> None:
...
+        pdf = tdf.to_pandas()
+        for tup in pdf.itertuples():
+            assert tup.img == tup.img_literal
+
+        # check that literal and non-literal image cols are equal (ie no loss of fidelity)
         html_str = tbl.show(n=100)._repr_html_()
         print(html_str)
         # TODO: check html_str
 
+    def test_insert_bad_images(self, test_client: pt.Client) -> None:
```
mkornacker, line 124: good idea to do it in a separate test case and test both paths.

pixeltable/type_system.py
mkornacker: Good point, and it sounds like it's worth thinking this through.

For computed columns, errors are explicitly recognized and recorded (as separate columns errortype and -msg, which  are exposed as properties of a column: eg, "computed_col_name.errortype"). They're also handled during expression evaluation (take a look at Evaluator.eval()). Something similar could be done for input data, ie, it just records the errors and skips all derived computed columns, but otherwise keeps going. This should probably be a table parameter, it may  not be appropriate in every situation.

But definitely worth adding if this makes it more attractive for researchers or for experimentation in general.

pixeltable/tests/test_table.py
orm011: I've removed the comment in line 120, which is stale. the rest of lines, 121-123 are not really my change.

pixeltable/type_system.py
orm011: Not sure, the converse also is interesting.  Sometimes PIL is not able to read an image but some other library can. I've seen this with some tiff files. For the most part PIL is ok.

It seems like flexibility for this sort of rough edge of image formats is something to keep in mind in terms of allowing user input on how to load. but maybe out of scope for this review.

pixeltable/exec.py
orm011: ok, changed.

pixeltable/tests/test_table.py
```diff
@@ -94,15 +95,58 @@ def test_create_images(self, test_client: pt.Client) -> None:
...
+            catalog.Column('category', StringType(nullable=False)),
+            catalog.Column('split', StringType(nullable=False)),
+        ]
+        tbl = test_client.create_table('test', cols)
+        rows, col_names = read_data_file('imagenette2-160', 'manifest_bad.csv', ['img'])
+
+        # check insert with bad image file fails
+        with pytest.raises(exc.Error) as exc_info:
+            tbl.insert(rows, columns=col_names)
+            assert 'not a valid PIL image' in str(exc_info.value)
```
mkornacker: This doesn't get executed, you need to unindent it.

pixeltable/tests/test_table.py
```diff
@@ -94,15 +95,58 @@ def test_create_images(self, test_client: pt.Client) -> None:
...
+
+        # replace row with literal
+        bad_row = rows[0]
+        img_idx = col_names.index('img')
+        bad_row[img_idx] = b'bad image literal'
+
+        # check insert with bad image literal fails
+        with pytest.raises(exc.Error) as exc_info:
+            tbl.insert(rows, columns=col_names)
+            assert 'not a valid PIL image' in str(exc_info.value)
```
mkornacker: Same here.

mkornacker (approved):
Fix the test asserts before push.

mkornacker (approved):

orm011:
Does something else need to be done to merge it?

PR #29, opened 2023-10-13
extend syntax to allow user specified alias in select via kwargs

orm011:


pixeltable/catalog/table_base.py
```diff
@@ -58,12 +58,12 @@ def df(self) -> 'pixeltable.dataframe.DataFrame':
         from pixeltable.dataframe import DataFrame
         return DataFrame(self.tbl_version)
 
-    def select(self, *items: 'exprs.Expr') -> 'pixeltable.dataframe.DataFrame':
+    def select(self, *items: 'exprs.Expr', **named_items: 'exprs.Expr') -> 'pixeltable.dataframe.DataFrame':
```
mkornacker: These shouldn't be exprs.Expr, you can also pass in dicts and lists (eg, see test_inline_dict()).

pixeltable/dataframe.py
```diff
@@ -118,47 +117,86 @@ def finalize_exec(self) -> None:
         if self.filter is not None:
             self.filter.release()
 
+def _expand_select_list(tbl : catalog.TableVersion,
```
mkornacker: Reformat to conform to prevailing style (eg, TableVersion.update())

Also, the name doesn't really match what you're doing (you're not necessarily expanding).

pixeltable/dataframe.py
```diff
@@ -118,47 +117,86 @@ def finalize_exec(self) -> None:
...
     def __init__(
             self, tbl: catalog.TableVersion,
-            select_list: Optional[List[exprs.Expr]] = None,
+            select_list:  Optional[List[Tuple[exprs.Expr, Optional[str]]]] = None,
             where_clause: Optional[exprs.Predicate] = None,
             group_by_clause: Optional[List[exprs.Expr]] = None,
             order_by_clause: Optional[List[Tuple[exprs.Expr, bool]]] = None,  # List[(expr, asc)]
             limit: Optional[int] = None,):
         self.tbl = tbl
-        # exprs contain execution state and therefore cannot be shared
```
mkornacker, line 131: This comment explains the need for the following calls to deepcopy()

pixeltable/dataframe.py
```diff
@@ -118,47 +117,86 @@ def finalize_exec(self) -> None:
...
+        expanded_list = []
+        for col in tbl.columns():
+            if tbl.is_frame_col(col):
+                expanded_list.append((exprs.FrameColumnRef(col), None))
+            else:
+                expanded_list.append((exprs.ColumnRef(col), None))
+    else:
+        expanded_list = select_list
+    out = []
+    for i, tup in enumerate(expanded_list):
```
mkornacker: for i, (expr, alias) in ...

is easier to read

pixeltable/dataframe.py
```diff
@@ -195,10 +236,10 @@ def collect(self) -> DataFrameResultSet:
         except sql.exc.DBAPIError as e:
             raise exc.Error(f'Error during SQL execution:\n{e}')
 
-        col_names = [expr.display_name() for expr in self.select_list]
-        # replace ''
-        col_names = [n if n != '' else f'col_{i}' for i, n in enumerate(col_names)]
-        return DataFrameResultSet(data_rows, col_names, [expr.col_type for expr in self.select_list])
+        expanded_select_list = _expand_select_list(self.tbl, self.select_list)
```
mkornacker: Given that the caller always separates the exprs from the aliases, might be good to return them separately.

pixeltable/dataframe.py
```diff
@@ -170,6 +208,9 @@ def show(self, n: int = 20) -> DataFrameResultSet:
 
     def head(self, n: int = 20) -> DataFrameResultSet:
         return self.show(n)
+    
+    def _column_names(self):
```
mkornacker: Given that you refer to aliases elsewhere, best to use that term here as well for consistency.

pixeltable/dataframe.py
```diff
@@ -287,30 +328,46 @@ def __repr__(self) -> str:
...
+            else:
+                select_list.append((exprs.Literal(raw_expr), alias))
+            expr = select_list[-1][0]
             if expr.col_type.is_invalid_type():
-                raise exc.Error(f'Invalid type: {expr}')
+                raise exc.Error(f'Invalid type: {raw_expr}')
             # TODO: check that ColumnRefs in expr refer to self.tbl
+        # check user provided aliases would not conflict among themselves 
+        # or with auto-generated ones
+        seen = set()
```
mkornacker: type hint

pixeltable/dataframe.py
```diff
@@ -287,30 +328,46 @@ def __repr__(self) -> str:
...
             if expr.col_type.is_invalid_type():
-                raise exc.Error(f'Invalid type: {expr}')
+                raise exc.Error(f'Invalid type: {raw_expr}')
             # TODO: check that ColumnRefs in expr refer to self.tbl
+        # check user provided aliases would not conflict among themselves 
+        # or with auto-generated ones
+        seen = set() 
+        for (_, alias) in _expand_select_list(self.tbl, select_list):
+            if alias in seen:
+                raise exc.Error(f'Repeated alias: {alias}')
```
mkornacker: Might not be obvious to a user what the two offending expressions are. Better to list those explicitly.

pixeltable/tests/test_dataframe.py
```diff
@@ -25,6 +31,34 @@ def test_select_where(self, test_tbl: catalog.Table) -> None:
...
+            _ = t.select(t.c1, **{'foo.bar': t.c2}).show(0)
+        assert 'Invalid alias' in str(exc_info.value)
+
+        # catch repeated alias from user input
+        with pytest.raises(exc.Error) as exc_info:
+            _ = t.select(t.c2, c2=t.c1).show(0)
+        assert 'Repeated alias' in str(exc_info.value)
+
+        with pytest.raises(exc.Error) as exc_info:
+            _ = t.select(t.c2, t.c2).show(0)
```
mkornacker: Hm, this should probably be legal.

In the end, the alias is simply meant for displaying the output, so maybe duplicates don't matter.

mkornacker (changes requested):

pixeltable/catalog/table_base.py
orm011: I'm almost sure this is the python convention,  

the list and dict[str,] for *args and **kwargs are implicit in python, and typechecking tools like mypy,  expect that (see link below) list[expr]would mean list of list of expr/

https://docs.python.org/3/library/typing.html (they link to the mypy page explaining  near the top of the page)
 https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html, but also have examples in there)

mkornacker: What I meant was: it needs to be Union[Expr, Any]

pixeltable/tests/test_dataframe.py
orm011: We had a chat about selections a couple of weeks ago. My impression was users potentially use the aliases downstream. 

Eg. the PyTorch dataset interface, any parquet export, and any downstream code we write will need to specify what happens in the case of repeated columns.  Some of it is not obvious, eg, python dictionaries do not allow repeated keys.

If the ambiguity will be an issue everywhere downstream, doesn't it save us work not having to think of those cases downstream?

mkornacker: Yeah, good point. I was just thinking about how that's handled in sql, where this is legal, but it wouldn't be legal in a Create Table As Select statement, so there's no 1x1 analogy here.

orm011: There's a related issue with stuff like t.video.url, where the auto-generated column name includes dots. I decided against looking into this, since it delays stuff, but I notice it would not be a valid python identifier. That by itself is not an issue but makes me wonder if it would be a valid standalone column identifier in a table, eg. I'm wondering what happens if a user decided to create a view with this expression. Can one write queries against that view using a python API?

orm011: I'm leaving that out of scope for this code review, as that behavior has not been changed (though a user currently cannot use select("video.url" ) as a user provided alias, as python won't allow it as a key in a kwarg).

pixeltable/dataframe.py
orm011: I use the name for column names, and alias for the syntactically provided names "select(alias=expr)", i've  now changed the code to be more consistent with this.

pixeltable/catalog/table_base.py
```diff
@@ -58,12 +58,15 @@ def df(self) -> 'pixeltable.dataframe.DataFrame':
         from pixeltable.dataframe import DataFrame
         return DataFrame(self.tbl_version)
 
-    def select(self, *items: 'exprs.Expr') -> 'pixeltable.dataframe.DataFrame':
+    def select(
+                self, *items: Union['exprs.Expr', dict[str,'exprs.Expr'], list['exprs.Expr']],
```
mkornacker: Is this for inline dicts and arrays? It's not correct, because those can also contain constants like numbers or strings.

Best to use Any here.

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
...
+        out_names.append(column_name) # user provided alias
+        seen_out_names.add(column_name)
+    assert len(out_exprs) == len(out_names)
+    assert set(out_names) == seen_out_names
+    return out_exprs, out_names
+
+def _select_list_check_rep(
+        select_list: Optional[List[Tuple[exprs.Expr, Optional[str]]]],
+) -> None:
+    """ validate basic select list types.
```
mkornacker: Use standard format (no space after """, capitalize).

pixeltable/exprs.py
```diff
@@ -94,6 +94,8 @@ def __str__(self) -> str:
         if self == self.MOD:
             return '%'
 
+def is_valid_column_name(name: str) -> bool:
```
mkornacker: This shouldn't live here, it's really a catalog thing.

pixeltable/exprs.py
```diff
@@ -153,11 +155,12 @@ def bind_rel_paths(self, mapper: Optional[JsonMapper] = None) -> None:
         for c in self.components:
             c.bind_rel_paths(mapper)
 
-    def display_name(self) -> str:
+    def default_column_name(self) -> Optional[str]:
         """
-        Displayed column name in DataFrame. '': assigned by DataFrame
+        Default column name in DataFrame
```
mkornacker: Follow style guide docstring convention:
https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

pixeltable/exprs.py
```diff
@@ -502,8 +505,9 @@ def __getattr__(self, name: str) -> Expr:
 
         return super().__getattr__(name)
 
-    def display_name(self) -> str:
-        return str(self)
+    def default_column_name(self) -> str:
+        ret =  str(self)
```
mkornacker: Collapse into single line

pixeltable/tests/test_dataframe.py
```diff
@@ -15,6 +15,14 @@ def test_select_where(self, test_tbl: catalog.Table) -> None:
         res2 = t.where(t.c2 < 10).select(t.c1, t.c2, t.c3).show(0)
         assert res1 == res2
 
+        res3 = t.where(t.c2 < 10).select(c1=t.c1, c2=t.c2, c3=t.c3).show(0)
```
mkornacker, line 18: Would be good to test that if you create different aliases the result is still the same.

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
         if self.filter is not None:
             self.filter.release()
 
+def _expand_select_list_resolve_names(
```
mkornacker: Chatty function name, and basically describing the implementation.

Something like 'normalize_select_list' might be better.

Also, is this useful outside of DataFrame?

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
...
+                column_name = default_name
+                if default_name in seen_out_names:
+                    # already used, then add suffix until unique name is found
+                    for j in range(1, len(out_names)+1):
+                        mod_name = f'{default_name}_{j}'
+                        if mod_name not in seen_out_names:
+                            column_name = mod_name
+                            break
+            else: # no default name, eg some expressions
+                column_name = f"col_{i}"
```
mkornacker: use ' for consistency

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
...
+    for i, (expr, alias) in enumerate(expanded_list):
+        if alias is None:
+             # no user provided alias: use default
+             # add suffix if needed so there are no duplicates
+            default_name = expr.default_column_name()
+            if default_name is not None:
+                column_name = default_name
+                if default_name in seen_out_names:
+                    # already used, then add suffix until unique name is found
+                    for j in range(1, len(out_names)+1):
```
mkornacker: formatting (" + 1")

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
         if self.filter is not None:
             self.filter.release()
 
+def _expand_select_list_resolve_names(
+        tbl : catalog.TableVersion, 
+        select_list : Optional[List[Tuple[exprs.Expr, Optional[str]]]],
+    ) -> Tuple[list[exprs.Expr], list[str]]:
+    """
+    expand select list information with all columns and their names
+    """
```
mkornacker: use blank lines to divide into the logic into logically separate sections

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
...
+    if select_list is None:
+        expanded_list = []
+        for col in tbl.columns():
+            if tbl.is_frame_col(col):
+                expanded_list.append((exprs.FrameColumnRef(col), None))
+            else:
+                expanded_list.append((exprs.ColumnRef(col), None))
+    else:
+        expanded_list = select_list
+    out_exprs : List[exprs.Expr] = []
```
mkornacker: formatting

https://google.github.io/styleguide/pyguide.html#221-type-annotated-code

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
...
+        if alias is None:
+             # no user provided alias: use default
+             # add suffix if needed so there are no duplicates
+            default_name = expr.default_column_name()
+            if default_name is not None:
+                column_name = default_name
+                if default_name in seen_out_names:
+                    # already used, then add suffix until unique name is found
+                    for j in range(1, len(out_names)+1):
+                        mod_name = f'{default_name}_{j}'
```
mkornacker: assign to column_name directly and save line 152

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
         if self.filter is not None:
             self.filter.release()
 
+def _expand_select_list_resolve_names(
+        tbl : catalog.TableVersion, 
+        select_list : Optional[List[Tuple[exprs.Expr, Optional[str]]]],
+    ) -> Tuple[list[exprs.Expr], list[str]]:
+    """
+    expand select list information with all columns and their names
```
mkornacker: Doesn't state what it returns.

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
...
+                        if mod_name not in seen_out_names:
+                            column_name = mod_name
+                            break
+            else: # no default name, eg some expressions
+                column_name = f"col_{i}"
+        else: # user provided alias, no attempt to rename
+            column_name = alias
+
+        out_exprs.append(expr)
+        out_names.append(column_name) # user provided alias
```
mkornacker: obsolete comment?

pixeltable/dataframe.py
```diff
@@ -170,6 +229,10 @@ def show(self, n: int = 20) -> DataFrameResultSet:
 
     def head(self, n: int = 20) -> DataFrameResultSet:
         return self.show(n)
+    
+    def _column_names(self):
```
mkornacker: missing type annotation

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
...
 
     def _exec(self) -> Generator[exprs.DataRow, None, None]:
         """Returned value: list of select list values"""
-        if self.select_list is None:
-            # select all columns
-            self.select_list = [
-                exprs.FrameColumnRef(col) if self.tbl.is_frame_col(col) else exprs.ColumnRef(col)
-                for col in self.tbl.columns()
-            ]
+        exps, _ = _expand_select_list_resolve_names(self.tbl, self.select_list)
```
mkornacker: these are select list items (or select list exprs)

pixeltable/dataframe.py
```diff
@@ -118,47 +118,106 @@ def finalize_exec(self) -> None:
...
         if self.group_by_clause is None:
             self.group_by_clause = []
         if self.order_by_clause is None:
             self.order_by_clause = []
-        for item in self.select_list:
+        for item in exps:
             item.bind_rel_paths(None)
-        plan, self.select_list = Planner.create_query_plan(
-            self.tbl, self.select_list, where_clause=self.where_clause, group_by_clause=self.group_by_clause,
+        plan, select_list_tmp = Planner.create_query_plan(
```
mkornacker, line 219: at this point, you're replace the select list items (and exps shouldn't be referenced anymore)

pixeltable/dataframe.py
```diff
@@ -195,10 +258,9 @@ def collect(self) -> DataFrameResultSet:
         except sql.exc.DBAPIError as e:
             raise exc.Error(f'Error during SQL execution:\n{e}')
 
-        col_names = [expr.display_name() for expr in self.select_list]
-        # replace ''
-        col_names = [n if n != '' else f'col_{i}' for i, n in enumerate(col_names)]
-        return DataFrameResultSet(data_rows, col_names, [expr.col_type for expr in self.select_list])
+        exps, col_names = _expand_select_list_resolve_names(self.tbl, self.select_list)
```
mkornacker: you're doing this in two places now, which leaves room for inconsistency to creep in.

might be best to do this before the exec() call and pass it in as a parameter

pixeltable/dataframe.py
```diff
@@ -264,7 +326,7 @@ def _description(self) -> pd.DataFrame:
         assert len(heading_vals) == len(info_vals)
         return pd.DataFrame({'Heading': heading_vals, 'Info': info_vals})
 
-    def _description_html(self) -> pd.DataFrame:
+    def _description_html(self) -> Styler:
```
mkornacker: use qualified name. Styler on its own doesn't make it very clear that this is pandas stuff

pixeltable/dataframe.py
```diff
@@ -287,30 +349,51 @@ def __repr__(self) -> str:
     def _repr_html_(self) -> str:
         return self._description_html()._repr_html_()
 
-    def select(self, *items: exprs.Expr) -> DataFrame:
+    def select(self,
```
mkornacker: indentation

pixeltable/dataframe.py
```diff
@@ -287,30 +349,51 @@ def __repr__(self) -> str:
...
-            if isinstance(expr, dict):
-                select_list[i] = expr = exprs.InlineDict(expr)
-            elif isinstance(expr, list):
-                select_list[i] = expr = exprs.InlineArray(tuple(expr))
-            elif not isinstance(expr, exprs.Expr):
-                select_list[i] = expr = exprs.Literal(expr)
+        for (alias, _) in named_items.items():
+            if not isinstance(alias, str) or not is_valid_column_name(alias):
+                raise exc.Error(f'Invalid alias: {alias}')
+        base_list = [(expr, None) for expr in items] + [(expr, k) for (k,expr) in named_items.items()]
```
mkornacker, line 366: formatting

pixeltable/dataframe.py
```diff
@@ -287,30 +349,51 @@ def __repr__(self) -> str:
     def _repr_html_(self) -> str:
         return self._description_html()._repr_html_()
 
-    def select(self, *items: exprs.Expr) -> DataFrame:
+    def select(self, 
+               *items: Union[exprs.Expr, dict[str,exprs.Expr], list[exprs.Expr]],
+               **named_items : Union[exprs.Expr, dict[str,exprs.Expr], list[exprs.Expr]],
+            ) -> DataFrame:
```
mkornacker: here (and elsewhere, where applicable): use blank lines to divide logic into cohesive blocks

pixeltable/dataframe.py
```diff
@@ -287,30 +349,51 @@ def __repr__(self) -> str:
...
+                raise exc.Error(f'Invalid type: {raw_expr}')
             # TODO: check that ColumnRefs in expr refer to self.tbl
+        # check user provided aliases would not conflict among themselves
+        # or with auto-generated ones
+        seen : set[str] = set()
+        _, names = _expand_select_list_resolve_names(self.tbl, select_list)
+        for (i, name) in enumerate(names):
+            if name in seen:
+                repeated_names = [j for j, x in enumerate(names) if x == name]
+                raise exc.Error(f'Repeated column name {name} at positions {repeated_names}')
```
mkornacker: in line 360 you're using 'alias'

pixeltable/dataframe.py
```diff
@@ -287,30 +349,51 @@ def __repr__(self) -> str:
...
+                select_list.append((exprs.Literal(raw_expr), alias))
+            expr = select_list[-1][0]
             if expr.col_type.is_invalid_type():
-                raise exc.Error(f'Invalid type: {expr}')
+                raise exc.Error(f'Invalid type: {raw_expr}')
             # TODO: check that ColumnRefs in expr refer to self.tbl
+        # check user provided aliases would not conflict among themselves
+        # or with auto-generated ones
+        seen : set[str] = set()
+        _, names = _expand_select_list_resolve_names(self.tbl, select_list)
```
mkornacker: you already have _column_names()

pixeltable/exprs.py
mkornacker, line 97: This shouldn't live here, it's really a catalog thing.

pixeltable/tests/test_dataframe.py
mkornacker, line 18: Should also have a test that the result doesn't change if you change the aliases.

mkornacker (changes requested):

pixeltable/catalog/table_base.py
orm011: Ah true. I'll change it to Any.

pixeltable/dataframe.py
orm011: Not useful outside of dataframe, on the other hand, they dont depend on a fully formed 'self' DataFrame object, and are useful in the constructor to check the select list has no problems. 
I made them static methods so it is clear they don't assume a fully formed object, not sure what else to do.

pixeltable/dataframe.py
orm011: Read the note but don't yet understand the issue.

pixeltable/dataframe.py
orm011: Not sure I follow. 
What's the underlying reason for the assignment to the self.select_list ?
Just to recap a previous conversation we had about this assignment: my understanding from an earlier conversation around this was that replacing the select list was not needed.

pixeltable/dataframe.py
orm011: Not sure I follow.

pixeltable/dataframe.py
orm011: Not sure I follow. format string related? or something else.

pixeltable/dataframe.py
orm011: I was using alias to refer to user provided names for columns.  Early on, they are user provided names, 
later on they are a mix of user provided and deduced names.

Since I see no benefit to keeping that distinction, I'll just use name everywhere, including error messages etc.

pixeltable/catalog/table_base.py
orm011: I think Union[Expr,Any] = Any, so I'll just use Any as you mention later.

pixeltable/tests/test_dataframe.py
orm011: Current version makes it legal, but makes a new name for the repeated column to keep names unique. 
There's a test checking this is legal.

pixeltable/exprs.py
orm011: Ok. There are different files there, not sure where is a suitable place.

pixeltable/tests/test_dataframe.py
orm011: Could you expand on what should be the same? 
do you mean the values of the different columns based on the same column?
eg (t.c1, c1prime=t.c1) c1 and c1prime should have equal values? or something else.

pixeltable/dataframe.py
orm011: not sure I follow. I use exps bc exprs is used by the module.

pixeltable/exprs.py
orm011: Ok. I've placed them in the catalog. I figured the current regexp based checks should replaced with these shared validation methods to stay consistent.

added a small test.

pixeltable/dataframe.py
orm011: Since the type signature has changed back to a one liner,  I guess its ok now either way.

pixeltable/catalog/column.py
```diff
@@ -10,9 +9,7 @@
 from pixeltable import exceptions as exc
 from pixeltable.metadata import schema
 from pixeltable.type_system import ColumnType, StringType
-
-_ID_RE = r'[a-zA-Z]\w*'
-_PATH_RE = f'{_ID_RE}(\\.{_ID_RE})*'
+from pixeltable.catalog.validation import is_valid_identifier
```
mkornacker, line 12: isn't import .validation enough?

pixeltable/catalog/validation.py
```diff
@@ -0,0 +1,11 @@
...
+    return name.isidentifier() and not name.startswith('_')
+
+def is_valid_path(path: str, empty_is_valid : bool) -> bool:
+    if path == '':
+        return empty_is_valid
+        
+    for part in path.split('.'):
+        if not is_valid_identifier(part):
+            return False
+    return True
```
mkornacker, line 11: blank line at end of file missing

pixeltable/dataframe.py
```diff
@@ -119,46 +118,116 @@ def finalize_exec(self) -> None:
             self.filter.release()
 
 
+
 class DataFrame:
     def __init__(
             self, tbl: catalog.TableVersion,
-            select_list: Optional[List[exprs.Expr]] = None,
+            select_list:  Optional[List[Tuple[exprs.Expr, Optional[str]]]] = None,
```
mkornacker, line 125: extra space: ":  Optional"

pixeltable/dataframe.py
```diff
@@ -119,46 +118,116 @@ def finalize_exec(self) -> None:
...
         self.order_by_clause = copy.deepcopy(order_by_clause)
         self.limit_val = limit
 
+    @staticmethod
+    def _select_list_check_rep(
+        select_list: Optional[List[Tuple[exprs.Expr, Optional[str]]]],
+    ) -> None:
+        """Validate basic select list types. 
+        """
+        if select_list is not None: # basic check for valid select list
```
mkornacker, line 154: if select_list is None:
    return
assert ...
to reduce indentation levels

pixeltable/dataframe.py
```diff
@@ -119,46 +118,116 @@ def finalize_exec(self) -> None:
...
+                assert isinstance(ent, tuple)
+                assert len(ent) == 2
+                assert isinstance(ent[0], exprs.Expr)
+                assert ent[1] is None or isinstance(ent[1], str)
+                if isinstance(ent[1], str):
+                    assert is_valid_identifier(ent[1])
+
+    @staticmethod
+    def _normalize_select_list(
+        tbl : catalog.TableVersion,
```
mkornacker, line 166: tbl:

pixeltable/dataframe.py
```diff
@@ -119,46 +118,116 @@ def finalize_exec(self) -> None:
...
+
+    @staticmethod
+    def _normalize_select_list(
+        tbl : catalog.TableVersion, 
+        select_list : Optional[List[Tuple[exprs.Expr, Optional[str]]]],
+    ) -> Tuple[List[exprs.Expr], List[str]]:
+        """
+        Expand select list information with all columns and their names
+        Returns: 
+            tupel consisting of list of expressions, list of names
```
mkornacker, line 172: tuple
also, should say what is being return (not just the type): a list of select list items and their respective names

pixeltable/dataframe.py
mkornacker: space after ,
(k, expr)

pixeltable/dataframe.py
```diff
@@ -119,46 +118,116 @@ def finalize_exec(self) -> None:
...
         self.tbl = tbl
+
+        # select list logic
+        DataFrame._select_list_check_rep(select_list) # check select list without expansion
         # exprs contain execution state and therefore cannot be shared
-        self.select_list = copy.deepcopy(select_list)  # None: implies all cols
+        select_list = copy.deepcopy(select_list)
+        exps, column_names = DataFrame._normalize_select_list(tbl, select_list)
+        DataFrame._select_list_check_rep(list(zip(exps, column_names))) # check select list after expansion to catch early
+        self._exps = exps
```
mkornacker, line 138: You're basically caching something here, and it can deviate from the actual self.select_list (which might be none).

Choose a more meaningful name (select_list_exprs?) and prefix with a comment that these two are always set, even when select_list is None (which is an important part of their semantics).

pixeltable/dataframe.py
```diff
@@ -264,7 +334,7 @@ def _description(self) -> pd.DataFrame:
         assert len(heading_vals) == len(info_vals)
         return pd.DataFrame({'Heading': heading_vals, 'Info': info_vals})
 
-    def _description_html(self) -> pd.DataFrame:
+    def _description_html(self) -> 'pandas.io.formats.style.Styler':
```
mkornacker, line 337: does -> pd.io.formats.style.Styler
not work?

pixeltable/dataframe.py
```diff
@@ -287,30 +357,52 @@ def __repr__(self) -> str:
...
+                select_list.append((exprs.Literal(raw_expr), name))
+            expr = select_list[-1][0]
             if expr.col_type.is_invalid_type():
-                raise exc.Error(f'Invalid type: {expr}')
+                raise exc.Error(f'Invalid type: {raw_expr}')
             # TODO: check that ColumnRefs in expr refer to self.tbl
+
+        # check user provided names do not conflict among themselves
+        # or with auto-generated ones
+        seen : Set[str] = set()
```
mkornacker, line 388: "seen: Set"

pixeltable/exprs.py
```diff
@@ -153,11 +153,12 @@ def bind_rel_paths(self, mapper: Optional[JsonMapper] = None) -> None:
         for c in self.components:
             c.bind_rel_paths(mapper)
 
-    def display_name(self) -> str:
+    def default_column_name(self) -> Optional[str]:
         """
-        Displayed column name in DataFrame. '': assigned by DataFrame
+        Returns: 
+            None if this expression lacks a default name, or a valid identifier (catalog.is_valid_identifer) otherwise
```
mkornacker, line 159: "or a valid identifier (according to catalog.is_valid_identifier) otherwise"

pixeltable/tests/test_dataframe.py
mkornacker: The result of 
t.select(a=t.c1, b=t.c2, c=t.c3) should be the same

mkornacker (approved):

mkornacker:
commit 1b11f35ca10275cd66847af3d5c85d7d13900d41
Author: Oscar Moll <orm@csail.mit.edu>
Date:   Thu Oct 12 13:57:24 2023 -0400

    extend syntax to allow alias in select

PR #34, opened 2023-11-03
use pyav to encode straight to h264, remove docker ffmpeg dep.

orm011:
…ove ffmpeg docker dep.

The larger goal is to move toward not requiring docker. there's a separate, orthogonal pull request that enables this for postgres.

In this case, pyav is a pip installable python binding for libav.
As a side effect of using this, you also dont need to render the video twice to get around the licensing issue you mentioned.

videos look fine.


pixeltable/functions/__init__.py
```diff
@@ -102,31 +102,34 @@ def value(self) -> float:
 
 class VideoAggregator:
     def __init__(self):
-        self.video_writer = None
-        self.size = None
+        '''follows https://pyav.org/docs/develop/cookbook/numpy.html#generating-video'''
+        self.container = None
+        self.stream = None
+        self.fps = 24
```
mkornacker: We're using fps=25 at the moment, so let's stick with that.

pixeltable/functions/__init__.py
```diff
@@ -102,31 +102,34 @@ def value(self) -> float:
 
 class VideoAggregator:
     def __init__(self):
-        self.video_writer = None
-        self.size = None
+        '''follows https://pyav.org/docs/develop/cookbook/numpy.html#generating-video'''
```
mkornacker: """

pixeltable/functions/__init__.py
```diff
@@ -102,31 +102,34 @@ def value(self) -> float:
...
+        self.container = None
```
mkornacker: missing typing.

pixeltable/functions/__init__.py
```diff
@@ -102,31 +102,34 @@ def value(self) -> float:
...
-            self.video_writer = cv2.VideoWriter(str(self.tmp_file), cv2.VideoWriter_fourcc(*'mp4v'), 25, self.size)
-
-        frame_array = np.array(frame)
-        frame_array = cv2.cvtColor(frame_array, cv2.COLOR_RGB2BGR)
-        self.video_writer.write(frame_array)
-
+            self.container = av.open(str(self.out_file), mode="w")
+            self.stream = self.container.add_stream("mpeg4", rate=self.fps)
+            self.stream.width = frame.width
+            self.stream.height = frame.height
```
mkornacker, line 122: do you need stream.pix_fmt = "yuv420p" anywhere?

mkornacker (commented):
Have you run the tutorial notebooks by hand to see if everything works?

pixeltable/functions/__init__.py
orm011: Doesn't seem like it. The generated videos seem to have yuv420p either way.

orm011:
> Have you run the tutorial notebooks by hand to see if everything works?

I ran the 'Comparing Object Detection Models for Video.ipynb'. 
That one does work (the generated video is available and looks ok).

There is currently a bug in master though, where the .show() does not seem to include the url for the video in the video tag. 
That's a separate issue though

pixeltable/functions/__init__.py
```diff
@@ -102,29 +103,32 @@ def value(self) -> float:
 
 class VideoAggregator:
     def __init__(self):
-        self.video_writer = None
-        self.size = None
+        """follows https://pyav.org/docs/develop/cookbook/numpy.html#generating-video"""
+        self.container : av.container.OutputContainer = None
```
mkornacker: self.container: Optional[...] = None

pixeltable/functions/__init__.py
```diff
@@ -102,29 +103,32 @@ def value(self) -> float:
...
+        self.stream : av.stream.Stream = None
```
mkornacker: same

mkornacker (approved):

PR #36, opened 2023-11-09
to pytorch

orm011:


orm011:
NB some questions about how to deal with other types (arrays). hasn't been tested on that. but should be ok for images.

pixeltable/dataframe.py
```diff
@@ -454,3 +465,54 @@ def __getitem__(self, index: object) -> DataFrame:
...
+        assert self.group_by_clause == [] or self.group_by_clause is None, 'non trivial group_by not yet supported in this method'
+        assert self.order_by_clause == [] or self.order_by_clause is None, 'non trivial order_by not yet supported in this method'
+
+        d = {
+            'type': 'DataFrame',
+            'tbl_id': str(self.tbl.id),
+            'tbl_version': self.tbl.version,
+            'select_list': [(e.as_dict(), name) for (e, name) in self.select_list] if self.select_list is not None else None,
+            'where_clause': self.where_clause.as_dict() if self.where_clause is not None else None,
+            'group_by_clause': None,
```
mkornacker: Or simply implement that here? It's a one-liner, no?

pixeltable/dataframe.py
```diff
@@ -454,3 +465,54 @@ def __getitem__(self, index: object) -> DataFrame:
...
+            'order_by_clause': None,
+            'column_types': {name: col_type.as_dict() for (name, col_type) in zip(self.get_column_names(), self.get_column_types())}
```
mkornacker: I think this is redundant, because you're already capturing the table id and version (which subsumes the schema version). If so, remove. If not, explain in a comment.

pixeltable/dataframe.py
```diff
@@ -454,3 +465,54 @@ def __getitem__(self, index: object) -> DataFrame:
...
+                        'pt' means image columns return as a CxHxW tensor with values in [0,1] and type torch.float32.
+                            (the format output by torchvision.transforms.ToTensor())
+            Returns:
+                A pytorch IterableDataset: Columns become fields of the dataset, 
+                where rows are returned as a dictionary compatible with torch.utils.data.DataLoader default collation.
+        """
+        from pixeltable.utils.parquet import save_parquet # pylint: disable=import-outside-toplevel
+        from pixeltable.utils.pytorch import PixeltablePytorchDataset # pylint: disable=import-outside-toplevel
+
+        mstr = json.dumps(self.as_dict(), sort_keys=True)
```
mkornacker: what does mstr stand for?

also, sort_keys=True is needed for correctness, would be worth a quick comment

pixeltable/dataframe.py
```diff
@@ -454,3 +465,54 @@ def __getitem__(self, index: object) -> DataFrame:
...
+            Convert the dataframe to a pytorch IterableDataset suitable for parallel loading
+             with torch.utils.data.DataLoader. 
+
+            This method serializes data so it can be read from disk efficiently, repeatedly without 
+            re-executing the query. This data is cached to disk for future re-use.
+
+            Args:
+                image_format: format of the images. Can be 'pt' (pytorch tensor) or 'np' (numpy array).
+                        'np' means image columns return as an RGB uint8 array of shape HxWxC.
+                        'pt' means image columns return as a CxHxW tensor with values in [0,1] and type torch.float32.
```
mkornacker, line 503: regarding 'pt': I initially assumed this referred to pixeltable somehow, because I mostly do (and would expect others to do the same)
import pixeltable as pt

Should we refer to pytorch as 'torch' throughout? Or does that go against common practices? I always see 'import torch' in tutorials and such.

pixeltable/type_system.py
```diff
@@ -409,6 +417,10 @@ def to_sql(self) -> str:
 
     def to_sa_type(self) -> str:
         return sql.String
+    
+    def to_arrow_type(self) -> Any:
```
mkornacker: -> pyarrow.DataType

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +112,94 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+
+        t = img_tbl
+        tsize = t.count()
+        assert tsize > 0
+
+        # crop images to uniform size to test tensor stacking on the loader side
+        # H = 224, W = 256
+        # left top right bottom
+        # convert so they are all 3 channel RGB
+        dfbase : DataFrame = t.select(t.category, t.row_id,
```
mkornacker: Let's leave out hints in assignments.

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +112,94 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+        import torch.utils.data
+        import numpy as np
+        import pickle
+
+        t = img_tbl
+        tsize = t.count()
+        assert tsize > 0
+
+        # crop images to uniform size to test tensor stacking on the loader side
+        # H = 224, W = 256
```
mkornacker: uncomment and then use those constants below?

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +112,94 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+        long_size = dflong.count()
+
+        cols = dflong.get_column_names()
+        
+        assert long_size > 0
+        assert long_size < tsize #check non-trivial selection
+
+        dslong = dflong.to_pytorch_dataset(image_format='np')
+        dslongpt = dflong.to_pytorch_dataset(image_format='pt')
+        elts = []
```
mkornacker: unused

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +112,94 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+            assert arr.dtype == torch.float32
+            assert (0.0 <= arr).all()
+            assert (arr <= 1.0).all()
+
+            elts.append(elt)
+
+        assert len(elts) == long_size
+
+        ## now test interaction with dataloader
+        def _test_loader(ds, size : int, **kwargs):
```
mkornacker: type hint missing

pixeltable/dataframe.py
```diff
@@ -454,3 +465,54 @@ def __getitem__(self, index: object) -> DataFrame:
...
+        """
+            Convert the dataframe to a pytorch IterableDataset suitable for parallel loading
+             with torch.utils.data.DataLoader. 
+
+            This method serializes data so it can be read from disk efficiently, repeatedly without 
+            re-executing the query. This data is cached to disk for future re-use.
+
+            Args:
+                image_format: format of the images. Can be 'pt' (pytorch tensor) or 'np' (numpy array).
+                        'np' means image columns return as an RGB uint8 array of shape HxWxC.
```
mkornacker, line 502: what's the use case for non-torch tensors in a torch dataset? and why the different data types?

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +112,94 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+            dl = torch.utils.data.DataLoader(ds, **kwargs)
+
+            loaded_ids = set()
+            for b in dl:
```
mkornacker: b?

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+        Internal method to stream dataframe data to parquet format.
+        Does not materialize the dataset to memory.
+
+        It preserves pixeltable type metadata in a json file, which would otherwise 
+        not be available in the parquet format.
+
+        Images are stored inline in a compressed format in their parquet file.
+
+        NB. pixeltable types
```
mkornacker: 'NB': remove (if it's not worth noting it shouldn't be included anyway)

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+    column_types = df.get_column_types()
+
+    arrow_schema = pa.schema([pa.field(name, column_types[i].to_arrow_type()) for i, name in enumerate(column_names)])
+    # store the changes atomically
+    with transactional_folder(dest_path) as temp_path:
+        # dump metadata json file so we can inspect what was the source of the parquet file later on.
+        json.dump(df.as_dict(), (temp_path / '.pixeltable.json').open('w')) # pylint: disable=protected-access
+
+        batch_num = 0
+        current_value_batch : Dict[str, deque] = {k:deque() for k in column_names}
```
mkornacker, line 53: It'll get fixed when we have the auto-formatter in place, but in the meantime, could you just change everything, incl. type hints, to be like 'a<no space>:<space>...'?

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+            stacked_arr = np.stack(value_batch[field.name])
+            pydict[field.name] = pa.FixedShapeTensorArray.from_numpy_ndarray(stacked_arr)
+        else:
+            pydict[field.name] = value_batch[field.name]
+
+    tab = pa.Table.from_pydict(pydict, schema=schema)
+    pq.write_table(tab, output_path)
+
+
+def save_parquet(df : DataFrame, dest_path : Path, partition_size_bytes : int = 100_000_000):
```
mkornacker: 100MB would be 100 * 1024 * 1024

return type missing

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+
+        It preserves pixeltable type metadata in a json file, which would otherwise 
+        not be available in the parquet format.
+
+        Images are stored inline in a compressed format in their parquet file.
+
+        NB. pixeltable types
+            - string ok
+            - image ok (inlined)
+            - video ok (path only)
```
mkornacker: what happens with the other data types? at least int/float/bool should work, no?

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+    
+        Args:
+            partition_size_bytes: maximum target size for each chunk. Default 100MB.
+        Returns:
```
mkornacker: returns None

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+                        val.save(buf, format='PNG')
+                        val = buf.getvalue()
+                    else:
+                        assert False, f'unknown image type {type(val)}'
+                    length = len(val)
+                elif col_type.is_string_type():
+                    length = len(val)
+                elif col_type.is_video_type():
+                    if data_row.file_paths is not None and data_row.file_paths[e.slot_idx] is not None: 
+                        # if there is a file, read directly to preserve information
```
mkornacker: you're not reading

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+                        val = data_row.file_paths[e.slot_idx]
+                    else:
+                        assert False, f'unknown video type {type(val)}'
```
mkornacker, line 83: confusing error message, the type doesn't really matter here.

does it even make sense to include file names directly? should 'video' not be a supported type for the purpose of pytorch conversion?

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+                    length = len(val)
+                elif col_type.is_json_type():
+                    val = json.dumps(val)
+                    length = len(val)
+                elif col_type.is_array_type():
+                    length = val.nbytes
+                elif col_type.is_int_type():
+                    length = 8
+                elif col_type.is_float_type():
+                    length = 4
```
mkornacker: are python floats not 8 bytes by default?

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+            if current_byte_estimate > partition_size_bytes:
+                assert batch_num < 100_000, 'wrote too many parquet files, unclear ordering'
+                _write_batch(current_value_batch, arrow_schema,  temp_path / f'part-{batch_num:05d}.parquet')
+                batch_num += 1
+                current_value_batch = {k:deque() for k in column_names}
+                current_byte_estimate = 0
+
+        _write_batch(current_value_batch, arrow_schema, temp_path / f'part-{batch_num:05d}.parquet')
+
+def get_part_metadata(path : Path) -> List[Tuple[str, int]]:
```
mkornacker, line 112: what does 'part' mean here?

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,132 @@
...
+from collections import deque
+
+from pixeltable.dataframe import DataFrame
+from typing import List, Tuple, Any, Dict
+
+from pixeltable.utils.transactional_folder import transactional_folder
+
+
+def _write_batch(value_batch : Dict[str, deque], schema : pa.Schema, output_path : Path) -> None:
+    pydict = {}
```
mkornacker, line 14: type hint (when unclear from initialization)

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+from pixeltable.type_system import ColumnType
+from pixeltable.utils.parquet import get_part_metadata
+
+def _cumsum(lst):
+    acc = [0]
+    for x in lst:
+        acc.append(acc[-1] + x)
+    return acc
+
+class PixeltablePytorchDataset(torch.utils.data.IterableDataset):
```
mkornacker, line 23: Do we need to include 'Pixeltable' here? It's a very unwieldy name.

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+    NB. This class must inherit from torch.utils.data.IterableDataset for it 
+    to work with torch.utils.data.DataLoader.
+    """
+    def __init__(
+        self,
+        folder: Path,
+        image_format: str,
+    ):
+        """
+        Args:
```
mkornacker, line 35: doesn't match the actual parameters

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+            column_types: dict mapping column names to ColumnType
+            parquet_path: path to directory containing parquet files
+            image_format: 'np' or 'pt'. 'np' is RGB uint8 array,
+                        'pt' is result of torchvision.transforms.ToTensor()
+        """
+        super().__init__()
+
+        json_path = folder / '.pixeltable.json'
```
mkornacker: if we're using this file for more than just debugging, we should give it a more descriptive name

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+
+class PixeltablePytorchDataset(torch.utils.data.IterableDataset):
+    """
+    PyTorch dataset interface for pixeltable data. 
+    NB. This class must inherit from torch.utils.data.IterableDataset for it 
+    to work with torch.utils.data.DataLoader.
+    """
+    def __init__(
+        self,
+        folder: Path,
```
mkornacker: we use 'dir' or 'directory' elsewhere, let's stick with that for the sake of consistency

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+import math
+from pathlib import Path
+import PIL.Image
+import json
+from typing import Dict, Tuple, Generator, Any
+
+from pixeltable.type_system import ColumnType
+from pixeltable.utils.parquet import get_part_metadata
+
+def _cumsum(lst):
```
mkornacker, line 17: type hints

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+        else:
+            assert False, f"unknown image repr {self.image_format}"
+
+    def __iter__(self) -> Generator[Dict[str, Any], None, None]:
+        worker_info = torch.utils.data.get_worker_info()
+        if worker_info is None:  # single-process data loading, return the full iterator
+            start_row = 0
+            end_row = self._totals[-1]
+        else:  # in a worker process
+            num_workers = [
```
mkornacker, line 95: isn't this rows per worker? the code below is confusing when taking the name at face value.

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+                math.floor(self._totals[-1] / float(worker_info.num_workers))
+                for _ in range(worker_info.num_workers)
+            ]
+            assert self._totals[-1] - sum(num_workers) < worker_info.num_workers
+            for i in range(self._totals[-1] - sum(num_workers)):
```
mkornacker, line 100: not completely obvious at first glance what's happening here; prefix with brief comment ('distribute remaining rows'?)

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+        with json_path.open() as f:
+            json_data = json.load(f)
+        
+        column_types = json_data['column_types']
+        self.column_types = {k: ColumnType.from_dict(v) for k, v in column_types.items()}
+        self.pqpath = folder
+        self.image_format = image_format
+        assert image_format in ["np", "pt"]
+        self.part_metadata = get_part_metadata(self.pqpath)
+        self._totals = _cumsum([x[1] for x in self.part_metadata])
```
mkornacker, line 51: not very descriptive (totals of what?)

also: it has #partitions + 1 values, which is somewhat subtle

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+            start_rows = _cumsum(num_workers)
+            start_row = start_rows[worker_info.id]
+            end_row = start_rows[worker_info.id + 1]
+
+        if start_row == end_row:
+            return iter([])  # type: ignore
+        else:
+            return self._iter_range(start_row, end_row)
+
+    def _iter_range(self, start_row, end_row) -> Generator[Dict[str, Any], None, None]:
```
mkornacker: type hints

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+        self.column_types = {k: ColumnType.from_dict(v) for k, v in column_types.items()}
+        self.pqpath = folder
+        self.image_format = image_format
+        assert image_format in ["np", "pt"]
+        self.part_metadata = get_part_metadata(self.pqpath)
+        self._totals = _cumsum([x[1] for x in self.part_metadata])
+
+    def _get_start_position(self, row_start: int) -> Tuple[int, int]:
+        """
+        Returns the starting block and row within the block, based sizes of each part
```
mkornacker: block == file?

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+            prev_acc = acc
+
+        assert False, "unreachable"
+
+    def _unmarshall(self, k: str, v: Any) -> Any:
+        if not self.column_types[k].is_image_type():
+            return v
+
+        assert isinstance(v, bytes)
+        pil = PIL.Image.open(io.BytesIO(v))
```
mkornacker: let's not use variable names that might conflict with module names at some point in the future, you'd get some really weird/subtle bugs

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,139 @@
...
+
+        if self.image_format == "np":
+            import numpy as np  # pylint: disable = import-outside-toplevel
+            return np.array(pil)
+        elif self.image_format == "pt":
+            import torchvision  # pylint: disable = import-outside-toplevel
+            return torchvision.transforms.ToTensor()(pil)
```
mkornacker: is there some way to save directly in the correct format and avoid the transformation?

pixeltable/utils/transactional_folder.py
```diff
@@ -0,0 +1,34 @@
+from contextlib import contextmanager
+import pixeltable.exceptions as exc
+from typing import Any, Generator
+from pathlib import Path
+import shutil
+
+
+@contextmanager
+def transactional_folder(folder_path: Path) -> Generator[Path, Any, Any]:
```
mkornacker: I didn't understand that this was a temp directory until I read the code.

Also, let's stick with 'directory' instead of folder, pathlib also talks about directories.

mkornacker (changes requested):

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +112,94 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
         t = test_tbl
         res = t.select(1.0).where(t.c2 < 10).show(0)
         assert res.rows == [[1.0]] * 10
+
```
mkornacker, line 119: missing test cases:
- using the same dataset twice in a row uses the cache
- adding a row to the table invalidates the cached version
- changing the select list invalidates the cached version

pixeltable/dataframe.py
orm011: right, i see.
Huggingface, which tries to be multi-library, has a convention over tensor formats. 'pt' is pytorch, 'tf' tensorflow, 'np' numpy, etc, mostly I was re-using this. 

I believe its clear to most non-pixeltable developers. 
to other people it may not work well. maybe we should use 'pxt' or something like that if we ever have
a pixeltable specific. 

In terms of import pixeltable as pt, probably ok to just use pt for pixeltable. or we should use px (still 2 chars but less likely to be misunderstood by the pytorch world).

pixeltable/utils/pytorch.py
orm011: The main issue right now is that the only reasonable way to 
preserve images that I can think of is a compressed format ('png' or maybe lossy).

The storage blow-up of saving images as rbg in float32 could be awful
The Tensor transform is relatively cheap.

We can revisit this. Transforms generally are something we need to think about.

mkornacker: Ah yes, good point.

pixeltable/dataframe.py
orm011: While Torch dataloader output batches are generally torch tensors, Torch dataset is a general class, it is not uncommon to get numpy stuff out of them, and torch dataloader  does not assume everything coming in to it is a torch tensor: It handles numpy just as well. (more accurately, this is part of what the collate_fn argument does), so this gives us some flexibility. In terms of why I return them: 

For image types,  numpy rep of images can be easily converted to PIL.Image, allowing  easy inspection of the data.  This is one big reason to keep them.  Pytorch image tensors aren't just different  in type, they have their own conventions on channel ordering and scaling.  This pytorch representation is ok when passing it to a network, but not good if you want to view the image or if you wanted something non-standard.  A second reason to keep numpy is that up until you make it a numpy array, all conversions can be lossless. so it is easy to check can check values are the same the user put in. Once you convert to pytorch tensor it is not clear the conversion was lossless and it would be harder to test.

For array types, it makes sense to pass np array because that is our native image representation, and pyarrow also understands this format: If we convert to pytorch ourselves then we need to write tests checking the right conversions have happened, and that we haven't lost information.  So making everything pytorch seems to add complexity, but for little in return, because most downstream classes including torch dataloader are ok with numpy arrays and handle conversion on their own. By keeping them in numpy neither us nor users need to understand implicit conversions from our end: the values should be identical to what they saved.

pixeltable/dataframe.py
```diff
@@ -305,17 +313,18 @@ def categorical_map(self) -> Dict[str, int]:
             return result
 
     def _description(self) -> pd.DataFrame:
-        """Return a description of this DataFrame as a pandas DataFrame.
-        The DataFrame has two columns, heading and info, which list the contents of each 'component'
-        (select list, where clause, ...) vertically when printed.
+        """
```
mkornacker: Can you move this comment to the user-facing function, so that it shows up in the auto-generated API docs?

pixeltable/dataframe.py
```diff
@@ -454,3 +463,62 @@ def __getitem__(self, index: object) -> DataFrame:
         if isinstance(index, list):
             return self.select(*index)
         raise TypeError(f'Invalid index type: {type(index)}')
+    
+    def as_dict(self) -> Dict[str, Any]:
```
mkornacker: This doesn't need to be user-facing, no? _as_dict()? Otherwise it shows up in the generated API docs.

pixeltable/dataframe.py
```diff
@@ -454,3 +463,62 @@ def __getitem__(self, index: object) -> DataFrame:
...
+            'limit_val': self.limit_val,
+        }
+        return d
+
+    def to_pytorch_dataset(self, image_format : str = 'pt') -> 'torch.utils.data.IterableDataset':
+        """
+            Convert the dataframe to a pytorch IterableDataset suitable for parallel loading
+             with torch.utils.data.DataLoader. 
+
+            This method serializes data so it can be read from disk efficiently, repeatedly without
```
mkornacker: efficiently and repeatedly

pixeltable/dataframe.py
```diff
@@ -454,3 +463,62 @@ def __getitem__(self, index: object) -> DataFrame:
...
+                        'np' means image columns return as an RGB uint8 array of shape HxWxC.
+                        'pt' means image columns return as a CxHxW tensor with values in [0,1] and type torch.float32.
+                            (the format output by torchvision.transforms.ToTensor())
+            Returns:
+                A pytorch IterableDataset: Columns become fields of the dataset, 
+                where rows are returned as a dictionary compatible with torch.utils.data.DataLoader default collation.
+
+            Constraints:
+            1. Limited support for null values:
+                  Null array values are not currently supported by this method, these will trigger an error in this method.
```
mkornacker: This is literally a two-line fix in DataRow.__getitem__(). Make it in this PR? This is really a bug, not a constraint.

pixeltable/tests/conftest.py
```diff
@@ -75,6 +75,10 @@ def test_tbl_exprs(test_tbl: catalog.Table, test_stored_fn: pt.Function) -> List
         test_stored_fn(t.c2),
     ]
 
+@pytest.fixture(scope='function')
+def all_datatype_tbl(test_client: pt.Client) -> catalog.Table:
```
mkornacker, line 79: Let's call it all_dataypes_tbl() to be more grammatically correct.

pixeltable/tests/utils.py
```diff
@@ -158,6 +159,30 @@ def create_test_tbl(client: pt.Client) -> catalog.Table:
...
+        catalog.Column('c_float', FloatType(nullable=True)),
+        catalog.Column('c_image', ImageType(nullable=True)),
+        catalog.Column('c_int', IntType(nullable=True)),
+        catalog.Column('c_json', JsonType(nullable=True)),
+        catalog.Column('c_string', StringType(nullable=True)),
+        catalog.Column('c_timestamp', TimestampType(nullable=True)),
+        catalog.Column('c_video', VideoType(nullable=True)),
+    ]
+    tbl = test_client.create_table('all_datatype_tbl', cols)
+    example_rows = create_table_data(tbl, num_rows=11)
```
mkornacker, line 178: why 11?

pixeltable/utils/parquet.py
```diff
@@ -0,0 +1,126 @@
...
+                        # if no file available, eg. bc it is computed, convert to png
+                        buf = io.BytesIO()
+                        val.save(buf, format='PNG')
+                        val = buf.getvalue()
+                    else:
+                        assert False, f'unknown image type {type(val)}'
+                    length = len(val)
+                elif col_type.is_string_type():
+                    length = len(val)
+                elif col_type.is_video_type():
```
mkornacker, line 79: Let's remove support for video values, they don't really make sense. If you care about local file paths, you can simply select(t.video_col.localpath), which is a string value and supported. If you do select(t.video_col), you're not even guaranteed to get a value, even if you have a valid file url in there (it may not have been downloaded yet).

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +115,163 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+            3. changing the select list invalidates the cached version
+        """
+        t = all_datatype_tbl
+
+        t.drop_column('c_video') # null value video column triggers internal assertions in DataRow
+        # see https://github.com/mkornacker/pixeltable/issues/38
+
+        t.drop_column('c_array') # no support yet for null array values in the pytorch dataset
+
+        def _get_mtimes(dir: Path):
```
mkornacker, line 266: missing return type

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +115,163 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+
+        # check invalidation on insert
+        t_size = t.count()
+        t.insert([[t_size, ]], columns=['row_id'])
+        ds3 = t.to_pytorch_dataset(image_format='pt')
+        assert ds3.path != ds1.path, 'different path should be used'
+
+        # check select list invalidation
+        ds4 = t.select(t.row_id).to_pytorch_dataset(image_format='pt')
+        assert ds4.path != ds1.path, 'different select list, hence different path should be used'
```
mkornacker: You need to compare to ds3 here, not ds1

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def _iter_range(self, start_row : int, end_row : int) -> Generator[Dict[str, Any], None, None]:
+        (part_no, iter_start) = self._get_start_position(start_row)
+        total = end_row - start_row
+
+        acc = 0
+        part_pos = part_no
+        iter_pos = iter_start
+
+        def _to_column_dict(tab : pa.Table) -> dict[str, Any]:
```
mkornacker: We use Dict everywhere.

pixeltable/utils/pytorch.py
```diff
@@ -0,0 +1,156 @@
...
+            column_dict = {}
```
mkornacker, line 122: type hint missing

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +115,163 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+        df = t.select(
+            t.row_id,
+            t.c_image,
+            c_image_xformed=t.c_image.resize([220, 224]).convert('RGB')
+        ).where(t.row_id < 1)
+
+        ds = df.to_pytorch_dataset(image_format='np')
+        ds_ptformat = df.to_pytorch_dataset(image_format='pt')
+
+        elts = []
```
mkornacker: This is kind of a write-only variable, aside from the > 0 check at the end. Remove and replace with a counter?

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +115,163 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+
+            arr_pt = elt_pt['c_image']
+            assert torch.is_tensor(arr_pt)
+
+            arr_pt = elt_pt['c_image_xformed']
+            assert torch.is_tensor(arr_pt)
+            assert arr_pt.shape == (3, 224, 220)
+            assert arr_pt.dtype == torch.float32
+            assert (0.0 <= arr_pt).all()
+            assert (arr_pt <= 1.0).all()
```
mkornacker, line 192: Is there some way to check that the c_image and c_image_xformed values are "the same" (within some epsilon)? If it's difficult, leave to-do. If it's easy, add it.

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +115,163 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+        x = pickle.dumps(ds)
+        _ = pickle.loads(x)
+
+        # test we get all rows
+        def check_recover_all_rows(ds, size : int, **kwargs):
+            dl = torch.utils.data.DataLoader(ds, **kwargs)
+            loaded_ids = set()
+            for b in dl:
+                for row_id in b['row_id']:
+                    val = int(row_id)
```
mkornacker: comment: why do you need cast?

pixeltable/tests/test_dataframe.py
```diff
@@ -110,3 +115,163 @@ def test_select_literal(self, test_tbl: catalog.Table) -> None:
...
+
+        # check different number of workers
+        check_recover_all_rows(ds, size=df_size, batch_size=3, num_workers=0) # within this process
+        check_recover_all_rows(ds, size=df_size, batch_size=3, num_workers=2) # two separate processes
+
+        # check edge case where some workers get no rows
+        short_size = 1
+        df_short = df.where(t.row_id < short_size)
+        ds_short = df_short.to_pytorch_dataset(image_format='pt')
+        check_recover_all_rows(ds_short, size=short_size, batch_size=13, num_workers=short_size+1)
```
mkornacker, line 251: 13?

mkornacker (changes requested):
There were also some remaining unaddressed comments from the last round.

PR #37, opened 2023-11-09
Update __init__.py

orm011:
use h264 instead of mpeg4, checked browser can actually play it after making that change.


mkornacker:
Let's change the description to "use h264 instead of mpeg4".

PR #49, opened 2023-12-09
Dataset shuffle

orm011:
Opening for comments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/49)

### DIFF
--- a/pixeltable/dataframe.py
+++ b/pixeltable/dataframe.py
@@ -579,7 +579,7 @@ class DataFrame:
         else:
             return write_coco_dataset(self, dest_path)
 
-    def to_pytorch_dataset(self, image_format: str = 'pt') -> 'torch.utils.data.IterableDataset':
+    def to_pytorch_dataset(self, image_format: str = 'pt', shuffle : bool = True) -> 'torch.utils.data.IterableDataset':
         """
         Convert the dataframe to a pytorch IterableDataset suitable for parallel loading
         with torch.utils.data.DataLoader.
@@ -624,13 +624,20 @@ class DataFrame:
         from pixeltable.utils.parquet import save_parquet # pylint: disable=import-outside-toplevel
         from pixeltable.utils.pytorch import PixeltablePytorchDataset # pylint: disable=import-outside-toplevel
 
-        summary_string = json.dumps(self._as_dict()) 
+        if shuffle:
+            if not self.order_by_clause is None:
+                raise exc.Error('order_by() cannot be used with shuffle')
+            df = self.order_by(exprs.SqlRandom(), asc=True)
+        else:
+            df = self
+
+        summary_string = json.dumps(df._as_dict())
         cache_key = hashlib.sha256(summary_string.encode()).hexdigest()
     
         dest_path = (Env.get().dataset_cache_dir / f'df_{cache_key}').with_suffix('.parquet') # pylint: disable = protected-access
         if dest_path.exists(): # fast path: use cache
             assert dest_path.is_dir()
         else:
-            save_parquet(self, dest_path)
+            save_parquet(df, dest_path)
 
         return PixeltablePytorchDataset(path=dest_path, image_format=image_format)
--- a/pixeltable/tests/test_dataframe.py
+++ b/pixeltable/tests/test_dataframe.py
@@ -6,6 +6,8 @@ import numpy as np
 from pathlib import Path
 import bs4
 import requests
+import shutil
+import math
 
 from pycocotools.coco import COCO
 
@@ -33,7 +35,7 @@ class TestDataFrame:
         assert res1 == res4
 
         _ = t.where(t.c2 < 10).select(t.c2, t.c2).show(0) # repeated name no error
-        
+
         # duplicate select list
         with pytest.raises(exc.Error) as exc_info:
             _ = t.select(t.c1).select(t.c2).show(0)
@@ -216,7 +218,7 @@ class TestDataFrame:
         for tup in ds:
             for col in df.get_column_names():
                 assert col in tup
-        
+
             arrval = tup['c_array']
             assert isinstance(arrval, np.ndarray)
             col_type = type_dict['c_array']
@@ -232,6 +234,37 @@ class TestDataFrame:
             assert torch.is_tensor(tup['c_image'])
             assert isinstance(tup['c_video'], str)
             assert isinstance(tup['c_json'], dict)
+
+    def test_to_pytorch_dataset_shuffle(self, all_datatypes_tbl: catalog.Table) -> None:
+        """ Test shuffle flag is honored """
+        t = all_datatypes_tbl
+        assert t.count() > 20
+        # using a large number so that shuffle is unlikely to result in same input order
+        # 1/ 20! chance of spurious failure
+        df = t.select(t.row_id)
+
+        # shuffle is true by default. check that the order is different
+        ds = df.to_pytorch_dataset()
+        def get_ids_ordered(ds):
+            return [tup['row_id'] for tup in ds]
+        order1 = get_ids_ordered(ds)
+
+        # delete the cache and generate dataset again. order should be different
+        shutil.rmtree(str(ds.path))
+        ds2 = df.to_pytorch_dataset()
+        order2 = get_ids_ordered(ds2)
+        assert set(order1) == set(order2)
+        assert order1 != order2
+
+        # check order is respected
+        ds_ordered = df.order_by(t.row_id).to_pytorch_dataset(shuffle=False)
+        order3 = get_ids_ordered(ds_ordered)
+        assert order3 == list(range(df.count()))
+
+        # check error when order by is used with shuffle
+        with pytest.raises(exc.Error) as exc_info:
+            _ = df.order_by(t.row_id).to_pytorch_dataset()
+        assert 'order_by() cannot be used with shuffle' in str(exc_info.value)
 
     def test_to_pytorch_image_format(self, all_datatypes_tbl: catalog.Table) -> None:
         """ tests the image_format parameter is honored
@@ -294,11 +327,15 @@ class TestDataFrame:
         """
         import torch.utils.data
         @pt.udf(param_types=[pt.JsonType()], return_type=pt.JsonType())
-        def restrict_json_for_default_collate(obj):
+        def restrict_json_for_default_collate(obj) -> Dict[str,Any]:
             keys = ['id', 'label', 'iscrowd', 'bounding_box']
             return {k: obj[k] for k in keys}
-        
         t = all_datatypes_tbl
+        num_rows = 11 # prime number larger than batch size
+        batch_size = 3 # relatively prime to num_rows
+        assert math.gcd(num_rows, batch_size) == 1
+
+        # test every type is compatible with default collate_fn
         df = t.select(
             t.row_id,
             t.c_int,
@@ -312,9 +349,12 @@ class TestDataFrame:
             c_json = restrict_json_for_default_collate(t.c_json.detections[0]),
             # images must be uniform shape for pytorch collate_fn to not fail
             c_image=t.c_image.resize([220, 224]).convert('RGB')
-        )
+        ).where(t.row_id < num_rows)
+        # using a prime number of rows to ensure not fully divisible by batch size
         df_size = df.count()
+        assert df_size == num_rows
         ds = df.to_pytorch_dataset(image_format='pt')
+
         # test serialization:
         #  - pickle.dumps() and pickle.loads() must work so that
         #   we can use num_workers > 0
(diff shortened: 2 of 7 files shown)


pixeltable/dataframe.py
```diff
@@ -580,13 +580,19 @@ def to_pytorch_dataset(self, image_format : str = 'pt') -> 'torch.utils.data.Ite
         from pixeltable.utils.parquet import save_parquet # pylint: disable=import-outside-toplevel
         from pixeltable.utils.pytorch import PixeltablePytorchDataset # pylint: disable=import-outside-toplevel
 
-        summary_string = json.dumps(self._as_dict()) 
+        if shuffle:
+            assert self.order_by_clause is None or self.order_by_clause == [], 'order_by() cannot be used with shuffle=True'
```
mkornacker: we should never have an empty order by clause

pixeltable/tests/test_dataframe.py
```diff
@@ -285,7 +285,8 @@ def restrict_json_for_default_collate(obj):
             c_image=t.c_image.resize([220, 224]).convert('RGB')
         )
         df_size = df.count()
-        ds = df.to_pytorch_dataset(image_format='pt')
+        assert df_size >= 23, 'want to test enough tuples'
```
mkornacker: 23, huh?

pixeltable/tests/test_dataframe.py
```diff
@@ -295,18 +296,26 @@ def restrict_json_for_default_collate(obj):
         # test we get all rows
         def check_recover_all_rows(ds, size : int, **kwargs):
```
mkornacker, line 365: return type missing

pixeltable/tests/test_dataframe.py
```diff
@@ -285,7 +285,8 @@ def restrict_json_for_default_collate(obj):
             c_image=t.c_image.resize([220, 224]).convert('RGB')
         )
         df_size = df.count()
-        ds = df.to_pytorch_dataset(image_format='pt')
+        assert df_size >= 23, 'want to test enough tuples'
+        ds = df.to_pytorch_dataset(image_format='pt')
```
mkornacker: since you're explicitly testing the shuffling, you should supply the parameter explicitly as well

pixeltable/tests/test_dataframe.py
```diff
@@ -295,18 +296,26 @@ def restrict_json_for_default_collate(obj):
...
-            assert loaded_ids == set(range(size))
+            assert set(loaded_ids) == set(range(size))
+            return loaded_ids
+
+        loaded_ids = check_recover_all_rows(ds, size=df_size, batch_size=3, num_workers=0) # within this process
+        assert loaded_ids != list(range(df_size)) # check ordering is not preserved
+
+        ds_notshuffled = df.to_pytorch_dataset(image_format='pt', shuffle=False)
+        loaded_ids = check_recover_all_rows(ds_notshuffled, size=df_size, batch_size=3, num_workers=0)
+        assert loaded_ids == list(range(df_size)) # check ordering is preserved
```
mkornacker: This isn't guaranteed to work, because the ordering might change, depending on what Postgres decides. In this case, you should supply an explicit order_by().

pixeltable/tests/test_dataframe.py
```diff
@@ -295,18 +296,26 @@ def restrict_json_for_default_collate(obj):
...
 
         # check different number of workers
-        check_recover_all_rows(ds, size=df_size, batch_size=3, num_workers=0) # within this process
-        check_recover_all_rows(ds, size=df_size, batch_size=3, num_workers=2) # two separate processes
+        loaded_ids = check_recover_all_rows(ds, size=df_size, batch_size=3, num_workers=2) # two separate processes
+        assert loaded_ids != list(range(df_size)) # check ordering is not preserved
```
mkornacker: missing: check for failure of order_by() + shuffle

mkornacker (changes requested):

pixeltable/tests/test_dataframe.py
orm011: I'm also testing that the default argument is indeed shuffle, which we wouldn't notice if changed if we explicitly test positive.

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/mkornacker/pixeltable/49)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/dataframe.py` line 584 at r1](https://reviewable.io/reviews/mkornacker/pixeltable/49#gh-1421825348:-NoTtB1eB_P9Sd3NhYR0:b-896fix) ([raw file](https://github.com/mkornacker/pixeltable/blob/f0032e86f3ea58194bbdbb7703343a0bcf81e9b8/pixeltable/dataframe.py#L584)):*

we should never have an empty order by clause
</blockquote></details>

Done.
___
*[`pixeltable/tests/test_dataframe.py` line 288 at r1](https://reviewable.io/reviews/mkornacker/pixeltable/49#gh-1421826540:-NoTzWpY92MIhtcg6m8a:b-7h91p9) ([raw file](https://github.com/mkornacker/pixeltable/blob/f0032e86f3ea58194bbdbb7703343a0bcf81e9b8/pixeltable/tests/test_dataframe.py#L288)):*

23, huh?
</blockquote></details>

added comment. prime number plus large enough.

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/mkornacker/pixeltable/49)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/tests/test_dataframe.py` line 297 at r1](https://reviewable.io/reviews/mkornacker/pixeltable/49#gh-1421826737:-NoTzcyI8RM4gjaxKX6c:b-896fix) ([raw file](https://github.com/mkornacker/pixeltable/blob/f0032e86f3ea58194bbdbb7703343a0bcf81e9b8/pixeltable/tests/test_dataframe.py#L297)):*

return type missing
</blockquote></details>

Done.

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/mkornacker/pixeltable/49)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/tests/test_dataframe.py` line 314 at r1](https://reviewable.io/reviews/mkornacker/pixeltable/49#gh-1421828408:-NoTzeTu3m7at-FI_1on:b-896fix) ([raw file](https://github.com/mkornacker/pixeltable/blob/f0032e86f3ea58194bbdbb7703343a0bcf81e9b8/pixeltable/tests/test_dataframe.py#L314)):*

This isn't guaranteed to work, because the ordering might change, depending on what Postgres decides. In this case, you should supply an explicit order_by().
</blockquote></details>

Done.
___
*[`pixeltable/tests/test_dataframe.py` line 318 at r1](https://reviewable.io/reviews/mkornacker/pixeltable/49#gh-1421828557:-NoTzfDtAdmMXFFbd78t:b-896fix) ([raw file](https://github.com/mkornacker/pixeltable/blob/f0032e86f3ea58194bbdbb7703343a0bcf81e9b8/pixeltable/tests/test_dataframe.py#L318)):*

missing: check for failure of order_by() + shuffle
</blockquote></details>

Done.

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/mkornacker/pixeltable/49)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/tests/test_dataframe.py` line 289 at r1](https://reviewable.io/reviews/mkornacker/pixeltable/49#gh-1421828092:-NoZZnI-9MqpC1LpvjIf:b3f5gaf) ([raw file](https://github.com/mkornacker/pixeltable/blob/f0032e86f3ea58194bbdbb7703343a0bcf81e9b8/pixeltable/tests/test_dataframe.py#L289)):*

I'm also testing that the default argument is indeed shuffle, which we wouldn't notice if changed if we explicitly test positive.
</blockquote></details>

broke this test up into its own thing, to untangle dataloader stuff from dataset shuffle, which was complicating this dataloader test
 (see new test )
___
*[`pixeltable/tests/test_dataframe.py` line 314 at r1](https://reviewable.io/reviews/mkornacker/pixeltable/49#gh-1421828408:-NoZ_1az8WGsA1BlmHfT:b-cep6bg) ([raw file](https://github.com/mkornacker/pixeltable/blob/f0032e86f3ea58194bbdbb7703343a0bcf81e9b8/pixeltable/tests/test_dataframe.py#L314)):*

Done.
</blockquote></details>

see new test
___
*[`pixeltable/tests/test_dataframe.py` line 318 at r1](https://reviewable.io/reviews/mkornacker/pixeltable/49#gh-1421828557:-NoZ_2i22yu9CFtDjGR6:b-cep6bg) ([raw file](https://github.com/mkornacker/pixeltable/blob/f0032e86f3ea58194bbdbb7703343a0bcf81e9b8/pixeltable/tests/test_dataframe.py#L318)):*

Done.
</blockquote></details>

see new test

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/mkornacker/pixeltable/49)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/tests/test_dataframe.py` line 238 at r3](https://reviewable.io/reviews/mkornacker/pixeltable/49#-NoZ_6Oi-6ETC6yInRwK:-NoZ_6Oi-6ETC6yInRwL:b-adefgb) ([raw file](https://github.com/mkornacker/pixeltable/blob/885433e33bf5fc88d7696c682283785f3e151704/pixeltable/tests/test_dataframe.py#L238)):*
> ```Python
>             assert isinstance(tup['c_json'], dict)
> 
>     def test_to_pytorch_dataset_shuffle(self, all_datatypes_tbl: catalog.MutableTable) -> None:
> ```

new test to test shuffle separately, without needing a dataloader

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mkornacker/pixeltable/49)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @orm011)

___
*[`pixeltable/dataframe.py` line 630 at r4](https://reviewable.io/reviews/mkornacker/pixeltable/49#-NpNhuo90-6Cx8jSqrqk:-NpNhuo90-6Cx8jSqrql:b-xd5q8l) ([raw file](https://github.com/mkornacker/pixeltable/blob/8ad181548e15f1da362f40ec86c250b0b068be3f/pixeltable/dataframe.py#L630)):*
> ```Python
>             if not self.order_by_clause is None:
>                 raise exc.Error('order_by() cannot be used with shuffle')
>             df = self.order_by(exprs.SqlRandom(), asc=True)
> ```

Doesn't this mean that even if you call this with shuffle=True, you'll get the same order as last time when it's coming from the cache? I think that would lead to a lot of confusion and would probably also make the feature not be useful.

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/mkornacker/pixeltable/49)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/dataframe.py` line 630 at r4](https://reviewable.io/reviews/mkornacker/pixeltable/49#-NpNhuo90-6Cx8jSqrqk:-NpQItsL2_-SoS6Z_rUL:b-lx5c8y) ([raw file](https://github.com/mkornacker/pixeltable/blob/8ad181548e15f1da362f40ec86c250b0b068be3f/pixeltable/dataframe.py#L630)):*

Doesn't this mean that even if you call this with shuffle=True, you'll get the same order as last time when it's coming from the cache? I think that would lead to a lot of confusion and would probably also make the feature not be useful.
</blockquote></details>

Not sure I follow, but my interpretation of what you are asking is a scenario where 1.  a call is made without shuffle, then 2. a call is made with shuffle, the question being what's the cache behavior.

The reasonable behavior is for the output in 2. to be shuffled.
BC. the cache key is for`df`, and not for `self`, the intention is these two variations have different cached version, as would a separate 'order by '.

mkornacker:
My point was: if you call to_dataset() twice in a row with shuffle=True, you'll get the same dataset back, with the exact same row order. I don't think this is what people would expect or want.

orm011 (commented):
I can see why this may seem counter-intuitive, at the same time, I think if I were using it to train a model, its actually good to have a way to re-use it, mostly because the idea of shuffling is so that each mini-batch can be seen as providing an unbiased 
"poll" estimate of the Loss gradient. So, re-using the same shuffling is not bad, depending on what you are doing, eg if you want to try different experiments, re-using the same shuffling rather than redoing it every time is not bad.  The alternative, of forcing a re-shuffle seems less desirable.

One option is to add a shuffle seed parameter, and have be set to a default. If the user really cares for a different order, they can change the seed. This enables both scenarios, but I'd say the default seed is ok, if someone really cares to pay that extra price then they can go out of their way.  

> *[Reviewable](https://reviewable.io/reviews/mkornacker/pixeltable/49#-:-NpRiwJH52kgg7YaNoAG:bkji00u)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @mkornacker)

Issue #6, opened 2023-05-26
2-level hierarchy for .pixeltable/media


mkornacker:
Introduce a layer of subdirectories in order to scale to a very large number of stored media files.

aaron-siegel:
Done in PR #64

PR #64, opened 2024-01-16
Multi-level media paths

aaron-siegel:
Introduces a multi-level directory structure for the pixeltable `media_dir` as discussed in issue https://github.com/mkornacker/pixeltable/issues/6.

The standardized media names have been changed from

`f'{tbl_id.hex}_{col_id}_{version}_{id.hex}{ext}'`

to

`tbl_id.hex / version / id.hex[0:3] / f'{tbl_id.hex}_{col_id}_{version}_{id.hex}{ext}'`

Also, `MediaStore.get_path` has been renamed to `create_media_path` (the name `get_path` is misleading, since a new UUID is generated on each call).

Testing done:
`pytest -v`
Also expanded the image table unit test to exercise more of the functionality of MediaStore.

I am assuming no migration or backward-compatibility tools will be needed, but please chime in if they are.


pixeltable/tests/test_table.py
```diff
@@ -111,6 +114,24 @@ def test_create_image_table(self, test_client: pt.Client) -> None:
...
+        assert(MediaStore.count(tbl.id) == 2 * n_sample_rows)
+
+        # Test MediaStore.stats()
+        stats = list(filter(lambda x: x[0] == tbl.id, MediaStore.stats()))
+        assert len(stats) == 2                 # Two columns
+        assert stats[0][2] == n_sample_rows    # Each column has n_sample_rows associated images
+        assert stats[1][2] == n_sample_rows
+
+        # Test version-specific MediaStore.delete
+        MediaStore.delete(tbl.id, version=1)
```
mkornacker: You should do this with a tbl.revert() (which will remove the column and delete the media files). MediaStore.delete() shouldn't be called directly.

Issue #53, opened 2023-12-21
Allow descriptions of schema elements


mkornacker:
Example: Client.create_table('tbl', schema=..., description='test data for ...')

Record descriptions for:
- tables
- views
- snapshots
- columns

PR #77, opened 2024-02-09
Remove TableParameters and move its functionality to TableSchemaVersionMd; add Table.comment() property

aaron-siegel:
Removes `TableParameters` relocates its functionality to `TableSchemaVersionMd` so that table attributes can be versioned. Introduces an optional `comment` field to `TableSchemaVersionMd`. Surfaces `comment` and `num_retained_versions` as properties in `Table`.

Also includes some improvements to the DB migration test framework that makes it more resilient against test failures (previously, a test failure in test_migration could "bleed out" and cause other tests to fail).

Partially resolves issue #53.

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/77)


pixeltable/catalog/table.py
```diff
@@ -167,7 +167,12 @@ def describe(self) -> None:
             print(self.__repr__())
 
     def __repr__(self) -> str:
-        return self._description().to_string(index=False)
+        structure_as_str = self._description().to_string(index=False)
+        if self.tbl_version.attributes.description != '':
+            descr = f'{self.tbl_version.attributes.description}\n'
+        else:
+            descr = ''
+        return f'table \'{self.name}\'\n{descr}{structure_as_str}'
```
mkornacker: use display_name() instead of hardcoding "table"

pixeltable/tests/test_table.py
```diff
@@ -85,6 +85,23 @@ def test_create(self, test_client: pt.Client) -> None:
...
+        cl = test_client
+        schema = {'c': StringType(nullable=False)}
+        num_retained_versions = 20
+        description = "This is a table."
+        tbl = cl.create_table('test_attrs', schema, num_retained_versions=num_retained_versions, description=description)
+        assert(tbl.tbl_version.attributes.num_retained_versions == num_retained_versions)
+        assert(tbl.tbl_version.attributes.description == description)
+        new_num_retained_versions = 30
+        new_description = "This is an updated table."
+        tbl.tbl_version.update_attributes(num_retained_versions=new_num_retained_versions, description=new_description)
```
mkornacker: TableVersion is not a user-facing class.

This should be done via Table, and we should probably call it something like set_attr or set_attribute (to keep with Python vernacular).

In a similar vein, it would make sense to add a Table.description() -> str.

pixeltable/catalog/table_version.py
```diff
@@ -383,6 +383,21 @@ def rename_column(self, old_name: str, new_name: str) -> None:
...
+        )
+
+        # we're creating a new schema version
+        ts = time.time()
+        self.version += 1
+        preceding_schema_version = self.schema_version
+        self.schema_version = self.version
+        with Env.get().engine.begin() as conn:
+            self._update_md(ts, preceding_schema_version, conn)
+        _logger.info(f'Updated attributes: {self.attributes}, new version: {self.version}')
```
mkornacker: Would be more interesting to log only the attributes that were changed, with the value.

pixeltable/catalog/table_version.py
```diff
@@ -48,6 +49,7 @@ def __init__(
         self.id = id
         self.name = tbl_md.name
         self.version = version
+        self.attributes = schema_version_md.attributes
```
mkornacker: Or 'attrs', to stick with Python's getattr/setattr precedent.

pixeltable/catalog/table.py
```diff
@@ -142,6 +142,15 @@ def column_types(self) -> Dict[str, ts.ColumnType]:
         """Return the names of the columns in this table."""
         return {c.name: c.col_type for c in self.tbl_version_path.columns()}
 
+    def get_attrs(self) -> schema.TableAttributes:
```
mkornacker: schema.py is also not user-facing.

maybe just a get_attr(attr_name)? or no getter at all, and just comment() and num_retained_versions() instead?

pixeltable/catalog/table_version.py
```diff
@@ -383,6 +383,22 @@ def rename_column(self, old_name: str, new_name: str) -> None:
             self._update_md(ts, preceding_schema_version, conn)
         _logger.info(f'Renamed column {old_name} to {new_name} in table {self.name}, new version: {self.version}')
 
+    def set_attrs(self, num_retained_versions: Optional[int] = None, comment: Optional[str] = None):
+        self.attrs = TableAttributes(
+            num_retained_versions = self.attrs.num_retained_versions if num_retained_versions is None else num_retained_versions,
```
mkornacker: no spaces around =

pixeltable/catalog/table.py
aaron-siegel: Hmm... I can see an argument for a user-facing TableAttributes struct. Saying "it's part of schema.py, so we shouldn't have a user-facing TableAttributes struct" seems like putting the cart before the horse.

get_attr(attr_name) adds user friction because they have to remember all the attribute names, and separate methods for each attribute seems less clean and doesn't scale as well.

Two other options:
1. simply move TableAttributes out of schema.py (and into table.py, say, or somewhere else) (my preference)
2. have a separate struct that's equivalent to TableAttributes but is user-facing, and table_version.py does the conversion

pixeltable/catalog/table_version.py
aaron-siegel: Fixed, though I hate that convention :)

pixeltable/catalog/table.py
```diff
@@ -168,8 +184,14 @@ def describe(self) -> None:
         except NameError:
             print(self.__repr__())
 
+    # TODO: Display comments in _repr_html()
```
mkornacker, line 187: We should follow up on that pretty quickly, _repr_html() is what runs in a notebook.

pixeltable/metadata/schema.py
```diff
@@ -179,6 +180,7 @@ class TableSchemaVersionMd:
     schema_version: int
     preceding_schema_version: Optional[int]
     columns: Dict[int, SchemaColumn]  # col_id -> SchemaColumn
+    table_attrs: TableAttributes
```
mkornacker: The reason why TableParameters existed in the first place was that it was part of TableMd, but it's really just another needless level of indirection.

Why don't we correct that here and inline the attributes in TableSchemaVersionMd (because they really are part of a particular schema version) and ditch TableParameters/-Attributes altogether.

pixeltable/tests/conftest.py
```diff
@@ -38,10 +38,40 @@ def init_env(tmp_path_factory) -> None:
...
+    clean_db()
     cl = pt.Client(reload=True)
     cl.logging(level=logging.DEBUG, to_stdout=True)
     yield cl
-    cl.reset_catalog()
+
+
+def clean_db(restore_tables=True):
+    from pixeltable.env import Env
+    # (aaron-siegel) The logic from Client.reset_catalog() has been moved here, so that it
```
mkornacker: Why include your name?

pixeltable/tests/conftest.py
```diff
@@ -38,10 +38,40 @@ def init_env(tmp_path_factory) -> None:
...
+    # (such as test_migration.py) may leave the DB in a broken state, from which the
+    # client is uninstantiable.
+    clean_db()
     cl = pt.Client(reload=True)
     cl.logging(level=logging.DEBUG, to_stdout=True)
     yield cl
-    cl.reset_catalog()
+
+
+def clean_db(restore_tables=True):
```
mkornacker: type annotations missing

pixeltable/tests/conftest.py
```diff
@@ -38,10 +38,40 @@ def init_env(tmp_path_factory) -> None:
...
+    from pixeltable.env import Env
+    # (aaron-siegel) The logic from Client.reset_catalog() has been moved here, so that it
+    # does not rely on instantiating a Client object. As before, UUID-named data tables will
```
mkornacker, line 53: uuid-named?

Also, the intention is to delete the data as well. I forgot why I commented it out.

It requires TableVersion.drop() in order to clean out media files as well.

pixeltable/catalog/table.py
aaron-siegel: Sure. I just wasn't sure how to manipulate the Styler properly etc. I'd propose leaving this as a TODO for now, since nothing is actually broken (and you can still see the comment via t.comment())

pixeltable/tests/conftest.py
aaron-siegel: Fixed, thanks

pixeltable/tests/conftest.py
aaron-siegel: I've tried to get in the habit of including my name on TODOs. But you're right, this isn't a TODO. And git blame can be used if it's really needed. I'll remove it

mkornacker (approved):

PR #79, opened 2024-02-10
Together API

aaron-siegel:
Implements the Together AI completion API as a Pixeltable function. In addition:
- Introduces a new configuration file `$PIXELTABLE_HOME/config.yaml` to store API keys (and anything else we might want to add over time)
- Ensures that Python lists work properly as inline arguments to `JsonType` function parameters
- Refactors the unit tests to skip tests for packages that are not installed, and reactivates the openai tests.


pixeltable/env.py
```diff
@@ -136,6 +141,16 @@ def set_up(self, echo: bool = False) -> None:
         self._log_dir = self._home / 'logs'
         self._tmp_dir = self._home / 'tmp'
         self._pgdata_dir = Path(os.environ.get('PIXELTABLE_PGDATA', str(self._home / 'pgdata')))
+        self._config_file = self._home / 'pixeltable.yaml'
```
mkornacker: config.yaml?

precede w/ blank line?

also, do we need to keep this or is it only used once?

pixeltable/env.py
```diff
@@ -68,6 +69,10 @@ def __init__(self):
         self._log_to_stdout = False
         self._module_log_level: Dict[str, int] = {}  # module name -> log level
 
+        # config
+        self._config_file: Optional[Path] = None
+        self._config: Optional[Dict] = None
```
mkornacker: does the Dict have a structure?

pixeltable/env.py
```diff
@@ -227,11 +242,28 @@ def _create_nos_client(self) -> None:
             func.FunctionRegistry.get().register_module(mod)
 
     def _create_openai_client(self) -> None:
-        if not 'OPENAI_API_KEY' in os.environ:
+        if 'openai' in self._config and 'default' in self._config['openai'] and 'api_key' in self._config['openai']['default']:
```
mkornacker: what's the structure of the yaml file?

pixeltable/func/together_function.py
```diff
@@ -0,0 +1,77 @@
...
+        return 1
+
+    def invoke(self, arg_batches: List[List[Any]], kwarg_batches: Dict[str, List[Any]]) -> List[Any]:
+        endpoint = together
+        for e in self.spec.call:
+            endpoint = getattr(endpoint, e)
+        assert len(kwarg_batches) > 0
+        assert len(kwarg_batches[next(iter(kwarg_batches))]) == 1
+        kwargs = {k: v[0] for k, v in kwarg_batches.items()}
+        output_dict = endpoint(**kwargs)
```
mkornacker, line 71: is the missing
        output = endpoint(**kwargs)
        output_dict = output.dict()
the only difference to OpenAIFunction?

leave a "TODO: unify with OpenAIFunction" in the class or module comment? we should probably do the integration for fireworks and replicate and then create a base class.

aaron-siegel: There is one other difference,

cl = env.Env.get().openai_client
endpoint = cl

is replaced by 

endpoint = together

(There is no client object in the together API, everything is just static methods. That's poor SDK design but it is what it is. This can be easily abstracted away of course)

pixeltable/env.py
aaron-siegel: Here's what I have right now:

```
openai:
  default:
    api_key: ...
together:
  default:
    api_key: ...
```

`default` is to allow for the possibility in the future of multiple concurrent client instances (with different API keys)

An alternative could be:

```
openai:
  - name: default
    api_key: ...
together:
  - name: default
    api_key: ...
```

Or just assume 1 client instance per provider:

```
openai:
  api_key: ...
together:
  api_key: ...

Comment (mkornacker) [reply pixeltable/env.py]: Why have more than one api key? For higher throughput?

What are the names good for? (ie, why not make it a list of keys, if we want to support multiple keys)

File: pixeltable/env.py (hunk 2 above)
Comment (mkornacker) [reply pixeltable/env.py]: missed this?

File: pixeltable/env.py (hunk 1 above)
Comment (aaron-siegel) [reply pixeltable/env.py]: Renamed/reformatted as proposed. I kind of want to keep it. It's useful for debugging, and it's a lightweight structure. But I don't feel super strongly about it

File: pixeltable/func/together_function.py [hunk 5]
```diff
@@ -0,0 +1,77 @@
...
+
+import numpy as np
+import jmespath
+
+from .signature import Signature, Parameter
+from .function_md import FunctionMd
+from .external_function import ExternalFunction
+import pixeltable.env as env
+import pixeltable.type_system as ts
+import together
```
mkornacker: let's move this into invoke(), to avoid an import without the package being installed

pixeltable/env.py
aaron-siegel: fixed

pixeltable/env.py
aaron-siegel: Per conversation with @mkornacker I simplified the config format to:
```
openai:
  api_key: ...
together:
  api_key: ...
```

pixeltable/env.py
```diff
@@ -68,6 +69,10 @@ def __init__(self):
         self._log_to_stdout = False
         self._module_log_level: Dict[str, int] = {}  # module name -> log level
 
+        # config
+        self._config_file: Optional[Path] = None
+        self._config: Optional[Dict[str, any]] = None
```
mkornacker: It should be Any, not any.

(What does 'any' resolve to?)

pixeltable/env.py
```diff
@@ -227,11 +247,28 @@ def _create_nos_client(self) -> None:
...
+        if 'together' in self._config and 'api_key' in self._config['together']:
+            api_key = self._config['together']['api_key']
+        else:
+            api_key = os.environ.get('TOGETHER_API_KEY')
+        if api_key is None:
+            self._logger.info('Together client not initialized (no API key configured).')
+            return
+        import together
+        self._logger.info('Initializing Together client.')
+        together.api_key = api_key
```
mkornacker, line 271: seriously? :)

pixeltable/func/together_function.py
mkornacker: ?

aaron-siegel: Good call (sorry missed this comment earlier)

pixeltable/env.py
aaron-siegel: Oops. This was a typo. I'm surprised it runs! (re any vs. Any: I don't know. But now I'm curious.)

pixeltable/env.py
aaron-siegel: Yeah. I told you it was half-baked :)

mkornacker (approved):

PR #83, opened 2024-02-23
 Remove NUL from strings during insert

orm011:
shows up in HF datasets and its legal in python, etc. but not in Postgres.


orm011:
added it to the parquet request.

PR #85, opened 2024-03-02
Enable pytest in CI for pixeltable repo

aaron-siegel:
Sets up a CI workflow that runs pytest in response to any PR or commit to master.

This PR also reorganizes `pyproject.toml` into extras that can be selectively enabled for CI or other purposes, and it introduces a `poetry.toml` config, mainly in order to set `in-project = true`, which I think is good practice anyway.


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,40 @@
...
+    branches: [master]
+
+jobs:
+  ci:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, macos-latest]
+        # TODO: pixeltable currently fails on python 3.12.
+        python-version: ["3.9", "3.10", "3.11"]
```
mkornacker: Can we include 3.8? I think that's still the most frequently used version.

Could be in a follow-on, but then let's leave a todo.

pyproject.toml
```diff
@@ -29,6 +29,28 @@ pgvector = "^0.2.1"
 av = ">=10.0.0"
 pgserver = "0.0.4"
 
+# Optional dependencies (for extras)
```
mkornacker: But does this mean that a user installation (as opposed to dev) now also has to include all of that? I think we should avoid that.

Also, I got rid of extras recently. Maybe we need to think this through better, but it felt like going with explicit extras would mean we'd end up with dozens of extras.

aaron-siegel: I think it's ok to have lots of extras. We can consolidate them into groups or batches (like I did with `misc`) if they're unlikely to cause trouble. But we've had several situations recently where we've wanted to selectively enable various dependencies and I think it's useful to be able to do that. The all-or-nothing approach was too inflexible.

A default install

`poetry install`

will behave identically to before (extras will NOT be installed). A dev install

`poetry install --with dev`

will behave differently from before (extras will also NOT be installed). Extras can be selectively added via (say)

`poetry install --extras "misc openai"`

or

`poetry install --with dev --extras "misc openai"`

If you want to get everything, you can do

`poetry install --with dev --all-extras`

which would be identical to the old `--with-dev`.

If you're still concerned, we should probably discuss on a call.

.github/workflows/pytest.yml
aaron-siegel: Sure, I'll add 3.8.

pyproject.toml
mkornacker: It's not clear to me that we won't end up with one extra per external package, in which case you might as well not have the extra. We need to figure out how to let developers run tests locally, which the dev group was supposed to accomplish.

Let's discuss on a call.

mkornacker (approved):

PR #86, opened 2024-03-05
Set up linting tools for the pixeltable repo

aaron-siegel:
**Proposal:** Introduce the following tools into pixeltable CI:

- **mypy** (type-checking tool)
- **pylint** (linting tool)
- **ruff** (linting and formatting tool)

We may be able to ditch **pylint**, but I'm leaving it in for now, as it has some checks that seem useful that don't appear to be in **ruff**. (But ruff is highly configurable, so I could be wrong. We can get rid of pylint later if we decide it's not adding anything useful.)

**Proposal:** When a PR is opened:

- Lint ONLY the files that are affected by the PR
- Do NOT auto-reformat on commit (as this would introduce changes that are not visible in the PR diff); authors may choose to run `./scripts/format-changed-files.py` before submitting the PR
- However, DO use `ruff format --check` to report on whether the affected files conform to ruff's style conventions

When linting fails in CI, there are several possible things we could do.

1. Do nothing (always pass CI); or
2. Fail CI if `ruff format --check` fails, but not on other issues; or
3. Fail on any linting issue; or
4. Fail on any linting or type-checking issue.

I propose 2. It is relatively unburdensome to PR authors (at worst, all they have to do is run the provided `format-changed-files` script and the PR will pass). 3 or 4 would be very annoying at first; I suggest we make a best effort to address code structure issues but not feel obligated (initially) to resolve 100% of them. Later, once we have our codebase and linter configuration in better alignment, we can always choose to make CI more strict.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
     - name: Checkout repo
       uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
     - name: Install python
       uses: actions/setup-python@v5
       with:
@@ -20,7 +22,7 @@ jobs:
     - name: Install pylint
       run: |
         python -m pip install --upgrade pip
-        pip install pylint
+        pip install pylint flake8 black
     - name: Analyze the code with linting tools
       # This will lint ONLY the files that are affected by the PR.
       run: ./scripts/lint-changed-files.sh
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,7 +19,7 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install pylint
+    - name: Install linting tools
       run: |
         python -m pip install --upgrade pip
         pip install pylint flake8 black
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,25 @@
+name: Lint python files changed by a PR
+
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.11"]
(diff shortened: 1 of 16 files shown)


pyproject.toml
```diff
@@ -12,7 +12,7 @@ exclude = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = "^3.8.1"
```
mkornacker, line 19: comment: why not 3.8?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.11"]
```
mkornacker, line 12: why 3.11?

scripts/lint-changed-files.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
mkornacker, line 1: Should we also add references to this to the Makefile, so developers can run it locally?

pyproject.toml
aaron-siegel: One of the packages (either pylint or flake8 - can't remember which) requires >= 3.8.1.

.github/workflows/lint.yml
aaron-siegel: No particular reason. I don't think it really matters. This is just the version of python that will host the linting tools. I went with 3.11 because it's stable and relatively future-proof, but I'm just as happy to change it if there's a reason to.

I don't see a need to run the linting tools against multiple versions of python (they're static analyzers).

scripts/lint-changed-files.sh
aaron-siegel: Not sure what you mean here?

(It's meant to be run locally, so that the same logic is running locally and in CI... just not sure what's needed from the makefile)

Issue #78, opened 2024-02-10
.astype() function


mkornacker:
In certain cases it's useful to convert data to a different type. In pandas and numpy this is done with astype().

For Pixeltable, it would be useful to be able to do that for expressions, as in
t.select(t.dict_col.path_a.path_b.astype(pxt.StringType()))

aaron-siegel:
Resolved by https://github.com/pixeltable/pixeltable/pull/87.

PR #87, opened 2024-03-06
Add astype() and apply() methods to Expr

aaron-siegel:
Introduces two new methods to `pixeltable.exprs.Expr`:
- `astype(self, new_type: ts.ColumnType) -> Expr` explicitly casts a column as a different type
- `apply(self, fn: Callable) -> FunctionCall` dynamically constructs a `FunctionCall` that wraps the specified `Callable`

This resolves issue #78.


pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,68 @@
+import json
+from typing import Optional, Dict, List, Tuple, Any
+
+import sqlalchemy as sql
+
+from pixeltable.exprs import DataRow, Expr, RowBuilder
```
mkornacker: We're now standardizing on imports like so (see column_property_ref.py, for example):
local:
from .data_row import DataRowKornacker
elsewhere

pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,68 @@
...
+            self.id = self._create_id
+        else:
+            raise RuntimeError(f'Expression of type `{self._underlying.col_type}` cannot be cast to `{self.col_type}`')
+
+    @property
+    def _underlying(self):
+        return self.components[0]
+
+    def _equals(self, other: Expr) -> bool:
+        if isinstance(other, TypeCast):
```
mkornacker: This check isn't necessary (see Expr.equals()).

pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,68 @@
...
+            return self._underlying == other._underlying and self.col_type == other.col_type
```
mkornacker: The col_type check should be part of Expr.equals()

pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,68 @@
...
+
+class TypeCast(Expr):
+
+    def __init__(self, underlying: Expr, new_type: ColumnType):
+        super().__init__(new_type)
+        if self.col_type.is_string_type() \
+                or self.col_type.is_json_type() and underlying.col_type.is_string_type():
+            # It's a valid type conversion
+            self.components = [underlying]
+            self.id = self._create_id
```
mkornacker: This needs to be 'id = self._create_id()'

I'm not sure how to catch this other than to run a type checker (which we should look into; my guess is we'll get a ton of false positives).

Also, this is missing coverage by the serialization test (test_exprs:test_serialization).

pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,68 @@
...
+    def _equals(self, other: Expr) -> bool:
+        if isinstance(other, TypeCast):
+            return self._underlying == other._underlying and self.col_type == other.col_type
+        else:
+            return False
+
+    def _id_attrs(self) -> List[Tuple[str, Any]]:
+        return super()._id_attrs() + [('new_type', self.col_type)]
+
+    def sql_expr(self) -> Optional[sql.ClauseElement]:
```
mkornacker, line 32: Explain in doc string why we're not returning a sql expr

pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,68 @@
...
+        return None
+        # underlying_sql_expr = self._underlying.sql_expr()
+        # if underlying_sql_expr is None:
+        #     return None
+        # else:
+        #     return underlying_sql_expr.cast(self.col_type.to_sa_type)
+
+    # For now, we support only String -> JSON and anything -> String.
```
mkornacker: move to class docstring

pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,68 @@
+import json
+from typing import Optional, Dict, List, Tuple, Any
+
+import sqlalchemy as sql
+
+from pixeltable.exprs import DataRow, Expr, RowBuilder
+from pixeltable.type_system import ColumnType
+
+
+class TypeCast(Expr):
```
mkornacker, line 11: missing docstring

pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,68 @@
...
+        # underlying_sql_expr = self._underlying.sql_expr()
+        # if underlying_sql_expr is None:
+        #     return None
+        # else:
+        #     return underlying_sql_expr.cast(self.col_type.to_sa_type)
+
+    # For now, we support only String -> JSON and anything -> String.
+    # TODO: Support other type conversions as needed.
+    def eval(self, data_row: DataRow, row_builder: RowBuilder) -> None:
+        original_val = data_row[self._underlying.slot_idx]
```
mkornacker, line 40: Use ColumnType.create_literal() here and extend that for JsonType and StringType.

pixeltable/tests/test_exprs.py
```diff
@@ -356,6 +358,28 @@ def test_arrays(self, test_tbl: catalog.Table) -> None:
...
         print(_)
 
+    def test_type_cast(self, test_tbl: catalog.Table) -> None:
+        t = test_tbl
+        # For each column c2, ..., c7, we create a new column ci_as_str that casts it to
+        # a string, then check that each row is correctly converted
+        for col_id in range(2, 8):
+            col_name = f'c{col_id}'
+            as_str_col_name = f'c{col_id}_as_str'
+            t.add_column(**{as_str_col_name: getattr(t, col_name).astype(StringType())})
```
mkornacker: status = t.add_column(...)
assert status.num_excs == 0

pixeltable/tests/test_exprs.py
```diff
@@ -356,6 +358,28 @@ def test_arrays(self, test_tbl: catalog.Table) -> None:
...
+            data = t.select(*[getattr(t, col_name), getattr(t, as_str_col_name)]).collect()
```
mkornacker: better: t[col_name]

pixeltable/tests/test_exprs.py
```diff
@@ -356,6 +358,28 @@ def test_arrays(self, test_tbl: catalog.Table) -> None:
...
+            is_json_type = getattr(t, col_name).col_type.is_json_type()
```
mkornacker: This isn't part of the user-visible API. We should probably make that clearer with private members.

You can use Table.column_types()[col_name] to get this.

pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,68 @@
...
+from pixeltable.exprs import DataRow, Expr, RowBuilder
+from pixeltable.type_system import ColumnType
+
+
+class TypeCast(Expr):
+
+    def __init__(self, underlying: Expr, new_type: ColumnType):
+        super().__init__(new_type)
+        if self.col_type.is_string_type() \
+                or self.col_type.is_json_type() and underlying.col_type.is_string_type():
```
mkornacker: This check doesn't really work, because a json-typed column can give you a string value.

See docs/tutorials/openai_demo.ipynb (look at column eval_output at the very bottom; it has type json but contains strings).

We want to extend the tutorial to cast t.eval.choices[0].message.content to an actual dict (it already has column type json, but a single string is actually a valid json value; an explicit cast should try to convert the string to a dict).

mkornacker (changes requested):

pixeltable/exprs/type_cast.py
aaron-siegel: ah, thanks. I misunderstood the pattern. `return True` should suffice here.

pixeltable/exprs/type_cast.py
aaron-siegel: Oops ...

I'm surprised this worked at all, it's such an egregious type violation :/

I added some type hints to the initializer. I also added the missing test cases to `test_serialization`, but to my surprise, `test_serialization` still passes even with the horribly bogus `self.id`. I'm not sure what this means.

pixeltable/tests/test_exprs.py
aaron-siegel: Added, though it's not strictly necessary (the test will fail later if something goes wrong with column creation)

pixeltable/tests/test_exprs.py
aaron-siegel: Hmm. I can change this I guess, but I'd argue it *should* be part of the user-visible API. Or at least, there should be some visible surrogate.

`column.my_type`

is a much more natural pattern than

`my_table.column_types()[col_name]`

i.e., as a user I'd like to think of the column as an object that knows about it's type.

pixeltable/exprs/type_cast.py
aaron-siegel: Ok, I think this comment is a signal that there's something I'm not understanding about what `eval` is supposed to be doing. Probably easier to clarify on a call when you have a moment.

pixeltable/exprs/type_cast.py
aaron-siegel: I'm slightly confused by this. What type of cast are you talking about?

If I understand this comment right, you're concerned about the case where the column is a JsonType, and we want to cast it to a JsonType?

I'd argue that a cast from type `T` to `T` should always be a no-op. If it's a JsonType column and it holds a string because a string is valid JSON, then it should still hold a string after the cast.

I'd like to better understand the use case you're concerned about.

aaron-siegel: Note: I realize that the first cut of this didn't allow converting `T` to `T` at all; I've now fixed this (I think)

pixeltable/exprs/type_cast.py
```diff
@@ -0,0 +1,53 @@
...
+
+class TypeCast(Expr):
+    """
+    An `Expr` that represents a type conversion from an underlying `Expr` to
+    a specified `ColumnType`.
+    """
+    def __init__(self, underlying: Expr, new_type: ts.ColumnType):
+        super().__init__(new_type)
+        self.components: List[Expr] = [underlying]
+        self.id: Optional[int] = self._create_id()
```
mkornacker, line 19: Side note: let's discuss whether we want to require type hints in cases where the type can be inferred.

pixeltable/tests/test_exprs.py
```diff
@@ -356,6 +358,44 @@ def test_arrays(self, test_tbl: catalog.Table) -> None:
...
+    def test_astype(self, test_tbl: catalog.Table) -> None:
+        t = test_tbl
+        # Convert int to float
+        status = t.add_column(c2_as_float=t.c2.astype(FloatType()))
+        assert status.num_excs == 0
+        data = t.select(t.c2, t.c2_as_float).collect()
+        for row in data:
+            assert isinstance(row['c2'], int)
+            assert isinstance(row['c2_as_float'], float)
+            assert row['c2'] == row['c2_as_float']
```
mkornacker, line 371: This is missing failure scenarios (with pytest.raises()...)

Also, test astype() on something other than a ColumnRef.

pixeltable/exprs/expr.py
```diff
@@ -479,3 +511,9 @@ def _make_arithmetic_expr(self, op: ArithmeticOperator, other: object) -> 'pixel
             return ArithmeticExpr(op, self, Literal(other))  # type: ignore[arg-type]
         raise TypeError(f'Other must be Expr or literal: {type(other)}')
 
+
+_known_applicator_types = {
```
mkornacker: add explanatory comment

pixeltable/exprs/expr.py
```diff
@@ -388,6 +390,36 @@ def from_dict_list(cls, dict_list: List[Dict]) -> List[Expr]:
...
+        else:
+            # No type hint
+            fn_type = None
+
+        if fn_type is None:
+            raise RuntimeError(
+                f'Column type of `{fn.__name__}` cannot be inferred. Use `.apply({fn.__name__}, col_type=...)` to specify.'
+            )
+
+        function = make_function(return_type=fn_type, param_types=[self.col_type], eval_fn=lambda x: fn(x))
```
mkornacker: This is missing a bunch of verification/other logic, such as
- does this have exactly one non-default parameter?
- deal with this being a keyword-only parameter
- if fn has type annotations, do type checking

pixeltable/tests/test_exprs.py
```diff
@@ -356,6 +358,44 @@ def test_arrays(self, test_tbl: catalog.Table) -> None:
...
+            str_col_name = f'c{col_id}_str'
+            back_to_json_col_name = f'c{col_id}_back_to_json'
+            status = t.add_column(**{str_col_name: t[col_name].apply(json.dumps)})
+            assert status.num_excs == 0
+            status = t.add_column(**{back_to_json_col_name: t[str_col_name].apply(json.loads)})
+            assert status.num_excs == 0
+            data = t.select(t[col_name], t[str_col_name], t[back_to_json_col_name]).collect()
+            for row in data:
+                assert row[str_col_name] == json.dumps(row[col_name])
+                assert row[back_to_json_col_name] == row[col_name]
```
mkornacker, line 418: Missing: failure scenarios

Also, make sure that the tests completely cover the newly-added code (ie, all branches)

mkornacker (changes requested):

pixeltable/exprs/expr.py
aaron-siegel: I added some validation for parameter counts. I'm going to pass on doing type checking for now (we can discuss in person).

pixeltable/exprs/type_cast.py
aaron-siegel: I think they should be optional (can be used when it helps make the code clearer or more readable, but not required).

pixeltable/exprs/expr.py
```diff
@@ -1,21 +1,25 @@
...
 
-from .globals import ComparisonOperator, LogicalOperator, LiteralPythonTypes, ArithmeticOperator
-from .data_row import DataRow
 import pixeltable
-import pixeltable.exceptions as excs
 import pixeltable.catalog as catalog
+import pixeltable.exceptions as excs
 import pixeltable.type_system as ts
+from pixeltable.func import make_function
+from .data_row import DataRow
```
mkornacker, line 21: Why the reordering?

pixeltable/exprs/expr.py
```diff
@@ -392,6 +396,74 @@ def from_dict_list(cls, dict_list: List[Dict]) -> List[Expr]:
     def _from_dict(cls, d: Dict, components: List[Expr]) -> Expr:
         assert False, 'not implemented'
 
+    def astype(self, new_type: ts.ColumnType) -> 'pixeltable.exprs.TypeCast':
+        from pixeltable.exprs import TypeCast
+        return TypeCast(self, new_type)
+
+    def apply(self, fn: Callable, *, col_type: Optional[ts.ColumnType] = None) -> 'pixeltable.exprs.FunctionCall':
```
mkornacker, line 403: I think most of this logic should live as a classmethod or something like that in FunctionCall.

pixeltable/tests/test_exprs.py
```diff
@@ -356,6 +359,112 @@ def test_arrays(self, test_tbl: catalog.Table) -> None:
...
+
+        def f1(x):
+            return str(x)
+
+        # Now test that a function without a return type throws an exception ...
+        with pytest.raises(RuntimeError):
```
mkornacker: with ... as exc_info:
    ...
assert 'blabla' in str(exc_info.value)

is more specific

pixeltable/tests/test_exprs.py
```diff
@@ -356,6 +359,112 @@ def test_arrays(self, test_tbl: catalog.Table) -> None:
...
+            t['c2'].apply(f1)
+
+        # ... but works if the type is specified explicitly.
+        status = t.add_column(c2_str_f1=t['c2'].apply(f1, col_type=StringType()))
```
mkornacker: t.c2.apply(...) should work here

pixeltable/tests/test_exprs.py
```diff
@@ -356,6 +359,112 @@ def test_arrays(self, test_tbl: catalog.Table) -> None:
...
+        status = t.add_column(c2_str_f2=t['c2'].apply(f2))
+        assert status.num_excs == 0
+
+        # Test various validation failures.
+
+        def f3(x, y) -> str:
+            return f'{x}{y}'
+
+        with pytest.raises(RuntimeError):
+            t['c2'].apply(f3)       # Too many required parameters
```
mkornacker: wny the spaces before '#'?

pixeltable/tests/test_exprs.py
```diff
@@ -356,6 +359,112 @@ def test_arrays(self, test_tbl: catalog.Table) -> None:
...
+
+        def f5(**kwargs) -> str:
+            return ""
+
+        with pytest.raises(RuntimeError):
+            t['c2'].apply(f5)       # No positional parameters
+
+        # Ensure these varargs signatures are acceptable
+
+        def f6(x, **kwargs) -> str:
```
mkornacker, line 464: I think support kwargs (eg, apply(json.loads, parse_float=...) would be useful. At least leave a todo in apply()

pixeltable/tests/conftest.py
```diff
@@ -74,6 +74,13 @@ def test_tbl_exprs(test_tbl: catalog.Table, test_stored_fn: pt.Function) -> List
         (t.c2 > 5) | (t.c1 == 'test'),
         t.c7['*'].f5 >> [R[3], R[2], R[1], R[0]],
         t.c8[0, 1:],
+        t.c1.astype(StringType()),
+        t.c2.astype(StringType()),
+        t.c3.astype(StringType()),
+        t.c4.astype(StringType()),
+        t.c5.astype(StringType()),
+        t.c6.astype(StringType()),
+        t.c1.astype(JsonType()),
```
mkornacker: Also add an apply()

pixeltable/exprs/expr.py
aaron-siegel: I'm using "optimize imports" in PyCharm to put them in a canonical order. It's pretty standard and (I think) it's also the ordering that linters expect: PEP8 ordering for import categories, alphabetical within each category. It will also remove unused imports.

pixeltable/exprs/expr.py
aaron-siegel: Ok

pixeltable/tests/test_exprs.py
aaron-siegel: No good reason. Will fix. A formatter would enforce this :)

pixeltable/tests/test_exprs.py
aaron-siegel: Aha. So `expr.apply(json.loads, parse_float=...)` would call `json.loads(x, parse_float=...)` on each element x?

Note the "functional" way of doing this would be

`expr.apply(lambda x: json.loads(x, parse_float=...))`

or

```
def my_loader(x): json.loads(x, parse_float=...)
expr.apply(my_loader)
```

either of which of course already works. But your suggestion feels more "pythonic" and propagates the function name in a more concise way.

On the other hand, the lambda pattern is more general, since it also handles situations involving additional positional parameters, e.g.

```
def f(a, b): return a + b
expr.apply(lambda x: f(x, 3))
```

Of course we can also envision ways of handling this without the lambda by further extending apply().

I'll leave a TODO for now and we can ruminate on it some more.

pixeltable/exprs/expr.py
aaron-siegel: I tried putting it there and it didn't seem right. The logic is so expression-specific. I factored it out into a Expr._make_applicator_function() which is a peer to Expr._make_arithmetic_expr() and Expr._make_comparison() etc., which seems appropriate to me. Let me know what you think.

pixeltable/tests/test_exprs.py
aaron-siegel: Good catch. Fixed

pixeltable/tests/conftest.py
aaron-siegel: Oops, good catch. I completely forgot to update this when the semantics of `astype` changed.

I'm surprised nothing broke here. What is this test artifact actually being used for?

pixeltable/exprs/expr.py
```diff
@@ -483,3 +496,87 @@ def _make_arithmetic_expr(self, op: ArithmeticOperator, other: object) -> 'pixel
             return ArithmeticExpr(op, self, Literal(other))  # type: ignore[arg-type]
         raise TypeError(f'Other must be Expr or literal: {type(other)}')
 
+    def _make_applicator_function(self, fn: Callable, col_type: Optional[ts.ColumnType]) -> 'pixeltable.func.Function':
```
mkornacker, line 499: function comment missing

pixeltable/exprs/expr.py
```diff
@@ -483,3 +496,87 @@ def _make_arithmetic_expr(self, op: ArithmeticOperator, other: object) -> 'pixel
...
+            params = inspect.signature(fn).parameters
+            params_iter = iter(params.values())
+            first_param = next(params_iter) if len(params) >= 1 else None
+            second_param = next(params_iter) if len(params) >= 2 else None
+            # Check that fn has at least one positional parameter
+            if len(params) == 0 or first_param.kind in (inspect.Parameter.KEYWORD_ONLY, inspect.Parameter.VAR_KEYWORD):
+                raise excs.Error(
+                    f'Function `{fn.__name__}` has no positional parameters.'
+                )
+            # Check that fn has at most one required parameter, i.e., its second parameter
```
mkornacker, line 551: So we're okay with 
def f(x=1, y=2):
...
t.col.apply(f)?

Feels a bit weird. I'm wondering if we should require exactly one required parameter.

aaron-siegel: Yeah I considered this case and intentionally decided to support it. It's the most logical behavior to me - if I were a user, I'd think it was weird for apply() to reject a function just because its first parameter has a default. I don't feel super strongly about it though.

pixeltable/exprs/expr.py
```diff
@@ -483,3 +496,95 @@ def _make_arithmetic_expr(self, op: ArithmeticOperator, other: object) -> 'pixel
             return ArithmeticExpr(op, self, Literal(other))  # type: ignore[arg-type]
         raise TypeError(f'Other must be Expr or literal: {type(other)}')
 
+    def _make_applicator_function(self, fn: Callable, col_type: Optional[ts.ColumnType]) -> 'pixeltable.func.Function':
+        """
+        Creates a unary pixeltable ``Function`` that encapsulates a python ``Callable``. The result type of
```
mkornacker: Just FYI, I think single backticks are sufficient.

mkornacker (approved):

pixeltable/exprs/expr.py
aaron-siegel: Ok. I think I was parroting something else I saw in the codebase. Anyway, I switched to single backticks. Thanks!

PR #88, opened 2024-03-07
Fix many instances of dot-completion by adding missing imports

aaron-siegel:
Fix many instances of dot-completion by adding missing imports (and in some cases, fully qualifying types where a direct import would introduce a circular type reference).

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -9,7 +9,6 @@ import json
 
 import pandas as pd
 import sqlalchemy as sql
-import torch
 
 from .schema_object import SchemaObject
 from .column import Column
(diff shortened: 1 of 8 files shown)


pixeltable/catalog/table.py
```diff
@@ -9,13 +9,15 @@
 
 import pandas as pd
 import sqlalchemy as sql
+import torch
```
mkornacker: This is problematic, because it'll require torch to be installed, which is not a Pixeltable dependency.

Maybe the torch-specific functionality should be in a pixeltable.utils.torch module (and we move what's currently in utils into something like internal_utils).

In order to get these changes in, let's remove this import and deal with to_pytorch_dataset() later.

pixeltable/type_system.py
```diff
@@ -11,6 +11,7 @@
 import av
 import numpy as np
 import PIL.Image
+import pyarrow
```
mkornacker: Another import we don't want at the top level. Dot-completion shouldn't matter here, plus all the pyarrow functions are going away anyway with Oscar's changes.

mkornacker (changes requested):

pixeltable/catalog/table.py
aaron-siegel: Good call.

I removed the `torch` dependency and added a TODO noting that a refactor will be necessary.

pixeltable/type_system.py
aaron-siegel: Ok, done.

PR #92, opened 2024-03-10
Fix usage of mutable default arguments in the codebase

aaron-siegel:


pixeltable/exprs/row_builder.py
```diff
@@ -269,8 +271,10 @@ def get_dependencies(self, targets: List[Expr], exclude: List[Expr] = []) -> Lis
         result_ids.sort()
         return [self.unique_exprs[id] for id in result_ids]
 
-    def create_eval_ctx(self, targets: List[Expr], exclude: List[Expr] = []) -> EvalCtx:
+    def create_eval_ctx(self, targets: List[Expr], exclude: List[Expr] = None) -> EvalCtx:
```
mkornacker: Optional[...] missing

mkornacker (approved):

pixeltable/exprs/row_builder.py
aaron-siegel: Good eye!!

pixeltable/exec/data_row_batch.py
```diff
@@ -9,19 +9,22 @@
...
+
     def __init__(self, tbl: catalog.TableVersion, row_builder: exprs.RowBuilder, len: int = 0):
         self.tbl_id = tbl.id
         self.tbl_version = tbl.version
         self.row_builder = row_builder
         self.img_slot_idxs = [e.slot_idx for e in row_builder.unique_exprs if e.col_type.is_image_type()]
         # non-image media slots
         self.media_slot_idxs = [
-            e.slot_idx for e in row_builder.unique_exprs
+            e.slot_idx
```
mkornacker: Is that the auto-formatter? I'm not sure this is a good use of vertical space.

pixeltable/exec/data_row_batch.py
```diff
@@ -33,7 +36,8 @@ def __init__(self, tbl: catalog.TableVersion, row_builder: exprs.RowBuilder, len
     def add_row(self, row: Optional[exprs.DataRow] = None) -> exprs.DataRow:
         if row is None:
             row = exprs.DataRow(
-                self.row_builder.num_materialized, self.img_slot_idxs, self.media_slot_idxs, self.array_slot_idxs)
+                self.row_builder.num_materialized, self.img_slot_idxs, self.media_slot_idxs, self.array_slot_idxs
+            )
```
mkornacker: Also not in love with this.

pixeltable/exec/sql_scan_node.py
```diff
@@ -47,8 +53,9 @@ def __init__(
             sql_subexprs = iter_arg.subexprs(filter=lambda e: e.sql_expr() is not None, traverse_matches=False)
             [self.sql_exprs.append(e) for e in sql_subexprs]
         self.filter = filter
-        self.filter_eval_ctx = \
+        self.filter_eval_ctx = (
```
mkornacker: I feel like this is an argument for trying out other auto-formatter options. I already volunteered.

pixeltable/exec/sql_scan_node.py
```diff
@@ -192,7 +203,7 @@ def __next__(self) -> DataRowBatch:
             if needs_row:
                 output_row = output_batch.add_row()
             if self.num_pk_cols > 0:
-                output_row.set_pk(tuple(sql_row[-self.num_pk_cols:]))
+                output_row.set_pk(tuple(sql_row[-self.num_pk_cols :]))
```
mkornacker: This is completely non-standard formatting.

mkornacker (changes requested):
Let's discuss offline whether we want to run things through the auto-formatter just yet.

pixeltable/exec/data_row_batch.py
aaron-siegel: I backed out the formatting changes for now, until we can discuss. There are some options that can be configured, which we can look into. Note that this one could be condensed onto a single line (which would be even more compact vertically) and then I'm guessing the formatter would leave it alone.

pixeltable/exec/data_row_batch.py
aaron-siegel: FWIW I prefer it.

pixeltable/exec/sql_scan_node.py
aaron-siegel: Again, I prefer this. These are all following standard (PEP8) style conventions.

pixeltable/exec/sql_scan_node.py
aaron-siegel: Actually I don't think that's true. In compound expressions within a slice, it's normal to put a space between the expression and the colon.

See: https://peps.python.org/pep-0008/

In particular:

"In an extended slice, both colons must have the same amount of spacing applied. Exception: when a slice parameter is omitted, the space is omitted".

I believe the formatter is doing the correct thing in this case.

aaron-siegel: Here's an example straight out of PEP8:

`ham[:: step_fn(x)]`

Note the "one-sided" spacing when the slice has a single parameter that's a compound expression.

mkornacker (approved):

PR #94, opened 2024-03-13
Improve type safety of various parts of the pixeltable codebase

aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/94)


pixeltable/dataframe.py
```diff
@@ -417,27 +421,27 @@ def __repr__(self) -> str:
     def _repr_html_(self) -> str:
         return self._description_html()._repr_html_()
 
-    def select(self, *items: Any, **named_items : Any) -> DataFrame:
+    def select(self, *items: exprs.Expr, **named_items: exprs.Expr) -> DataFrame:
```
mkornacker: That's actually not correct, because you can do stuff like select([1, 2, t.int_col]). But there should be a comment explaining that.

pixeltable/client.py
```diff
@@ -210,7 +212,7 @@ def create_view(
         _logger.info(f'Created view {path_str}')
         return view
 
-    def get_table(self, path: str) -> catalog.Table:
+    def get_table(self, path: str) -> catalog.SchemaObject:
```
mkornacker: Why not Table?

pixeltable/env.py
```diff
@@ -37,7 +36,7 @@ def get(cls) -> Env:
             cls._instance = Env()
         return cls._instance
 
-    def __init__(self):
+    def __init__(self) -> None:
```
mkornacker, line 39: When does a type checker complain about the absence of -> None?

pixeltable/env.py
```diff
@@ -251,6 +251,7 @@ def _create_nos_client(self) -> None:
             func.FunctionRegistry.get().register_module(mod)
 
     def _create_openai_client(self) -> None:
+        assert self._config is not None
```
mkornacker, line 254: Maybe we should run set_up() in init. Or initialize to something non-None. It doesn't really make sense for these things to be None.

pixeltable/store.py
```diff
@@ -186,9 +193,9 @@ def add_column(self, col: catalog.Column, conn: sql.engine.Connection) -> None:
         message).
         """
         assert col.is_stored
-        stmt = sql.text(f'ALTER TABLE {self._storage_name()} ADD COLUMN {col.storage_name()} {col.col_type.to_sql()}')
-        log_stmt(_logger, stmt)
-        conn.execute(stmt)
+        clause = sql.text(f'ALTER TABLE {self._storage_name()} ADD COLUMN {col.storage_name()} {col.col_type.to_sql()}')
```
mkornacker: Why the rename? This is a statement, not a clause, from a SQL perspective.

pixeltable/dataframe.py
aaron-siegel: Oh. Ok I'll fix this

pixeltable/client.py
aaron-siegel: The typechecker doesn't know it's a Table, because it's fetched from a PathDict which returns a SchemaObject.

But I agree it should be a Table and probably the best pattern in this case is
```
obj = self.catalog.paths[p]
assert isinstance(obj, Table)
return obj
```

I'll fix it.

pixeltable/env.py
aaron-siegel: Oh hmm, I think it should be ignoring this one. Maybe I put it in there casually rather than in response to a type checker message?

Ordinarily it ignores untyped methods unless you specify `--disallow-untyped-defs`.

pixeltable/env.py
aaron-siegel: Yeah I agree, it's been bugging me. I'm not trying to fix that kind of thing on this pass though.

pixeltable/store.py
aaron-siegel: There's another variable named `stmt` later in the same function, which takes an unrelated type (a `tuple`). mypy doesn't like it (rightly so, IMO) when the same variable is reused in the same scope for unrelated types.

pixeltable/env.py
aaron-siegel: It does indeed complain about this. I don't know why. FWIW I prefer it with `-> None` but don't feel super strongly about it and there may be a way to disable that warning.

pixeltable/store.py
mkornacker: Oh right. Probably best to call those stmt_str (they're strings).

aaron-siegel: I found a different, much better way to address this. See the updated PR (posting in a few minutes)

PR #95, opened 2024-03-15
Add Batch capability to udf decorator and refactor Inference APIs accordingly

aaron-siegel:
- Integrates `BatchedFunction`s into the `udf` decorator
- Enhances `udf` decorator with type inference (you can now specify the udf signature with Python type hints instead of explicit pixeltable types)
- Refactors Inference APIs to use the enhanced `udf` decorator

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95)


orm011 (commented):
I like the idea of using the same pxt.function() as much as possible to implement these calls.  Also the pxt.udf to pxt.function may be something users understand better.  The guiding principle of using the user-facing API  as much as possible could save us a lot of work.  

I also would suggest we follow this guiding principle in trying out developing small UDFs on a notebook, just to experience what difficulties come up. is it easy to visualize the outputs of the udf on a table as one develops the UDF, and to call other UDF,  or does one need to extract data to python variables and only work on those.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95#-:-NszrQv84-aNU3DIcKh3:b-5rzxmq)* status: 0 of 11 files reviewed, all discussions resolved

pixeltable/functions/openai.py
```diff
@@ -0,0 +1,120 @@
...
+    return result.dict()['choices'][0]['message']['content']
+
+
+@pxt.udf()
+def vision(
+        prompt: str,
+        image: PIL.Image.Image,
+        model: str = 'gpt-4-vision-preview'
+) -> str:
+    bytes_arr = io.BytesIO()
```
mkornacker, line 51: In general, I don't think it's workable to mix general cpu pre/post-processing with logic that needs to run asynchronously or on the gpu.

pixeltable/functions/openai.py
```diff
@@ -0,0 +1,120 @@
...
+    return result.choices[0].message.content
+
+
+@pxt.udf()
+def moderations_create(input: str, model: Optional[str] = None) -> dict:
+    result = env.Env().get().openai_client.moderations.create(input=input, model=model)
+    return result.dict()
+
+
+# TODO(aaron-siegel): Implement batching
```
mkornacker: It's not obvious to me what that would look like.

pixeltable/functions/openai.py
```diff
@@ -0,0 +1,120 @@
...
+        max_tokens: Optional[int] = None,
+        n: Optional[int] = None,
+        presence_penalty: Optional[float] = None,
+        response_format: Optional[dict] = None,
+        seed: Optional[int] = None,
+        top_p: Optional[float] = None,
+        temperature: Optional[float] = None
+) -> str:
+    messages = [
+        {'role': 'system', 'content': 'You are a helpful assistant.'},
```
mkornacker: I question whether people would find this useful or whether they'd rather provide the system context themselves.

pixeltable/functions/openai.py
aaron-siegel: Per discussion with Marcel, this can be handled later by decomposing into two `Function`s and chaining them with an `ExprTemplate`.

pixeltable/functions/openai.py
aaron-siegel: See the new `huggingface.py` for a proposal.

orm011:
*[`pixeltable/env.py` line 81 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqK49XEiqBG4SzY_ty:-NtqK49XEiqBG4SzY_tz:buqhj9u) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/env.py#L81)):*
> ```Python
> 
>         # config
>         self._config_file: Optional[Path] = None
> ```

Are there situations where there is no config? why is it optional.

orm011:
*[`pixeltable/env.py` line 88 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqK1Z59PALwodI4RDw:-NtqK1Z59PALwodI4RDx:b-haxpv5) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/env.py#L88)):*
> ```Python
>         self._stdout_handler.setFormatter(logging.Formatter(self._log_fmt_str))
>         self._initialized = False
> ```

Why not just use .config. If it ever needs to change, @property can be added.

orm011:
*[`pixeltable/env.py` line 408 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqKLwp5hJ9EIZ_xyri:-NtqKLwp5hJ9EIZ_xyrj:bn8e54) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/env.py#L408)):*
> ```Python
> 
>     @property
>     def openai_client(self) -> Optional['openai.OpenAI']:
> ```

if this was inited, it should be not None from the start (should be inited in constructor), otherwise, this is an error (like with engine()), no?

orm011 (dismissed):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95)* status: 0 of 20 files reviewed, 8 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/functions/huggingface.py` line 13 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqLd5y5BGObbIyvEFa:-NtqLd5y5BGObbIyvEFb:b8eq77s) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/functions/huggingface.py#L13)):*
> ```Python
> 
> @pxt.udf(batch_size=32, return_type=ts.ArrayType((None,), dtype=ts.FloatType()))
> def sentence_transformer(sentences: Batch[str], *, model_id: str, normalize_embeddings: bool = False) -> Batch[np.ndarray]:
> ```

typically, a batch of np.ndarray is an np.ndarray of a larger rank, with the first dimension being the batch dimension. this has strong advantages in terms of memory locality and allocation overhead. Is there a reason why we are choosing differently?
___
*[`pixeltable/functions/huggingface.py` line 25 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqLz1m2wHHOgeW2l23:-NtqLz1m2wHHOgeW2l24:b-s09y9b) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/functions/huggingface.py#L25)):*
> ```Python
> 
> @pxt.udf
> def sentence_transformer_list(sentences: list, *, model_id: str, normalize_embeddings: bool = False) -> list:
> ```

is there some reason why these two API (list or batch) must have different names? from a user point of view, seems odd.
___
*[`pixeltable/functions/openai.py` line 60 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1530714763:-NtqMFg1DxRmKmPLvJvu:bkq0cjs) ([raw file](https://github.com/pixeltable/pixeltable/blob/43313b439025323e0ce5e4112171906cd832c46a/pixeltable/functions/openai.py#L60)):*

I question whether people would find this useful or whether they'd rather provide the system context themselves.
</blockquote></details>

+1 to Marcel's observation, 
___
*[`pixeltable/functions/openai.py` line 85 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1530711632:-NtqMPeH9JYvB9GxeMej:bm6eq05) ([raw file](https://github.com/pixeltable/pixeltable/blob/43313b439025323e0ce5e4112171906cd832c46a/pixeltable/functions/openai.py#L85)):*

Per discussion with Marcel, this can be handled later by decomposing into two `Function`s and chaining them with an `ExprTemplate`.
</blockquote></details>

In reponse to Marcel's observation. We should take seriously something along the lines of  https://github.com/orm011/pixeltable/blob/udf_prototype/docs/tutorials/example_func.ipynb. 
Within that fuction, we can tell whether something is GPU-runnable or by checking if we are calling an nn.Module. This is not mutually exclusive with having ExprTemplate.
In my view, the natural way to write a function that takes eg an image and applies a model is made up of a preproc-step, a model call, and a post porocess step. Generally these would run in different defices, with asynchrony ideal in a batch setting. Being able to write this as a function and test it as a function, and then have it be decomposed for production run on btaches seems small but is  huge added value, in my opinion. pushing us closer to the vision of avoiding re-writing when moving from notebook experiments to prod runs.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95)* status: 0 of 20 files reviewed, 8 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/env.py` line 81 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqK49XEiqBG4SzY_ty:-NtqZ1IJ9GL2UXySWugh:bowvkao) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/env.py#L81)):*

Are there situations where there is no config? why is it optional.
</blockquote></details>

Due to lazy initialization.

We can revisit this, but it's out of scope for this PR.
___
*[`pixeltable/env.py` line 88 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqK1Z59PALwodI4RDw:-NtqZG3DA7i61Y-INsqB:bs5l0ty) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/env.py#L88)):*

Why not just use .config. If it ever needs to change, @property can be added.
</blockquote></details>

Out of scope for this PR.

Let's try to keep review comments focused on things that are relevant to the PR. More general questions about codebase architecture we can spin up as a separate conversation.
___
*[`pixeltable/env.py` line 408 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqKLwp5hJ9EIZ_xyri:-NtqZW4o5g5MGXW-nrYu:bl6i7kp) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/env.py#L408)):*

if this was inited, it should be not None from the start (should be inited in constructor), otherwise, this is an error (like with engine()), no?
</blockquote></details>

I'm planning to refactor the initialization logic for OpenAI etc. as a follow up PR (once this one is merged).
___
*[`pixeltable/functions/huggingface.py` line 13 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqLd5y5BGObbIyvEFa:-NtqZcpl7WDpDRnEP3DB:b-c7xq7d) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/functions/huggingface.py#L13)):*

typically, a batch of np.ndarray is an np.ndarray of a larger rank, with the first dimension being the batch dimension. this has strong advantages in terms of memory locality and allocation overhead. Is there a reason why we are choosing differently?
</blockquote></details>

We can look into this, but it would require some architecture changes in how the pixeltable core does batching. Again, out of scope for this PR (which is just a refactoring of the interface pattern, not intended to address functional changes in how pixeltable does batching).
___
*[`pixeltable/functions/huggingface.py` line 25 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtqLz1m2wHHOgeW2l23:-NtqZsq_Bu575C--Jt2K:b-5sexr0) ([raw file](https://github.com/pixeltable/pixeltable/blob/9d194361bc1b0c4991927686ed706cf035f35bd3/pixeltable/functions/huggingface.py#L25)):*

is there some reason why these two API (list or batch) must have different names? from a user point of view, seems odd.
</blockquote></details>

The reason is that the types are different (one takes a JsonType() list, the other takes StringType()s as a batch). If we want to give them the same name AND preserve type-safety, then we'll need a way to handle function overloads in pixeltable, which is a larger conversation.
___
*[`pixeltable/functions/openai.py` line 60 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1530714763:-Ntq_Fuc1hEMGysnvKk9:byfoigp) ([raw file](https://github.com/pixeltable/pixeltable/blob/43313b439025323e0ce5e4112171906cd832c46a/pixeltable/functions/openai.py#L60)):*

+1 to Marcel's observation,
</blockquote></details>

I can take this out if it's really controversial, but I also don't think it's doing a lot of harm. One intended purpose is to facilitate streamlined tutorials that demonstrate the usefulness of pixeltable features without distracting ppl with the nuances of the OpenAI API. Of course, the fully general API will be made available either way.

We should probably involve Pierre in this discussion. He might have opinions since there's a product question here (how far do we go in regards to providing "streamlined" wrapper APIs in addition to the base APIs? Does this make the product easier to use, or just more confusing?)
___
*[`pixeltable/functions/openai.py` line 85 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1530711632:-NtqaA_99CXkFQZ0mg-G:bdwdnj2) ([raw file](https://github.com/pixeltable/pixeltable/blob/43313b439025323e0ce5e4112171906cd832c46a/pixeltable/functions/openai.py#L85)):*

In reponse to Marcel's observation. We should take seriously something along the lines of  https://github.com/orm011/pixeltable/blob/udf_prototype/docs/tutorials/example_func.ipynb. 
Within that fuction, we can tell whether something is GPU-runnable or by checking if we are calling an nn.Module. This is not mutually exclusive with having ExprTemplate.
In my view, the natural way to write a function that takes eg an image and applies a model is made up of a preproc-step, a model call, and a post porocess step. Generally these would run in different defices, with asynchrony ideal in a batch setting. Being able to write this as a function and test it as a function, and then have it be decomposed for production run on btaches seems small but is  huge added value, in my opinion. pushing us closer to the vision of avoiding re-writing when moving from notebook experiments to prod runs.
</blockquote></details>

Important discussion, but let's take it offline.

orm011:
*[`pixeltable/functions/openai.py` line 111 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtwVU2F0KGe_l8pWDum:-NtwVU2F0KGe_l8pWDun:b-zdmnou) ([raw file](https://github.com/pixeltable/pixeltable/blob/47556959f0460cfb97a1d69feec50113a12aadd8/pixeltable/functions/openai.py#L111)):*
> ```Python
> 
> 
> @pxt.udf(batch_size=32, return_type=ts.ArrayType((None,), dtype=ts.FloatType()))
> ```

I see this is carried over from the previous version, but I'm wondering where we're getting these batch sizes from.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95)* status: 0 of 27 files reviewed, 16 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/type_system.py` line 245 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-Ntx7DAK760HW7QfSoqr:-Ntx7DAK760HW7QfSoqs:bu7gt46) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/type_system.py#L245)):*
> ```Python
>         if t in _python_type_to_column_type:
>             return _python_type_to_column_type[t]
>         elif isinstance(t, typing._UnionGenericAlias) and t.__args__[1] is type(None):
> ```

explanatory comment missing, this is somewhat subtle
___
*[`pixeltable/func/udf.py` line 77 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-Ntx8hey9DR3MxDgl7WA:-Ntx8hey9DR3MxDgl7WB:b6ecq5q) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/func/udf.py#L77)):*
> ```Python
> 
> 
> # Constructs a `CallableFunction` or `BatchedFunction`, depending on the
> ```

move to docstring
___
*[`pixeltable/func/udf.py` line 82 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-Ntx8o3V11OmPHGZS5gA:-Ntx8o3V11OmPHGZS5gB:be6pfmo) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/func/udf.py#L82)):*
> ```Python
> # `substitute_fn`.
> def make_function(
>     eval_fn: Callable,
> ```

Why eval_fn and not py_fn? I chose eval_fn initially to contrast it with the aggregation-related functions that were hanging out in Function, but that's no longer the case.
___
*[`pixeltable/functions/fireworks.py` line 40 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxA-NeCwMrgzutIrdI:-NtxA-NeCwMrgzutIrdJ:bueh7ys) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/functions/fireworks.py#L40)):*
> ```Python
> 
> 
> def initialize():
> ```

use blank lines sensibly

(here and elsewhere)
___
*[`pixeltable/functions/openai.py` line 60 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1530714763:-NtxBMGf9z87zb4o7izg:b4s8v3f) ([raw file](https://github.com/pixeltable/pixeltable/blob/43313b439025323e0ce5e4112171906cd832c46a/pixeltable/functions/openai.py#L60)):*

I can take this out if it's really controversial, but I also don't think it's doing a lot of harm. One intended purpose is to facilitate streamlined tutorials that demonstrate the usefulness of pixeltable features without distracting ppl with the nuances of the OpenAI API. Of course, the fully general API will be made available either way.

We should probably involve Pierre in this discussion. He might have opinions since there's a product question here (how far do we go in regards to providing "streamlined" wrapper APIs in addition to the base APIs? Does this make the product easier to use, or just more confusing?)
</blockquote></details>

Yeah, let's discuss this.
___
*[`pixeltable/functions/openai.py` line 16 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxAxG28mQPq079Y1UX:-NtxAxG28mQPq079Y1UY:bzq226) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/functions/openai.py#L16)):*
> ```Python
> 
> @pxt.udf
> def completions_create(
> ```

Do we really want to call all of these things xxx_create?
___
*[`pixeltable/tests/test_functions.py` line 263 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxBoGZ2DOeuyj9mkfH:-NtxBoG_-G0GHG7bvtAn:b-xcs09c) ([raw file](https://github.com/pixeltable/pixeltable/blob/05a4ce54a0d626cf2603c94ee6343b4dedc1392c/pixeltable/tests/test_functions.py#L263)):*
> ```Python
>         verify_row(t.tail(1)[0])
> 
>         with pytest.raises(ValueError) as exc_info:
> ```

Leave as a todo somewhere that we also test bad parameters (for all external API functions)
___
*[`pixeltable/tests/test_udf.py` line 16 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxCBFC7asXzDVc23uc:-NtxCBFC7asXzDVc23ud:bxark1h) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/tests/test_udf.py#L16)):*
> ```Python
>     # correct error messages.
>     def test_invalid_udfs(self):
> ```

This blank line violates the Google Python style guide

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95)* status: 0 of 27 files reviewed, 16 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/type_system.py` line 245 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-Ntx7DAK760HW7QfSoqr:-NtxQguw7RFPjhBpkieU:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/type_system.py#L245)):*

explanatory comment missing, this is somewhat subtle
</blockquote></details>

Done.
___
*[`pixeltable/func/udf.py` line 77 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-Ntx8hey9DR3MxDgl7WA:-NtxR2NvCN4blr9ctrRO:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/func/udf.py#L77)):*

move to docstring
</blockquote></details>

Done.
___
*[`pixeltable/func/udf.py` line 82 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-Ntx8o3V11OmPHGZS5gA:-NtxR3aEDuJwO9b8jO9-:bxjm8k6) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/func/udf.py#L82)):*

Why eval_fn and not py_fn? I chose eval_fn initially to contrast it with the aggregation-related functions that were hanging out in Function, but that's no longer the case.
</blockquote></details>

I don't love `py_fn`. There are a lot of python functions floating around, and I'd prefer something more specific. I agree `eval_fn` isn't great either, though. How does `decorated_fn` sound?
___
*[`pixeltable/functions/fireworks.py` line 40 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxA-NeCwMrgzutIrdI:-NtxRbz36GxQjdQbUzrF:b7tvicx) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/functions/fireworks.py#L40)):*

use blank lines sensibly

(here and elsewhere)
</blockquote></details>

Fixed.

Personally I like the way the blank lines break up the structure of complicated functions. But I agree it's more important to follow the style conventions.
___
*[`pixeltable/functions/openai.py` line 60 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1530714763:-NtxSpIN3JlT-uqXG-QX:b-1c912l) ([raw file](https://github.com/pixeltable/pixeltable/blob/43313b439025323e0ce5e4112171906cd832c46a/pixeltable/functions/openai.py#L60)):*

Yeah, let's discuss this.
</blockquote></details>

I'll take it out for now, so it doesn't block the PR.
___
*[`pixeltable/functions/openai.py` line 111 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtwVU2F0KGe_l8pWDum:-NtxStyX8eenhAc5o-75:brlvxt3) ([raw file](https://github.com/pixeltable/pixeltable/blob/47556959f0460cfb97a1d69feec50113a12aadd8/pixeltable/functions/openai.py#L111)):*

I see this is carried over from the previous version, but I'm wondering where we're getting these batch sizes from.
</blockquote></details>

Not sure there's much science behind them. Marcel may have more insight. But I'd consider them a placeholder for now.
___
*[`pixeltable/functions/openai.py` line 16 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxAxG28mQPq079Y1UX:-NtxSN-Y7pdaWXVOqCw6:b50zbgo) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/functions/openai.py#L16)):*

Do we really want to call all of these things xxx_create?
</blockquote></details>

No. It's not a good name. I take the blame for it :)

Once this PR is done I'm going to go through all the OpenAI, Together, and Fireworks code and polish it. Trying to avoid scope creep on this PR though (especially for stuff that preceded it)
___
*[`pixeltable/tests/test_functions.py` line 263 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxBoGZ2DOeuyj9mkfH:-NtxUQHn0Y0Wc65jr0EV:bvk76e6) ([raw file](https://github.com/pixeltable/pixeltable/blob/05a4ce54a0d626cf2603c94ee6343b4dedc1392c/pixeltable/tests/test_functions.py#L263)):*

Leave as a todo somewhere that we also test bad parameters (for all external API functions)
</blockquote></details>

I'll create a ticket for it.
___
*[`pixeltable/tests/test_udf.py` line 16 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxCBFC7asXzDVc23uc:-NtxT7bS23wqp28t8WGs:bt2vr1b) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/tests/test_udf.py#L16)):*

This blank line violates the Google Python style guide
</blockquote></details>

Fixed.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95)* status: 0 of 27 files reviewed, 16 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/func/udf.py` line 82 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-Ntx8o3V11OmPHGZS5gA:-NtxZWHEEV24t_FYQ9qD:b-pvfzkq) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/func/udf.py#L82)):*

I don't love `py_fn`. There are a lot of python functions floating around, and I'd prefer something more specific. I agree `eval_fn` isn't great either, though. How does `decorated_fn` sound?
</blockquote></details>

I changed it to `decorated_fn`. Let me know what you think. I tried `py_fn` also, but the it gets really confusing because `py_fn` is also the name of the thing passed to `CallableFunction`, and that `py_fn` is sometimes the same as `decorated_fn`, but sometimes not. So it's a lot clearer to call it something else, and `decorated_fn` is unambiguous since no matter how it later gets interpreted, it's always the thing being decorated.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95)* status: 0 of 27 files reviewed, 16 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/functions/openai.py` line 16 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxAxG28mQPq079Y1UX:-NtxsVXr7YoWjP8i4jY0:b-jeu3yw) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/functions/openai.py#L16)):*

No. It's not a good name. I take the blame for it :)

Once this PR is done I'm going to go through all the OpenAI, Together, and Fireworks code and polish it. Trying to avoid scope creep on this PR though (especially for stuff that preceded it)
</blockquote></details>

Could we please keep the function names we already had? That way the existing notebooks don't break.

Also, in the OpenAI API the endpoint is "chat/completions":
https://platform.openai.com/docs/api-reference/chat/create

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95)* status: 0 of 28 files reviewed, 16 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/functions/openai.py` line 16 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/95#-NtxAxG28mQPq079Y1UX:-Nu-_wZV-HWkI3tiSbcK:bxuser) ([raw file](https://github.com/pixeltable/pixeltable/blob/edb275f95a1252460770f8792b76029f1971e9a2/pixeltable/functions/openai.py#L16)):*

Could we please keep the function names we already had? That way the existing notebooks don't break.

Also, in the OpenAI API the endpoint is "chat/completions":
https://platform.openai.com/docs/api-reference/chat/create
</blockquote></details>

Fixed. I went through and changed all the pixeltable function names to align them with the corresponding REST APIs. Note the prior name of the OpenAI endpoint was inconsistent (it was `chat_completion` not `chat_completions`), so I updated the notebook to be consistent.

pixeltable/func/batched_function.py
```diff
@@ -18,3 +20,31 @@ def get_batch_size(self, *args: Any, **kwargs: Any) -> Optional[int]:
     def invoke(self, arg_batches: List[List[Any]], kwarg_batches: Dict[str, List[Any]]) -> List[Any]:
         """Invoke the function for the given batch and return a batch of results"""
         raise NotImplementedError
+
+
+class ExplicitBatchedFunction(BatchedFunction):
```
mkornacker, line 25: Missing class comment

pixeltable/functions/fireworks.py
```diff
@@ -0,0 +1,61 @@
...
+    kwargs_not_none = dict(filter(lambda x: x[1] is not None, kwargs.items()))
+    print(kwargs_not_none)
+    return fireworks.client.Completion.create(
+        model=model,
+        prompt_or_messages=prompt,
+        **kwargs_not_none
+    ).dict()
+
+
+def initialize():
```
mkornacker, line 40: Shouldn't this be happening in Env? (similar to _create_together_client)

pixeltable/tests/test_functions.py
```diff
@@ -116,19 +116,34 @@ def skip_test_if_no_openai_client() -> None:
...
         t.add_column(output_text=t.output.output.choices[0].text)
         t.insert([{'input': 'I am going to the '}])
         result = t.select(t.output_text).collect()['output_text'][0]
         assert len(result) > 0
 
+    def test_fireworks(self, test_client: pt.Client) -> None:
+        skip_test_if_not_installed('fireworks')
+        try:
+            from pixeltable.functions.fireworks import initialize
+            initialize()
```
mkornacker, line 136: We shouldn't have to call this explicitly.

pixeltable/tests/test_udf.py
```diff
@@ -0,0 +1,44 @@
...
+import pixeltable as pxt
+import pixeltable.exceptions as excs
+from pixeltable.func import Batch
+
+
+class TestUdf:
+
+    # Test that various invalid udf definitions generate
+    # correct error messages.
+    def test_invalid_udfs(self):
```
mkornacker: Move to test_function.py?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/95)* status: 0 of 28 files reviewed, 20 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/func/batched_function.py` line 25 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1541755326:-Nu0DgyGEvW7Or9JSfkz:b-4sghf6) ([raw file](https://github.com/pixeltable/pixeltable/blob/24ac8776106fc89f4ebbde71f1c83f7c8915f3a8/pixeltable/func/batched_function.py#L25)):*

Missing class comment
</blockquote></details>

Done. Incidentally, I view `ExplicitBatchedFunction` as a temporary hack. My hope is that it can just become `BatchedFunction` once NOS is eliminated. (NOS is now the only other subclass of `BatchedFunction`, and I'd argue that going forward all batched functions should be defined via the decorator, and we should extend the decorator with added capabilities as needed.)
___
*[`pixeltable/functions/fireworks.py` line 40 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1541795900:-Nu0E3MIDzOPo55YsfNG:bbmthn9) ([raw file](https://github.com/pixeltable/pixeltable/blob/24ac8776106fc89f4ebbde71f1c83f7c8915f3a8/pixeltable/functions/fireworks.py#L40)):*

Shouldn't this be happening in Env? (similar to _create_together_client)
</blockquote></details>

I don't love the existing Env pattern for OpenAI/Together. As we go from 2 of these to 3 to 6 or more, Env is going to become a big mess. I was aiming for better locality here, but I'm aware it's still not perfect.

I have something in mind that I'm going to propose as a separate PR to unify these patterns. If it's ok with you, I'd like to not block this PR on resolving that (trying to avoid scope creep for this PR).

Another option would be to remove Fireworks from this PR entirely, but that's extra work and it's not really doing any harm, so my preference is to fix-forward.
___
*[`pixeltable/tests/test_functions.py` line 136 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1541800753:-Nu0EceK05x82ydWyDS0:b-8dh2r4) ([raw file](https://github.com/pixeltable/pixeltable/blob/24ac8776106fc89f4ebbde71f1c83f7c8915f3a8/pixeltable/tests/test_functions.py#L136)):*

We shouldn't have to call this explicitly.
</blockquote></details>

Again, let me address this in a downstream PR if that's ok.
___
*[`pixeltable/tests/test_udf.py` line 15 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/95#gh-1541807322:-Nu0ElZ74VniOaaRhPHv:b838xy6) ([raw file](https://github.com/pixeltable/pixeltable/blob/24ac8776106fc89f4ebbde71f1c83f7c8915f3a8/pixeltable/tests/test_udf.py#L15)):*

Move to test_function.py?
</blockquote></details>

Sure, I don't feel super strongly about it.

mkornacker (approved):

PR #96, opened 2024-03-16
Restructuring of the Function class hierarchy

mkornacker:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/96)


pixeltable/functions/image.py
```diff
@@ -0,0 +1,16 @@
+import base64
+
+import PIL.Image
+
+from pixeltable.type_system import ImageType, StringType
+import pixeltable.func as func
+
+
+@func.udf(return_type=StringType(), param_types=[ImageType(), StringType()])
```
aaron-siegel, line 9: This is a very welcome change. I much prefer the decorator pattern to the older ways of doing things.

pixeltable/functions/pil/image.py
```diff
@@ -55,73 +46,62 @@ def _crop_return_type(bound_args: Dict[str, Any]) -> ColumnType:
...
+    PIL.Image.Image.getprojection, return_type=JsonType(), param_types=[ImageType()])
+histogram = func.make_module_function(
+    PIL.Image.Image.histogram, return_type=JsonType(), param_types=[ImageType(), ImageType(), JsonType()])
+quantize = func.make_module_function(
+    PIL.Image.Image.quantize, return_type=ImageType(),
+    param_types=[ImageType(), IntType(), IntType(nullable=True), IntType(), IntType(nullable=True), IntType()])
+reduce = func.make_module_function(
+    PIL.Image.Image.reduce, return_type=ImageType(), param_types=[ImageType(), IntType(), JsonType()])
+transpose = func.make_module_function(
+    PIL.Image.Image.transpose, return_type=_caller_return_type, param_types=[ImageType(), IntType()])
```
aaron-siegel: I have to say I don't love this pattern. In many of these cases, the relevant function can be replaced by a use of `apply`, e.g.,

`t.img.getprojection`

is identical in effect to

`t.img.apply(PIL.Image.Image.getprojection)`

We should think through our general philosophy regarding the "pixeltable library". If our general design pattern is to explicitly define such operations in the pixeltable library, we risk ending up with a large, sprawling stdlib that we have to maintain, one that interposes a new library the users have to come to terms with, and that will always be incomplete. An alternative approach is to use `apply` for these "nuts and bolts" operations that end up being no-op delegations to the underlying python library, and (with perhaps a few exceptions) reserve explicit pixeltable library declarations for AI related functionality.

I also don't love the make_* pattern. I'd much prefer to see all operations defined with decorators, using a common pattern.

pixeltable/functions/video.py
```diff
@@ -30,7 +30,15 @@
...
-def _extract_audio(
+
+_extract_audio_param_types = [
+    ts.VideoType(nullable=False),
+    ts.IntType(nullable=False),
+    ts.StringType(nullable=False),
+    ts.StringType(nullable=False)
+]
+@func.udf(return_type=ts.AudioType(nullable=True), param_types=_extract_audio_param_types)
+def extract_audio(
```
aaron-siegel, line 41: Yes! This makes sense to me.

pixeltable/func/callable_function.py
```diff
@@ -0,0 +1,117 @@
...
+    # determine whether this is a named function in a module
+    if py_fn.__module__ == '__main__' or not py_fn.__name__.isidentifier():
+        return make_stored_function(py_fn, return_type=return_type, param_types=param_types)
+    else:
+        return make_module_function(
+            py_fn, return_type=return_type, param_types=param_types, module_path=py_fn.__module__)
+
+
+def udf(*, return_type: ts.ColumnType, param_types: List[ts.ColumnType], py_fn: Optional[Callable] = None) -> Callable:
+    """Returns decorator to create a CallableFunction from a function definition.
```
aaron-siegel, line 88: I generally find it cleaner to implement decorators as classes with a `__call__` method, rather than as functions. (Note a decorator can be any `Callable`.)

I'd also prefer to call this `pxt.func` rather than `func.udf`. The term "UDF" sounds archaic to my ears; in my view, they're all just functions.

aaron-siegel (commented):
This is a large change, and I've only left some cursory initial comments. I'll give it another read-through on Monday.

I'm very happy to see a lot of the cumbersome function definition patterns replaced with decorator interfaces, which I find much cleaner!

My number one point of design feedback is regarding the "class hierarchy" of `Function` objects. I definitely think the bespoke `OpenAIFunction`, `TogetherFunction`, `HuggingfaceFunction` etc should go away. I don't see what utility they add. See #95 for a sense of what I'd envision here.

But I'd go a step further and question whether a `BatchedFunction` subclass is the right abstraction. Python often eschews subclasses as a way to provide augmented functionality, and I suspect they may be cumbersome here. An alternative pattern would be to provide augmented functionality via mixins with a decorator interface. In this pattern, one would first define a function in the usual way,

```
@func.udf
def my_function(...):
```

then later declare a batching capability,

```
@my_function.batcher
def my_function_batcher(...):
```

a la property getters and setters. This isn't a super well thought out proposal yet, more of a vague sense that a decorator pattern would be more flexible and easier for users to follow / imitate in their own code.

I also wonder whether we need so many library functions, or if it wouldn't be better to use `apply` in cases where the library function is a no-op delegate to the underlying python library. There's a more detailed comment about this below.

This is a huge change that's going to block things if it bakes for too long, so I'm not expecting the above suggestions to be resolved before committing it; we could commit this PR and then factor the suggestions out as discussion points for further iteration.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/96)* status: 0 of 44 files reviewed, 11 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/expr.py` line 526 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-Ntkrp1D9K6esQ0jzhAK:-Ntkrp1D9K6esQ0jzhAL:b-13e2ac) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/exprs/expr.py#L526)):*
> ```Python
> 
>         if fn_type is None:
>             raise excs.Error((
> ```

Why is there an extra set of parentheses here?
___
*[`pixeltable/exprs/function_call.py` line 240 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-Ntks1W9-_Nqb_6v8QB4:-Ntks1W9-_Nqb_6v8QB5:b9n0ajq) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/exprs/function_call.py#L240)):*
> ```Python
>     @property
>     def is_window_fn_call(self) -> bool:
>         return isinstance(self.fn, func.AggregateFunction) and self.fn.allows_window and \
> ```

I actually prefer `self.fn.is_aggregate` to the use of `isinstance`. It's generally best practice to avoid relying on `isinstance` checks when it's feasible to do so; in this case, why not just use `self.is_agg_fn_call()`?
___
*[`pixeltable/exprs/function_call.py` line 250 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtksbkNCN44U_rBhv08:-NtksbkNCN44U_rBhv09:bplmt4p) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/exprs/function_call.py#L250)):*
> ```Python
>     @property
>     def is_agg_fn_call(self) -> bool:
>         return isinstance(self.fn, func.AggregateFunction)
> ```

I'd think a better pattern would be to define

`Function.is_agg_fn()`

or some such, as an abstract method, which is overriden to be `True` in `AggregateFunction` and `False` in `CallableFunction` and `BatchedFunction`. This would be a more logical object-oriented design and avoid the use of `isinstance` checks altogether.
___
*[`pixeltable/func/aggregate_function.py` line 46 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkzCkG8MK4_rqG639k:-NtkzCkG8MK4_rqG639l:b-7elw20) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L46)):*
> ```Python
>         init_params = [
>             inspect.Parameter(p.name, inspect.Parameter.KEYWORD_ONLY, default=p.default)
>             for p in list(inspect.signature(self.agg_cls.__init__).parameters.values())[1:]  # leave out self
> ```

for p in itertools.islice(inspect.signature(self.agg_cls.__init__).parameters.values(), 1)

will avoid the unnecessary list creation
___
*[`pixeltable/func/aggregate_function.py` line 51 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkzbQoEfRb5RkPmpB5:-NtkzbQoEfRb5RkPmpB6:b2d89he) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L51)):*
> ```Python
>         duplicate_params = set(p.name for p in init_params) & set(p.name for p in update_params)
>         if len(duplicate_params) > 0:
>             raise excs.Error((
> ```

Extra paren? Is this necessary because of the multi-segment string?
___
*[`pixeltable/func/aggregate_function.py` line 59 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkuNRe91tx8DSfHLX5:-NtkuNRe91tx8DSfHLX6:bfquk9q) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L59)):*
> ```Python
> 
>         params = [Parameter(p.name, update_types[i], p.kind) for i, p in enumerate(update_params)]
>         params.extend([Parameter(p.name, init_types[i], p.kind) for i, p in enumerate(init_params)])
> ```

Use generator instead of explicit list

`params.extend(Parameter(p.name, init_types[i], p.kind) for i, p in enumerate(init_params))`
___
*[`pixeltable/func/aggregate_function.py` line 86 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkurSz6ia0wSusS0iL:-NtkurT-AwfLJkFoaY7h:be5apj) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L86)):*
> ```Python
>             if not self.allows_window:
>                 raise excs.Error(f'Order_by invalid with an aggregate function that does not allow windows')
>             order_by_clause = kwargs['order_by']
> ```

Define 'order_by' and 'group_by' as string constants?
___
*[`pixeltable/func/aggregate_function.py` line 87 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkuzzT6mOTsgvGGi03:-NtkuzzT6mOTsgvGGi04:bt6p6lb) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L87)):*
> ```Python
>                 raise excs.Error(f'Order_by invalid with an aggregate function that does not allow windows')
>             order_by_clause = kwargs['order_by']
>             del kwargs['order_by']
> ```

Instead of lookup/del, it's a little cleaner to use

`order_by_clause = kwargs.pop('order_by', None)`

to remove 'order_by' from `kwargs` and assign it to `order_by_clause`, defaulting to `None` if it's not in `kwargs`.
___
*[`pixeltable/func/aggregate_function.py` line 102 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkzwJ81uRi7mrWv3DM:-NtkzwJ81uRi7mrWv3DN:b-z42z8w) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L102)):*
> ```Python
>         if 'group_by' in kwargs:
>             if not self.allows_window:
>                 raise excs.Error(f'Group_by invalid with an aggregate function that does not allow windows')
> ```

A lot of these error messages could be made more informative by stating more context in the message, e.g., in this case, state the function name:

Group_by invalid with aggregate function {my_func}, which does not allow windows
___
*[`pixeltable/func/callable_function.py` line 106 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkxpOK5WDTD7SLZFEl:-NtkxpOK5WDTD7SLZFEm:b-3sz42n) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/callable_function.py#L106)):*
> ```Python
> 
>     # the decorated function is only used for the signature/path and never executed
>     def dummy_decorator(dummy_py_fn: Callable) -> CallableFunction:
> ```

I'd propose doing away with `dummy_decorator` and always using concrete function bodies in udfs. That's a more transparent pattern, more consistent with what's expected out of a decorator, and visible to type checkers. There's a more detailed comment about this (with an example) elsewhere in the review.

(I'm fine merging the PR without resolving this one; we can discuss later when I fold in #95)
___
*[`pixeltable/functions/pil/image.py` line 17 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkwPP2EQ8UmZJq0xhF:-NtkwPP2EQ8UmZJq0xhG:b-1qi4d5) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/functions/pil/image.py#L17)):*
> ```Python
>     py_fn=PIL.Image.alpha_composite, return_type=ImageType(), param_types=[ImageType(), ImageType()])
> def alpha_composite(img1: PIL.Image.Image, img2: PIL.Image.Image) -> PIL.Image.Image:
>     pass
> ```

This strikes me as clunky, and a misuse of the decorator pattern. You're defining a "phantom" method only to throw it away and replace it with something else. It complicates the semantics of the `udf` decorator: sometimes it does something with the method body; sometimes it doesn't, depending on the arguments given.

I much prefer the following:
```
@func.udf(return_type=ImageType(), param_types=[ImageType(), ImageType(), ImageType()])
def composite(img1: PIL.Image.Image, img2: PIL.Image.Image, mask: PIL.Image.Image) -> PIL.Image.Image:
    return PIL.Image.composite(img1, img2, mask)
```
It's crystal-clear, and it aligns the usage of `@func.udf` under a simple, common pattern, with no real downsides. A bonus is that the type checker can verify that the signature of `PIL.Image.composite` is properly adapted. (As it stands, the signature of `PIL.Image.composite` is cloaked in indirection.)

I'd propose getting rid of the `py_fn` parameter and `dummy_decorator` and standardizing on concrete function bodies for all udfs.

NOTE: I'm fine if we merge this PR without resolving this one. If you agree with the suggestion I can roll it in to #95, since I'll need to refactor the udf pattern again anyway. (Or we can discuss further once this PR is merged.)

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/96)* status: 0 of 44 files reviewed, 11 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exprs/expr.py` line 526 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-Ntkrp1D9K6esQ0jzhAK:-NtmrrGt10J1zSYy55j8:bm863vl) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/exprs/expr.py#L526)):*

Why is there an extra set of parentheses here?
</blockquote></details>

It's needed for multi-line f-strings
___
*[`pixeltable/func/aggregate_function.py` line 46 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkzCkG8MK4_rqG639k:-NtmtKKZCf1lRUAmQf_I:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L46)):*

for p in itertools.islice(inspect.signature(self.agg_cls.__init__).parameters.values(), 1)

will avoid the unnecessary list creation
</blockquote></details>

Done.
___
*[`pixeltable/func/aggregate_function.py` line 51 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkzbQoEfRb5RkPmpB5:-NtmsS3f5ycER0FcP_Jm:b2nhswy) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L51)):*

Extra paren? Is this necessary because of the multi-segment string?
</blockquote></details>

Yes
___
*[`pixeltable/func/aggregate_function.py` line 59 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkuNRe91tx8DSfHLX5:-NtmtNbo219ZUIvSyXn3:b-ym6mgn) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L59)):*

Use generator instead of explicit list

`params.extend(Parameter(p.name, init_types[i], p.kind) for i, p in enumerate(init_params))`
</blockquote></details>

I want to use a list here, because it doesn't have any performance implications and lists can be examined in the debugger (generators can't).
___
*[`pixeltable/func/aggregate_function.py` line 86 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkurSz6ia0wSusS0iL:-NtmtzDzAlTgcUQWEoq7:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L86)):*

Define 'order_by' and 'group_by' as string constants?
</blockquote></details>

Done.
___
*[`pixeltable/func/aggregate_function.py` line 87 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkuzzT6mOTsgvGGi03:-Ntmu_4oExVVLEwU6LXz:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L87)):*

Instead of lookup/del, it's a little cleaner to use

`order_by_clause = kwargs.pop('order_by', None)`

to remove 'order_by' from `kwargs` and assign it to `order_by_clause`, defaulting to `None` if it's not in `kwargs`.
</blockquote></details>

Done.
___
*[`pixeltable/func/aggregate_function.py` line 102 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkzwJ81uRi7mrWv3DM:-NtmvkkT6Y6fK0xdNxL5:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/aggregate_function.py#L102)):*

A lot of these error messages could be made more informative by stating more context in the message, e.g., in this case, state the function name:

Group_by invalid with aggregate function {my_func}, which does not allow windows
</blockquote></details>

Done.
___
*[`pixeltable/func/callable_function.py` line 106 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/96#gh-1527211217:-Ntmvq261gla3tlh7End:b41xl1x) ([raw file](https://github.com/pixeltable/pixeltable/blob/ddaf229dc5d67102f809689ab2bc245143730548/pixeltable/func/callable_function.py#L106)):*

I generally find it cleaner to implement decorators as classes with a `__call__` method, rather than as functions. (Note a decorator can be any `Callable`.)

I'd also prefer to call this `pxt.func` rather than `func.udf`. The term "UDF" sounds archaic to my ears; in my view, they're all just functions.
</blockquote></details>

Let's discuss this in a team mtg, those seem like decisions beyond this PR.
___
*[`pixeltable/func/callable_function.py` line 106 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtkxpOK5WDTD7SLZFEl:-NtmwFvvEqHfMarEAww2:b-ph4ygc) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/func/callable_function.py#L106)):*

I'd propose doing away with `dummy_decorator` and always using concrete function bodies in udfs. That's a more transparent pattern, more consistent with what's expected out of a decorator, and visible to type checkers. There's a more detailed comment about this (with an example) elsewhere in the review.

(I'm fine merging the PR without resolving this one; we can discuss later when I fold in #95)
</blockquote></details>

I chose this for a reason, let's discuss offline.
___
*[`pixeltable/functions/pil/image.py` line 107 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/96#gh-1527209757:-NtmwWBHBXGvsUfIoail:b6r7qdm) ([raw file](https://github.com/pixeltable/pixeltable/blob/ddaf229dc5d67102f809689ab2bc245143730548/pixeltable/functions/pil/image.py#L107)):*

I have to say I don't love this pattern. In many of these cases, the relevant function can be replaced by a use of `apply`, e.g.,

`t.img.getprojection`

is identical in effect to

`t.img.apply(PIL.Image.Image.getprojection)`

We should think through our general philosophy regarding the "pixeltable library". If our general design pattern is to explicitly define such operations in the pixeltable library, we risk ending up with a large, sprawling stdlib that we have to maintain, one that interposes a new library the users have to come to terms with, and that will always be incomplete. An alternative approach is to use `apply` for these "nuts and bolts" operations that end up being no-op delegations to the underlying python library, and (with perhaps a few exceptions) reserve explicit pixeltable library declarations for AI related functionality.

I also don't love the make_* pattern. I'd much prefer to see all operations defined with decorators, using a common pattern.
</blockquote></details>

regarding removing make_callable_function():
I find 
fn = func.udf(return_type=col.col_type, param_types=[arg.col_type for arg in args])(col.compute_func)
harder to read than
        fn = func.make_callable_function(
            col.compute_func, return_type=col.col_type, param_types=[arg.col_type for arg in args])

regarding having image methods instead of using apply(): this is consistent with convenience functions in other libraries. For instance, in pandas you can write
df.groupby('A')['B'].sum()
instead of
df.groupby('A').agg({'B': 'sum'})
and I'm willing to bet that 99% of users will go with the former.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/96)* status: 0 of 44 files reviewed, 5 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/expr.py` line 526 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-Ntkrp1D9K6esQ0jzhAK:-NtnwRQc8NrvMUnOHXMG:b-gvjk1) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/exprs/expr.py#L526)):*

It's needed for multi-line f-strings
</blockquote></details>

Are you sure? I tested a multi-line f-string in the interpreter and it worked fine with single parens:
```
x = 4
y = 5
raise RuntimeError(
    f'{x}Test1'
    f'{y}Test2'
)
```
___
*[`pixeltable/func/callable_function.py` line 106 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/96#gh-1527211217:-NtnxTCb4X9sPNG7PKTC:bp6sqw5) ([raw file](https://github.com/pixeltable/pixeltable/blob/ddaf229dc5d67102f809689ab2bc245143730548/pixeltable/func/callable_function.py#L106)):*

Let's discuss this in a team mtg, those seem like decisions beyond this PR.
</blockquote></details>

Yeah. Actually you can ignore this comment, it's from a week ago and I've since come around on it.
___
*[`pixeltable/functions/pil/image.py` line 69 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/96#-NtnyzIGC1w4H-mb7RpR:-NtnyzIGC1w4H-mb7RpS:bbh4047) ([raw file](https://github.com/pixeltable/pixeltable/blob/14eeea93702b1fd432651fda7a34602cb0bf0c2a/pixeltable/functions/pil/image.py#L69)):*
> ```Python
> 
> # Image.resize()
> def resize_return_type(bound_args: Dict[str, Any]) -> ColumnType:
> ```

I'm confused about what these `*_return_type` methods are doing. The `return_type` param of `func.udf` is typed as a `ColumnType`, but we're passing it a `Callable`?

aaron-siegel (approved):
Approved, I'm ok with resolving any remaining comments / open questions downstream.

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/96)* status: 0 of 44 files reviewed, 5 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exprs/expr.py` line 526 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/96#-Ntkrp1D9K6esQ0jzhAK:-NtqJ6x30U7pAfYAQmqq:bbmr2tb) ([raw file](https://github.com/pixeltable/pixeltable/blob/b23c1afef88833612578e9a88474ca187a4816d0/pixeltable/exprs/expr.py#L526)):*

Are you sure? I tested a multi-line f-string in the interpreter and it worked fine with single parens:
```
x = 4
y = 5
raise RuntimeError(
    f'{x}Test1'
    f'{y}Test2'
)
```
</blockquote></details>

You're right! I'll fix this now and everything else as I run across it.

PR #97, opened 2024-03-20
Update dependencies for python 3.12 compatibility

aaron-siegel:
Updates dependencies for python 3.12 and adds 3.12 to CI.

The only significant change is to bump pillow to >= 10.0.

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/97)


orm011 (commented):
Maybe we should split the task of removing 3.9 support with providing 3.12. eg. Pytorch is currently only providing "experimental" support for 3.12. 
One larger issue I ran into just a few minutes ago is how narrow the allowed versions for dependencies are. This causes false-positive conflicts when installing other packages, like unstructured.

I wonder if trying to support 3.12 right now will make it harder to get the  false-positive conflict problem in order soon, without providing much upside:  (possibly needing to specify different ranges depending on the python version  as done for numpy  in your diff below).  

It seems to me false-positive version conflict is a blocker for open source release, while 3.12 is not

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/97#-:-NtSfgAwEZgFsRGhUmGW:byn6t4e)* status: 0 of 4 files reviewed, all discussions resolved

pyproject.toml
```diff
@@ -16,10 +16,10 @@ exclude = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.8.1"
-numpy = "^1.24.1"
-pandas = "^1.5.3"
-pillow = "^9.4.0"
+python = "^3.9"
+numpy = "^1.26.0"
+pandas = "^2.2.0"
```
mkornacker: Does 2.0.0 work here or do we really need 2.2.0? No need to be overly restrictive.

Same for pillow (10.0 vs 10.2)

orm011 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/97)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @aaron-siegel)

___
*[`poetry.lock` line 773 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/97#-NtSkW6rBr9pCZAWhpxr:-NtSkW6rBr9pCZAWhpxs:b2kx1we) ([raw file](https://github.com/pixeltable/pixeltable/blob/04294fd530865942722b752e480c8aeb73c20671/poetry.lock#L773)):*
> ```TOML
>     {version = ">=1.16,<2.0", markers = "python_version <= \"3.11\""},
>     {version = ">=1.26.0rc1,<2.0", markers = "python_version >= \"3.12\""},
> ]
> ```

question: 
if we simply provided ">= 1.16,<2.0" as numpy versions, would dependency resolution at install time (on the client side) use the appropriate numpy version depending on its python version ? If so, we wouldn't need to be this explicit, and we could focus on widening the version as much as possible.

orm011 (commented):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/97)* status: all files reviewed, 1 unresolved discussion (waiting on @aaron-siegel)

orm011 (changes requested):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/97)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`poetry.lock` line 3410 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/97#-Nvbo1y5-ULmQqu1nMhj:-Nvbo1y6Dnk5ANkk4rEM:b3ufssi) ([raw file](https://github.com/pixeltable/pixeltable/blob/3aa3c7f3959d2728de5b190a07bae787fc3493cf/poetry.lock#L3410)):*
> ```TOML
> [package.dependencies]
> numpy = [
>     {version = ">=1.22.4,<2", markers = "python_version < \"3.11\""},
> ```

How come we need this case-by-case breakdown for numpy but not for other libraries?  Wouldn't this be something the dependency resolution takes care of?

mkornacker (approved):

orm011 (approved):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/97)* status: all files reviewed, 3 unresolved discussions (waiting on @aaron-siegel)

PR #98, opened 2024-03-21
First stab at pdf document support.

orm011:
1. Add handling for pdf document type validation and loading. Add text iteration.
2. Add example data.
3. Add example notebook.

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/98)


pixeltable/utils/documents.py
```diff
@@ -9,31 +9,56 @@ class DocumentHandle:
     format: ts.DocumentType.DocumentFormat
     bs_doc: Optional['bs4.BeautifulSoup'] = None
     md_ast: Optional[Dict] = None
+    pdf_doc: Optional['pdfminer.PDFDocument'] = None
```
mkornacker: This isn't right, get_pdf_handle() returns a str

mkornacker (commented):
Tests?

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/98)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`pixeltable/utils/documents.py` line 12 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1536540994:-Nv7hbtW09an81ZS12j9:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/60b47daa5338c49d3c13e96b3bb9025d30d6e75e/pixeltable/utils/documents.py#L12)):*

This isn't right, get_pdf_handle() returns a str
</blockquote></details>

Done.

orm011:
I've added pdf and markdown now. (Also a notebook)

docs/tutorials/pdf_demo.ipynb
```diff
@@ -0,0 +1,150 @@
+{
```
aaron-siegel: We should be using hyphenated filenames in notebooks (and other files that are going to show up as doc pages). Underscores in URLs are just nasty...

pixeltable/iterators/document.py
```diff
@@ -29,44 +28,59 @@ class Separator(enum.Enum):
...
+    page_number: int = 0
+    bbox: Optional[Dict[str,float]] = None
+
 
 @dataclasses.dataclass
 class DocumentSection:
     """A single document chunk, according to some of the splitting criteria"""
     text: Optional[str]
-    md: Optional[DocumentSectionMd]
+    metadata: Optional[DocumentSectionMetadata]
```
aaron-siegel: Should these have `None` as a default?

pixeltable/iterators/document.py
```diff
@@ -29,44 +28,59 @@ class Separator(enum.Enum):
...
-        with open(document, 'r', encoding='utf8') as fh:
-            s = fh.read()
-            self._doc_handle = get_document_handle(s)
-            assert self._doc_handle is not None
-        self._separators = [Separator[s.upper()] for s in separators.split(',')]
-        self._md_fields = [ChunkMetadata[m.upper()] for m in metadata.split(',')] if len(metadata) > 0 else []
+        self._doc_handle = get_document_handle(document)
+        assert self._doc_handle is not None
+        self._separators = [Separator[s.strip().upper()] for s in separators.split(',')]
+        self._metadata_fields = [ChunkMetadata[m.strip().upper()] for m in metadata.split(',')] if len(metadata) > 0 else []
```
aaron-siegel: What happens if `separators` is not valid input? Where does that get validated and what kind of exception does the user see?

aaron-siegel: Why is the `if len(metadata) > 0 else []` necessary? Won't the behavior be identical if we simply remove that clause?

pixeltable/iterators/document.py
```diff
@@ -79,9 +93,13 @@ def __init__(
         if self._doc_handle.format == DocumentType.DocumentFormat.HTML:
             assert self._doc_handle.bs_doc is not None
             self._sections = self._html_sections()
-        else:
+        elif self._doc_handle.format == DocumentType.DocumentFormat.MD:
             assert self._doc_handle.md_ast is not None
             self._sections = self._markdown_sections()
+        else:
```
aaron-siegel: Add a comment indicating that this implies `self._doc_handle.format == DocumentType.DocumentFormat.PDF`

pixeltable/iterators/document.py
```diff
@@ -106,14 +124,14 @@ def input_schema(cls) -> Dict[str, ColumnType]:
     def output_schema(cls, *args: Any, **kwargs: Any) -> Tuple[Dict[str, ColumnType], List[str]]:
         schema = {'text': StringType()}
         if 'metadata' in kwargs and len(kwargs['metadata']) > 0:
-            md_fields = kwargs['metadata'].split(',')
+            md_fields = [field.strip() for field in kwargs['metadata'].split(',')]
```
aaron-siegel: Yuck, it just seems nasty and error prone that we're parsing this out a second time. And that we're using **kwargs here. Can we take the opportunity to refactor and do this right? Factor the `metadata` and `separators` parameter parsing out to avoid duplicated logic? and use typesafe method signatures when feasible? (I understand that some of these are legacy issues)

pixeltable/iterators/document.py
```diff
@@ -266,13 +288,33 @@ def process_element(el: Dict) -> Generator[DocumentSection, None, None]:
             yield from process_element(el)
         yield from emit()
 
+    def _pdf_sections(self) -> Generator[DocumentSection, None, None]:
+        """Create DocumentSections reflecting the pdf-specific separators"""
+        import fitz
+        doc : fitz.Document = self._doc_handle.pdf_doc
```
aaron-siegel: PEP8 style is no space before the :

pixeltable/tests/data/documents/layout-parser-paper.mmd
```diff
@@ -0,0 +1,268 @@
+# LayoutParser: A Unified Toolkit for Deep Learning Based Document Image Analysis
+
+Zejiang Shen
+
```
aaron-siegel, line 4: What is `.mmd`? Is this another kind of document? How is this represented in the document types enum?

pixeltable/utils/documents.py
```diff
@@ -9,31 +9,56 @@ class DocumentHandle:
...
         return DocumentHandle(format=ts.DocumentType.DocumentFormat.MD, md_ast=md_ast)
-    return None
+    # NB: get_html_handle must be last, because it will return a handle for md files as well,
+    # and we want to prefer the md handle.
+    bs_doc = get_html_handle(path)
+    if bs_doc is not None:
+        return DocumentHandle(format=ts.DocumentType.DocumentFormat.HTML, bs_doc=bs_doc)
 
-def get_html_handle(s: str) -> Optional['bs4.BeautifulSoup']:
+    return pdf_doc
```
aaron-siegel: What's this for?

pixeltable/utils/documents.py
```diff
@@ -9,31 +9,56 @@ class DocumentHandle:
...
+            return None
+        if doc.find() is None:
+            return None
+        return doc
 
-def get_markdown_handle(s: str) -> Optional[Dict]:
+def get_markdown_handle(path: str) -> Optional[Dict]:
     import mistune
+    with open(path, 'r', encoding='utf8') as file:
+        s = file.read()
```
aaron-siegel: If we have to try both markdown and HTML, we end up reading the entire file twice into a string. That's inefficient? If we're going to read the file into memory anyway, can't we do it once and then try both MD and html against the string?

aaron-siegel (changes requested):

pixeltable/iterators/document.py
orm011: Not sure, everywhere is setting them explicitly.

pixeltable/utils/documents.py
orm011: not needed anymore

orm011:
*[`pixeltable/tests/data/documents/layout-parser-paper.mmd` line 4 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562803784:-NvxiCBj5sVqr6RGPFG4:bwgecpq) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/pixeltable/tests/data/documents/layout-parser-paper.mmd#L4)):*

What is `.mmd`? Is this another kind of document? How is this represented in the document types enum?
</blockquote></details>

A kind of markdown.

orm011:
*[`pixeltable/iterators/document.py` line 83 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562792616:-NvxjhC2CHHc6AexkPc9:b-w4zgby) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/pixeltable/iterators/document.py#L83)):*

What happens if `separators` is not valid input? Where does that get validated and what kind of exception does the user see?
</blockquote></details>

have fixed up the parsing and added test for error msg

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/98)* status: 0 of 13 files reviewed, 11 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/iterators/document.py` line 83 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562793757:-Nvxi29_BxImBpS0N15b:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/pixeltable/iterators/document.py#L83)):*

Why is the `if len(metadata) > 0 else []` necessary? Won't the behavior be identical if we simply remove that clause?
</blockquote></details>

Done.
___
*[`pixeltable/iterators/document.py` line 99 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562795298:-Nvxi3Co7fBi7FgSfMXg:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/pixeltable/iterators/document.py#L99)):*

Add a comment indicating that this implies `self._doc_handle.format == DocumentType.DocumentFormat.PDF`
</blockquote></details>

Done.
___
*[`pixeltable/iterators/document.py` line 127 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562798064:-Nvxi4Yb3fZIkpvcWcUM:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/pixeltable/iterators/document.py#L127)):*

Yuck, it just seems nasty and error prone that we're parsing this out a second time. And that we're using **kwargs here. Can we take the opportunity to refactor and do this right? Factor the `metadata` and `separators` parameter parsing out to avoid duplicated logic? and use typesafe method signatures when feasible? (I understand that some of these are legacy issues)
</blockquote></details>

Done.
___
*[`pixeltable/iterators/document.py` line 294 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562801174:-Nvxi8jI2kmcAjL3BYd-:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/pixeltable/iterators/document.py#L294)):*

PEP8 style is no space before the :
</blockquote></details>

Done.
___
*[`pixeltable/utils/documents.py` line 46 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562807879:-NvxiOvm7jrFrUh_9YJO:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/pixeltable/utils/documents.py#L46)):*

If we have to try both markdown and HTML, we end up reading the entire file twice into a string. That's inefficient? If we're going to read the file into memory anyway, can't we do it once and then try both MD and html against the string?
</blockquote></details>

Done.
___
*[`docs/tutorials/pdf_demo.ipynb` line 1 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562785208:-NvxiQw81d_vei5sE29f:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/docs/tutorials/pdf_demo.ipynb#L1)):*

We should be using hyphenated filenames in notebooks (and other files that are going to show up as doc pages). Underscores in URLs are just nasty...
</blockquote></details>

Done.

orm011:
*[`pixeltable/iterators/document.py` line 47 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562787808:-Nvxjs6vAIWfL3VKwnkh:b-pwmnlg) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/pixeltable/iterators/document.py#L47)):*

Not sure, I'll leave it as not none by default since its only used internally, and each user can decide what it should be.
</blockquote></details>

I meant: I ended up not providing a default so that this is explicitly decided at the point of use

mkornacker (changes requested):
> Reviewed 3 of 10 files at r4.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/98)* status: 3 of 13 files reviewed, 26 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/type_system.py` line 882 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwLz0fW6zfUU8haBath:-NwLz0fW6zfUU8haBati:b-hetmar) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/type_system.py#L882)):*
> ```Python
> # functions. (Since Python functions do not necessarily have type hints, this
> # should always be an optional/convenience inference.)
> _python_type_to_column_type: dict[type, ColumnType] = {
> ```

This went away
___
*[`pixeltable/iterators/document.py` line 182 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM00Uf32KC0yfn8msq:-NwM00Uf32KC0yfn8msr:b-tn7jb1) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L182)):*
> ```Python
>         if Separator.TOKEN_LIMIT in separators or Separator.CHAR_LIMIT in separators:
>             if 'token_limit' in separators and 'char_limit' in separators:
>                 raise Error('Cannot specify both "token_limit" and "char_limit" separators')
> ```

Then let's use Separator.TOKEN_LIMIT.name.lower(), etc., here
___
*[`pixeltable/iterators/document.py` line 333 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM1ckw1Q0WK41B87Xh:-NwM1ckw1Q0WK41B87Xi:bqrx2w7) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L333)):*
> ```Python
>         emit_on_page = Separator.PAGE in self._separators or emit_on_paragraph
> 
>         accumulated_text = []  # invariant: all elements are ftfy clean and non-empty
> ```

separate following function w/ blank line
___
*[`pixeltable/iterators/document.py` line 348 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM1zISBsyJI2iMnyLa:-NwM1zISBsyJI2iMnyLb:b92a8x2) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L348)):*
> ```Python
>                 x1, y1, x2, y2, text, _, _ = block
>                 _add_cleaned_text(text)
>                 if accumulated_text and emit_on_paragraph:
> ```

So basically a block is a paragraph? Is that always true? Add comment.
___
*[`pixeltable/iterators/document.py` line 349 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM1xcU72wX5Yk5kzZH:-NwM1xcU72wX5Yk5kzZI:blqgiav) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L349)):*
> ```Python
>                 _add_cleaned_text(text)
>                 if accumulated_text and emit_on_paragraph:
>                     bbox = {'x1': x1, 'y1': y1,
> ```

single line
___
*[`pixeltable/iterators/document.py` line 360 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2Inj7flm7nsHlx3h:-NwM2Ink9ZIaHdhWjN9s:bh92j5d) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L360)):*
> ```Python
> 
>         if accumulated_text and not emit_on_page:
>             yield DocumentSection(text=_emit_text(),
> ```

Weird indentation
___
*[`pixeltable/iterators/document.py` line 387 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2OCl4bShntqJvkT8:-NwM2OCl4bShntqJvkT9:b-57tggz) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L387)):*
> ```Python
>             while start_idx < len(tokens):
>                 end_idx = min(start_idx + self._limit, len(tokens))
>                 while end_idx > start_idx:  # find a cutoff point that doesnt break utf encoding of sequence
> ```

doesn't
___
*[`pixeltable/iterators/document.py` line 397 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2gft5XL7x9qQMB1q:-NwM2gft5XL7x9qQMB1r:b-rl1emf) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L397)):*
> ```Python
> 
>                 assert end_idx > start_idx
>                 assert text
> ```

text is not None? I like being explicit
___
*[`pixeltable/tests/test_document.py` line 31 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2rNz1GAtPa-WUw9R:-NwM2rNz1GAtPa-WUw9S:b-lvetp7) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L31)):*
> ```Python
>     full_text = '\n'.join(full_text_parts)
>     return ftfy.fix_text(full_text)
> ```

Single-line separation is standard, no?
___
*[`pixeltable/tests/test_document.py` line 33 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2z4s2zv7x1ug9GUF:-NwM2z4s2zv7x1ug9GUG:bipobqx) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L33)):*
> ```Python
> 
> 
> def normalize(text: str):
> ```

missing return type (fix that here and elsewhere)
___
*[`pixeltable/tests/test_document.py` line 34 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2xm00SsQKa_aL9RW:-NwM2xm00SsQKa_aL9RX:b-yswiky) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L34)):*
> ```Python
> 
> def normalize(text: str):
>     """ for comparison """
> ```

unclear comment
___
*[`pixeltable/tests/test_document.py` line 84 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM3Qu9BHgi98mhQQ95:-NwM3Qu9BHgi98mhQQ96:b93l7aw) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L84)):*
> ```Python
>             assert 'Invalid separator' in str(exc_info.value)
> 
>         invalid_metadata = ['chapter']  # non existent metadata
> ```

additional test case: mix valid with invalid (same for the separators)
___
*[`pixeltable/tests/test_document.py` line 90 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM3mY50zggW2uhn1d-:-NwM3mY50zggW2uhn1d0:b-7vmkhz) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L90)):*
> ```Python
>             assert 'Invalid metadata' in str(exc_info.value)
> 
>     def test_pdf_splitter(self, test_client: pxt.Client) -> None:
> ```

why can't test_doc_splitter be extended? looks like copy & paste
___
*[`pixeltable/tests/test_document.py` line 103 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM3apq74NdzITrc2rg:-NwM3apq74NdzITrc2rh:b-8vh407) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L103)):*
> ```Python
>             level2 = ['', 'char_limit', 'token_limit']
> 
>             for sep1 in level1:
> ```

add comments explaining what is being tested
___
*[`pixeltable/utils/documents.py` line 15 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM44f71j-H9K1gMopW:-NwM44f71j-H9K1gMopX:b-m96rlm) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/utils/documents.py#L15)):*
> ```Python
> 
> def get_document_handle(path: str) -> Optional[DocumentHandle]:
>     # NB: try pdf first, because correct PDF must be opened in binary mode
> ```

Get rid of NB (here now, everywhere else, and going forward in all future code). If it's not worth noting, it shouldn't be here in the first place.

orm011:
*[`pixeltable/tests/test_document.py` line 90 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM3mY50zggW2uhn1d-:-NwQ13YB7Xl1_DKRvIEH:b-g6737k) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L90)):*

why can't test_doc_splitter be extended? looks like copy & paste
</blockquote></details>

test_doc_splitter is trying to do too much at the same time, and its gotten unwieldy,  it skips a bunch of checks that should be done. 
for example, metadata checks. these are different for every type of document. in the pdf test I can focus on the pdf functionality.

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/98)* status: 3 of 14 files reviewed, 26 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/type_system.py` line 882 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwLz0fW6zfUU8haBath:-NwPdvhRCFjHL04iqsUV:brpmqb3) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/type_system.py#L882)):*

This went away
</blockquote></details>

removed
___
*[`pixeltable/iterators/document.py` line 182 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM00Uf32KC0yfn8msq:-NwMHUDl1swjnbTsnI2S:bs7wumx) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L182)):*

Then let's use Separator.TOKEN_LIMIT.name.lower(), etc., here
</blockquote></details>

Done. 
___
*[`pixeltable/iterators/document.py` line 333 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM1ckw1Q0WK41B87Xh:-NwMHnuoBz-RjNOFrWvI:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L333)):*

separate following function w/ blank line
</blockquote></details>

Done.
___
*[`pixeltable/iterators/document.py` line 348 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM1zISBsyJI2iMnyLa:-NwMHqK90kL4ygu9rVIm:b-m7kjx4) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L348)):*

So basically a block is a paragraph? Is that always true? Add comment.
</blockquote></details>

It's not exactly the same concept, but pymupdf agrees is close. I've added a comment.
___
*[`pixeltable/iterators/document.py` line 349 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM1xcU72wX5Yk5kzZH:-NwMJ1de5_zVRe2xDmFU:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L349)):*

single line
</blockquote></details>

Done.
___
*[`pixeltable/iterators/document.py` line 360 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2Inj7flm7nsHlx3h:-NwMJ7eaBJ2ufoCHyBmw:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L360)):*

Weird indentation
</blockquote></details>

Done.
___
*[`pixeltable/iterators/document.py` line 387 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2OCl4bShntqJvkT8:-NwMJpdU3eC_iKaFYOF7:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L387)):*

doesn't
</blockquote></details>

Done.
___
*[`pixeltable/iterators/document.py` line 397 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2gft5XL7x9qQMB1q:-NwMJsHPEz-VqUHbz-3b:bpwodsw) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/iterators/document.py#L397)):*

text is not None? I like being explicit
</blockquote></details>

for a string type, its boolean value means text is not None and text != '', its different. 
___
*[`pixeltable/tests/test_document.py` line 31 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2rNz1GAtPa-WUw9R:-NwMKkueBXSpJcMGC_Dr:bwzkvel) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L31)):*

Single-line separation is standard, no?
</blockquote></details>

unsure about what you mean?
___
*[`pixeltable/tests/test_document.py` line 33 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2z4s2zv7x1ug9GUF:-NwMKnp819Hd-zQRIfij:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L33)):*

missing return type (fix that here and elsewhere)
</blockquote></details>

Done.
___
*[`pixeltable/tests/test_document.py` line 34 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM2xm00SsQKa_aL9RW:-NwMLC-pC9bYNrmsX6VB:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L34)):*

unclear comment
</blockquote></details>

Done.
___
*[`pixeltable/tests/test_document.py` line 84 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM3Qu9BHgi98mhQQ95:-NwMQK2WDtKVeN5RBc-a:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L84)):*

additional test case: mix valid with invalid (same for the separators)
</blockquote></details>

Done.
___
*[`pixeltable/tests/test_document.py` line 90 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM3mY50zggW2uhn1d-:-NwQHZ1C93y-WmevR0GR:b-cx3lu8) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L90)):*

test_doc_splitter is trying to do too much at the same time, and its gotten unwieldy,  it skips a bunch of checks that should be done. 
for example, metadata checks. these are different for every type of document. in the pdf test I can focus on the pdf functionality.
</blockquote></details>

I just spent some time merging them, some improvements made it down to test_doc_splitter (eg, test also with sep1 = '', which is valid but was not being tested. And test token limits are honored.  

This test is a bit complicated, and the doc abstraction too. for example, token limit and char limit stuff is really an add on on top of the semantic splits,  and in general would be more useful as a re-usable component on its own. Anyway, thats for some other time. The Document type is a little odd, its kind of a 'union' type.
___
*[`pixeltable/tests/test_document.py` line 103 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM3apq74NdzITrc2rg:-NwQ9JBjDpFC2xFMyrGt:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/tests/test_document.py#L103)):*

add comments explaining what is being tested
</blockquote></details>

Done.
___
*[`pixeltable/utils/documents.py` line 15 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwM44f71j-H9K1gMopW:-NwPcxGi5CeDAVEAFc5U:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/utils/documents.py#L15)):*

Get rid of NB (here now, everywhere else, and going forward in all future code). If it's not worth noting, it shouldn't be here in the first place.
</blockquote></details>

Done.

pixeltable/iterators/document.py
```diff
@@ -86,37 +86,41 @@ class DocumentSplitter(ComponentIterator):
...
         `limit`: the maximum number of tokens or characters in each chunk if 'token_limit' or 'char_limit' is specified.
     """
     METADATA_COLUMN_TYPES = {
-        ChunkMetadata.TITLE: StringType(),
-        ChunkMetadata.HEADING: JsonType(),
-        ChunkMetadata.SOURCELINE: IntType(),
-        ChunkMetadata.PAGE: IntType(),
-        ChunkMetadata.BOUNDING_BOX: JsonType(),
+        ChunkMetadata.TITLE: StringType(nullable=True),
+        ChunkMetadata.HEADING: JsonType(nullable=True),
```
mkornacker, line 99: It should actually be headings, not the singular (because we're surfacing the hierarchy)

pixeltable/iterators/document.py
```diff
@@ -86,37 +86,41 @@ class DocumentSplitter(ComponentIterator):
```
mkornacker, line 87: Can you also explain in the class comment the use of ftfy.fix_text() (ie, where does it need to get applied/what's the pattern; this will be useful for someone forking this class)?

pixeltable/iterators/document.py
```diff
@@ -86,37 +86,41 @@ class DocumentSplitter(ComponentIterator):
...
+        ChunkMetadata.TITLE: StringType(nullable=True),
+        ChunkMetadata.HEADING: JsonType(nullable=True),
+        ChunkMetadata.SOURCELINE: IntType(nullable=True),
+        ChunkMetadata.PAGE: IntType(nullable=True),
+        ChunkMetadata.BOUNDING_BOX: JsonType(nullable=True),
     }
 
     def __init__(
-            self, document: str, *, separators: str, limit: int = 0, overlap: int = 0, metadata: str = '',
+            self, document: str, *, separators: str, limit: Optional[int] = None, overlap: Optional[int] = None, metadata: str = '',
```
mkornacker, line 106: what's the reason for the Optional[int] for limit and overlap? It doesn't seem to change the usage or the logic.

pixeltable/iterators/document.py
```diff
@@ -170,17 +174,20 @@ def output_schema(cls, *args: Any, **kwargs: Any) -> Tuple[Dict[str, ColumnType]
         if Separator.TOKEN_LIMIT in separators:
             Env.get().require_package('tiktoken')
 
-        if 'limit' in kwargs or 'overlap' in kwargs:
+        limit = kwargs.get('limit', None)
```
mkornacker: don't you have to cast to int?

pixeltable/iterators/document.py
```diff
@@ -170,17 +174,20 @@ def output_schema(cls, *args: Any, **kwargs: Any) -> Tuple[Dict[str, ColumnType]
...
                 raise Error('"limit" must be an integer > 0')
-            if 'overlap' in kwargs and int(kwargs['overlap']) < 0:
+            if overlap is not None and overlap < 0:
                 raise Error('"overlap" must be an integer >= 0')
         if Separator.TOKEN_LIMIT in separators or Separator.CHAR_LIMIT in separators:
-            if 'token_limit' in separators and 'char_limit' in separators:
+            if Separator.TOKEN_LIMIT in separators and Separator.CHAR_LIMIT in separators:
                 raise Error('Cannot specify both "token_limit" and "char_limit" separators')
-            if 'limit' not in kwargs:
+            if 'limit' not in kwargs or kwargs['limit'] is None:
```
mkornacker: why is the None check not enough?

pixeltable/iterators/document.py
```diff
@@ -82,7 +82,8 @@ def _parse_metadata(metadata: str) -> List[ChunkMetadata]:
 class DocumentSplitter(ComponentIterator):
     """Iterator over pieces of a document. The document is split into chunks based on the specified separators.
     The iterator output tuples are of schema {'text': StringType()}, but can include additional metadata fields if specified
-    in the `metadata` argument as explained below.
+    in the `metadata` argument as explained below, the chunk text is passed through `ftfy.fix_text` to fix up common problems
```
mkornacker: below. The chunk text

mkornacker (approved):

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/98)* status: 1 of 15 files reviewed, 32 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/iterators/document.py` line 85 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1581631987:-NwSV2KLB2hWeJG5x9_x:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/c2d8a4d56eba0a5d7af1a0f82771595f6670a76d/pixeltable/iterators/document.py#L85)):*

below. The chunk text
</blockquote></details>

Done.

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/98)* status: 1 of 15 files reviewed, 32 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/type_system.py` line 882 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/98#-NwLz0fW6zfUU8haBath:-NwSeZ7V-2t0EmL1DDK4:b-n0vw2s) ([raw file](https://github.com/pixeltable/pixeltable/blob/a1503fd25b9aadcc9b0647d34da2896df927b3e2/pixeltable/type_system.py#L882)):*

removed
</blockquote></details>

I think you need to mark this as completed.
___
*[`pixeltable/utils/documents.py` line 46 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/98#gh-1562807879:-NwSefab-nP2fn0v7qzu:b-dr222c) ([raw file](https://github.com/pixeltable/pixeltable/blob/9e6ab9d147af3b2ba2086e41cf4518a9d28c4976/pixeltable/utils/documents.py#L46)):*

Done.
</blockquote></details>

I believe this needs to be marked completed

aaron-siegel (approved):

mkornacker (approved):

PR #100, opened 2024-03-22
Improve tqdm display and fix spurious warning

aaron-siegel:
Improves the appearance of the tqdm progress bar in Jupyter notebooks. Removes the spurious tqdm warning when pixeltable initializes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/100)


pyproject.toml
```diff
@@ -22,6 +22,7 @@ pandas = "^1.5.3"
 pillow = "^9.4.0"
 opencv-python-headless = "^4.7.0.68"
 tqdm = "^4.64.1"
+ipywidgets = "^8.1.0"
```
mkornacker: Is this overly restrictive? We might be able to deal with an older version.

mkornacker (approved):

pyproject.toml
aaron-siegel: It might be. I could try an older version. I'm not sure how critical that is though, I think the dependency graph for ipywidgets is somewhat orthogonal to most of our core dependencies. In general we could spend a lot of time bisecting versions and such, but I'm not convinced that's a good use of time unless there's a known problem with a dependency conflict.

aaron-siegel: In this case, the only packages that are affected are:
```
ipywidgets
jupyterlab-widgets
widgetsnbextension
```

All widgets-related, all new. Nothing existing has to be upgraded. So I'd argue we're actually best off using the latest version in this case.

mkornacker (approved):

PR #101, opened 2024-03-22
Shorthand notation for `ArrayType` instances with wildcard dimensions

aaron-siegel:
Introduces a new shorthand notation for `ArrayType` instances where every dimension is a wildcard (`None`).

`ArrayType(3, dtype=FloatType())`

is equivalent to

`ArrayType((None, None, None), dtype=FloatType())`

and so forth.

This makes `ArrayType`s easier to define and read, in the very common case where all the dimensions are wildcards.

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/101)

### DIFF
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -587,9 +587,12 @@ class JsonType(ColumnType):
 
 class ArrayType(ColumnType):
     def __init__(
-            self, shape: Tuple[Union[int, None], ...], dtype: ColumnType, nullable: bool = False):
+            self, shape: Union[int, Tuple[Union[int, None], ...]], dtype: ColumnType, nullable: bool = False):
         super().__init__(self.Type.ARRAY, nullable=nullable)
-        self.shape = shape
+        if isinstance(shape, int):
+            self.shape = tuple(None for i in range(0, shape))
+        else:
+            self.shape = shape
         assert dtype.is_int_type() or dtype.is_float_type() or dtype.is_bool_type() or dtype.is_string_type()
         self.dtype = dtype._type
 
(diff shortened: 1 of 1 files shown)


pixeltable/type_system.py
```diff
@@ -608,7 +611,12 @@ def _as_dict(self) -> Dict:
         return result
 
     def __str__(self) -> str:
-        return f'{self._type.name.lower()}({self.shape}, dtype={self.dtype.name})'
+        if all(self.shape[n] is None for n in range(0, len(self.shape))):
+            # Display shorthand if all dimensions are unspecified
+            shape_repr = len(self.shape)
```
mkornacker, line 616: Will this be intuitive for users? Is anything else following this convention?

mkornacker (commented):
It's not completely clear that this doesn't create a larger problem than it solves (everybody is used to 'a.shape' returning a tuple; now we have to get them accustomed to something else).

Also: no tests?

aaron-siegel:
Note `a.shape` still returns a tuple, so

`ArrayType(3, dtype=FloatType()).shape`

will still return `(None, None, None)`. Only the string representation is different.

We could also scope the change down by allowing the shorthand notation in the initializer, but not changing the string representation.

I will say though that as a user, I cringe whenever I see

`array((None,), dtype=FLOAT)`

and I find

`array(1, dtype=FLOAT)`

to be far more intuitive.

Obviously I'd add tests before committing anything. Consider this a preview; I don't want to invest time in writing tests and documentation unless we decide this is a change worth pursuing.

PR #104, opened 2024-03-26
Parameterized expressions as a subclass of Function

mkornacker:
- new class ExprTemplate, which contains an Expr tree with Parameter exprs and has a Function interface
- extending FunctionCall to handle ExprTemplates

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/104)


aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/104)* status: 0 of 8 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/function_call.py` line 101 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu04T7--74CHo2Qavb6:-Nu04T705CbOvA3-peLX:b-u4r4mm) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/exprs/function_call.py#L101)):*
> ```Python
>             self.fn_expr_idx = len(self.components) - 1
>         else:
>             self.fn_expr_idx = sys.maxsize
> ```

Semantics would be clearer to me to use an Optional[int] rather than a sentinel value
___
*[`pixeltable/exprs/parameter.py` line 12 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu04j3uA14UPk2qWIVG:-Nu04j3uA14UPk2qWIVH:b-chx51) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/exprs/parameter.py#L12)):*
> ```Python
> 
> 
> class Parameter(Expr):
> ```

I'd suggest `VariableExpr` rather than `Parameter` as the class name.

There are a lot of things floating around called `Parameter`. In my view, these are variables; when they get put into an `ExprTemplate`, then the `ExprTemplate` acquires a parameter for each uniquely named variable. That is, expressions have variables, functions have parameters.

And they really are two different things: there's a many-to-one relationship: a variable with the same name can occur several times in an expression, but it gets linked to a single parameter (I assume).
___
*[`pixeltable/exprs/parameter.py` line 25 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu05nQnBwo9-ef8bxWR:-Nu05nQnBwo9-ef8bxWS:b-fl0p69) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/exprs/parameter.py#L25)):*
> ```Python
>         return super()._id_attrs() + [('name', self.name)]
> 
>     def default_column_name(self) -> Optional[str]:
> ```

This might be better:

```
def default_column_name(self) -> NoReturn:
    raise NotImplementedError(...)
```

Likewise for sql_expr and other functions that are expected never to be called.
___
*[`pixeltable/func/expr_template.py` line 10 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu06DuS9iKPXbyLHZPv:-Nu06DuS9iKPXbyLHZPw:b2obicx) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L10)):*
> ```Python
> 
> 
> class ExprTemplate(Function):
> ```

Since this is a function and not an expression, it would be less confusing if it had `Function` in its name rather than `Expr`.

One option is `LambdaFunction` which to me captures what this is all about.
___
*[`pixeltable/func/expr_template.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu06_NK3Yrw8FRyijpb:-Nu06_NK3Yrw8FRyijpc:b-hrnqma) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L13)):*
> ```Python
>     """A parameterized expression from which an executable Expr is created with a function call."""
>     def __init__(
>             self, expr: 'pixeltable.exprs.Expr', py_signature: inspect.Signature, self_path: Optional[str] = None,
> ```

You need to `import pixeltable` or the forward reference will give a type error. (PyCharm should be warning you about this, unless you disabled warnings?)
___
*[`pixeltable/func/expr_template.py` line 19 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu07UTq2j4jS651iNoa:-Nu07UTq2j4jS651iNob:b2g1wov) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L19)):*
> ```Python
>         self.self_name = name
>         self.param_exprs = list(expr.subexprs(expr_class=exprs.Parameter))
>         self.param_exprs_by_name = {p.name: p for p in self.param_exprs}
> ```

What happens if there are two with the same name? The logical treatment would be to evaluate all like-named occurrences under a single parameter, but I don't think that's what's happening here.

Suggestion is to make `param_exprs_by_name` one-to-many using a group-by instead of a comprehension. Something like this:
```
self.param_exprs_by_name = dict(itertools.groupby(self.param_exprs, lambda p: p.name))
```
___
*[`pixeltable/func/expr_template.py` line 32 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu087Od3XN1d8LTwXzC:-Nu087Od3XN1d8LTwXzD:b-66dby4) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L32)):*
> ```Python
>             except TypeError as e:
>                 msg = str(e)
>                 raise excs.Error(f"Default value for parameter '{py_param.name}': {msg[0].lower() + msg[1:]}")
> ```

Include function name in error message for clarity?
___
*[`pixeltable/func/expr_template.py` line 50 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu08esp501jOnvyR61S:-Nu08XA0EAcIVZX3RX0H:b6letbn) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L50)):*
> ```Python
>         result = self.expr.copy()
>         for param_name, arg in bound_args.items():
>             param_expr = self.param_exprs_by_name[param_name]
> ```

If the same variable occurs twice in an expression this will mask later occurrences and the subsequent assertion will fire (see preceding comment)
___
*[`pixeltable/func/expr_template.py` line 81 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu0As0e6TP6gFnYFUbu:-Nu0As0e6TP6gFnYFUbv:b-wvcke3) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L81)):*
> ```Python
> 
> 
> def expr_udf(*, param_types: List[ts.ColumnType]) -> Callable:
> ```

Suggestion: Move this to udf.py after merging #95. We can probably unify some of the logic too.
___
*[`pixeltable/tests/test_function.py` line 256 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu0BGQk-4xkMUurybO7:-Nu0BGQk-4xkMUurybO8:b-3bqhv1) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/tests/test_function.py#L256)):*
> ```Python
>             return x + y
>         res1 = t.select(out=add2(t.c2)).order_by(t.c2).collect()
>         pass
> ```

Spurious `pass`?

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/104)* status: 0 of 8 files reviewed, 10 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exprs/parameter.py` line 12 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu04j3uA14UPk2qWIVG:-Nu0CHxjE26t3Gcxc0S7:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/exprs/parameter.py#L12)):*

I'd suggest `VariableExpr` rather than `Parameter` as the class name.

There are a lot of things floating around called `Parameter`. In my view, these are variables; when they get put into an `ExprTemplate`, then the `ExprTemplate` acquires a parameter for each uniquely named variable. That is, expressions have variables, functions have parameters.

And they really are two different things: there's a many-to-one relationship: a variable with the same name can occur several times in an expression, but it gets linked to a single parameter (I assume).
</blockquote></details>

Done.
___
*[`pixeltable/exprs/parameter.py` line 25 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu05nQnBwo9-ef8bxWR:-Nu0DIlg2d3t-jZ7-N5H:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/exprs/parameter.py#L25)):*

This might be better:

```
def default_column_name(self) -> NoReturn:
    raise NotImplementedError(...)
```

Likewise for sql_expr and other functions that are expected never to be called.
</blockquote></details>

Done.
___
*[`pixeltable/func/expr_template.py` line 10 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu06DuS9iKPXbyLHZPv:-Nu0EFy9-ttFlC0Z5ygt:bc6x19c) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L10)):*

Since this is a function and not an expression, it would be less confusing if it had `Function` in its name rather than `Expr`.

One option is `LambdaFunction` which to me captures what this is all about.
</blockquote></details>

ExprTemplateFunction? It's also really just an expr template.
___
*[`pixeltable/func/expr_template.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu06_NK3Yrw8FRyijpb:-Nu0ED794UkuEG_0HieX:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L13)):*

You need to `import pixeltable` or the forward reference will give a type error. (PyCharm should be warning you about this, unless you disabled warnings?)
</blockquote></details>

Done.
___
*[`pixeltable/func/expr_template.py` line 19 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu07UTq2j4jS651iNoa:-Nu0Emk74tPwrTpnGm7R:bwbl17k) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L19)):*

What happens if there are two with the same name? The logical treatment would be to evaluate all like-named occurrences under a single parameter, but I don't think that's what's happening here.

Suggestion is to make `param_exprs_by_name` one-to-many using a group-by instead of a comprehension. Something like this:
```
self.param_exprs_by_name = dict(itertools.groupby(self.param_exprs, lambda p: p.name))
```
</blockquote></details>

In the current code, there cannot be two with the same name, because the names come from the decorated function's signature. But there should be  a check here.
___
*[`pixeltable/func/expr_template.py` line 32 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu087Od3XN1d8LTwXzC:-Nu0FPbM6a213azQKB45:b8iwiqd) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L32)):*

Include function name in error message for clarity?
</blockquote></details>

Not here, because this is the c'tor, called directly by the decorator (ie, there is only one function this could apply to).
___
*[`pixeltable/func/expr_template.py` line 50 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu08esp501jOnvyR61S:-Nu0FdDKEaujpQRUh6vl:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L50)):*

If the same variable occurs twice in an expression this will mask later occurrences and the subsequent assertion will fire (see preceding comment)
</blockquote></details>

Done.
___
*[`pixeltable/func/expr_template.py` line 81 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu0As0e6TP6gFnYFUbu:-Nu0Gc3oDQYu2nDEVRVq:bb71kmx) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L81)):*

Suggestion: Move this to udf.py after merging #95. We can probably unify some of the logic too.
</blockquote></details>

Okay
___
*[`pixeltable/tests/test_function.py` line 256 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu0BGQk-4xkMUurybO7:-Nu0FmsLDkNm3LPss16o:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/tests/test_function.py#L256)):*

Spurious `pass`?
</blockquote></details>

Done.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/104)* status: 0 of 10 files reviewed, 10 unresolved discussions (waiting on @mkornacker and @udf)

___
*[`pixeltable/func/expr_template.py` line 10 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu06DuS9iKPXbyLHZPv:-Nu1rUTx8VjIOOO5uIZp:b2nhll9) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L10)):*

ExprTemplateFunction? It's also really just an expr template.
</blockquote></details>

Ok.
___
*[`pixeltable/func/expr_template.py` line 19 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu07UTq2j4jS651iNoa:-Nu1rvWVEJx1qvV9uA7n:brmjdk9) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L19)):*

In the current code, there cannot be two with the same name, because the names come from the decorated function's signature. But there should be  a check here.
</blockquote></details>

Maybe I don't fully understand what's going on then. What would be the effect of this?

```
@pxt.expr_udf(param_types=[IntType()])
    def add(x: int) -> int:
        return x + x
```

It mentions `x` twice in the expression.

Oh, I think I see now. It will be two references to the same `Variable` object.

Add a test like this that mentions the same variable twice, just to validate that it's doing the right thing in this case?
___
*[`pixeltable/func/expr_template.py` line 81 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu0As0e6TP6gFnYFUbu:-Nu1sMVF1yh8H2RYLtXJ:b-y1734h) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L81)):*

Okay
</blockquote></details>

This suggestion is obsolete I guess, since we moved @udf back to function.py.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/104)* status: 0 of 10 files reviewed, 10 unresolved discussions (waiting on @mkornacker and @udf)

___
*[`pixeltable/func/expr_template.py` line 81 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/104#-Nu0As0e6TP6gFnYFUbu:-Nu2SNGZAmb5qZsytC1z:b-knxllf) ([raw file](https://github.com/pixeltable/pixeltable/blob/795916b73e55a6c6172d8632f8a79042708d2d7b/pixeltable/func/expr_template.py#L81)):*

This suggestion is obsolete I guess, since we moved @udf back to function.py.
</blockquote></details>

Ignore that comment! We only moved the tests into test_function, not the code itself. I don't see a way to delete a comment in reviewable (!)

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/104)* status: 0 of 17 files reviewed, 5 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/func/batched_function.py` line 40 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/104#-NuQK-MeAThD-uQFRsx8:-NuQK-MeAThD-uQFRsx9:b1ol8it) ([raw file](https://github.com/pixeltable/pixeltable/blob/5ecc9b8a275355828d8eca56b6711ee052514b90/pixeltable/func/batched_function.py#L40)):*
> ```Python
>     def invoke(self, arg_batches: List[List[Any]], kwarg_batches: Dict[str, List[Any]]) -> List[Any]:
>         """Invoke the function for the given batch and return a batch of results"""
>         constant_param_names = [p.name for p in self.signature.constant_parameters]
> ```

Incidentally, I think we should switch to specifying batched_parameters as the "special" parameters, rather than constant_parameters. (Not something that's actionable for this review - just a note. I'm planning to make this change in the future, after NOS goes away)
___
*[`pixeltable/exprs/variable.py` line 27 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/104#-NuQJUf4E26JikB2jDX8:-NuQJUf4E26JikB2jDX9:b-kc9ofx) ([raw file](https://github.com/pixeltable/pixeltable/blob/5ecc9b8a275355828d8eca56b6711ee052514b90/pixeltable/exprs/variable.py#L27)):*
> ```Python
>         raise NotImplementedError()
> 
>     def _equals(self, other: Variable) -> bool:
> ```

Check that col_type is the same as well?
___
*[`pixeltable/func/expr_template_function.py` line 51 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/104#-NuQKQFtDa3WdyEKlq2W:-NuQKQFtDa3WdyEKlq2X:b-ceso5i) ([raw file](https://github.com/pixeltable/pixeltable/blob/5ecc9b8a275355828d8eca56b6711ee052514b90/pixeltable/func/expr_template_function.py#L51)):*
> ```Python
>         # apply defaults, otherwise we might have Parameters left over
>         bound_args.update(
>             {param_name: default for param_name, default in self.defaults.items() if param_name not in bound_args})
> ```

Can this be an iterator? (Get rid of explicit dict comprehension?)

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/104)* status: 0 of 17 files reviewed, 6 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/tests/test_function.py` line 224 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/104#-NuQM_2h7yzMRUEvwBA6:-NuQM_2h7yzMRUEvwBA7:b4ulx0h) ([raw file](https://github.com/pixeltable/pixeltable/blob/5ecc9b8a275355828d8eca56b6711ee052514b90/pixeltable/tests/test_function.py#L224)):*
> ```Python
>         assert 'reserved' in str(exc_info.value)
> 
>     def test_expr_udf(self, test_tbl: catalog.Table) -> None:
> ```

Add a test that uses a variable more than once e.g.

def add(x: int) -> int:
    return x + x

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/104)* status: 0 of 17 files reviewed, 6 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/tests/test_function.py` line 224 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/104#-NuQM_2h7yzMRUEvwBA6:-NuRXDb-EkcOjegXfHKg:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/5ecc9b8a275355828d8eca56b6711ee052514b90/pixeltable/tests/test_function.py#L224)):*

Add a test that uses a variable more than once e.g.

def add(x: int) -> int:
    return x + x
</blockquote></details>

Done.
___
*[`pixeltable/exprs/variable.py` line 27 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/104#-NuQJUf4E26JikB2jDX8:-NuRW7AL49uuxFsfZvGN:b-a8hm47) ([raw file](https://github.com/pixeltable/pixeltable/blob/5ecc9b8a275355828d8eca56b6711ee052514b90/pixeltable/exprs/variable.py#L27)):*

Check that col_type is the same as well?
</blockquote></details>

Not here, that should happen in Expr.equals(), but that's out of scope.
___
*[`pixeltable/func/expr_template_function.py` line 51 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/104#-NuQKQFtDa3WdyEKlq2W:-NuRWekGEI-3aR6Sz7bx:b-d0z3s1) ([raw file](https://github.com/pixeltable/pixeltable/blob/5ecc9b8a275355828d8eca56b6711ee052514b90/pixeltable/func/expr_template_function.py#L51)):*

Can this be an iterator? (Get rid of explicit dict comprehension?)
</blockquote></details>

I don't think that's possible. Also, this wouldn't have any performance benefits.

PR #111, opened 2024-04-01
move depedencies currently needed for import to work

orm011:
A few deps were missing on a clean env.
Some were missing, some are meant to be dev, but are being imported from pxt. Cleaning that can be separate PR. 

mamba create <>
pip install <>
python -c 'import pixeltable as pxt; cl = pxt.Client()'   

Import fails without them (on clean env)

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/111)


aaron-siegel:
Are you sure you ran `poetry lock` with `--no-update`? There are a LOT of changes.

Also, let's create a JIRA ticket to track down why those 3 dependencies (that should only be needed for dev) are breaking the import.

orm011:
solved in another PR

PR #123, opened 2024-04-09
Inference API updates

aaron-siegel:
Lots of updates to the OpenAI, Together, and Fireworks APIs, including:
- Mostly complete implementations of OpenAI and Together (except model training), including image and audio endpoints
- A common pattern for registering clients with the environment and surfacing API keys
- Retries in OpenAI to account for rate limiting (using the `tenacity` library - not a great solution, since it will synchronously block the pixeltable executor; eventually pixeltable should orchestrate its own retries)
- Lots of new and expanded unit tests

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/123)


pixeltable/functions/openai.py
```diff
@@ -1,51 +1,160 @@
...
 
 
+def openai_client() -> openai.OpenAI:
+    return env.Env.get().get_client('openai', lambda api_key: openai.OpenAI(api_key=api_key))
+
+
+# Exponential backoff decorator using tenacity.
+# TODO(aaron-siegel): Right now this hardwires random exponential backoff with defaults suggested
+# by OpenAI. Should we investigate making this more customizable in the future?
+def _retry(fn: Callable) -> Callable:
```
mkornacker, line 26: What's the reason for doing this here? Did you run into some throttling problems?

pixeltable/functions/openai.py
```diff
@@ -61,28 +170,85 @@ def vision(
...
+@_retry
+def image_generations(
+        prompt: str,
+        *,
+        model: Optional[str] = None,
+        quality: Optional[str] = None,
+        size: Optional[str] = None,
+        style: Optional[str] = None,
+        user: Optional[str] = None
+) -> PIL.Image.Image:
```
mkornacker, line 217: Do we know what the resolution is? Would be nice to return that as part of the type.

pixeltable/functions/together.py
```diff
@@ -1,27 +1,121 @@
...
+        prompt: str,
+        *,
+        model: str,
+        steps: Optional[int] = None,
+        seed: Optional[int] = None,
+        height: Optional[int] = None,
+        width: Optional[int] = None,
+        negative_prompt: Optional[str] = None,
+) -> PIL.Image.Image:
+    result = together_client().images.generate(
```
mkornacker, line 109: What is the type of result? It would be good to decompose this into the endpoint call and the post-processing, in the form of an expr udf.

pixeltable/tests/functions/test_fireworks.py
```diff
@@ -0,0 +1,40 @@
...
+        )
+        t['output_2'] = chat_completions(
+            messages=messages,
+            model='accounts/fireworks/models/llama-v2-7b-chat',
+            max_tokens=300,
+            top_k=40,
+            top_p=0.9,
+            temperature=0.7
+        )
+        t.insert(input="How's everything going today?")
```
mkornacker: check status for exceptions

pixeltable/tests/functions/test_fireworks.py
```diff
@@ -0,0 +1,40 @@
...
+        results = t.collect()
+        assert len(results['output'][0]['choices'][0]['message']['content']) > 0
+        assert len(results['output_2'][0]['choices'][0]['message']['content']) > 0
+
+    @staticmethod
+    def skip_test_if_no_fireworks_client() -> None:
```
mkornacker, line 37: What is this for?

pixeltable/type_system.py
```diff
@@ -240,16 +240,22 @@ def infer_literal_type(cls, val: Any) -> Optional[ColumnType]:
...
     def from_python_type(cls, t: type) -> Optional[ColumnType]:
-        if t in _python_type_to_column_type:
+        if typing.get_origin(t) is typing.Union:
+            union_args = typing.get_args(t)
+            if union_args[1] is type(None):
+                # `t` is a type of the form Optional[T] (equivalently, Union[T, None]).
+                # We treat it as the underlying type but with nullable=True.
+                underlying = cls.from_python_type(union_args[0])
+                if underlying is not None:
+                    result = copy(underlying)
```
mkornacker: Why the copy?

pixeltable/type_system.py
```diff
@@ -240,16 +240,22 @@ def infer_literal_type(cls, val: Any) -> Optional[ColumnType]:
...
-        elif isinstance(t, typing._UnionGenericAlias) and t.__args__[1] is type(None):
-            # `t` is a type of the form Optional[T] (equivalently, Union[T, None]).
-            # We treat it as the underlying type but with nullable=True.
-            if t.__args__[0] in _python_type_to_column_type:
-                underlying = copy(_python_type_to_column_type[t.__args__[0]])
-                underlying.nullable = True
-                return underlying
-
+        else:
+            base = typing.get_origin(t)  # Discard type parameters
```
mkornacker: What is this for?

pyproject.toml
```diff
@@ -19,7 +19,7 @@ exclude = [
 python = ">=3.9,<4.0"
 numpy = "^1.26"
 pandas = ">=2.0,<3.0"
-pillow = "^9.4.0"
+pillow = ">=10.1"
```
mkornacker: Is there some specific reason for 10.1, as opposed to 10.0?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/123)* status: 0 of 16 files reviewed, 8 unresolved discussions (waiting on @mkornacker)

___
*[`pyproject.toml` line 22 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561994700:-NvFY8327E5B_XkJXPLq:byg9l32) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pyproject.toml#L22)):*

Is there some specific reason for 10.1, as opposed to 10.0?
</blockquote></details>

No specific reason. I can downgrade this. It's an interesting problem, though. The wider our version windows, the more flexible pxt is wrt other libraries; but it also means our tests are less exhaustive (more likely there will be some breaking combination of library versions we haven't tested).
___
*[`pixeltable/type_system.py` line 250 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561994325:-NvFZ4NrEZqM710jyDIy:b-5qrh26) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/type_system.py#L250)):*

Why the copy?
</blockquote></details>

We have to set `result.nullable = True` for the `Optional` type. If we don't make a copy, it will set `result.nullable = True` on the dictionary instance, which will cause the incorrect type to be returned for anything non-nullable in the future.

I added a code comment to clarify this.
___
*[`pixeltable/type_system.py` line 256 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561994405:-NvFZPAZ99VX-AT-t2NP:b-64uh97) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/type_system.py#L256)):*

What is this for?
</blockquote></details>

To ensure that parameterized types such as `list[int]` get mapped correctly to pixeltable types.

Without this, `list` would get mapped to a `JsonType()` (for example), but not `list[int]`. This ensures that any `list[T]` gets properly mapped. I added a code comment to clarify this.
___
*[`pixeltable/functions/openai.py` line 26 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561986726:-NvFZi-S4xXWT4fFTXHe:b-qvgplk) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/functions/openai.py#L26)):*

What's the reason for doing this here? Did you run into some throttling problems?
</blockquote></details>

Yes, without this, rate limited accounts will run into problems. OpenAI recommends rate limiting as best general practice. See:
https://platform.openai.com/docs/guides/rate-limits?context=tier-free
___
*[`pixeltable/functions/openai.py` line 216 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561988230:-NvF_-Qy19wRf-v9m84l:bo0szs7) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/functions/openai.py#L216)):*

Do we know what the resolution is? Would be nice to return that as part of the type.
</blockquote></details>

Nope. It depends on the parameters (and model choice), unfortunately.
___
*[`pixeltable/functions/together.py` line 108 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561989514:-NvF_9bT-0QOxO3KJxag:b-xedakg) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/functions/together.py#L108)):*

What is the type of result? It would be good to decompose this into the endpoint call and the post-processing, in the form of an expr udf.
</blockquote></details>

Json.

Maybe the decomposition could be a TODO? There are similar considerations with the OpenAI api.
___
*[`pixeltable/tests/functions/test_fireworks.py` line 29 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561990093:-NvF_i3fE2qO2aXYgDxK:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/tests/functions/test_fireworks.py#L29)):*

check status for exceptions
</blockquote></details>

Done.
___
*[`pixeltable/tests/functions/test_fireworks.py` line 35 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561990508:-NvF_qOM6TiWIbQrW8FG:bgd5rva) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/tests/functions/test_fireworks.py#L35)):*

What is this for?
</blockquote></details>

The main reason: When a pull request runs in CI, it won't have access to the API keys (for unavoidable security reasons). This ensures that the test will skip rather than fail if no API key is present. (CI will run again post-merge, this time with access to the API keys, ensuring that the test eventually runs in CI.)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/123)* status: 0 of 16 files reviewed, 11 unresolved discussions (waiting on @aaron-siegel)

___
*[`pyproject.toml` line 22 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561994700:-NvOa8Xf5LZtowDVCqs7:bjrpp4k) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pyproject.toml#L22)):*

No specific reason. I can downgrade this. It's an interesting problem, though. The wider our version windows, the more flexible pxt is wrt other libraries; but it also means our tests are less exhaustive (more likely there will be some breaking combination of library versions we haven't tested).
</blockquote></details>

Let's try to be more inclusive here
___
*[`pixeltable/type_system.py` line 250 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561994325:-NvOaW6s-HavLUsgMlUW:b-23hm8q) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/type_system.py#L250)):*

We have to set `result.nullable = True` for the `Optional` type. If we don't make a copy, it will set `result.nullable = True` on the dictionary instance, which will cause the incorrect type to be returned for anything non-nullable in the future.

I added a code comment to clarify this.
</blockquote></details>

I'm only now picking up on this. Returning shared type instance is generally unsafe.
___
*[`pixeltable/type_system.py` line 255 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/123#-NvOaynA499lVATChRJp:-NvOaynA499lVATChRJq:b-5aww7q) ([raw file](https://github.com/pixeltable/pixeltable/blob/b2ec31dec3589477bd43111d6f5a94cf2d696303/pixeltable/type_system.py#L255)):*
> ```Python
>                     return result
>         elif t in _python_type_to_column_type:
>             return _python_type_to_column_type[t]
> ```

This should return the copy.
___
*[`pixeltable/functions/openai.py` line 26 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561986726:-NvObwQuA7QssFPTuFTp:bazv4km) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/functions/openai.py#L26)):*

Yes, without this, rate limited accounts will run into problems. OpenAI recommends rate limiting as best general practice. See:
https://platform.openai.com/docs/guides/rate-limits?context=tier-free
</blockquote></details>

Good to know.

My guess is that the correct place to do this is at the ExecNode level (ie, where you have a clear picture of all the endpoints that are being used in a single query).
___
*[`pixeltable/functions/openai.py` line 216 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561988230:-NvOcTbG-z09uMG8T6Km:b-ii12e4) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/functions/openai.py#L216)):*

Nope. It depends on the parameters (and model choice), unfortunately.
</blockquote></details>

But this information is available at runtime, no?
___
*[`pixeltable/functions/together.py` line 108 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561989514:-NvOcjBT056JSMfGyurJ:b-3uo7nu) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/functions/together.py#L108)):*

Json.

Maybe the decomposition could be a TODO? There are similar considerations with the OpenAI api.
</blockquote></details>

Leave a todo then (here and in the other places where it applies).
___
*[`pixeltable/tests/functions/test_fireworks.py` line 35 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/123#gh-1561990508:-NvOcwWS8e34zYw0ncd_:b-1qjfzg) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f893985faa43124e9eeb8426eb069bc0bc8c4a4/pixeltable/tests/functions/test_fireworks.py#L35)):*

The main reason: When a pull request runs in CI, it won't have access to the API keys (for unavoidable security reasons). This ensures that the test will skip rather than fail if no API key is present. (CI will run again post-merge, this time with access to the API keys, ensuring that the test eventually runs in CI.)
</blockquote></details>

Best to leave that comment in the code
___
*[`pixeltable/tests/functions/test_openai.py` line 27 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/123#-NvOdsxN1kQMhJlCYyv4:-NvOdsxN1kQMhJlCYyv5:b-4ed23b) ([raw file](https://github.com/pixeltable/pixeltable/blob/b2ec31dec3589477bd43111d6f5a94cf2d696303/pixeltable/tests/functions/test_openai.py#L27)):*
> ```Python
>             t.speech, model='whisper-1', prompt='Translate the recording from Spanish into English.', temperature=0.7
>         ))
>         t.insert(input='I am a banana.')
> ```

call insert() with a list and check status (that way, you test a multi-row transaction)

re checking status: add that to every insert() call
___
*[`pixeltable/tests/functions/test_openai.py` line 69 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/123#-NvOeeeh8zpTz6EvQtJ_:-NvOeeeh8zpTz6EvQtJa:b3m5qbj) ([raw file](https://github.com/pixeltable/pixeltable/blob/b2ec31dec3589477bd43111d6f5a94cf2d696303/pixeltable/tests/functions/test_openai.py#L69)):*
> ```Python
>         assert len(result['chat_output_4'][0]['choices'][0]['message']['content']) > 0
> 
>         # TODO This should probably not be throwing an exception, but rather logging the error in
> ```

This cannot be a todo, it breaks the insert semantics. Also, ExprEvalError is not meant to be user-facing.

Why is there an exception here?

pixeltable/type_system.py
```diff
@@ -240,16 +240,25 @@ def infer_literal_type(cls, val: Any) -> Optional[ColumnType]:
...
-
+        if typing.get_origin(t) is typing.Union:
+            union_args = typing.get_args(t)
+            if union_args[1] is type(None):
+                # `t` is a type of the form Optional[T] (equivalently, Union[T, None]).
+                # We treat it as the underlying type but with nullable=True.
+                underlying = cls.from_python_type(union_args[0])
+                if underlying is not None:
+                    # Make a copy to ensure we're not mutating the global dictionary entry
+                    result = copy(underlying)
```
mkornacker: Why is this still necessary? from_python_type() should never return a copy().

mkornacker (commented):
ColumnType.from_python_type() cannot return shared data. The dict you're using there should not be used directly outside of from_python_type(). It should probably be a class member (and should really be local to from_python_type()).

pixeltable/type_system.py
mkornacker: I meant to say: from_python_type() should never return a shared ColumnType instance

mkornacker (approved):

PR #124, opened 2024-04-10
Initial support for embedding indices

mkornacker:
This includes:
- index creation with Table.add_index()
- index maintenance during insert/update/delete operations

This excludes
- retrieval
- generalization of indexing (with an Index base class and an EmbeddingIndex subclass, in preparation for FullTextIndex, etc.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/124)


aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/124)* status: 0 of 32 files reviewed, 11 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/store.py` line 175 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCnroA7F4oyFnjPgPv:-NvCnroA7F4oyFnjPgPw:b-rsgk09) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/store.py#L175)):*
> ```Python
>             .where(self.v_max_col > self.tbl_version.version)
>         try:
>             needs_close = conn is None
> ```

This seems like more lines of code, and less clarity, than the simpler
```
if conn is None:
    with env.Env.get().engine.begin() as new_conn:
        result = new_conn.execute(stmt).scalar_one()
else:
        result = conn.execute(stmt).scalar_one()
assert isinstance(result, int)
return result
```
with no real benefit.
___
*[`pixeltable/catalog/table.py` line 479 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCorwZ8sXtvHAuNWG4:-NvCorwZ8sXtvHAuNWG5:b-1f4tiw) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table.py#L479)):*
> ```Python
>         self.tbl_version_path.tbl_version.rename_column(old_name, new_name)
> 
>     def add_index(
> ```

I don't love this API. The pseudo-typing (`idx_type` being a `str`) and use of `kwargs` obfuscate what's actually going on.

Much cleaner, in my opinion, is
```
def add_index(self, col_name: str, *, idx_name: Option[str] = None, idx: Index)
```
where the subclass of `Index` defines the subparameter structure.

I'm aware of the desire to keep bespoke classes out of the API but I think this is a case that warrants them, and we can mostly hide them from the user (the same way CallableFunction/BatchedFunction/etc are effectively hidden).

We don't need to resolve this in this PR but I think it's something that merits attention fairly soon.
___
*[`pixeltable/catalog/table.py` line 485 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCqDtp360xlscadUWS:-NvCqDtp360xlscadUWT:b-v54wyd) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table.py#L485)):*
> ```Python
>         Args:
>             col_name: name of column to index
>             idx_name: name of index (if not provided, a name will be generated)
> ```

Can we add more clarity to this in the docstring? What is the structure of the default names that are generated? Are they random or sequential, do they incorporate the column name, etc.

Also, what is the constraint on index names - globally unique, table-unique, or column-unique? It's not clear.
___
*[`pixeltable/catalog/table.py` line 525 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCrBNeAqM9PAhkwseG:-NvCrBNeAqM9PAhkwseH:b-sy74o5) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table.py#L525)):*
> ```Python
>         if (column_name is None) == (idx_name is None):
>             raise excs.Error('Exactly one of column_name or idx_name must be provided')
> ```

Why is this constraint necessary?

Why shouldn't I be able to provide both, as long as they're consistent? (i.e., the index name actually belongs to the specified column?)
___
*[`pixeltable/catalog/table_version.py` line 298 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCry1mCPOv5gqbdzzz:-NvCry1mCPOv5gqbe---:b-oegnd7) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L298)):*
> ```Python
>     def _update_md(
>             self, ts: float, preceding_schema_version: Optional[int], conn: sql.engine.Connection,
>             is_new_version: bool = True
> ```

Add to docstring, maybe with a note that the reason for having this (I assume?) is b/c adding indices doesn't trigger a new version
___
*[`pixeltable/catalog/table_version.py` line 330 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCsY_O6mcaHjOzXlRe:-NvCsFxoDZo9f8u4mvx7:b-sho1j) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L330)):*
> ```Python
>         idx_id = self.next_idx_id
>         self.next_idx_id += 1
>         assert idx_type == 'embedding'
> ```

See comment above; it'd be much cleaner to take EmbeddingIndex as an argument to add_index.
___
*[`pixeltable/catalog/table_version.py` line 373 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCtP4Y861Heqhtyidn:-NvCtP4Y861Heqhtyido:bv8fjdk) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L373)):*
> ```Python
>             idx.create_index(self._store_idx_name(idx_id), val_col, conn)
> 
>         # TODO: create index
> ```

Wait. Isn't that what just happened? There's some dissonance here between

`idx.create_index(...)`

and

`# TODO: create index`

It'd be helpful to have more clarity in the TODO about what is and isn't implemented.
___
*[`pixeltable/catalog/table_version.py` line 383 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCszBC8_cmpGByfFFO:-NvCszBC8_cmpGByfFFP:b6g2oay) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L383)):*
> ```Python
>         del self.index_md[idx_id]
>         del self.idxs_by_name[idx_md.name]
>         # TODO: drop index and index column
> ```

This seems like a pretty major TODO! Are these things we need to do before this feature can be useful?

Also, we should get in the habit of marking TODOs with an author, as in

# TODO(mkornacker): ...

This is fairly standard practice (and in fact linting tools will alert on it)
___
*[`pixeltable/catalog/table_version.py` line 471 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCtysNBJlMfd0LrPl6:-NvCtysNBJlMfd0LrPl7:bly9974) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L471)):*
> ```Python
>                 self.cols.pop()
>                 for col in cols:
>                     # remove columns that we already added
> ```

I haven't fully digested the relevant code/patterns, so interpret this question as me trying to educate myself:

It seems as though the sequence of steps is
- update pxt structures with column changes
- execute the SQL
- if the SQL fails, back out the changes to the pxt structures

It seems simpler/more maintainable to do:
- execute the SQL
- if the SQL succeeds, update pxt structures

What's the reason for doing it the first way (or am I misinterpreting what's happening here)?
___
*[`pixeltable/exec/cache_prefetch_node.py` line 92 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCvCWbDYMDp17t3ooy:-NvCvCWbDYMDp17t3ooz:bm6jo9s) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/exec/cache_prefetch_node.py#L92)):*
> ```Python
>         extension = ''
>         if parsed.path != '':
>             p = Path(urllib.parse.unquote(urllib.request.url2pathname(parsed.path)))
> ```

I caught a whole bunch of these when I did the Windows compatibility fixes.

The fact that more of them are still lurking in the code suggests they're not being tested :(
___
*[`pixeltable/index/embedding_index.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCvs7o9V3rSUzaImr2:-NvCvs7o9V3rSUzaImr3:b472a8k) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/index/embedding_index.py#L13)):*
> ```Python
> 
> 
> class EmbeddingIndex:
> ```

This is very close to the kind of structure that I'd envisioned in my earlier note (great!): an ensemble of Functions into the same vector space.

Main suggestion: factor out everything except `Column` into an `Embedding` class, so that

`EmbeddingIndex` = `Embedding` + `Column`.

Then `Embedding`s can be more general + library-supplied and we get a nice clean API (e.g. for the huggingface stuff).

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/124)* status: 0 of 32 files reviewed, 11 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/store.py` line 175 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCnroA7F4oyFnjPgPv:-NvOK3Ms6wcW-29Y4o6r:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/store.py#L175)):*

This seems like more lines of code, and less clarity, than the simpler
```
if conn is None:
    with env.Env.get().engine.begin() as new_conn:
        result = new_conn.execute(stmt).scalar_one()
else:
        result = conn.execute(stmt).scalar_one()
assert isinstance(result, int)
return result
```
with no real benefit.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 479 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCorwZ8sXtvHAuNWG4:-NvOK9kc5m2cYo8MIVjm:b-qmap8m) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table.py#L479)):*

I don't love this API. The pseudo-typing (`idx_type` being a `str`) and use of `kwargs` obfuscate what's actually going on.

Much cleaner, in my opinion, is
```
def add_index(self, col_name: str, *, idx_name: Option[str] = None, idx: Index)
```
where the subclass of `Index` defines the subparameter structure.

I'm aware of the desire to keep bespoke classes out of the API but I think this is a case that warrants them, and we can mostly hide them from the user (the same way CallableFunction/BatchedFunction/etc are effectively hidden).

We don't need to resolve this in this PR but I think it's something that merits attention fairly soon.
</blockquote></details>

I'm purposely avoiding objects and classes in the user-facing API. Given that there's going to be only a small number of index types, I don't see any harm in using names here. But we should run this by some users and see what their preference is.
___
*[`pixeltable/catalog/table.py` line 485 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCqDtp360xlscadUWS:-NvOLDsz0XRysx6XSfME:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table.py#L485)):*

Can we add more clarity to this in the docstring? What is the structure of the default names that are generated? Are they random or sequential, do they incorporate the column name, etc.

Also, what is the constraint on index names - globally unique, table-unique, or column-unique? It's not clear.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 525 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCrBNeAqM9PAhkwseG:-NvOLLQN0ktifwop5kEc:bm0og2p) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table.py#L525)):*

Why is this constraint necessary?

Why shouldn't I be able to provide both, as long as they're consistent? (i.e., the index name actually belongs to the specified column?)
</blockquote></details>

Why would you want to provide both?
___
*[`pixeltable/catalog/table_version.py` line 298 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCry1mCPOv5gqbdzzz:-NvOLRAo6CSzdQu9HB7F:b-1kisi5) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L298)):*

Add to docstring, maybe with a note that the reason for having this (I assume?) is b/c adding indices doesn't trigger a new version
</blockquote></details>

I'm going to revise this logic...
___
*[`pixeltable/catalog/table_version.py` line 330 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCsY_O6mcaHjOzXlRe:-NvOOnRDBOpbQZHurqfY:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L330)):*

See comment above; it'd be much cleaner to take EmbeddingIndex as an argument to add_index.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table_version.py` line 373 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCtP4Y861Heqhtyidn:-NvOP1xx6jpRBhODw0vV:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L373)):*

Wait. Isn't that what just happened? There's some dissonance here between

`idx.create_index(...)`

and

`# TODO: create index`

It'd be helpful to have more clarity in the TODO about what is and isn't implemented.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table_version.py` line 471 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCtysNBJlMfd0LrPl6:-NvOPh5VCfgV5DvL8uuM:b8ljfdd) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L471)):*

I haven't fully digested the relevant code/patterns, so interpret this question as me trying to educate myself:

It seems as though the sequence of steps is
- update pxt structures with column changes
- execute the SQL
- if the SQL fails, back out the changes to the pxt structures

It seems simpler/more maintainable to do:
- execute the SQL
- if the SQL succeeds, update pxt structures

What's the reason for doing it the first way (or am I misinterpreting what's happening here)?
</blockquote></details>

Good point, I'll clarify in the code.
___
*[`pixeltable/index/embedding_index.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCvs7o9V3rSUzaImr2:-NvOQjCr2qMmwWL1u_nr:b-sydcpw) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/index/embedding_index.py#L13)):*

This is very close to the kind of structure that I'd envisioned in my earlier note (great!): an ensemble of Functions into the same vector space.

Main suggestion: factor out everything except `Column` into an `Embedding` class, so that

`EmbeddingIndex` = `Embedding` + `Column`.

Then `Embedding`s can be more general + library-supplied and we get a nice clean API (e.g. for the huggingface stuff).
</blockquote></details>

The index class also represents the matching operations that can be expressed on indexed columns, and how those are translated to sqlalchemy.

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/124)* status: 0 of 35 files reviewed, 11 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/type_system.py` line 366 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvmubaV634hGaQ7EmlX:-NvmubaV634hGaQ7EmlY:b-x4idga) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/type_system.py#L366)):*
> ```Python
>         Return corresponding SQLAlchemy type.
>         """
>         assert False
> ```

Huh? Why is there stuff happening after the `assert False`?
___
*[`pixeltable/catalog/table.py` line 479 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCorwZ8sXtvHAuNWG4:-Nvmur-59n4ul6FgkTc6:b-bzewsa) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table.py#L479)):*

I'm purposely avoiding objects and classes in the user-facing API. Given that there's going to be only a small number of index types, I don't see any harm in using names here. But we should run this by some users and see what their preference is.
</blockquote></details>

Don't we already have objects and classes in the user-facing API though? I don't see how this is any different than, say, FrameIterator. I get where you're coming from, but it's also important to have clean abstractions and APIs and thoughtful code modularity.

Also, I think it can be done in a way that doesn't directly expose the objects and classes to the user, cf. my earlier proposal. Maybe it's easiest for me to just put up a PR to illustrate what I'm thinking, once this one has been merged.
___
*[`pixeltable/catalog/table.py` line 525 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCrBNeAqM9PAhkwseG:-Nvmv_r8892Rauz5O6mT:b7xz1k7) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table.py#L525)):*

Why would you want to provide both?
</blockquote></details>

I don't have a specific use case in mind, it just feels like an illogical API to refuse to accept both if they're consistent.
___
*[`pixeltable/catalog/table_version.py` line 218 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvmwFZmAjZ3SJ1rnfDP:-NvmwFZmAjZ3SJ1rnfDQ:b-yn1283) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/catalog/table_version.py#L218)):*
> ```Python
>         # create columns first, so the indices can reference them
>         self._init_cols(tbl_md, schema_version_md)
>         self._init_idxs(tbl_md)
> ```

I prefer `_init_indices` etc. Matter of style, of course, but I like to strive for consistent & clear naming
___
*[`pixeltable/catalog/table_version.py` line 255 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvmwcPH6BHuZx6yqBV2:-NvmwcPIAq7oKXYnSBdm:b-b947a) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/catalog/table_version.py#L255)):*
> ```Python
> 
>     def _init_idxs(self, tbl_md: schema.TableMd) -> None:
>         self.idx_md = tbl_md.index_md
> ```

Example of inconsistent naming (I prefer index/indices, but whatever the convention is we should be consistent)
___
*[`pixeltable/catalog/table_version.py` line 257 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-Nvmwr2yC89_coEYFb_d:-Nvmwr2yC89_coEYFb_e:b2v1qbl) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/catalog/table_version.py#L257)):*
> ```Python
>         self.idx_md = tbl_md.index_md
>         self.idxs_by_name = {}
>         import pixeltable.index as index_module
> ```

Why not just keep it as `pixeltable.index`? `index_module` isn't really much shorter. (I feel like I'm nitpicking on naming conventions in this review, sorry)
___
*[`pixeltable/catalog/table_version.py` line 266 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-Nvmxtgb0Xp38gIwehy8:-Nvmxtgb0Xp38gIwehy9:bwfzxjf) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/catalog/table_version.py#L266)):*
> ```Python
>             # instantiate index object
>             cls_name = md.class_fqn.rsplit('.', 1)[-1]
>             cls = getattr(index_module, cls_name)
> ```

Interesting that you chose to hard-code `pixeltable.index` as the module, rather than storing the module as a parameter in the metadata. I had to implement something similar for Remotes, and went with the opposite (store both module name and class name and reconstruct them both dynamically).

I'd be curious to understand why you chose to go this way? We should try to be consistent (and maybe even abstract this out; it seems to be a common pattern)
___
*[`pixeltable/catalog/table_version.py` line 480 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-Nvmyqkv--wUE9agpMzn:-Nvmyqkv--wUE9agpMzo:b-1sb9cz) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/catalog/table_version.py#L480)):*
> ```Python
>         if print_stats:
>             plan.ctx.profile.print(num_rows=row_count)
>         # TODO: what to do about system columns with exceptions?
> ```

Annotate the TODO:

`# TODO(mkornacker): ...`
___
*[`pixeltable/index/base.py` line 11 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-Nvmze6sE7C8haYJ19nD:-Nvmze6t8bXQiyasbDqK:b-njqz17) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/index/base.py#L11)):*
> ```Python
> 
> 
> class IndexBase(abc.ABC):
> ```

`BaseIndex`?
___
*[`pixeltable/index/embedding_index.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCvs7o9V3rSUzaImr2:-NvmzoUL78cIPo8hAIaW:b-lbtpvt) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/index/embedding_index.py#L13)):*

The index class also represents the matching operations that can be expressed on indexed columns, and how those are translated to sqlalchemy.
</blockquote></details>

I understand! I'm specifically not proposing changing the semantics of `EmbeddingIndex`.

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/124)* status: 0 of 35 files reviewed, 11 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/type_system.py` line 366 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvmubaV634hGaQ7EmlX:-Nvsol7D83waFoOrXQnX:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/type_system.py#L366)):*

Huh? Why is there stuff happening after the `assert False`?
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table_version.py` line 383 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/124#-NvCszBC8_cmpGByfFFO:-NvsqALC57VOJO96Mh86:b3hxlbs) ([raw file](https://github.com/pixeltable/pixeltable/blob/b13c9b768a5889bd1d775e785d55e3d389b38c26/pixeltable/catalog/table_version.py#L383)):*

This seems like a pretty major TODO! Are these things we need to do before this feature can be useful?

Also, we should get in the habit of marking TODOs with an author, as in

# TODO(mkornacker): ...

This is fairly standard practice (and in fact linting tools will alert on it)
</blockquote></details>

Resolved the todo
___
*[`pixeltable/catalog/table_version.py` line 257 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-Nvmwr2yC89_coEYFb_d:-NvspQPV02aih1BEA7nm:b-ce9l3q) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/catalog/table_version.py#L257)):*

Why not just keep it as `pixeltable.index`? `index_module` isn't really much shorter. (I feel like I'm nitpicking on naming conventions in this review, sorry)
</blockquote></details>

Because it's being used specifically as a module object in combination with getattr(), which is not how you'd normally use an imported symbol.
___
*[`pixeltable/catalog/table_version.py` line 266 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-Nvmxtgb0Xp38gIwehy8:-Nvspq2j-cKZ5vPKJ1By:b-wm4moz) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/catalog/table_version.py#L266)):*

Interesting that you chose to hard-code `pixeltable.index` as the module, rather than storing the module as a parameter in the metadata. I had to implement something similar for Remotes, and went with the opposite (store both module name and class name and reconstruct them both dynamically).

I'd be curious to understand why you chose to go this way? We should try to be consistent (and maybe even abstract this out; it seems to be a common pattern)
</blockquote></details>

I don't foresee index subclasses living outside of the index module, and having additional flexibility here doesn't buy anything.
___
*[`pixeltable/catalog/table_version.py` line 480 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-Nvmyqkv--wUE9agpMzn:-NvsqJEJ2ZSqbTmvTNuu:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/catalog/table_version.py#L480)):*

Annotate the TODO:

`# TODO(mkornacker): ...`
</blockquote></details>

Done.
___
*[`pixeltable/index/base.py` line 11 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/124#-Nvmze6sE7C8haYJ19nD:-NvsqP49EhMlDBub6cbD:b9oy068) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd0526d8f1105f308cdf4752052f085159a2cf2e/pixeltable/index/base.py#L11)):*

`BaseIndex`?
</blockquote></details>

To me that sounds like something different (an index on a base table).

PR #128, opened 2024-04-17
switch from cv2 to pyav

orm011:
There are two good reasons to switch from cv2 to pyav, 
1.) cv2 seek to frame, as well as cv2 frame numbers are known to be buggy, from personal experience but also see:
https://forum.opencv.org/t/cv2-cap-prop-pos-frames-sometimes-starts-at-a-keyframe-sometimes-doesnt/5380/9
(ie, you can iterate sequentially to frame 20, and this is not going to always be the same as seeking to frame 20), ie you can seek to frame 140 but actually be somewhere else.  This is highly video dependent, ie, in some videos  and formats there is no issue, but in some there are. 

2) somewhat less importantly, less issues with all the extra deps cv2 has (eg SSL deps), vs pyav: libav and codecs.

One complication is that pyav does not provide a way to get a sequential frame number, one can only infer this from iterating directly from the beginning and decoding everything. 

The only seemingly reliable id field for a frame is the presentation timestamp, or `pts`, an integer field. It is also  the id you can use for seeking without decoding, with libav. Hence I made the decision that its better if we abstain from offering users seeking to exact frame number since we cannot honor that, and instead we should persist the PTS in the iterator table, so one can seek to that frame using av.

NB. I further believe we should remove the frame_idx field from the iterator, as it is not possible to reconstruct this field after a 'seek' operation ('set_pos'): the container cannot tell you which sequential frame you are in just from the frame itself, but can tell you which presentation time stamp you are in. (this is not yet removed in this PR)

1. remove cv2 dep by using only pyav (libav bindings)
2. rework video iterator to work with pyav, and provide pts (presentation timestamp) for seeking.
3. suppress libav logging (shows up in .to_image() method for any frame decoded if libav does not offer an optimized version of yuv 2 rgb conversion for that archicture, currently showing up in validation and also in iteration)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/128)

PR #129, opened 2024-04-17
Fix broken retries in unit tests

aaron-siegel:
Increases the retry window and number of retry attempts for OpenAI unit tests. This is necessary because of the strict rate limits on my low-tier personal account. We can continue to iterate on the right rate limit settings if need be - I didn't think them through too much; I just want to get the unit tests working again in CI.

Also contains an unrelated tweak for a slightly flaky unit test.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/129)

mkornacker (approved):

PR #130, opened 2024-04-17
Disable dall-e-3 tests

aaron-siegel:
Disable testing of OpenAI's dall-e-3 model. It's expensive, slow, and prone to triggering rate limits.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/130)

PR #131, opened 2024-04-19
Fix broken udf in eval.py

aaron-siegel:
If `from __future__ import annotations` appears in a .py file, then any udfs in that file will fail to load. This is because for any function `fn` defined in that file, `inspect.signature(fn)` will contain strings rather than type objects.

This PR does not fix the underlying issue, but merely removes `from __future__ import annotations` from eval.py, remediating the broken udf in that file.

It also contains a unit test for `from_python_type` that turned out to be unrelated to the actual issue, but is nonetheless a good thing to have.

### DIFF
--- a/pixeltable/functions/eval.py
+++ b/pixeltable/functions/eval.py
@@ -1,4 +1,3 @@
-from __future__ import annotations
 from typing import List, Tuple, Dict
 from collections import defaultdict
 import sys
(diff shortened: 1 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/131)

mkornacker (approved):

PR #133, opened 2024-04-19
configure pyav logging when initializing pixeltable 

orm011:
sent to separate log as agreed. 
you need to import pixeltable for this to work, if I do that, notebook noise disappears.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/133)

aaron-siegel (approved):

PR #134, opened 2024-04-19
YOLOX integration

aaron-siegel:
Integrates YOLOX directly with Pixeltable, providing a lightweight solution for object detection that doesn't rely on Docker.

Also updates the object-detection-in-pixeltable tutorial to use the new YOLOX integration.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)

orm011:
*[`pixeltable/dataframe.py` line 115 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/134#-NwAiONa9PGQj67IrZdD:-NwAiONa9PGQj67IrZdE:bb1g329) ([raw file](https://github.com/pixeltable/pixeltable/blob/dad90c288ef668258688927b8a8178e82e0eec97/pixeltable/dataframe.py#L115)):*
> ```Python
>             '''
> 
>     def _format_video(self, file_path: str) -> str:
> ```

We can use pyav for this, I'd recommend we stop adding more code using it.

orm011:
*[`pixeltable/dataframe.py` line 115 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/134#-NwAiONa9PGQj67IrZdD:-NwAijyCEa454XEUNNdT:b-7jnddc) ([raw file](https://github.com/pixeltable/pixeltable/blob/dad90c288ef668258688927b8a8178e82e0eec97/pixeltable/dataframe.py#L115)):*

We can use pyav for this, I'd recommend we stop adding more code using it.
</blockquote></details>

using cv2 i mean.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)* status: 0 of 23 files reviewed, 3 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/dataframe.py` line 115 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/134#-NwAiONa9PGQj67IrZdD:-NwAoV500YbsABGGMKGR:b-1cedbg) ([raw file](https://github.com/pixeltable/pixeltable/blob/dad90c288ef668258688927b8a8178e82e0eec97/pixeltable/dataframe.py#L115)):*

using cv2 i mean.
</blockquote></details>

Ok let's make this the last use of cv2? I'd prefer not to block on this b/c the pyav changes aren't in yet. Let's make it a separate problem to refactor to use pyav?
___
*[`pixeltable/ext/functions/yolox.py` line 40 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/134#-NwAknDX948WP-Hcyh0N:-NwAomhX2SWs4mBhTdEn:b-kxq7iz) ([raw file](https://github.com/pixeltable/pixeltable/blob/dad90c288ef668258688927b8a8178e82e0eec97/pixeltable/ext/functions/yolox.py#L40)):*

I'm almost sure a copy is made in the right device unless it is already in the correct device, but the original tensor is not mutated. Is the device part tested?
</blockquote></details>

Yes I tested the device part. There isn't a pytest test for it though; maybe there should be, but it'd take some work to make it portable.

I don't fully understand your comment - is there a specific suggestion here?

orm011 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)* status: 0 of 23 files reviewed, 5 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/functions/huggingface.py` line 60 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/134#-NwAnJCy50ty5mogGiTQ:-NwAnJCy50ty5mogGiTR:bwp8fl1) ([raw file](https://github.com/pixeltable/pixeltable/blob/dad90c288ef668258688927b8a8178e82e0eec97/pixeltable/functions/huggingface.py#L60)):*
> ```Python
> 
> @pxt.udf(batch_size=32, return_type=ts.ArrayType((512,), dtype=ts.FloatType(), nullable=False))
> def clip_text(text: Batch[str], *, model_id: str) -> Batch[np.ndarray]:
> ```

probably, we would also want a device here. It's maybe out of scope to the yolox goals but maybe a ticket needs to be added
___
*[`pixeltable/ext/functions/yolox.py` line 64 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/134#-NwAlqnJ-jw1WvCrueF9:-NwAlqnJ-jw1WvCrueFA:b-q6iok7) ([raw file](https://github.com/pixeltable/pixeltable/blob/dad90c288ef668258688927b8a8178e82e0eec97/pixeltable/ext/functions/yolox.py#L64)):*
> ```Python
> 
> def __images_to_tensors(images: Iterable[PIL.Image.Image], exp: Exp) -> Iterator[torch.Tensor]:
>     for image in images:
> ```

The double underscore is a deliberate choice?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)* status: 0 of 23 files reviewed, 5 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/functions/huggingface.py` line 60 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/134#-NwAnJCy50ty5mogGiTQ:-NwBFDv471TWoWVf_mqo:bte2pmo) ([raw file](https://github.com/pixeltable/pixeltable/blob/dad90c288ef668258688927b8a8178e82e0eec97/pixeltable/functions/huggingface.py#L60)):*

probably, we would also want a device here. It's maybe out of scope to the yolox goals but maybe a ticket needs to be added
</blockquote></details>

Yeah I think we probably do. I'll create a ticket for it
___
*[`pixeltable/ext/functions/yolox.py` line 64 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/134#-NwAlqnJ-jw1WvCrueF9:-NwBF60H0Y3Vi9xFkfwW:b-rwh59s) ([raw file](https://github.com/pixeltable/pixeltable/blob/dad90c288ef668258688927b8a8178e82e0eec97/pixeltable/ext/functions/yolox.py#L64)):*

The double underscore is a deliberate choice?
</blockquote></details>

Yeah, we should talk about this. It's the standard Python convention for private members. We need to have a broader conversation around Python style and consistency.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)* status: 0 of 23 files reviewed, 5 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/ext/functions/yolox.py` line 40 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/134#-NwAknDX948WP-Hcyh0N:-NwCQtRIBsKuFVmI6F8P:b-oqb2bk) ([raw file](https://github.com/pixeltable/pixeltable/blob/dad90c288ef668258688927b8a8178e82e0eec97/pixeltable/ext/functions/yolox.py#L40)):*

Yes I tested the device part. There isn't a pytest test for it though; maybe there should be, but it'd take some work to make it portable.

I don't fully understand your comment - is there a specific suggestion here?
</blockquote></details>

Follow up on this one: this was, indeed, not properly moving the data to gpu. Thanks @orm011. I'm removing the `device` parameter for now.

pixeltable/ext/functions/yolox.py
```diff
@@ -0,0 +1,92 @@
...
+        else:
+            results.append({
+                'bboxes': [(output[:4] / ratio).tolist() for output in outputs[0]],
+                'scores': [output[4].item() * output[5].item() for output in outputs[0]],
+                'labels': [int(output[6]) for output in outputs[0]]
+            })
+    return results
+
+
+def __images_to_tensors(images: Iterable[PIL.Image.Image], exp: Exp) -> Iterator[torch.Tensor]:
```
mkornacker: Just curious: why the '__' instead just '_'?

pixeltable/functions/eval.py
```diff
@@ -156,16 +156,16 @@ def calculate_image_tpfp(
         ts.JsonType(nullable=False)
     ])
 def eval_detections(
-        pred_bboxes: List[List[int]], pred_classes: List[int], pred_scores: List[float],
-        gt_bboxes: List[List[int]], gt_classes: List[int]
+        pred_bboxes: List[List[int]], pred_labels: List[int], pred_scores: List[float],
```
mkornacker, line 159: What's the point of this rename?

pixeltable/functions/huggingface.py
```diff
@@ -58,43 +59,64 @@ def cross_encoder_list(sentence1: str, sentences2: list, *, model_id: str) -> li
 @pxt.udf(batch_size=32, return_type=ts.ArrayType((512,), dtype=ts.FloatType(), nullable=False))
 def clip_text(text: Batch[str], *, model_id: str) -> Batch[np.ndarray]:
     env.Env.get().require_package('transformers')
+    import torch
     from transformers import CLIPModel, CLIPProcessor
 
     model = _lookup_model(model_id, CLIPModel.from_pretrained)
     assert model.config.projection_dim == 512
+    model.eval()
```
mkornacker: I think this should go into _lookup_model()

pixeltable/functions/huggingface.py
```diff
@@ -58,43 +59,64 @@ def cross_encoder_list(sentence1: str, sentences2: list, *, model_id: str) -> li
...
+        threshold: float = 0.5,
+        device: str = 'auto'
+) -> Batch[dict]:
     env.Env.get().require_package('transformers')
+    device = resolve_torch_device(device)
+    import torch
     from transformers import DetrImageProcessor, DetrForObjectDetection
 
     model = _lookup_model(model_id, lambda x: DetrForObjectDetection.from_pretrained(x, revision='no_timm'))
+    model = model.to(device)
```
mkornacker: How does that work? Don't you also have to move the input tensors to the device?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)* status: 0 of 23 files reviewed, 9 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/functions/eval.py` line 159 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577084164:-NwClEyA65qalCZ4ShDK:bscxopc) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/functions/eval.py#L159)):*

What's the point of this rename?
</blockquote></details>

They're called "labels" pretty much everywhere else (huggingface, yolo etc) so I was trying to be consistent.
___
*[`pixeltable/functions/huggingface.py` line 67 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577084771:-NwCllMg8XlTRggZoQWS:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/functions/huggingface.py#L67)):*

I think this should go into _lookup_model()
</blockquote></details>

Done.
___
*[`pixeltable/functions/huggingface.py` line 109 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577085729:-NwClt1i-NkjKgJSF6Wq:bre49d2) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/functions/huggingface.py#L109)):*

How does that work? Don't you also have to move the input tensors to the device?
</blockquote></details>

They get moved below: in the statement
```
outputs = model(**inputs.to(device))
```
___
*[`pixeltable/ext/functions/yolox.py` line 59 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577083239:-NwCkPt-134s00X0q3vo:b86b9jk) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/ext/functions/yolox.py#L59)):*

Just curious: why the '__' instead just '_'?
</blockquote></details>

Double underscore is the standard convention for private members in python. We've been a bit inconsistent about this in the code base (sometimes using single underscore, sometimes nothing even though it's not intended as a public API).

Single underscore is intended for protected methods (can be accessed by a subclass). We should probably have a broader discussion to align on this (and other questions around codebase hygiene) and then maybe have a codebase cleanup blitz at some point.

I'll make these single underscores for now just to preserve some semblance of consistency. I've been inconsistent about it too...

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)* status: 0 of 23 files reviewed, 9 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/functions/huggingface.py` line 67 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577084771:-NwD48qh5fyIA2xDusyn:bszcwdi) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/functions/huggingface.py#L67)):*

Done.
</blockquote></details>

This didn't work - not all models support model.eval(). I backed it out.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)* status: 0 of 23 files reviewed, 9 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/functions/huggingface.py` line 67 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577084771:-NwDBUVSCAmlaKcPi2Nz:b-jv1czg) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/functions/huggingface.py#L67)):*

This didn't work - not all models support model.eval(). I backed it out.
</blockquote></details>

Which ones don't, sentence_transformer and cross_encoder? I think that's an argument for having separate model caches, not moving this back here.
___
*[`pixeltable/functions/huggingface.py` line 109 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577085729:-NwDBlUE-XqEo5gPyME_:b7ymozj) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/functions/huggingface.py#L109)):*

They get moved below: in the statement
```
outputs = model(**inputs.to(device))
```
</blockquote></details>

Is this required prior to every inference call?

Also, what's the motivation for enabling gpu execution for this model family specifically (as opposed to all the other ones)?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)* status: 0 of 23 files reviewed, 9 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/functions/huggingface.py` line 67 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577084771:-NwGYYAj6RT9Fa96E6Ih:b-7zoerg) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/functions/huggingface.py#L67)):*

Which ones don't, sentence_transformer and cross_encoder? I think that's an argument for having separate model caches, not moving this back here.
</blockquote></details>

I made it into a parameter of _lookup_model.
___
*[`pixeltable/functions/huggingface.py` line 109 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577085729:-NwGYjeHEqQzbM0heth-:b-7dh33x) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/functions/huggingface.py#L109)):*

Is this required prior to every inference call?

Also, what's the motivation for enabling gpu execution for this model family specifically (as opposed to all the other ones)?
</blockquote></details>

I moved this into _lookup_model as well. I believe it's basically a no-op if you call it repeatedly, *unless* someone bounces back and forth between devices, in which case it will have huge negative performance impact. In any case, moving it to _lookup_model both cleans up the code and solves the hypothetical performance problem.

mkornacker (approved):

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134)* status: 0 of 23 files reviewed, 4 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/functions/huggingface.py` line 67 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/134#gh-1577084771:-NwKwMKM0nNjA9pHnHdB:brzql9m) ([raw file](https://github.com/pixeltable/pixeltable/blob/37c2d897375a778e7f0f4f63bf7d503c97150a2f/pixeltable/functions/huggingface.py#L67)):*

I made it into a parameter of _lookup_model.
</blockquote></details>

.eval is a method of the nn.Module type.  https://pytorch.org/docs/stable/generated/torch.nn.Module.html, i'd say, if it is an nn.Module, we should call .eval(), as it disables some expensive operations. 

If something is not of type nn.Module, because they probably still hold an nn.module inside, I think its worth looking at what's offered. Or it may be possible they are already in eval().  Certain layers will act differently on .eval and train(), we should double check that for the ones that don't provide it, its because its taken care of somehow, or well have unusual bugs where results are not reproducible. (eg Normalization layers keep moving averages of certain quantities in .train(), these numbers would be updated for every input if eval() is not set.

orm011 (approved):
:lgtm:

> Reviewed 4 of 13 files at r1, 10 of 12 files at r2, 1 of 7 files at r3, 1 of 2 files at r4, 1 of 2 files at r5, 4 of 6 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/134#-:-NwLC2drDm03Rf9TkUJ9:bnfp4nl)* status: 22 of 23 files reviewed, 4 unresolved discussions (waiting on @mkornacker)

PR #139, opened 2024-04-24
Label studio integration

aaron-siegel:
Initial implementation of Label Studio integration.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)

pixeltable/catalog/table.py
```diff
@@ -692,3 +693,134 @@ def revert(self) -> None:
             raise excs.Error('Cannot revert a snapshot')
         self._check_is_dropped()
         self.tbl_version_path.tbl_version.revert()
+
+    def link_remote(
```
mkornacker: As discussed, let's avoid new classes in the API. We can follow the same pattern here as for embedding indices (eg, add_embedding_index(<embedding index-specific parameters); in this case, something like link_to_labelstudio(...)

pixeltable/catalog/table.py
```diff
@@ -692,3 +693,134 @@ def revert(self) -> None:
...
+            self,
+            remote: 'pixeltable.datatransfer.Remote',
+            col_mapping: Optional[dict[str, str]] = None
+    ) -> None:
+        """
+        Links the specified `Remote` to this table. Once a remote is linked, it can be synchronized with
+        this `Table` by calling [`Table.sync_remotes()`]. A record of the link
```
mkornacker: The fact that the link is recorded in the metadata doesn't need to be stated here, that's implied (all metadata is persisted).

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)* status: 0 of 39 files reviewed, 2 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/catalog/table.py` line 704 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/139#gh-1608940999:-NyS-te01EPjRktRShOi:b6al1ov) ([raw file](https://github.com/pixeltable/pixeltable/blob/8ef271c813e125c95a2b40063b1ca37eb5b63c48/pixeltable/catalog/table.py#L704)):*

The fact that the link is recorded in the metadata doesn't need to be stated here, that's implied (all metadata is persisted).
</blockquote></details>

I'd argue it does need to be stated. This wasn't obvious to Mohsin. Of course it's obvious to you and me, but for customers it bears repeating.

Mohsin's first question was "when I restart Pixeltable, how do I re-link the label studio project to my table?" It was news to him that he didn't have to.

mkornacker (commented):
Let's discuss tomorrow how to get it over the finish line.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139#-:-NzfW2Z31F4bovZxexDq:blyz5p1)* status: 0 of 42 files reviewed, 10 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/catalog/table.py` line 721 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeqAWh15IFC0WRJLte:-NzeqAWh15IFC0WRJLtf:bpq8asd) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/catalog/table.py#L721)):*
> ```Python
>         print(f'Linked remote {remote} to table `{self.get_name()}`.')
> 
>     def unlink_remote(
> ```

Let's go with add and drop, since we already have that for columns, and that's generally the terminology used by dbms's.
___
*[`pixeltable/catalog/table.py` line 757 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzerPhC02NsRkSwnfbV:-NzerPhC02NsRkSwnfbW:bvlvn99) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/catalog/table.py#L757)):*
> ```Python
>                 raise excs.Error(
>                     f'Column `{t_col}` does not exist in Table `{self.get_name()}`. '
>                     '(Specify `col_mapping` explicitly?)'
> ```

Confusing, because you passed in the col_mapping. 
___
*[`pixeltable/datatransfer/label_studio.py` line 211 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzexBSs30ggXFAxwLWr:-NzexBSs30ggXFAxwLWs:b-bhul8v) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/datatransfer/label_studio.py#L211)):*
> ```Python
>         localpath_col_opt = [t[media_col_name].localpath] if is_stored else []
>         # Select the media column, rectanglelabels columns, and localpath (if appropriate)
>         rows = t.select(t[media_col_name], *[t[col] for col in t_rl_cols], *localpath_col_opt)
> ```

This is not a good idea, this should be limited to only retrieving new rows. Running this query isn't guaranteed to be free, you might end up downloading a large number of media files
___
*[`pixeltable/datatransfer/remote.py` line 10 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeNpVr6AEchqEawtWO:-NzeNpVr6AEchqEawtWP:b-rybxpg) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/datatransfer/remote.py#L10)):*
> ```Python
> 
> 
> class Remote(abc.ABC):
> ```

Not a good name. It's already used by git to mean something specific and different. And it's also not part of standard dbms terminology.
___
*[`pixeltable/datatransfer/remote.py` line 13 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeMpomA2hjTXUuvUWo:-NzeMponECtG6SkGKVvv:b6w5qao) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/datatransfer/remote.py#L13)):*
> ```Python
>     """
>     Abstract base class that represents a remote data store. Subclasses of `Remote` provide
>     functionality for synchronizing between Pixeltable tables and stateful remote stores.
> ```

It's unclear that this can be used for all external data stores. I generally stay away from introducing abstractions when the actual requirements aren't known, and this is definitely the case here. This will work for labeling projects, and that should be the scope of this class. We can always expand the abstraction or create more general ones later, when we're integrating other data stores such as dbms's, etc., and understand the requirements
___
*[`pixeltable/exprs/data_row.py` line 205 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeO9fJAyfP-CB4Gmvc:-NzeO9fJAyfP-CB4Gmvd:b-98zluu) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/exprs/data_row.py#L205)):*
> ```Python
> 
>     @property
>     def vmin(self) -> int:
> ```

Let's go with v_min to stay with current conventions
___
*[`pixeltable/functions/huggingface.py` line 149 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeOSpW3l-CRqF-0AyU:-NzeOSpW3l-CRqF-0AyV:bmface2) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/functions/huggingface.py#L149)):*
> ```Python
> 
> @pxt.udf
> def detr_to_coco(image: PIL.Image.Image, detr_info: dict[str, Any]) -> dict[str, Any]:
> ```

This needs to live somewhere else, it's not HF-specific.
___
*[`pixeltable/functions/openai.py` line 19 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzfVHea6SRI4o9zyl6J:-NzfVHea6SRI4o9zyl6K:b-i99t59) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/functions/openai.py#L19)):*
> ```Python
> 
> 
> @env.Env.get().register_client('openai')
> ```

What is the purpose of this? I'm finding this harder to follow than what it replaces, and it doesn't appear to save any code.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)* status: 0 of 42 files reviewed, 10 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/catalog/table.py` line 721 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeqAWh15IFC0WRJLte:-Nzg1D4N30xGKrcFVELI:b-hs5vmp) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/catalog/table.py#L721)):*

Let's go with add and drop, since we already have that for columns, and that's generally the terminology used by dbms's.
</blockquote></details>

Oh, ok. I changed it to `link` and `unlink` to avoid mentioning remotes, but `add_X` / `drop_X` is fine too, we just have to figure out what X is.
___
*[`pixeltable/datatransfer/label_studio.py` line 211 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzexBSs30ggXFAxwLWr:-Nzg1OTP12IgfxgNB95V:b9orwnl) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/datatransfer/label_studio.py#L211)):*

This is not a good idea, this should be limited to only retrieving new rows. Running this query isn't guaranteed to be free, you might end up downloading a large number of media files
</blockquote></details>

Ok. Let's talk about how to address it
___
*[`pixeltable/datatransfer/remote.py` line 10 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeNpVr6AEchqEawtWO:-Nzg1Qw1-dbvpXHd9CQo:b-fc0k47) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/datatransfer/remote.py#L10)):*

Not a good name. It's already used by git to mean something specific and different. And it's also not part of standard dbms terminology.
</blockquote></details>

I can't change the name on this PR for schema reasons, but I can hide it from the user and then change the name in the next PR.
___
*[`pixeltable/datatransfer/remote.py` line 13 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeMpomA2hjTXUuvUWo:-Nzg3tQL9ToOr1JAFYB4:b-7acv79) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/datatransfer/remote.py#L13)):*

It's unclear that this can be used for all external data stores. I generally stay away from introducing abstractions when the actual requirements aren't known, and this is definitely the case here. This will work for labeling projects, and that should be the scope of this class. We can always expand the abstraction or create more general ones later, when we're integrating other data stores such as dbms's, etc., and understand the requirements
</blockquote></details>

Ok. Happy to discuss what might be a better name and/or change the docs.
___
*[`pixeltable/exprs/data_row.py` line 205 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeO9fJAyfP-CB4Gmvc:-Nzg1_735ynX1RKZR2WB:b-acmps2) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/exprs/data_row.py#L205)):*

Let's go with v_min to stay with current conventions
</blockquote></details>

If it's `v_min`, shouldn't it also be `row_id`?
___
*[`pixeltable/functions/huggingface.py` line 149 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeOSpW3l-CRqF-0AyU:-Nzg1hfiEc1X2SV3MRpz:b-jnryvz) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/functions/huggingface.py#L149)):*

This needs to live somewhere else, it's not HF-specific.
</blockquote></details>

Actually, I think it is! It's specific to the DETR huggingface models, as far as I know. If you're aware of some standard that it aligns with, let me know and I can move it (but then the name shouldn't be `detr_to_coco`, it should be `X_to_coco` where X is whatever standard is being implemented)
___
*[`pixeltable/functions/openai.py` line 19 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzfVHea6SRI4o9zyl6J:-Nzg20zY6Yz7fhoa3BTL:b-lcpzzj) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/functions/openai.py#L19)):*

What is the purpose of this? I'm finding this harder to follow than what it replaces, and it doesn't appear to save any code.
</blockquote></details>

That's true in the case of OpenAI, but the old pattern only worked for SDK clients with exactly one parameter and with the parameter name "api_key". That won't work for Label Studio, which has two parameters (url and api_key). And I think we should standardize on a uniform pattern for SDK client registration. This is a nice clean pattern that works in the general case.

I did find a way to clean up the decorator syntax a bit, let me know what you think.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)* status: 0 of 42 files reviewed, 19 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/env.py` line 66 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznJVWM21T7lvimuESP:-NznJVWM21T7lvimuESQ:bsepgj4) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L66)):*
> ```Python
>         self._http_address: Optional[str] = None
> 
>         self._registered_clients: dict[str, Any] = {}
> ```

incorrect type decl
___
*[`pixeltable/env.py` line 365 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznJZ4G-WIbkBxKto7s:-NznJZ4H9l9G9jEyyVNP:b49lp13) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L365)):*
> ```Python
>             - name: The name of the client
>         """
>         cl: ApiClient = self._registered_clients[name]
> ```

we don't use type decls here
___
*[`pixeltable/env.py` line 372 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznIse8DX-ZPadVj6fN:-NznIse92nClBmRQ5HLz:bcw98vl) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L372)):*
> ```Python
>         # if not, look in Pixeltable config from `config.yaml`.
> 
>         init_kwargs = {}
> ```

missing type decl
___
*[`pixeltable/env.py` line 381 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznKG2mAI2JCbvzYX0U:-NznKG2mAI2JCbvzYX0V:bfb746m) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L381)):*
> ```Python
>             if param not in init_kwargs or init_kwargs[param] == '':
>                 raise excs.Error(
>                     f'`{name}` client not initialized: parameter `{param}` is not configured.\n'
> ```

I'm finding this harder to read with all the backticks. Do we need those?
___
*[`pixeltable/env.py` line 542 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznJdZS6vHjaFmYv6IA:-NznJdZS6vHjaFmYv6IB:b-fxzv0m) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L542)):*
> ```Python
> 
> @dataclass
> class ApiClient:
> ```

move to top
___
*[`pixeltable/catalog/table.py` line 726 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznKm8YD5GiT1g867SR:-NznKm8YD5GiT1g867SS:b-oi7qgd) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L726)):*
> ```Python
>             self,
>             remote: 'pixeltable.datatransfer.Remote',
>             col_mapping: Optional[dict[str, str]] = None
> ```

As discussed, move the data store-specific parameters and logic into the data store-specific class and call this _link(). It should only have the remote parameter.
___
*[`pixeltable/catalog/table.py` line 738 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznLSBD77YDJWWctbRe:-NznLSBD77YDJWWctbRf:b-6k916u) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L738)):*
> ```Python
>         """
>         push_cols = remote.get_push_columns()
> ```

Given that we already have import and export elsewhere, why not reuse those concepts here and call them export_/import_cols.
___
*[`pixeltable/catalog/table.py` line 751 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznM4Sd2ZCvv-6dMK6k:-NznM4Sd2ZCvv-6dMK6l:b-wjszxe) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L751)):*
> ```Python
>             self,
>             *,
>             remotes: Optional['pixeltable.datatransfer.Remote' | list['pixeltable.datatransfer.Remote']] = None,
> ```

Remove this parameter and assert that there's only one remote. In the next round we can add the optional name/names parameter
___
*[`pixeltable/catalog/table.py` line 760 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznMb-4AAI1nv1F0TJY:-NznMb-4AAI1nv1F0TJZ:b-4xck7b) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L760)):*
> ```Python
>             remotes: If specified, will unlink only the specified `Remote` or list of `Remote`s. If not specified,
>                 will unlink all of this table's `Remote`s.
>             ignore_errors (bool): If `True`, no exception will be thrown if the specified `Remote` is not linked
> ```

What's the point of that?
___
*[`pixeltable/catalog/table.py` line 780 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznN21p-X0asVU9z-k8:-NznN21p-X0asVU9z-k9:b-tmzxhf) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L780)):*
> ```Python
>             print(f'Unlinked remote {remote} from table `{self.get_name()}`.')
> 
>     def _validate_remote(
> ```

Move all data store-specific logic to data store class.
___
*[`pixeltable/catalog/table.py` line 838 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznOTIR-5rYADXf1GJf:-NznOTIR-5rYADXf1GJg:bwk0rpt) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L838)):*
> ```Python
>             self,
>             *,
>             remotes: Optional['pixeltable.datatransfer.Remote' | list['pixeltable.datatransfer.Remote']] = None,
> ```

Remove parameter and assert there's only a single remote. This will be replaced with names.
___
*[`pixeltable/catalog/table.py` line 839 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznNlB_D-dKz_URipkP:-NznNlBaEKYtpehNEWmU:b-6ot091) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L839)):*
> ```Python
>             *,
>             remotes: Optional['pixeltable.datatransfer.Remote' | list['pixeltable.datatransfer.Remote']] = None,
>             push: bool = True,
> ```

export/import
___
*[`pixeltable/catalog/table.py` line 849 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznODj-5ih8-kfyO2g_:-NznODj-5ih8-kfyO2ga:b-ho51ed) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L849)):*
> ```Python
>                 will sync all linked `Remote`s.
>             push: If `True`, data from this table will be pushed to the remotes during synchronization.
>             pull: If `True`, data from this table will be pulled from the remotes during synchronization.
> ```

Ambiguous phrasing. "data from the remote will be imported into this table ..."
___
*[`pixeltable/functions/huggingface.py` line 149 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeOSpW3l-CRqF-0AyU:-NznPUtUAtRynMSzDLKk:b7ebyev) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/functions/huggingface.py#L149)):*

Actually, I think it is! It's specific to the DETR huggingface models, as far as I know. If you're aware of some standard that it aligns with, let me know and I can move it (but then the name shouldn't be `detr_to_coco`, it should be `X_to_coco` where X is whatever standard is being implemented)
</blockquote></details>

Are you saying that running a non-HF DETR model gives you different output?

https://mmdetection.readthedocs.io/en/v2.16.0/_modules/mmdet/models/detectors/detr.html

aaron-siegel (commented):
> Reviewed 1 of 8 files at r7.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)* status: 0 of 42 files reviewed, 19 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/env.py` line 66 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznJVWM21T7lvimuESP:-Nznbh2v361gTOB5MDlW:bnj1s4s) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L66)):*

incorrect type decl
</blockquote></details>

Fixed, thanks
___
*[`pixeltable/env.py` line 365 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznJZ4G-WIbkBxKto7s:-NznbmW27MFUvyECr1Fi:b-7ql1td) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L365)):*

we don't use type decls here
</blockquote></details>

fixed
___
*[`pixeltable/env.py` line 381 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznKG2mAI2JCbvzYX0U:-NzncAacDKKnGDn01pSi:b7bj9hl) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L381)):*

I'm finding this harder to read with all the backticks. Do we need those?
</blockquote></details>

I've always used them to denote identifiers in user-facing messages. Most systems quote identifiers in error messages in some way. e.g. Python uses single quotes:
```
TypeError: 'int' object is not callable
```
I don't have a strong preference on backticks in particular but I do think we should quote identifiers in some way, when they appear inline in an error message like this.
___
*[`pixeltable/env.py` line 542 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznJdZS6vHjaFmYv6IA:-Nzncq420s12OgsE27vR:b-dze1by) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L542)):*

move to top
</blockquote></details>

Why? Is this following some particular convention? I think it's standard for the principal class (in this case `Env`) to be declared before its auxiliary classes, although to my knowledge PEP8 is agnostic about this.
___
*[`pixeltable/catalog/table.py` line 757 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzerPhC02NsRkSwnfbV:-Nzndc264PPyOnf-K0tn:b-886gsx) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/catalog/table.py#L757)):*

Confusing, because you passed in the col_mapping.
</blockquote></details>

Fixed
___
*[`pixeltable/catalog/table.py` line 726 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznKm8YD5GiT1g867SR:-NznpPIx5gZETmjZEypk:b-f8h98u) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L726)):*

As discussed, move the data store-specific parameters and logic into the data store-specific class and call this _link(). It should only have the remote parameter.
</blockquote></details>

I'll do this in the next PR since it doesn't affect the API.
___
*[`pixeltable/catalog/table.py` line 738 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznLSBD77YDJWWctbRe:-NznpNJzCCefMvi34u77:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L738)):*

Given that we already have import and export elsewhere, why not reuse those concepts here and call them export_/import_cols.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 751 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznM4Sd2ZCvv-6dMK6k:-NznpMaF4vSSQ_kuQ5w7:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L751)):*

Remove this parameter and assert that there's only one remote. In the next round we can add the optional name/names parameter
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 760 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznMb-4AAI1nv1F0TJY:-NzndT-A7hEwmW_vGODf:b-gqhxb2) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L760)):*

What's the point of that?
</blockquote></details>

Just to be consistent with drop_table etc. If we're going to be using names, I definitely think we should have it as a parameter.
___
*[`pixeltable/catalog/table.py` line 780 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznN21p-X0asVU9z-k8:-NznpDa-AkJZeE1d2PxN:b-lwnwe0) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L780)):*

Move all data store-specific logic to data store class.
</blockquote></details>

I'll do this in the next PR (it doesn't impact the public-facing API since this is a protected method)
___
*[`pixeltable/catalog/table.py` line 838 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznOTIR-5rYADXf1GJf:-Nznp9so8PO-D0uD6ppi:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L838)):*

Remove parameter and assert there's only a single remote. This will be replaced with names.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 839 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznNlB_D-dKz_URipkP:-NznpAZ14ltaE64fSvK3:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L839)):*

export/import
</blockquote></details>

Done.
___
*[`pixeltable/functions/huggingface.py` line 149 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzeOSpW3l-CRqF-0AyU:-NzndnJi7vmlgkWuPfb4:b-kx6era) ([raw file](https://github.com/pixeltable/pixeltable/blob/9506500de45cc93b5a6ccc840b5e44333580a851/pixeltable/functions/huggingface.py#L149)):*

Are you saying that running a non-HF DETR model gives you different output?

https://mmdetection.readthedocs.io/en/v2.16.0/_modules/mmdet/models/detectors/detr.html
</blockquote></details>

I haven't looked at mmdetect specifically, but yeah, the DETR model doesn't create the JSON struct. As I understand it, that's done by the huggingface library. So unless they're both conforming to some standard, there's no reason to expect the output format to be the same.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)* status: 0 of 42 files reviewed, 9 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/env.py` line 542 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznJdZS6vHjaFmYv6IA:-NzoX7ef0vJn8q7Z3ga-:bvtnv4w) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/env.py#L542)):*

Why? Is this following some particular convention? I think it's standard for the principal class (in this case `Env`) to be declared before its auxiliary classes, although to my knowledge PEP8 is agnostic about this.
</blockquote></details>

I guess I personally find it easier to see this before it's being used and would always put it at the top.
___
*[`pixeltable/catalog/table.py` line 726 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznKm8YD5GiT1g867SR:-NzoXW0P7mod-p7SIrQO:b3uga7z) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L726)):*

I'll do this in the next PR since it doesn't affect the API.
</blockquote></details>

It would affect the API, because the col_mapping parameter will go away.

This doesn't look like much work to me.
___
*[`pixeltable/catalog/table.py` line 849 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznODj-5ih8-kfyO2g_:-NzoXwzc1Da8EhxQcbLO:b5zolwy) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L849)):*

Ambiguous phrasing. "data from the remote will be imported into this table ..."
</blockquote></details>

Isn't that exactly not the case? import == True means data will be imported *into* this table.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)* status: 0 of 42 files reviewed, 16 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`tests/datatransfer/test_label_studio.py` line 54 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzoa3H3CmbPEgnH_HjD:-Nzoa3H3CmbPEgnH_HjE:b-g8d6ql) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L54)):*
> ```Python
>         skip_test_if_not_installed('label_studio_sdk')
>         from pixeltable.datatransfer.label_studio import LabelStudioProject
>         remote = LabelStudioProject.create(title='test_remote_project', label_config=self.test_config_2)
> ```

Given that we decided to move to the create_ls_project() function, this interface isn't user-facing anymore. 
___
*[`tests/datatransfer/test_label_studio.py` line 64 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzob4I4Cr9I8Hw-PmB3:-Nzob4I4Cr9I8Hw-PmB4:b4nwyda) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L64)):*
> ```Python
> 
>         with pytest.raises(excs.Error) as exc_info:
>             _ = LabelStudioProject.create(
> ```

This should be tested via the user-facing function create_label_studio_project()
___
*[`tests/datatransfer/test_label_studio.py` line 96 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzoaUPr9WKtbtI-Oqeo:-NzoaUPr9WKtbtI-Oqep:b-5c6zjj) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L96)):*
> ```Python
>         skip_test_if_not_installed('label_studio_sdk')
>         t = ls_image_table
>         from pixeltable.datatransfer.label_studio import LabelStudioProject
> ```

This shouldn't be needed.
___
*[`tests/datatransfer/test_label_studio.py` line 125 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzobn779XD1kxzVoUU7:-Nzobn779XD1kxzVoUU8:bjv3aq) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L125)):*
> ```Python
>         t = pxt.get_table('test_ls_sync')
>         t.sync()
>         annotations = t.collect()['annotations_col']
> ```

You're expecting a particular order, which means you need an order_by (otherwise Postgres might decide to return data in a different order and the test breaks for no obvious reason)
___
*[`tests/datatransfer/test_label_studio.py` line 180 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzodk8U5OGN6cccCV_b:-Nzodk8U5OGN6cccCV_c:bgcg5q4) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L180)):*
> ```Python
>         from pixeltable.datatransfer.label_studio import LabelStudioProject
> 
>         remote = LabelStudioProject.create('test_sync_errors_project', self.test_config)
> ```

Use create_label_studio_project() to ensure that you're testing the same code paths that a user would exercise.
___
*[`tests/datatransfer/test_label_studio.py` line 241 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzoedj07Jfl7LAUAug7:-Nzoedj07Jfl7LAUAug8:b-txi87s) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L241)):*
> ```Python
>     _logger.info('Setting up a venv the Label Studio pytext fixture.')
>     subprocess.run('python -m venv target/ls-env'.split(' '), check=True)
>     if platform.system() == 'Windows':
> ```

Why this? I thought this doesn't run on Windows anyway.
___
*[`tests/datatransfer/test_remote.py` line 18 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzof3aP5m8ayxrjNoN3:-Nzof3aP5m8ayxrjNoN4:b9ycu8v) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_remote.py#L18)):*
> ```Python
>         t = pxt.create_table('test_remote', schema)
> 
>         remote1 = MockRemote(
> ```

Why not use the public API for this?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)* status: 0 of 42 files reviewed, 16 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/catalog/table.py` line 726 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznKm8YD5GiT1g867SR:-NzogL4E1ohr74LMSjdg:bbjz0hn) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L726)):*

It would affect the API, because the col_mapping parameter will go away.

This doesn't look like much work to me.
</blockquote></details>

see my comment on slack. It's considerably more work than you realize, and it doesn't affect the API since _link is now protected.
___
*[`pixeltable/catalog/table.py` line 849 at r11](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NznODj-5ih8-kfyO2g_:-NzogIvD7NX9Fhqlcyf8:b-f70qm1) ([raw file](https://github.com/pixeltable/pixeltable/blob/88595e409af3e966d1be4815b4e86fc56142510b/pixeltable/catalog/table.py#L849)):*

Isn't that exactly not the case? import == True means data will be imported *into* this table.
</blockquote></details>

Oops typo. I'll fix this. Good catch
___
*[`tests/datatransfer/test_label_studio.py` line 125 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzobn779XD1kxzVoUU7:-NzogeK-CgDEQN6-zoUp:bnz8qa3) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L125)):*

You're expecting a particular order, which means you need an order_by (otherwise Postgres might decide to return data in a different order and the test breaks for no obvious reason)
</blockquote></details>

Ah good catch, I'll fix this

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)* status: 0 of 42 files reviewed, 16 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`tests/datatransfer/test_label_studio.py` line 96 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-NzoaUPr9WKtbtI-Oqeo:-NzoiRUB51Unz8EliR6u:b-886gsx) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L96)):*

This shouldn't be needed.
</blockquote></details>

Fixed
___
*[`tests/datatransfer/test_label_studio.py` line 241 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzoedj07Jfl7LAUAug7:-NzoiTZz3N5tlabIy-1H:b-pi3lmg) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L241)):*

Why this? I thought this doesn't run on Windows anyway.
</blockquote></details>

It's part of an earlier attempt to get it working on Windows. I have visions of that effort eventually succeeding, so I'd like to leave this in
___
*[`tests/datatransfer/test_remote.py` line 18 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzof3aP5m8ayxrjNoN3:-NzoibzC4ooumc_qPsiJ:bv9rmb6) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_remote.py#L18)):*

Why not use the public API for this?
</blockquote></details>

Not sure what this comment is asking, the goal here is to isolate the remote validation behavior (which i guess will become the project validation behavior) without depending on label studio or another specific implementation

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/139)* status: 0 of 42 files reviewed, 16 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`tests/datatransfer/test_label_studio.py` line 241 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzoedj07Jfl7LAUAug7:-Nzok7Bo4TFnp0oRmgsz:b-76zay0) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_label_studio.py#L241)):*

It's part of an earlier attempt to get it working on Windows. I have visions of that effort eventually succeeding, so I'd like to leave this in
</blockquote></details>

Leave explanatory comment
___
*[`tests/datatransfer/test_remote.py` line 18 at r12](https://reviewable.io/reviews/pixeltable/pixeltable/139#-Nzof3aP5m8ayxrjNoN3:-NzokSN4DM-UtCLJRh5n:b-ju4ulp) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd3d661d455a50b4941b2bdd2edcf288acce1c37/tests/datatransfer/test_remote.py#L18)):*

Not sure what this comment is asking, the goal here is to isolate the remote validation behavior (which i guess will become the project validation behavior) without depending on label studio or another specific implementation
</blockquote></details>

In general, it's better to use the public-facing API, to make sure that the error condition bubbles up in the right way. Leave todo.

mkornacker (approved):

PR #140, opened 2024-04-25
fix ci runner versions

orm011:
Ensure we test both an old x86 mac  (which is equivalent to current macos-latest )
as well as a newer arm mac (mostly to detect issues with any present and future binary dependencies of pixeltable, the ptyhon side should be ok). 
Also name runners so the meaning doesnt change (macos-latest is switching form x86 to arm64, or did in pgserver)

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/140)

orm011:
made similar changes  in pdf branch so it passes tests.

PR #141, opened 2024-04-25
Resolves an issue where Pixeltable fails to resolve function references in metadata

aaron-siegel:
Resolves an issue where Pixeltable fails to resolve function references in metadata.

If a table contains a computed column with a udf, loading metadata will throw an exception unless either:
- the udf is in pixeltable.functions, or
- the udf is fully imported in __init.py__ through every ancestor module.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/141)

PR #146, opened 2024-04-27
An `import_pandas()` method for converting a Pandas DataFrame into an InsertableTable

aaron-siegel:
Introduces a new `pixeltable.io` package with an `import_pandas()` function. Also included are `import_csv()` and `import_excel()` convenience methods, which delegate to the corresponding Pandas importers.

Reorganizes the existing importers (huggingface/parquet) into the new `pixeltable.io` package.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)

orm011:
*[`pixeltable/datatransfer/pandas.py` line 98 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDgkgK-gw4p9nZd76L:-NxDgkgK-gw4p9nZd76M:b-76vfue) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L98)):*
> ```Python
>             val = bool(val)
>         if pxt_type.is_string_type():
>             val = str(val)
> ```

seems like this would take in any type, since everything can be converted to string in python. ie, you would end up with random objects <obj at 0xq3r2354> in the DB. wouldnt it be better to check the value is actually a string? I think we may want to look at the values for objects (this enables images etc) rather than only the pandas schema.
We can delay the part of using values to determine types rather than only schemas, but maybe not just call str(), just check it is indeed a string to begin with.

orm011 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: 0 of 12 files reviewed, 7 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pyproject.toml` line 85 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDZi_D-WFVn_LluaYt:-NxDZi_D-WFVn_LluaYu:b74pvg1) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pyproject.toml#L85)):*
> ```TOML
> transformers = "^4.20"
> datasets = ">=2.15.0"
> openpyxl = ">=3.1"  # Excel (.xlsx) support
> ```

I'm thinking we may want to adopt a convention of a comment to explain why a dep or test dep is needed. Its often not obvious, and we seem to adopt a different attitude to this than to code, where non-obvious is meant to require a comment. eg. why did we need `regex` 
___
*[`pixeltable/client.py` line 174 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxD_l_R7BLxrBLBUN4T:-NxD_l_R7BLxrBLBUN4U:b-s37pue) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/client.py#L174)):*
> ```Python
>             self,
>             table_path: str,
>             io,
> ```

what's the reason for using io here, but filepath_or_buffer for csv?
___
*[`pixeltable/client.py` line 192 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDaB4k7D7eeZ09FBP1:-NxDaB4k7D7eeZ09FBP2:bm6iy4a) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/client.py#L192)):*
> ```Python
>             df: pd.DataFrame,
>             *,
>             schema: Optional[dict[str, ts.ColumnType]] = None
> ```

for parquet, we use 'schema_override', the exact name doesn't matter to me, should we aim for consistency?  
___
*[`pixeltable/datatransfer/pandas.py` line 22 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDdXYn2fauF2AdDjuh:-NxDdXYn2fauF2AdDjui:baljgqx) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L22)):*
> ```Python
>     else:
>         # Validate the specified schema
>         if len(schema) != len(df.columns):
> ```

For parquet and HF imports, we allowed users to provide a schema for a subset, that way they can simply override mapping only for those columns that would be ambiguous, eg a column of image paths, or a schema for a column where the default was somehow not appropriate (eg int vs float). Note that pandas will make int columns be floats in the presence of None (which it represents as NaN), but the DB can manage nullable without resorting to this.
___
*[`pixeltable/datatransfer/pandas.py` line 76 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDfTXi5aH1y9T5BLZr:-NxDfTXi5aH1y9T5BLZs:bingxmv) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L76)):*
> ```Python
>     Infers a Pixeltable type based on a Numpy dtype.
>     """
>     if np.issubdtype(np_dtype, np.integer):
> ```

I believe we only handle 32 bit ints, ie 64 bit ones would be lossy. this is recurring theme in our imports.
___
*[`pixeltable/datatransfer/pandas.py` line 82 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDftSb1Y6uA12RYpOG:-NxDftSc2rUtG2M_ECnh:bxnzaca) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L82)):*
> ```Python
>     if np.issubdtype(np_dtype, np.bool_):
>         return pxt.BoolType()
>     if np_dtype == np.object_ or np.issubdtype(np_dtype, np.character):
> ```

a pandas DF with images would also show up as object. 
Is there a reason we use np types as opposed to pandas types? I think pandas has its own namespace.

orm011 (changes requested):
> Reviewed 1 of 4 files at r1, 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: all files reviewed, 8 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/datatransfer/pandas.py` line 76 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDfTXi5aH1y9T5BLZr:-NxDizuSDowHNMb1clTl:bw0fcsb) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L76)):*

I believe we only handle 32 bit ints, ie 64 bit ones would be lossy. this is recurring theme in our imports.
</blockquote></details>

(same problem with Float vs Double)
___
*[`pixeltable/tests/datatransfer/test_pandas.py` line 9 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDjRg6EGShne4chCDH:-NxDjRg6EGShne4chCDI:bjwrp4k) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/tests/datatransfer/test_pandas.py#L9)):*
> ```Python
> 
> 
> class TestPandas:
> ```

Looks like none of the tests include a case of file paths eg to images or one of the other types. I think those are realistic use cases we should be testing.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: 9 of 12 files reviewed, 8 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pyproject.toml` line 85 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDZi_D-WFVn_LluaYt:-NxE9ccJ3vmJlQa2Jc1e:bu3hzb6) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pyproject.toml#L85)):*

I'm thinking we may want to adopt a convention of a comment to explain why a dep or test dep is needed. Its often not obvious, and we seem to adopt a different attitude to this than to code, where non-obvious is meant to require a comment. eg. why did we need `regex`
</blockquote></details>

I like this idea.
___
*[`pixeltable/client.py` line 174 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxD_l_R7BLxrBLBUN4T:-NxE9eSNDvUQ_jds3eC6:b-d3uz6y) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/client.py#L174)):*

what's the reason for using io here, but filepath_or_buffer for csv?
</blockquote></details>

I'm just following the nomenclature in the pandas library (mirroring the parameter names in `pd.read_csv` and `pd.read_excel`).
___
*[`pixeltable/client.py` line 192 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDaB4k7D7eeZ09FBP1:-NxE9kt5DLZRCVMo9tox:botiyeg) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/client.py#L192)):*

for parquet, we use 'schema_override', the exact name doesn't matter to me, should we aim for consistency?
</blockquote></details>

Sure... I prefer just `schema`, I guess.
___
*[`pixeltable/datatransfer/pandas.py` line 22 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDdXYn2fauF2AdDjuh:-NxEA90g6iTKJ7WXAExQ:b-m272dw) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L22)):*

For parquet and HF imports, we allowed users to provide a schema for a subset, that way they can simply override mapping only for those columns that would be ambiguous, eg a column of image paths, or a schema for a column where the default was somehow not appropriate (eg int vs float). Note that pandas will make int columns be floats in the presence of None (which it represents as NaN), but the DB can manage nullable without resorting to this.
</blockquote></details>

Right... there's an open question whether we should map pandas NaN to None or NaN (as they're different in pixeltable).

I don't know how I feel about allowing for a partial schema mapping. Both these questions maybe entail a discussion at the eng meeting. I'll add them as agenda items.
___
*[`pixeltable/datatransfer/pandas.py` line 76 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDfTXi5aH1y9T5BLZr:-NxEBDmD-EnxZEPJaN0v:b-xvpry9) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L76)):*

(same problem with Float vs Double)
</blockquote></details>

This doesn't seem to be true.
```
t = cl.create_table('test', {'n': pxt.IntType()})
t.insert(n=2**62)
t.show()
```

```
                     n
0  4611686018427387904
```

I agree there's an issue with doubles vs. floats but I don't think the loss of precision there is such a big deal (and anyway, what's the alternative?)
___
*[`pixeltable/datatransfer/pandas.py` line 82 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDftSb1Y6uA12RYpOG:-NxEB8vn9Jak5hUNoAXS:bta9ydd) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L82)):*

a pandas DF with images would also show up as object. 
Is there a reason we use np types as opposed to pandas types? I think pandas has its own namespace.
</blockquote></details>

Most of the tutorials I found use numpy types? Are they not the same? Can you point me to some examples/references on what you have in mind here?
___
*[`pixeltable/datatransfer/pandas.py` line 98 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDgkgK-gw4p9nZd76L:-NxEC9-r7Z4B5b3s2OJq:bvn3cs) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L98)):*

seems like this would take in any type, since everything can be converted to string in python. ie, you would end up with random objects <obj at 0xq3r2354> in the DB. wouldnt it be better to check the value is actually a string? I think we may want to look at the values for objects (this enables images etc) rather than only the pandas schema.
We can delay the part of using values to determine types rather than only schemas, but maybe not just call str(), just check it is indeed a string to begin with.
</blockquote></details>

Hmm, I'm not necessarily opposed to this, but it needs some discussion. Of course it's easy to check whether it's a string, but the more interesting question is what to do when it's not. What are your thoughts here? The options I can think of are:
- simply convert it to a string (that's the current behavior)
- throw an exception (this seems exceedingly fussy - picture someone trying to load a dataset that's 99% fine but contains one stray unknown object. if I were a user and pixeltable said "sorry, no", it'd drive me nuts)
- set the value to None and put something informative in the cell's errormsg

The reason I went with the present implementation is I think it's clearest to the user ... I could make a case that putting "<obj at 0xq3r2354>" into the table is preferable to doing something fussy / less clear.
___
*[`pixeltable/tests/datatransfer/test_pandas.py` line 9 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDjRg6EGShne4chCDH:-NxEWNqG65lH4sFAMWBr:b-kyn3hf) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/tests/datatransfer/test_pandas.py#L9)):*

Looks like none of the tests include a case of file paths eg to images or one of the other types. I think those are realistic use cases we should be testing.
</blockquote></details>

Good idea. I added a test with some image paths and an ImageType() schema to verify that they load properly.

orm011:
*[`pixeltable/datatransfer/pandas.py` line 82 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDftSb1Y6uA12RYpOG:-NxIqDN2BHdaX0q0wIyy:b96kywq) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L82)):*

Most of the tutorials I found use numpy types? Are they not the same? Can you point me to some examples/references on what you have in mind here?
</blockquote></details>

this has some pointers, https://pandas.pydata.org/docs/reference/arrays.html
this one caught my attention, it may be the most relevant, eg for handling of strings. 
https://pandas.pydata.org/docs/reference/api/pandas.api.types.infer_dtype.html

orm011 (changes requested):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: all files reviewed, 6 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/datatransfer/pandas.py` line 76 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDfTXi5aH1y9T5BLZr:-NxImNPo8__lCbtK4zIE:b384l03) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L76)):*

This doesn't seem to be true.
```
t = cl.create_table('test', {'n': pxt.IntType()})
t.insert(n=2**62)
t.show()
```

```
                     n
0  4611686018427387904
```

I agree there's an issue with doubles vs. floats but I don't think the loss of precision there is such a big deal (and anyway, what's the alternative?)
</blockquote></details>

looks like we map pxt.Integer to sql.BigInteger. But np.uint64 would map where.  We may need 
to move pixeltable to use sql double precision for storage instead of float so the conversion is not lossy. At least this way there is a consistent story for int and float.  Later on pxt can offer more types. (float, double, int, long int), or we can just add them.

Didn't we have a chat about this same issue for HF datasets? in that case we ended up erroring. Not saying it was a good decision. 
___
*[`pixeltable/datatransfer/pandas.py` line 82 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDftSb1Y6uA12RYpOG:-NxIsTyF8HEdZNXOGLDe:brijtsq) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L82)):*

this has some pointers, https://pandas.pydata.org/docs/reference/arrays.html
this one caught my attention, it may be the most relevant, eg for handling of strings. 
https://pandas.pydata.org/docs/reference/api/pandas.api.types.infer_dtype.html
</blockquote></details>

Maybe that method deals with nans better and saves us some work.
___
*[`pixeltable/datatransfer/pandas.py` line 98 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDgkgK-gw4p9nZd76L:-NxIspZL2ZjYV_ukF-fF:bgp1pqq) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L98)):*

Hmm, I'm not necessarily opposed to this, but it needs some discussion. Of course it's easy to check whether it's a string, but the more interesting question is what to do when it's not. What are your thoughts here? The options I can think of are:
- simply convert it to a string (that's the current behavior)
- throw an exception (this seems exceedingly fussy - picture someone trying to load a dataset that's 99% fine but contains one stray unknown object. if I were a user and pixeltable said "sorry, no", it'd drive me nuts)
- set the value to None and put something informative in the cell's errormsg

The reason I went with the present implementation is I think it's clearest to the user ... I could make a case that putting "<obj at 0xq3r2354>" into the table is preferable to doing something fussy / less clear.
</blockquote></details>

I guess I agree it may be best to take the string and then make sure it is easy to clean up the loaded table after the fact, or to add type annotations to specific columns (ie, these paths are to images)

I'm thinking of a situation of a column of  PIL. Image objects though (not paths but in-memory pil images). What should be the behavior then? probably not making it a string. We should infer it is an image.

pixeltable/datatransfer/pandas.py
```diff
@@ -0,0 +1,147 @@
+from typing import Optional, Any, Iterable
```
mkornacker: Move this into utils/pandas.py and get rid of pixeltable/datatransfer, this feels like a utility.

mkornacker (changes requested):

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: 9 of 15 files reviewed, 7 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/datatransfer/pandas.py` line 98 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/146#-NxDgkgK-gw4p9nZd76L:-NxJJpYECrxKnzgbVMRv:b-dz7vrq) ([raw file](https://github.com/pixeltable/pixeltable/blob/4371af3d13f9db9237ba5caff8eae79771cfbc43/pixeltable/datatransfer/pandas.py#L98)):*

I guess I agree it may be best to take the string and then make sure it is easy to clean up the loaded table after the fact, or to add type annotations to specific columns (ie, these paths are to images)

I'm thinking of a situation of a column of  PIL. Image objects though (not paths but in-memory pil images). What should be the behavior then? probably not making it a string. We should infer it is an image.
</blockquote></details>

We could add this as a follow-up feature. How often is it going to happen though? Is it really a common pattern to store PIL images in pandas dataframes? We'll never expect to see one when loading from csv/excel, right? Or does excel have embedded images or some such?
___
*[`pixeltable/datatransfer/pandas.py` line 1 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1592871356:-NxJJ2Rd4MJM2ec1ZXiO:bl5g57a) ([raw file](https://github.com/pixeltable/pixeltable/blob/4c3c3aef081b399a040418b5176921d3d5a320ba/pixeltable/datatransfer/pandas.py#L1)):*

Move this into utils/pandas.py and get rid of pixeltable/datatransfer, this feels like a utility.
</blockquote></details>

We should discuss this. I think it we should have a package for data importers. There are going to be a lot of them, and I'd argue it's correct to view them as features, not utilities. (Typically a utility is a small unit of code that supports multiple features across different parts of the codebase; that's not what this is.) `utils` is going to get very fat very quickly if we start putting stuff like this into it.

We don't need to call it `datatransfer` but I do think we need something other than a `utils` catch-all.

pixeltable/io/pandas.py
```diff
@@ -0,0 +1,144 @@
...
+
+import numpy as np
+
+import pandas as pd
+import pixeltable as pxt
+import pixeltable.exceptions as excs
+import pixeltable.type_system as ts
+
+
+def import_pandas(
```
mkornacker, line 11: missing docstring

pixeltable/io/pandas.py
```diff
@@ -0,0 +1,144 @@
...
+        df: pd.DataFrame,
+        *,
+        schema: Optional[dict[str, pxt.ColumnType]] = None
+) -> pxt.catalog.InsertableTable:
+    if schema is None:
+        # Infer schema
+        schema = _df_to_pxt_schema(df)
+    else:
+        # Validate the specified schema
+        if len(schema) != len(df.columns):
```
mkornacker: I don't think we should insist on that. In pd.read_csv(), the dtype parameter lets you override specific columns. It's pretty annoying to have to specify the full schema of a 20-column df when only 3 columns have dtype object and need to be overridden.

pixeltable/io/pandas.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def import_csv(
+        table_path: str,
+        filepath_or_buffer,
+        schema: Optional[dict[str, ts.ColumnType]] = None,
+        **kwargs
+) -> pxt.catalog.InsertableTable:
+    """
+    Creates a new `InsertableTable` from a csv file. This is a convenience method and is equivalent
```
mkornacker: Let's reference Table here, not InsertableTable. I don't think the user should have to be aware of the latter.

pixeltable/io/pandas.py
```diff
@@ -0,0 +1,144 @@
...
+    return schema
+
+
+def _normalize_pxt_col_name(pd_name: str) -> str:
+    """
+    Normalizes an arbitrary DataFrame column name into a valid Pixeltable identifier by:
+    - replacing any non-ascii or non-alphanumeric characters with an underscore _
+    - prefixing the result with the letter 'c' if it starts with an underscore or a number
+    """
+    id = ''.join(
```
mkornacker: 4 lines where 1 is perfectly readable...

pixeltable/io/pandas.py
```diff
@@ -0,0 +1,144 @@
...
+            val = str(val)
+        elif pxt_type.is_timestamp_type():
+            if pd.isnull(val):
+                # pandas has the bespoke 'NaT' type for a missing timestamp; postgres is very
+                # much not-ok with it. (But if we convert it to None and then load out the
+                # table contents as a pandas DataFrame, it will correctly restore the 'NaT'!)
+                val = None
+            else:
+                val = pd.Timestamp(val).to_pydatetime()
+        yield col_name, val
```
mkornacker: I don't think having a generator over a row serves any useful purpose here, you're never going to be dealing with 1000s of columns. Rows are generally assumed to be dict[str, Any], and it would be best to stick to that convention here as well.

pixeltable/io/parquet.py
```diff
@@ -132,11 +141,26 @@ def import_parquet(
     table_path: str,
     *,
     parquet_path: str,
-    schema_override: Optional[Dict[str, ts.ColumnType]],
+    schema_override: Optional[Dict[str, ts.ColumnType]] = None,
     **kwargs,
-) -> 'catalog.InsertableTable':
+) -> pxt.catalog.InsertableTable:
+    """Create a new `InsertableTable` from a Parquet file or set of files. Requires pyarrow to be installed.
```
mkornacker: Table

pixeltable/tests/io/test_pandas.py
```diff
@@ -0,0 +1,114 @@
...
+import numpy as np
+
+import pixeltable as pxt
+from ..utils import skip_test_if_not_installed
+
+
+class TestPandas:
+
+    def test_pandas_csv(self, reset_db) -> None:
+        from pixeltable.io import import_csv
```
mkornacker: Are we really opposed to having those functions be visible in the top-level namespace?

pixeltable/tests/io/test_pandas.py
```diff
@@ -0,0 +1,114 @@
...
+        _assert_equals_with_nans(result_set['float_n'], [1.0, 5.0, float('nan'), 1.0])
+        assert result_set['bool'] == [True, False, True, True]
+        assert result_set['string'] == ['fish', 'cake', 'salad', 'egg']
+        assert result_set['string_n'] == ['fish', 'cake', None, 'egg']
+        assert result_set['ts'] == [datetime.datetime(2024, 5, n) for n in range(3, 7)]
+        assert result_set['ts_n'] == [datetime.datetime(2024, 5, 3), None, None, datetime.datetime(2024, 5, 6)]
+
+    def test_pandas_images(self, reset_db) -> None:
+        skip_test_if_not_installed('boto3')  # This test relies on s3 URLs
+        from pixeltable.io.pandas import import_csv
```
mkornacker: Let's at the very least have all these be in pixeltable.io (not pixeltable.io.pandas).

Or maybe even just pixeltable.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: 8 of 26 files reviewed, 12 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/datatransfer/pandas.py` line 1 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1592871356:-NxViolU8Htzq8gM-bej:b-yx1srm) ([raw file](https://github.com/pixeltable/pixeltable/blob/4c3c3aef081b399a040418b5176921d3d5a320ba/pixeltable/datatransfer/pandas.py#L1)):*

We should discuss this. I think it we should have a package for data importers. There are going to be a lot of them, and I'd argue it's correct to view them as features, not utilities. (Typically a utility is a small unit of code that supports multiple features across different parts of the codebase; that's not what this is.) `utils` is going to get very fat very quickly if we start putting stuff like this into it.

We don't need to call it `datatransfer` but I do think we need something other than a `utils` catch-all.
</blockquote></details>

[Just making a note that per our conversation at standup, we're calling this `pixeltable.io`]
___
*[`pixeltable/io/pandas.py` line 11 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596206071:-NxVin3LCAeKlzzcw2D4:bv6s96d) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/pandas.py#L11)):*

missing docstring
</blockquote></details>

Fixed. Good catch.
___
*[`pixeltable/io/pandas.py` line 22 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596206673:-NxVhRZu6cpQmXeEMqIx:b2nhll9) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/pandas.py#L22)):*

I don't think we should insist on that. In pd.read_csv(), the dtype parameter lets you override specific columns. It's pretty annoying to have to specify the full schema of a 20-column df when only 3 columns have dtype object and need to be overridden.
</blockquote></details>

Ok.
___
*[`pixeltable/io/pandas.py` line 43 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596209165:-NxVg_-x-Of9K9SpR6wu:b-qvbx2k) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/pandas.py#L43)):*

Let's reference Table here, not InsertableTable. I don't think the user should have to be aware of the latter.
</blockquote></details>

Ok. I was just trying to be consistent with the existing documentation in the other importers (parquet and huggingface) but I can change those too.
___
*[`pixeltable/io/pandas.py` line 91 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596209662:-NxVgnDS6t516PRl_NgY:b-t8caud) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/pandas.py#L91)):*

4 lines where 1 is perfectly readable...
</blockquote></details>

sigh.

I actually think it's clearer when separated out like this. I find it harder to read when there's too many if/then/else/for/in etc keywords in a row, without any extra delineation. So this was very much intentional.

I can change it if you feel strongly otherwise, but I also don't think it's a productive use of time to have these debates. Better in my opinion to let the formatter decide, and move on to more important things.
___
*[`pixeltable/io/pandas.py` line 144 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596211271:-NxVhaFNEurEQDlf1UKT:b-8bx4np) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/pandas.py#L144)):*

I don't think having a generator over a row serves any useful purpose here, you're never going to be dealing with 1000s of columns. Rows are generally assumed to be dict[str, Any], and it would be best to stick to that convention here as well.
</blockquote></details>

It's not clear to me what you're suggesting here. The only invocation of this function gets wrapped in a dict immediately. What's the change you're asking for?
___
*[`pixeltable/io/parquet.py` line 147 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596211586:-NxVhdECDDuOhaizzhAK:b-gzht81) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/parquet.py#L147)):*

Table
</blockquote></details>

The use of `InsertableTable` here precedes this PR but I can change it.
___
*[`pixeltable/tests/io/test_pandas.py` line 12 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596212111:-NxViNUP7yuH5JalBUEo:b-32pjdu) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/tests/io/test_pandas.py#L12)):*

Are we really opposed to having those functions be visible in the top-level namespace?
</blockquote></details>

See comment below
___
*[`pixeltable/tests/io/test_pandas.py` line 73 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596212723:-NxVhlG08KLbxzJTvtU-:bj4lsiu) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/tests/io/test_pandas.py#L73)):*

Let's at the very least have all these be in pixeltable.io (not pixeltable.io.pandas).

Or maybe even just pixeltable.
</blockquote></details>

They ARE imported into `pixeltable.io`. See the immediately preceding comment! Either `pixeltable.io` or `pixeltable.io.pandas` will work here. I'll change it to the former, to be sure the import has test coverage. (I think wrote the test before creating the `pixeltable.io` namespace; that's the only reason it's qualified like it is)

I feel reasonably strongly that they should be in `pixeltable.io` but not `pixeltable`. We could discuss this at standup/eng meeting but I think they're perfectly discoverable in `pixeltable.io`, and it's consistent with how other libraries organize their APIs.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: 8 of 26 files reviewed, 12 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/io/pandas.py` line 91 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596209662:-NxVk548Av-Gk36Fh3dl:bieh9go) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/pandas.py#L91)):*

sigh.

I actually think it's clearer when separated out like this. I find it harder to read when there's too many if/then/else/for/in etc keywords in a row, without any extra delineation. So this was very much intentional.

I can change it if you feel strongly otherwise, but I also don't think it's a productive use of time to have these debates. Better in my opinion to let the formatter decide, and move on to more important things.
</blockquote></details>

I don't. It was just a gripe.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: 8 of 26 files reviewed, 12 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/io/pandas.py` line 144 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596211271:-NxVkDSP4ua7sxm1fHl1:b1odbpc) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/pandas.py#L144)):*

It's not clear to me what you're suggesting here. The only invocation of this function gets wrapped in a dict immediately. What's the change you're asking for?
</blockquote></details>

Why not just have this return a dict? It would make the logic completely self-contained and  easier to understand, and the performance implications are meaningless.
___
*[`pixeltable/tests/io/test_pandas.py` line 73 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596212723:-NxVkiOm8rdu5e9RqOQb:b-prftl2) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/tests/io/test_pandas.py#L73)):*

They ARE imported into `pixeltable.io`. See the immediately preceding comment! Either `pixeltable.io` or `pixeltable.io.pandas` will work here. I'll change it to the former, to be sure the import has test coverage. (I think wrote the test before creating the `pixeltable.io` namespace; that's the only reason it's qualified like it is)

I feel reasonably strongly that they should be in `pixeltable.io` but not `pixeltable`. We could discuss this at standup/eng meeting but I think they're perfectly discoverable in `pixeltable.io`, and it's consistent with how other libraries organize their APIs.
</blockquote></details>

In pandas, read_csv() is in the top-level namespace.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: 8 of 26 files reviewed, 12 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/io/pandas.py` line 22 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596206673:-NxVnmivD2G1UHXUIpOc:bgrmzk) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/pandas.py#L22)):*

Ok.
</blockquote></details>

Actually, I think this needs a little more thought. If the schemas are not one-to-one, then:
- Are the keys of the dict the Pandas column names or the Pixeltable column names?
- Do we want to provide any way to override the column *names*, or is the user stuck with them? (The advantage of the one-to-one schema substitution is you can change both the names AND the types.)

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: 8 of 26 files reviewed, 12 unresolved discussions (waiting on @mkornacker and @orm011)

___
*[`pixeltable/tests/io/test_pandas.py` line 73 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596212723:-NxVoqTT4Sp2c-Qoem3T:bexn6p9) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/tests/io/test_pandas.py#L73)):*

In pandas, read_csv() is in the top-level namespace.
</blockquote></details>

True, but `read_csv()` is a lot closer to the core of Pandas than to Pixeltable. I was thinking of libraries like LlamaIndex or LangChain, where loaders are typically in a subpackage.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: 8 of 26 files reviewed, 12 unresolved discussions (waiting on @aaron-siegel and @orm011)

___
*[`pixeltable/io/pandas.py` line 22 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596206673:-NxXexx9Ev0DdEepgki6:b-ffib50) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/io/pandas.py#L22)):*

Actually, I think this needs a little more thought. If the schemas are not one-to-one, then:
- Are the keys of the dict the Pandas column names or the Pixeltable column names?
- Do we want to provide any way to override the column *names*, or is the user stuck with them? (The advantage of the one-to-one schema substitution is you can change both the names AND the types.)
</blockquote></details>

read_csv() uses parameters usecols, names, dtype for that, eg, read_csv(usecols=[0, 1, 2], names=['col_1', 'col_2', 'col_3'], dtype={'col_1': ...})

Seems like a good pattern.

orm011 (commented):
> Reviewed 15 of 18 files at r6, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/146)* status: all files reviewed, 12 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/tests/io/test_pandas.py` line 73 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/146#gh-1596212723:-NxYEJje1LCFPU-jGAv7:b-n1tudu) ([raw file](https://github.com/pixeltable/pixeltable/blob/06b2d55951670de298d2b10a1056124af9d60a79/pixeltable/tests/io/test_pandas.py#L73)):*

True, but `read_csv()` is a lot closer to the core of Pandas than to Pixeltable. I was thinking of libraries like LlamaIndex or LangChain, where loaders are typically in a subpackage.
</blockquote></details>

The question here is how we will make it discoverable. auto complete from pxt wont work anymore afaik? That was the motivation for top level. 

Does pxt.io. autocomplete work without needing to separately import pxt.io by the user?

mkornacker (approved):

PR #147, opened 2024-04-27
Making Expr.slot_idx an Optional[int]

mkornacker:
Now using None to indicate an invalid index, instead of -1.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/147)

aaron-siegel (approved):
Good idea

PR #148, opened 2024-04-30
Function call return type resolution based on function args

mkornacker:
- simplifies the mechanism that used to be part of Signature (Signature.return_type could be a Callable) and moves it into Function.call_return_type()
- adding call type resolution to various function modules


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/148)

aaron-siegel (changes requested):
Overall lgtm with a couple comments.

Per conversation on Wed 5/1 we are planning to follow this up with some refactoring to use a decorator pattern for defining dynamic return types (similar to the pattern for `@property` / `@singledispatch` in the Python API).

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/148#-:-Nwu3pO28qppH9VGZkXU:b7ysy4l)* status: 0 of 20 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

___
*[`poetry.lock` line 1 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/148#-Nwu-FpH6ardbsaykoIb:-Nwu-FpH6ardbsaykoIc:b-o74qmg) ([raw file](https://github.com/pixeltable/pixeltable/blob/22c341d83cf6f52800fad82f992125fff064dc14/poetry.lock#L1)):*
> ```TOML
> # This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
> ```

Per our conversation earlier this week: don't forget to upgrade to poetry 1.8.2
___
*[`pixeltable/functions/openai.py` line 195 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/148#-Nwu4Q8UCw72AyFm1fMq:-Nwu4Q8UCw72AyFm1fMr:b3rhm9r) ([raw file](https://github.com/pixeltable/pixeltable/blob/22c341d83cf6f52800fad82f992125fff064dc14/pixeltable/functions/openai.py#L195)):*
> ```Python
>         if model not in _embedding_dimensions_cache:
>             # TODO: this will fail because Env hasn't been initialized
>             result = openai_client().embeddings.create(input='This is a test', model=model, encoding_format='float')
> ```

This is clever, but I wonder if it will surprise the user by initiating an API call at an unexpected time. (For example, if I use this function in a computed column, it will cause the OpenAI API to be contacted even if the table contains no data.)

There are other considerations. Calling OpenAI takes an indeterminate amount of time. Are we anticipating a future where the return type can be defined asynchronously and is 'unknown' until then? Or are we going to accept that there may be a bottleneck here?

What if the call to the API fails? It definitely needs to be guarded (it could fail for any number of reasons). But even if it's guarded, it's inherently nondeterministic whether a particular API call to OpenAI succeeds. This is (obviously) fine when we're executing a computation, but here we're introducing nondeterminism into the *definition* of a workflow: once the computed column has been defined, the repercussions are long-lived, in the sense that the optimizer's behavior will depend indefinitely on what happens to this single API call. (If I understand things right.)

So I guess I'd feel more comfortable not doing this, and relying on the embedding being in the predefined dimensions cache, or the dimension being explicitly specified by the user. Likewise for Together/Fireworks/other remote APIs. A couple options would be:
- Fall back on dimensionless ArrayType
- Refuse to instantiate (throw an exception) if the embedding is not in the cache and no explicit dimension is specified

I could also envision a separate `infer_dimensions` helper method (or parameter?), that does this clever thing, but in a way that explicitly makes clear it may take a long time and may fail, but I think that would benefit from some more thought.

aaron-siegel (approved):

PR #149, opened 2024-05-01
Suggestion for handling dynamic embedding dims [PROPOSAL]

aaron-siegel:
Here's a suggestion for an alternate way of handling embedding dimensions that vary with the model parameter. It's a simpler change (not reliant on "parameter-dependent" return types), with an added advantage: it can be used directly in index creation, in most cases saving the user from having to deal with `expr_udf` boilerplate. The user can just directly write

`t.add_embedding_index('category', text_embed=together.embedding_fn('my_model'))`

(for example).

Ad hoc creation of `expr_udf`s should be safe, since (unlike with ordinary `udf`s) we'll never actually need the path to the function.

**This is an example/template only - not a working implementation.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/149)

PR #150, opened 2024-05-01
Fixes a bug when loading DB schemas containing an `InlineArray`

aaron-siegel:
PR #147 introduced a bug that causes DB schema loading to fail when the schema contains an `InlineArray`. The bug was not caught in CI due to an unrelated bug in `test_migration.py`. This PR fixes both bugs and increments the schema version.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/150)

pixeltable/exprs/inline_array.py
```diff
@@ -100,7 +100,7 @@ def _from_dict(cls, d: Dict, components: List[Expr]) -> Expr:
         assert 'elements' in d
         arg: List[Any] = []
         for idx, val in d['elements']:
-            if idx is not None:
+            if idx is not None and idx >= 0:  # Older schemas might have -1 instead of None
```
mkornacker, line 105: You'd need to apply the same logic to InlineDict.

But let's talk about this. I understand it's a fix, but I don't love having stuff like that hang around the codebase forever.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/150)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`pixeltable/exprs/inline_array.py` line 103 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/150#gh-1586787194:-Nwv6n1C6fyXGafJlgEK:b-etuk5f) ([raw file](https://github.com/pixeltable/pixeltable/blob/8f8bc58400fcec93538ad07730d567ed54de0f6f/pixeltable/exprs/inline_array.py#L103)):*

You'd need to apply the same logic to InlineDict.

But let's talk about this. I understand it's a fix, but I don't love having stuff like that hang around the codebase forever.
</blockquote></details>

I added a TODO that this needs to be fixed.

mkornacker (approved):

PR #151, opened 2024-05-02
Fix setuptools version to 69.1

aaron-siegel:
Fixes the setuptools version to 69.1. There is some anecdotal evidence that setuptools >= 69.2 causes problems with Github Actions.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/151)

mkornacker (approved):

PR #153, opened 2024-05-03
Ensure http server works in windows, add pdf thumbnail + link for doc type

orm011:
also, modify unit tests to be modules and to use relative imports of utils and conftest (needed for them to work without in non-dev env)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/153)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/153)* status: 0 of 31 files reviewed, 7 unresolved discussions (waiting on @orm011)

___
*[`pixeltable/dataframe.py` line 155 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-BnIFE4nSq1Li_v8s:-Nx-BnIFE4nSq1Li_v8t:b8h29zu) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/dataframe.py#L155)):*
> ```Python
>                 p = doc.get_page_pixmap(0)
>                 while p.width > max_width or p.height > max_height:
>                     p.shrink(1)
> ```

Add a comment explaining what this does (resizes by 1/2 on each dimension). It's a very strange API  on PyMuPDF's part and not obvious at all (I had to dig through the docs to understand it).
___
*[`pixeltable/tests/test_dataframe.py` line 200 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-DVRrB94yoeqNzJrW:-Nx-DNbI81hX_ZHU8UgO:b-pbsk5h) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/tests/test_dataframe.py#L200)):*
> ```Python
>                                                           'doc': pxt.DocumentType()})
> 
>         pdf_docs = [ f for f in get_documents() if f.endswith('.pdf') ]
> ```

Better would be (since you just need one):

`pdf_doc = next(f for f in get_documents() if f.endswith('.pdf'))`
___
*[`pixeltable/tests/test_dataframe.py` line 210 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-DnRk9G31mNzlDu8l:-Nx-DnRl0HRfVE-wmAUM:b8yushc) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/tests/test_dataframe.py#L210)):*
> ```Python
>         assert len(video_tags) == 1
>         audio_tags = doc.find_all('audio')
>         assert len(audio_tags) == 1
> ```

Spurious duplicate assertion
___
*[`pixeltable/utils/http_server.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-EusH9mFJDvpnGl_U:-Nx-EusH9mFJDvpnGl_V:bkfs38p) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/utils/http_server.py#L13)):*
> ```Python
> 
> 
> def get_file_uri(http_address: str, file_path: str) -> str:
> ```

Are you sure we need to do all this manually? The usual way of doing this is with `urllib.request.pathname2url`. Parsing filepaths and URLs is fraught and we should strive to avoid doing it manually whenever feasible. At minimum, let's add a comment explaining why the library methods are inadequate and we have to resort to string manipulations.
___
*[`pixeltable/utils/http_server.py` line 26 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-FqXO6p6AH3mhQzGR:-Nx-FqXO6p6AH3mhQzGS:btdvm6v) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/utils/http_server.py#L26)):*
> ```Python
> 
> def path_to_parts(uri_path: str) -> tuple[str, str]:
>     """Split a URI path into an anchor and a path"""
> ```

Likewise, the usual way of doing this is

```
parsed = urllib.parse.urlparse(uri_path)
urllib.parse.unquote(urllib.request.url2pathname(parsed.path))
```

I'd like to understand why we can't use that here.
___
*[`pixeltable/utils/http_server.py` line 41 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-GCEcBXF3HHlFK-7u:-Nx-GCEcBXF3HHlFK-7v:b-oof86l) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/utils/http_server.py#L41)):*
> ```Python
> 
> 
> class AbsolutePathHandler(http.server.SimpleHTTPRequestHandler):
> ```

I think this may explain a bunch of my other comments. Are we using absolute pathnames in our URLs? I don't think that's such a good idea. Typically a webserver would use pathnames relative to some webdocs root. If that's a controversial suggestion, then I'd like to understand what's the rationale for using absolute pathnames, and have a discussion about alternatives.
___
*[`pixeltable/utils/http_server.py` line 53 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-EOnI7izIIjoOnQWB:-Nx-EOnI7izIIjoOnQWC:b-h7f4lc) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/utils/http_server.py#L53)):*
> ```Python
>             Translate a /-separated PATH to the local filename syntax.
>             overrides http.server.SimpleHTTPRequestHandler.translate_path
>             it will translate http://localhost/c:/foo/bar to c:/foo/bar
> ```

This all seems really nasty and full of potential gotchas. Why are we getting URLs like `http://localhost/c:/foo/bar` in the first place? That seems like something that should never happen. (`file://c:/foo/bar` is normal, but we shouldn't be seeing drive letters in http urls...)

orm011 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/153)* status: 0 of 31 files reviewed, 7 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/dataframe.py` line 155 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-BnIFE4nSq1Li_v8s:-Nx-NZqV5B_4epBUU4CM:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/dataframe.py#L155)):*

Add a comment explaining what this does (resizes by 1/2 on each dimension). It's a very strange API  on PyMuPDF's part and not obvious at all (I had to dig through the docs to understand it).
</blockquote></details>

Done.
___
*[`pixeltable/tests/test_dataframe.py` line 200 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-DVRrB94yoeqNzJrW:-Nx-Nk3sBUtmlnWJDdAo:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/tests/test_dataframe.py#L200)):*

Better would be (since you just need one):

`pdf_doc = next(f for f in get_documents() if f.endswith('.pdf'))`
</blockquote></details>

Done.
___
*[`pixeltable/tests/test_dataframe.py` line 210 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-DnRk9G31mNzlDu8l:-Nx-NmzQ-NVyMSeFvtq4:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/tests/test_dataframe.py#L210)):*

Spurious duplicate assertion
</blockquote></details>

Done.
___
*[`pixeltable/utils/http_server.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-EusH9mFJDvpnGl_U:-Nx3Bqqa4FQeh6CyLl3M:bpyd4er) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/utils/http_server.py#L13)):*

Are you sure we need to do all this manually? The usual way of doing this is with `urllib.request.pathname2url`. Parsing filepaths and URLs is fraught and we should strive to avoid doing it manually whenever feasible. At minimum, let's add a comment explaining why the library methods are inadequate and we have to resort to string manipulations.
</blockquote></details>

Thanks for the suggestions, I've switched. 

 I originally modeled it on http.server.SimpleHTTPServer, which itself does not use this. Also looked around in http.parse for it, since quote() and unquote() are there, somehow that http.request has what is basically a parsing method.
___
*[`pixeltable/utils/http_server.py` line 26 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-FqXO6p6AH3mhQzGR:-Nx3CDzEDLEldHJAEKD_:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/utils/http_server.py#L26)):*

Likewise, the usual way of doing this is

```
parsed = urllib.parse.urlparse(uri_path)
urllib.parse.unquote(urllib.request.url2pathname(parsed.path))
```

I'd like to understand why we can't use that here.
</blockquote></details>

Done.
___
*[`pixeltable/utils/http_server.py` line 41 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-GCEcBXF3HHlFK-7u:-Nx-JXhB8ma-768-nxaR:b5c6pec) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/utils/http_server.py#L41)):*

I think this may explain a bunch of my other comments. Are we using absolute pathnames in our URLs? I don't think that's such a good idea. Typically a webserver would use pathnames relative to some webdocs root. If that's a controversial suggestion, then I'd like to understand what's the rationale for using absolute pathnames, and have a discussion about alternatives.
</blockquote></details>

Yes, currently we use absolute paths. The only thing we are adding in this code reivew is ability to do it in windows, where there is no equivalent of unix root. SimpleHTTPServer is ok with being root in unix, but there are multiple roots in Windows. That's all this is.

Remember, this is for use with the notebook in localhost, a web-app which which already has full access to everything in the machine, you're not gaining any more access than you had before. 

Currently, pixeltable will take any file path you give it, files live anywhere and we don't move them.  
___
*[`pixeltable/utils/http_server.py` line 53 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-EOnI7izIIjoOnQWB:-Nx-Kxvq8NTZwSNJQBYp:b-cm6bcj) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/utils/http_server.py#L53)):*

This all seems really nasty and full of potential gotchas. Why are we getting URLs like `http://localhost/c:/foo/bar` in the first place? That seems like something that should never happen. (`file://c:/foo/bar` is normal, but we shouldn't be seeing drive letters in http urls...)
</blockquote></details>

I've replaced the parsing logic so it uses the methods from the python library you recommended,  they take care of parsing gotchas. pathname2url does preserve //c:/foo/bar.

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/153)* status: 0 of 31 files reviewed, 7 unresolved discussions (waiting on @orm011)

___
*[`pixeltable/dataframe.py` line 155 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-BnIFE4nSq1Li_v8s:-NxE8MW1745yBSnTaG65:bw6e1li) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/dataframe.py#L155)):*

Done.
</blockquote></details>

Are you sure shrink(1).shrink(1) is less than shrink(2)? That wasn't my interpretation of the docs. I thought shrink(2) reduces each dimension by a factor of 2^2 = 4. Maybe double check this information to ensure the comment is accurate. (Although, I think it's sufficient to just point out that p.shrink(1) halves each dimension)

Also (very pedantic comment) the verb form of "half" is "halve"
___
*[`pixeltable/utils/http_server.py` line 41 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/153#-Nx-GCEcBXF3HHlFK-7u:-NxE9Gqq885pPqZesjfD:b-tbjz9i) ([raw file](https://github.com/pixeltable/pixeltable/blob/a56018dd387a9797b49ea22ad705eefa3b791011/pixeltable/utils/http_server.py#L41)):*

Yes, currently we use absolute paths. The only thing we are adding in this code reivew is ability to do it in windows, where there is no equivalent of unix root. SimpleHTTPServer is ok with being root in unix, but there are multiple roots in Windows. That's all this is.

Remember, this is for use with the notebook in localhost, a web-app which which already has full access to everything in the machine, you're not gaining any more access than you had before. 

Currently, pixeltable will take any file path you give it, files live anywhere and we don't move them.
</blockquote></details>

Got it, ok.

PR #156, opened 2024-05-04
.isin() pseudo-function/InPredicate

mkornacker:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/156)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/156)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/table.py` line 101 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0Llmv9flmcIZW8r_x:-Nx0Llmv9flmcIZW8r_y:b-3ddj0y) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/catalog/table.py#L101)):*
> ```Python
> 
>     def group_by(self, *items: 'exprs.Expr') -> 'pixeltable.dataframe.DataFrame':
>         """Return a DataFrame for this table."""
> ```

wrong docstring?
___
*[`pixeltable/exprs/in_predicate.py` line 16 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0LrBQ1sR7XHdQMVaE:-Nx0LrBQ1sR7XHdQMVaF:baxdaok) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/exprs/in_predicate.py#L16)):*
> ```Python
> 
> class InPredicate(Predicate):
>     def __init__(self, lhs: Expr, value_list: list[Any]):
> ```

This seems unnecessarily restrictive. Instead of `value_list: list`, why not `values: Union[Iterable, dict]`?

I believe that the only change this would require is to replace
```
self.value_list = value_list
```
by
```
if isinstance(values, dict):
    self.value_list = list(values.keys())
else:
    self.value_list = list(values)
```
and everything else should just work.

Sooner or later someone is going to use `isin` on a set or dict and be surprised that it doesn't work as expected.
___
*[`pixeltable/exprs/in_predicate.py` line 26 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0Mfpb9vDS5HpUN_kJ:-Nx0Mfpb9vDS5HpUN_kK:b-8lsge) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/exprs/in_predicate.py#L26)):*
> ```Python
>             except TypeError:
>                 raise excs.Error(
>                     f'isin(): list item {val!r} is not compatible with the type of {lhs}, which is {lhs.col_type}')
> ```

I don't see why this constraint should be necessary?
___
*[`pixeltable/tests/test_exprs.py` line 359 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0NUqU2fRN1toHFg4e:-Nx0NUqU2fRN1toHFg4f:b-yt3pgi) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/tests/test_exprs.py#L359)):*
> ```Python
>         print(_)
> 
>     def test_in(self, test_tbl: catalog.Table) -> None:
> ```

Also add a computed column that uses `isin` to `pixeltable/tool/create_test_db_dump.py`
___
*[`pixeltable/tests/test_exprs.py` line 366 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0NLq8CnmVxkNkCqpT:-Nx0NLq8CnmVxkNkCqpU:b-hz9f40) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/tests/test_exprs.py#L366)):*
> ```Python
>         with pytest.raises(excs.Error) as excinfo:
>             # incompatible values
>             _ = t.where(t.c2.isin([1, 'a'])).collect()
> ```

cf. the preceding comment; why shouldn't this be allowed?

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/156)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/catalog/table.py` line 101 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0Llmv9flmcIZW8r_x:-Nx0_Ix9EaYiFL1GTfUN:bipmnac) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/catalog/table.py#L101)):*

wrong docstring?
</blockquote></details>

It's the same as all the other ones. 
___
*[`pixeltable/exprs/in_predicate.py` line 16 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0LrBQ1sR7XHdQMVaE:-Nx0_wuq6jI9xb8bquUN:b8ceuqr) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/exprs/in_predicate.py#L16)):*

This seems unnecessarily restrictive. Instead of `value_list: list`, why not `values: Union[Iterable, dict]`?

I believe that the only change this would require is to replace
```
self.value_list = value_list
```
by
```
if isinstance(values, dict):
    self.value_list = list(values.keys())
else:
    self.value_list = list(values)
```
and everything else should just work.

Sooner or later someone is going to use `isin` on a set or dict and be surprised that it doesn't work as expected.
</blockquote></details>

We can do Iterable[Any], but I don't see how a dict fits in.

Pandas does not accept dicts for isin():
https://pandas.pydata.org/docs/reference/api/pandas.Series.isin.html
___
*[`pixeltable/exprs/in_predicate.py` line 26 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0Mfpb9vDS5HpUN_kJ:-Nx0aQnp6LtXl85H8Oqn:b11ljgb) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/exprs/in_predicate.py#L26)):*

I don't see why this constraint should be necessary?
</blockquote></details>

How does it make sense not to require that? Also, you can't translate a list containing bad values to SQL, you (understandably) end up with an error.
___
*[`pixeltable/tests/test_exprs.py` line 359 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0NUqU2fRN1toHFg4e:-Nx0cVYS6p8Eza-r89xW:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/tests/test_exprs.py#L359)):*

Also add a computed column that uses `isin` to `pixeltable/tool/create_test_db_dump.py`
</blockquote></details>

Done.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/156)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/in_predicate.py` line 16 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0LrBQ1sR7XHdQMVaE:-Nx1-B8jA0oxZ7qc-7VW:b-7bt15w) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/exprs/in_predicate.py#L16)):*

We can do Iterable[Any], but I don't see how a dict fits in.

Pandas does not accept dicts for isin():
https://pandas.pydata.org/docs/reference/api/pandas.Series.isin.html
</blockquote></details>

Pandas does accept a dict. It's considered list-like with keys as elements. For example:
```
In[14]: pd.Series(['a', 'b']).isin({'a': 5, 'c': 1})
Out[14]: 
0     True
1    False
```
```
In[15]: pd.api.types.is_list_like({'a': 5, 'c': 1})
Out[15]: True
```
This aligns with the behavior of the python `in` keyword. I'd argue Pixeltable should follow Python/Pandas and accept a dict.
___
*[`pixeltable/exprs/in_predicate.py` line 26 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/156#-Nx0Mfpb9vDS5HpUN_kJ:-Nx10DFD5BWhHsJhOHmr:b-grqr8p) ([raw file](https://github.com/pixeltable/pixeltable/blob/449fba68e5676d2cfa0daf7da09c494fa6a35266/pixeltable/exprs/in_predicate.py#L26)):*

How does it make sense not to require that? Also, you can't translate a list containing bad values to SQL, you (understandably) end up with an error.
</blockquote></details>

Pandas `isin` doesn't require it. Neither does Python `in`. I'm guessing that's the behavior most users would expect? I'd be surprised if I tried to do (say) `t.col isin [3, 'a']` and got a type error.

I agree we can't translate bad values to SQL, but we can just filter them, right? If the types don't match, then we know they can't contribute to the truth value of the expression in any case. That is, instead of this:
```
        for val in value_list:
            try:
                lhs.col_type.validate_literal(val)
            except TypeError:
                raise excs.Error(
                    f'isin(): list item {val!r} is not compatible with the type of {lhs}, which is {lhs.col_type}')
```
We could instead do this:
```
        value_list = [val for val in value_list if lhs.col_type.is_valid_literal(val)]
```
Everything that remains has the proper type, and anything that was discarded couldn't affect the expression result.

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/156)* status: 0 of 22 files reviewed, 7 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/in_predicate.py` line 18 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/156#-NxsCTS96o3x-IBBGYaY:-NxsCTSA1b7eEN21Ro0X:b-3hbf9u) ([raw file](https://github.com/pixeltable/pixeltable/blob/bf518637b8a618ea22fda960fc01e3acda08cb7e/pixeltable/exprs/in_predicate.py#L18)):*
> ```Python
>     """Predicate corresponding to the SQL IN operator."""
> 
>     def __init__(self, lhs: Expr, value_set_literal: Optional[Iterable] = None, value_set_expr: Optional[Expr] = None):
> ```

Type of `value_set_literal` is no longer correct here b/c we're accepting `dict`s (it's `Optional[Union[list, dict]]` at the moment, but see comment below) 
___
*[`pixeltable/exprs/in_predicate.py` line 24 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/156#-NxsBx7HA4l-1P4lTKcW:-NxsBx7HA4l-1P4lTKcX:b-3yeptc) ([raw file](https://github.com/pixeltable/pixeltable/blob/bf518637b8a618ea22fda960fc01e3acda08cb7e/pixeltable/exprs/in_predicate.py#L24)):*
> ```Python
>         super().__init__()
> 
>         self.value_list: Optional[list[Any]] = None  # only contains values of the correct type
> ```

As a matter of style, I think `list` is preferred to `list[Any]` (they are exact synonyms in Python, I believe)
___
*[`pixeltable/exprs/in_predicate.py` line 47 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/156#-NxsD6FfD46YEqg7xpzB:-NxsD6FfD46YEqg7xpzC:b-qi9xw8) ([raw file](https://github.com/pixeltable/pixeltable/blob/bf518637b8a618ea22fda960fc01e3acda08cb7e/pixeltable/exprs/in_predicate.py#L47)):*
> ```Python
> 
>     def _normalize_value_set(self, value_set: Any, filter_type_mismatches: bool = True) -> Iterable:
>         if not isinstance(value_set, (list, dict)):
> ```

If we want to approximate Python/Pandas, we should also accept a `set` or `ndarray` here. I think at minimum we should accept a `set` (easiest might be an `Iterable`)
___
*[`pixeltable/exprs/in_predicate.py` line 69 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/156#-NxsDchyBNWcm9Hvd7Yi:-NxsDchyBNWcm9Hvd7Yj:b14618s) ([raw file](https://github.com/pixeltable/pixeltable/blob/bf518637b8a618ea22fda960fc01e3acda08cb7e/pixeltable/exprs/in_predicate.py#L69)):*
> ```Python
> 
>     def _equals(self, other: InPredicate) -> bool:
>         return self.value_list == other.value_list
> ```

The semantics of this might not be correct in all cases, but that's probably ok depending on how `_equals` is being used? (I don't have an entirely clear intuition about it)

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/156)* status: 0 of 22 files reviewed, 7 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exprs/in_predicate.py` line 18 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/156#-NxsCTS96o3x-IBBGYaY:-NxsOyDY-eOhOFchOuX6:b-xuejmf) ([raw file](https://github.com/pixeltable/pixeltable/blob/bf518637b8a618ea22fda960fc01e3acda08cb7e/pixeltable/exprs/in_predicate.py#L18)):*
I'm getting
>>> from typing import Iterable
>>> d = {1: 'a'}
>>> isinstance(d, Iterable)
True
___
*[`pixeltable/exprs/in_predicate.py` line 24 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/156#-NxsBx7HA4l-1P4lTKcW:-NxsOB8KBo9CXKJ1i5bj:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/bf518637b8a618ea22fda960fc01e3acda08cb7e/pixeltable/exprs/in_predicate.py#L24)):*

As a matter of style, I think `list` is preferred to `list[Any]` (they are exact synonyms in Python, I believe)
</blockquote></details>

Done.
___
*[`pixeltable/exprs/in_predicate.py` line 47 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/156#-NxsD6FfD46YEqg7xpzB:-NxsPlw71OFFj02Bo6FX:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/bf518637b8a618ea22fda960fc01e3acda08cb7e/pixeltable/exprs/in_predicate.py#L47)):*

If we want to approximate Python/Pandas, we should also accept a `set` or `ndarray` here. I think at minimum we should accept a `set` (easiest might be an `Iterable`)
</blockquote></details>

Done.
___
*[`pixeltable/exprs/in_predicate.py` line 69 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/156#-NxsDchyBNWcm9Hvd7Yi:-NxsPqV7AU5wtZCSrcos:b-o0vf7g) ([raw file](https://github.com/pixeltable/pixeltable/blob/bf518637b8a618ea22fda960fc01e3acda08cb7e/pixeltable/exprs/in_predicate.py#L69)):*

The semantics of this might not be correct in all cases, but that's probably ok depending on how `_equals` is being used? (I don't have an entirely clear intuition about it)
</blockquote></details>

It's comparing two predicates, and the value_lists are getting normalized. What are cases where this doesn't work?

PR #157, opened 2024-05-06
Move tests

orm011:
Its helpful to have tests be outside of the installed tree, so you can run them as client code and can modify them in place, and see how they do, eg when you want to debug or run against the pip-release of pixeltable.

When debugging in windows, it was confusing to me which test was running, the one included with pip, or the one from the repo.  I had a few false starts there bc when I modifed the tests in place I was not seeing changes.

Additionally, currently it looks like we distribute the tests but not the data, so you basically need to clone the git repo to test anything that uses this data. 

Also remove docs from build, assuming this is not needed within the release (including potentially large notebooks , test data etc)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/157)

mkornacker (approved):

aaron-siegel (approved):

PR #159, opened 2024-05-07
Together demo with 2 issues

pierrebrunelle:
I'm struggling with two things:
- Chat Completion: I can't figure out why it doesn't generate the response and can't retrieve it.
- Image Generation: I can only render the images if I restart the runtime after a first pip install cause of PIL dependency.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/159)

pierrebrunelle:
@aaron-siegel should be ready to review/merge after the changes we discussed.

PR #160, opened 2024-05-07
Widen pillow dependency requirement

aaron-siegel:
This fixes an issue where colab complains that the kernel has to be restarted after pip-installing Pixeltable.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/160)

pyproject.toml
```diff
@@ -19,7 +19,7 @@ exclude = [
 python = ">=3.9,<4.0"
 numpy = ">=1.26"
 pandas = ">=2.0,<3.0"
-pillow = ">=10.0"
+pillow = ">=9.3.0"
```
mkornacker, line 22: We can't do that. The signature of at least one function changed between 9 and 10.

aaron-siegel: I ran the unit tests and they all passed.

We HAVE to do this or it causes unacceptable user friction in colab. Tell me which function changed and I'll look at it.

pixeltable/tests/functions/test_pil.py
```diff
@@ -0,0 +1,39 @@
...
+from ..utils import get_image_files
+
+
+class TestPil:
+
+    def to_img(self, x: Any) -> PIL.Image:
+        return x
+
+    def test_pil(self, img_tbl: Table) -> None:
+        #mask_img = next(f for f in get_image_files() if f.endswith('n03888257_1389.JPEG'))
```
mkornacker: remove

pixeltable/tests/functions/test_pil.py
```diff
@@ -0,0 +1,39 @@
...
+from pixeltable.functions.pil.image import *
+from ..utils import get_image_files
+
+
+class TestPil:
+
+    def to_img(self, x: Any) -> PIL.Image:
+        return x
+
+    def test_pil(self, img_tbl: Table) -> None:
```
mkornacker, line 9: leave TODO: we should compare the actual outputs.

mkornacker (approved):

PR #162, opened 2024-05-08
remove ColumnType::to_sql, which is not used other than once as a constant

orm011:
helpful as we try to understand the innards of the pxt


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/162)

mkornacker (approved):

PR #163, opened 2024-05-13
Use nullable=True as the default for Pixeltable types

aaron-siegel:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/163)

PR #169, opened 2024-05-15
Query functions

mkornacker:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)

orm011:
*[`tests/test_function.py` line 246 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxl4Zo298cIe2tgtR2:-Nxxl4Zo298cIe2tgtR3:b3bohyo) ([raw file](https://github.com/pixeltable/pixeltable/blob/c2e9b6b000dfed09b9d9612d1fcd343276bc80b9/tests/test_function.py#L246)):*
> ```Python
> 
>     @pxt.query
>     def lt_x(t, x: int) -> int:
> ```

I find this pattern of defining this test query function as a method within a test class confusing, what's the type of its first argument? (and what is the effect of the decorator in this context).  

Can we just move it outside the class to make it clear what this is (I'm kind of using it as documentation)

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)* status: 0 of 23 files reviewed, 1 unresolved discussion (waiting on @orm011 and @udf)

___
*[`tests/test_function.py` line 246 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxl4Zo298cIe2tgtR2:-NxxnUnxEMrwAHVUIdHL:b5qexg3) ([raw file](https://github.com/pixeltable/pixeltable/blob/c2e9b6b000dfed09b9d9612d1fcd343276bc80b9/tests/test_function.py#L246)):*

I find this pattern of defining this test query function as a method within a test class confusing, what's the type of its first argument? (and what is the effect of the decorator in this context).  

Can we just move it outside the class to make it clear what this is (I'm kind of using it as documentation)
</blockquote></details>

We can do that, but I'm not quite following: how will that make anything clearer? It's going to be the same symbol, but you'll use it without the 'self.'

The type of the first argument is Table. Happy to add that.

The effect of the decorator is to turn it into an object that can then be used with Table.query(), analogously to @udf.

orm011:
*[`tests/test_function.py` line 246 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxl4Zo298cIe2tgtR2:-NxxoohU0GJROyK5XywZ:b-eudcit) ([raw file](https://github.com/pixeltable/pixeltable/blob/c2e9b6b000dfed09b9d9612d1fcd343276bc80b9/tests/test_function.py#L246)):*

We can do that, but I'm not quite following: how will that make anything clearer? It's going to be the same symbol, but you'll use it without the 'self.'

The type of the first argument is Table. Happy to add that.

The effect of the decorator is to turn it into an object that can then be used with Table.query(), analogously to @udf.
</blockquote></details>

What I mean is: if you didn't have that decorator, self.lt_x(1) would bind an instance of  TestFunction to `t` and thats what t would be. The decorator somehow makes this work with t being a table, but its really not obvious to me what's going on here.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)* status: 0 of 23 files reviewed, 7 unresolved discussions (waiting on @mkornacker, @orm011, and @udf)

___
*[`pixeltable/dataframe.py` line 346 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-NxxudOi45BNYJXV_ZdD:-NxxudOi45BNYJXV_ZdE:bpm6og9) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/dataframe.py#L346)):*
> ```Python
> 
>     def _vars(self) -> dict[str, exprs.Variable]:
>         """Return a dict mapping parameter name to Variable"""
> ```

Something like this: would be clearer (if I correctly understand what it's doing): "Returns a dict mapping parameter names to Variables, ranging over all Variables appearing in expressions that define this DataFrame."
___
*[`pixeltable/dataframe.py` line 366 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-NxxvJA24Fm66lnYbaMq:-NxxvJA24Fm66lnYbaMr:benpsw9) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/dataframe.py#L366)):*
> ```Python
> 
>     def parameters(self) -> dict[str, ColumnType]:
>         """Return a dict mapping parameter name to parameter type"""
> ```

Same comment as above.
___
*[`pixeltable/dataframe.py` line 459 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxw-ayCCH2mpbHPqj9:-Nxxw-ayCCH2mpbHPqjA:benn5r8) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/dataframe.py#L459)):*
> ```Python
>                 exprs.Expr.list_substitute(order_by_exprs, var_expr, arg_expr)
>         select_list = list(zip(select_list_exprs, self._column_names))
>         order_by_clause: Optional[List[Tuple[exprs.Expr, bool]]] = None
> ```

use lowercase list/tuple
___
*[`pixeltable/func/signature.py` line 22 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-NxxxfUI7dJg8XzGvlkR:-NxxxfUI7dJg8XzGvlkS:bkojdjp) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/func/signature.py#L22)):*
> ```Python
>     col_type: Optional[ts.ColumnType]  # None for variable parameters
>     kind: enum.Enum  # inspect.Parameter.kind; inspect._ParameterKind is private
>     # for some reason, this needs to precede is_batched in the dataclass definition
> ```

What happens if it doesn't?
___
*[`pixeltable/func/signature.py` line 104 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxzdj0BeVI4vxiFTTQ:-Nxxzdj0BeVI4vxiFTTR:ba3u1ds) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/func/signature.py#L104)):*
> ```Python
>         result = {
>             'return_type': self.get_return_type().as_dict(),
>             'parameters': [p.as_dict() for p in self.parameters.values()],
> ```

These are some pretty substantial schema changes; we'll need to bump the schema version, right?
___
*[`pixeltable/func/signature.py` line 160 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxy7GF4eJXZ7bk8L8Y:-Nxxy7GF4eJXZ7bk8L8Z:b814zhd) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/func/signature.py#L160)):*
> ```Python
>             param_types: Optional[List[ts.ColumnType]] = None
>     ) -> List[Parameter]:
>         assert (py_fn is None) != (py_params is None)
> ```

I see this pattern being used a lot, and I don't love it. Isn't it more typical in Python to have a single (non-Optional) parameter of type

`Union[Callable, list[inspect.Parameter]]`

? To me that makes the intended semantics much clearer.
___
*[`tests/test_function.py` line 246 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxl4Zo298cIe2tgtR2:-Nxy-RbL18k20wo3EbVn:b-f94095) ([raw file](https://github.com/pixeltable/pixeltable/blob/c2e9b6b000dfed09b9d9612d1fcd343276bc80b9/tests/test_function.py#L246)):*

What I mean is: if you didn't have that decorator, self.lt_x(1) would bind an instance of  TestFunction to `t` and thats what t would be. The decorator somehow makes this work with t being a table, but its really not obvious to me what's going on here.
</blockquote></details>

We should definitely add the type hint. I kinda need to wrap my head around this too though.

I think of `@udf` as an operator that "lifts" a Python function, which acts on values, to a Pixeltable Function, which acts on columns of values.

Is it correct to think of `@query` as doing the same thing but with an additional `Table` parameter: the Python function acts on (Table, values) and the lifted Function acts on (Table, columns)?

udf:
Hi, I was accidentally mentioned here. Not sure how to opt out of the review on reviewable, or if i even have to.

aaron-siegel (commented):
We have a Python decorator `@udf` that matches your username; it seems that github tags you every time we mention it. I'm not sure how to disable this behavior.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169#-:-Ny1BRKk6mtMwuscWK39:b-sii5cd)* status: 0 of 27 files reviewed, 7 unresolved discussions (waiting on @mkornacker, @orm011, and @udf)

udf:
Don't worry, I'm used to it. Putting it in monospace makes it not mention me, but if I'm tagged once then reviewable will see me as a mentionee and keep tagging me.

udf (commented):
> Reviewed 15 of 23 files at r1, 5 of 5 files at r2, 5 of 7 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)* status: all files reviewed, 7 unresolved discussions (waiting on @mkornacker and @orm011)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)* status: 7 of 31 files reviewed, 15 unresolved discussions (waiting on @mkornacker, @orm011, and @udf)

___
*[`pixeltable/dataframe.py` line 378 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-CYEaQDFqlc2IrcnY9:-O-CYEaQDFqlc2IrcnYA:bp9p74e) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/dataframe.py#L378)):*
> ```Python
>         return {name: var.col_type for name, var in vars.items()}
> 
>     def _exec(self, conn: Optional[sql.engine.Connection] = None) -> Generator[exprs.DataRow, None, None]:
> ```

I think `Iterator[DataRow]` the preferred synonym for `Generator[DataRow, None, None]`?
___
*[`pixeltable/dataframe.py` line 407 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-CYfEL5YrxADsmbid3:-O-CYfEL5YrxADsmbid4:b5337ct) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/dataframe.py#L407)):*
> ```Python
>         def exec_plan(conn: sql.engine.Connection) -> Generator[exprs.DataRow, None, None]:
>             plan.ctx.set_conn(conn)
>             plan.open()
> ```

Better to use a context manager pattern here e.g. something like:
```
with plan.open(conn):
    for row_batch in plan:
        # etc
```
It's generally preferred to a try/finally pattern with explicit close.

Not necessary to make the change for this review, but I think we should note this as a future refactoring of the planner code
___
*[`pixeltable/dataframe.py` line 735 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-CZzhMAgHvTMKXQhTa:-O-CZzhMAgHvTMKXQhTb:biwm2r2) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/dataframe.py#L735)):*
> ```Python
>         d = {
>             '_classname': 'DataFrame',
>             'tbl': self.tbl.as_dict(),
> ```

Does this entail a schema version change?
___
*[`pixeltable/catalog/table.py` line 56 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-C_S_C455BoeO2-BA5:-O-C_S_C455BoeO2-BA6:bu7psah) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L56)):*
> ```Python
> 
>     @property
>     def _tbl_version(self) -> TableVersion:
> ```

I wasn't aware this existed until now!

But why is it `self.tbl_version_path` but `self._tbl_version`? Shouldn't they both (or neither) have underscores for consistency? That's probably why I never discovered it.
___
*[`pixeltable/catalog/table.py` line 61 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-C_rKjC5wkOxIjWM6c:-O-C_rKjC5wkOxIjWM6d:bf4z99a) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L61)):*
> ```Python
> 
>     def __hash__(self) -> int:
>         return hash(self._tbl_version().id)
> ```

Yikes! `_tbl_version` is now a property so this needs to be changed. This suggests this code path isn't being tested.
___
*[`pixeltable/catalog/table.py` line 76 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-Ca4XRDSNDjxfsxQUO:-O-Ca4XRDSNDjxfsxQUP:b-h0sf93) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L76)):*
> ```Python
>         return getattr(self.tbl_version_path, name)
> 
>     def __getitem__(self, index: object) -> Union['pixeltable.exprs.ColumnRef', 'pixeltable.dataframe.DataFrame']:
> ```

`__getitem__` needs to have `pixeltable.func.QueryTemplateFunction` as a type in its union too, correct? Or is this access pattern not supported for query functions?
___
*[`pixeltable/catalog/table.py` line 734 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-CaJVPAiFveaUQWp6v:-O-CaJVPAiFveaUQWp6w:bivb876) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L734)):*
> ```Python
> 
>     @overload
>     def query(self, *, param_types: Optional[List[ts.ColumnType]] = None) -> Callable: ...
> ```

Use a qualified return type

`Callable[[Callable], QueryTemplateFunction]`

so that the typechecker will know what the decorator spits out.
___
*[`pixeltable/catalog/table.py` line 764 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-CbXwhEeouLb6PQNLp:-O-CbXwhEeouLb6PQNLq:bt1clhb) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L764)):*
> ```Python
>         else:
>             assert len(args) == 0 and len(kwargs) == 1 and 'param_types' in kwargs
>             return lambda py_fn: decorator(py_fn, kwargs['param_types'])
> ```

This is kind of a weird pattern and might confuse the typechecker (and me).

It'd be clearer to use the name `make_query_template` (or some such) for the thing we're now calling `decorator`, and change this to:

```
        if len(args) == 1:
            assert len(kwargs) == 0 and callable(args[0])
            return make_query_template(args[0], None)
        else:
            assert len(args) == 0 and len(kwargs) == 1 and 'param_types' in kwargs
            def decorator(py_fn: Callable) -> 'pixeltable.func.QueryTemplateFunction':
                return make_query_template(py_fn, kwargs['param_types'])
            return decorator
```
It's slightly more verbose, but (1) the decorator now has proper type hints and (2) the thing we're calling `decorator` actually is a decorator.

Same comment for the code in `expr_udf` (which I guess I overlooked when I did that code review)
___
*[`pixeltable/exprs/column_ref.py` line 116 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-Cd00l8rXW1AQqRuWQ:-O-Cd00l8rXW1AQqRuWR:b2gxh4a) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/exprs/column_ref.py#L116)):*
> ```Python
> 
>     @classmethod
>     def _from_dict(cls, d: Dict, _: List[Expr]) -> Expr:
> ```

use lowercase dict/list

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)* status: 5 of 33 files reviewed, 15 unresolved discussions (waiting on @aaron-siegel, @orm011, and @udf)

___
*[`pixeltable/dataframe.py` line 346 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-NxxudOi45BNYJXV_ZdD:-Nxy4eTp1IA3h3evOuhj:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/dataframe.py#L346)):*

Something like this: would be clearer (if I correctly understand what it's doing): "Returns a dict mapping parameter names to Variables, ranging over all Variables appearing in expressions that define this DataFrame."
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 366 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-NxxvJA24Fm66lnYbaMq:-Nxy4usBEdMyYLoYP_A-:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/dataframe.py#L366)):*

Same comment as above.
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 459 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxw-ayCCH2mpbHPqj9:-Nxy45ewCeDYAu0xFfsj:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/dataframe.py#L459)):*

use lowercase list/tuple
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 378 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-CYEaQDFqlc2IrcnY9:-O-Xzraz54t8dpJy39Uz:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/dataframe.py#L378)):*

I think `Iterator[DataRow]` the preferred synonym for `Generator[DataRow, None, None]`?
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 735 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-CZzhMAgHvTMKXQhTa:-O-Y-0_yEa6jx4W556AY:b2nhswy) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/dataframe.py#L735)):*

Does this entail a schema version change?
</blockquote></details>

Yes
___
*[`pixeltable/catalog/table.py` line 56 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-C_S_C455BoeO2-BA5:-O-Y-aSl5mlhGdIsddzJ:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L56)):*

I wasn't aware this existed until now!

But why is it `self.tbl_version_path` but `self._tbl_version`? Shouldn't they both (or neither) have underscores for consistency? That's probably why I never discovered it.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 61 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-C_rKjC5wkOxIjWM6c:-O-Y-d7d5nuRzG62NoxT:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L61)):*

Yikes! `_tbl_version` is now a property so this needs to be changed. This suggests this code path isn't being tested.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 76 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-Ca4XRDSNDjxfsxQUO:-O-Y01VVEX6erX7Ccn9r:bfyb44s) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L76)):*

`__getitem__` needs to have `pixeltable.func.QueryTemplateFunction` as a type in its union too, correct? Or is this access pattern not supported for query functions?
</blockquote></details>

That's in getattr() (I hate that it collapses lines w/o changes; if it didn't, you'd have seen getattr() directly above it).
___
*[`pixeltable/catalog/table.py` line 764 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-CbXwhEeouLb6PQNLp:-O-Y1CaD5pjNScdgAP7A:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L764)):*

This is kind of a weird pattern and might confuse the typechecker (and me).

It'd be clearer to use the name `make_query_template` (or some such) for the thing we're now calling `decorator`, and change this to:

```
        if len(args) == 1:
            assert len(kwargs) == 0 and callable(args[0])
            return make_query_template(args[0], None)
        else:
            assert len(args) == 0 and len(kwargs) == 1 and 'param_types' in kwargs
            def decorator(py_fn: Callable) -> 'pixeltable.func.QueryTemplateFunction':
                return make_query_template(py_fn, kwargs['param_types'])
            return decorator
```
It's slightly more verbose, but (1) the decorator now has proper type hints and (2) the thing we're calling `decorator` actually is a decorator.

Same comment for the code in `expr_udf` (which I guess I overlooked when I did that code review)
</blockquote></details>

Done.
___
*[`pixeltable/exprs/column_ref.py` line 116 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-Cd00l8rXW1AQqRuWQ:-O-Y1UuX4JEYJ5vkYzLn:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/exprs/column_ref.py#L116)):*

use lowercase dict/list
</blockquote></details>

Done.
___
*[`pixeltable/func/signature.py` line 22 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-NxxxfUI7dJg8XzGvlkR:-Nxy5KDJE5CqIti2AxgU:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/func/signature.py#L22)):*

What happens if it doesn't?
</blockquote></details>

Done.
___
*[`pixeltable/func/signature.py` line 104 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxzdj0BeVI4vxiFTTQ:-Nxy5U6b32MLqQYvTe_7:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/func/signature.py#L104)):*

These are some pretty substantial schema changes; we'll need to bump the schema version, right?
</blockquote></details>

Done.
___
*[`pixeltable/func/signature.py` line 160 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxy7GF4eJXZ7bk8L8Y:-Nxy5ayA418KGpwfC7Qe:b-va4p2q) ([raw file](https://github.com/pixeltable/pixeltable/blob/6336b95a75f0e2552d63d39f3d6266f9ded34854/pixeltable/func/signature.py#L160)):*

I see this pattern being used a lot, and I don't love it. Isn't it more typical in Python to have a single (non-Optional) parameter of type

`Union[Callable, list[inspect.Parameter]]`

? To me that makes the intended semantics much clearer.
</blockquote></details>

I absolutely *hate* unions, especially in this case, where they really are two completely different things (which to me indicates that they should not be called the same thing). Also, doing that will/would make typechecking useless.
___
*[`tests/test_function.py` line 246 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxl4Zo298cIe2tgtR2:-Nxy7Y8q6m6I4N5N1jZJ:bkgu049) ([raw file](https://github.com/pixeltable/pixeltable/blob/c2e9b6b000dfed09b9d9612d1fcd343276bc80b9/tests/test_function.py#L246)):*

We should definitely add the type hint. I kinda need to wrap my head around this too though.

I think of `@udf` as an operator that "lifts" a Python function, which acts on values, to a Pixeltable Function, which acts on columns of values.

Is it correct to think of `@query` as doing the same thing but with an additional `Table` parameter: the Python function acts on (Table, values) and the lifted Function acts on (Table, columns)?
</blockquote></details>

Let's discuss verbally.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)* status: 4 of 35 files reviewed, 9 unresolved discussions (waiting on @mkornacker, @orm011, and @udf)

___
*[`pixeltable/dataframe.py` line 453 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-bP1s_1kBoxuhY9Vim:-O-bP1saEtCQbK0kH_tl:bg3yiku) ([raw file](https://github.com/pixeltable/pixeltable/blob/237cdeabecd3c59f4cefa66d8ac914c7f1816c3a/pixeltable/dataframe.py#L453)):*
> ```Python
>         where_clause = copy.deepcopy(self.where_clause)
>         group_by_clause = copy.deepcopy(self.group_by_clause)
>         order_by_exprs = copy.deepcopy(
> ```

Two random stylistic comments:

```[copy.deepcopy(order_by_expr) for order_by_expr, _ in self.order_by_clause]```

is a little clearer to me, with the advantage that we're not creating a list just to copy it. And,

```<expression> if self.order_by_clause is not None else None```

can be shortened to

```<expression> if self.order_by_clause else None```

(I'm assuming Expr instances are always truthy.) I've started preferring the latter, though I'm not aware of an established convention for this.
___
*[`pixeltable/catalog/table.py` line 76 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-Ca4XRDSNDjxfsxQUO:-O-bQeDtBbeI3rBD7ynd:bjiw1qt) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L76)):*

That's in getattr() (I hate that it collapses lines w/o changes; if it didn't, you'd have seen getattr() directly above it).
</blockquote></details>

I understand, but doesn't `__getitem__` need it too? I'm assuming that `t['func_name']` can resolve a query function just like `t.func_name`; if not, I think it should (otherwise the lack of parity between them is a little confusing)
___
*[`pixeltable/catalog/table.py` line 764 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-CbXwhEeouLb6PQNLp:-O-bRiC41T0WOO4uZI2o:b-lb4oqp) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L764)):*

Done.
</blockquote></details>

lgtm. And ah, I guess the lambda is ok because the type hint is on the overloaded method, so we don't need another type hint here.
___
*[`pixeltable/catalog/table.py` line 70 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-bR48RAHUAo_5gfMvn:-O-bR48RAHUAo_5gfMvo:b-seg0vj) ([raw file](https://github.com/pixeltable/pixeltable/blob/237cdeabecd3c59f4cefa66d8ac914c7f1816c3a/pixeltable/catalog/table.py#L70)):*
> ```Python
>             self, name: str
>     ) -> Union['pixeltable.exprs.ColumnRef', 'pixeltable.func.QueryTemplateFunction']:
>         """Return a ColumnRef for the given column name.
> ```

docstring should be updated since this can return more than a `ColumnRef` now
___
*[`pixeltable/metadata/converters/convert_15.py` line 16 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-bUDXM5yHJVGbHRkzD:-O-bUDXM5yHJVGbHRkzE:b6q9zbj) ([raw file](https://github.com/pixeltable/pixeltable/blob/237cdeabecd3c59f4cefa66d8ac914c7f1816c3a/pixeltable/metadata/converters/convert_15.py#L16)):*
> ```Python
> 
> 
> def convert_15(engine: sql.engine.Engine) -> None:
> ```

Also add some `@query` instances to `create_test_db_dump`. (`test_migration` also tests loading the v15 dump from scratch, in addition to migrating older dumps, so we get validation of the `@query` metadata write/read loop)
___
*[`pixeltable/metadata/converters/convert_15.py` line 34 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-c9N8u2CYCmRqdwBWb:-O-c9N8u2CYCmRqdwBWc:bmlmjs6) ([raw file](https://github.com/pixeltable/pixeltable/blob/237cdeabecd3c59f4cefa66d8ac914c7f1816c3a/pixeltable/metadata/converters/convert_15.py#L34)):*
> ```Python
>         col_type = ts.ColumnType.from_dict(col_type_dict) if col_type_dict is not None else None
>         default = py_params[name].default
>         kind = inspect._ParameterKind(kind_int)  # is there a way to avoid referecing a private type?
> ```

there doesn't appear to be, other than maybe?
```
KINDS = [POSITIONAL_ONLY, POSITIONAL_OR_KEYWORD, VAR_POSITIONAL, KEYWORD_ONLY, VAR_KEYWORD]

# later:
kind = KINDS[kind_int]
```

I don't think construction of `inspect.Parameter`s is meant to be a public facing operation; the intent is they only show up as the output of `inspect.signature` and so on. It might be risky to regard the `_ParameterKind` enum as anything other than an implementation detail of the `inspect.Parameter.POSITIONAL_ONLY` (etc) constants.

(I do see that this is only being used in the converter & scrupulously avoided elsewhere)
___
*[`tests/test_function.py` line 246 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/169#-Nxxl4Zo298cIe2tgtR2:-O-bTi8VAv9yoXb9-0_5:b-qgjkd8) ([raw file](https://github.com/pixeltable/pixeltable/blob/c2e9b6b000dfed09b9d9612d1fcd343276bc80b9/tests/test_function.py#L246)):*

Let's discuss verbally.
</blockquote></details>

My previous comment is obsolete; the `@query` pattern has changed a lot since then, and I'm happy with the outcome.

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)* status: 4 of 35 files reviewed, 9 unresolved discussions (waiting on @aaron-siegel, @orm011, and @udf)

___
*[`pixeltable/dataframe.py` line 453 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-bP1s_1kBoxuhY9Vim:-O-cU3uVAd9qO8FbdZo6:b-rt2ig1) ([raw file](https://github.com/pixeltable/pixeltable/blob/237cdeabecd3c59f4cefa66d8ac914c7f1816c3a/pixeltable/dataframe.py#L453)):*

Two random stylistic comments:

```[copy.deepcopy(order_by_expr) for order_by_expr, _ in self.order_by_clause]```

is a little clearer to me, with the advantage that we're not creating a list just to copy it. And,

```<expression> if self.order_by_clause is not None else None```

can be shortened to

```<expression> if self.order_by_clause else None```

(I'm assuming Expr instances are always truthy.) I've started preferring the latter, though I'm not aware of an established convention for this.
</blockquote></details>

Done.

I left the explicit comparison to None in place, though - the Google style guide wants that.
___
*[`pixeltable/catalog/table.py` line 76 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-Ca4XRDSNDjxfsxQUO:-O-cUU4hDM59kTpNhODA:b-abpgrs) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L76)):*

I understand, but doesn't `__getitem__` need it too? I'm assuming that `t['func_name']` can resolve a query function just like `t.func_name`; if not, I think it should (otherwise the lack of parity between them is a little confusing)
</blockquote></details>

Can we leave this out of this? It's unclear whether we want to keep getitem for anything other than column references, because you can have non-id column names. But that doesn't apply to queries.
___
*[`pixeltable/catalog/table.py` line 70 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-bR48RAHUAo_5gfMvn:-O-cVuFh38MJQS4cvY5P:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/237cdeabecd3c59f4cefa66d8ac914c7f1816c3a/pixeltable/catalog/table.py#L70)):*

docstring should be updated since this can return more than a `ColumnRef` now
</blockquote></details>

Done.
___
*[`pixeltable/metadata/converters/convert_15.py` line 16 at r8](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-bUDXM5yHJVGbHRkzD:-O-cV_bm2HcZcGYJ_2-3:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/237cdeabecd3c59f4cefa66d8ac914c7f1816c3a/pixeltable/metadata/converters/convert_15.py#L16)):*

Also add some `@query` instances to `create_test_db_dump`. (`test_migration` also tests loading the v15 dump from scratch, in addition to migrating older dumps, so we get validation of the `@query` metadata write/read loop)
</blockquote></details>

Done.

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)* status: 4 of 35 files reviewed, 9 unresolved discussions (waiting on @mkornacker, @orm011, and @udf)

___
*[`pixeltable/catalog/table.py` line 76 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-Ca4XRDSNDjxfsxQUO:-O-dr4exEc9LYDRS1x-p:bjm4wc) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L76)):*

Can we leave this out of this? It's unclear whether we want to keep getitem for anything other than column references, because you can have non-id column names. But that doesn't apply to queries.
</blockquote></details>

Sure, we don't have to resolve it in this PR. But I tend to think it'll confuse ppl if `t.my_symbol` and `t['my_symbol']` behave similarly, but not the same. The vast majority of users' exposure to these patterns will involve dereferencing columns, and as a result they'll expect them to be synonyms, and likely be surprised to find that's only the case for columns and not queries. (I'd be surprised.)
___
*[`pixeltable/tool/create_test_db_dump.py` line 252 at r9](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-dsJV53MW1q19NRN7H:-O-dsJV53MW1q19NRN7I:bi32sj9) ([raw file](https://github.com/pixeltable/pixeltable/blob/6dee7971cf3e7259e59e307ea69bc33457f1ab56/pixeltable/tool/create_test_db_dump.py#L252)):*
> ```Python
>             #return t.where(t.c2 < i)
>             return t.where(t.c2 < i).select(t.c1, t.c2)
>         add_column('query_output', t.q1(t.c2))
> ```

it'd be good to have an instance that uses `similarity` too, since `SimilarityExpr` can't (currently) be assigned directly to a column, and therefore isn't yet covered by the db dumps (that is, I believe the only way to get a `SimilarityExpr` into the dict is through a `@query`, but pls correct me if that's wrong).

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/169)* status: 4 of 35 files reviewed, 9 unresolved discussions (waiting on @aaron-siegel, @orm011, and @udf)

___
*[`pixeltable/catalog/table.py` line 76 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-Ca4XRDSNDjxfsxQUO:-O-fZ8dA8rBaljBBqWh1:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/032c090086ea017f8a9799bfc22e3a40668e38c5/pixeltable/catalog/table.py#L76)):*

Sure, we don't have to resolve it in this PR. But I tend to think it'll confuse ppl if `t.my_symbol` and `t['my_symbol']` behave similarly, but not the same. The vast majority of users' exposure to these patterns will involve dereferencing columns, and as a result they'll expect them to be synonyms, and likely be surprised to find that's only the case for columns and not queries. (I'd be surprised.)
</blockquote></details>

Done.
___
*[`pixeltable/tool/create_test_db_dump.py` line 252 at r9](https://reviewable.io/reviews/pixeltable/pixeltable/169#-O-dsJV53MW1q19NRN7H:-O-fYRK9AOp3O9uOtayg:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/6dee7971cf3e7259e59e307ea69bc33457f1ab56/pixeltable/tool/create_test_db_dump.py#L252)):*

it'd be good to have an instance that uses `similarity` too, since `SimilarityExpr` can't (currently) be assigned directly to a column, and therefore isn't yet covered by the db dumps (that is, I believe the only way to get a `SimilarityExpr` into the dict is through a `@query`, but pls correct me if that's wrong).
</blockquote></details>

Done.

PR #170, opened 2024-05-16
New pattern for specifying iterators in views

aaron-siegel:
Following Marcel's suggestion, this basically does the simplest thing that gets the new pattern established. We can sort out getting dot/tab-completion working for iterator args as a separate problem.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/170)

mkornacker (approved):
Nice!

PR #177, opened 2024-05-17
more concise array and string formatting

orm011:
make array columns print only a few elements by using numpys array formatting functions
also: make string columns print an abridged string if the original string is too long

rendered rag notebook here:
https://github.com/orm011/pixeltable/blob/concise_format/docs/release/tutorials/rag-operations.ipynb

compare with:
https://github.com/pixeltable/pixeltable/blob/master/docs/release/tutorials/rag-operations.ipynb


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/177)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/177)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @orm011)

___
*[`pixeltable/dataframe.py` line 55 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/177#-Ny7d7B8677cainBy32t:-Ny7d7B8677cainBy32u:b-6lp8my) ([raw file](https://github.com/pixeltable/pixeltable/blob/ff9cc8733d05d9f78b5451b35f42eadd6cb20837/pixeltable/dataframe.py#L55)):*
> ```Python
>             ts.AudioType: self._format_audio,
>             ts.DocumentType: self._format_document,
>             ts.ArrayType: self._format_array,
> ```

We also need a `_format_dict` and `_format_list`, and to process them recursively. Arrays/lists inside of dicts are fairly common.
___
*[`pixeltable/dataframe.py` line 148 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/177#-Ny7kObB5mwYfl6nioU8:-Ny7kObB5mwYfl6nioU9:bsr7v4s) ([raw file](https://github.com/pixeltable/pixeltable/blob/ff9cc8733d05d9f78b5451b35f42eadd6cb20837/pixeltable/dataframe.py#L148)):*
> ```Python
>     def _format_array(self, arr: List[Any]) -> str:
>         arr = np.array(arr)
>         return np.array2string(arr, precision=3, threshold=8, separator=',', edgeitems=3)
> ```

I'd argue for increasing threshold and edgeitems a bit, maybe to 16/6? We could think about making this a configurable parameter in the pixeltable env (but obviously that's not urgent)
___
*[`pixeltable/dataframe.py` line 152 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/177#-Ny7kZpJ1-zazTaiVGIx:-Ny7kZpJ1-zazTaiVGIy:b-62hslw) ([raw file](https://github.com/pixeltable/pixeltable/blob/ff9cc8733d05d9f78b5451b35f42eadd6cb20837/pixeltable/dataframe.py#L152)):*
> ```Python
>     def _format_string(self, string: str) -> str:
>         if len(string) > 250:
>             return f'{string[:120]}...{string[-120:]}'
> ```

Suggestion: use `......` as the separator instead of `...`

The single ellipsis is sufficient for arrays but could get buried in the string.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/177)* status: 0 of 7 files reviewed, 22 unresolved discussions (waiting on @orm011)

___
*[`docs/release/tutorials/rag-operations.ipynb` line 1442 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYgXAJD7zoVOXelkPM:-NyYgXAJD7zoVOXelkPN:b-j4jj2n) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/docs/release/tutorials/rag-operations.ipynb#L1442)):*
> ```
>        "      <td><pre style=\"text-align:left; background-color:transparent; margin:0;\">{1: Marc Chagall}</pre></td>\n",
>        "      <td>1015</td>\n",
>        "      <td><pre style=\"text-align:left; background-color:transparent; margin:0;\">[ 6.518e-05, 3.302e-01,-3.144e-01, 1.680e-01,-1.229e-01, 3.993e-01,...,\n",
> ```

I don't love that scientific notation is being used for values like 0.330 (just picking an example). It's more compact to write it without scientific notation! Do we have control over this (i.e. only use scientific notation when it's actually more compact than a decimal of equal precision?)
___
*[`docs/release/tutorials/rag-operations.ipynb` line 1589 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYh-Sg84PBXRe7JSni:-NyYh-Sg84PBXRe7JSnj:b-ltna8y) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/docs/release/tutorials/rag-operations.ipynb#L1589)):*
> ```
>        "\n",
>        "                                        minilm_embed  \\\n",
>        "0  [-0.26239732, -0.11875591, -0.1327097, 0.04825...   \n",
> ```

Why are we still seeing high-precision numbers in the notebooks?
___
*[`pixeltable/utils/formatter.py` line 49 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYi4X-62VpHjMxW9vd:-NyYi4X-62VpHjMxW9ve:b-b1o1ez) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L49)):*
> ```Python
> 
> 
> from typing import Union
> ```

This should go at the top of the file
___
*[`pixeltable/utils/formatter.py` line 52 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYiaMVBrAvmdFpLpVu:-NyYiaMVBrAvmdFpLpVv:bytol7s) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L52)):*
> ```Python
> 
> 
> def contains_only_simple_elements(values: Union[list[Any], dict[Any]]) -> bool:
> ```

dict has two type parameters, right?
___
*[`pixeltable/utils/formatter.py` line 55 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYikGUBea0iKJJbK3v:-NyYikGUBea0iKJJbK3w:b-m3c80r) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L55)):*
> ```Python
>     """whether all values of this array or dict are themselves elementary (no nested lists or dicts)"""
>     if isinstance(values, dict):
>         values = values.values()
> ```

This is really awkward. It'd be a clearer separation of logic to take an `Iterable` as the parameter rather than `Union[list, dict]`, and then when using it on a dict, call `contains_only_simple_elements(my_dict.values())`.

Also, I'd argue it'd be clearer to negate the function and call it `contains_collection` or some such; that also has the benefit of avoiding introducing the bespoke term `simple` (not a huge deal, but it was an extra/avoidable cognitive step for me to figure out what that meant)
___
*[`pixeltable/utils/formatter.py` line 60 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYnGED5wbMiec3_kul:-NyYnGED5wbMiec3_kum:bmawf9y) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L60)):*
> ```Python
> 
> 
> class PixeltableFormatter:
> ```

We definitely need tests for this. There's a lot of new logic in here.
___
*[`pixeltable/utils/formatter.py` line 70 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYjoXz0dEoaYSfix6j:-NyYjoXz0dEoaYSfix6k:bi52cjv) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L70)):*
> ```Python
>         return np.array2string(np.array(val), precision=FLOAT_PRECISION)
> 
>     def _format_array(self, arr: list[Any]) -> str:
> ```

why's this taking a list as input?
___
*[`pixeltable/utils/formatter.py` line 71 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYk6e27Y7q3WJD3cKt:-NyYk6e27Y7q3WJD3cKu:b-cn6mvw) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L71)):*
> ```Python
> 
>     def _format_array(self, arr: list[Any]) -> str:
>         """for numerical arrays only. to format an array type, use format_json, which also includes styling"""
> ```

I'd like to get away from docstrings that are really just notes (vs. proper documentation).

It's not necessary to have docstrings for private methods, so the simplest thing to do would be convert this to a code comment. If we do want to include the docstrings, let's write them properly (clear description of what the function does, complete sentences, correct punctuation, etc.)
___
*[`pixeltable/utils/formatter.py` line 88 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYk03U6NqJu8N7F4nW:-NyYk03U6NqJu8N7F4nX:bmqet18) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L88)):*
> ```Python
>             If the string is too long, show only the first and last STRING_EDGEITEMS characters.
> 
>             TODO: render enclosing quotes also? but need to escape inner quotes
> ```

These TODOs should definitely be in comments, not the docstring
___
*[`pixeltable/utils/formatter.py` line 96 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYkv1p6iYgQOg7R2CN:-NyYkv1p6iYgQOg7R2CO:bjcd0a3) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L96)):*
> ```Python
> 
>     def _format_json_helper(self, obj: Any, level: int) -> str:
>         def get_prefix(level):
> ```

why's this a nested function? since it depends only on level, can't it just be a variable? (`prefix = ' ' * (level * INDENT)`)
___
*[`pixeltable/utils/formatter.py` line 103 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYlJzN0CwachW0bJq8:-NyYlJzN0CwachW0bJq9:b7tqh8s) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L103)):*
> ```Python
> 
>         if isinstance(obj, np.ndarray):  # sometimes we get a numpy array, which is technically not a list
>             return self._format_json_helper(obj.tolist(), level=level)
> ```

better to say `obj = obj.tolist()` than to do a recursive call
___
*[`pixeltable/utils/formatter.py` line 112 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYljViCLbaym-ceNvP:-NyYljViCLbaym-ceNvQ:btsoqub) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L112)):*
> ```Python
>             # 3. list with at least one nested list or dict
>             #   => insert newlines between elements to help read them.
>             arr = as_numeric_ndarray(obj)
> ```

Actually (cf. my previous comment), this seems weird. we're converting to a list, then back to an ndarray again? Can't this be reorganized so as to avoid that?
___
*[`pixeltable/utils/formatter.py` line 156 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYnRuJEQKoTfBDC3Ec:-NyYnRuJEQKoTfBDC3Ed:bvydwc5) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L156)):*
> ```Python
>             return f'{get_prefix(level)}{self.format_float(obj)}'
>         elif isinstance(obj, str):
>             return f'{get_prefix(level)}{self._format_string(obj)}'
> ```

Why is `_format_string` a protected method but not `format_float`?

Are strings going to be properly quoted when they appear inside JSON?
___
*[`pixeltable/utils/formatter.py` line 160 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYneoKEjFWcHQc_9Ep:-NyYneoKEjFWcHQc_9Eq:bqz5y1p) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L160)):*
> ```Python
>             return f'{get_prefix(level)}{obj}'
>         else:
>             raise AssertionError(f'Unexpected type within json: {type(obj)}')
> ```

Pretty sure `AssertionError` should never be referenced explicitly (use `assert` keyword for assertions).
___
*[`pixeltable/utils/formatter.py` line 164 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYpMnx6jpsR_qbdF6w:-NyYpMnx6jpsR_qbdF6x:beu3kpc) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L164)):*
> ```Python
>     def format_json(self, obj: Any) -> str:
>         """ Formats a json object for display in a notebook cell so that it is easier to read, and
>         so that sub-objects are consistent with the top-level elementary types like int, float and arrays.
> ```

Let's try to avoid vague language in docstrings ("elementary types like int, float, and arrays"). Are those the only three types involved? If so, remove the word "like". If not, list out the types explicitly (and clarify the meaning of "arrays", e.g., something like "the elementary types int, float, bool, and np.ndarray".
___
*[`pixeltable/utils/formatter.py` line 166 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYoSDW-C-HbHffPgvR:-NyYoSDW-C-HbHffPgvS:b2pnk0b) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L166)):*
> ```Python
>         so that sub-objects are consistent with the top-level elementary types like int, float and arrays.
> 
>         Implementation notes:
> ```

I don't think this will render properly as a docstring? I suggest using markdown bullets
___
*[`pixeltable/utils/formatter.py` line 168 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYoGui4U0w-jyvakuW:-NyYoGui4U0w-jyvakuX:bq3bsl7) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L168)):*
> ```Python
>         Implementation notes:
>             Uses <pre> tag so that :
>                 1) json is displayed in a monospace font
> ```

Are we sure this is what we want?
___
*[`pixeltable/utils/formatter.py` line 171 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYo7Hw484BXxFcZk4p:-NyYo7Hw484BXxFcZk4q:bqv1is) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L171)):*
> ```Python
>                 2) whitespace is respected (newlines and spaces)
>             Set style so that:
>                 1) json is left-aligned  (overrides jupyter default to right-aligned)
> ```

Very happy to see this change! Are we doing the same for strings?
___
*[`pixeltable/utils/formatter.py` line 178 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYpDTe9mfwhJl2V8b1:-NyYpDTe9mfwhJl2V8b2:b-plhsp) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L178)):*
> ```Python
>             to avoid excessive vertical space.
>         """
>         return f"""<pre style="text-align:left; background-color:transparent; margin:0;">{self._format_json_helper(obj, level=0)}</pre>"""
> ```

Make this a multiline string for clarity and consistency (following format_img/format_video/etc below)

aaron-siegel (commented):
High level comments.
1. The formatter definitely needs tests.
2. I'm surprised we're doing the JSON rendering by hand? Seems potentially fraught/full of edge cases. Have we looked to see if there are any libraries out there that pretty-print JSON?

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/177#-:-NyYqmsZAostAkLWlodR:bj3m3jr)* status: 0 of 7 files reviewed, 22 unresolved discussions (waiting on @orm011)

orm011 (commented):
1. I will add some tests, I have some I used during development but as they are visual, they are in a notebook.

2. I tried the built in json pretty printer first, the main dealbreaker, is it does not understand that we want to treat nested lists that happen to be numeric arrays just like numpy does, and that we ant to keep consistency between types within the json, and those in other columns, which were things you expressed.  Now we have that.  I also know there are some web-based json displays but they would still have a similar issue with consistency.   


My high level retrospective on this, the time went into:
1.  dealing with json so that we can control consistency between json and other types when they appear outside, with multiple tests written for this (will add to test suite, currently in a notebook)  (1 day)
2.  figuring out how to ensure the pretty-printing we get from numpy and our own json printer, or any text-based json pretty printer for that matter, can still render ok when put in HTML (using <pre> tags was the solution in the end, which lets us control formatting with whitespace).  .5 days
3. string types escaping (which we now escape at a fine-grain).  .5 days. In retrospective, this was a much higher priority issue than I initially thought, given we were effectively allowing injection of randomly scraped code (eg from html docs) into the notebook of users.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/177#-:-Ny_ohyNDxyiS4c3X2eI:bjk1j3z)* status: 0 of 9 files reviewed, 22 unresolved discussions (waiting on @aaron-siegel)

___
*[`docs/release/tutorials/rag-operations.ipynb` line 1442 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYgXAJD7zoVOXelkPM:-Ny_rH9PE8XhDz24BUO1:bhh990j) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/docs/release/tutorials/rag-operations.ipynb#L1442)):*

I don't love that scientific notation is being used for values like 0.330 (just picking an example). It's more compact to write it without scientific notation! Do we have control over this (i.e. only use scientific notation when it's actually more compact than a decimal of equal precision?)
</blockquote></details>

This is the standard numpy setting. I believe it has to do with keeping values aligned. 
___
*[`docs/release/tutorials/rag-operations.ipynb` line 1589 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYh-Sg84PBXRe7JSni:-Ny_sDXQ9XtRG-sqmfVf:b-c6i1ny) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/docs/release/tutorials/rag-operations.ipynb#L1589)):*

Why are we still seeing high-precision numbers in the notebooks?
</blockquote></details>

They're not in the rendered notebook:
https://github.com/orm011/pixeltable/blob/concise_format/docs/release/tutorials/rag-operations.ipynb
___
*[`pixeltable/dataframe.py` line 55 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/177#-Ny7d7B8677cainBy32t:-Ny8-iq37qQ6Z-yKFt8H:b-a5rzeo) ([raw file](https://github.com/pixeltable/pixeltable/blob/ff9cc8733d05d9f78b5451b35f42eadd6cb20837/pixeltable/dataframe.py#L55)):*

We also need a `_format_dict` and `_format_list`, and to process them recursively. Arrays/lists inside of dicts are fairly common.
</blockquote></details>

Yeah, we need to look into json col, probably needs its own PR early next week?if this works 
___
*[`pixeltable/dataframe.py` line 148 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/177#-Ny7kObB5mwYfl6nioU8:-Ny801hr6M4m5kvPqM1C:b-czlzq9) ([raw file](https://github.com/pixeltable/pixeltable/blob/ff9cc8733d05d9f78b5451b35f42eadd6cb20837/pixeltable/dataframe.py#L148)):*

I'd argue for increasing threshold and edgeitems a bit, maybe to 16/6? We could think about making this a configurable parameter in the pixeltable env (but obviously that's not urgent)
</blockquote></details>

yeah, we need to expose some way of controlling these.
___
*[`pixeltable/dataframe.py` line 152 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/177#-Ny7kZpJ1-zazTaiVGIx:-Ny_sVfI3FZ3sVAMvoVY:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ff9cc8733d05d9f78b5451b35f42eadd6cb20837/pixeltable/dataframe.py#L152)):*

Suggestion: use `......` as the separator instead of `...`

The single ellipsis is sufficient for arrays but could get buried in the string.
</blockquote></details>

Done.
___
*[`pixeltable/utils/formatter.py` line 49 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYi4X-62VpHjMxW9vd:-Ny_sWdY2sbp-LmFqsPB:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L49)):*

This should go at the top of the file
</blockquote></details>

Done.
___
*[`pixeltable/utils/formatter.py` line 52 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYiaMVBrAvmdFpLpVu:-Ny_sXSW7Bkuz3mC7mKW:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L52)):*

dict has two type parameters, right?
</blockquote></details>

Done.
___
*[`pixeltable/utils/formatter.py` line 55 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYikGUBea0iKJJbK3v:-Ny_kDY13fEvCiMR7bjQ:b4mo326) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L55)):*

This is really awkward. It'd be a clearer separation of logic to take an `Iterable` as the parameter rather than `Union[list, dict]`, and then when using it on a dict, call `contains_only_simple_elements(my_dict.values())`.

Also, I'd argue it'd be clearer to negate the function and call it `contains_collection` or some such; that also has the benefit of avoiding introducing the bespoke term `simple` (not a huge deal, but it was an extra/avoidable cognitive step for me to figure out what that meant)
</blockquote></details>

I can switch it to just look at the values as you say.
This is all in the context of json, where there are arrays and dictionaries,  there are no arbitrary iterables, but to call it as you would like i guess iterable would be the option.
___
*[`pixeltable/utils/formatter.py` line 60 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYnGED5wbMiec3_kul:-Ny_sYaW811WGHA7YuAR:btapvre) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L60)):*

We definitely need tests for this. There's a lot of new logic in here.
</blockquote></details>

will add. 
___
*[`pixeltable/utils/formatter.py` line 70 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYjoXz0dEoaYSfix6j:-Ny_kpmEDu04W7SAKeQH:bhzrg1w) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L70)):*

why's this taking a list as input?
</blockquote></details>

because its input is supposed to be a json array.
___
*[`pixeltable/utils/formatter.py` line 71 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYk6e27Y7q3WJD3cKt:-Ny_kxMD39jDnTGU-dPV:b-80tlqk) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L71)):*

I'd like to get away from docstrings that are really just notes (vs. proper documentation).

It's not necessary to have docstrings for private methods, so the simplest thing to do would be convert this to a code comment. If we do want to include the docstrings, let's write them properly (clear description of what the function does, complete sentences, correct punctuation, etc.)
</blockquote></details>

Sure.
___
*[`pixeltable/utils/formatter.py` line 88 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYk03U6NqJu8N7F4nW:-Ny_l5Dv6z-e2q92l3Tb:bb74ce2) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L88)):*

These TODOs should definitely be in comments, not the docstring
</blockquote></details>

Sure
___
*[`pixeltable/utils/formatter.py` line 96 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYkv1p6iYgQOg7R2CN:-Ny_l6uq8YuAzzNbbuA7:b-yngngx) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L96)):*

why's this a nested function? since it depends only on level, can't it just be a variable? (`prefix = ' ' * (level * INDENT)`)
</blockquote></details>

Because it may get called with different arguments within a single function.
___
*[`pixeltable/utils/formatter.py` line 103 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYlJzN0CwachW0bJq8:-Ny_lAoOAS5PIOxgDuOT:bbqg15w) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L103)):*

better to say `obj = obj.tolist()` than to do a recursive call
</blockquote></details>

see next comment
___
*[`pixeltable/utils/formatter.py` line 112 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYljViCLbaym-ceNvP:-Ny_lGiB3OrijKCqSlBH:b-kyq01e) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L112)):*

Actually (cf. my previous comment), this seems weird. we're converting to a list, then back to an ndarray again? Can't this be reorganized so as to avoid that?
</blockquote></details>

Not quite. That's why its easier to reason recursively: at that point we don't have much information about what the array has in it. The numeric ndarray part is for numeric arrays only. This could be an np.array() of other things, and those are not treated the same way.  

Instead of duplicating the logic, I decided it makes sense to simply unwrap such numpy arrays and then have them fully follow the logic intended for json lists, this makes the semantics obvious even if it adds a new call. This was a real issue, sometimes we nest random numpy arrays inside 'json'. (in reality this is not really Json, its our own nested object datatype, an extension of json, that we serialize to json, as json would not take in numpy arrays, something we should discuss)

It may turn out to be a numeric array in the end, but thats ok. this is not perf critical and can be revisited if we find rendering somehow takes long.
___
*[`pixeltable/utils/formatter.py` line 156 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYnRuJEQKoTfBDC3Ec:-Ny_mGZs2PSBOKZ0PjdQ:b-sar9xu) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L156)):*

Why is `_format_string` a protected method but not `format_float`?

Are strings going to be properly quoted when they appear inside JSON?
</blockquote></details>

Should they be properly quoted when outside of json? Right now I keep them consistent inside and outside json, I don't quote newlines.  But I do escape HTML characters. We can switch, but my main goal was consistency, as you expressed.
___
*[`pixeltable/utils/formatter.py` line 160 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYneoKEjFWcHQc_9Ep:-Ny_mgLV1LRYzD4PaCVU:b-3pfqqd) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L160)):*

Pretty sure `AssertionError` should never be referenced explicitly (use `assert` keyword for assertions).
</blockquote></details>

Ruff with the settings from the repo flagged me for saying assert False and suggested this fix
___
*[`pixeltable/utils/formatter.py` line 164 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYpMnx6jpsR_qbdF6w:-Ny_mom05XWVv1Ae52kJ:bb7oryy) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L164)):*

Let's try to avoid vague language in docstrings ("elementary types like int, float, and arrays"). Are those the only three types involved? If so, remove the word "like". If not, list out the types explicitly (and clarify the meaning of "arrays", e.g., something like "the elementary types int, float, bool, and np.ndarray".
</blockquote></details>

sure
___
*[`pixeltable/utils/formatter.py` line 166 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYoSDW-C-HbHffPgvR:-Ny_myRaApP0OU1v8T9K:b66ctzu) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L166)):*

I don't think this will render properly as a docstring? I suggest using markdown bullets
</blockquote></details>

I'll just convert it to comments. 
___
*[`pixeltable/utils/formatter.py` line 168 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYoGui4U0w-jyvakuW:-Ny_n1hF805Q01tkVf5v:b-pmrsur) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L168)):*

Are we sure this is what we want?
</blockquote></details>

Re. the monospace part,  numpy formatting is designed with a monospace  world in mind.
___
*[`pixeltable/utils/formatter.py` line 171 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYo7Hw484BXxFcZk4p:-Ny_nNhTAzP4a71msFJ6:b-j0icov) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L171)):*

Very happy to see this change! Are we doing the same for strings?
</blockquote></details>

Yes, goal was consistency across all types, and most of our types can go in a json.
___
*[`pixeltable/utils/formatter.py` line 178 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/177#-NyYpDTe9mfwhJl2V8b1:-Ny_nVc4AN-8ryK5H0-8:b-pilt7a) ([raw file](https://github.com/pixeltable/pixeltable/blob/f013bb4859018173b38629f553cca9c5a23e00c7/pixeltable/utils/formatter.py#L178)):*

Make this a multiline string for clarity and consistency (following format_img/format_video/etc below)
</blockquote></details>

I'll add a comment explaining this instead? I think it probably needs to stay as it is.
The  way <pre> works is that inside that tag ends, all whitepsace is important, so breaking it up will add whitespace to all
rendered cells. (I started off prettier but had to fix this.)

orm011 (commented):
current state of things

object detection tutorial (json, arrays)
before: https://github.com/orm011/pixeltable/blob/04f73ac/docs/release/tutorials/object-detection-in-videos.ipynb
after: https://github.com/orm011/pixeltable/blob/concise_format/docs/release/tutorials/object-detection-in-videos.ipynb

rag operation tutorial (strings, arrays)
before:https://github.com/orm011/pixeltable/blob/04f73ac/docs/release/tutorials/rag-operations.ipynb
after: https://github.com/orm011/pixeltable/blob/concise_format/docs/release/tutorials/rag-operations.ipynb

anm tutorial (strings)
before:https://github.com/orm011/pixeltable/blob/query/docs/tutorials/document-demo-anm.ipynb
after: https://github.com/orm011/pixeltable/blob/concise_format/docs/drafts/document-demo-anm.ipynb (up to cell 14, as there is no query in this branch)

This las demo tells me:
Its possible we need to escape string whitespace (render without newlines). But the strange markup is gone.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/177#-:-Ny_siJfAWAGwa5VGrAP:b11idt4)* status: 0 of 9 files reviewed, 22 unresolved discussions (waiting on @aaron-siegel)

PR #180, opened 2024-05-18
New Working with Data (Video Data) Tutorial

pierrebrunelle:
This is an extract of the Object Detection Video Tutorial center around only focus on how to work with Frame and Video and then linking to the next tutorials


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/180)

PR #182, opened 2024-05-20
Bug fix: exception in insert()

mkornacker:
This happens after dropping a non-nullable column from a table, reloading the catalog and then running insert(), at which point
TableVersion.get_required_col_names() would include this column with a None name.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/182)

aaron-siegel (approved):

PR #183, opened 2024-05-22
Prep for type-checking work

aaron-siegel:
- Add proper type signatures to decorators
- Make mypy somewhat more lenient
- Add pytype as a dependency for comparison


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/183)

mkornacker (approved):

PR #186, opened 2024-05-27
new BtreeIndex class

mkornacker:
This replaces the default Postgres B-tree indices that were created directly on scalar and media columns with explicitly maintained indices using the standard index maintenance mechanisms (eg, is only maintained for the current version).

Includes:
- BtreeIndex (subclass of IndexBase)
- bug fixes for maintaining index value columns during updates


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/186)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/186)* status: 0 of 23 files reviewed, 8 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/table.py` line 537 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NyzxRl88uf5DNZXl4lh:-NyzxRl88uf5DNZXl4li:bvkq8nl) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/catalog/table.py#L537)):*
> ```Python
>     def drop_index(
>             self, *, column_name: Optional[str] = None, idx_name: Optional[str] = None,
>             _idx_class: Optional[Type[index.IndexBase]] = None
> ```

Why does `_idx_class` need to be a parameter here? Is it possible to have two indices on the same column with the same name, but different classes?

If it's absolutely necessary, then I'd argue it should be a protected method (`_drop_index`, with a separate `drop_btree_index` etc)
___
*[`pixeltable/catalog/table_version.py` line 456 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NyzybH0Bs6QEqPEV2Zh:-NyzybH0Bs6QEqPEV2Zi:bzad224) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/catalog/table_version.py#L456)):*
> ```Python
>         return status
> 
>     def _add_columns(self, cols: List[Column], conn: sql.engine.Connection, print_stats: bool = False) -> UpdateStatus:
> ```

I'm very happy to see `_update_md` being factored out of `_add_columns` and `_drop_columns` (In fact I was going to suggest this! Stored proxy columns for remotes is another use cases where a table version increment adds/removes columns *and also* does a bunch of other stuff, and it's much cleaner to factor out `_update_md` as a separate step.)
___
*[`pixeltable/catalog/table_version.py` line 1009 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-Nyzz1l-1CuXNubKrSdo:-Nyzz1l-1CuXNubKrSdp:b2op2ta) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/catalog/table_version.py#L1009)):*
> ```Python
> 
>     def get_idx_val_columns(self, cols: Iterable[Column]) -> Set[Column]:
>         result: set[Column] = set()
> ```

This could be done as a comprehension:
```
result = {
    info.val_col
    for col in cols
    for info in col.get_idx_info().values()
}
```
___
*[`pixeltable/catalog/table_version.py` line 1015 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NyzzovV1EoKcwA39-6-:-Nyzzjvq1-z-vCOLmpMw:b7r7zcy) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/catalog/table_version.py#L1015)):*
> ```Python
>         return result
> 
>     def get_dependent_columns(self, cols: List[Column]) -> Set[Column]:
> ```

This doesn't actually depend on `self`, right?

It'd make more sense to me as a method of `Column` (with the trivial comprehension inlined when multiple columns are involved).

I realize that it's a legacy thing and this PR doesn't actually change this method in any material way (just makes it typesafe)
___
*[`pixeltable/exprs/column_ref.py` line 112 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-Nz--9o17pbDc_5bWFqm:-Nz--9o17pbDc_5bWFqn:b-23abxp) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/exprs/column_ref.py#L112)):*
> ```Python
>         tbl_version = catalog.Catalog.get().tbl_versions[(tbl_id, version)]
>         # don't use tbl_version.cols_by_id here, this might be a snapshot reference to a column that was then dropped
>         col = [col for col in tbl_version.cols if col.id == col_id][0]
> ```

The preferred way to do this is

`col = next(col for col in tbl_version.cols if col.id == col_id)`

(avoids creating a list only to deference it)
___
*[`pixeltable/exprs/comparison.py` line 34 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-Nz--j9o6uDmKoy-vDdv:-Nz--j9o6uDmKoy-vDdw:b740dvu) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/exprs/comparison.py#L34)):*
> ```Python
>             self.components = [op1, op2]
> 
>         if self.is_search_arg_comparison and self._op2.col_type.is_string_type() and len(self._op2.val) > 64:
> ```

`64` should be a defined constant somewhere, not hardcoded here. [EDIT: I see it's defined in `BtreeIndex`, just reference that]

I'd argue it should be greater than 64, at least 128. Or we can use the hash-proxy trick to allow exact lookups against strings of length > 64, with the length limit applying only to inequalities.
___
*[`pixeltable/index/btree.py` line 12 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-Nz-1OXP1sLvxahWUqd_:-Nz-1OXP1sLvxahWUqda:b-oqeqmv) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/index/btree.py#L12)):*
> ```Python
> 
> 
> class BtreeIndex(IndexBase):
> ```

We need some test cases to check that equality and inequality work properly for indexed string columns with a mix of short and long strings.
___
*[`pixeltable/index/btree.py` line 23 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-Nz-1DW20xx7OSqqDblq:-Nz-1DW20xx7OSqqDblr:by3yog5) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/index/btree.py#L23)):*
> ```Python
>             return None
>         if len(s) > BtreeIndex.MAX_STRING_LEN:
>             return None
> ```

We should be truncating strings, not filtering them out, right? Otherwise inequalities won't work correctly. Unless I'm missing something.

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/186)* status: 0 of 23 files reviewed, 8 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/catalog/table.py` line 537 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NyzxRl88uf5DNZXl4lh:-Nz-GS0o1_TBbIBeLSgE:b-l9gb0n) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/catalog/table.py#L537)):*

Why does `_idx_class` need to be a parameter here? Is it possible to have two indices on the same column with the same name, but different classes?

If it's absolutely necessary, then I'd argue it should be a protected method (`_drop_index`, with a separate `drop_btree_index` etc)
</blockquote></details>

You can have two indices on the same column, but only if they have different classes.

It's more convenient to specify the column name rather than the index name (in particular right now, since we don't have the functionality to list existing indices).

We can have a separate _drop_index(), though.
___
*[`pixeltable/catalog/table_version.py` line 1009 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-Nyzz1l-1CuXNubKrSdo:-Nz-HVq49Kmrez6G7euD:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/catalog/table_version.py#L1009)):*

This could be done as a comprehension:
```
result = {
    info.val_col
    for col in cols
    for info in col.get_idx_info().values()
}
```
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table_version.py` line 1015 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NyzzovV1EoKcwA39-6-:-Nz-IC096Ue4vnPCOIj_:b14j44u) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/catalog/table_version.py#L1015)):*

This doesn't actually depend on `self`, right?

It'd make more sense to me as a method of `Column` (with the trivial comprehension inlined when multiple columns are involved).

I realize that it's a legacy thing and this PR doesn't actually change this method in any material way (just makes it typesafe)
</blockquote></details>

I'd rather keep it here, it doesn't make the code any more complex and it's always called in the context of a table version, and it might need to be restricted to a particular table version in the future.
___
*[`pixeltable/exprs/column_ref.py` line 112 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-Nz--9o17pbDc_5bWFqm:-Nz-J0LR0l2RSurKuLaE:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/exprs/column_ref.py#L112)):*

The preferred way to do this is

`col = next(col for col in tbl_version.cols if col.id == col_id)`

(avoids creating a list only to deference it)
</blockquote></details>

Done.
___
*[`pixeltable/exprs/comparison.py` line 34 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-Nz--j9o6uDmKoy-vDdv:-Nz-JQ697JCGgs3wQxqC:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/exprs/comparison.py#L34)):*

`64` should be a defined constant somewhere, not hardcoded here. [EDIT: I see it's defined in `BtreeIndex`, just reference that]

I'd argue it should be greater than 64, at least 128. Or we can use the hash-proxy trick to allow exact lookups against strings of length > 64, with the length limit applying only to inequalities.
</blockquote></details>

Done.
___
*[`pixeltable/index/btree.py` line 23 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-Nz-1DW20xx7OSqqDblq:-Nz-JWeG4J82bDPk2Wth:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/index/btree.py#L23)):*

We should be truncating strings, not filtering them out, right? Otherwise inequalities won't work correctly. Unless I'm missing something.
</blockquote></details>

Done.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/186)* status: 0 of 26 files reviewed, 9 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/env.py` line 312 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NzV2LMT5XC_CtinJMMe:-NzV2LMT5XC_CtinJMMf:btefk5g) ([raw file](https://github.com/pixeltable/pixeltable/blob/4f837ad1bb4a308b11ace08fa3994bf0365f4206/pixeltable/env.py#L312)):*
> ```Python
>                 stmt = (
>                     f"CREATE DATABASE {preparer.quote(self._db_name)} "
>                     "ENCODING 'utf-8' LC_COLLATE 'C' LC_CTYPE 'C' TEMPLATE template0"
> ```

Is this required for proper behavior of the Pixeltable DB? How is it going to work with existing installations?
___
*[`pixeltable/env.py` line 319 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NzV2Umv-OjssuTuO2ph:-NzV2Umv-OjssuTuO2pi:bqwvgbn) ([raw file](https://github.com/pixeltable/pixeltable/blob/4f837ad1bb4a308b11ace08fa3994bf0365f4206/pixeltable/env.py#L319)):*
> ```Python
> 
>         # enable pgvector
>         db_url = self._db_server.get_uri(database=self._db_name)
> ```

It'd be clearer to use a different variable name here (`postgres_db_url` vs `pxt_db_url`, say)
___
*[`pixeltable/plan.py` line 291 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NzV2wDVEV5Vy1_tcty0:-NzV2wDVEV5Vy1_tcty1:b-oal4wa) ([raw file](https://github.com/pixeltable/pixeltable/blob/4f837ad1bb4a308b11ace08fa3994bf0365f4206/pixeltable/plan.py#L291)):*
> ```Python
>         return (
>             plan,
>             [f'{c.tbl.name}.{c.name}' for c in updated_cols + list(recomputed_user_cols)],
> ```

`list(recomputed_user_cols)` is redundant? (it's already a list right?)
___
*[`pixeltable/catalog/table.py` line 537 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NyzxRl88uf5DNZXl4lh:-NzV39Le4tn64c-tYUoC:b-alt95c) ([raw file](https://github.com/pixeltable/pixeltable/blob/e2ecc0fac535def68ac488abdbfad8288dad7d97/pixeltable/catalog/table.py#L537)):*

drop_index() is still useful as a user-facing function. If you have only one index on a column, it's unambiguous and easier to remember than providing the index name.

I removed the _idx_class parameter in the latest changes.
</blockquote></details>

Ok, that works.
___
*[`pixeltable/index/btree.py` line 16 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NzV5rBy6a26sS-PPCt3:-NzV5rBy6a26sS-PPCt4:b-g3jd79) ([raw file](https://github.com/pixeltable/pixeltable/blob/4f837ad1bb4a308b11ace08fa3994bf0365f4206/pixeltable/index/btree.py#L16)):*
> ```Python
>     Interface to B-tree indices in Postgres.
>     """
>     MAX_STRING_LEN = 256
> ```

`MAX_INDEXED_STRING_LEN` or some such might be a better name (it's somewhat implied since it's part of `BtreeIndex` but I'd argue for the extra clarity here)
___
*[`tests/test_index.py` line 266 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NzV4oMq1N-KZG6ynyu2:-NzV4oMq1N-KZG6ynyu3:bvrok4e) ([raw file](https://github.com/pixeltable/pixeltable/blob/4f837ad1bb4a308b11ace08fa3994bf0365f4206/tests/test_index.py#L266)):*
> ```Python
>         n = 255
>         data = [create_random_str(n) for _ in range(self.BTREE_TEST_NUM_ROWS)]
>         t = self.run_btree_test(data, pxt.StringType())
> ```

This isn't a complete test (unless I'm missing something), since it doesn't contain longer strings. We should have a test that includes some longer strings of length >= MAX_STRING_LEN in the column, to check that indexing/truncation still works.

To properly test the edge case, there should also be a test that includes both `s` and `s + 'a'` (say), where `s` is a string of length exactly `MAX_STRING_LEN-1`, and then tests both `>=` and `>` against `s`.

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/186)* status: 0 of 26 files reviewed, 9 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/env.py` line 312 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NzV2LMT5XC_CtinJMMe:-NzVTNOnDIxKDhD3nbSM:b-x6x16i) ([raw file](https://github.com/pixeltable/pixeltable/blob/4f837ad1bb4a308b11ace08fa3994bf0365f4206/pixeltable/env.py#L312)):*

Is this required for proper behavior of the Pixeltable DB? How is it going to work with existing installations?
</blockquote></details>

It is, and we'll need to tell Obvio to re-create their db
___
*[`pixeltable/env.py` line 319 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NzV2Umv-OjssuTuO2ph:-NzVTYzL74q2F4z-epB1:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/4f837ad1bb4a308b11ace08fa3994bf0365f4206/pixeltable/env.py#L319)):*

It'd be clearer to use a different variable name here (`postgres_db_url` vs `pxt_db_url`, say)
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 291 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NzV2wDVEV5Vy1_tcty0:-NzVUNYH1reZ5tZ1keFq:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/4f837ad1bb4a308b11ace08fa3994bf0365f4206/pixeltable/plan.py#L291)):*

`list(recomputed_user_cols)` is redundant? (it's already a list right?)
</blockquote></details>

Done.
___
*[`tests/test_index.py` line 266 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/186#-NzV4oMq1N-KZG6ynyu2:-NzVWc839KA_kQkYBV9A:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/4f837ad1bb4a308b11ace08fa3994bf0365f4206/tests/test_index.py#L266)):*

This isn't a complete test (unless I'm missing something), since it doesn't contain longer strings. We should have a test that includes some longer strings of length >= MAX_STRING_LEN in the column, to check that indexing/truncation still works.

To properly test the edge case, there should also be a test that includes both `s` and `s + 'a'` (say), where `s` is a string of length exactly `MAX_STRING_LEN-1`, and then tests both `>=` and `>` against `s`.
</blockquote></details>

Done.

aaron-siegel (approved):

PR #188, opened 2024-05-30
Use a decorator pattern for schema converters

aaron-siegel:
Right now, in order to define a schema converter, we have to mention the version number three times:

```
def convert_n(engine): ...

register_converter(n, convert_n)
```

By using a decorator pattern, we can reduce this to once:
```
@register_converter(version=n)
def _(engine): ...
```


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/188)

PR #191, opened 2024-06-09
bug fix: missing default values for non-computed view columns

mkornacker:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/191)

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/191)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exec/in_memory_data_node.py` line 36 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/191#-O-2SKuS0_mvyjmAYwjB:-O-2SKuS0_mvyjmAYwjC:b-fmbafy) ([raw file](https://github.com/pixeltable/pixeltable/blob/8d1531fbbbfb7bffa77dee71b397584063a45b28/pixeltable/exec/in_memory_data_node.py#L36)):*
> ```Python
>     def _open(self) -> None:
>         """Create row batch and populate with self.input_rows"""
>         user_column_info = {
> ```

suggestion, call these `user_columns_by_name` and `output_columns_by_idx` or some such (they look so similar that i misread the code at first and overlooked that the keys are different)
___
*[`pixeltable/exprs/row_builder.py` line 258 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/191#-O-2RJ-s0l9i-U86kG8v:-O-2RJ-s0l9i-U86kG8w:b-alvl08) ([raw file](https://github.com/pixeltable/pixeltable/blob/8d1531fbbbfb7bffa77dee71b397584063a45b28/pixeltable/exprs/row_builder.py#L258)):*
> ```Python
>         if exclude is None:
>             exclude = []
>         if len(targets) == 0:
> ```

This won't work for a general `Iterable`. If you're changing the function signature to take `Iterable`s and you want to use `len` (or make multiple traversals over the inputs, as happens below) then you need to put

```
targets = list(targets)
exclude = list(exclude)
```

before their first use
___
*[`tests/test_view.py` line 529 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/191#-O-2Rq-M0KTmf-jCwE6S:-O-2Rq-M0KTmf-jCwE6T:b-ha41oa) ([raw file](https://github.com/pixeltable/pixeltable/blob/8d1531fbbbfb7bffa77dee71b397584063a45b28/tests/test_view.py#L529)):*
> ```Python
>         check(s, v, view_s)
> 
>     def test_column_defaults(self, reset_db) -> None:
> ```

Might be useful to add a comment that mentions the bug that led to this test (as a way of documenting the pattern that it's testing for)

PR #194, opened 2024-06-14
Fix handling of grayscale images in YoloX and DETR models

aaron-siegel:
Fixes handling of grayscale images in YoloX and DETR models. Updates the `get_image_files()` test utility to ensure that the list is deterministic (platform-independent) and that small initial segments of the image list are guaranteed to contain images representing all modes.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/194)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/194)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @aaron-siegel)

___
*[`pixeltable/functions/util.py` line 21 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/194#-O-Mc5wn7CW9vyAwGXU6:-O-Mc5wn7CW9vyAwGXU7:bbn31vu) ([raw file](https://github.com/pixeltable/pixeltable/blob/21cd10c9732635536baee0a462f453e5e306de31/pixeltable/functions/util.py#L21)):*
> ```Python
>     """
>     if image.mode == '1' or image.mode == 'L':
>         return image.convert('RGB')
> ```

I don't think this is performance-transparent, and it exceeds what is needed (which is simply to have 3-dim ndarrays).

mkornacker (approved):

PR #198, opened 2024-06-19
Tab completion cleanup, Phase 1

aaron-siegel:
Cleans up tab completion for:
- All public modules (pxt, pxt.io, pxt.iterators, pxt.functions, pxt.functions.*)
- Tables and other schema objects

As a bonus, tab-completion now works for columns (if you type `my_table.` and press <TAB>, the currently defined columns will appear in the list). Also, dict-completion works the same way (if you type `my_table[` and press <TAB>).

Also included in this PR is a small cleanup/refactor of `pxt.functions.image` (including removing `pxt.functions.pil.image` and folding it in).

Still TODO (not in this PR):
- DataFrames
- Exprs (in particular ColumnRefs)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/198)

pixeltable/utils/code.py
```diff
@@ -0,0 +1,27 @@
+import types
```
mkornacker, line 1: code.py?

Might be a good idea to leave a comment describing what should go in here.

mkornacker (approved):

pixeltable/utils/code.py
aaron-siegel: Sure, I'll do that. I'm also open to suggestions for a better name for the .py file :/

PR #199, opened 2024-06-20
Add retry logic to "install dependencies" CI step

aaron-siegel:
Adds retry logic to the "install dependencies" CI step. This should reduce the frequency of CI failures due to connection errors on the runners.

It isn't perfect (in particular, a connection error could still occur after the tests have started - say, while downloading a model) but hopefully will help somewhat.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/199)

PR #206, opened 2024-06-27
Some API enhancements

aaron-siegel:
- Allow updates and deletes from a DataFrame: `t.where(t.c1 < 10).delete()`
- Allow view creation from a DataFrame: `pxt.create_view('my_view', t.where(t.c1 < 10))`
- Allow empty select statement: `t.select()`
- Rename `t.df()` to `t._df()` (the user-facing way to do this is now `t.select()`)

Moves a bunch of update/delete logic from `Table` into `TableVersion`.

Reference doc: https://www.notion.so/pixeltable/Pixeltable-API-Suggestions-0cf550f777f245f2a82654125b9ba465

### DIFF
--- a/pixeltable/catalog/globals.py
+++ b/pixeltable/catalog/globals.py
@@ -7,6 +7,8 @@ _logger = logging.getLogger('pixeltable')
 
 # name of the position column in a component view
 POS_COLUMN_NAME = 'pos'
+_ROWID_COLUMN_NAME = '_rowid'
+
 
 @dataclasses.dataclass
 class UpdateStatus:
--- a/pixeltable/catalog/table_version.py
+++ b/pixeltable/catalog/table_version.py
@@ -703,13 +703,6 @@ class TableVersion:
         _logger.info(f'TableVersion {self.name}: new version {self.version}')
         return result
 
-    def update(
-            self, update_targets: dict[Column, 'pixeltable.exprs.Expr'],
-            where_clause: Optional['pixeltable.exprs.Predicate'] = None, cascade: bool = True
-    ) -> UpdateStatus:
-        with Env.get().engine.begin() as conn:
-            return self._update(conn, update_targets, where_clause, cascade)
-
     def batch_update(
             self, batch: list[dict[Column, 'pixeltable.exprs.Expr']], rowids: list[Tuple[int, ...]],
             cascade: bool = True
--- a/pixeltable/catalog/table_version.py
+++ b/pixeltable/catalog/table_version.py
@@ -28,6 +28,7 @@ from ..func.globals import resolve_symbol
 
 _logger = logging.getLogger('pixeltable')
 
+
 class TableVersion:
     """
     TableVersion represents a particular version of a table/view along with its physical representation:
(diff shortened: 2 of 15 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/206)

mkornacker (changes requested):
Let's discuss tomorrow.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/206#-:-O0fUOJl10weakCFrl38:bhdwmln)* status: 0 of 14 files reviewed, 8 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/dataframe.py` line 713 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/206#-O0fShb36yLLVlU8HPr2:-O0fShb36yLLVlU8HPr3:bdx2lvj) ([raw file](https://github.com/pixeltable/pixeltable/blob/7da111d55084ea88fb890e04db16072831606cc8/pixeltable/dataframe.py#L713)):*
> ```Python
>     def update(self, value_spec: dict[str, Any], cascade: bool = True) -> UpdateStatus:
>         self._validate_writable('update')
>         return self.tbl.update(value_spec, where=self.where_clause, cascade=cascade)
> ```

When update() is legal (= it's a mutable table), self.tbl is the same as self.tbl.tbl_version.path, no?
___
*[`pixeltable/dataframe.py` line 717 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/206#-O0fS3e47Pk0kKg_IxrD:-O0fS3e47Pk0kKg_IxrE:bsg5fei) ([raw file](https://github.com/pixeltable/pixeltable/blob/7da111d55084ea88fb890e04db16072831606cc8/pixeltable/dataframe.py#L717)):*
> ```Python
>     def delete(self) -> UpdateStatus:
>         self._validate_writable('delete')
>         return self.tbl.delete(where=self.where_clause)
> ```

I believe this would still allow you to call delete() on a df over a view, which isn't legal. For delete() to be legal, self.tbl has length 1.
___
*[`pixeltable/dataframe.py` line 719 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/206#-O0W5o4h3qVjeDwQh3gg:-O0W5o4h3qVjeDwQh3gh:bfc4d6l) ([raw file](https://github.com/pixeltable/pixeltable/blob/7da111d55084ea88fb890e04db16072831606cc8/pixeltable/dataframe.py#L719)):*
> ```Python
>         return self.tbl.delete(where=self.where_clause)
> 
>     def _validate_writable(self, op_name: str) -> None:
> ```

function comment missing

writable?
___
*[`pixeltable/globals.py` line 81 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/206#-O0fNpqQDifzvFnaUWsa:-O0fNpqQDifzvFnaUWsb:b-cqzxq2) ([raw file](https://github.com/pixeltable/pixeltable/blob/7da111d55084ea88fb890e04db16072831606cc8/pixeltable/globals.py#L81)):*
> ```Python
> def create_view(
>     path_str: str,
>     base: Union[catalog.Table, DataFrame],
> ```

This should be called something else, 'base' is used exclusively for TableVersions.
___
*[`pixeltable/globals.py` line 130 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/206#-O0W682aERB8xdFtql7f:-O0W682aERB8xdFtql7g:b9ff7iu) ([raw file](https://github.com/pixeltable/pixeltable/blob/7da111d55084ea88fb890e04db16072831606cc8/pixeltable/globals.py#L130)):*
> ```Python
>         base._validate_writable('create_view')
>         tbl_version_path = base.tbl
>         filter = base.where_clause if filter is None else base.where_clause & filter
> ```

I don't know if combining where clauses here is a good idea. I don't find that particularly intuitive (I'm sure some people would assume the filter overrides the dataframe). Probably best to disallow a separate filter rather than create a surprise - it doesn't serve a purpose anyway.
___
*[`pixeltable/catalog/table_version_path.py` line 33 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/206#-O0W7rpF1NNw2dg8EB11:-O0W7rpF1NNw2dg8EB12:b-t7rcwm) ([raw file](https://github.com/pixeltable/pixeltable/blob/7da111d55084ea88fb890e04db16072831606cc8/pixeltable/catalog/table_version_path.py#L33)):*
> ```Python
>     """
> 
>     __ROWID_COLUMN_NAME = '_rowid'
> ```

Move into catalog/globals.py, it should have been there in the first place.
___
*[`pixeltable/catalog/table_version_path.py` line 292 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/206#-O0fRpz59Lw9ef1wR14w:-O0fRpz59Lw9ef1wR14x:brztd1l) ([raw file](https://github.com/pixeltable/pixeltable/blob/7da111d55084ea88fb890e04db16072831606cc8/pixeltable/catalog/table_version_path.py#L292)):*
> ```Python
>         return update_targets
> 
>     def delete(self, where: Optional['pixeltable.exprs.Predicate'] = None) -> UpdateStatus:
> ```

What's the reason for moving this? I think it's better to keep this in one place (cross-class calls make the code more complex/harder to follow).
___
*[`tests/test_dataframe.py` line 243 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/206#-O0fU-Xz3pgeLMuQYaL0:-O0fU-Y-4ZpOuGuKSm4_:by4fsgf) ([raw file](https://github.com/pixeltable/pixeltable/blob/7da111d55084ea88fb890e04db16072831606cc8/tests/test_dataframe.py#L243)):*
> ```Python
>         assert t.count() == 0
> 
>         # select_list
> ```

Missing error cases:
- delete from view
- update snapshot
- delete from snapshot

pixeltable/catalog/globals.py
```diff
@@ -7,6 +7,8 @@
 
 # name of the position column in a component view
 POS_COLUMN_NAME = 'pos'
+_ROWID_COLUMN_NAME = '_rowid'
```
mkornacker, line 10: ROWID_COLUMN_NAME

Also, add explanatory comment

pixeltable/catalog/table_version.py
```diff
@@ -704,15 +705,34 @@ def _insert(
         return result
 
     def update(
-            self, update_targets: dict[Column, 'pixeltable.exprs.Expr'],
-            where_clause: Optional['pixeltable.exprs.Predicate'] = None, cascade: bool = True
+        self, value_spec: dict[str, Any], where: Optional['exprs.Predicate'] = None, cascade: bool = True
     ) -> UpdateStatus:
+        """Update rows in this TableVersionPath.
```
mkornacker, line 710: Update comment

pixeltable/catalog/table_version.py
```diff
@@ -768,24 +787,61 @@ def _update(
...
+                if not col.col_type.matches(value_expr.col_type):
+                    raise excs.Error((
+                        f'Type of value {val!r} ({value_expr.col_type}) is not compatible with the type of column '
+                        f'{col_name} ({col.col_type})'
+                    ))
+            update_targets[col] = value_expr
+
+        return update_targets
+
+    def propagate_update(
```
mkornacker, line 844: This can be private again

pixeltable/catalog/table_version.py
```diff
@@ -819,22 +875,31 @@ def _propagate_update(
...
         with Env.get().engine.begin() as conn:
-            num_rows = self._delete(analysis_info.sql_where_clause, base_versions=[], conn=conn, timestamp=time.time())
+            num_rows = self.propagate_delete(analysis_info.sql_where_clause, base_versions=[], conn=conn, timestamp=time.time())
 
         status = UpdateStatus(num_rows=num_rows)
         return status
 
-    def _delete(
-            self, where: Optional['pixeltable.exprs.Predicate'], base_versions: List[Optional[int]],
+    def propagate_delete(
```
mkornacker, line 901: This can also be private.

mkornacker (approved):

PR #211, opened 2024-07-02
Concise formatting for strings, floats, arrays, & json

aaron-siegel:
Implements concise formatting for strings, floats, arrays, and json structs. This is based on @orm011 's PR https://github.com/pixeltable/pixeltable/pull/177 but with some significant differences.

- limit floats to 3 digits of precision
- escape strings for both HTML and MathJax
- abbreviate strings to 1000 characters for StringType, 300 characters if nested inside JsonType
- no multi-line rendering, attempted alignments, or font changes (this is deliberate)

RAG demo and Object Detection demo both look significantly cleaner now.

Also fixes a bug in the Object Detection demo.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/211)

pixeltable/utils/formatter.py
```diff
@@ -0,0 +1,211 @@
...
+
+_FLOAT_PRECISION = 3
+_LIST_THRESHOLD = 16
+_LIST_EDGEITEMS = 6
+_STRING_SEP = ' ...... '
+_STRING_MAX_LEN = 1000
+_NESTED_STRING_MAX_LEN = 300
+
+
+class PixeltableFormatter:
```
mkornacker: Can we just call this Formatter, given that it's already part of Pixeltable? PixeltableFormatter is pretty long.

Also: missing class comment (ie, what's the scope of what belongs in here). Why rows and columns and an http address?

pixeltable/utils/formatter.py
```diff
@@ -0,0 +1,211 @@
...
+import cv2
+import numpy as np
+
+import io
+import pixeltable.type_system as ts
+from pixeltable.utils.http_server import get_file_uri
+
+_logger = logging.getLogger('pixeltable')
+
+_FLOAT_PRECISION = 3
```
mkornacker: Any reason not to move those into the formatter class?

pixeltable/utils/formatter.py
```diff
@@ -0,0 +1,211 @@
...
+                components.extend(cls.__format_json_rec(x) for x in val[-_LIST_EDGEITEMS:])
+            return '[' + ', '.join(components) + ']'
+        if isinstance(val, dict):
+            kv_pairs = (f'{cls.__format_json_rec(k)}: {cls.__format_json_rec(v)}' for k, v in val.items())
+            return '{' + ', '.join(kv_pairs) + '}'
+        return json.dumps(val)
+
+    def format_img(self, img: Image.Image) -> str:
+        """
+        Create <img> tag for Image object.
```
mkornacker, line 136: So this class is an html formatter.

pixeltable/utils/formatter.py
```diff
@@ -0,0 +1,211 @@
...
+                <img src="data:image/jpeg;base64,{img_base64}" width="{width}" />
+            </div>
+            """
+
+    def format_video(self, file_path: str) -> str:
+        thumb_tag = ''
+        # Attempt to extract the first frame of the video to use as a thumbnail,
+        # so that the notebook can be exported as HTML and viewed in contexts where
+        # the video itself is not accessible.
+        # TODO(aaron-siegel): If the video is backed by a concrete external URL,
```
mkornacker, line 160: Why add your name? Anyone should be able to work on that (or feel compelled to).

pixeltable/utils/formatter.py
```diff
@@ -0,0 +1,211 @@
...
+        # should we link to that instead?
```
mkornacker, line 161: Why is that useful?

tests/test_formatter.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_format(self, init_env):
+        formatter = PixeltableFormatter(10, 10, Env.get().http_address)
+
+        assert formatter.format_float(0.4171780) == '0.417'
+        assert formatter.format_float(1401.19018) == '1401.19'
+
+        # Test HTML and MathJax escaping
+        test_string = '<div>"bacon & eggs", $9\n</div>'
+        escaped = '&lt;div&gt;&quot;bacon &amp; eggs&quot;, \\$9\n&lt;/div&gt;'
```
mkornacker, line 18: Is there some easy way to produce the escaped string, eg, via a standard library? That would be more reassuring than hardwiring it here.

pixeltable/utils/formatter.py
```diff
@@ -0,0 +1,211 @@
...
+import pixeltable.type_system as ts
+from pixeltable.utils.http_server import get_file_uri
+
+_logger = logging.getLogger('pixeltable')
+
+_FLOAT_PRECISION = 3
+_LIST_THRESHOLD = 16
+_LIST_EDGEITEMS = 6
+_STRING_SEP = ' ...... '
+_STRING_MAX_LEN = 1000
```
mkornacker: 1000? Is that practical? Why not 100?

tests/test_formatter.py
```diff
@@ -0,0 +1,36 @@
...
+        long_string = "abcdefghijklm" * 77
+        assert len(long_string) == 1001
+        result = formatter.format_string(long_string)
+        assert len(result) == 1000
+        assert result == f'{long_string[:496]} ...... {long_string[-496:]}'
+        assert result == formatter.format_json(long_string)
+
+        # Test a JSON list
+        items = [0.4171780, test_string, long_string, 84, True]
+        escaped_json = '&lt;div&gt;\\&quot;bacon &amp; eggs\\&quot;, \\$9\\n&lt;/div&gt;'
```
mkornacker, line 35: Comment: why is this different than line 15?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/211)* status: 0 of 6 files reviewed, 8 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/utils/formatter.py` line 19 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662765327:-O0oCZqcEBVs6AAP9h-p:b-7yhbno) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L19)):*

Any reason not to move those into the formatter class?
</blockquote></details>

They're global constants, not instance variables.
___
*[`pixeltable/utils/formatter.py` line 23 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662778874:-O0oCATr1I2nwRNVCf4c:b-44i8kr) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L23)):*

1000? Is that practical? Why not 100?
</blockquote></details>

Yes, I tested it with various values and 1000 looks great in notebooks such as the RAG demo. Smaller values hide too much in use cases such as document chunks. Note this only applies to standalone strings; strings nested within lists/dicts will use `_NESTED_STRING_MAX_LEN` instead.

We could make this column-dependent since we have access to num_cols (shorten strings more if num_cols > 2, say). But my main goal here is "better than before" and we can always iterate on it later if we don't like what we see.

Check out the RAG demo if you want to see what the 1000-len strings look like. I think they're just right.
___
*[`pixeltable/utils/formatter.py` line 27 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662759124:-O0oC96S5kkmFndZS1fv:bo22vtg) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L27)):*

Can we just call this Formatter, given that it's already part of Pixeltable? PixeltableFormatter is pretty long.

Also: missing class comment (ie, what's the scope of what belongs in here). Why rows and columns and an http address?
</blockquote></details>

Sure.

Rows/columns are used to size the image/video embeds dynamically. http_address points to the Pixeltable http server, so that URLs can be constructed for proper media references. (None of that functionality is new to this PR - it was merely refactored here)

I get the "missing class comment" feedback pretty often, but there isn't any consistency around class comments for non-user-facing classes in the existing codebase. `DataFrame` and `DataFrameResultSet` don't have any! I think it'd be useful to document our expectations around practices for this sort of thing ("Pixeltable coding practices" doc or some such), particularly in cases where the existing codebase isn't a suitable model.
___
*[`pixeltable/utils/formatter.py` line 113 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662768129:-O0oGPY13PUQDMSi_6Cg:b-49lg5q) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L113)):*

So this class is an html formatter.
</blockquote></details>

yes, I clarified that in the docstring.
___
*[`pixeltable/utils/formatter.py` line 137 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662770213:-O0oCgKq0xU23pU2mMXX:b-62mg85) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L137)):*

Why add your name? Anyone should be able to work on that (or feel compelled to).
</blockquote></details>

Standard best practice for TODOs. It doesn't mean "only aaron-siegel can work on it". It means "aaron-siegel left this comment here". Most linting tools enforce this by default (you'll get a linting error if there's a TODO without a name).
___
*[`pixeltable/utils/formatter.py` line 138 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662770760:-O0oCwYwCT_wnVbKoajB:bd8v9dt) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L138)):*

Why is that useful?
</blockquote></details>

So the video will actually show up in the notebook if it's shared outside of the user's deployment. (Right now you get just a thumbnail)

Note that none of the video/audio/image stuff is new to this PR; it's been around for awhile and was merely refactored here. The string/float/JSON/array stuff is new.
___
*[`tests/test_formatter.py` line 15 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662777073:-O0oD92uAEZB49nvFxIs:b-a7ayfj) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/tests/test_formatter.py#L15)):*

Is there some easy way to produce the escaped string, eg, via a standard library? That would be more reassuring than hardwiring it here.
</blockquote></details>

Of course there is (that's how it's getting produced in the code) but the whole point is to verify that the libraries are doing the thing we expect. e.g., if we upgrade the libraries at some point and the behavior changes unexpectedly, then the test should notify us of that.
___
*[`tests/test_formatter.py` line 30 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662781127:-O0oDT22EA_zX7DGikZI:b-bttoee) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/tests/test_formatter.py#L30)):*

Comment: why is this different than line 15?
</blockquote></details>

Added comment

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/211)* status: 0 of 6 files reviewed, 8 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/utils/formatter.py` line 19 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662765327:-O0oRbG52-A8tSphp_zo:b76eibg) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L19)):*

They're global constants, not instance variables.
</blockquote></details>

I get that, but do they need to be accessible outside the class?
___
*[`pixeltable/utils/formatter.py` line 27 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662759124:-O0oRgDG06ZHBiDg1GSa:b6i5bzm) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L27)):*

Sure.

Rows/columns are used to size the image/video embeds dynamically. http_address points to the Pixeltable http server, so that URLs can be constructed for proper media references. (None of that functionality is new to this PR - it was merely refactored here)

I get the "missing class comment" feedback pretty often, but there isn't any consistency around class comments for non-user-facing classes in the existing codebase. `DataFrame` and `DataFrameResultSet` don't have any! I think it'd be useful to document our expectations around practices for this sort of thing ("Pixeltable coding practices" doc or some such), particularly in cases where the existing codebase isn't a suitable model.
</blockquote></details>

When I'm asking for clarification in a review comment, I usually mean for that to happen in a comment in the code (not in the review response).

Agreed we need to come up with some "code of conduct" guidelines for the code.
___
*[`pixeltable/utils/formatter.py` line 137 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662770213:-O0oSC-l8yZZHFyFYEXI:b-ixljdv) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L137)):*

Standard best practice for TODOs. It doesn't mean "only aaron-siegel can work on it". It means "aaron-siegel left this comment here". Most linting tools enforce this by default (you'll get a linting error if there's a TODO without a name).
</blockquote></details>

We have 'blame' for code attribution. It's fine to leave it, but it doesn't feel like something useful or that needs enforcement.

pixeltable/utils/formatter.py
```diff
@@ -0,0 +1,224 @@
...
+
+class PixeltableFormatter:
+    """
+    A factory for constructing Pandas HTML formatters for Pixeltable data. The formatters are used to customize
+    the rendering of `DataFrameResultSet`s in notebooks.
+
+    Args:
+        num_rows: Number of rows in the DataFrame being rendered.
+        num_cols: Number of columns in the DataFrame being rendered.
+        http_address: Root address of the Pixeltable HTTP server (used to construct URLs for media references).
```
mkornacker, line 28: Should this not come from Env?

pixeltable/utils/formatter.py
```diff
@@ -0,0 +1,224 @@
...
+_LIST_THRESHOLD = 16
+_LIST_EDGEITEMS = 6
+_STRING_SEP = ' ...... '
+_STRING_MAX_LEN = 1000
+_NESTED_STRING_MAX_LEN = 300
+
+
+class PixeltableFormatter:
+    """
+    A factory for constructing Pandas HTML formatters for Pixeltable data. The formatters are used to customize
```
mkornacker: If the scope is explicitly pandas, why not call this PandasFormatter (and get_formatter())?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/211)* status: 0 of 6 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/utils/formatter.py` line 27 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662759124:-O0oWDNa7zqfQF7O0Gm_:b-n2il0e) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L27)):*

When I'm asking for clarification in a review comment, I usually mean for that to happen in a comment in the code (not in the review response).

Agreed we need to come up with some "code of conduct" guidelines for the code.
</blockquote></details>

Right, I added it to the code too
___
*[`pixeltable/utils/formatter.py` line 29 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662907722:-O0oWFdJ2epVYugR-SWf:b32w1t) ([raw file](https://github.com/pixeltable/pixeltable/blob/2720bcc0a5c4191f56823981c3c13f4b16c81be3/pixeltable/utils/formatter.py#L29)):*

If the scope is explicitly pandas, why not call this PandasFormatter (and get_formatter())?
</blockquote></details>

Ok
___
*[`pixeltable/utils/formatter.py` line 35 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662906781:-O0oWHD1D58d_4FlRvNB:b-vtcyay) ([raw file](https://github.com/pixeltable/pixeltable/blob/2720bcc0a5c4191f56823981c3c13f4b16c81be3/pixeltable/utils/formatter.py#L35)):*

Should this not come from Env?
</blockquote></details>

It does (it gets passed here from Env in the DataFrameResultSet); I don't see a reason not to parameterize it though

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/211)* status: 0 of 6 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/utils/formatter.py` line 19 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662765327:-O0oXN3V8gCyNg8IOsg9:b-ybsfo1) ([raw file](https://github.com/pixeltable/pixeltable/blob/11a1cc5824e08a6d849a3d273cdb2329700bd893/pixeltable/utils/formatter.py#L19)):*

I get that, but do they need to be accessible outside the class?
</blockquote></details>

Fixed this

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/211)* status: 0 of 6 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/utils/formatter.py` line 29 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/211#gh-1662907722:-O0oagpR5SQRsryvxk68:bt6cor9) ([raw file](https://github.com/pixeltable/pixeltable/blob/2720bcc0a5c4191f56823981c3c13f4b16c81be3/pixeltable/utils/formatter.py#L29)):*

Ok
</blockquote></details>

Actually, I think much of the stuff in here is useful more generally. So I'm going to stick with `Formatter` / `get_pandas_formatter`. I also refactored the tests to exercise the `get_pandas_formatter` code path.

PR #216, opened 2024-07-17
API changes to `Table` and `SchemaObject`

aaron-siegel:
- Renames `Table.get_views()` to `Table.list_views()`, returning strings instead of objects
- Renames `SchemaObject.get_name()` to `SchemaObject.name` and converts it to a property
- Adds `SchemaObject.parent` and `SchemaObject.path`


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/216)

pixeltable/catalog/table.py
```diff
@@ -82,14 +82,22 @@ def __getitem__(
...
 
-    def get_views(self, *, recursive: bool = False) -> list['Table']:
+    def list_views(self, *, recursive: bool = True) -> list[str]:
         """
-        All views and snapshots of this `Table`.
+        Returns a list of all views and snapshots of this `Table`.
+
+        Args:
+            recursive: If `False`, returns only the immediate successor views of this `Table`. If `True`, returns
+                all sub-views, including this `Table` itself.
```
mkornacker: I don't know if anyone would expect to get the table itself back.

pixeltable/catalog/table.py
```diff
@@ -82,14 +82,22 @@ def __getitem__(
...
         """
+        return [t.path for t in self._get_views(recursive=recursive)]
+
+    def _get_views(self, *, recursive: bool = True) -> list['Table']:
+        dependents = catalog.Catalog.get().tbl_dependents[self._get_id()]
         if recursive:
-            return [self] + [t for view in self.get_views(recursive=False) for t in view.get_views(recursive=True)]
+            return [self] + [t for view in dependents for t in view._get_views(recursive=True)]
```
mkornacker: Let's not add self here.

pixeltable/catalog/table_version.py
```diff
@@ -557,7 +557,7 @@ def drop_column(self, name: str) -> None:
             )
         # See if this column has a dependent store. We need to look through all stores in all
         # (transitive) views of this table.
-        transitive_views = Catalog.get().tbls[self.id].get_views(recursive=True)
+        transitive_views = Catalog.get().tbls[self.id]._get_views(recursive=True)
```
mkornacker: It doesn't really make sense for TableVersion to be interacting with Table, the latter is only meant as the user-facing interface to tables. get_views() needs to be in TableVersion, not Table.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/216)* status: 0 of 14 files reviewed, 3 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/table.py` line 91 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/216#gh-1681971775:-O22hcpF0z3Xf2LeZ9u0:b-ze9akn) ([raw file](https://github.com/pixeltable/pixeltable/blob/cab14c0af1a1edad12a358a3c13d1c6f5e5d8b60/pixeltable/catalog/table.py#L91)):*

I don't know if anyone would expect to get the table itself back.
</blockquote></details>

I certainly would!
___
*[`pixeltable/catalog/table.py` line 98 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/216#gh-1681976744:-O22h_nn5WZHgfrmd_mS:b-xejx1n) ([raw file](https://github.com/pixeltable/pixeltable/blob/cab14c0af1a1edad12a358a3c13d1c6f5e5d8b60/pixeltable/catalog/table.py#L98)):*

Let's not add self here.
</blockquote></details>

As a mathematician, I'm appalled :)

It's not the hill I want to die on, but I do want to discuss and understand the reasoning here.
___
*[`pixeltable/catalog/table_version.py` line 560 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/216#gh-1681980504:-O22hopS-hGJzVI0h4A2:bjdxv4d) ([raw file](https://github.com/pixeltable/pixeltable/blob/cab14c0af1a1edad12a358a3c13d1c6f5e5d8b60/pixeltable/catalog/table_version.py#L560)):*

It doesn't really make sense for TableVersion to be interacting with Table, the latter is only meant as the user-facing interface to tables. get_views() needs to be in TableVersion, not Table.
</blockquote></details>

Ok. This isn't new to this PR though. It's been in there for awhile. The only thing this PR does is rename the method. Should it be a blocking issue to address it here?

Do you have an opinion on the right way to do this resolution?

mkornacker (approved):

PR #219, opened 2024-07-18
Update OpenAI models used in tests and notebooks; various other housecleaning

aaron-siegel:
Various housecleaning:
- Update tests and notebooks to use `gpt-4o-mini` OpenAI model instead of `gpt-3.5-turbo` and `gpt-4-vision-preview`
- Resurrect GPT-4 vision test
- Update notebooks to use `string_embed` instead of `text_embed`
- Use new `drop_dir(force=True)` pattern in all notebooks that were modified, instead of dropping tables individually


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/219)

mkornacker (approved):

PR #223, opened 2024-07-22
Basic Data Manipulation

pierrebrunelle:
- Adding a new "fundamentals" tutorial.
- The last section regarding Views and Snapshots could be removed if tackled in a separate tutorial and not used as gentle transition to that other notebook.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/223)

PR #237, opened 2024-08-02
Allow method call and property access syntax on arbitrary types (not just images)

aaron-siegel:
Enables method/property references on arbitrary Pixeltable expressions, not just expressions of type `IMAGE`. For example:

- `t.str_col.contains('x')` (method access)
- `t.float_col.floor` (property access)

UDFs must be explicitly registered for method/property access via the `member_access` parameter of the `@udf` decorator.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/237)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/237)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/func/callable_function.py` line 28 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/237#-O3K_hHD5h11mruXr3pt:-O3K_hHD5h11mruXr3pu:b2eixmd) ([raw file](https://github.com/pixeltable/pixeltable/blob/db6923995b6e6d814c6e465e1869c5637f84979f/pixeltable/func/callable_function.py#L28)):*
> ```Python
>         self_name: Optional[str] = None,
>         batch_size: Optional[int] = None,
>         allow_member_access: bool = False,
> ```

Maybe just is_member_fn or something like that? allow_member_access is a bit mysterious. (and Python class member fns can always be called in a standard pattern)
___
*[`pixeltable/func/callable_function.py` line 102 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/237#-O3Kbs2t4DJaZd9pWmW9:-O3Kbs2t4DJaZd9pWmWA:b-dxsugf) ([raw file](https://github.com/pixeltable/pixeltable/blob/db6923995b6e6d814c6e465e1869c5637f84979f/pixeltable/func/callable_function.py#L102)):*
> ```Python
>     def to_store(self) -> tuple[dict, bytes]:
>         md = {
>             'signature': self.signature.as_dict(),
> ```

This needs to capture the new metadata.
___
*[`pixeltable/func/function_registry.py` line 136 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/237#-O3KcJWLBJXdIYjyHxqs:-O3KcJWLBJXdIYjyHxqt:b-99ghio) ([raw file](https://github.com/pixeltable/pixeltable/blob/db6923995b6e6d814c6e465e1869c5637f84979f/pixeltable/func/function_registry.py#L136)):*
> ```Python
>     #         return self.module_fns[fqn]
> 
>     def lookup_type_method(self, base_type: ts.ColumnType.Type, name: str) -> Optional[Function]:
> ```

Brief docstring to explain when it returns None
___
*[`pixeltable/func/udf.py` line 29 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/237#-O3KcklE0LCCtJKmWny4:-O3KcklE0LCCtJKmWny5:b5dodmu) ([raw file](https://github.com/pixeltable/pixeltable/blob/db6923995b6e6d814c6e465e1869c5637f84979f/pixeltable/func/udf.py#L29)):*
> ```Python
>         batch_size: Optional[int] = None,
>         substitute_fn: Optional[Callable] = None,
>         member_access: Optional[str] = None,
> ```

I think having two bool parameters (is_member, is_property) is a bit easier to grasp.

pixeltable/exprs/method_ref.py
```diff
@@ -0,0 +1,65 @@
...
+import pixeltable.type_system as ts
+from pixeltable.exprs import Expr, FunctionCall
+from pixeltable.func import FunctionRegistry, CallableFunction
+from .data_row import DataRow
+from .row_builder import RowBuilder
+
+
+class MethodRef(Expr):
+    """
+    A method reference. This represents access to an unevaluated method of a Pixeltable expression, such as
```
mkornacker: That sounds misleading - the expression classes have their own methods, which have nothing to do with our Functions. I think omitting that sentence makes this comment easier to follow.

pixeltable/exprs/method_ref.py
```diff
@@ -0,0 +1,65 @@
...
+    `t.img_col.rotate` or `t.str_col.contains`. Equivalently, it represents a `Function` instance with its
+    first parameter bound to a base expression.
+
+    When a `MethodRef` is called, it returns a `FunctionCall` with the base expression as the first argument.
```
mkornacker, line 18: I'm pretty opposed to backticks around references to classes in code comments. There is no precedent for that in this code base or any other one that I've ever worked on, and it doesn't make the code more readable and this isn't part of any externally-facing docs.

pixeltable/exprs/method_ref.py
```diff
@@ -0,0 +1,65 @@
...
+    The effective arity of a `MethodRef` is one less than the arity of the underlying `Function`.
```
mkornacker, line 19: Feel free to add a 'todo: should this even be an Expr?'

pixeltable/exprs/method_ref.py
```diff
@@ -0,0 +1,65 @@
...
+        return self.base_expr == other.base_expr and self.method_name == other.method_name
+
+    def _id_attrs(self) -> list[tuple[str, Any]]:
+        return super()._id_attrs() + [('method_name', self.method_name)]
+
+    def sql_expr(self) -> Optional[sql.ClauseElement]:
+        return None
+
+    def eval(self, data_row: DataRow, row_builder: RowBuilder) -> None:
+        base_val = data_row[self.base_expr.slot_idx]
```
mkornacker: Why is this necessary? I would expect that only the actual function call gets evaluated, but never this.

pixeltable/func/function.py
```diff
@@ -19,9 +21,13 @@ class Function(abc.ABC):
     via the member self_path.
     """
 
-    def __init__(self, signature: Signature, self_path: Optional[str] = None):
+    def __init__(self, signature: Signature, self_path: Optional[str] = None, is_method: bool = False, is_property: bool = False):
+        if (is_method or is_property) and self_path is None:
+            raise excs.Error('Stored functions cannot be declared using `is_method` or `is_property`')
```
mkornacker: Shouldn't this be checked in the decorator (because the error message is referencing parameter names of the decorator)?

pixeltable/func/function_registry.py
```diff
@@ -65,10 +64,17 @@ def clear_cache(self) -> None:
     #             fn_path = f'{caller_path}.{name}'  # fully-qualified name
     #             self.module_fns[fn_path] = obj
 
-    def register_function(self, fqn: str, fn: Function) -> None:
+    def register_function(self, fqn: str, fn: Function, register_as_method: bool) -> None:
```
mkornacker: Is register_as_method not simply (fn.is_method or fn.is_property)?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/237)* status: 0 of 13 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/func/callable_function.py` line 28 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/237#-O3K_hHD5h11mruXr3pt:-O3YoKPpCSEe7A1T3XB2:b-zcia08) ([raw file](https://github.com/pixeltable/pixeltable/blob/db6923995b6e6d814c6e465e1869c5637f84979f/pixeltable/func/callable_function.py#L28)):*

Maybe just is_member_fn or something like that? allow_member_access is a bit mysterious. (and Python class member fns can always be called in a standard pattern)
</blockquote></details>

Changed to `is_method`.
___
*[`pixeltable/func/callable_function.py` line 102 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/237#-O3Kbs2t4DJaZd9pWmW9:-O3YoPopAxda48BnHoXP:bj8b80j) ([raw file](https://github.com/pixeltable/pixeltable/blob/db6923995b6e6d814c6e465e1869c5637f84979f/pixeltable/func/callable_function.py#L102)):*

This needs to capture the new metadata.
</blockquote></details>

I simply prevented stored functions from using method call syntax (at least for now - until we think through the implications some more).
___
*[`pixeltable/func/function.py` line 26 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/237#gh-1703001928:-O3YobsSCjlF2ggaQzr3:b-dab5kn) ([raw file](https://github.com/pixeltable/pixeltable/blob/aeab67a1551d7521d0f1acbfc879f48fc5ad6792/pixeltable/func/function.py#L26)):*

Shouldn't this be checked in the decorator (because the error message is referencing parameter names of the decorator)?
</blockquote></details>

It seemed safest to do it here? Just in case Functions get instantiated some way other than through the decorator?
___
*[`pixeltable/func/function_registry.py` line 136 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/237#-O3KcJWLBJXdIYjyHxqs:-O3YowKp2BGpsR4UbaCy:bb6un1j) ([raw file](https://github.com/pixeltable/pixeltable/blob/db6923995b6e6d814c6e465e1869c5637f84979f/pixeltable/func/function_registry.py#L136)):*

Brief docstring to explain when it returns None
</blockquote></details>

Done
___
*[`pixeltable/func/function_registry.py` line 67 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/237#gh-1703003129:-O3YovA0AnxO9CpDymsO:b-47vjvg) ([raw file](https://github.com/pixeltable/pixeltable/blob/aeab67a1551d7521d0f1acbfc879f48fc5ad6792/pixeltable/func/function_registry.py#L67)):*

Is register_as_method not simply (fn.is_method or fn.is_property)?
</blockquote></details>

Good call, fixed
___
*[`pixeltable/func/udf.py` line 29 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/237#-O3KcklE0LCCtJKmWny4:-O3YoxCG724-oBoaGGdp:bb6un1j) ([raw file](https://github.com/pixeltable/pixeltable/blob/db6923995b6e6d814c6e465e1869c5637f84979f/pixeltable/func/udf.py#L29)):*

I think having two bool parameters (is_member, is_property) is a bit easier to grasp.
</blockquote></details>

Done
___
*[`pixeltable/exprs/method_ref.py` line 15 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/237#gh-1702998575:-O3YnaoVERfpa7T0wSo1:b-bt85em) ([raw file](https://github.com/pixeltable/pixeltable/blob/aeab67a1551d7521d0f1acbfc879f48fc5ad6792/pixeltable/exprs/method_ref.py#L15)):*

That sounds misleading - the expression classes have their own methods, which have nothing to do with our Functions. I think omitting that sentence makes this comment easier to follow.
</blockquote></details>

Ok I took it out.
___
*[`pixeltable/exprs/method_ref.py` line 19 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/237#gh-1702998788:-O3Ymw9v7saSp66-LPEM:b9e9vwe) ([raw file](https://github.com/pixeltable/pixeltable/blob/aeab67a1551d7521d0f1acbfc879f48fc5ad6792/pixeltable/exprs/method_ref.py#L19)):*

I'm pretty opposed to backticks around references to classes in code comments. There is no precedent for that in this code base or any other one that I've ever worked on, and it doesn't make the code more readable and this isn't part of any externally-facing docs.
</blockquote></details>

It's false that there's no precedent. Check out, for example, right here in the python SDK:

```python
import json
help(json.dumps)
```

It's the general practice I've used in the past for docstrings, and personally I do think it makes them more readable, and it ensures consistency with best markdown practices for code documentation.

Can we factor this out as a more general discussion topic and not try to resolve it in this PR?
___
*[`pixeltable/exprs/method_ref.py` line 20 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/237#gh-1702999429:-O3YnvKf6k4HeIjTAAEA:bb6un1j) ([raw file](https://github.com/pixeltable/pixeltable/blob/aeab67a1551d7521d0f1acbfc879f48fc5ad6792/pixeltable/exprs/method_ref.py#L20)):*

Feel free to add a 'todo: should this even be an Expr?'
</blockquote></details>

Done
___
*[`pixeltable/exprs/method_ref.py` line 58 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/237#gh-1703000931:-O3YnzsX0z755WTj3saL:b-si7sgk) ([raw file](https://github.com/pixeltable/pixeltable/blob/aeab67a1551d7521d0f1acbfc879f48fc5ad6792/pixeltable/exprs/method_ref.py#L58)):*

Why is this necessary? I would expect that only the actual function call gets evaluated, but never this.
</blockquote></details>

It's not. I just cargo-culted it from the old `ImageMemberAccess` without really thinking about it. I took it out

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/237)* status: 0 of 13 files reviewed, 10 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/func/function.py` line 26 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/237#gh-1703001928:-O3Yyb_uBmKfNHuvu0Yq:b-7zdxgo) ([raw file](https://github.com/pixeltable/pixeltable/blob/aeab67a1551d7521d0f1acbfc879f48fc5ad6792/pixeltable/func/function.py#L26)):*

It seemed safest to do it here? Just in case Functions get instantiated some way other than through the decorator?
</blockquote></details>

I agree this should be checked here. But I think the error should get raised in the function called by the user (udf()) and this should just assert.
___
*[`pixeltable/exprs/method_ref.py` line 19 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/237#gh-1702998788:-O3YupiT2S62j0V0gN8f:bzbq2v9) ([raw file](https://github.com/pixeltable/pixeltable/blob/aeab67a1551d7521d0f1acbfc879f48fc5ad6792/pixeltable/exprs/method_ref.py#L19)):*

It's false that there's no precedent. Check out, for example, right here in the python SDK:

```python
import json
help(json.dumps)
```

It's the general practice I've used in the past for docstrings, and personally I do think it makes them more readable, and it ensures consistency with best markdown practices for code documentation.

Can we factor this out as a more general discussion topic and not try to resolve it in this PR?
</blockquote></details>

Okay, but then take it out of this one. It's not part of the code base as it is today.

MethodRef is not something that the user will ever consciously interact with, so I don't think json.dumps is a good analogy.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/237)* status: 0 of 13 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/func/function.py` line 26 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/237#gh-1703001928:-O3ZmJXu5c1qPJ0kzQVQ:b-b7zlvk) ([raw file](https://github.com/pixeltable/pixeltable/blob/aeab67a1551d7521d0f1acbfc879f48fc5ad6792/pixeltable/func/function.py#L26)):*

I agree this should be checked here. But I think the error should get raised in the function called by the user (udf()) and this should just assert.
</blockquote></details>

Ah, that makes sense. Done.

mkornacker (approved):

PR #239, opened 2024-08-04
Timestamp functions plus image function docstrings

mkornacker:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/239)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/239)* status: 0 of 11 files reviewed, 7 unresolved discussions (waiting on @mkornacker)

___
*[`mkdocs.yml` line 34 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3YqSqH-mwe5RPgT4bu:-O3YqSqH-mwe5RPgT4bv:bc7oi7x) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/mkdocs.yml#L34)):*
> ```YAML
>   - Pixeltable Functions:
>     - Huggingface: api/functions/huggingface.md
>     - ImageType: api/functions/image.md
> ```

I'd favor using the package names `image`, `timestamp`, etc for section headings, rather than `ImageType` and `TimestampType` etc.
___
*[`pixeltable/functions/string.py` line 598 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3Yqrjc4NOq9O7sfN9B:-O3Yqrjc4NOq9O7sfN9C:bukx6wx) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/pixeltable/functions/string.py#L598)):*
> ```Python
> 
> @func.udf
> def zfill(self: str, width: int) -> str:
> ```

I think I was mistaken, and these don't actually need to be called `self` to pass muster with mkdocs, so go with whatever you think makes the most sense
___
*[`pixeltable/functions/timestamp.py` line 34 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3Yr1Oe5aWPSurMVdyB:-O3Yr1Oe5aWPSurMVdyC:bye1oms) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/pixeltable/functions/timestamp.py#L34)):*
> ```Python
> def month(self: datetime) -> int:
>     """
>     Between 1 and 12 inclusive.
> ```

I think this would be clearer if it stated what the property actually is, e.g.,

The month component of the given datetime, between 1 and 12 (inclusive).
___
*[`pixeltable/functions/timestamp.py` line 157 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3YrYvX1c-ytzIe-2Ez:-O3YrYvX1c-ytzIe-2F-:b-qc7d5j) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/pixeltable/functions/timestamp.py#L157)):*
> ```Python
>     """
>     d = self.date()
>     return datetime(d.year, d.month, d.day)
> ```

It's generally fraught to contrive dates into special cases of datetimes; this is why python keeps them separate. Pixeltable's type system allows either; shouldn't we take advantage of that and just return self.date() here?

I think there's a case to be made for a Pixeltable Date type, separate from Timestamp. Particularly if we want to handle timeseries data in the future, there's a lot of value to getting this right.

In general there are four distinct entities one encounters, that are all different, and we should think through how we want to handle/represent them in pixeltable:
1. Absolute timestamp (= a number of microseconds since the epoch)
2. A date
3. A time of day
4. A wall clock datetime (tuple of 2 & 3)

(1) and (4) differ because (4) is sensitive to time zone and (1) is not. The absolute timestamp boundaries of (2) are timezone sensitive and are not necessarily 24 hours apart (due to daylight savings adjustments). A lot of datetime libraries get tripped up by these types of considerations. I'm not proposing that we need 4 different types in Pixeltable, necessarily, but we should think through this.
___
*[`pixeltable/functions/timestamp.py` line 163 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3Ys4Z8ATUuV71GWGWe:-O3Ys4Z9CZ9FtY_7CmaL:b-gnxflg) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/pixeltable/functions/timestamp.py#L163)):*
> ```Python
> def time(self: datetime) -> datetime:
>     """
>     Return the time part of the datetime, with microseconds set to 0.
> ```

Doesn't look like microseconds are set to 0 to me
___
*[`pixeltable/functions/timestamp.py` line 168 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3Yrv0xBf395-5Xy_eo:-O3Yrv0xBf395-5Xy_ep:bcjzkn2) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/pixeltable/functions/timestamp.py#L168)):*
> ```Python
>     """
>     t = self.time()
>     return datetime(1, 1, 1, t.hour, t.minute, t.second, t.microsecond)
> ```

Same concern as above; is this a postgres thing to mock up times as datetimes on 1/1/1?
___
*[`tests/functions/test_timestamp.py` line 16 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3YsTrT3VyZF7n4GUxq:-O3YsTrT3VyZF7n4GUxr:bqoirj1) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/tests/functions/test_timestamp.py#L16)):*
> ```Python
>     ]
> 
>     def test_all(self, reset_db) -> None:
> ```

More descriptive name here? `test_timestamp_udfs`?

aaron-siegel (changes requested):
I don't see any mention of timezones in this PR. How are they handled in Pixeltable? Is it always UTC or does system timezone factor in some how?

Some libraries/database systems default to system timezone which can be very problematic - I always favor either using UTC or being explicit about timezone.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/239#-:-O3YvFnC0kvL3gffFp78:b7u6jdz)* status: 0 of 11 files reviewed, 9 unresolved discussions (waiting on @mkornacker)

___
*[`tests/functions/test_timestamp.py` line 21 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3YupKxChQi5chKSEjo:-O3YupKxChQi5chKSEjp:b-h7mdz7) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/tests/functions/test_timestamp.py#L21)):*
> ```Python
>         validate_update_status(t.insert({'dt': dt} for dt in test_dts), expected_rows=len(test_dts))
> 
>         from pixeltable.functions.timestamp import (
> ```

Easier to just do `from pixeltable.functions.timestamp import *` here?
___
*[`tests/functions/test_timestamp.py` line 51 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3Yvdi0BB6uOR2aNt9_:-O3Yvdi0BB6uOR2aNt9a:b-qci31c) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/tests/functions/test_timestamp.py#L51)):*
> ```Python
> 
>         for pxt_fn, dt_fn, args, kwargs in test_params:
>             assert t.select(out=pxt_fn(t.dt, *args, **kwargs)).collect()['out'] ==\
> ```

Random style note, I believe best practice is not to use `\`, but to enclose in parens. that is, instead of
```python
assert a == \
    b
```
it's preferred to do
```python
assert (a ==
    b)
```

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/239)* status: 0 of 11 files reviewed, 9 unresolved discussions (waiting on @aaron-siegel)

___
*[`mkdocs.yml` line 34 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3YqSqH-mwe5RPgT4bu:-O3YzdCy3qbDKkfFBjmd:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/mkdocs.yml#L34)):*

I'd favor using the package names `image`, `timestamp`, etc for section headings, rather than `ImageType` and `TimestampType` etc.
</blockquote></details>

Done.
___
*[`pixeltable/functions/string.py` line 598 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3Yqrjc4NOq9O7sfN9B:-O3YzfO8416c6lsVtDAT:baq2bcf) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/pixeltable/functions/string.py#L598)):*

I think I was mistaken, and these don't actually need to be called `self` to pass muster with mkdocs, so go with whatever you think makes the most sense
</blockquote></details>

No, you weren't.
___
*[`pixeltable/functions/timestamp.py` line 34 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3Yr1Oe5aWPSurMVdyB:-O3YzwDy-iK21V48PJL0:b-w56n4f) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/pixeltable/functions/timestamp.py#L34)):*

I think this would be clearer if it stated what the property actually is, e.g.,

The month component of the given datetime, between 1 and 12 (inclusive).
</blockquote></details>

This is simply copying from the official Python docs, which is presumably good enough.

https://docs.python.org/3/library/datetime.html#datetime.datetime.month
___
*[`pixeltable/functions/timestamp.py` line 157 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3YrYvX1c-ytzIe-2Ez:-O3Z-blz12oEsoUIzVFJ:btpjokk) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/pixeltable/functions/timestamp.py#L157)):*

It's generally fraught to contrive dates into special cases of datetimes; this is why python keeps them separate. Pixeltable's type system allows either; shouldn't we take advantage of that and just return self.date() here?

I think there's a case to be made for a Pixeltable Date type, separate from Timestamp. Particularly if we want to handle timeseries data in the future, there's a lot of value to getting this right.

In general there are four distinct entities one encounters, that are all different, and we should think through how we want to handle/represent them in pixeltable:
1. Absolute timestamp (= a number of microseconds since the epoch)
2. A date
3. A time of day
4. A wall clock datetime (tuple of 2 & 3)

(1) and (4) differ because (4) is sensitive to time zone and (1) is not. The absolute timestamp boundaries of (2) are timezone sensitive and are not necessarily 24 hours apart (due to daylight savings adjustments). A lot of datetime libraries get tripped up by these types of considerations. I'm not proposing that we need 4 different types in Pixeltable, necessarily, but we should think through this.
</blockquote></details>

I'll leave these out for now.
___
*[`tests/functions/test_timestamp.py` line 16 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3YsTrT3VyZF7n4GUxq:-O3Z-wsjBLktC9vgCGLJ:b-4b2ab1) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/tests/functions/test_timestamp.py#L16)):*

More descriptive name here? `test_timestamp_udfs`?
</blockquote></details>

This is specifically in tests/functions, so the fact that it's testing functions is implied.
___
*[`tests/functions/test_timestamp.py` line 21 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3YupKxChQi5chKSEjo:-O3Z09Y_6os_JZU_59gF:b-3pnmnc) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/tests/functions/test_timestamp.py#L21)):*

Easier to just do `from pixeltable.functions.timestamp import *` here?
</blockquote></details>

Didn't work. I got an error that * is only allowed at module level.
___
*[`tests/functions/test_timestamp.py` line 51 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3Yvdi0BB6uOR2aNt9_:-O3Z1KHrAjeWLU2NKjCg:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/tests/functions/test_timestamp.py#L51)):*

Random style note, I believe best practice is not to use `\`, but to enclose in parens. that is, instead of
```python
assert a == \
    b
```
it's preferred to do
```python
assert (a ==
    b)
```
</blockquote></details>

Done.

mkornacker (commented):
They've basically been ignored so far. We should think this through, but I don't think this should hold up getting these functions in.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/239#-:-O3Z69IoBR7Y6s89TZjC:bk05efm)* status: 0 of 11 files reviewed, 9 unresolved discussions (waiting on @aaron-siegel)

aaron-siegel (changes requested):
Ok. keeping out the conversion methods until we think through it (which I see you've done) seems like a reasonable way to go

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/239#-:-O3_0WNiCuRefQaSxFJI:b-yeo3h1)* status: 0 of 11 files reviewed, 9 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/functions/timestamp.py` line 34 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/239#-O3Yr1Oe5aWPSurMVdyB:-O3_0GglBAbNRHd9Vuws:b1p5p3i) ([raw file](https://github.com/pixeltable/pixeltable/blob/4323221ee6687f57dee24e0abfa3383071745aac/pixeltable/functions/timestamp.py#L34)):*

This is simply copying from the official Python docs, which is presumably good enough.

https://docs.python.org/3/library/datetime.html#datetime.datetime.month
</blockquote></details>

Ok. Yeah, if that's what the Python docs say, we should just copy them. sigh

aaron-siegel (approved):

PR #241, opened 2024-08-07
Enable method call syntax for all string, timestamp, and video UDFs

aaron-siegel:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/241)

pixeltable/functions/timestamp.py
```diff
@@ -17,7 +17,7 @@
 from pixeltable.utils.code import local_public_names
 
 
-@func.udf
+@func.udf(is_method=True)
```
mkornacker, line 20: This (and month, day, etc.) are properties for Python's datetime, let's make them properties here as well.

mkornacker (approved):

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/241)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`pixeltable/functions/timestamp.py` line 20 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/241#gh-1707898216:-O3iwv-tA2iBNcWzv9Fn:b73sr5y) ([raw file](https://github.com/pixeltable/pixeltable/blob/6015757f53894936f1babf54e88f5121029de625/pixeltable/functions/timestamp.py#L20)):*

This (and month, day, etc.) are properties for Python's datetime, let's make them properties here as well.
</blockquote></details>

I'd thought about that, but it's not going to work if we want to make them timezone-configurable. They'll need to take an optional timezone parameter when we make the API timezone aware.

PR #243, opened 2024-08-08
Update fundamentals tutorial

aaron-siegel:
Updates the fundamentals tutorial with two new examples:
- using `contains()` as the predicate in a `where` clause
- using `update()` with an expression (rather than a constant)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/243)

mkornacker (approved):

PR #248, opened 2024-08-14
Remove support for Python `datetime.date` type, until we can add a proper Pixeltable `DateType`

aaron-siegel:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/248)

pixeltable/tool/create_test_db_dump.py
```diff
@@ -61,7 +61,7 @@ def dump_db(self) -> None:
         info_dict = {'pixeltable-dump': {
             'metadata-version': md_version,
             'git-sha': git_sha,
-            'datetime': datetime.datetime.utcnow(),
+            'datetime': datetime.datetime.now(tz=datetime.timezone.utc),
```
mkornacker, line 64: comment why not utcnow() here

mkornacker (approved):

pixeltable/tool/create_test_db_dump.py
aaron-siegel: It's deprecated as of Python 3.12. Does that really need a comment?

PR #254, opened 2024-08-17
Performance improvements for StoreBase.load_column()

mkornacker:
- existing logic: individual Update statement per row
- new logic: create and populate temp table with new values, then update the target table via a join to the temp table

Anecdotal performance comparison for adding a computed image column to a table with 100k rows: throughput goes up from ~1.4k rows/s to ~16k rows/s.

### DIFF
--- a/pixeltable/store.py
+++ b/pixeltable/store.py
@@ -223,35 +223,70 @@ class StoreBase:
         """
         num_excs = 0
         num_rows = 0
-        for row_batch in exec_plan:
-            num_rows += len(row_batch)
-            for result_row in row_batch:
-                values_dict: Dict[sql.Column, Any] = {}
 
-                if col.is_computed:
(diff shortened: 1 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/254)

aaron-siegel (approved):
Seems like a pretty nice pattern. Is this a common design pattern in sql DB architecture or is it something you invented to solve this problem?

It'd be great to see some data (either as code comments or in the PR description) on how big of a difference this makes (before-and-after perf comparison).

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/254#-:-O4Xuopy7PDa5CK-E4HP:b36doe7)* status: 0 of 2 files reviewed, all discussions resolved

mkornacker:
I just updated the comment.

PR #258, opened 2024-08-20
Aggregate make_list()

mkornacker:
Also adds an implementation for JsonType.supertype().

### DIFF
--- a/docs/release/api/functions/json.md
+++ b/docs/release/api/functions/json.md
@@ -1,0 +1,1 @@
+## ::: pixeltable.functions.json
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -31,6 +31,7 @@ nav:
     - Table: api/table.md
   - Built-in Functions and Operators:
     - image: api/functions/image.md
+    - json: api/functions/json.md
     - string: api/functions/string.md
     - timestamp: api/functions/timestamp.md
     - video: api/functions/video.md
--- a/pixeltable/functions/__init__.py
+++ b/pixeltable/functions/__init__.py
@@ -1,4 +1,4 @@
-from . import fireworks, huggingface, image, openai, string, together, video, timestamp
+from . import fireworks, huggingface, image, openai, string, together, video, timestamp, json
 from .globals import *
 from pixeltable.utils.code import local_public_names
 
--- a/pixeltable/functions/json.py
+++ b/pixeltable/functions/json.py
@@ -1,0 +1,46 @@
+"""
+Pixeltable [UDFs](https://pixeltable.readme.io/docs/user-defined-functions-udfs) for `JsonType`.
+
+Example:
+```python
+import pixeltable as pxt
+
+t = pxt.get_table(...)
+t.select(pxt.functions.json.make_list()).collect()
+```
+"""
+
+from typing import Any
+
+import pixeltable.func as func
+import pixeltable.type_system as ts
+from pixeltable.utils.code import local_public_names
+
+
+@func.uda(
+    update_types=[ts.JsonType(nullable=True)],
+    value_type=ts.JsonType(),
+    requires_order_by=False,
+    allows_window=False,
+)
+class make_list(func.Aggregator):
+    """
+    Collects arguments into a list.
+    """
+    def __init__(self):
+        self.output: list[Any] = []
+
+    def update(self, obj: Any) -> None:
+        if obj is None:
+            return
+        self.output.append(obj)
+
+    def value(self) -> list[Any]:
+        return self.output
+
+
+__all__ = local_public_names(__name__)
+
+
+def __dir__():
+    return __all__
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -519,6 +519,28 @@ class JsonType(ColumnType):
         super().__init__(self.Type.JSON, nullable=nullable)
         self.type_spec = type_spec
 
+    def supertype(self, other: ColumnType) -> Optional[JsonType]:
+        if self.type_spec is None:
+            # we don't have a type spec and can accept anything accepted by other
+            return type(self)(nullable=(self.nullable or other.nullable))
+        if other.type_spec is None:
+            # we have a type spec but other doesn't
+            return type(self)(nullable=(self.nullable or other.nullable))
+
+        # we both have type specs; the supertype's type spec is the union of the two
+        type_spec = deepcopy(self.type_spec)
+        for other_field_name, other_field_type in other.type_spec.items():
+            if other_field_name not in type_spec:
+                type_spec[other_field_name] = other_field_type.copy()
+            else:
+                # both type specs have this field
+                field_type = type_spec[other_field_name].supertype(other_field_type)
+                if field_type is None:
+                    # conflicting types
+                    return None
+                type_spec[other_field_name] = field_type
+        return type(self)(type_spec, nullable=(self.nullable or other.nullable))
+
     def _as_dict(self) -> Dict:
         result = super()._as_dict()
         if self.type_spec is not None:
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -522,10 +522,10 @@ class JsonType(ColumnType):
     def supertype(self, other: ColumnType) -> Optional[JsonType]:
         if self.type_spec is None:
             # we don't have a type spec and can accept anything accepted by other
-            return type(self)(nullable=(self.nullable or other.nullable))
+            return JsonType(nullable=(self.nullable or other.nullable))
         if other.type_spec is None:
             # we have a type spec but other doesn't
-            return type(self)(nullable=(self.nullable or other.nullable))
+            return JsonType(nullable=(self.nullable or other.nullable))
 
         # we both have type specs; the supertype's type spec is the union of the two
         type_spec = deepcopy(self.type_spec)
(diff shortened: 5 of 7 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/258)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/258)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/type_system.py` line 517 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4kneg0E2yT0h7I4YNh:-O4kneg0E2yT0h7I4YNi:bz8h1z) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/pixeltable/type_system.py#L517)):*
> ```Python
> 
> class JsonType(ColumnType):
>     # TODO: type_spec also needs to be able to express lists
> ```

We should consider having separate `ListType` and `DictType` column types that are subtypes of JsonType. Then the hierarchy would be
```
JsonType
| IntType
| FloatType
| StringType
| BoolType
| ListType
| DictType
```
___
*[`pixeltable/type_system.py` line 518 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4kn2CI-SP-WcGcrHHu:-O4kn2CI-SP-WcGcrHHv:b-l45con) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/pixeltable/type_system.py#L518)):*
> ```Python
> class JsonType(ColumnType):
>     # TODO: type_spec also needs to be able to express lists
>     def __init__(self, type_spec: Optional[Dict[str, ColumnType]] = None, nullable: bool = False):
> ```

Add a comment here saying what `type_spec` is. I gather that it means "if specified, then the JSON dict can ONLY contain keys that appear in `type_spec`, and the values of those keys must match the given type" but this isn't obvious from the name `type_spec`.

(For example, a valid alternative interpretation might be, "if specified, then the JSON dict can contain any keys, but IF a key that appears in `type_spec` is in the dict, THEN its value must match the given type".)
___
*[`pixeltable/type_system.py` line 525 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4klkh-A5MFEzJRD12w:-O4klkh-A5MFEzJRD12x:b-hckttu) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/pixeltable/type_system.py#L525)):*
> ```Python
>         if self.type_spec is None:
>             # we don't have a type spec and can accept anything accepted by other
>             return type(self)(nullable=(self.nullable or other.nullable))
> ```

Why use `type(self)` here? I find it confusing; it'd be clearer to state `JsonType` explicitly.
___
*[`pixeltable/type_system.py` line 540 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4kmixw4fvmXw4WSwNr:-O4kmixw4fvmXw4WSwNs:bp4tr84) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/pixeltable/type_system.py#L540)):*
> ```Python
>                 if field_type is None:
>                     # conflicting types
>                     return None
> ```

Shouldn't the fallback be
```
JsonType(nullable=(self.nullable or other.nullable))
```
rather than `None`? An unspec'ed `JsonType` should be supertype of any `JsonType`s including ones with conflicting specs.
___
*[`pixeltable/functions/json.py` line 26 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4kp3otBcEM_ASIeCpO:-O4kp3otBcEM_ASIeCpP:btaxltz) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/pixeltable/functions/json.py#L26)):*
> ```Python
>     allows_window=False,
> )
> class make_list(func.Aggregator):
> ```

It feels like `make_list` is pretty impoverished if we can't turn a group of `StringType()` cells into a list (for example).

I'd argue we really need to make JsonType a supertype of the scalar types. Maybe not in this PR, but soon.
___
*[`tests/test_exprs.py` line 821 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4koPPd8jpD0822Vd_8:-O4koPPd8jpD0822Vd_9:bfxdgj4) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/tests/test_exprs.py#L821)):*
> ```Python
>         val = res[0]['out']
>         assert len(val) == t.count()
>         # we can't compare the values directly, because the order of the elements in the list is not guaranteed
> ```

turn them into sets and compare them that way. alternatively, put an order_by(t.c2) in the select; this would additionally test that make_list respects order_by. Or have separate invocations with and without order_by, and test the value in the latter.

Regardless, I think it's important that we test the value for correctness in some way.
___
*[`tests/test_types.py` line 90 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4koeF2CrOS0jigQLzZ:-O4koeF2CrOS0jigQLz_:b9jo5kg) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/tests/test_types.py#L90)):*
> ```Python
>             (JsonType(type_spec={'a': IntType()}), JsonType(), JsonType()),
>             (JsonType(type_spec={'a': IntType()}), JsonType(type_spec={'b': StringType()}), JsonType(type_spec={'a': IntType(), 'b': StringType()})),
>             (JsonType(type_spec={'a': IntType()}), JsonType(type_spec={'a': StringType()}), None),
> ```

This should produce JsonType() as a supertype, not None (see comment above)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/258)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/type_system.py` line 517 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4kneg0E2yT0h7I4YNh:-O4kpYZq-H0gV9w9yLZ0:b-qojat1) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/pixeltable/type_system.py#L517)):*

We should consider having separate `ListType` and `DictType` column types that are subtypes of JsonType. Then the hierarchy would be
```
JsonType
| IntType
| FloatType
| StringType
| BoolType
| ListType
| DictType
```
</blockquote></details>

Sorry, that should be
```
JsonType
| FloatType
  | IntType
    | BoolType
| StringType
| ListType
| DictType
```

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/258)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/type_system.py` line 525 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4klkh-A5MFEzJRD12w:-O4nBV9B0ulspebQItz7:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/pixeltable/type_system.py#L525)):*

Why use `type(self)` here? I find it confusing; it'd be clearer to state `JsonType` explicitly.
</blockquote></details>

Done.
___
*[`pixeltable/type_system.py` line 540 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4kmixw4fvmXw4WSwNr:-O4nBWjo9lHGvtB6qDj4:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/pixeltable/type_system.py#L540)):*

Shouldn't the fallback be
```
JsonType(nullable=(self.nullable or other.nullable))
```
rather than `None`? An unspec'ed `JsonType` should be supertype of any `JsonType`s including ones with conflicting specs.
</blockquote></details>

Done.
___
*[`pixeltable/functions/json.py` line 26 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4kp3otBcEM_ASIeCpO:-O4nBaHSB_RAgAEfBa8f:b-hy66j3) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/pixeltable/functions/json.py#L26)):*

It feels like `make_list` is pretty impoverished if we can't turn a group of `StringType()` cells into a list (for example).

I'd argue we really need to make JsonType a supertype of the scalar types. Maybe not in this PR, but soon.
</blockquote></details>

We talked about PolymorphicAggregateFunction, which this would be.
___
*[`tests/test_exprs.py` line 821 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4koPPd8jpD0822Vd_8:-O4nDtbL0sKmkHhqJu0X:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/tests/test_exprs.py#L821)):*

turn them into sets and compare them that way. alternatively, put an order_by(t.c2) in the select; this would additionally test that make_list respects order_by. Or have separate invocations with and without order_by, and test the value in the latter.

Regardless, I think it's important that we test the value for correctness in some way.
</blockquote></details>

Done.
___
*[`tests/test_types.py` line 90 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/258#-O4koeF2CrOS0jigQLzZ:-O4nDuTy8zJi6H7YWBaa:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0a288ee9b1af10a0b0cffefb217a43c896ff668e/tests/test_types.py#L90)):*

This should produce JsonType() as a supertype, not None (see comment above)
</blockquote></details>

Done.

PR #264, opened 2024-08-26
Type system performance improvements

aaron-siegel:
On a sample dataset of ~500k rows, achieves a > 90% reduction in running time of the type inference logic in `import_rows()` on an M3 Max CPU, from 32.1 seconds to 2.56 seconds.

Most of the reduction (from 32 -> ~6 seconds) is achieved by eliminating the use of `deepcopy()` in `ColumnType.copy()`. The remainder is achieved by reducing object creation and dynamic evaluation in `ColumnType` and subclasses.

It is assumed that `ColumnType` instances are immutable. I believe that further optimizations are possible by making `ColumnType` instances formally immutable (as frozen dataclasses, say) and minimizing object creation. But those changes would be more invasive, and I'm pretty sure this is enough to get type inference off the critical performance path.

UPDATED: Optimized `JsonType._validate_literal`, reducing the time to 2.3 seconds (93% reduction vs. baseline)

### DIFF
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -169,25 +169,23 @@ class ColumnType:
         return isinstance(other, ColumnType) and self.matches(other) and self.nullable == other.nullable
 
     def is_supertype_of(self, other: ColumnType, ignore_nullable: bool = False) -> bool:
-        operand = self.copy(nullable=True) if ignore_nullable else self
-        return operand.supertype(other) == operand
+        if ignore_nullable:
+            supertype = self.supertype(other)
+            if supertype is None:
+                return False
+            return supertype.matches(self)
+        else:
+            return self.supertype(other) == self
 
     def matches(self, other: ColumnType) -> bool:
         """Two types match if they're equal, aside from nullability"""
-        if type(self) != type(other):
-            return False
-
-        for member_var in vars(self).keys():
-            if member_var == '_nullable':
-                continue
-            if getattr(self, member_var) != getattr(other, member_var):
-                return False
-        return True
+        # Default: just compare base types (this works for all types whose only parameter is nullable)
+        return self._type == other._type
 
     def supertype(self, other: ColumnType) -> Optional[ColumnType]:
         if self == other:
             return self
-        if self.copy(nullable=True) == other.copy(nullable=True):
+        if self.matches(other):
             return self.copy(nullable=(self.nullable or other.nullable))
 
         if self.is_invalid_type():
@@ -217,7 +215,7 @@ class ColumnType:
         if isinstance(val, datetime.datetime):
             return TimestampType(nullable=nullable)
         if isinstance(val, PIL.Image.Image):
-            return ImageType(width=val.width, height=val.height, mode=val.mode)
+            return ImageType(width=val.width, height=val.height, mode=val.mode, nullable=nullable)
         if isinstance(val, np.ndarray):
             col_type = ArrayType.from_literal(val, nullable=nullable)
             if col_type is not None:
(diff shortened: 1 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/264)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/264)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/type_system.py` line 171 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/264#-O5FuP-q8EGDGd_ujCJe:-O5FuP-q8EGDGd_ujCJf:brc02ep) ([raw file](https://github.com/pixeltable/pixeltable/blob/fa9db9afed4946bae10b3ad8e9e5cea9dfaec0b7/pixeltable/type_system.py#L171)):*
> ```Python
>         return isinstance(other, ColumnType) and self.matches(other) and self.nullable == other.nullable
> 
>     def is_supertype_of(self, other: ColumnType, ignore_nullable: bool = False) -> bool:
> ```

This should probably not do any object construction at all but simply do checks (and subclasses can specialize their own _is_supertype_of())
___
*[`pixeltable/type_system.py` line 182 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/264#-O5Ftt516SSzqUW8_TJw:-O5Ftt516SSzqUW8_TJx:bu62cpc) ([raw file](https://github.com/pixeltable/pixeltable/blob/fa9db9afed4946bae10b3ad8e9e5cea9dfaec0b7/pixeltable/type_system.py#L182)):*
> ```Python
>     def matches(self, other: ColumnType) -> bool:
>         """Two types match if they're equal, aside from nullability"""
>         # Default: just compare base types (this works for all types whose only parameter is nullable)
> ```

Add check that this is a scalar type.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/264)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/type_system.py` line 171 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/264#-O5FuP-q8EGDGd_ujCJe:-O5GZxxW0KAovn595yPA:bsurnhv) ([raw file](https://github.com/pixeltable/pixeltable/blob/fa9db9afed4946bae10b3ad8e9e5cea9dfaec0b7/pixeltable/type_system.py#L171)):*

This should probably not do any object construction at all but simply do checks (and subclasses can specialize their own _is_supertype_of())
</blockquote></details>

I agree, but it's out of scope for this PR - it's not on the critical path to the problem at hand, and cost of object creation is tiny compared to the perf problems solved by this PR. `ColumnType` needs more substantial work to minimize object creation and it'll be easiest to consider it in the context of that future project
___
*[`pixeltable/type_system.py` line 182 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/264#-O5Ftt516SSzqUW8_TJw:-O5GZXK902hBhegxYDEN:b-taxvxl) ([raw file](https://github.com/pixeltable/pixeltable/blob/fa9db9afed4946bae10b3ad8e9e5cea9dfaec0b7/pixeltable/type_system.py#L182)):*

Add check that this is a scalar type.
</blockquote></details>

No- default implementation works for various non-scalars as well (Video, Audio)

pixeltable/type_system.py
```diff
@@ -552,11 +552,20 @@ def print_value(self, val: Any) -> str:
 
     def _validate_literal(self, val: Any) -> None:
         if not isinstance(val, dict) and not isinstance(val, list):
+            # TODO In the future we should accept scalars too, which would enable us to remove this top-level check
```
mkornacker, line 555: remove todo (let's discuss this first...)

mkornacker (approved):

PR #267, opened 2024-08-31
Clean up the Table class namespace; disallow conflicting column names

aaron-siegel:
Several changes:
- Disallow column names that conflict with a system method of the `Table` or `InsertableTable` class, such as `add_column` or `insert`
- Clean up the `Table` class namespace (e.g., remove the `name` property, since it is likely to be a popular column name)
- Unify the code paths for `t.my_col` and `t['my_col']`

Also, more substantively: there are two ways to create columns: using `t.add_column()`, and setitem style `t['my_col'] = ...`. For `add_column()`, additional properties may be specified,
```python
t.add_column(my_col=pxt.ImageType(), stored=False)
```
Pixeltable has allowed the following syntax for setitem style:
```python
t['my_col'] = {'type': pxt.ImageType(), 'stored': False}
```
**I propose to disallow this.** The reason is that a dictionary in this position means something completely different as an argument to `add_column()`: it means "construct an `InlineDict`".

It is very confusing for the two forms to behave mostly identically, but then very differently specifically when you pass them a dict! This means that setitem will be a feature-incomplete synonym, but that's already true (it doesn't return an `UpdateStatus`), and I don't see harm in simply saying that `add_column()` is the API to use if you want to fine-tune the column parameters.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/267)

pixeltable/catalog/globals.py
```diff
@@ -1,23 +1,26 @@
...
+from typing import Optional
 
 _logger = logging.getLogger('pixeltable')
 
 # name of the position column in a component view
-POS_COLUMN_NAME = 'pos'
+_POS_COLUMN_NAME = 'pos'
 _ROWID_COLUMN_NAME = '_rowid'
 
+# Set of symbols that are predefined in the `InsertableTable` class (and are therefore not allowed as column names).
```
mkornacker, line 12: Why InsertableTable (and not Table, etc.)?

pixeltable/catalog/table.py
```diff
@@ -60,7 +62,7 @@ def __hash__(self) -> int:
 
     def _check_is_dropped(self) -> None:
         if self._is_dropped:
-            raise excs.Error(f'{self.display_name()} {self.name} has been dropped')
+            raise excs.Error(f'{self._display_name()} {self._name} has been dropped')
```
mkornacker, line 76: I was expecting an implementation of get_metadata() for this class that adds the table schema to the SchemaObject get_metadata() output.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/267)* status: 0 of 23 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/globals.py` line 11 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/267#gh-1739555009:-O5aXl6_5YNdcYLgWWr9:b-665hj4) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8fcc1dd596d2b6d9505ac2ed43c0ccb82310789/pixeltable/catalog/globals.py#L11)):*

Why InsertableTable (and not Table, etc.)?
</blockquote></details>

Because InsertableTable inherits everything from Table and SchemaObject in its __dir__. (It's also a superset of the symbols in View, though I suppose we could add View explicitly just in case that ever changes.)
___
*[`pixeltable/catalog/table.py` line 65 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/267#gh-1739556204:-O5aY7srCpeZNy5p0jI3:b-7c1va4) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8fcc1dd596d2b6d9505ac2ed43c0ccb82310789/pixeltable/catalog/table.py#L65)):*

I was expecting an implementation of get_metadata() for this class that adds the table schema to the SchemaObject get_metadata() output.
</blockquote></details>

Done. Did the same for `base`, `comment`, `num_retained_versions`,  and `version`. We should consider taking those out of the namespace too.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/267)* status: 0 of 23 files reviewed, 2 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/catalog/table.py` line 65 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/267#gh-1739556204:-O5dW9gw-9D9Ihgkw5Qk:bf0awv4) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8fcc1dd596d2b6d9505ac2ed43c0ccb82310789/pixeltable/catalog/table.py#L65)):*

Done. Did the same for `base`, `comment`, `num_retained_versions`,  and `version`. We should consider taking those out of the namespace too.
</blockquote></details>

I think you should do that in this PR, version and comment sound like they could be popular column names.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/267)* status: 0 of 23 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/table.py` line 65 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/267#gh-1739556204:-O5dxZJk-JHUOm0Xa8yM:bxorxlv) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8fcc1dd596d2b6d9505ac2ed43c0ccb82310789/pixeltable/catalog/table.py#L65)):*

I think you should do that in this PR, version and comment sound like they could be popular column names.
</blockquote></details>

Done. I removed the setters for now (I don't think that's the right API for altering table attributes anyway; we need a more general "alter table" mechanism that needs to be thought through)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/267)* status: 0 of 23 files reviewed, 8 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/catalog/globals.py` line 11 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/267#gh-1739555009:-O5j7aW899fIHpgBHfdH:b1c5ltr) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8fcc1dd596d2b6d9505ac2ed43c0ccb82310789/pixeltable/catalog/globals.py#L11)):*

Because InsertableTable inherits everything from Table and SchemaObject in its __dir__. (It's also a superset of the symbols in View, though I suppose we could add View explicitly just in case that ever changes.)
</blockquote></details>

It should really be 'Table class hierarchy'. The fact that InsertableTable is a superset right now is not by design.
___
*[`pixeltable/catalog/globals.py` line 49 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5jC-I87-HDMREnLLGp:-O5jC-I87-HDMREnLLGq:b-bc1enp) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/globals.py#L49)):*
> ```Python
>     global _PREDEF_SYMBOLS
>     if _PREDEF_SYMBOLS is None:
>         _PREDEF_SYMBOLS = set(dir(InsertableTable))
> ```

Also add symbols from View.
___
*[`pixeltable/catalog/table.py` line 53 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5j9K-s5I4ZEl-qIauy:-O5j9K-s5I4ZEl-qIauz:bfxq590) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L53)):*
> ```Python
>     def get_metadata(self) -> dict[str, Any]:
>         md = super().get_metadata()
>         md['base'] = self._base._path if self._base is not None else None
> ```

I think only views should expose this (and an 'is_snapshot' flag). We should probably have a 'type': ['basetable', 'view'] entry as well.
___
*[`pixeltable/catalog/table.py` line 55 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5jA-fo4EkLP_SL20im:-O5jA-fo4EkLP_SL20in:bhkomia) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L55)):*
> ```Python
>         md['base'] = self._base._path if self._base is not None else None
>         md['schema'] = self.column_types()
>         md['version'] = self._version
> ```

There's also a schema version that's distinct from the table version.
___
*[`pixeltable/catalog/table.py` line 175 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5jAK-B56zXCDbBBBOC:-O5jAK-B56zXCDbBBBOD:bvubkg2) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L175)):*
> ```Python
>         return self._df().count()
> 
>     def column_names(self) -> list[str]:
> ```

make private (I don't think users would want to call this; they can get a schema listing with __repr__() and if they want the schema as a dict they can call get_metadata()
___
*[`pixeltable/catalog/table.py` line 179 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5j99iH2xlAezw1W7-r:-O5j99iH2xlAezw1W7-s:b-m9wikd) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L179)):*
> ```Python
>         return [c.name for c in self._tbl_version_path.columns()]
> 
>     def column_types(self) -> dict[str, ts.ColumnType]:
> ```

This should be called schema() and should also be private
___
*[`pixeltable/catalog/table.py` line 183 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5jACFb89Pse8uf1kCX:-O5jACFb89Pse8uf1kCY:b-e004af) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L183)):*
> ```Python
>         return {c.name: c.col_type for c in self._tbl_version_path.columns()}
> 
>     def query_names(self) -> list[str]:
> ```

Also make private.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/267)* status: 0 of 23 files reviewed, 8 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/table.py` line 53 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5j9K-s5I4ZEl-qIauy:-O5sw1mVBOGoyBb4_Pf9:b69i9kt) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L53)):*

I think only views should expose this (and an 'is_snapshot' flag). We should probably have a 'type': ['basetable', 'view'] entry as well.
</blockquote></details>

I'd argue it's most logical to always include 'base' in the metadata, and set it to None for a base table.

I added a 'kind' field ('type' is already very overloaded terminology) with values 'base_type', 'view', and 'snapshot' (this makes more sense to me than a separate is_snapshot field).
___
*[`pixeltable/catalog/table.py` line 55 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5jA-fo4EkLP_SL20im:-O5swOwF27FUyuWf_jyR:b-3otmos) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L55)):*

There's also a schema version that's distinct from the table version.
</blockquote></details>

Is that something that we want to be user-facing?
___
*[`pixeltable/catalog/table.py` line 175 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5jAK-B56zXCDbBBBOC:-O5swTfb5guvHBsHrFYf:b-ie0gwm) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L175)):*

make private (I don't think users would want to call this; they can get a schema listing with __repr__() and if they want the schema as a dict they can call get_metadata()
</blockquote></details>

Done. Also made it (and the others) into a property.
___
*[`pixeltable/catalog/table.py` line 179 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5j99iH2xlAezw1W7-r:-O5swSvUBnA7cmKNCznu:bon65rr) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L179)):*

This should be called schema() and should also be private
</blockquote></details>

Done. I could see an argument for getting rid of column_names entirely and having schema.keys() be the preferred way to get the column names (but I haven't made this change yet)
___
*[`pixeltable/catalog/table.py` line 183 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/267#-O5jACFb89Pse8uf1kCX:-O5swUPQDrVOqlw-X_i2:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0f68e53ea800ae806836d57752b58c49210dcdf5/pixeltable/catalog/table.py#L183)):*

Also make private.
</blockquote></details>

Done.

mkornacker (approved):

PR #273, opened 2024-09-04
Set up Nightly CI

aaron-siegel:
Sets up a nightly (actually, thrice-weekly) CI workflow that `pip install`s Pixeltable along with various combinations of optional dependencies and ensures that the tests run.

This isn't intended to detect regressions due to code changes, but rather, regressions that happen due to something else changing in the environment (e.g., a third party releasing a new version of a library that gets preferentially `pip install`ed but is backward-incompatible).

I've already found (and fixed) a couple bugs this way:
- `img.resize()` is broken with the latest Pillow release
- import errors when using torch/transformers with pyarrow

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,77 @@
+name: Run pytest
+
+on:
+  schedule:
+    - cron: "19 2 * * 1,3,5"  # Monday, Wednesday, Friday at 02:19 UTC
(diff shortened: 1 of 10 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/273)

mkornacker (approved):

PR #277, opened 2024-09-07
Type-checking correctness for `catalog`, `functions`, and `ext` packages

aaron-siegel:
Also includes some miscellaneous other fixes, including remediating all type issues that are visible as errors in vscode.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/277)

pixeltable/catalog/table.py
```diff
@@ -282,7 +288,7 @@ def __setitem__(self, col_name: str, spec: Union[ts.ColumnType, exprs.Expr]) ->
             raise excs.Error(f'Column name must be a string, got {type(col_name)}')
         if not isinstance(spec, (ts.ColumnType, exprs.Expr)):
             raise excs.Error(f'Column spec must be a ColumnType or an Expr, got {type(spec)}')
-        self.add_column(**{col_name: spec})
+        self.add_column(type=None, stored=None, print_stats=False, **{col_name: spec})
```
mkornacker, line 291: comment: why does it need those explicitly?

pixeltable/catalog/table_version.py
```diff
@@ -119,12 +150,12 @@ def __init__(
         cat.tbl_versions[(self.id, self.effective_version)] = self
 
         # init schema after we determined whether we're a component view, and before we create the store table
-        self.cols: list[Column] = []  # contains complete history of columns, incl dropped ones
-        self.cols_by_name: dict[str, Column] = {}  # contains only user-facing (named) columns visible in this version
-        self.cols_by_id: dict[int, Column] = {}  # contains only columns visible in this version, both system and user
+        self.cols = []  # contains complete history of columns, incl dropped ones
```
mkornacker: preserve comments

pixeltable/catalog/table_version.py
```diff
@@ -802,7 +833,7 @@ def _validate_update_spec(
         return update_targets
 
     def propagate_update(
-            self, plan: Optional[exec.ExecNode], where_clause: Optional[sql.ClauseElement],
+            self, plan: Optional['exec.ExecNode'], where_clause: Optional[sql.ColumnElement[bool]],
```
mkornacker: Does it require 'bool'? If not, let's leave it off, we can't statically type-check exprs anyway.

pixeltable/catalog/table_version_path.py
```diff
@@ -82,9 +83,7 @@ def find_tbl_version(self, id: UUID) -> Optional[TableVersion]:
 
     def __getattr__(self, col_name: str) -> 'pixeltable.exprs.ColumnRef':
         """Return a ColumnRef for the given column name."""
-        from pixeltable.exprs import ColumnRef, RowidRef
-        if col_name == _POS_COLUMN_NAME and self.is_component_view():
```
mkornacker, line 86: Where did this go?

pixeltable/exprs/expr.py
```diff
@@ -215,12 +216,12 @@ def is_bound_by(self, tbl: catalog.TableVersionPath) -> bool:
                 return False
         return True
 
-    def retarget(self, tbl: catalog.TableVersionPath) -> Expr:
+    def retarget(self, tbl: catalog.TableVersionPath) -> Self:
```
mkornacker, line 229: This is only available starting with 3.11.

pixeltable/functions/string.py
```diff
@@ -12,8 +12,9 @@
 ```
 """
 
-from typing import Any, Optional
+from typing import Any, Literal, Optional
```
mkornacker: Where is Literal being used?

pixeltable/metadata/schema.py
```diff
@@ -64,8 +65,8 @@ class DirMd:
 class Dir(Base):
     __tablename__ = 'dirs'
 
-    id = sql.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4, nullable=False)
-    parent_id = sql.Column(UUID(as_uuid=True), ForeignKey('dirs.id'), nullable=True)
+    id: orm.Mapped[uuid.UUID] = orm.mapped_column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4, nullable=False)
```
mkornacker, line 68: requires explanatory comment

Makefile
```diff
@@ -77,7 +77,7 @@ nbtest: install
 
 .PHONY: typecheck
 typecheck: install
-	@mypy pixeltable/*.py
+	@mypy pixeltable/*.py pixeltable/catalog pixeltable/ext pixeltable/functions
```
mkornacker, line 80: dummy comment; ignore

mkornacker (changes requested):

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/277)* status: 0 of 29 files reviewed, 8 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/table.py` line 291 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755933375:-O6b0SfH9ePO69SkotCn:b-u6ztjm) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table.py#L291)):*

comment: why does it need those explicitly?
</blockquote></details>

mypy complains if it sees a `**{...}` argument and the call doesn't also mention the explicit named parameters.

The alternative would be a `# type: ignore` directive.
___
*[`pixeltable/catalog/table_version.py` line 153 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755934407:-O6b0wal-8wnrOB3K_4C:bb7hquh) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table_version.py#L153)):*

preserve comments
</blockquote></details>

huh?
___
*[`pixeltable/catalog/table_version.py` line 836 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755937598:-O6b1SNK3oEMWx9sXQ_M:bxabxgm) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table_version.py#L836)):*

Does it require 'bool'? If not, let's leave it off, we can't statically type-check exprs anyway.
</blockquote></details>

It's fine without the `bool`; I removed it
___
*[`pixeltable/catalog/table_version_path.py` line 0 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755939306:-O6b1fiW9Mhi-2Tw9f0_:buiraz9) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table_version_path.py#L0)):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map [this GitHub inline comment thread](https://github.com/pixeltable/pixeltable/pull/277#discussion_r1755939306) to the right spot — sorry!)*

Where did this go?
</blockquote></details>

I assume this is referring to the `RowidRef` thing?

It isn't actually being used anywhere (removing this clause broke no tests), and is statically AND semantically false, since (1) the static result type of this method is `ColumnRef`, and (2) clients of this method freely assume they're getting a `ColumnRef`.

If you think it needs to stick around, we should discuss
___
*[`pixeltable/exprs/expr.py` line 219 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755941321:-O6b25na8lgbqNpJOz8p:bhskpyq) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/exprs/expr.py#L219)):*

This is only available starting with 3.11.
</blockquote></details>

Not true. It's available on 3.9+ via the `typing_extensions` package, which is how I've imported it.

BTW I do all my development and testing on 3.9 for exactly this reason.
___
*[`pixeltable/functions/string.py` line 15 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755942742:-O6b2Y24BhlfP_73enrW:buronh3) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/functions/string.py#L15)):*

Where is Literal being used?
</blockquote></details>

oops, it's not. It's residue from a failed experiment.

Fixed
___
*[`pixeltable/metadata/schema.py` line 68 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755944532:-O6b2ieb8ABVfKa6blUh:bticc1h) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/metadata/schema.py#L68)):*

requires explanatory comment
</blockquote></details>

I don't know if I agree. We're going to be using this pattern throughout the codebase as we do this refactor. The "explanatory comment" would be a link to this doc:

https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#whatsnew-20-orm-declarative-typing

which explains the correct (type-safe) way to define these mappings in SQLAlchemy 2.0. Do we really need to put that URL in the code base every time we follow this pattern? If not, then why here specifically (the only thing that's special about it is it happens to be the first one we've modernized?)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/277)* status: 0 of 33 files reviewed, 7 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/catalog/table.py` line 291 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755933375:-O6dqXlQ6I3MxAWjAcnU:b3e4e3m) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table.py#L291)):*

mypy complains if it sees a `**{...}` argument and the call doesn't also mention the explicit named parameters.

The alternative would be a `# type: ignore` directive.
</blockquote></details>

I meant: comment in the code. Otherwise the next person might remove that again.

Maybe type: ignore is better...
___
*[`pixeltable/catalog/table_version.py` line 153 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755934407:-O6dqMRrBHD1UzyCSXpW:btrlleu) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table_version.py#L153)):*

huh?
</blockquote></details>

Each of the initializations has a comment. Those need to move to the corresponding class-level declaration.
___
*[`pixeltable/catalog/table_version_path.py` line 0 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755939306:-O6dr6fwDbIf2P9g89hJ:b-pf953p) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table_version_path.py#L0)):*

I assume this is referring to the `RowidRef` thing?

It isn't actually being used anywhere (removing this clause broke no tests), and is statically AND semantically false, since (1) the static result type of this method is `ColumnRef`, and (2) clients of this method freely assume they're getting a `ColumnRef`.

If you think it needs to stick around, we should discuss
</blockquote></details>

Okay, we need to add a test, but this is required functionality. At least the object detection notebook is referencing this.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/277)* status: 0 of 33 files reviewed, 7 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/table.py` line 291 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755933375:-O6eIK7X4OWwvQ5uVhja:b-rr41ox) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table.py#L291)):*

I meant: comment in the code. Otherwise the next person might remove that again.

Maybe type: ignore is better...
</blockquote></details>

Well the next person won't be able to just remove it, since that will now trigger a type checking failure in CI.

It's bothering you a lot more than it's bothering me so you make the call on what to do.
___
*[`pixeltable/catalog/table_version.py` line 153 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755934407:-O6eIiWxANUFu58ZIO-C:b-yglom6) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table_version.py#L153)):*

Each of the initializations has a comment. Those need to move to the corresponding class-level declaration.
</blockquote></details>

This was taken care of when merging your latest PR (which did the same thing)
___
*[`pixeltable/catalog/table_version_path.py` line 0 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/277#gh-1755939306:-O6eIoKU5o24ThZB2eGw:bojuz74) ([raw file](https://github.com/pixeltable/pixeltable/blob/480fff23f54205d656a8067c0e3bb1ace4837019/pixeltable/catalog/table_version_path.py#L0)):*

Okay, we need to add a test, but this is required functionality. At least the object detection notebook is referencing this.
</blockquote></details>

That's not true. The object detection notebook is running in CI now (and passing). It appears that without it, `t.pos` resolves as a `ColumnRef` instead and still gives the right answers.

If we still need it as a separate class, then either we need to (1) make RowidRef a subclass of ColumnRef (or give them a common superclass) or (2) add RowidRef to all the upstream Unions (up to Table, which makes it user-facing) or (3) replace all the Unions with Expr.

mkornacker (approved):

PR #278, opened 2024-09-07
Initial support for converting FunctionCalls to SQL

mkornacker:
- this only works for scalar functions at the moment, not AggregateFunctions
- new class SqlElementCache: this will enable componsition of Select statements via a With clause, which is needed for creating Select statements for DataFrames with aggregate function calls

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -546,6 +546,7 @@ class Table(SchemaObject):
             metric: str = 'cosine'
     ) -> None:
         """Add an index to the table.
+
         Args:
             col_name: name of column to index
             idx_name: name of index, which needs to be unique for the table; if not provided, a name will be generated
--- a/pixeltable/catalog/table_version.py
+++ b/pixeltable/catalog/table_version.py
@@ -44,6 +44,40 @@ class TableVersion:
         have TableVersions reference those
     - mutable TableVersions record their TableVersionPath, which is needed for expr evaluation in updates
     """
+
+    id: UUID
+    name: str
+    version: int
+    comment: str
+    num_retained_versions: int
+    schema_version: int
+    view_md: Optional[schema.ViewMd]
+    is_snapshot: bool
+    effective_version: Optional[int]
+    path: Optional[pxt.catalog.TableVersionPath]
+    base: Optional[TableVersion]
+    next_col_id: int
+    next_idx_id: int
+    next_rowid: int
+    predicate: Optional[exprs.Expr]
+    mutable_views: List[TableVersion]
+    iterator_cls: Optional[Type[ComponentIterator]]
+    iterator_args: Optional[exprs.InlineDict]
+    num_iterator_cols: int
+
+    # contains complete history of columns, incl dropped ones
+    cols: list[Column]
+    # contains only user-facing (named) columns visible in this version
+    cols_by_name: dict[str, Column]
+    # contains only columns visible in this version, both system and user
+    cols_by_id: dict[int, Column]
+    # needed for _create_tbl_md()
+    idx_md: dict[int, schema.IndexMd]
+    # contains only actively maintained indices
+    idxs_by_name: dict[str, TableVersion.IndexInfo]
+
+    external_stores: dict[str, pxt.io.ExternalStore]
+
     @dataclasses.dataclass
     class IndexInfo:
         id: int
@@ -95,13 +129,13 @@ class TableVersion:
         from pixeltable import exprs
         predicate_dict = None if not is_view or tbl_md.view_md.predicate is None else tbl_md.view_md.predicate
         self.predicate = exprs.Expr.from_dict(predicate_dict) if predicate_dict is not None else None
-        self.mutable_views: List[TableVersion] = []  # targets for update propagation
+        self.mutable_views = []  # targets for update propagation
         if self.base is not None and not self.base.is_snapshot and not self.is_snapshot:
             self.base.mutable_views.append(self)
 
         # component view-specific initialization
-        self.iterator_cls: Optional[Type[ComponentIterator]] = None
-        self.iterator_args: Optional[exprs.InlineDict] = None
+        self.iterator_cls = None
+        self.iterator_args = None
         self.num_iterator_cols = 0
         if is_view and tbl_md.view_md.iterator_class_fqn is not None:
             module_name, class_name = tbl_md.view_md.iterator_class_fqn.rsplit('.', 1)
@@ -119,12 +153,12 @@ class TableVersion:
         cat.tbl_versions[(self.id, self.effective_version)] = self
 
         # init schema after we determined whether we're a component view, and before we create the store table
-        self.cols: list[Column] = []  # contains complete history of columns, incl dropped ones
-        self.cols_by_name: dict[str, Column] = {}  # contains only user-facing (named) columns visible in this version
-        self.cols_by_id: dict[int, Column] = {}  # contains only columns visible in this version, both system and user
-        self.idx_md = tbl_md.index_md  # needed for _create_tbl_md()
-        self.idxs_by_name: dict[str, TableVersion.IndexInfo] = {}  # contains only actively maintained indices
-        self.external_stores: dict[str, pxt.io.ExternalStore] = {}
+        self.cols = []
+        self.cols_by_name = {}
+        self.cols_by_id = {}
+        self.idx_md = tbl_md.index_md
+        self.idxs_by_name = {}
+        self.external_stores = {}
 
         self._init_schema(tbl_md, schema_version_md)
 
@@ -719,8 +753,9 @@ class TableVersion:
             plan, updated_cols, recomputed_cols = (
                 Planner.create_update_plan(self.path, update_spec, [], where, cascade)
             )
+            from pixeltable.exprs import SqlElementCache
             result = self.propagate_update(
-                plan, where.sql_expr() if where is not None else None, recomputed_cols,
+                plan, where.sql_expr(SqlElementCache()) if where is not None else None, recomputed_cols,
                 base_versions=[], conn=conn, timestamp=time.time(), cascade=cascade, show_progress=True)
             result.updated_cols = updated_cols
             return result
@@ -802,7 +837,7 @@ class TableVersion:
         return update_targets
 
     def propagate_update(
-            self, plan: Optional[exec.ExecNode], where_clause: Optional[sql.ClauseElement],
+            self, plan: Optional[exec.ExecNode], where_clause: Optional[sql.ColumnElement],
             recomputed_view_cols: List[Column], base_versions: List[Optional[int]], conn: sql.engine.Connection,
             timestamp: float, cascade: bool, show_progress: bool = True
     ) -> UpdateStatus:
@@ -843,6 +878,7 @@ class TableVersion:
         assert self.is_insertable()
         from pixeltable.exprs import Expr
         from pixeltable.plan import Planner
+        sql_where_clause: Optional[Expr] = None
         if where is not None:
             if not isinstance(where, Expr):
                 raise excs.Error(f"'where' argument must be a predicate, got {type(where)}")
@@ -850,10 +886,10 @@ class TableVersion:
             # for now we require that the updated rows can be identified via SQL, rather than via a Python filter
             if analysis_info.filter is not None:
                 raise excs.Error(f'Filter {analysis_info.filter} not expressible in SQL')
+            sql_where_clause = analysis_info.sql_where_clause
 
-        analysis_info = Planner.analyze(self.path, where)
         with Env.get().engine.begin() as conn:
-            num_rows = self.propagate_delete(analysis_info.sql_where_clause, base_versions=[], conn=conn, timestamp=time.time())
+            num_rows = self.propagate_delete(sql_where_clause, base_versions=[], conn=conn, timestamp=time.time())
 
         status = UpdateStatus(num_rows=num_rows)
         return status
@@ -867,7 +903,8 @@ class TableVersion:
         Returns:
             number of deleted rows
         """
-        sql_where_clause = where.sql_expr() if where is not None else None
+        from pixeltable.exprs import SqlElementCache
+        sql_where_clause = where.sql_expr(SqlElementCache()) if where is not None else None
         num_rows = self.store_tbl.delete_rows(
             self.version + 1, base_versions=base_versions, match_on_vmin=False, where_clause=sql_where_clause,
             conn=conn)
--- a/pixeltable/exec/aggregation_node.py
+++ b/pixeltable/exec/aggregation_node.py
@@ -21,8 +21,9 @@ class AggregationNode(ExecNode):
         self.input = input
         self.group_by = group_by
         self.input_exprs = list(input_exprs)
-        self.agg_fn_calls = agg_fn_calls
         self.agg_fn_eval_ctx = row_builder.create_eval_ctx(agg_fn_calls, exclude=input_exprs)
+        # we need to make sure to refer to the same exprs that RowBuilder.eval() will use
+        self.agg_fn_calls = self.agg_fn_eval_ctx.target_exprs
         self.output_batch = DataRowBatch(tbl, row_builder, 0)
 
     def _reset_agg_state(self, row_num: int) -> None:
--- a/pixeltable/exec/component_iteration_node.py
+++ b/pixeltable/exec/component_iteration_node.py
@@ -19,7 +19,7 @@ class ComponentIterationNode(ExecNode):
         super().__init__(input.row_builder, [], [], input)
         self.view = view
         iterator_args = [view.iterator_args.copy()]
-        self.row_builder.substitute_exprs(iterator_args)
+        self.row_builder.set_slot_idxs(iterator_args)
         self.iterator_args = iterator_args[0]
         assert isinstance(self.iterator_args, exprs.InlineDict)
         self.iterator_args_ctx = self.row_builder.create_eval_ctx([self.iterator_args])
--- a/pixeltable/exec/sql_node.py
+++ b/pixeltable/exec/sql_node.py
@@ -35,7 +35,7 @@ class SqlNode(ExecNode):
         self.sql_exprs = exprs.ExprSet(select_list)
         # unstored iter columns: we also need to retrieve whatever is needed to materialize the iter args
         for iter_arg in row_builder.unstored_iter_args.values():
-            sql_subexprs = iter_arg.subexprs(filter=lambda e: sql_elements[e] is not None, traverse_matches=False)
+            sql_subexprs = iter_arg.subexprs(filter=sql_elements.contains, traverse_matches=False)
             for e in sql_subexprs:
                 self.sql_exprs.add(e)
         super().__init__(row_builder, self.sql_exprs, [], None)  # we materialize self.sql_exprs
--- a/pixeltable/exec/sql_node.py
+++ b/pixeltable/exec/sql_node.py
@@ -45,7 +45,7 @@ class SqlNode(ExecNode):
         for rowid_ref in [e for e in self.sql_exprs if isinstance(e, exprs.RowidRef)]:
             rowid_ref.set_tbl(tbl)
 
-        sql_select_list = [sql_elements[e] for e in self.sql_exprs]
+        sql_select_list = [sql_elements.get(e) for e in self.sql_exprs]
         assert len(sql_select_list) == len(self.sql_exprs)
         assert all(e is not None for e in sql_select_list)
         self.set_pk = set_pk
@@ -232,10 +232,10 @@ class SqlScanNode(SqlNode):
             if isinstance(e, exprs.SimilarityExpr):
                 order_by_clause.append(e.as_order_by_clause(asc))
             else:
-                order_by_clause.append(sql_elements[e].desc() if not asc else sql_elements[e])
+                order_by_clause.append(sql_elements.get(e).desc() if not asc else sql_elements.get(e))
 
         if where_clause is not None:
-            sql_where_clause = sql_elements[where_clause]
+            sql_where_clause = sql_elements.get(where_clause)
             assert sql_where_clause is not None
             self.stmt = self.stmt.where(sql_where_clause)
         if len(order_by_clause) > 0:
--- a/pixeltable/exec/sql_node.py
+++ b/pixeltable/exec/sql_node.py
@@ -18,7 +18,7 @@ class SqlNode(ExecNode):
 
     def __init__(
             self, tbl: catalog.TableVersionPath, row_builder: exprs.RowBuilder,
-            select_list: Iterable[exprs.Expr], set_pk: bool = False
+            select_list: Iterable[exprs.Expr], sql_elements: exprs.SqlElementCache, set_pk: bool = False
     ):
         """
         Initialize self.stmt with expressions derived from select_list.
@@ -35,8 +35,9 @@ class SqlNode(ExecNode):
         self.sql_exprs = exprs.ExprSet(select_list)
         # unstored iter columns: we also need to retrieve whatever is needed to materialize the iter args
         for iter_arg in row_builder.unstored_iter_args.values():
-            sql_subexprs = iter_arg.subexprs(filter=lambda e: e.sql_expr() is not None, traverse_matches=False)
-            [self.sql_exprs.append(e) for e in sql_subexprs]
+            sql_subexprs = iter_arg.subexprs(filter=lambda e: sql_elements[e] is not None, traverse_matches=False)
+            for e in sql_subexprs:
+                self.sql_exprs.add(e)
         super().__init__(row_builder, self.sql_exprs, [], None)  # we materialize self.sql_exprs
 
         # change rowid refs against a base table to rowid refs against the target table, so that we minimize
@@ -44,7 +45,7 @@ class SqlNode(ExecNode):
         for rowid_ref in [e for e in self.sql_exprs if isinstance(e, exprs.RowidRef)]:
             rowid_ref.set_tbl(tbl)
 
-        sql_select_list = [e.sql_expr() for e in self.sql_exprs]
+        sql_select_list = [sql_elements[e] for e in self.sql_exprs]
         assert len(sql_select_list) == len(self.sql_exprs)
         assert all(e is not None for e in sql_select_list)
         self.set_pk = set_pk
@@ -204,7 +205,8 @@ class SqlScanNode(SqlNode):
             set_pk: if True, sets the primary for each DataRow
             exact_version_only: tables for which we only want to see rows created at the current version
         """
-        super().__init__(tbl, row_builder, select_list, set_pk=set_pk)
+        sql_elements = exprs.SqlElementCache()
+        super().__init__(tbl, row_builder, select_list, sql_elements, set_pk=set_pk)
         # create Select stmt
         if order_by_items is None:
             order_by_items = []
@@ -230,10 +232,10 @@ class SqlScanNode(SqlNode):
             if isinstance(e, exprs.SimilarityExpr):
                 order_by_clause.append(e.as_order_by_clause(asc))
             else:
-                order_by_clause.append(e.sql_expr().desc() if not asc else e.sql_expr())
+                order_by_clause.append(sql_elements[e].desc() if not asc else sql_elements[e])
 
         if where_clause is not None:
-            sql_where_clause = where_clause.sql_expr()
+            sql_where_clause = sql_elements[where_clause]
             assert sql_where_clause is not None
             self.stmt = self.stmt.where(sql_where_clause)
         if len(order_by_clause) > 0:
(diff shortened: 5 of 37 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/278)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/278)* status: 0 of 37 files reviewed, 15 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/plan.py` line 19 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Iq2ic1bmjjgZNN3yy:-O6Iq2ic1bmjjgZNN3yz:b862r31) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L19)):*
> ```Python
> 
> def _get_combined_ordering(
>         o1: list[tuple[exprs.Expr, bool]], o2: list[tuple[exprs.Expr, bool]]
> ```

Also remove `List` and `Tuple` as imports, if you're lowercasing them throughout a file
___
*[`pixeltable/plan.py` line 84 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6RfRAe-Uyd5dhTG8Gr:-O6RfRAe-Uyd5dhTG8Gs:bg43m6w) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L84)):*
> ```Python
>         self.order_by_clause = [(e.resolve_computed_cols(), asc) for e, asc in order_by_clause]
> 
>         # not executable
> ```

is this comment still relevant?
___
*[`pixeltable/plan.py` line 89 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rg2ku6Iv3sGsGTCag:-O6Rg2ku6Iv3sGsGTCah:bita6f3) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L89)):*
> ```Python
>         if where_clause is not None:
>             where_clause_conjuncts, self.filter = where_clause.split_conjuncts(
>                 lambda e: self.sql_elements[e] is not None)
> ```

The expression `sql_elements[e] is not None` gets used a lot; I think it'd be clearer to capture it with a method, say, `sql_elements.has_sql(e)`. Bonus is that we could (I think?) get rid of the lambda (here and elsewhere), as in
```python
where_clause.split_conjuncts(self.sql_elements.has_sql)
```
___
*[`pixeltable/plan.py` line 209 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Iqc0I5L0PMgnV85QV:-O6Iqc0I5L0PMgnV85QW:b-95nj4x) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L209)):*
> ```Python
>             cls, tbl: catalog.TableVersionPath, where_clause: Optional[exprs.Expr] = None
>     ) -> sql.Select:
>         stmt = sql.select(sql.func.count('*'))
> ```

why do we need a `'*'` here? feels like a contrived attempt to mimic SQL in python. the parameterless `count()` is more natural to me
___
*[`pixeltable/exprs/expr.py` line 373 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rh4EtE2NTJPTS1N_d:-O6Rh4EtE2NTJPTS1N_e:btlyanu) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/expr.py#L373)):*
> ```Python
> 
>     @abc.abstractmethod
>     def sql_expr(self, sql_elements: 'pixeltable.exprs.SqlElementCache') -> Optional[sql.ColumnElement]:
> ```

This will give a type error if `pixeltable` is not imported. One solution would be to put this at the top of the file:
```python
if TYPE_CHECKING:
    from pixeltable import exprs
```
Then you can make this `sql_elements: 'exprs.SqlElementCache'`.
___
*[`pixeltable/exprs/row_builder.py` line 236 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6RkC2725p68IgsYQ4j:-O6RkC2725p68IgsYQ4k:bwtrtyu) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/row_builder.py#L236)):*
> ```Python
>         if remove_duplicates:
>             seen: set[int] = set()  # seen Expr.ids
>             deduped = [e for e in expr_list if e.id not in seen and not seen.add(e.id)]
> ```

btw a standard way of doing this is
```
deduped = list({a: None for a in expr_list}.keys())
```
which will efficiently dedup the list in an order preserving way.

kinda sad that the Python API doesn't include a `list.unique()` method or some such.
___
*[`pixeltable/exprs/row_builder.py` line 237 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rjh_h2t0aZeLxcUab:-O6Rjh_h2t0aZeLxcUac:bjmaaxe) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/row_builder.py#L237)):*
> ```Python
>             seen: set[int] = set()  # seen Expr.ids
>             deduped = [e for e in expr_list if e.id not in seen and not seen.add(e.id)]
>             expr_list[:] = deduped
> ```

we're modifying the parameter in place? i think this should be called out in the function comment for clarity
___
*[`pixeltable/exprs/row_builder.py` line 239 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6RjORJ2c7bs4gCwYeF:-O6RjORJ2c7bs4gCwYeG:btwbyst) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/row_builder.py#L239)):*
> ```Python
>             expr_list[:] = deduped
> 
>     def _set_slot_idxs_aux(self, e: Expr) -> None:
> ```

use double underscore: this is definitely a private method

i generally prefer `__set_slot_idxs_r` in this situation (`aux` is kinda weird) but don't feel strongly about it
___
*[`pixeltable/exprs/sql_element_cache.py` line 8 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6IrlXc2keV1jA1Jxg1:-O6IrlXc2keV1jA1Jxg2:bectscl) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/sql_element_cache.py#L8)):*
> ```Python
> 
> 
> class SqlElementCache:
> ```

The name `SqlElementCache` seems slightly misleading to me, since it's driving the actual evaluation, and part of its purpose is to capture separate evaluation contexts (whereas "cache" suggests we're temporarily storing some invariant for efficiency).

Maybe `SqlContext`?
___
*[`pixeltable/exprs/sql_element_cache.py` line 11 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6RulfGESfF67Lc768o:-O6RulfGESfF67Lc768p:b-6zeh03) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/sql_element_cache.py#L11)):*
> ```Python
>     """Cache of sql.ColumnElements for exprs"""
> 
>     cache: dict[int, sql.ColumnElement]  # key: Expr.id
> ```

shouldn't this be a `dict[int, Optional[sql.ColumnElement]]` ?
___
*[`pixeltable/func/function.py` line 91 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Is07QBA8sHCFchm46:-O6Is07QBA8sHCFchm47:b-yanj7k) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/func/function.py#L91)):*
> ```Python
>         pass
> 
>     def _to_sql(self, *args: Any, **kwargs: Any) -> Optional['sqlalchemy.ColumnElement']:
> ```

This should probably just be a member variable rather than a method `def`. Defining a method via `def` and then later overriding it with an assignment is pretty confusing.

That is:

```
__sql_conversion_fn: Callable[..., Optional[sql.ColumnElement]] = None
```
and then:
```
def to_sql(...):
    self.__sql_conversion_fn = fn
```

See how `_conditional_return_type` is implemented, for example.
___
*[`pixeltable/func/function.py` line 100 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6IsoE59rffj_NA2yJG:-O6IsoE59rffj_NA2yJH:b723v93) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/func/function.py#L100)):*
> ```Python
> 
>     def to_sql(
>             self, fn: Callable[..., Optional['sqlalchemy.ColumnElement']]
> ```

The quoted type shouldn't be necessary here; just put `import sqlalchemy as sql` at top of file and use a proper type reference.
___
*[`pixeltable/functions/timestamp.py` line 33 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6ItwRe2dzu0tQmpzbv:-O6ItwRe2dzu0tQmpzbw:b-g9uul3) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/functions/timestamp.py#L33)):*
> ```Python
> 
> 
> @year.to_sql
> ```

Every function that has a `to_sql` should have a test that executes it two ways: first in an expression that will be converted to SQL, then in an expression that can only be evaluated in Python.

(in this case, for example, one might enhance `test_all` so that in addition to calling `pxt_fn(t.dt, ...)`, we also call something like `pxt_fn(t.dt.apply(lambda x: x))`.)
___
*[`pixeltable/functions/timestamp.py` line 35 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6It1OiCSc3TdYtn3tC:-O6It1OiCSc3TdYtn3tD:bx3kmu9) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/functions/timestamp.py#L35)):*
> ```Python
> @year.to_sql
> def _(self: sql.ColumnElement) -> sql.ColumnElement:
>     import sqlalchemy as sql
> ```

spurious import (here and elsewhere) - it's already imported at top of file
___
*[`pixeltable/functions/timestamp.py` line 50 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6ItVipD9iTOPa1mBnP:-O6ItVipD9iTOPa1mBnQ:b-jwotfy) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/functions/timestamp.py#L50)):*
> ```Python
> 
> @month.to_sql
> def _(self: 'sqlalchemy.ColumnElement') -> 'sqlalchemy.ColumnElement':
> ```

unnecessary quoted type (here and elsewhere)

this will actually give a type error as printed, since `sqlalchemy` was imported as `sql`

(in fact, i'd expect it to generate an inline warning in pycharm, unless you've disabled them)

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/278)* status: 0 of 37 files reviewed, 14 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/plan.py` line 19 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Iq2ic1bmjjgZNN3yy:-O6SLJRO8xC8V13BPQVR:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L19)):*

Also remove `List` and `Tuple` as imports, if you're lowercasing them throughout a file
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 84 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6RfRAe-Uyd5dhTG8Gr:-O6SLdTN0MNaIYbntdTo:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L84)):*

is this comment still relevant?
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 89 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rg2ku6Iv3sGsGTCag:-O6SN2cp6Y7cEB924tYw:b1kirrj) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L89)):*

The expression `sql_elements[e] is not None` gets used a lot; I think it'd be clearer to capture it with a method, say, `sql_elements.has_sql(e)`. Bonus is that we could (I think?) get rid of the lambda (here and elsewhere), as in
```python
where_clause.split_conjuncts(self.sql_elements.has_sql)
```
</blockquote></details>

replaced with contains()
___
*[`pixeltable/plan.py` line 209 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Iqc0I5L0PMgnV85QV:-O6SNknw9NBMgfUbHddj:b-9dqgpn) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L209)):*

why do we need a `'*'` here? feels like a contrived attempt to mimic SQL in python. the parameterless `count()` is more natural to me
</blockquote></details>

not needed, removed
___
*[`pixeltable/exprs/expr.py` line 373 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rh4EtE2NTJPTS1N_d:-O6SOXdS7g20Yp8G2KTr:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/expr.py#L373)):*

This will give a type error if `pixeltable` is not imported. One solution would be to put this at the top of the file:
```python
if TYPE_CHECKING:
    from pixeltable import exprs
```
Then you can make this `sql_elements: 'exprs.SqlElementCache'`.
</blockquote></details>

Done.
___
*[`pixeltable/exprs/row_builder.py` line 236 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6RkC2725p68IgsYQ4j:-O6SUsijASe-E4uStBvd:b-wi1ni4) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/row_builder.py#L236)):*

btw a standard way of doing this is
```
deduped = list({a: None for a in expr_list}.keys())
```
which will efficiently dedup the list in an order preserving way.

kinda sad that the Python API doesn't include a `list.unique()` method or some such.
</blockquote></details>

We already have ExprSet...
___
*[`pixeltable/exprs/row_builder.py` line 237 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rjh_h2t0aZeLxcUab:-O6SPkEC2s87MHK41zSK:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/row_builder.py#L237)):*

we're modifying the parameter in place? i think this should be called out in the function comment for clarity
</blockquote></details>

Done.
___
*[`pixeltable/exprs/row_builder.py` line 239 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6RjORJ2c7bs4gCwYeF:-O6SPuzg6myK2R6Yp4s0:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/row_builder.py#L239)):*

use double underscore: this is definitely a private method

i generally prefer `__set_slot_idxs_r` in this situation (`aux` is kinda weird) but don't feel strongly about it
</blockquote></details>

Done.
___
*[`pixeltable/exprs/sql_element_cache.py` line 8 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6IrlXc2keV1jA1Jxg1:-O6SQDOV0RuqMJV6PTSu:b5fzscn) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/sql_element_cache.py#L8)):*

The name `SqlElementCache` seems slightly misleading to me, since it's driving the actual evaluation, and part of its purpose is to capture separate evaluation contexts (whereas "cache" suggests we're temporarily storing some invariant for efficiency).

Maybe `SqlContext`?
</blockquote></details>

I was thinking about that, but SqlContext seems a bit grandiose and vague. How about we defer this for now and see how it evolves?
___
*[`pixeltable/exprs/sql_element_cache.py` line 11 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6RulfGESfF67Lc768o:-O6SQ1tcBsPM8UVa5jJp:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/sql_element_cache.py#L11)):*

shouldn't this be a `dict[int, Optional[sql.ColumnElement]]` ?
</blockquote></details>

Done.
___
*[`pixeltable/func/function.py` line 91 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Is07QBA8sHCFchm46:-O6SSaIY6gm3U1blz2ii:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/func/function.py#L91)):*

This should probably just be a member variable rather than a method `def`. Defining a method via `def` and then later overriding it with an assignment is pretty confusing.

That is:

```
__sql_conversion_fn: Callable[..., Optional[sql.ColumnElement]] = None
```
and then:
```
def to_sql(...):
    self.__sql_conversion_fn = fn
```

See how `_conditional_return_type` is implemented, for example.
</blockquote></details>

Done.
___
*[`pixeltable/func/function.py` line 100 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6IsoE59rffj_NA2yJG:-O6SROeI7QZOOpzDirsG:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/func/function.py#L100)):*

The quoted type shouldn't be necessary here; just put `import sqlalchemy as sql` at top of file and use a proper type reference.
</blockquote></details>

Done.
___
*[`pixeltable/functions/timestamp.py` line 35 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6It1OiCSc3TdYtn3tC:-O6SSqJcCXZvUGW6Ejw7:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/functions/timestamp.py#L35)):*

spurious import (here and elsewhere) - it's already imported at top of file
</blockquote></details>

Done.
___
*[`pixeltable/functions/timestamp.py` line 50 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6ItVipD9iTOPa1mBnP:-O6STAs408ZGVDqcBftJ:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/functions/timestamp.py#L50)):*

unnecessary quoted type (here and elsewhere)

this will actually give a type error as printed, since `sqlalchemy` was imported as `sql`

(in fact, i'd expect it to generate an inline warning in pycharm, unless you've disabled them)
</blockquote></details>

Done.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/278)* status: 0 of 37 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/plan.py` line 89 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rg2ku6Iv3sGsGTCag:-O6XyXi_DxilY2at_OtJ:b-d1r3fw) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L89)):*

replaced with contains()
</blockquote></details>

`contains()` is a little misleading, because its normal interpretation is "is contained in the cache". But an Expr could be contained in the cache, but still return None, because it previously had its sql_expr() evaluated, with a result of None. `has_sql()` seems clearer to me (or really, anything that clearly says "this Expr returned a positive result when it was converted to sql."
___
*[`pixeltable/exec/sql_node.py` line 38 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Xzz0UCTGOGVmLxqCA:-O6Xzz0UCTGOGVmLxqCB:bt2o3d1) ([raw file](https://github.com/pixeltable/pixeltable/blob/8d9f0d070f8e8b9b93f1df5f0209c5af94639569/pixeltable/exec/sql_node.py#L38)):*
> ```Python
>         # unstored iter columns: we also need to retrieve whatever is needed to materialize the iter args
>         for iter_arg in row_builder.unstored_iter_args.values():
>             sql_subexprs = iter_arg.subexprs(filter=lambda e: sql_elements[e] is not None, traverse_matches=False)
> ```

Replace lambda with sql_elements.has_sql (or whatever we want to call it)
___
*[`pixeltable/exprs/sql_element_cache.py` line 8 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6IrlXc2keV1jA1Jxg1:-O6az7VBB8jD_DqhdLm7:b32wtd) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/exprs/sql_element_cache.py#L8)):*

I was thinking about that, but SqlContext seems a bit grandiose and vague. How about we defer this for now and see how it evolves?
</blockquote></details>

ok
___
*[`pixeltable/func/function.py` line 46 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6azvE_CgHhqRV8BqlI:-O6azTwUEfi9GvNqBS5p:b-ykpsm1) ([raw file](https://github.com/pixeltable/pixeltable/blob/8d9f0d070f8e8b9b93f1df5f0209c5af94639569/pixeltable/func/function.py#L46)):*
> ```Python
>         self.is_property = is_property
>         self._conditional_return_type = None
>         self._to_sql = self.__to_sql
> ```

Why isn't the default `None`? That seems more logical to me than this pattern. Yes it would involve an extra line of code when `_to_sql` gets called, but I think that'd be clearer to anyone reading the code.
___
*[`pixeltable/func/function.py` line 111 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6b-8X04JYQEYNtc3Mj:-O6b-8X04JYQEYNtc3Mk:bgn92bg) ([raw file](https://github.com/pixeltable/pixeltable/blob/8d9f0d070f8e8b9b93f1df5f0209c5af94639569/pixeltable/func/function.py#L111)):*
> ```Python
>         return fn
> 
>     def __to_sql(self, *args: Any, **kwargs: Any) -> Optional[sql.ColumnElement]:
> ```

See above comment - I think we should get rid of this and have the default `_to_sql` be simply `None`.

If you feel strongly about keeping it, better to call it `__default_to_sql` or some such, and then the type of `_to_sql` is non-Optional.

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/278)* status: 0 of 37 files reviewed, 10 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/plan.py` line 89 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rg2ku6Iv3sGsGTCag:-O6b7vT0CvUQj0gvAva2:b-rlsko0) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L89)):*

`contains()` is a little misleading, because its normal interpretation is "is contained in the cache". But an Expr could be contained in the cache, but still return None, because it previously had its sql_expr() evaluated, with a result of None. `has_sql()` seems clearer to me (or really, anything that clearly says "this Expr returned a positive result when it was converted to sql."
</blockquote></details>

This is going with the customary interpretation of 'None' meaning 'doesn't have it' (as in dict.get()). The cache simply remembers the absence of a ColumnElement by placing an entry in a physical dict, but that's really internal to SqlElementCache.

I added comments to SqlElementCache.
___
*[`pixeltable/exec/sql_node.py` line 38 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Xzz0UCTGOGVmLxqCA:-O6b9N-T6RyAZWoglzD6:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/8d9f0d070f8e8b9b93f1df5f0209c5af94639569/pixeltable/exec/sql_node.py#L38)):*

Replace lambda with sql_elements.has_sql (or whatever we want to call it)
</blockquote></details>

Done.
___
*[`pixeltable/func/function.py` line 46 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6azvE_CgHhqRV8BqlI:-O6b9VXC7TsFC7EhaQlm:b-bw0n6c) ([raw file](https://github.com/pixeltable/pixeltable/blob/8d9f0d070f8e8b9b93f1df5f0209c5af94639569/pixeltable/func/function.py#L46)):*

Why isn't the default `None`? That seems more logical to me than this pattern. Yes it would involve an extra line of code when `_to_sql` gets called, but I think that'd be clearer to anyone reading the code.
</blockquote></details>

I think that's more error-prone. I find this completely clear, but added a comment to __to_sql() to clarify that it's the default implementation.
___
*[`pixeltable/func/function.py` line 111 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6b-8X04JYQEYNtc3Mj:-O6bA_-W-jc-xKes91ok:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/8d9f0d070f8e8b9b93f1df5f0209c5af94639569/pixeltable/func/function.py#L111)):*

See above comment - I think we should get rid of this and have the default `_to_sql` be simply `None`.

If you feel strongly about keeping it, better to call it `__default_to_sql` or some such, and then the type of `_to_sql` is non-Optional.
</blockquote></details>

Done.
___
*[`pixeltable/functions/timestamp.py` line 33 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6ItwRe2dzu0tQmpzbv:-O6bCqYz2E6HDBv04-7U:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/functions/timestamp.py#L33)):*

Every function that has a `to_sql` should have a test that executes it two ways: first in an expression that will be converted to SQL, then in an expression that can only be evaluated in Python.

(in this case, for example, one might enhance `test_all` so that in addition to calling `pxt_fn(t.dt, ...)`, we also call something like `pxt_fn(t.dt.apply(lambda x: x))`.)
</blockquote></details>

Done.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/278)* status: 0 of 37 files reviewed, 3 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/plan.py` line 89 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rg2ku6Iv3sGsGTCag:-O6bNDBC64umJa6CIhpU:b-88he9v) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L89)):*

This is going with the customary interpretation of 'None' meaning 'doesn't have it' (as in dict.get()). The cache simply remembers the absence of a ColumnElement by placing an entry in a physical dict, but that's really internal to SqlElementCache.

I added comments to SqlElementCache.
</blockquote></details>

```
d = {'a': None}
```
Now `'a' in d` returns `True`, but `'b' in d` returns `False`. I'd argue `contains()` implies semantics that mirror `in`, not `get()`.

aaron-siegel (approved):

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/278)* status: 0 of 37 files reviewed, 3 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/plan.py` line 89 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/278#-O6Rg2ku6Iv3sGsGTCag:-O6coglO2Nb2euofHGzF:b-nf90r4) ([raw file](https://github.com/pixeltable/pixeltable/blob/d11569132f93263b6afe2d26776f62add3c8d468/pixeltable/plan.py#L89)):*

```
d = {'a': None}
```
Now `'a' in d` returns `True`, but `'b' in d` returns `False`. I'd argue `contains()` implies semantics that mirror `in`, not `get()`.
</blockquote></details>

I changed SqlElementCache.__getitem__ to SqlElementCache.get

PR #282, opened 2024-09-17
Make `mistune` an optional dependency

aaron-siegel:
Converts `mistune` into an optional (dev) dependency, fixing the pip errors in colab.
Also includes two other changes:
- Makes the error messages for `require_package()` more informative
- When `require_package()` is called, double-checks whether the package has been
installed. This ensures that after a "library not found" error, the user can `pip install` the
dependency and re-run the Pixeltable operation without having to restart their Python
session.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/282)

mkornacker (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/282)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @aaron-siegel)

___
*[`pixeltable/env.py` line 646 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/282#-O71JaxY1mFWlMI50-Q3:-O71JaxY1mFWlMI50-Q4:boyy9wi) ([raw file](https://github.com/pixeltable/pixeltable/blob/2e4a74909858be733a1886a2797a95ee58b31714/pixeltable/env.py#L646)):*
> ```Python
>     is_installed: bool
>     library_name: str
>     version: Optional[list[int]] = None
> ```

comment: what does this list represent?

PR #250, opened 2024-08-15
Time Zone Support

aaron-siegel:

PR #283, opened 2024-09-18
Fix circularity in DB migration for schema version 19 -> 20

aaron-siegel:
PR #250 introduced an issue where database migration could enter an infinite loop in certain situations, because it was trying to use the catalog to drive the migration before the catalog had been fully loaded. This issue was not caught in testing because the unit tests uses a different catalog initialization mechanism (`reload_catalog()`).

This fix changes the migration script to rely only on literal metadata, and not on the Catalog, to drive the migration.

TESTING DONE: Manually create a v19 Pixeltable database and test the upgrade on that installation. Verify that the migration tests still pass with the new migration script.

PRODUCTION IMPACT: None. (PR #250 has not gone into production yet.)

FOLLOW-UP WORK: Consider adding additional tests that run the DB migrations on fresh Pixeltable instances in separate Python processes.

### DIFF
--- a/pixeltable/metadata/converters/convert_19.py
+++ b/pixeltable/metadata/converters/convert_19.py
@@ -1,35 +1,34 @@
 import datetime
-from typing import TYPE_CHECKING, Any, Optional
+from typing import Any, Optional
 
 import sqlalchemy as sql
 
 import pixeltable as pxt
-from pixeltable.metadata import register_converter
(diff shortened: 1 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/283)

mkornacker (approved):

PR #294, opened 2024-09-29
Add UUID Type to Pixeltable Type System

denise-k:
### PR Summary: Add UUID Type to Pixeltable Type System

**Description:**

This pull request introduces a new `UuidType` to the Pixeltable type system to provide native support for UUIDs.

**Changes:**

1. **New UUID Type:**
 
   - [x]  Added a new `UuidType` class extending `ColumnType` to manage UUID values.
     - [x] Implemented the `to_sa_type()` method to map the UUID type to SQLAlchemy's `sql.UUID()`.
     - [x] Implemented the `_validate_literal()` method to ensure the value is an instance of `uuid.UUID`.
     - [x] Implemented the `_create_literal()` method to support creating UUIDs from strings, bytes, or integers.

2. **Type Enum Update:**
   - [x] Updated the `ColumnType.Type` enum to include a new `UUID` type.

3. **Type Factory Method Update:**
   - [x] Updated the `make_type` method in `ColumnType` to handle the new `UUID` type by returning an instance of `UuidType`.

4. **Test:**
   - [x] Added a unit test to `test_types.py` to ensure that `UuidType` behaves as expected.

### DIFF
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -7,6 +7,7 @@ import json
 import typing
 import urllib.parse
 import urllib.request
+import uuid
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Mapping, Optional, Sequence, Tuple, Union
 
@@ -33,6 +34,7 @@ class ColumnType:
         VIDEO = 8
         AUDIO = 9
         DOCUMENT = 10
+        UUID = 11
 
         # exprs that don't evaluate to a computable value in Pixeltable, such as an Image member function
         INVALID = 255
@@ -162,6 +164,8 @@ class ColumnType:
             return AudioType()
         if t == cls.Type.DOCUMENT:
             return DocumentType()
+        if t == cls.Type.UUID:
+            return UuidType()
 
     def __str__(self) -> str:
         return self._type.name.lower()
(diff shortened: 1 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/294)

mkornacker:
Hi Denise,

Thanks for your contribution! Could you give us some more context on your use case of storing UUIDs in tables (and how treating them as strings doesn't work for that)?

denise-k:
Hi Marcel,

Happy to share a bit more! We use [combined-time UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier#Combined_Time-GUID) as primary/secondary keys in our tables. The UUID built-in would be nice for pixeltable because:
- Stricter input validation gives us peace of mind.
- It helps us keep a standardized output form for UUIDs when coercing to str (lowercase letters, hyphens in the right place, etc).
- There are several built-ins for accessing UUID as an int, or bytes, which would be easier than having to cast multiple times.

I'm sure all of these could be built around a `StringType` too, just thought the `UuidType` might be helpful for someone else in the future.

PR #307, opened 2024-10-01
Allow Nones as input to iterators (and skip over rows with Nones)

aaron-siegel:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/307)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/307)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exec/component_iteration_node.py` line 48 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/307#-O8DtBz8EsK2p8JoRXXG:-O8DtBz8EsK2p8JoRXXH:b-kq31a2) ([raw file](https://github.com/pixeltable/pixeltable/blob/1d8f2343241b1892687670ccddff82690ad48c89/pixeltable/exec/component_iteration_node.py#L48)):*
> ```Python
>                 # specified and are not null. If any of them are null, then we skip this row (i.e., we emit 0
>                 # output rows for this input row).
>                 if self.__all_required_args_specified(iterator_args):
> ```

to avoid excessive indentation:
if not ...:
    continue
___
*[`pixeltable/exec/component_iteration_node.py` line 84 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/307#-O8DsWOqAUiF9-VDsz6_:-O8DsWOqAUiF9-VDsz6a:b-ticcl4) ([raw file](https://github.com/pixeltable/pixeltable/blob/1d8f2343241b1892687670ccddff82690ad48c89/pixeltable/exec/component_iteration_node.py#L84)):*
> ```Python
>         """Returns true if all required (non-nullable) iterator arguments are specified and not null."""
>         for param_name, param_type in self.view.iterator_cls.input_schema().items():
>             if not param_type.nullable and iterator_args.get(param_name, None) is None:
> ```

This isn't quite correct: init() could specify a default value for a non-null parameter.

Probably best to call inspect.Signature.bind() to get the actual bindings.

Would also be good to cover that with a test case.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/307)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exec/component_iteration_node.py` line 48 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/307#-O8DtBz8EsK2p8JoRXXG:-O8ELQak95nTSpvh8mTD:b-eab5m2) ([raw file](https://github.com/pixeltable/pixeltable/blob/1d8f2343241b1892687670ccddff82690ad48c89/pixeltable/exec/component_iteration_node.py#L48)):*

to avoid excessive indentation:
if not ...:
    continue
</blockquote></details>

I prefer a more functional style; excessive use of control statements like `continue` complicates control flow and makes the logic harder to follow (whereas indentation is purely aesthetic).

I'd argue a better way to deal with excessive indentation would be to break up the function. I went ahead and did this; I think the result is clearer code AND less indentation.
___
*[`pixeltable/exec/component_iteration_node.py` line 84 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/307#-O8DsWOqAUiF9-VDsz6_:-O8Em__b5iBAd6udqOLW:b-aldlxm) ([raw file](https://github.com/pixeltable/pixeltable/blob/1d8f2343241b1892687670ccddff82690ad48c89/pixeltable/exec/component_iteration_node.py#L84)):*

This isn't quite correct: init() could specify a default value for a non-null parameter.

Probably best to call inspect.Signature.bind() to get the actual bindings.

Would also be good to cover that with a test case.
</blockquote></details>

Ah, good catch!

It's actually simpler than this; checking for required arguments & default resolution happen earlier in the flow (at iterator creation time). So all we need to do here is check whether any of the directly specified args are Nones.

Fixed.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/307)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exec/component_iteration_node.py` line 84 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/307#-O8DsWOqAUiF9-VDsz6_:-O8EmowG0jQxxNohLUqK:bp5qg1o) ([raw file](https://github.com/pixeltable/pixeltable/blob/1d8f2343241b1892687670ccddff82690ad48c89/pixeltable/exec/component_iteration_node.py#L84)):*

Ah, good catch!

It's actually simpler than this; checking for required arguments & default resolution happen earlier in the flow (at iterator creation time). So all we need to do here is check whether any of the directly specified args are Nones.

Fixed.
</blockquote></details>

(Also: I added an additional test case to `test_document.py` that captures this scenario, and refactored `test_document.py` somewhat in the process.)

mkornacker (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/307)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exec/component_iteration_node.py` line 69 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/307#-O8JNaU9-1t2vNR3H33V:-O8JNaU9-1t2vNR3H33W:bi1l674) ([raw file](https://github.com/pixeltable/pixeltable/blob/66528b99638e0f14e476bc7d732bf532b2f4a7bb/pixeltable/exec/component_iteration_node.py#L69)):*
> ```Python
>         """
>         Returns true if all non-nullable iterator arguments are not `None`.
>         """
> ```

input_schema = ..._cls.input_schema()
before loop, then use that
___
*[`tests/test_document.py` line 17 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/307#-O8JOEm6AbRo0rUpqWB-:-O8JOEm6AbRo0rUpqWB0:bxdjq3) ([raw file](https://github.com/pixeltable/pixeltable/blob/66528b99638e0f14e476bc7d732bf532b2f4a7bb/tests/test_document.py#L17)):*
> ```Python
> 
> 
> def _check_pdf_metadata(rec, sep1, metadata: list[str]):
> ```

Add missing type hints?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/307)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exec/component_iteration_node.py` line 69 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/307#-O8JNaU9-1t2vNR3H33V:-O8JvJB20ITfkAMJQtnz:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/66528b99638e0f14e476bc7d732bf532b2f4a7bb/pixeltable/exec/component_iteration_node.py#L69)):*

input_schema = ..._cls.input_schema()
before loop, then use that
</blockquote></details>

Done.
___
*[`tests/test_document.py` line 17 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/307#-O8JOEm6AbRo0rUpqWB-:-O8JvKQR58cP4W2T1D-Y:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/66528b99638e0f14e476bc7d732bf532b2f4a7bb/tests/test_document.py#L17)):*

Add missing type hints?
</blockquote></details>

Done.

PR #315, opened 2024-10-03
Fix Together AI image endpoint

aaron-siegel:
Together AI changed their image endpoint without warning. Instead of returning a base64-encoded image in the JSON response, their API now returns a URL pointing to the generated image. (The Together AI API docs still reference the old pattern.)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/315)

PR #319, opened 2024-10-04
Type-checking correctness for several additional packages.

aaron-siegel:
Type-checking correctness for `pixeltable.func`, `pixeltable.index`, `pixeltable.io`, `pixeltable.iterators`, `pixeltable.metadata`, and `pixeltable.util`.

This leaves only `exec`, `exprs`, and `tool` unchecked.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/319)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/319)* status: 0 of 43 files reviewed, 10 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/catalog/table.py` line 98 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SdWsk6u25wmLyOO_z:-O8SdWsk6u25wmLyOOa-:bu3jfai) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/catalog/table.py#L98)):*
> ```Python
> 
>     @overload
>     def __getitem__(self, index: Union[slice, exprs.Expr]) -> 'pxt.DataFrame': ...
> ```

Why slice?
___
*[`pixeltable/catalog/table.py` line 106 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8ShLr1C1m2G_niiwOu:-O8ShLr1C1m2G_niiwOv:bovbw0f) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/catalog/table.py#L106)):*
> ```Python
>             return getattr(self, index)
>         else:
>             return self._df()[index]
> ```

We should probably retire this altogether.
___
*[`pixeltable/func/query_template_function.py` line 62 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8Sf9qv7Tk4bFHafj1j:-O8Sf9qv7Tk4bFHafj1k:bcktrt4) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/func/query_template_function.py#L62)):*
> ```Python
>     # QueryTemplateFunction
>     def __getattr__(self, item: str) -> 'exprs.Expr':
>         raise AttributeError(f"'QueryTemplateFunction' object has no attribute '{item}'")
> ```

Maybe RuntimeError would be more appropriate here. This presumably should never be called.
___
*[`pixeltable/index/btree.py` line 51 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SfWvhDEOfzeekE9LW:-O8SfWvhDEOfzeekE9LX:b711tgi) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/index/btree.py#L51)):*
> ```Python
> 
> @udf
> def _str_filter(s: Optional[str]) -> Optional[str]:
> ```

Does it cause problems if it's part of BtreeIndex? This doesn't have meaning outside that class.
___
*[`pixeltable/io/external_store.py` line 244 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8ShZzqEI7bxrylSYr7:-O8ShZzqEI7bxrylSYr8:b-s41243) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/io/external_store.py#L244)):*
> ```Python
>                 )
>             col_ref = table[t_col]
>             assert isinstance(col_ref, exprs.ColumnRef)
> ```

Why assert and not cast()? (where should we use one vs the other?)
___
*[`pixeltable/metadata/schema.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8Si5sODTEjnStZrHCI:-O8Si5sODTEjnStZrHCJ:bdk7ku9) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L13)):*
> ```Python
> 
> Base: type = declarative_base()
> base_metadata = Base.metadata  # type: ignore[attr-defined]
> ```

Why is this needed?
___
*[`pixeltable/metadata/schema.py` line 29 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SiReJAlEOict5n33N:-O8SiReJAlEOict5n33O:b-gl8pem) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L29)):*
> ```Python
>             # Handling Optional types
>             non_none_args = [arg for arg in type_args if arg is not type(None)]
>             if len(non_none_args) == 1:
> ```

precede w/ assert len(...) == 1 and remove if
___
*[`pixeltable/metadata/schema.py` line 42 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SicB23pCqmyjV8Ckl:-O8SicB23pCqmyjV8Ckm:b-nr9va3) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L42)):*
> ```Python
>             return tuple(md_from_dict(arg_type, elem) for arg_type, elem in zip(type_args, data))  # type: ignore[return-value]
>         else:
>             assert False
> ```

Is this a logic error? Should this raise something?
___
*[`pixeltable/metadata/converters/util.py` line 71 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SittG6n1W1ZwNwR_l:-O8SittG6n1W1ZwNwR_m:b-g0afg) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/converters/util.py#L71)):*
> ```Python
> ) -> Any:
>     if isinstance(md, dict):
>         updated_dict = {}
> ```

type decl?
___
*[`pixeltable/utils/help.py` line 0 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SjSmG0GziCFKbcr7N:-O8SjSmG0GziCFKbcr7O:b-9gifm5) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/utils/help.py#L0)):*
What about help()?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/319)* status: 0 of 43 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/table.py` line 98 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SdWsk6u25wmLyOO_z:-O8cMzhiDycBTBnhEj66:b-alm5pe) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/catalog/table.py#L98)):*

Why slice?
</blockquote></details>

Ah good question. I guess I was reacting to the code comment ("return a DataFrame for the given slice"). Looking in `DataFrame`, I see that it actually takes a list or tuple, not a slice. Fixed.

Separately, why are `list` and `tuple` separate cases in the implementation of `DataFrame.__getitem__`? They behave identically, so it should probably just take a `Sequence` instead.
___
*[`pixeltable/catalog/table.py` line 106 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8ShLr1C1m2G_niiwOu:-O8cO_hKBYmhlujrr9OY:b-2hviaf) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/catalog/table.py#L106)):*

We should probably retire this altogether.
</blockquote></details>

Sure. I'm striving to avoid functional changes in these type-checking PRs though, so let's factor this out as a separate issue
___
*[`pixeltable/func/query_template_function.py` line 62 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8Sf9qv7Tk4bFHafj1j:-O8cP_nu63kA3scDvcux:b-e9yz9w) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/func/query_template_function.py#L62)):*

Maybe RuntimeError would be more appropriate here. This presumably should never be called.
</blockquote></details>

It can't unfortunately. It's part of the `__getattr__` contract that the only exception type raised is `AttributeError`. Breaking this contract will break tab-completion in Jupyter in weird ways.

We need to rethink this pattern though. Just implementing `__getattr__` for `QueryTemplateFunction` isn't enough.
___
*[`pixeltable/index/btree.py` line 51 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SfWvhDEOfzeekE9LW:-O8cPnrYEQ60zVFbje6l:b-7slmyd) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/index/btree.py#L51)):*

Does it cause problems if it's part of BtreeIndex? This doesn't have meaning outside that class.
</blockquote></details>

Yes it causes problems. Every function in a class _must_ have `self` (or `cls`) as its first parameter. So a UDF without a `self` parameter can't appear inside a class.

I don't see this as being an issue, since marking it as module-protected carries the same implication as putting it inside the class (and neither one has any sort of runtime enforcement).
___
*[`pixeltable/io/external_store.py` line 244 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8ShZzqEI7bxrylSYr7:-O8cQSd62aYNNxGzaMNL:b-ary9dh) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/io/external_store.py#L244)):*

Why assert and not cast()? (where should we use one vs the other?)
</blockquote></details>

I think `assert` is generally preferred, but I've occasionally used `cast` when it's more convenient, for example, when it's a compound expression that's being cast and I don't want to introduce extra code by assigning it to a variable and then asserting on that variable.
___
*[`pixeltable/metadata/schema.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8Si5sODTEjnStZrHCI:-O8cQjNYB2oIBR7XhnY4:bnb177s) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L13)):*

Why is this needed?
</blockquote></details>

`Base` has to be marked as a `type`, or mypy will throw an absolute fit when it gets used as a type hint. But a `type` doesn't have a `metadata` attribute, and Python doesn't let you do something like
```
Base: Intersection[type, DeclarativeMeta]
```
I did find a way to avoid the `# type: ignore` though, which is an improvement. Pls chime in if you see a better way to do this.
___
*[`pixeltable/metadata/schema.py` line 29 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SiReJAlEOict5n33N:-O8cSMHT21o-wstpFQUB:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L29)):*

precede w/ assert len(...) == 1 and remove if
</blockquote></details>

Done.
___
*[`pixeltable/metadata/schema.py` line 42 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SicB23pCqmyjV8Ckl:-O8cSgACCg_WYcF-yj0I:b-y7b5me) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L42)):*

Is this a logic error? Should this raise something?
</blockquote></details>

i don't think so? I think it can only result from some internal coding error (it would basically mean we used an unsupported type in defining our metadata structs, right? or am i missing something?)
___
*[`pixeltable/metadata/converters/util.py` line 71 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SittG6n1W1ZwNwR_l:-O8cT0DY9sVzhxlOKa18:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/converters/util.py#L71)):*

type decl?
</blockquote></details>

Done.
___
*[`pixeltable/utils/help.py` line  at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SjSmG0GziCFKbcr7N:-O8cTJnd36HvvpzEtQqA:b-h9emei) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/utils/help.py#L)):*

What about help()?
</blockquote></details>

We found a way to surface UDF docs via the builtin Python `help()`. That obviates the need for a separate `pxt.help()`, right?

pixeltable/catalog/table.py
```diff
@@ -1009,3 +1012,14 @@ def __dir__(self) -> list[str]:
 
     def _ipython_key_completions_(self) -> list[str]:
         return list(self._schema.keys()) + self._query_names
+
+    class QueryScope:
```
mkornacker: Since you already need to introduce this, why not have it record the QueryTemplateFunctions itself (so you only need one query-related member var in Table).

Also, please move this to the top, that makes reading the file top-to-bottom easier.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/319)* status: 0 of 46 files reviewed, 11 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/table.py` line 1016 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/319#gh-1797291137:-O8xRJRsA33xCqU3CI_g:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/86fbfe210a052413017d6415bbdcd47d9e5b8084/pixeltable/catalog/table.py#L1016)):*

Since you already need to introduce this, why not have it record the QueryTemplateFunctions itself (so you only need one query-related member var in Table).

Also, please move this to the top, that makes reading the file top-to-bottom easier.
</blockquote></details>

Done.

mkornacker (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/319)* status: 0 of 46 files reviewed, 10 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/metadata/schema.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8Si5sODTEjnStZrHCI:-O9HvpnsCNpuyijDdDwJ:b-uba6we) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L13)):*

`Base` has to be marked as a `type`, or mypy will throw an absolute fit when it gets used as a type hint. But a `type` doesn't have a `metadata` attribute, and Python doesn't let you do something like
```
Base: Intersection[type, DeclarativeMeta]
```
I did find a way to avoid the `# type: ignore` though, which is an improvement. Pls chime in if you see a better way to do this.
</blockquote></details>

Best to put that explanation into the code, it wasn't obvious to me...
___
*[`pixeltable/metadata/schema.py` line 29 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SiReJAlEOict5n33N:-O9HwAqs2P9qeWOl77kJ:b-896fig) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L29)):*

Done.
</blockquote></details>

Done?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/319)* status: 0 of 46 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/metadata/schema.py` line 13 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8Si5sODTEjnStZrHCI:-O9IjaxC8PYXqSoD9ng5:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L13)):*

Best to put that explanation into the code, it wasn't obvious to me...
</blockquote></details>

Done.
___
*[`pixeltable/metadata/schema.py` line 29 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/319#-O8SiReJAlEOict5n33N:-O9IjXXX6rY1D-Xb32ar:b-hefk6z) ([raw file](https://github.com/pixeltable/pixeltable/blob/0595b1bb6a977ce18ffbc89b1b181d791816b4cc/pixeltable/metadata/schema.py#L29)):*

Done?
</blockquote></details>

Yes - I switched to using `assert len` and no `if`. Did I miss something?

PR #320, opened 2024-10-06
Converting aggregate FunctionCalls to SQL

mkornacker:
Also:
- new class ExprDict
- fixes a bug in TableVersion.insert(df), which didn't generate correct rowids
- changes limit=0 to mean 'return no rows' instead of 'return all rows'


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/320)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/320)* status: 0 of 28 files reviewed, 25 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/plan.py` line 109 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cIzn_BCv0IdMvs5E-:-O8cIzn_BCv0IdMvs5E0:b-styobp) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L109)):*
> ```Python
>             exprs.Expr.list_subexprs(
>                 candidates, expr_class=exprs.FunctionCall,
>                 filter=lambda e: bool(e.is_agg_fn_call and not e.is_window_fn_call)))
> ```

The enclosing `bool` shouldn't be necessary here?
___
*[`pixeltable/plan.py` line 131 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cJPGJ5suE3yMfxF_z:-O8cJPGJ5suE3yMfxFa-:b-4ear9j) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L131)):*
> ```Python
>         # check that filter doesn't contain aggregates
>         if self.filter is not None:
>             if bool(
> ```

Is the type checker demanding these `bool`s?
___
*[`pixeltable/plan.py` line 132 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cJpXWDlyJN-CATyOq:-O8cJpXWDlyJN-CATyOr:b-51p101) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L132)):*
> ```Python
>         if self.filter is not None:
>             if bool(
>                 next(self.filter.subexprs(expr_class=exprs.FunctionCall, filter=lambda e: _is_agg_fn_call(e)), False)
> ```

This might be cleaner:
```
if any(_is_agg_fn_call(e) for e in self.filter.subexprs(expr_class=exprs.FunctionCall))
```
___
*[`pixeltable/plan.py` line 190 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8gv1OMCfEU99_UP_XA:-O8gv1OMCfEU99_UP_XB:b-69s23l) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L190)):*
> ```Python
>         row_builder.set_slot_idxs(self.agg_fn_calls)
>         row_builder.set_slot_idxs(self.agg_order_by)
> ```

extraneous newline
___
*[`pixeltable/plan.py` line 196 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8gvRf74kHNL_x5KYYb:-O8gvRf74kHNL_x5KYYc:bdtl026) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L196)):*
> ```Python
>         for fn_call in self.window_fn_calls:
>             # window functions require ordering by the group_by/order_by clauses
>             gb, ob = fn_call.get_window_sort_exprs()
> ```

i suggest using more descriptive names than `gb` and `ob` - maybe `gb_sort_key_exprs` etc.?
___
*[`pixeltable/plan.py` line 525 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nOX4r5c3aGGrO0eLv:-O8nOX4r5c3aGGrO0eLw:b-2mwtvv) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L525)):*
> ```Python
> 
>         if verify_agg:
>             ordering: Ordering = []
> ```

What's this doing? If it's a type checker thing, you shouldn't need the extraneous declaration (just annotate with `order: Ordering` the first time it's assigned)
___
*[`pixeltable/plan.py` line 533 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nP-qh7w-xyIbP_tPL:-O8nP-qh7w-xyIbP_tPM:befuji8) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L533)):*
> ```Python
>             for fn_call in analyzer.agg_fn_calls:
>                 # agg functions with an ordering requirement are implicitly ascending
>                 ordering = [(e, None) for e in analyzer.group_by_clause] + [(e, True) for e in fn_call.get_agg_order_by()]
> ```

How would this work if I wanted to aggregate along a different ordering? e.g. I have an aggregator `my_agg` with an ordering requirement, and I do something like

```t.order_by(t.my_col, asc=False).select(my_agg(t.my_col))```

The code would seem to imply I'd get an "Incompatible ordering requirements", but it feels like this should be a valid pattern.

Rather than making aggregators with ordering requirements implicitly ascending, might we instead consider requiring an explicit order_by in order to use them? i.e., make it an error to use an ordered aggregator without a preceding order_by?
___
*[`pixeltable/plan.py` line 549 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8gwoWG3yfYiTvjsi6f:-O8gwoWG3yfYiTvjsi6g:b-xx2fgq) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L549)):*
> ```Python
>     def _determine_ordering(cls, analyzer: Analyzer) -> list[tuple[exprs.Expr, bool]]:
>         """Returns the exprs for the ORDER BY clause of the SqlScanNode"""
>         order_by_items: list[tuple[exprs.Expr, Optional[bool]]] = []
> ```

Is this not just a `list[Ordering]`?
___
*[`pixeltable/plan.py` line 710 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8gwy_54zMWhzCmXl-e:-O8gwy_54zMWhzCmXl-f:b-4qe8do) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L710)):*
> ```Python
>         sql_elements = analyzer.sql_elements
>         is_python_agg = (
>                 not sql_elements.contains(analyzer.agg_fn_calls) or not sql_elements.contains(analyzer.window_fn_calls)
> ```

extraneous indentation
___
*[`pixeltable/plan.py` line 735 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nSuOSDoKjWC5KQnz2:-O8nSuOSDoKjWC5KQnz3:bhcu16a) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L735)):*
> ```Python
>         sql_scan_exprs = exprs.ExprSet(candidates)
>         # we don't want to materialize Literals via a Select
>         sql_scan_exprs -= exprs.ExprSet(e for e in sql_scan_exprs if isinstance(e, exprs.Literal))
> ```

Can this be folded into a single expression with the above?

`sql_scan_exprs = exprs.ExprSet(e for e in candidates if not isinstance(e, exprs.Literal))`
___
*[`pixeltable/plan.py` line 773 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nV-Gi9L-0okf--he_:-O8nV-Gi9L-0okf--hea:btyshx8) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L773)):*
> ```Python
>                 agg_output = exprs.ExprSet(
>                     # explicit cast to avoid typechecking error
>                     cast(list[exprs.Expr], analyzer.grouping_exprs + analyzer.agg_fn_calls + analyzer.window_fn_calls))
> ```

If you want to avoid the typecast, I think this works:

```python
exprs.ExprSet(
    itertools.chain(analyzer.grouping_exprs, analyzer.agg_fn_calls, analyzer.window_fn_calls)
)
```
The issue (I think) isn't that it needs a cast, but that `list` is not covariant in Python. The cast just causes mypy to skip past the expression that has the type error.

The following also seems to work, though in this case I'm not sure why it works and yours doesn't:
```
agg_list = analyzer.grouping_exprs + analyzer.agg_fn_calls + analyzer.window_fn_calls
agg_output = exprs.ExprSet(agg_list)
```
___
*[`pixeltable/store.py` line 136 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cKZoR7AXjiOdXXfge:-O8cKZoR7AXjiOdXXfgf:b-vs3crb) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/store.py#L136)):*
> ```Python
> 
>     def _move_tmp_media_files(
>             self, table_rows: list[Dict[str, Any]], media_cols: list[catalog.Column], v_min: int
> ```

Lowercase the `Dict` also (here and elsewhere - since you're already lowercasing `list` and `set` and `tuple` in this file)
___
*[`pixeltable/catalog/table_version.py` line 694 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nX0MKCYCy_0sQU8mZ:-O8nX0MKCYCy_0sQU8m_:b2hvquo) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/catalog/table_version.py#L694)):*
> ```Python
> 
>         # this is a base table; we generate rowids during the insert
>         def rowids() -> Iterator[int]:
> ```

Pretty neat what you can do with Python `yield`
___
*[`pixeltable/exec/aggregation_node.py` line 82 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nX_HoCjJ4vyx-4Mvv:-O8nX_HoCjJ4vyx-4Mvw:bqnl5jg) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/aggregation_node.py#L82)):*
> ```Python
>         self.output_batch.flush_imgs(None, self.stored_img_cols, self.flushed_img_slots)
>         _logger.debug(f'AggregateNode: consumed {num_input_rows} rows, returning {len(self.output_batch.rows)} rows')
>         yield self.output_batch
> ```

I'm confused, `yield` will only ever be called once? Does that imply every aggregator will only ever emit a single output batch containing the entire query results? I think this needs an explanatory comment
___
*[`pixeltable/exec/exec_node.py` line 51 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nZPkIEBRi4I80xfyA:-O8nZPkIEBRi4I80xfyB:bsx8dpn) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/exec_node.py#L51)):*
> ```Python
>             self.input.set_stored_img_cols(stored_img_cols)
> 
>     #@abc.abstractmethod
> ```

What are these comments?
___
*[`pixeltable/exec/exec_node.py` line 56 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nZ67w414nM5yrYW_a:-O8nZ67w414nM5yrYW_b:b-v6hy1p) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/exec_node.py#L56)):*
> ```Python
> 
>     #@abc.abstractmethod
>     def __next__(self) -> DataRowBatch:
> ```

Is `next` actually required if we're moving to an `iter` pattern? To me it's a more sensible abstraction for `ExecNode` to look like an `Iterable[DataRowBatch]` rather than an `Iterator[DataRowBatch]`.
___
*[`pixeltable/exec/sql_node.py` line 17 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cKs3vAJrcqcOQADYq:-O8cKs3vAJrcqcOQADYr:b-mpwn74) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L17)):*
> ```Python
> 
> 
> Ordering = list[tuple[exprs.Expr, Optional[bool]]]
> ```

Needs a comment explaining what this is. Probably best to use a NamedTuple here to clarify the meanings of the Expr and Optional[bool].

I find the name `Ordering` somewhat confusing; many languages use `Ordering` as a synonym for `Comparator`, which isn't what this is. Something like `OrderByKey` would be clearer.
___
*[`pixeltable/exec/sql_node.py` line 21 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8n_73NBGAwrJGDti_a:-O8n_73NBGAwrJGDti_b:bmfnetw) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L21)):*
> ```Python
> 
> def combine_orderings(orderings: Iterable[Ordering]) -> Optional[Ordering]:
>     """Returns an ordering that's compatible with orderings, or None if no such ordering exists"""
> ```

This needs elaboration. It isn't clear what "compatible with" means. By reading the code, I infer it means an ordering C such that for any X in orderings: the exprs of X are an initial segment of those of C, and if some expr X_i specifies a direction, then C_i must also specify the same direction. Then `combine_orderings` returns the least specific such C.

Assuming that's correct, it'd be useful to say it in the comment, which would make the code easier to read.
___
*[`pixeltable/exec/sql_node.py` line 132 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8naw5w7CU7Ccq7Dpb7:-O8naw5w7CU7Ccq7Dpb8:b-16iv6c) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L132)):*
> ```Python
> 
>     def _ordering_tbl_ids(self) -> set[UUID]:
>         return exprs.Expr.list_tbl_ids([e for e, _ in self.ordering])
> ```

extraneous list creation
___
*[`pixeltable/exec/sql_node.py` line 261 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nbgoR1EyumSsBWMcL:-O8nbgoR1EyumSsBWMcM:b-y1oszi) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L261)):*
> ```Python
>             for i, e in enumerate(self.select_list):
>                 slot_idx = e.slot_idx
>                 # certain numerical operations can produce Decimals (eg, SUM(<int column>)); we need to convert them
> ```

i think it's better to do this in SQL with a `sql.expression.cast` (in the implementation of `sum.to_sql` and so on). See `arithmetic_expr.py` lines 72, 86, 88, for example.
___
*[`pixeltable/exprs/expr_dict.py` line 11 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cLGGa8DbbUNdfmOw_:-O8cLGGa8DbbUNdfmOwa:b-mw0av4) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exprs/expr_dict.py#L11)):*
> ```Python
>     A dictionary that maps Expr instances to values of type T.
> 
>     We cannot use dict[Expr, T] because Expr.__eq__() serves a different purpose than the default __eq__.
> ```

A custom collection class seems like a heavyweight solution to this problem.

Would it be simpler to use a
```
dict[ExprWrapper, T]
```
where an `ExprWrapper` contains an `Expr` and implements its own `__eq__` and `__hash__` appropriately?
___
*[`pixeltable/functions/globals.py` line 31 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ncucw0bpusywg9hvW:-O8ncucw0bpusywg9hvX:bbeuw8s) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/functions/globals.py#L31)):*
> ```Python
> 
> 
> # disable type checking: mypy doesn't seem to understand that 'sum' is an instance of Function
> ```

This is ok for now but please create a ticket in JIRA to fix it eventually. (Worst case is through some custom logic in our mypy extension)
___
*[`pixeltable/functions/globals.py` line 34 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ncjHXADaG89JSE_mf:-O8ncjHXADaG89JSE_mg:bhfeb85) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/functions/globals.py#L34)):*
> ```Python
> @sum.to_sql  # type: ignore
> def _(val: sql.ColumnElement) -> Optional[sql.ColumnElement]:
>     return sql.sql.func.sum(val)
> ```

Here is where I suggest using a SQL cast to avoid the creation of `Decimal` objects:
```python
sql.sql.expression.cast(sql.sql.func.sum(val), sql.Integer)  # or sql.Float or whatever
```
___
*[`tests/test_exprs.py` line 154 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ndcj8DpT0Vq4mI5BM:-O8ndcj8DpT0Vq4mI5BN:b-5ii15d) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/tests/test_exprs.py#L154)):*
> ```Python
>         _ = t.where(t.c2 > 50).show()
>         print(_)
>         _ = t.where(t.c1n == None).show()
> ```

missed one? (s/show/collect/)
___
*[`tests/functions/test_timestamp.py` line 175 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ne0DtEmy4aYFuJDtS:-O8ne0DtEmy4aYFuJDtT:b-usl3a8) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/tests/functions/test_timestamp.py#L175)):*
> ```Python
>             res['out'] == [dt.replace(minute=0).astimezone(Env.get().default_time_zone) for dt in test_dts]
>         )
>         _ = (t.group_by(t.dt.month)
> ```

Let's make sure every SQL-ized aggregator is tested both in a SQL-izable call and in a Python-only call, to ensure the logic stays in sync between SQL/Python. Similar test cases are needed for `sum`, etc., as well.

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/320)* status: 0 of 29 files reviewed, 25 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/plan.py` line 109 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cIzn_BCv0IdMvs5E-:-O8rGcXU2QNCVUEXwZQw:b-dp1el7) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L109)):*

The enclosing `bool` shouldn't be necessary here?
</blockquote></details>

I couldn't get typechecking to pass without it, but happy to do something else here.
___
*[`pixeltable/plan.py` line 132 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cJpXWDlyJN-CATyOq:-O8woqKN-u-cO8WYx8Nl:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L132)):*

This might be cleaner:
```
if any(_is_agg_fn_call(e) for e in self.filter.subexprs(expr_class=exprs.FunctionCall))
```
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 190 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8gv1OMCfEU99_UP_XA:-O8wozyn5EskVyOmjL5b:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L190)):*

extraneous newline
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 196 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8gvRf74kHNL_x5KYYb:-O8wp86R7vJB5UxeKPcP:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L196)):*

i suggest using more descriptive names than `gb` and `ob` - maybe `gb_sort_key_exprs` etc.?
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 525 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nOX4r5c3aGGrO0eLv:-O8wpdd804jBNCD8TKvS:b-q4rpqw) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L525)):*

What's this doing? If it's a type checker thing, you shouldn't need the extraneous declaration (just annotate with `order: Ordering` the first time it's assigned)
</blockquote></details>

I'm finding this more informative.
___
*[`pixeltable/plan.py` line 533 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nP-qh7w-xyIbP_tPL:-O8wpP6qAnsQ1usn4QAc:bfrf84r) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L533)):*

How would this work if I wanted to aggregate along a different ordering? e.g. I have an aggregator `my_agg` with an ordering requirement, and I do something like

```t.order_by(t.my_col, asc=False).select(my_agg(t.my_col))```

The code would seem to imply I'd get an "Incompatible ordering requirements", but it feels like this should be a valid pattern.

Rather than making aggregators with ordering requirements implicitly ascending, might we instead consider requiring an explicit order_by in order to use them? i.e., make it an error to use an ordered aggregator without a preceding order_by?
</blockquote></details>

You are right that this should be a valid pattern, but it's not supported at the moment. Implementing that requires more drastic changes (you need to generate entirely different plans that include reordering operators).
___
*[`pixeltable/plan.py` line 549 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8gwoWG3yfYiTvjsi6f:-O8wpsH5-KwpU89WJpZz:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L549)):*

Is this not just a `list[Ordering]`?
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 710 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8gwy_54zMWhzCmXl-e:-O8wpz8E9BxCoIazNMws:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L710)):*

extraneous indentation
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 735 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nSuOSDoKjWC5KQnz2:-O8wqHeVCLYKHFwDAVLy:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L735)):*

Can this be folded into a single expression with the above?

`sql_scan_exprs = exprs.ExprSet(e for e in candidates if not isinstance(e, exprs.Literal))`
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 773 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nV-Gi9L-0okf--he_:-O8wr72n9FK3JBKUqRIC:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L773)):*

If you want to avoid the typecast, I think this works:

```python
exprs.ExprSet(
    itertools.chain(analyzer.grouping_exprs, analyzer.agg_fn_calls, analyzer.window_fn_calls)
)
```
The issue (I think) isn't that it needs a cast, but that `list` is not covariant in Python. The cast just causes mypy to skip past the expression that has the type error.

The following also seems to work, though in this case I'm not sure why it works and yours doesn't:
```
agg_list = analyzer.grouping_exprs + analyzer.agg_fn_calls + analyzer.window_fn_calls
agg_output = exprs.ExprSet(agg_list)
```
</blockquote></details>

Done.
___
*[`pixeltable/store.py` line 136 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cKZoR7AXjiOdXXfge:-O8wrEJJ5fCtmlmyzz_t:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/store.py#L136)):*

Lowercase the `Dict` also (here and elsewhere - since you're already lowercasing `list` and `set` and `tuple` in this file)
</blockquote></details>

Done.
___
*[`pixeltable/exec/aggregation_node.py` line 82 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nX_HoCjJ4vyx-4Mvv:-O8wso4g9gXqS9dXK3hl:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/aggregation_node.py#L82)):*

I'm confused, `yield` will only ever be called once? Does that imply every aggregator will only ever emit a single output batch containing the entire query results? I think this needs an explanatory comment
</blockquote></details>

Done.
___
*[`pixeltable/exec/exec_node.py` line 51 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nZPkIEBRi4I80xfyA:-O8wtA5gBRFuWj_DcsrI:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/exec_node.py#L51)):*

What are these comments?
</blockquote></details>

Done.
___
*[`pixeltable/exec/exec_node.py` line 56 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nZ67w414nM5yrYW_a:-O8wtAxp6juuB1UJkEJd:b-2s921n) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/exec_node.py#L56)):*

Is `next` actually required if we're moving to an `iter` pattern? To me it's a more sensible abstraction for `ExecNode` to look like an `Iterable[DataRowBatch]` rather than an `Iterator[DataRowBatch]`.
</blockquote></details>

It's not. Added an explanatory comment. I don't want to add the wholesale conversion to __iter__() to these changes, it's already a fair amount of code.
___
*[`pixeltable/exec/sql_node.py` line 17 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cKs3vAJrcqcOQADYq:-O8wy-BM4gR91kIBCdRV:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L17)):*

Needs a comment explaining what this is. Probably best to use a NamedTuple here to clarify the meanings of the Expr and Optional[bool].

I find the name `Ordering` somewhat confusing; many languages use `Ordering` as a synonym for `Comparator`, which isn't what this is. Something like `OrderByKey` would be clearer.
</blockquote></details>

Done.
___
*[`pixeltable/exec/sql_node.py` line 21 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8n_73NBGAwrJGDti_a:-O8wy-yb8IJBRtUOXf1Z:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L21)):*

This needs elaboration. It isn't clear what "compatible with" means. By reading the code, I infer it means an ordering C such that for any X in orderings: the exprs of X are an initial segment of those of C, and if some expr X_i specifies a direction, then C_i must also specify the same direction. Then `combine_orderings` returns the least specific such C.

Assuming that's correct, it'd be useful to say it in the comment, which would make the code easier to read.
</blockquote></details>

Done.
___
*[`pixeltable/exec/sql_node.py` line 132 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8naw5w7CU7Ccq7Dpb7:-O8wzfzq52o9n7bmHGtD:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L132)):*

extraneous list creation
</blockquote></details>

Done.
___
*[`pixeltable/exec/sql_node.py` line 261 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nbgoR1EyumSsBWMcL:-O8wzjlr45HdcGb6UFsN:b8mldwh) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L261)):*

i think it's better to do this in SQL with a `sql.expression.cast` (in the implementation of `sum.to_sql` and so on). See `arithmetic_expr.py` lines 72, 86, 88, for example.
</blockquote></details>

Added explanatory comment
___
*[`pixeltable/exprs/expr_dict.py` line 11 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cLGGa8DbbUNdfmOw_:-O8x-JhUAusrQDYIK2b3:b-28qncq) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exprs/expr_dict.py#L11)):*

A custom collection class seems like a heavyweight solution to this problem.

Would it be simpler to use a
```
dict[ExprWrapper, T]
```
where an `ExprWrapper` contains an `Expr` and implements its own `__eq__` and `__hash__` appropriately?
</blockquote></details>

I don't see this as heavyweight. It's only 55 lines, the code is easy to follow and self-contained, and this frees the caller from having to create ExprWrappers.
___
*[`pixeltable/functions/globals.py` line 31 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ncucw0bpusywg9hvW:-O8x-jxrAosBcPVnJTY_:b48zv44) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/functions/globals.py#L31)):*

This is ok for now but please create a ticket in JIRA to fix it eventually. (Worst case is through some custom logic in our mypy extension)
</blockquote></details>

Let's talk about that offline. It's not clear to me why that's particularly useful here.
___
*[`pixeltable/functions/globals.py` line 34 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ncjHXADaG89JSE_mf:-O8x-i9F6ul0snR_5ovG:bsvoyq3) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/functions/globals.py#L34)):*

Here is where I suggest using a SQL cast to avoid the creation of `Decimal` objects:
```python
sql.sql.expression.cast(sql.sql.func.sum(val), sql.Integer)  # or sql.Float or whatever
```
</blockquote></details>

See comments above.
___
*[`tests/test_exprs.py` line 154 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ndcj8DpT0Vq4mI5BM:-O8x-v1e1Y8RQfKNRViD:b-si56o6) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/tests/test_exprs.py#L154)):*

missed one? (s/show/collect/)
</blockquote></details>

I only replaced show(0) everywhere, because limit=0 now means 'return nothing' (this is useful for validating that a query is actually runnable).
___
*[`tests/functions/test_timestamp.py` line 175 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ne0DtEmy4aYFuJDtS:-O8x0MJJ-2jPsI7oxCZM:b6rdv5y) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/tests/functions/test_timestamp.py#L175)):*

Let's make sure every SQL-ized aggregator is tested both in a SQL-izable call and in a Python-only call, to ensure the logic stays in sync between SQL/Python. Similar test cases are needed for `sum`, etc., as well.
</blockquote></details>

That's already done, see test_agg().

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/320)* status: 0 of 29 files reviewed, 24 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/plan.py` line 525 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nOX4r5c3aGGrO0eLv:-O9HZ-brAW8NaKYXfG5N:b-et971k) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L525)):*

I'm finding this more informative.
</blockquote></details>

I checked the style guides and it's actually ok to write:
```python
ordering: OrderByClause
```
if you know that `ordering` is going to be defined prior to use in all succeeding code paths. I think that's the way to go, assuming the type checker is ok with it (the assignment to `[]` that never gets used is what was bothering me)
___
*[`pixeltable/plan.py` line 533 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nP-qh7w-xyIbP_tPL:-O9HZwcK8KS5a44QhUe0:b-b6gv8s) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L533)):*

You are right that this should be a valid pattern, but it's not supported at the moment. Implementing that requires more drastic changes (you need to generate entirely different plans that include reordering operators).
</blockquote></details>

Add a JIRA ticket so we don't lose track of this desirable enhancement?
___
*[`pixeltable/exec/sql_node.py` line 21 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8n_73NBGAwrJGDti_a:-O9H_laE-nG50V4r2qq6:b-z8tmjy) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L21)):*

Done.
</blockquote></details>

Wait, I don't think that comment is correct? The exprs don't need to be identical, one just needs to be an initial segment of the other, right? And if they're unequal, the output clause (the one that's "compatible with" the others) needs to be the longer one. i.e., "compatible with" is not a symmetric relationship, but the comment implies it is.
___
*[`pixeltable/exec/sql_node.py` line 261 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nbgoR1EyumSsBWMcL:-O9HaGhP6-ENd7GGqZI8:b79bttz) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L261)):*

Added explanatory comment
</blockquote></details>

Per our conversation last week, I now understand that it's unsafe to do the conversion in SQL due to possible loss of precision. We should say that in the comment. I'd also argue for adding a test that specifically would overflow the SQL decimal-to-int conversion, just to guard against someone trying to "optimize" this in the future
___
*[`pixeltable/exprs/expr_dict.py` line 11 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8cLGGa8DbbUNdfmOw_:-O9HajXp3747_YkBKVQg:b-p9qeex) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exprs/expr_dict.py#L11)):*

I don't see this as heavyweight. It's only 55 lines, the code is easy to follow and self-contained, and this frees the caller from having to create ExprWrappers.
</blockquote></details>

well, over 100 lines if you include the custom `ExprSet`, but i don't feel that strongly about it.
___
*[`pixeltable/exprs/expr_set.py` line 70 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O9Hb7PV1EB_EVRMSLw8:-O9Hb7PV1EB_EVRMSLw9:bdl5lhs) ([raw file](https://github.com/pixeltable/pixeltable/blob/506aa22d615a301a5e307f78d8ceee349e2dc9ef/pixeltable/exprs/expr_set.py#L70)):*
> ```Python
>         return self.difference(other)
> 
> ExprSet = ExprSet[Expr]  # type: ignore
> ```

I did a double-take when I saw this. Why alter the meaning of `ExprSet`? Seems super confusing (is there any precedent for doing this?)
___
*[`pixeltable/functions/globals.py` line 31 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ncucw0bpusywg9hvW:-O9HbPs25ZZIKKicEH6g:bcm37xb) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/functions/globals.py#L31)):*

Let's talk about that offline. It's not clear to me why that's particularly useful here.
</blockquote></details>

`# type: ignore` is generally undesirable, and it's an avoidable inconvenience for users defining aggregators. I don't think it'll be hard to fix.

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/320)* status: 0 of 29 files reviewed, 24 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/plan.py` line 525 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nOX4r5c3aGGrO0eLv:-O9IDuZx2aEs2F4idhV2:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L525)):*

I checked the style guides and it's actually ok to write:
```python
ordering: OrderByClause
```
if you know that `ordering` is going to be defined prior to use in all succeeding code paths. I think that's the way to go, assuming the type checker is ok with it (the assignment to `[]` that never gets used is what was bothering me)
</blockquote></details>

Done.
___
*[`pixeltable/plan.py` line 533 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nP-qh7w-xyIbP_tPL:-O9IEenJ29QsFnLmpWqx:b-uarmjo) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/plan.py#L533)):*

Add a JIRA ticket so we don't lose track of this desirable enhancement?
</blockquote></details>

Done (309)
___
*[`pixeltable/exec/sql_node.py` line 21 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8n_73NBGAwrJGDti_a:-O9IF8EQ0ZP0ymgdTH9b:b6z7fx6) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L21)):*

Wait, I don't think that comment is correct? The exprs don't need to be identical, one just needs to be an initial segment of the other, right? And if they're unequal, the output clause (the one that's "compatible with" the others) needs to be the longer one. i.e., "compatible with" is not a symmetric relationship, but the comment implies it is.
</blockquote></details>

Each item has an expr, and those exprs need to be identical.
___
*[`pixeltable/exec/sql_node.py` line 261 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8nbgoR1EyumSsBWMcL:-O9IGBQ0CYq4mFqT1hbk:bsnmhwl) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L261)):*

Per our conversation last week, I now understand that it's unsafe to do the conversion in SQL due to possible loss of precision. We should say that in the comment. I'd also argue for adding a test that specifically would overflow the SQL decimal-to-int conversion, just to guard against someone trying to "optimize" this in the future
</blockquote></details>

I'm not sure how to induce that, and I would say that's follow-on work.

The comment for sum.to_sql() explains why it doesn't do a conversion.
___
*[`pixeltable/exprs/expr_set.py` line 70 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O9Hb7PV1EB_EVRMSLw8:-O9IIEedBNRF5mi5Vgvg:b-w0c3o4) ([raw file](https://github.com/pixeltable/pixeltable/blob/506aa22d615a301a5e307f78d8ceee349e2dc9ef/pixeltable/exprs/expr_set.py#L70)):*

I did a double-take when I saw this. Why alter the meaning of `ExprSet`? Seems super confusing (is there any precedent for doing this?)
</blockquote></details>

Doesn't appear to be necessary after all
___
*[`pixeltable/functions/globals.py` line 31 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8ncucw0bpusywg9hvW:-O9IILPt3-lWFo7ZDfOT:bn3cmuj) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/functions/globals.py#L31)):*

`# type: ignore` is generally undesirable, and it's an avoidable inconvenience for users defining aggregators. I don't think it'll be hard to fix.
</blockquote></details>

I'll leave a todo

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/320)* status: 0 of 29 files reviewed, 24 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exec/sql_node.py` line 21 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8n_73NBGAwrJGDti_a:-O9Ik1rW9nrrlrjOM7OA:b-ytarw3) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L21)):*

Each item has an expr, and those exprs need to be identical.
</blockquote></details>

Right, but the clauses don't need to be the same length, do they? The `result` can be strictly longer than some of the elements of `clauses`. And none of the clauses can be strictly longer than the `result`. So it's not a symmetric relationship. At least that's how I'm reading the code.

mkornacker (commented):
> Dismissed @aaron-siegel from a discussion.
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/320)* status: 0 of 29 files reviewed, 23 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exec/sql_node.py` line 21 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8n_73NBGAwrJGDti_a:-O9KrHVI4-Q7uyQi26Ds:bckkgvy) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L21)):*

Right, but the clauses don't need to be the same length, do they? The `result` can be strictly longer than some of the elements of `clauses`. And none of the clauses can be strictly longer than the `result`. So it's not a symmetric relationship. At least that's how I'm reading the code.
</blockquote></details>

The relationship is symmetric. If ordering A is compatible with B, then B is also compatible with A.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/320)* status: 0 of 29 files reviewed, 22 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exec/sql_node.py` line 21 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/320#-O8n_73NBGAwrJGDti_a:-O9KybPSCNnJMXBlpuzy:bkw0d3f) ([raw file](https://github.com/pixeltable/pixeltable/blob/1a888cc4a34a5be51ef15d48232403bab82eeada/pixeltable/exec/sql_node.py#L21)):*

The relationship is symmetric. If ordering A is compatible with B, then B is also compatible with A.
</blockquote></details>

Ok yes, it still makes sense with that definition. Still, the lengths of the sequences may differ. Things would be clearer if that were called out in the comment. As currently written it's easy to interpret it as meaning the expr sequences must be the same length (which then becomes a point of confusion when reading the code).

aaron-siegel (approved):

PR #325, opened 2024-10-09
Allow @query to be redefined if in __main__ (notebook) scope

aaron-siegel:
It's a 1-line change. I don't see an easy way to test it, unfortunately; we'd have to simulate notebook scope somehow (maybe we can file that away as a future problem)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/325)

mkornacker (approved):

PR #327, opened 2024-10-09
Raise an exception if an error occurs during add_column()

aaron-siegel:
It's been a source of user frustration that exceptions are swallowed into the `errormsg` and `errortype` fields when raised during `add_column()` evaluation. This PR changes that behavior to raise an `excs.Error` instead.

Also adds a parameter `on_error: Literal['raise', 'continue']` parameter to `add_column()`. The default is `raise`, whereas `continue` implies the older behavior.

This is an inconsistent name with the existing `fail_on_exception` parameter of `insert`; we should decide which pattern is best.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/327)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/327)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/store.py` line 221 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/327#-O8nUGEj-whyT4-ED2Ie:-O8nUGEj-whyT4-ED2If:b1urkpk) ([raw file](https://github.com/pixeltable/pixeltable/blob/7470870c67d3abe6173fb43ab591825b67a3c8c8/pixeltable/store.py#L221)):*
> ```Python
>         value_expr_slot_idx: int,
>         conn: sql.engine.Connection,
>         on_error: Literal['raise', 'continue']
> ```

Let's do an ignore_errors bool flag here (and leave string flags to the user-facing part of the API).
___
*[`pixeltable/catalog/table.py` line 329 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/327#-O8nV8W64v88Oc1mREem:-O8nV8W64v88Oc1mREen:b-tnj4qf) ([raw file](https://github.com/pixeltable/pixeltable/blob/7470870c67d3abe6173fb43ab591825b67a3c8c8/pixeltable/catalog/table.py#L329)):*
> ```Python
>             stored: Optional[bool] = None,
>             print_stats: bool = False,
>             on_error: Literal['raise', 'continue'] = 'raise',
> ```

How about abort and ignore instead?

'raise' sounds like an implementation detail (and would be meaningless if we ever do SDKs for languages w/o exceptions).
___
*[`pixeltable/catalog/table_version.py` line 483 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/327#-O8nVyPL5mn-zeXuEDYZ:-O8nVyPL5mn-zeXuEDY_:bmyzi2o) ([raw file](https://github.com/pixeltable/pixeltable/blob/7470870c67d3abe6173fb43ab591825b67a3c8c8/pixeltable/catalog/table_version.py#L483)):*
> ```Python
>             _logger.info(f'Dropped index {idx_md.name} on table {self.name}')
> 
>     def add_column(self, col: Column, print_stats: bool, on_error: Literal['raise', 'continue']) -> UpdateStatus:
> ```

Let's do a bool flag here as well

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/327)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/store.py` line 221 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/327#-O8nUGEj-whyT4-ED2Ie:-O9JKF_9C6fpfUtQa0fA:bo8vgey) ([raw file](https://github.com/pixeltable/pixeltable/blob/7470870c67d3abe6173fb43ab591825b67a3c8c8/pixeltable/store.py#L221)):*

Let's do an ignore_errors bool flag here (and leave string flags to the user-facing part of the API).
</blockquote></details>

Per our discussion last week, leaving this as a string for now, though I suppose we could consider an enum instead.
___
*[`pixeltable/catalog/table.py` line 329 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/327#-O8nV8W64v88Oc1mREem:-O9JKBDwE4cQyMrkzQWE:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/7470870c67d3abe6173fb43ab591825b67a3c8c8/pixeltable/catalog/table.py#L329)):*

How about abort and ignore instead?

'raise' sounds like an implementation detail (and would be meaningless if we ever do SDKs for languages w/o exceptions).
</blockquote></details>

Done.

mkornacker (approved):

PR #328, opened 2024-10-12
Media validation on read

mkornacker:
This only contains the metadata changes, not the implementation of the on_read validation strategy.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/328)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/328)* status: 0 of 17 files reviewed, 3 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/globals.py` line 39 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9FkcFwE2V-7c5r5_pG:-O9FkcFwE2V-7c5r5_pH:b9xy8iw) ([raw file](https://github.com/pixeltable/pixeltable/blob/1f9f7bdf9e942036e881f04b71370c59a933e989/pixeltable/catalog/globals.py#L39)):*
> ```Python
> 
> 
> class MediaValidation(enum.Enum):
> ```

I'm surprised to see an enum being introduced here. Why not just keep it as a string? If you use `Literal[...]`, it's typesafe.

We've been inconsistent in the codebase about using strings vs. enums, I think this is something we should discuss. I'm not sure what we're actually gaining by using a custom enum vs. `Literal['on_read', 'on_write']`. We're going to end up with a lot of enums... if we want a separate `MediaValidation` type then something like

```python
MediaValidation = Literal['on_read', 'on_write']
```
makes more sense to me. Then you can use
```python
typing.get_args(MediaValidation)
```
to get the list of allowed values.
___
*[`pixeltable/metadata/converters/convert_21.py` line 15 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9FkBiz5I3sQwiU5een:-O9FkBj-3Wb0748AQbbi:bmfoye4) ([raw file](https://github.com/pixeltable/pixeltable/blob/1f9f7bdf9e942036e881f04b71370c59a933e989/pixeltable/metadata/converters/convert_21.py#L15)):*
> ```Python
> 
> def __update_table_schema_version_md(md: dict) -> None:
>     md['media_validation'] = 1  # catalog.MediaValidation.ON_WRITE
> ```

i'd much prefer to see strings as values here:
```python
md['media_validation'] = 'on_write'
```
It makes the metadata more human-readable, which is a good thing in general, e.g., for debugging. and it's the convention that's been used until now even when enums are used (e.g. index metric is stored as `metric.name.lower()`). what's the argument against doing this?
___
*[`pixeltable/metadata/converters/convert_21.py` line 19 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9FkZIV9fvnpg40ThDp:-O9FkZIV9fvnpg40ThDq:bs8x76r) ([raw file](https://github.com/pixeltable/pixeltable/blob/1f9f7bdf9e942036e881f04b71370c59a933e989/pixeltable/metadata/converters/convert_21.py#L19)):*
> ```Python
> 
> def __update_schema_column(md: dict) -> None:
>     md['media_validation'] = None
> ```

add newline at end of file

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/328)* status: 0 of 28 files reviewed, 28 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/globals.py` line 95 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uG45X3l6nDBupHsVI:-O9uG45X3l6nDBupHsVJ:bu6yzw1) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/globals.py#L95)):*
> ```Python
>             raise excs.Error('primary_key must be a single column name or a list of column names')
> 
>     if media_validation.upper() not in catalog.MediaValidation.__members__.keys():
> ```

`catalog.MediaValidation._member_names_` is shorter
___
*[`pixeltable/globals.py` line 181 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uGPm-2GmXtlKWw3v0:-O9uGPm-2GmXtlKWw3v1:b-df8s58) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/globals.py#L181)):*
> ```Python
>         iterator_class, iterator_args = iterator
> 
>     if media_validation.upper() not in catalog.MediaValidation.__members__.keys():
> ```

duplicated logic here; can this be factored out into a method of `MediaValidation`? (say, `MediaValidation.validated_lookup(val)` or some such, that does the check and then returns `cls[val]`)
___
*[`pixeltable/type_system.py` line 13 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uHgyy9cVfqqNThNbw:-O9uHgyy9cVfqqNThNbx:bx6yr8d) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/type_system.py#L13)):*
> ```Python
> from typing import Any, Iterable, Mapping, Optional, Sequence, Union
> 
> import PIL.Image
> ```

have you explored debugger options for using vscode yet? 

standardizing on vscode would reduce thrash in things like import sort order. vscode handles this in a very nice way (it uses isort which is configurable in pyproject.toml so you get a consistent set of import conventions applied at the project level)
___
*[`pixeltable/catalog/column.py` line 30 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uIEn4DpokUJf3I7_Y:-O9uIEn4DpokUJf3I7_Z:b-feilql) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/column.py#L30)):*
> ```Python
>     stored: bool
>     is_pk: bool
>     _media_validation: Optional[MediaValidation]  # if not set, TableVersion.media_validation applies
> ```

What convention are you using for determining when a member variable gets an underscore? I'd thought we weren't using them for non-public-facing classes.

Might be good to revisit/codify our code style conventions.
___
*[`pixeltable/catalog/table.py` line 250 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uJcxC7fIqTh9uAgK-:-O9uJcxC7fIqTh9uAgK0:b-2h9a2p) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table.py#L250)):*
> ```Python
> 
>     @property
>     def _media_validation(self) -> str:
> ```

it'd be more internally consistent to have this `@property` return an enum, and do the string conversion in `get_metadata()`.
___
*[`pixeltable/catalog/table.py` line 459 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uJo1m6F9XAWt086mr:-O9uJo1m6F9XAWt086ms:b-1heq9w) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table.py#L459)):*
> ```Python
> 
>         if 'media_validation' in spec:
>             if spec['media_validation'].upper() not in catalog.MediaValidation.__members__.keys():
> ```

another cargo-cult of this logic- it should definitely be factored out. the only difference is in the `Column {name}:` prefix in the error message, which could be a parameter of the validate method. (And maybe add `Table {name}` to the other instances for consistency)
___
*[`pixeltable/catalog/table_version.py` line 113 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uKMyuDVDH7giCP1Ka:-O9uKMyuDVDH7giCP1Kb:b-kps2mh) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table_version.py#L113)):*
> ```Python
>         is_view = tbl_md.view_md is not None
>         self.is_snapshot = (is_view and tbl_md.view_md.is_snapshot) or bool(is_snapshot)
>         self.media_validation = MediaValidation.__members__[schema_version_md.media_validation.upper()]
> ```

`.__members__` can be removed here
___
*[`pixeltable/catalog/table_version.py` line 299 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uKZZL9rj1ODizKbIx:-O9uKZZL9rj1ODizKbIy:bazcfxo) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table_version.py#L299)):*
> ```Python
>             col_name = schema_col_md.name if schema_col_md is not None else None
>             media_val = (
>                 MediaValidation.__members__[schema_col_md.media_validation.upper()]
> ```

remove `.__members__`
___
*[`pixeltable/catalog/table_version.py` line 300 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uKhkEECcZDMp2hE_0:-O9uKhkEECcZDMp2hE_1:b-e9gvjf) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table_version.py#L300)):*
> ```Python
>             media_val = (
>                 MediaValidation.__members__[schema_col_md.media_validation.upper()]
>                 #MediaValidation(schema_col_md.media_validation)
> ```

extraneous comment?
___
*[`pixeltable/catalog/table_version.py` line 729 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uKtWjCO35cnpYzpGt:-O9uKtWjCO35cnpYzpGu:b-5rg3r5) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table_version.py#L729)):*
> ```Python
>             with Env.get().engine.begin() as conn:
>                 return self._insert(
>                     plan, conn, time.time(), print_stats=print_stats, rowids=rowids(), abort_on_exc=fail_on_exception)
> ```

why `fail_on_exception` in one context but `abort_on_exc` in the other? kinda confusing
___
*[`pixeltable/exprs/column_property_ref.py` line 9 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uLB2A7DJJ4MtYN5n6:-O9uLB2A7DJJ4MtYN5n7:b3sdq2t) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_property_ref.py#L9)):*
> ```Python
> 
> import pixeltable.type_system as ts
> import pixeltable.catalog as catalog
> ```

`from pixeltable import catalog`
___
*[`pixeltable/exprs/column_property_ref.py` line 80 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uLoh-6Mf7d-znIc2a:-O9uLoh-6Mf7d-znIc2b:bpocul5) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_property_ref.py#L80)):*
> ```Python
>             assert data_row.has_val[self._col_ref.slot_idx]
>             data_row[self.slot_idx] = data_row.file_urls[self._col_ref.slot_idx]
>             return
> ```

it'd be cleaner to use `if/elif/.../else` with the `assert False` in the `else` block.
___
*[`pixeltable/exprs/column_property_ref.py` line 85 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uLx2v4EtqK7KA81E9:-O9uLx2v4EtqK7KA81EA:b-f6u598) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_property_ref.py#L85)):*
> ```Python
>             data_row[self.slot_idx] = data_row.file_paths[self._col_ref.slot_idx]
>             return
>         if self.prop == self.Property.ERRORTYPE or self.prop == self.Property.ERRORMSG:
> ```

replace with `self.is_error_prop()`?
___
*[`pixeltable/exprs/column_property_ref.py` line 90 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uMASeDgxEkTxVew96:-O9uMASeDgxEkTxVew97:b86sfxu) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_property_ref.py#L90)):*
> ```Python
>                 data_row[self.slot_idx] = None
>             else:
>                 data_row[self.slot_idx] = type(exc).__name__ if self.prop == self.Property.ERRORTYPE else str(exc)
> ```

would be easier to read without the nested ifs:

```python
if exc is None:
    # ...
elif self.prop == self.Property.ERRORTYPE:
    # ...
else:
    # ...
```
___
*[`pixeltable/exprs/column_ref.py` line 40 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uMgct6yIkvrNKzM7V:-O9uMgct6yIkvrNKzM7W:b-cg3gun) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L40)):*
> ```Python
>     pos_idx: Optional[int]
>     id: int
>     validates: bool  # if True, performs media validation
> ```

call this `perform_validation`?
___
*[`pixeltable/exprs/column_ref.py` line 42 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uOpBz4okxXbV-Isb4:-O9uOpBz4okxXbV-Isb5:b-ykrl3l) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L42)):*
> ```Python
>     validates: bool  # if True, performs media validation
> 
>     def __init__(self, col: catalog.Column, perform_validation: Optional[bool] = None):
> ```

so the `perform_validation` parameter is only referenced if it's a media type. Maybe put an assertion somewhere:

```python
assert col.col_type.is_media_type() or perform_validation is None
```
___
*[`pixeltable/exprs/column_ref.py` line 60 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uOuQPD1AufeGxLvlf:-O9uOuQPD1AufeGxLvlg:bpkuhis) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L60)):*
> ```Python
>             # we perform media validation if the column is a media type and the validation is set to ON_READ,
>             # unless we're told not to
>             self.validates = perform_validation if perform_validation is not None else (
> ```

Maybe `self.validates` should be a dynamic property, not a member variable:

Is there a future where we can `alter_column()` to change the validation strategy? I would think there is (I create my workflow, start using it, and later discover it's getting tedious to do all that validation on write as my app scales up). In that scenario and with this implementation, all my existing `ColumnRef`s still refer to the old strategy. This will remediate on the next catalog load, but that's a strange time for the changes to take effect.
___
*[`pixeltable/exprs/column_ref.py` line 75 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uPncvBQyNIBpNH2GA:-O9uOCwa1fEzHf70efzq:b-1s5uqd) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L75)):*
> ```Python
>         return (
>             super()._id_attrs()
>             + [('tbl_id', self.col.tbl.id), ('col_id', self.col.id), ('validates', self.validates)]
> ```

The attribute element should probably be the `perform_validation` initializer parameter, not the internal derived `validates` flag
___
*[`pixeltable/exprs/column_ref.py` line 108 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uQ8016F9zWhbN5MQA:-O9uQ8016F9zWhbN5MQB:b-ia41wb) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L108)):*
> ```Python
> 
>     def _equals(self, other: ColumnRef) -> bool:
>         return self.col == other.col and self.validates == other.validates
> ```

same comment as above
___
*[`pixeltable/exprs/column_ref.py` line 147 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uQOazC8yLlmKA987x:-O9uQOazC8yLlmKA987y:b-8d6r2n) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L147)):*
> ```Python
>             except excs.Error as exc:
>                 # propagate the exception, but ignore it otherwise;
>                 # media validation errors don't cause exceptions during query execution
> ```

this probably isn't the default we want? experience so far is that users get confused when pixeltable swallows errors and it's a better user experience if exceptions are surfaced by default. this should follow the `on_error` directive of the user-level call to `add_column` or `insert`.

would it be a lot of work to incorporate that into this PR? if so, let's make sure we have a ticket for it
___
*[`pixeltable/exprs/data_row.py` line 98 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uQz476ULby1UxpDdb:-O9uQz476ULby1UxpDdc:b345tr0) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/data_row.py#L98)):*
> ```Python
>         if slot_idx is not None:
>             return self.excs[slot_idx] is not None
>         return any(exc is not None for exc in self.excs)
> ```

confusing semantics. it'd be a lot clearer to have a separate `has_any_exc()` method.
___
*[`pixeltable/exprs/data_row.py` line 105 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uRM-X51QP84ApUsYV:-O9uRM-Y5Z1iS_eQ1QHl:b-f8tvd0) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/data_row.py#L105)):*
> ```Python
>         for exc in self.excs:
>             if exc is not None:
>                 return exc
> ```

also confusing. why not simply a `get_all_excs()` method that returns `[exc for exc in self.excs if exc is not None]`?
___
*[`pixeltable/index/btree.py` line 13 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uSxEr45DJiyA8BrYZ:-O9uSxEr45DJiyA8BrY_:b-tk1l97) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/index/btree.py#L13)):*
> ```Python
> 
> if TYPE_CHECKING:
>     import pixeltable
> ```

why is this needed? see below
___
*[`pixeltable/index/btree.py` line 22 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uTAdP9sIRTh3vqlEN:-O9uTAdP9sIRTh3vqlEO:bhbnowi) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/index/btree.py#L22)):*
> ```Python
>     MAX_STRING_LEN = 256
> 
>     value_expr: 'pixeltable.exprs.Expr'
> ```

just use `exprs.Expr` since that's already getting imported
___
*[`pixeltable/index/btree.py` line 34 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uTE8A2tPRb0e71Snv:-O9uTE8A2tPRb0e71Snw:bdvrqus) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/index/btree.py#L34)):*
> ```Python
>         if not c.col_type.is_scalar_type() and not c.col_type.is_media_type():
>             raise excs.Error(f'Index on column {c.name}: B-tree index requires scalar or media type, got {c.col_type}')
>         from pixeltable.exprs import ColumnRef, ColumnPropertyRef
> ```

why do we need this? `exprs` is already imported. refer to them as `exprs.columnRef` etc and we avoid the extraneous inline import
___
*[`tests/test_table.py` line 892 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uUDjS5TQGzjwNkqlO:-O9uUDjS5TQGzjwNkqlP:bz7qynm) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/tests/test_table.py#L892)):*
> ```Python
>         schema = {'id': pxt.Required[pxt.Int], 'img': pxt.Image}
>         t = pxt.create_table('test', schema)
>         t.insert([{'id': i, 'img': None} for i in range(10)])
> ```

unnecessary list creation

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/328)* status: 0 of 28 files reviewed, 28 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/globals.py` line 95 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uG45X3l6nDBupHsVI:-OA_b5uK7LdRzCpq-UAS:b-kh86e0) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/globals.py#L95)):*

`catalog.MediaValidation._member_names_` is shorter
</blockquote></details>

As per Perplexity that is not part of the public API, and you're supposed to use __members__. I'm also not seeing it in the docs, but I am seeing __members__:
https://docs.python.org/3/library/enum.html#enum.EnumType
___
*[`pixeltable/globals.py` line 181 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uGPm-2GmXtlKWw3v0:-OA_bXQ7BbOgS7ZmYC-X:bb6un1j) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/globals.py#L181)):*

duplicated logic here; can this be factored out into a method of `MediaValidation`? (say, `MediaValidation.validated_lookup(val)` or some such, that does the check and then returns `cls[val]`)
</blockquote></details>

Done
___
*[`pixeltable/type_system.py` line 13 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uHgyy9cVfqqNThNbw:-OA_bhAA1SeO7yDMrdvs:b32w12) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/type_system.py#L13)):*

have you explored debugger options for using vscode yet? 

standardizing on vscode would reduce thrash in things like import sort order. vscode handles this in a very nice way (it uses isort which is configurable in pyproject.toml so you get a consistent set of import conventions applied at the project level)
</blockquote></details>

No
___
*[`pixeltable/catalog/column.py` line 30 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uIEn4DpokUJf3I7_Y:-OA_blb5Dvwhg4EnG649:br1bxze) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/column.py#L30)):*

What convention are you using for determining when a member variable gets an underscore? I'd thought we weren't using them for non-public-facing classes.

Might be good to revisit/codify our code style conventions.
</blockquote></details>

There's a property media_validation, this avoids the collision.
___
*[`pixeltable/catalog/table.py` line 250 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uJcxC7fIqTh9uAgK-:-OA_cVviBdbkOcio1xKP:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table.py#L250)):*

it'd be more internally consistent to have this `@property` return an enum, and do the string conversion in `get_metadata()`.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 459 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uJo1m6F9XAWt086mr:-OAcD6YP86pr3hTSNmTx:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table.py#L459)):*

another cargo-cult of this logic- it should definitely be factored out. the only difference is in the `Column {name}:` prefix in the error message, which could be a parameter of the validate method. (And maybe add `Table {name}` to the other instances for consistency)
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table_version.py` line 113 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uKMyuDVDH7giCP1Ka:-OA_x8nMB9wccKetBefc:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table_version.py#L113)):*

`.__members__` can be removed here
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table_version.py` line 299 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uKZZL9rj1ODizKbIx:-OA_xJlzBBDJDl9oo-mE:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table_version.py#L299)):*

remove `.__members__`
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table_version.py` line 300 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uKhkEECcZDMp2hE_0:-OA_xK5E930kATf3-MsH:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/catalog/table_version.py#L300)):*

extraneous comment?
</blockquote></details>

Done.
___
*[`pixeltable/exprs/column_property_ref.py` line 9 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uLB2A7DJJ4MtYN5n6:-OA_yWaQ6Ovg1Y1u6ncT:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_property_ref.py#L9)):*

`from pixeltable import catalog`
</blockquote></details>

Done.
___
*[`pixeltable/exprs/column_property_ref.py` line 80 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uLoh-6Mf7d-znIc2a:-OA_ykbgDMRlWUm8Jzx1:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_property_ref.py#L80)):*

it'd be cleaner to use `if/elif/.../else` with the `assert False` in the `else` block.
</blockquote></details>

Done.
___
*[`pixeltable/exprs/column_property_ref.py` line 85 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uLx2v4EtqK7KA81E9:-OA_yqwj4gjGQhZCPsVk:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_property_ref.py#L85)):*

replace with `self.is_error_prop()`?
</blockquote></details>

Done.
___
*[`pixeltable/exprs/column_property_ref.py` line 90 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uMASeDgxEkTxVew96:-OA_zFoE3ke_TvdmIThX:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_property_ref.py#L90)):*

would be easier to read without the nested ifs:

```python
if exc is None:
    # ...
elif self.prop == self.Property.ERRORTYPE:
    # ...
else:
    # ...
```
</blockquote></details>

Done.
___
*[`pixeltable/exprs/column_ref.py` line 147 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uQOazC8yLlmKA987x:-OAa--MP6uRhf1WIW3_i:b-jkv4an) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L147)):*

this probably isn't the default we want? experience so far is that users get confused when pixeltable swallows errors and it's a better user experience if exceptions are surfaced by default. this should follow the `on_error` directive of the user-level call to `add_column` or `insert`.

would it be a lot of work to incorporate that into this PR? if so, let's make sure we have a ticket for it
</blockquote></details>

I think that needs further discussion. The implemented behavior mimics what a write-validated column would do, which seems like a good starting point.
___
*[`pixeltable/index/btree.py` line 13 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uSxEr45DJiyA8BrYZ:-OAa-n3A6841FqDmNDrA:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/index/btree.py#L13)):*

why is this needed? see below
</blockquote></details>

Done.
___
*[`pixeltable/index/btree.py` line 22 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uTAdP9sIRTh3vqlEN:-OAa-neA4cck-PL2dgAc:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/index/btree.py#L22)):*

just use `exprs.Expr` since that's already getting imported
</blockquote></details>

Done.
___
*[`pixeltable/index/btree.py` line 34 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uTE8A2tPRb0e71Snv:-OAa-oi88egEjHjG03gM:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/index/btree.py#L34)):*

why do we need this? `exprs` is already imported. refer to them as `exprs.columnRef` etc and we avoid the extraneous inline import
</blockquote></details>

Done.
___
*[`tests/test_table.py` line 892 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uUDjS5TQGzjwNkqlO:-OAa0Ccx1eXg8cJ5Q7Rz:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/tests/test_table.py#L892)):*

unnecessary list creation
</blockquote></details>

Done.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/328)* status: 0 of 28 files reviewed, 28 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/column_ref.py` line 147 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uQOazC8yLlmKA987x:-OAe5gWrBQCrPuHcd87X:bdv8stl) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L147)):*

I think that needs further discussion. The implemented behavior mimics what a write-validated column would do, which seems like a good starting point.
</blockquote></details>

Not anymore, right? We changed it so that write-validated columns raise an exception by default (`on_error = 'abort'`).

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/328)* status: 0 of 28 files reviewed, 28 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exprs/column_ref.py` line 40 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uMgct6yIkvrNKzM7V:-OAeh7Ib0H85aQSP-ypH:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L40)):*

call this `perform_validation`?
</blockquote></details>

Done.
___
*[`pixeltable/exprs/column_ref.py` line 42 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uOpBz4okxXbV-Isb4:-OAehTDk004eH0uwEBjG:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L42)):*

so the `perform_validation` parameter is only referenced if it's a media type. Maybe put an assertion somewhere:

```python
assert col.col_type.is_media_type() or perform_validation is None
```
</blockquote></details>

Done.
___
*[`pixeltable/exprs/column_ref.py` line 60 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uOuQPD1AufeGxLvlf:-OAei2HE1fL2hniyFdqs:b-28gjw5) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L60)):*

Maybe `self.validates` should be a dynamic property, not a member variable:

Is there a future where we can `alter_column()` to change the validation strategy? I would think there is (I create my workflow, start using it, and later discover it's getting tedious to do all that validation on write as my app scales up). In that scenario and with this implementation, all my existing `ColumnRef`s still refer to the old strategy. This will remediate on the next catalog load, but that's a strange time for the changes to take effect.
</blockquote></details>

To be addressed in the future.
___
*[`pixeltable/exprs/column_ref.py` line 75 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uPncvBQyNIBpNH2GA:-OAeiD0x6MtCVmuFzNlV:b796dfc) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L75)):*

The attribute element should probably be the `perform_validation` initializer parameter, not the internal derived `validates` flag
</blockquote></details>

That wouldn't make a difference, they'd be operationally identical.
___
*[`pixeltable/exprs/column_ref.py` line 147 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uQOazC8yLlmKA987x:-OAeiLBq7bAhwaME-9BH:b7dfydy) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/column_ref.py#L147)):*

Not anymore, right? We changed it so that write-validated columns raise an exception by default (`on_error = 'abort'`).
</blockquote></details>

Leaving that in place for now, as discussed.
___
*[`pixeltable/exprs/data_row.py` line 98 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uQz476ULby1UxpDdb:-OAeiO_NDTf6SZUolW65:b-hww6i2) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/data_row.py#L98)):*

confusing semantics. it'd be a lot clearer to have a separate `has_any_exc()` method.
</blockquote></details>

I think has_exc() without any parameter implies 'any'.
___
*[`pixeltable/exprs/data_row.py` line 105 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9uRM-X51QP84ApUsYV:-OAeiUoY7V6zV0aUTJNH:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/ce4b672622a34047d8aee84d59f19a05fe7c091e/pixeltable/exprs/data_row.py#L105)):*

also confusing. why not simply a `get_all_excs()` method that returns `[exc for exc in self.excs if exc is not None]`?
</blockquote></details>

Done.
___
*[`pixeltable/metadata/converters/convert_21.py` line 19 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/328#-O9FkZIV9fvnpg40ThDp:-OAeiYevEKRVBDOO9-Ub:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/1f9f7bdf9e942036e881f04b71370c59a933e989/pixeltable/metadata/converters/convert_21.py#L19)):*

add newline at end of file
</blockquote></details>

Done.

aaron-siegel (approved):

Issue #329, opened 2024-10-14
Some links are broken in Pixeltable Docs


denise-k:
Not necessarily a comprehensive list, just some ones I saw.
- In [Pixeltable Basics](https://docs.pixeltable.com/docs/pixeltable-basics), the link to Getting Started with Pixeltable is stale.
- In [Tables and Data Operations](https://docs.pixeltable.com/docs/tables-and-data-operations), the links to Pixeltable Fundamentals, Computed Columns, and Unstructured Data are stale.
- In [Computed Columns](https://docs.pixeltable.com/docs/computed-columns), the 2nd link to Tables and Data Operations is stale (1st is fine).
- In [UDFs](https://docs.pixeltable.com/docs/user-defined-functions-udfs), the link to Type System is stale.
- In [Object Detection in Videos](https://docs.pixeltable.com/docs/object-detection-in-videos), the link to
Pixeltable Basics is stale
- In [OpenAI tutorial](https://docs.pixeltable.com/docs/working-with-openai), the link to Pixeltable Basics is stale.

pierrebrunelle:
Thanks for reporting this. All fixed.

PR #330, opened 2024-10-15
Docs-fixed

pierrebrunelle:
Broken URLs fixed reported in #329


docs/release/howto/udfs-in-pixeltable.ipynb
```diff
@@ -345,7 +345,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "\r",
+      "\r\n",
```
aaron-siegel, line 348: When you save these files, Windows newlines get added all over the place. It's probably harmless, but we should really get you off Windows...

aaron-siegel (changes requested):
A bunch of those "blank" links are in there as placeholder reminders to fill them in when the relevant doc sections have been written. I'd prefer to not just delete them, but I agree it's not great to have broken links sitting around. Two options:
- Leave the links in, but route them to an "Under Construction" page with a message saying to check back in a few weeks for this content
- Remove the links and add a code comment in the notebook with something like:
   ```<!-- TODO: Add link to Expressions tutorial when completed -->```

docs/release/howto/udfs-in-pixeltable.ipynb
pierrebrunelle: 1. Let's not waste time building an **Under Construction** page and on top of that who wants to click on a link that redirects them to an "Under construction" page that leads to no additional insights. 
2. Adding a <!-- TODO to our actual live documentation makes it look unfinished. It's distracting to the reader. We should just track internally what notebooks/documentation we want to expand in the future and not expose that WIP through commentary in the live doc.

Let's remove the sections mentioning future materials as we do not know when we might get to work on them and our plans change.

Regarding Windows, as you said, it's harmless and it's good to have one of us running Windows for testing.

aaron-siegel: If you put a TODO in the format I proposed it will be an HTML comment that won't show up when the notebook cell is rendered.

pierrebrunelle: As discussed. Let's approve and merge it and add the TODO in HTML as we go where we feel like we need additions and/or as we build more tutorials.

aaron-siegel (approved):

PR #333, opened 2024-10-18
Back out changes to BtreeIndex.str_filter and mark as `@staticmethod`

aaron-siegel:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/333)

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/333)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`pixeltable/index/btree.py` line 18 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/333#-O9Vsty682cn76LdjEvh:-O9W-lD_0oVvlug8VoCh:b-i0ll1h) ([raw file](https://github.com/pixeltable/pixeltable/blob/31de72ed206bf89ee30e6392089ad8972b386ad3/pixeltable/index/btree.py#L18)):*

Why staticmethod first? Doesn't it end up getting an argument that's not even a function?
</blockquote></details>

It doesn't work if it's the other way around (mypy and tests both fail).

PR #334, opened 2024-10-18
Replicate adapter

aaron-siegel:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/334)

pixeltable/exprs/data_row.py
```diff
@@ -123,11 +123,18 @@ def __getitem__(self, index: object) -> Any:
         assert self.has_val[index], index
 
         if self.file_urls[index] is not None and index in self.img_slot_idxs:
-            # if we need to load this from a file, it should have been materialized locally
-            assert self.file_paths[index] is not None
-            if self.vals[index] is None:
-                self.vals[index] = PIL.Image.open(self.file_paths[index])
+            if self.file_urls[index].startswith('data:'):
+                # It's a base64-encoded data URL. Decode it now.
+                with urllib.request.urlopen(self.file_urls[index]) as response:
```
mkornacker: Didn't this already get downloaded by the CachePrefetchNode and the file is available in file_paths?

aaron-siegel: That’s what I expected as well, and it’s indeed the case when defining a computed column. But it fails with an assertion error if selecting the image cast dynamically without defining a new column, as is done in the example notebook. 

Maybe that’s a bug - and this probably needs more investigation. let’s discuss at standup Monday.

mkornacker: Let's punt on that bug right now (and change the notebook not to trigger it). Doing the download during expr evaluation would have very poor performance at the moment.

We'll fix that later in a larger overhaul of expr evaluation and parallelization.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/334)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`pixeltable/exprs/data_row.py` line 128 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/334#gh-1807115215:-O9zLpcW7RiFSwOz0Dc4:bu56mb3) ([raw file](https://github.com/pixeltable/pixeltable/blob/3cb75848b7b86fe5e34cc6cf95060b1eccb0b0fc/pixeltable/exprs/data_row.py#L128)):*

Let's punt on that bug right now (and change the notebook not to trigger it). Doing the download during expr evaluation would have very poor performance at the moment.

We'll fix that later in a larger overhaul of expr evaluation and parallelization.
</blockquote></details>

"change the notebook not to trigger it" = what specifically do you have in mind?

i don't see a way to meaningfully implement the replicate API without this. can we just keep the workaround in place for now and i'll add a TODO about it? it's not actually downloading anything - all it's doing is converting a b64-encoded `data:` url to an image. (note i specifically didn't enable this for http urls)

pixeltable/exprs/data_row.py
mkornacker: Okay, so this isn't really a url, it's the data inlined. I didn't understand that before.

How does the data: url end up in here? I'm assuming TypeCast gets a string and then calls ImageType.create_literal()? It seems like that's where the conversion should happen. file_urls should stick with file urls.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/334)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`pixeltable/exprs/data_row.py` line 128 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/334#gh-1807115215:-OA3XWod2iPzZuYhZR57:b-pdktvw) ([raw file](https://github.com/pixeltable/pixeltable/blob/3cb75848b7b86fe5e34cc6cf95060b1eccb0b0fc/pixeltable/exprs/data_row.py#L128)):*

Okay, so this isn't really a url, it's the data inlined. I didn't understand that before.

How does the data: url end up in here? I'm assuming TypeCast gets a string and then calls ImageType.create_literal()? It seems like that's where the conversion should happen. file_urls should stick with file urls.
</blockquote></details>

Well, it IS a url, just not an http url :)

But processing data: urls in create_literal() makes sense. I'll make that change

pixeltable/type_system.py
```diff
@@ -798,6 +799,19 @@ def _from_dict(cls, d: dict) -> ColumnType:
     def to_sa_type(self) -> sql.types.TypeEngine:
         return sql.String()
 
+    def _create_literal(self, val: Any) -> Any:
+        if isinstance(val, str) and val.startswith('data:'):
```
mkornacker, line 803: There should probably be a check for the correct media type here, given that the information is available (and would produce a more specific error message).

mkornacker (approved):
Another alternative (or: thing to add) would be to have an explicit udf (data_url_to_img(data_url: str) -> PIL.Image.Image or something like that).

We do have an explicit b64_encode(img) -> str udf in the image module (rather than doing that implicitly in a TypeCast).

pixeltable/exec/cache_prefetch_node.py
```diff
@@ -99,7 +99,7 @@ def _fetch_url(self, row: exprs.DataRow, slot_idx: int) -> Optional[str]:
                     if self.boto_client is None:
                         self.boto_client = get_client()
                     self.boto_client.download_file(parsed.netloc, parsed.path.lstrip('/'), str(tmp_path))
-            elif parsed.scheme == 'http' or parsed.scheme == 'https':
+            elif parsed.scheme == 'http' or parsed.scheme == 'https' or parsed.scheme == 'data':
```
mkornacker: Let's back this out as well. We should never see images stored as strings in the db.

aaron-siegel (commented):
Not sure about this one. URLs are special and I'd argue we need to handle them in a uniform way. An application might not know in advance whether an incoming URL is an `http` or `data` URL, and some APIs might reasonably return either depending on the situation.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/334#-:-OAGsOsTEgb_xvWuvPPt:b9b2gsr)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/type_system.py` line 803 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/334#gh-1817703174:-OAGs0eV4rsjOz0HsEte:bha8q64) ([raw file](https://github.com/pixeltable/pixeltable/blob/a002f2b0b28e5d0f368d97a1facf86c011e5716f/pixeltable/type_system.py#L803)):*

There should probably be a check for the correct media type here, given that the information is available (and would produce a more specific error message).
</blockquote></details>

I'm not sure what specifically is being suggested here - are you advocating for handling the below exception in a different way rather than falling through to `validate_literal`?
___
*[`pixeltable/exec/cache_prefetch_node.py` line 102 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/334#gh-1817706012:-OAGsBBw7W_Wwv-lJqdi:bwm0knm) ([raw file](https://github.com/pixeltable/pixeltable/blob/a002f2b0b28e5d0f368d97a1facf86c011e5716f/pixeltable/exec/cache_prefetch_node.py#L102)):*

Let's back this out as well. We should never see images stored as strings in the db.
</blockquote></details>

Hmmm. I agree with you that they should never be stored as strings in the db. But I also think we can't just reject them, for the same reason cited above: it's important to treat URLs in a uniform way because applications won't necessarily be able to predict which type of URL they receive. (Here I'm thinking of URLs that show up in computed image columns; I'm not anticipating anyone inserting `data:` URLs directly.)

Probably the best thing to do on seeing a `data:` URL is decode it, write the image to the media store, and then put a reference to the file in the DB. I'll back out this change, but create a ticket as a reminder that we need to address this somehow

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/334)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/type_system.py` line 803 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/334#gh-1817703174:-OAH4rRrE5n_ylCw0SeH:b-8nz98y) ([raw file](https://github.com/pixeltable/pixeltable/blob/a002f2b0b28e5d0f368d97a1facf86c011e5716f/pixeltable/type_system.py#L803)):*

I'm not sure what specifically is being suggested here - are you advocating for handling the below exception in a different way rather than falling through to `validate_literal`?
</blockquote></details>

Exactly. Check whether this is an image type, and if not return an error message that this is the wrong kind of data url.
___
*[`pixeltable/exec/cache_prefetch_node.py` line 102 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/334#gh-1817706012:-OAH5Lr95R8KWyE9KYmV:b-csemc3) ([raw file](https://github.com/pixeltable/pixeltable/blob/a002f2b0b28e5d0f368d97a1facf86c011e5716f/pixeltable/exec/cache_prefetch_node.py#L102)):*

Hmmm. I agree with you that they should never be stored as strings in the db. But I also think we can't just reject them, for the same reason cited above: it's important to treat URLs in a uniform way because applications won't necessarily be able to predict which type of URL they receive. (Here I'm thinking of URLs that show up in computed image columns; I'm not anticipating anyone inserting `data:` URLs directly.)

Probably the best thing to do on seeing a `data:` URL is decode it, write the image to the media store, and then put a reference to the file in the DB. I'll back out this change, but create a ticket as a reminder that we need to address this somehow
</blockquote></details>

Yes, we definitely can't store a data url containing an image in the varchar column that expects to be storing a file reference (it may well exceed the maximum varchar size).

We also don't want udfs to return data urls instead of pil.image.image. Explicit type casts seem to be the only way to handle data urls sensibly.

mkornacker (approved):

PR #339, opened 2024-10-23
Type-checking correctness for remaining Pixeltable packages

aaron-siegel:
- `mypy pixeltable` now succeeds
- Add `py.typed` marker to repo
- Bugfixes in `ArraySlice`


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/339)

pixeltable/exprs/array_slice.py
```diff
@@ -59,7 +59,7 @@ def _as_dict(self) -> Dict:
         return {'index': index, **super()._as_dict()}
 
     @classmethod
-    def _from_dict(cls, d: Dict, components: List[Expr]) -> Expr:
+    def _from_dict(cls, d: dict, components: list[Expr]) -> ArraySlice:
```
mkornacker, line 62: Why not return Self then? (here and elsewhere)

pixeltable/exprs/column_property_ref.py
```diff
@@ -33,15 +34,17 @@ def __init__(self, col_ref: ColumnRef, prop: Property):
     def default_column_name(self) -> Optional[str]:
         return str(self).replace('.', '_')
 
-    def _equals(self, other: ColumnRef) -> bool:
+    def _equals(self, other: ColumnPropertyRef) -> bool:
```
mkornacker, line 37: Self?

pixeltable/exprs/expr.py
```diff
@@ -226,12 +225,12 @@ def is_bound_by(self, tbl: catalog.TableVersionPath) -> bool:
                 return False
         return True
 
-    def retarget(self, tbl: catalog.TableVersionPath) -> Self:
+    def retarget(self, tbl: catalog.TableVersionPath) -> Union[Self, 'exprs.ColumnRef']:
```
mkornacker: Why is that Union important?

pixeltable/exprs/expr.py
```diff
@@ -329,11 +330,8 @@ def tbl_ids(self) -> set[UUID]:
         return {ref.col.tbl.id for ref in self.subexprs(ColumnRef)} | {ref.tbl.id for ref in self.subexprs(RowidRef)}
 
     @classmethod
-    def list_tbl_ids(cls, expr_list: list[Expr]) -> set[UUID]:
-        ids: set[UUID] = set()
-        for e in expr_list:
-            ids.update(e.tbl_ids())
-        return ids
+    def all_tbl_ids(cls, exprs: Iterable[Expr]) -> set[UUID]:
```
mkornacker: I don't like using 'exprs' as a variable, because that's the canonical module name.

pixeltable/exprs/method_ref.py
```diff
@@ -36,10 +37,10 @@ def _as_dict(self) -> dict:
         return {'method_name': self.method_name, **super()._as_dict()}
 
     @classmethod
-    def _from_dict(cls, d: dict, components: list[Expr]) -> Expr:
+    def _from_dict(cls, d: dict, components: list[Expr]) -> 'MethodRef':
         assert 'method_name' in d
         assert len(components) == 1
-        return cls(d['method_name'], components[0])
+        return cls(components[0], d['method_name'])
```
mkornacker, line 43: Underlining the need for adding catalog reloads to our tests...

pixeltable/exprs/row_builder.py
```diff
@@ -3,7 +3,7 @@
 import sys
 import time
 from dataclasses import dataclass
-from typing import Any, Dict, Iterable, List, Optional, Sequence, Set, Tuple
+from typing import Any, Iterable, Optional, Sequence
```
mkornacker, line 6: Could you leave this untouched? I already made those changes in the media validation PR, and I'd like to avoid having to resolve all those merge conflicts.

pixeltable/exprs/sql_element_cache.py
```diff
@@ -31,4 +31,4 @@ def contains(self, items: Union[Expr, Iterable[Expr]]) -> bool:
         """Returns True if every item has a (non-None) sql.ColumnElement."""
         if isinstance(items, Expr):
             return self.get(items) is not None
-        return all(self.get(e) is not None for e in items)
+        return all(self.get(e) is not None for e in cast(Iterable[Expr], items))
```
mkornacker: how about just prefixing the return with 
assert isinstance(...)?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/339)* status: 0 of 43 files reviewed, 7 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/array_slice.py` line 62 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/339#gh-1815087611:-O9zDq7NALfYjeF9JFSX:bqczrqn) ([raw file](https://github.com/pixeltable/pixeltable/blob/883ca0528365020dc278565f4b172c2642f2e8ff/pixeltable/exprs/array_slice.py#L62)):*

Why not return Self then? (here and elsewhere)
</blockquote></details>

`Self` is really intended for cases where the concrete type doesn't suffice: typically to enforce that all subclasses that override the method also specialize it. if a class has no subclasses, then it shouldn't be necessary, and i'd argue it's best to avoid it and use the concrete type.
___
*[`pixeltable/exprs/column_property_ref.py` line 37 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/339#gh-1815088293:-O9zEGjtDBsip3ukXTJi:bkzwfy1) ([raw file](https://github.com/pixeltable/pixeltable/blob/883ca0528365020dc278565f4b172c2642f2e8ff/pixeltable/exprs/column_property_ref.py#L37)):*

Self?
</blockquote></details>

same comment as above; no need for it here
___
*[`pixeltable/exprs/expr.py` line 228 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/339#gh-1815091848:-O9zZkowEnDJ4f3VXsk9:b3alerv) ([raw file](https://github.com/pixeltable/pixeltable/blob/883ca0528365020dc278565f4b172c2642f2e8ff/pixeltable/exprs/expr.py#L228)):*

Why is that Union important?
</blockquote></details>

Hmm good catch. Mypy was thrown off by the explicit special-casing of `ColumnRef`.

I've fixed this by moving the specialized logic out into a `ColumnRef._retarget()` override (which is a cleaner pattern anyway).
___
*[`pixeltable/exprs/expr.py` line 333 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/339#gh-1815105291:-O9zZz2CDGUSppdMpVcA:bt2vr1b) ([raw file](https://github.com/pixeltable/pixeltable/blob/883ca0528365020dc278565f4b172c2642f2e8ff/pixeltable/exprs/expr.py#L333)):*

I don't like using 'exprs' as a variable, because that's the canonical module name.
</blockquote></details>

Fixed.
___
*[`pixeltable/exprs/method_ref.py` line 43 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/339#gh-1815130954:-O9zEbEN9M17k7b9HoGd:b-60ia6i) ([raw file](https://github.com/pixeltable/pixeltable/blob/883ca0528365020dc278565f4b172c2642f2e8ff/pixeltable/exprs/method_ref.py#L43)):*

Underlining the need for adding catalog reloads to our tests...
</blockquote></details>

Actually not in this case: `MethodRef` is never actually written to the DB - it's not a true expression (I remember us talking about this awhile back), just a transient object that gets thrown away when the expression is formed. So in this case this is "dead code" and never actually exercised. Only mypy cares :)

If this bug were actually meaningful, I would expect the migration tests to fail.

Maybe we should make this explicit by having `MethodRef._from_dict()` etc throw a `NotImplementedError` instead.
___
*[`pixeltable/exprs/row_builder.py` line 6 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/339#gh-1815134630:-O9zF3nBB8MPQy5n_4pV:b-6jrehf) ([raw file](https://github.com/pixeltable/pixeltable/blob/883ca0528365020dc278565f4b172c2642f2e8ff/pixeltable/exprs/row_builder.py#L6)):*

Could you leave this untouched? I already made those changes in the media validation PR, and I'd like to avoid having to resolve all those merge conflicts.
</blockquote></details>

That's not really a fair request - you're asking me to do extra work in order to save you some work, and it'd be more effort to back this out than to just deal with the conflicts (most of which git will auto-resolve anyway).

Merge conflicts are a fact of life and IMO not a good reason to complicate our eng workflow, except in rare cases (e.g. colliding repo-wide changes that could trigger conflicts in many files).

If you're really worried about it then let's stage the merges so the media validation PR goes in first, and I'll deal with the merge conflicts. I just don't think they're a big deal; it's what, 10 minutes of work to deal with them?
___
*[`pixeltable/exprs/sql_element_cache.py` line 34 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/339#gh-1815139534:-O9zacSzBnoJKQggw2lW:b-fsf0b) ([raw file](https://github.com/pixeltable/pixeltable/blob/883ca0528365020dc278565f4b172c2642f2e8ff/pixeltable/exprs/sql_element_cache.py#L34)):*

how about just prefixing the return with 
assert isinstance(...)?
</blockquote></details>

mypy doesn't accept that (possibly it's a bug in mypy?) and i prefer the cast to a `type: ignore`.

FWIW i don't love this pattern, i'd prefer separate `contains()` and `contains_all()` methods, which would obviate the typing problem

mkornacker (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/339)* status: 0 of 43 files reviewed, 7 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exprs/sql_element_cache.py` line 34 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/339#gh-1815139534:-OA1eJDr2LMW6Thtwvb-:b6xmipd) ([raw file](https://github.com/pixeltable/pixeltable/blob/883ca0528365020dc278565f4b172c2642f2e8ff/pixeltable/exprs/sql_element_cache.py#L34)):*

mypy doesn't accept that (possibly it's a bug in mypy?) and i prefer the cast to a `type: ignore`.

FWIW i don't love this pattern, i'd prefer separate `contains()` and `contains_all()` methods, which would obviate the typing problem
</blockquote></details>

Feel free to make that change.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/339)* status: 0 of 43 files reviewed, 7 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/sql_element_cache.py` line 34 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/339#gh-1815139534:-OAYvtzy7dPRpUOaMkOk:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/883ca0528365020dc278565f4b172c2642f2e8ff/pixeltable/exprs/sql_element_cache.py#L34)):*

Feel free to make that change.
</blockquote></details>

Done.

PR #341, opened 2024-10-25
Update Whisper library version

aaron-siegel:
OpenAI finally (after almost a year) published a new pypi release of Whisper, so we no longer need to rely on a repo reference for the dev dependency.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/341)

mkornacker (approved):

PR #343, opened 2024-10-25
Support XML document type

aaron-siegel:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/343)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/343)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @aaron-siegel)

___
*[`pixeltable/utils/documents.py` line 62 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/343#-OA6dW4D1GSuwVJK-LCl:-OA6dW4D1GSuwVJK-LCm:b-vcmxd) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8c952518cad3b86b629b1f40a6acbdc4f591549/pixeltable/utils/documents.py#L62)):*
> ```Python
>     try:
>         with open(path, 'r', encoding='utf8') as fp:
>             doc = bs4.BeautifulSoup(fp, 'html.parser')
> ```

We should probably switch this to the lxml html parser, if available. 

This is what the bs4 docs recommend. See:
https://www.crummy.com/software/BeautifulSoup/bs4/doc/#installing-a-parser

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/343)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`pixeltable/utils/documents.py` line 62 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/343#-OA6dW4D1GSuwVJK-LCl:-OAEEpLzDq3yCct_5OtF:bvg2p2t) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8c952518cad3b86b629b1f40a6acbdc4f591549/pixeltable/utils/documents.py#L62)):*

We should probably switch this to the lxml html parser, if available. 

This is what the bs4 docs recommend. See:
https://www.crummy.com/software/BeautifulSoup/bs4/doc/#installing-a-parser
</blockquote></details>

That would involve adding `lxml` as a required dependency. Let's discuss that separately? I don't think it needs to be entangled with this PR since it's a separate issue

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/343)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @aaron-siegel)

___
*[`pixeltable/utils/documents.py` line 62 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/343#-OA6dW4D1GSuwVJK-LCl:-OAEjedEBrMSZwf5HkDk:b-s8xbsx) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8c952518cad3b86b629b1f40a6acbdc4f591549/pixeltable/utils/documents.py#L62)):*

That would involve adding `lxml` as a required dependency. Let's discuss that separately? I don't think it needs to be entangled with this PR since it's a separate issue
</blockquote></details>

I meant: you can check dynamically if lxml is available. If so, use their parser.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/343)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`pixeltable/utils/documents.py` line 62 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/343#-OA6dW4D1GSuwVJK-LCl:-OAEpcjqEHe1UnUh_ZYu:bgrzbwx) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8c952518cad3b86b629b1f40a6acbdc4f591549/pixeltable/utils/documents.py#L62)):*

I meant: you can check dynamically if lxml is available. If so, use their parser.
</blockquote></details>

I don't think that's a good solution. It means the observed behavior of the document splitter will change depending only on which dependencies are installed.

i.e., if the user `pip install`s lxml for an unrelated reason, then their document splitter workloads will suddenly behave differently, even if there is no change to their Pixeltable version, Beautifulsoup version, or Pixeltable config. That'd be pretty surprising.

We should either make lxml an explicit option that the user has to configure, or else make it the default and a required dependency.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/343)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`pixeltable/utils/documents.py` line 62 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/343#-OA6dW4D1GSuwVJK-LCl:-OAGgisA4Y3Q843XmxDV:b-jucv8m) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8c952518cad3b86b629b1f40a6acbdc4f591549/pixeltable/utils/documents.py#L62)):*

I don't think that's a good solution. It means the observed behavior of the document splitter will change depending only on which dependencies are installed.

i.e., if the user `pip install`s lxml for an unrelated reason, then their document splitter workloads will suddenly behave differently, even if there is no change to their Pixeltable version, Beautifulsoup version, or Pixeltable config. That'd be pretty surprising.

We should either make lxml an explicit option that the user has to configure, or else make it the default and a required dependency.
</blockquote></details>

We should maybe just add `lxml` as a required dependency. It's super-lightweight (adds no transitive dependencies), it's what beautifulsoup recommends for html (as you point out), and it wouldn't hurt to have xml natively supported as a doc type.

mkornacker (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/343)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @aaron-siegel)

___
*[`pixeltable/utils/documents.py` line 62 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/343#-OA6dW4D1GSuwVJK-LCl:-OAMvNdd8sAIjJhPbbNB:b-xx1nd6) ([raw file](https://github.com/pixeltable/pixeltable/blob/a8c952518cad3b86b629b1f40a6acbdc4f591549/pixeltable/utils/documents.py#L62)):*

We should maybe just add `lxml` as a required dependency. It's super-lightweight (adds no transitive dependencies), it's what beautifulsoup recommends for html (as you point out), and it wouldn't hurt to have xml natively supported as a doc type.
</blockquote></details>

Let's go with that, then.

PR #347, opened 2024-10-29
Ensure FunctionCall return types have correct nullability

aaron-siegel:
Ensures that a `FunctionCall` will have nullable return type if any of its non-nullable parameters have a nullable input. This conforms to Pixeltable's behavior of defaulting to `None` any time a `FunctionCall` sees such an input.

Previously, the column type was reported as non-nullable but `None`s were being generated, an inconsistency.

### DIFF
--- a/pixeltable/exprs/function_call.py
+++ b/pixeltable/exprs/function_call.py
@@ -50,10 +50,24 @@ class FunctionCall(Expr):
         if group_by_clause is None:
             group_by_clause = []
         signature = fn.signature
-        super().__init__(fn.call_return_type(bound_args))
+        return_type = fn.call_return_type(bound_args)
         self.fn = fn
         self.is_method_call = is_method_call
         self.normalize_args(fn.name, signature, bound_args)
+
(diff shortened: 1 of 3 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/347)

pixeltable/exprs/function_call.py
```diff
@@ -50,11 +50,25 @@ def __init__(
...
         self.normalize_args(fn.name, signature, bound_args)
 
+        # If `return_type` is non-nullable, but the function call has a nullable input to any of its non-nullable
+        # parameters, then we need to make it nullable. This is because Pixeltable defaults a function output to
+        # `None` when any of its non-nullable inputs are `None`.
+        for arg_name, arg in bound_args.items():
+            param = signature.parameters[arg_name]
+            if (
+                param.col_type is not None and not param.col_type.nullable
+                and isinstance(arg, Expr) and arg.col_type.nullable
```
mkornacker, line 65: isinstance() is redundant: after normalize_args() all args should be Expr instances.

mkornacker (approved):

PR #353, opened 2024-11-02
Test infrastructure improvements

aaron-siegel:
Several improvements to our test infra:

1. `make test` now runs `mypy` and `mkdocs build --strict` in addition to `pytest`. It was too easy to forget to run these checks locally and then get burned in CI. To run pytest only, use `make pytest`.

2. Use `pytest-xdist` to parallelize test execution. `make test` now enables this by default. Tests in CI will still run on a single process. I've limited parallelism to 6 concurrent processes for now (above this, there seems to be no benefit, and the tests actually take longer due to increased initialization costs). The time to run `make pytest` is reduced by about 2/3, from ~326 seconds to ~112 seconds on my machine.

    You can ensure that a particular subset of tests runs on the same worker with (for example) `@pytest.mark.xdist_group('label_studio')`.

3. Use `pytest-rerunfailures` to allow flaky tests to rerun multiple times. Some tests depend on downloading data or models from remote servers, which can be unreliable; it's especially annoying when they fail in CI. Flaky tests can now be marked with:

    `@pytest.mark.flaky(reruns=3)`

    which will cause them to run `n` times before failing. To rerun only on specific exceptions, use (for example):

    `@pytest.mark.flaky(reruns=3, only_rerun='HfHubHTTPError')`

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -160,7 +160,7 @@ jobs:
       - name: Ensure pytest is installed
         # This is necessary for running the tests without --with dev
         if: ${{ matrix.poetry-options == '' }}
-        run: pip install pytest coverage
+        run: pip install pytest pytest-xdist pytest-rerunfailures coverage
       - name: Run the unit tests
         if: ${{ matrix.test-category == 'py' }}
         # Run the tests with the 'expensive' marker only once, on ubuntu-22.04 with Python 3.9. The other tests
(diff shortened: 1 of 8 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/353)

mkornacker (approved):

PR #363, opened 2024-11-09
Partial bindings for `Function`

aaron-siegel:
Introduces a `Function.using()` method that creates a partial binding of the function parameters as an `ExprTemplateFunction`. If `f` is an n-ary `Function`, then

`f.using(a=..., b=..., ...)`

is an (n-k)-ary function whose parameters are all the parameters of `f` not among `a, b, ...`.

This is useful for specifying indices without having to create explicit `@expr_udf`s.

TODO: Is `using` what we want to call it?
TODO: Add tests for validation errors


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/363)

pixeltable/func/function.py
```diff
@@ -99,6 +103,39 @@ def conditional_return_type(self, fn: Callable[..., ts.ColumnType]) -> Callable[
         self._conditional_return_type = fn
         return fn
 
+    def using(self, **kwargs) -> 'ExprTemplateFunction':
```
mkornacker: **kwargs: Any?

pixeltable/func/function.py
```diff
@@ -99,6 +103,39 @@ def conditional_return_type(self, fn: Callable[..., ts.ColumnType]) -> Callable[
...
+                raise excs.Error(f'Expected type {param.col_type}; got {expr.col_type}')
+            bindings[k] = v  # Use the original value, not the Expr (The Expr is only for validation)
+
+        residual_params = [
+            p for p in self.signature.parameters.values() if p.name not in bindings
+        ]
+
+        # Bind each remaining parameter to a like-named variable
+        for param in residual_params:
+            if param.name not in bindings:
```
mkornacker: This should always be the case, as per line 123, no?

mkornacker (commented):
I think this requires more doc updates. 
- the UDF tutorial should describe this in detail
- the embedding and vector indices tutorial should link to that and also have some brief explanation; for example, in 'creating an index' it refers to UDFs as the embedding function parameters, but the example contains partial bindings; I think this will confuse the reader

pixeltable/func/function.py
aaron-siegel: Done.

pixeltable/func/function.py
aaron-siegel: Ah good catch. I reorganized the code at one point and forgot to remove this check :) Fixed.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/363)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

pixeltable/index/embedding_index.py
```diff
@@ -136,7 +136,12 @@ def _validate_embedding_fn(cls, embed_fn: func.Function, name: str, expected_typ
         """Validate the signature"""
         assert isinstance(embed_fn, func.Function)
         sig = embed_fn.signature
-        if len(sig.parameters) != 1 or sig.parameters_by_pos[0].col_type.type_enum != expected_type:
+
+        # The embedding function must be a 1-ary function of the correct type. But it's ok if the function signature
+        # has more than one parameter, as long as it has at most one *required* parameter.
+        if (len(sig.parameters) == 0
+            or len(sig.required_parameters) > 1
```
mkornacker, line 143: why not just len(required_params) == 1?

tests/test_function.py
```diff
@@ -296,6 +296,38 @@ def c(x: int, y: int) -> int:
...
+        assert pb2.signature.required_parameters[0].name == 'p2'
+
+        t = pxt.create_table('test', {'c1': pxt.String, 'c2': pxt.String, 'c3': pxt.String})
+        t.insert(c1='a', c2='b', c3='c')
+        t.add_computed_column(pb1=pb1(t.c1, t.c3))
+        t.add_computed_column(pb2=pb2(t.c2))
+        t.add_computed_column(pb3=pb3(p4='changed'))
+        res = t.select(t.pb1, t.pb2, t.pb3).collect()
+        assert res[0] == {'pb1': 'a y c default', 'pb2': 'x b z default', 'pb3': 'x y z changed'}
+
```
mkornacker, line 322: Also add test for 
t.select(pb1(p1=t.c1))
to make sure that this path also picks up on missing required params.

mkornacker (commented):
Could you also add a link in the RAG tutorial to the UDF tutorial in this paragraph:
The following command creates a vector index on the text column in the chunks table, using the E5 embedding model. Note that defining the index is sufficient in order to load it with the existing data (and also to update it when the underlying data changes, as we'll see later).

pixeltable/index/embedding_index.py
aaron-siegel: Because it's in theory fine if all the function parameters have defaults, as long as there is at least one parameter.

tests/test_function.py
aaron-siegel: Done.

aaron-siegel (commented):
Done - everything except updating the UDF tutorial, which I don't see as urgent (only use case at the moment is for embeddings). In the RAG tutorial, I linked to the embeddings how-to guide.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/363#-:-OBlMhYPCdHa8fH2lZig:b-fftrh9)* status: 0 of 16 files reviewed, 4 unresolved discussions (waiting on @mkornacker)

PR #366, opened 2024-11-10
Use torchaudio for preprocessing instead of librosa

aaron-siegel:
I'd initially used librosa for audio preprocessing in `speech2text_for_conditional_generation`. I'd assumed it was pulled in as a dependency by torchaudio; this turned out to be false (it was coming into our dependency tree from elsewhere).

This PR switches to use torchaudio for preprocessing instead. It's a little more work this way, but we avoid requiring librosa as a dependency in order to use it. (torchaudio is already needed by the model)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/366)

mkornacker (approved):

PR #372, opened 2024-11-13
PXT-330 Fix a bug in create_insert_plan

anushas:
This commit fixes a bug in Planner.create_insert_plan where an insert continues to include a column even after it was dropped. This was because it was using catalog.TableVersion.cols to get the list of columns, but that tracks history of all the columns in the table/view. Fix is to use catalog.TableVersion.cols_by_id that tracks the current list of columns.

Testing: new testcase that reproduces the bug passes with the fix.

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -883,6 +883,23 @@ class TestTable:
         for tup in t.collect():
             assert tup['c1'] == 'this is a python string'
 
+    def test_insert_bad_udf_bug_repro(self, reset_db) -> None:
+        @pxt.udf(_force_stored=True)
+        def bad_udf(x: str) -> str:
+            assert False
+
+        t = pxt.create_table('test', {'str_col': pxt.String})
+        t.add_column(bad=bad_udf(t.str_col))  # Succeeds because the table has no data
+        t.drop_column('bad')
+        # Before fix the following statement would fail with an exception
+        # even though 'bad' was dropped. This was due to a bug in Planner.create_insert_plan
+        # using catalog.TableVersion.cols to get the list of columns.
+        # catalog.TableVersion.cols tracks history of all the columns, including
+        # dropped ones. Fixed Planner.create_insert_plan to instead use
+        # catalog.TableVersion.cols_by_id that tracks the current list of columns
+        # to make this test pass.
+        t.insert(str_col='Hello there.')
+
     def test_query(self, reset_db) -> None:
         skip_test_if_not_installed('boto3')
         col_names = ['c1', 'c2', 'c3', 'c4', 'c5']
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -875,9 +875,9 @@ class TestTable:
             t.insert(c5=np.ndarray((3, 2)))
         assert 'expected ndarray((2, 3)' in str(exc_info.value)
 
-        # test that insert with a bad column succeeds when the
-        # table is empty and after the bad column is dropped
-        # because expression evaluation is skipped
+        # test that insert skips expression evaluation and does not
+        # populate any columns when the table is empty, and for any
+        # columns that are not part of the current schema.
         @pxt.udf(_force_stored=True)
         def bad_udf(x: str) -> str:
             assert False
@@ -894,7 +894,6 @@ class TestTable:
         assert t.count() == 1
         for tup in t.collect():
             assert tup['c1'] == 'this is a python string'
-
 
     def test_query(self, reset_db) -> None:
         skip_test_if_not_installed('boto3')
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -875,6 +875,18 @@ class TestTable:
             t.insert(c5=np.ndarray((3, 2)))
         assert 'expected ndarray((2, 3)' in str(exc_info.value)
 
+        # test that insert with a bad column succeeds when the
+        # table is empty and after the bad column is dropped
+        # because expression evaluation is skipped
+        @pxt.udf(_force_stored=True)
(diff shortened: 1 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/372)

aaron-siegel (changes requested):
Nice simple fix :)

Just some minor comments. I want @mkornacker to look at this one too before committing it.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/372#-:-OBaTEQL9qrEKr_jRKLE:b-acz5qj)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @anushas)

___
*[`tests/test_table.py` line 886 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/372#-OBaTNEHBLOJFAy3NY_H:-OBaTNEHBLOJFAy3NY_I:bjeflms) ([raw file](https://github.com/pixeltable/pixeltable/blob/49c26e3086f11eed33225c536fd54d81053b25aa/tests/test_table.py#L886)):*
> ```Python
>             assert tup['c1'] == 'this is a python string'
> 
>     def test_insert_bad_udf_bug_repro(self, reset_db) -> None:
> ```

Not sure we need a separate test for this? Maybe just add it to `test_add_column()`.
___
*[`tests/test_table.py` line 894 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/372#-OBaTggB1Zm7wjAf8JUF:-OBaTggB1Zm7wjAf8JUG:b-6lpotn) ([raw file](https://github.com/pixeltable/pixeltable/blob/49c26e3086f11eed33225c536fd54d81053b25aa/tests/test_table.py#L894)):*
> ```Python
>         t.add_column(bad=bad_udf(t.str_col))  # Succeeds because the table has no data
>         t.drop_column('bad')
>         # Before fix the following statement would fail with an exception
> ```

These comments aren't going to make as much sense out of context. I don't think they need to be a permanent part of the codebase - maybe just a shorter comment saying something like "test that expression evaluation is skipped for columns that have been dropped"?

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/372)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @aaron-siegel)

___
*[`tests/test_table.py` line 886 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/372#-OBaTNEHBLOJFAy3NY_H:-OBcb7N36OTyCZ3i8uMg:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/49c26e3086f11eed33225c536fd54d81053b25aa/tests/test_table.py#L886)):*

Not sure we need a separate test for this? Maybe just add it to `test_add_column()`.
</blockquote></details>

Done.
___
*[`tests/test_table.py` line 894 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/372#-OBaTggB1Zm7wjAf8JUF:-OBcb8_00_35mbWswG6-:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/49c26e3086f11eed33225c536fd54d81053b25aa/tests/test_table.py#L894)):*

These comments aren't going to make as much sense out of context. I don't think they need to be a permanent part of the codebase - maybe just a shorter comment saying something like "test that expression evaluation is skipped for columns that have been dropped"?
</blockquote></details>

Done.

tests/test_table.py
```diff
@@ -875,6 +875,18 @@ def test_insert(self, reset_db) -> None:
             t.insert(c5=np.ndarray((3, 2)))
         assert 'expected ndarray((2, 3)' in str(exc_info.value)
 
+        # test that insert with a bad column succeeds when the
```
mkornacker: Wording: this isn't really about a 'bad' column, we just want to make sure that it's not trying to populate columns that aren't part of the current schema.

tests/test_table.py
```diff
@@ -883,6 +895,7 @@ def test_insert_string_with_null(self, reset_db) -> None:
         for tup in t.collect():
             assert tup['c1'] == 'this is a python string'
 
+
```
mkornacker: Extra blank line

mkornacker (approved):

tests/test_table.py
anushas: done

tests/test_table.py
anushas: done.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/372)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @anushas and @mkornacker)

___
*[`tests/test_table.py` line 879 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/372#-OBg0Ydh5Q5e2j1l0IuT:-OBg0Ydh5Q5e2j1l0IuU:b-lf842t) ([raw file](https://github.com/pixeltable/pixeltable/blob/08eeb1ef069d1d553b88e5b293e483e99e286b65/tests/test_table.py#L879)):*
> ```Python
> 
>         # test that insert skips expression evaluation and does not
>         # populate any columns when the table is empty, and for any
> ```

It's mostly about testing that expression evaluation is skipped for columns that are not in the current schema (the fact that it doesn't evaluate expressions on an empty table is kind of a tautology; I'd just put that comment in the repro for clarity, not because it's important to the test)

"test that insert skips expression evaluation for columns that are not part of the current schema"

aaron-siegel (approved):

PR #377, opened 2024-11-17
Basic joins

mkornacker:
Added DataFrame.join(other, on=..., how=...), where 'other' must be a Table instance (support for 'other' to be a DataFrame is follow-on work).

This is breaking TestSnapshot::test_multiple_snapshot_paths, which looks like a bug in Planner.create_update_plan() (that was previously covered up by some logic that is now missing). I'll address that soon, but the core join logic can be reviewed now.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/377)

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/377)* status: 0 of 23 files reviewed, 16 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/dataframe.py` line 531 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCBAT116gym-s1ByUlz:-OCH0AGf2V6PyUs1Brp5:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b913e8cf2f1eadb712d7185483326576cbae5c19/pixeltable/dataframe.py#L531)):*

Give an example here; it's not clear a priori what's the difference between a "user-specified join condition" and a "join predicate". An example would help clarify.
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 602 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCBIXmpC66RBfg6XcYS:-OCGzpBkBSurkJNlFJxT:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b913e8cf2f1eadb712d7185483326576cbae5c19/pixeltable/dataframe.py#L602)):*

Suggestion: allow `on` to be a positional argument (I don't see a good reason not to allow this, and I'd find it convenient).
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 552 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCF4AYu0IsIk0vxe8x_:-OCH09FZ1XbXtpRbWk6r:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/dataframe.py#L552)):*

Suggestion: disallow the Sequence form for predicates and make the signature

`on: Union[exprs.Expr, Sequence[exprs.ColumnRef]]`

it's a simpler API, and a bit clearer, with no loss of expressive power (since a sequence of predicates can obviously be written just as easily, and more clearly, as a single predicate). This also helps address the ambiguity noted below
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 561 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCF3Zcr9pHD3BNJ4m4u:-OCH3Lp15PP2t-W7T1Qp:bt2vr1b) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/dataframe.py#L561)):*

There's an edge case when it comes to boolean-valued columns... I'm not at all sure which is the most logical treatment in that case.
</blockquote></details>

Fixed.
___
*[`pixeltable/dataframe.py` line 619 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCF4ruE-W-XqGya25Gf:-OCH-3amBTn0PfZjyp5e:b-1w3s0e) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/dataframe.py#L619)):*

Add a note clarifying that

`join(other, t.my_col)`

is exactly equivalent to

`join(other, t.my_col == other.my_col)`

(in some sense this is implied by the verbiage, but writing it out programatically would make it a lot clearer)
</blockquote></details>

I think the example below is doing exactly that.
___
*[`pixeltable/dataframe.py` line 631 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCF5CPc2d-Y4ioekPe1:-OCH-8H_5rJ2akufSywo:bl6f14s) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/dataframe.py#L631)):*

This was called `full_outer` in the literal definition above, but personally I prefer just `outer`, the meaning of which is clear to me and fairly standardized.
</blockquote></details>

The standard meaning of 'outer join'  is "any type of outer join: left, right, full". I changed it to full_outer.

https://en.wikipedia.org/wiki/Join_(SQL)#Outer_join
___
*[`pixeltable/dataframe.py` line 643 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCF5W5621KouemgXCEw:-OCH-kzC10v0sf9us-Qp:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/dataframe.py#L643)):*

spell out why it's ambiguous
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 654 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCF5bjh8Oeesf1RbBJz:-OCH06mZ3z5WRU-tMXgc:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/dataframe.py#L654)):*

this is the pattern that i don't see as serving any real purpose. isn't it just as easy, and somewhat clearer, to write:

```python
df = t.join(d, on=(t.d1 == d.pk1) & (t.d2 == d.pk2), how='left')
```
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 185 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCFFArv7I_7uYU1solU:-OCJnXLB0ZFIF64abwKD:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/catalog/table.py#L185)):*

use `JoinTypeLiterals` here
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 187 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCFFGrn9fifF-vqXh2o:-OCJnffZ5mffjtcZJKM9:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/catalog/table.py#L187)):*

spurious newline
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 188 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCFFLRTEY4XUUZFUSj7:-OCJniLAD0vTSDNdlbWX:bgaz3vo) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/catalog/table.py#L188)):*

incorrect docstring
</blockquote></details>

All of the dataframe methods say that. We should decide what we want to do there.
___
*[`pixeltable/exec/sql_node.py` line 234 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCFFwFa97vKWt56ip-v:-OCJiAzdAGloc2fOSLAy:b-r7oyqs) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/exec/sql_node.py#L234)):*

alternatively, could use a property setter
</blockquote></details>

I think that would make sense if there were a getter, but that doesn't apply here.
___
*[`pixeltable/exec/sql_node.py` line 240 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCFFyZv3UCpzN6lqb4l:-OCJiIIA9Ft4rEhbas9Q:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/exec/sql_node.py#L240)):*

incidentally: `filter` is a Python builtin, and the usual best practice is to avoid shadowing builtins; we might consider a different symbol (such as `filter_`) in all the places where we're currently using `filter`
</blockquote></details>

Done.
___
*[`pixeltable/exec/sql_node.py` line 463 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCFGrApBJHU9G1RaNN6:-OCJrfhsCRMHLBRzhzMl:b-a1p86l) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/exec/sql_node.py#L463)):*

I wonder if there's a point in even keeping CROSS around as a JoinType enum choice. (We should keep `'cross'` in the user-facing API, but converting it to an inner join with predicate `True` at an earlier stage would simplify the downstream logic. At minimum we could fill in the predicate early on rather than keeping it as Optional, which would avoid having to bury a special case deep inside the execution logic)
</blockquote></details>

As discussed, we need to record 'cross' in order to use it in __repr__. SqlJoinNode does the translation to SQL, so I think this is the correct place to handle that case.

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/377)* status: 0 of 23 files reviewed, 5 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/dataframe.py` line 148 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCB9iUhDBlBtaH1uwQu:-OCKrKc50IlUTA2dtP7S:biki44g) ([raw file](https://github.com/pixeltable/pixeltable/blob/b913e8cf2f1eadb712d7185483326576cbae5c19/pixeltable/dataframe.py#L148)):*

Do we actually need the `tbl` parameter any more? Might make sense to eliminate it in favor of a required `from_clause` parameter that subsumes a single-table clause as a special case (since this is an internal API, it makes sense to optimize for simplicity)
</blockquote></details>

Bumping this comment - it didn't get a reply, so I'm not sure if you missed it, or if you saw it but decided not to act on it.
___
*[`pixeltable/dataframe.py` line 531 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCBAT116gym-s1ByUlz:-OCKp_kJ81D0rBta0syL:b-rb385a) ([raw file](https://github.com/pixeltable/pixeltable/blob/b913e8cf2f1eadb712d7185483326576cbae5c19/pixeltable/dataframe.py#L531)):*

Done.
</blockquote></details>

I don't see anything new here? (Example and/or clarification in the docstring?)
___
*[`pixeltable/dataframe.py` line 631 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCF5CPc2d-Y4ioekPe1:-OCKq1yr7HsFB7zRPFvQ:b32w1t) ([raw file](https://github.com/pixeltable/pixeltable/blob/e88fff27eccb1adb3de423e1bd0f9b91c7e677d4/pixeltable/dataframe.py#L631)):*

The standard meaning of 'outer join'  is "any type of outer join: left, right, full". I changed it to full_outer.

https://en.wikipedia.org/wiki/Join_(SQL)#Outer_join
</blockquote></details>

Ok
___
*[`pixeltable/plan.py` line 63 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCKqa1CBt8jtXU9siuT:-OCKqa1CBt8jtXU9siuU:byj1yhl) ([raw file](https://github.com/pixeltable/pixeltable/blob/55b0f4f5002c04d9cef82ba4c9c1ff7a80415457/pixeltable/plan.py#L63)):*
> ```Python
>             raise excs.Error(f'{error_prefix} must be one of: [{val_strs}]')
> 
> JoinType.LiteralType = Literal[tuple(e.name.lower() for e in JoinType)]
> ```

Is the static typechecker going to understand what to do with this?

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/377)* status: 0 of 23 files reviewed, 5 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/dataframe.py` line 144 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCASYWO4qnUDNCbM3G5:-OCLefBgCkaZtPLX2-nG:bb6un1j) ([raw file](https://github.com/pixeltable/pixeltable/blob/b913e8cf2f1eadb712d7185483326576cbae5c19/pixeltable/dataframe.py#L144)):*

Why `full_outer` instead of the briefer `outer`? Just curious. (I'm accustomed to `outer` carrying that implication unless qualified with `left` or `right`)

Also, remove `'right'` for now if we're not going to support it.
</blockquote></details>

Done
___
*[`pixeltable/dataframe.py` line 148 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCB9iUhDBlBtaH1uwQu:-OCLS2C--MBuIvDt8xFr:b-jg43yu) ([raw file](https://github.com/pixeltable/pixeltable/blob/b913e8cf2f1eadb712d7185483326576cbae5c19/pixeltable/dataframe.py#L148)):*

Bumping this comment - it didn't get a reply, so I'm not sure if you missed it, or if you saw it but decided not to act on it.
</blockquote></details>

Oversight.
___
*[`pixeltable/dataframe.py` line 531 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCBAT116gym-s1ByUlz:-OCLSRdE8QuiEmw42cWI:bwzzh5c) ([raw file](https://github.com/pixeltable/pixeltable/blob/b913e8cf2f1eadb712d7185483326576cbae5c19/pixeltable/dataframe.py#L531)):*

I don't see anything new here? (Example and/or clarification in the docstring?)
</blockquote></details>

I removed the docstring.
___
*[`pixeltable/plan.py` line 63 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/377#-OCKqa1CBt8jtXU9siuT:-OCLSYek7iBT_18966PT:b-3v3kt) ([raw file](https://github.com/pixeltable/pixeltable/blob/55b0f4f5002c04d9cef82ba4c9c1ff7a80415457/pixeltable/plan.py#L63)):*

Is the static typechecker going to understand what to do with this?
</blockquote></details>

Good point, not likely. Changed.

PR #380, opened 2024-11-19
Support JSON Schemas for validating `JsonType` columns

aaron-siegel:
Allows a `JsonType` to be qualified with a JSON Schema. In particular, since every Pydantic model can be exported as a JSON Schema, this allows Pydantic models to be used for validation. (However, custom Python validators in the model will not be applied; only validators that are representable in a JSON Schema are currently applied.)

TODO: In the current PR, we do row-level validation, including field validators such as "greater than", on an insert(); but on an update(), we only do schema-validation against the type of the update expression, but not row-level field validation. Should we: (1) stick with this (asymmetric) pattern for now; (2) also do field validation on update(); or (3) skip field validators on insert(), only doing type validation?

TODO: What about batch_update?


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/380)

pixeltable/type_system.py
```diff
@@ -650,24 +661,134 @@ def _validate_literal(self, val: Any) -> None:
...
 
     def _create_literal(self, val: Any) -> Any:
         if isinstance(val, tuple):
             val = list(val)
         return val
 
+    def supertype(self, other: ColumnType) -> Optional[JsonType]:
+        basic_supertype = super().supertype(other)
+        if basic_supertype is not None:
+            # This will be much faster when it works
```
mkornacker: cryptic comment

pixeltable/type_system.py
```diff
@@ -650,24 +661,134 @@ def _validate_literal(self, val: Any) -> None:
...
+        superschema = self.__superschema(self.json_schema, other.json_schema)
+
+        return JsonType(
+            json_schema=(None if len(superschema) == 0 else superschema),
+            nullable=(self.nullable or other.nullable)
+        )
+
+    @classmethod
+    def __superschema(cls, a: dict[str, Any], b: dict[str, Any]) -> Optional[dict[str, Any]]:
+        # Defining a general type hierarchy over all JSON schemas would be a challenging problem. In order to keep
```
mkornacker, line 708: I'm finding this comment hard to follow. After reading it I still don't understand what it's trying to do.

pixeltable/type_system.py
aaron-siegel: lol i don't even know where that comment came from. it shouldn't be there. removed it

pixeltable/type_system.py
aaron-siegel: ok we are weeks outside of SLA on this PR. i don't have time right now to be revisiting code comments so let's either ship this PR or close it. you and pierre decide, i don't care at this point

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/380)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

pixeltable/type_system.py
aaron-siegel: Oh, I remember what the comment was for. clarified it

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/380)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

mkornacker (approved):

PR #382, opened 2024-11-19
Python 3.13 Support

aaron-siegel:
Adds Python 3.13 to CI. Updates to use the latest `pixeltable-pgserver` which is now publishing Python 3.13 wheels.

Not many changes were needed; the main change involved incrementing the versions of various libraries that we're using in our dev setup. Those version changes are buried in our dev `poetry.lock` and are not user-visible (no impact on `pip install` behavior).

WhisperX has to be taken out of Poetry due to an overly specific version constraint (it requires `av==11.*` which is not Python-3.13 compatible).


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/382)

mkornacker (approved):

PR #383, opened 2024-11-21
PXT-352 gracefully error when using a catalog.Table after it is dropped

anushas:
This commit changes the catalog.Table APIs to consistently check when the table is already dropped, and error out gracefully and early Before some APIs did the check, few simply suceeded as if the table was still active, and most failed with unfriendly errors errors during SQL execution later.

Testing: added a new testcase to verify.

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -46,7 +46,7 @@ class Table(SchemaObject):
     def __init__(self, id: UUID, dir_id: UUID, name: str, tbl_version_path: TableVersionPath):
         super().__init__(id, name, dir_id)
         self._is_dropped = False
-        self._tbl_version_path = tbl_version_path
+        self.__tbl_version_path = tbl_version_path
         self.__query_scope = self.QueryScope(self)
 
     class QueryScope:
@@ -116,13 +116,13 @@ class Table(SchemaObject):
     @property
     def _tbl_version(self) -> TableVersion:
         """Return TableVersion for just this table."""
-        return self.tbl_version_path.tbl_version
+        return self._tbl_version_path.tbl_version
 
     @property
-    def tbl_version_path(self) -> TableVersionPath:
+    def _tbl_version_path(self) -> TableVersionPath:
         """Return TableVersionPath for just this table."""
         self._check_is_dropped()
-        return self._tbl_version_path
+        return self.__tbl_version_path
 
     def __hash__(self) -> int:
         return hash(self._tbl_version.id)
@@ -135,7 +135,7 @@ class Table(SchemaObject):
         """Return a ColumnRef for the given name.
         """
         self._check_is_dropped()
-        return self._tbl_version_path.get_column_ref(name)
+        return self.__tbl_version_path.get_column_ref(name)
 
     @overload
     def __getitem__(self, name: str) -> 'pxt.exprs.ColumnRef': ...
@@ -176,7 +176,7 @@ class Table(SchemaObject):
         """Return a DataFrame for this table.
         """
         # local import: avoid circular imports
-        return pxt.DataFrame(self.tbl_version_path)
+        return pxt.DataFrame(self._tbl_version_path)
 
     @property
     def queries(self) -> 'Table.QueryScope':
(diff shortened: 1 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/383)

aaron-siegel (approved):
I note that we could simplify the code a bit by making `_tbl_version_path` into a `@property` and putting `_check_is_dropped()` in that property, which would enable us to remove it from most of the separate existing places (and might be good practice anyway, to safeguard any future methods that get added that use it)

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/383#-:-OCFJtt463Su9hLwgDWD:b-8aopc5)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @anushas)

___
*[`tests/test_table.py` line 1725 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/383#-OCFKJ0M8EEeXFkp_Cdn:-OCFKJ0M8EEeXFkp_Cdo:bdyzrdy) ([raw file](https://github.com/pixeltable/pixeltable/blob/4c3970d2dccaf85ea2eb3b4a229a10f1b6320adc/tests/test_table.py#L1725)):*
> ```Python
>         _ = tbl.select(tbl.id, tbl._name).collect()
> 
>     def test_table_api_on_dropped_table(self, reset_db: None) -> None:
> ```

Thanks for adding such thorough tests for this check!
___
*[`pixeltable/catalog/table.py` line 1162 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/383#-OCFJkAH-Z-k1QYbHPs1:-OCFJkAH-Z-k1QYbHPs2:b-svuf53) ([raw file](https://github.com/pixeltable/pixeltable/blob/4c3970d2dccaf85ea2eb3b4a229a10f1b6320adc/pixeltable/catalog/table.py#L1162)):*
> ```Python
>         Links the specified `ExternalStore` to this table.
>         """
> ```

This is redundant since it's already protected by the `_tbl_version` access.

anushas (commented):
That's a good point. If I add _tbl_version_path as a property, I get an AttributeError: can't set attribute in the catalog.Table.__init__ code where we set the self._tbl_version_path. To avoid that I would have to also create a property setter, but I am not sure if it is right to allow a setter (then it can be modified outside of instantiation?). I tried that but it led to a recursion call to the property. I need to understand this better. So deferred this for now.

> @property
def \_tbl\_version\_path(self) -> TableVersionPath:
   """Return TableVersionPath for just this table."""
    self.\_check\_is\_dropped()
    return self.\_tbl\_version\_path
>
> @\_tbl\_version\_path.setter
def \_tbl\_version\_path(self, tbl\_version\_path: TableVersionPath) -> None:
    self.\_tbl\_version\_path \= tbl\_version\_path

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/383#-:-OCIC_Vj7M7AaE97HVVx:b-pxb9is)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/catalog/table.py` line 1162 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/383#-OCFJkAH-Z-k1QYbHPs1:-OCICV9f6tW1Tjd1znic:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/4c3970d2dccaf85ea2eb3b4a229a10f1b6320adc/pixeltable/catalog/table.py#L1162)):*

This is redundant since it's already protected by the `_tbl_version` access.
</blockquote></details>

Done.

anushas:
Refreshed the PR to: a) Resolve merge conflict with pixeltbale/main. and b) Rename property tbl_version_path to _tbl_version_path since it is not a public property, and rename to the class member to __tbl_version_path.

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/383)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @anushas)

___
*[`pixeltable/catalog/table.py` line 138 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/383#-OCL8yzy5HZXREl4OmME:-OCL8yzy5HZXREl4OmMF:b-58lb30) ([raw file](https://github.com/pixeltable/pixeltable/blob/fd16f6fb05fef0683455210e5b4e7896dbb83ca8/pixeltable/catalog/table.py#L138)):*
> ```Python
>         """
>         return self.__tbl_version_path.get_column_ref(name)
> ```

should we refer to the property here (instead of the separate _check_is_dropped() check)?

PR #385, opened 2024-11-21
Lock av version in dev environment

aaron-siegel:
Locks the version of `av` used in the Pixeltable dev environment, keyed on Python version.

This should eliminate the need to do a `make clean` before `make install` when updating `poetry.lock`.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/385)

mkornacker (approved):

PR #387, opened 2024-11-22
Fix handling of Optional vs. Required in astype()

aaron-siegel:
- Ensures that astype() treats its type argument as if it appeared in a schema (nullable by default, Python builtins not allowed)
- Exprs of non-nullable type will remain non-nullable after an astype(), regardless of the astype() argument (since they are provably non-nullable)
- Casting from nullable to non-nullable is allowed and will generate errors if it encounters any Nones
- Adds unit tests to validate all of the above


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/387)

mkornacker (approved):

PR #394, opened 2024-11-27
PXT-361 - improve DataFrame docstrings

anushas:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/394)

aaron-siegel (changes requested):
General docstring style comment: in multiline docstrings, the first line should contain only the quote marks and nothing else, as in:

```
'''
This is a multiline docstring.

This is another line.
'''
```

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/394#-:-OD8P3OX6frvsYr6f1cL:b-w7iogn)* status: 0 of 1 files reviewed, 17 unresolved discussions (waiting on @anushas)

___
*[`pixeltable/dataframe.py` line 303 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Ky--5lNWus0wzB8B:-OD8Ky--5lNWus0wzB8C:b-88skov) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L303)):*
> ```Python
> 
>     def head(self, n: int = 10) -> DataFrameResultSet:
>         '''Get the first n rows of the DataFrame.
> ```

Let's keep the phrasing consistent with the similarly named methods in Table. The current docstring for `Table.head()` reads: "Return the first n rows inserted into this table." I prefer that one (in particular, it clarifies that "first" means "in insertion order"), but if we're going to update it let's do it consistently in both places.
___
*[`pixeltable/dataframe.py` line 320 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8L7hA115jbFfOj2sj:-OD8L7hBAbDt8RrPzo55:b-ovar6f) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L320)):*
> ```Python
> 
>     def tail(self, n: int = 10) -> DataFrameResultSet:
>         '''Get the last n rows of the DataFrame.
> ```

cf. `Table.tail()`
___
*[`pixeltable/dataframe.py` line 340 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8LwbB25uPN0Hb5yM_:-OD8LwbB25uPN0Hb5yMa:btfz5mt) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L340)):*
> ```Python
>     @property
>     def schema(self) -> dict[str, ColumnType]:
>         '''Get the dictionary representation of the schema of the DataFrame.
> ```

I prefer just "Get the schema"; it's clear that it's represented as a dictionary.
___
*[`pixeltable/dataframe.py` line 418 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8MA598OCSTIGpvKxn:-OD8MA598OCSTIGpvKxo:bfrcv4m) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L418)):*
> ```Python
> 
>     def count(self) -> int:
>         """Get the number of rows in the DataFrame.
> ```

cf. `Table.count()`
___
*[`pixeltable/dataframe.py` line 492 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8MDi96SlJL8jwY6sV:-OD8MDi96SlJL8jwY6sW:b-rg5od0) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L492)):*
> ```Python
> 
>     def select(self, *items: Any, **named_items: Any) -> DataFrame:
>         """ Select columns or expressions from the DataFrame.
> ```

cf `Table.select()`.

In this case I think the new wording is preferred and we should update `Table.select()` accordingly.
___
*[`pixeltable/dataframe.py` line 494 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8MMEd221E3kn0tvHv:-OD8MMEd221E3kn0tvHw:b-lwbgf1) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L494)):*
> ```Python
>         """ Select columns or expressions from the DataFrame.
> 
>         Error is raised if the select list is already specified, or if the expression is invalid.
> ```

Suggestions:

"Error" -> "An error"

"the expression is" -> "any of the specified expressions are"

Also, this should be moved to a "Raises:" block in the docstring (typically between "Returns:" and "Examples:")
___
*[`pixeltable/dataframe.py` line 506 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8N11lA0bdSr1nyEHk:-OD8N11lA0bdSr1nyEHl:bsatj0f) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L506)):*
> ```Python
>             Select the columns 'name' and 'age' from the DataFrame df:
> 
>             >>> df = df.select('name', 'age')
> ```

Give an example using kwargs syntax
___
*[`pixeltable/dataframe.py` line 557 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8NLysAHR_1GfbfqDO:-OD8NLysAHR_1GfbfqDP:b51dpqq) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L557)):*
> ```Python
> 
>     def where(self, pred: exprs.Expr) -> DataFrame:
>         """Filter rows based on a predicate.
> ```

cf. `Table.where()` (the new one is better)
___
*[`pixeltable/dataframe.py` line 559 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8NIVbDwaLY3yNI_fl:-OD8NIVbDwaLY3yNI_fm:b-180sav) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L559)):*
> ```Python
>         """Filter rows based on a predicate.
> 
>         Error is raised if the predicate is not a Pixeltable expression or if it does not return a boolean.
> ```

Error -> An error, and move to Raises:
___
*[`pixeltable/dataframe.py` line 565 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8NQ9FCio7WXp6gc9_:-OD8NQ9FCio7WXp6gc9a:buupn4y) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L565)):*
> ```Python
> 
>         Returns:
>             A new DataFrame with the spefied predicates added to the where-clause.
> ```

typo: specified

Also, the where clause won't be added, it will replace any existing one. We should probably revisit whether that is the correct behavior.
___
*[`pixeltable/dataframe.py` line 570 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Nr5y1_eE3gmVhj78:-OD8Nr5y1_eE3gmVhj79:b-yfdcjt) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L570)):*
> ```Python
>             Filter the DataFrame df to only include rows where the column 'age' is greater than 30:
> 
>             >>> df = df.where(df.age > 30)
> ```

This won't work (you can't dereference columns of a dataframe with `getattr` the way you can with a table)
___
*[`pixeltable/dataframe.py` line 724 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8NeRuCn5Fo6Pp1Zrd:-OD8NeRuCn5Fo6Pp1Zre:b-8v4z9h) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L724)):*
> ```Python
>         """ Add a group-by clause to this DataFrame.
> 
>         Error is raised if the group-by clause is already specified, or if the expression is invalid.
> ```

Error -> An error and move to Raises:
___
*[`pixeltable/dataframe.py` line 739 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8O5B16Q69RQ0gPDFq:-OD8O5B16Q69RQ0gPDFr:b-73bzym) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L739)):*
> ```Python
>             Group a DataFrame book by the 'genre' column:
> 
>             >>> df = book.group_by(genre)
> ```

Did you mean to put `genre` in quotes?
___
*[`pixeltable/dataframe.py` line 787 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8OBCx1ZTVSTeWtw-i:-OD8OBCx1ZTVSTeWtw-j:b-arv4v7) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L787)):*
> ```Python
>             Order a DataFrame book by two columns (price, pages) in descending order:
> 
>             >>> df = book.order_by(price, pages, asc=False)
> ```

Did you mean to quote the column names?
___
*[`pixeltable/dataframe.py` line 829 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8OJO8AekUlIJIS8ST:-OD8OJO8AekUlIJIS8SU:b-hl3c1r) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L829)):*
> ```Python
> 
>         Update rows in the DataFrame with the specified value_spec.
> ```

Should we mention that the underlying table must be a base table (not a view or snapshot)?
___
*[`pixeltable/dataframe.py` line 844 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8OkDk3txGz8mPWed4:-OD8OkDlDMclZ7FC4spm:b-x9zhd3) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L844)):*
> ```Python
>     def delete(self) -> UpdateStatus:
>         ''' Delete rows in the DataFrame.
> ```

Same comment as for update()
___
*[`pixeltable/dataframe.py` line 900 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8OoGB1plJWBJBtaY3:-OD8OoGB1plJWBJBtaY4:b-vh1i0v) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L900)):*
> ```Python
> 
>     @classmethod
>     def from_dict(cls, d: dict[str, Any]) -> 'DataFrame':
> ```

Is this intended to be a public-facing method? (If not, we should rename it to `_from_dict` so the docstring doesn't show up in our API docs)

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/394)* status: 0 of 1 files reviewed, 22 unresolved discussions (waiting on @aaron-siegel and @anushas)

___
*[`pixeltable/dataframe.py` line 303 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Ky--5lNWus0wzB8B:-OD8U4KQ8hUm3CYOHsg-:bas18uj) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L303)):*

Let's keep the phrasing consistent with the similarly named methods in Table. The current docstring for `Table.head()` reads: "Return the first n rows inserted into this table." I prefer that one (in particular, it clarifies that "first" means "in insertion order"), but if we're going to update it let's do it consistently in both places.
</blockquote></details>

I would argue that "the first n rows of the DataFrame" is simply incorrect - there is no concept of first outside of an order-by.

Also, head() does not work with joins (because it needs to refer to a specific table), and this should also be pointed out here.
___
*[`pixeltable/dataframe.py` line 340 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8LwbB25uPN0Hb5yM_:-OD8Uktp2MmGMBXbVUhG:b-bm429n) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L340)):*

I prefer just "Get the schema"; it's clear that it's represented as a dictionary.
</blockquote></details>

I don't think this should be exposed. ColumnType is not a user-facing class anymore.
___
*[`pixeltable/dataframe.py` line 900 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8OoGB1plJWBJBtaY3:-OD8XWkN6x0-TIWzvODY:b6kjk90) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L900)):*

Is this intended to be a public-facing method? (If not, we should rename it to `_from_dict` so the docstring doesn't show up in our API docs)
</blockquote></details>

This does not show up in the API docs, because it doesn't have a docstring.

It is *not* meant to be user-facing.
___
*[`pixeltable/dataframe.py` line 494 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Uz-aEb2YVcuZTbng:-OD8Uz-aEb2YVcuZTbnh:b-9vmurv) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L494)):*
> ```Python
>         """ Select columns or expressions from the DataFrame.
> 
>         Error is raised if the select list is already specified, or if the expression is invalid.
> ```


The Raises section is also mandated by the Google Python style guide.
___
*[`pixeltable/dataframe.py` line 506 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Vndl0ths2BFL4si4:-OD8Vndl0ths2BFL4si5:b-la6oa0) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L506)):*
> ```Python
>             Select the columns 'name' and 'age' from the DataFrame df:
> 
>             >>> df = df.select('name', 'age')
> ```

This will return the string literals 'name' and 'age'.
___
*[`pixeltable/dataframe.py` line 739 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8WFU-2FhPCj2LpfHC:-OD8WFU-2FhPCj2LpfHD:bcwinjy) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L739)):*
> ```Python
>             Group a DataFrame book by the 'genre' column:
> 
>             >>> df = book.group_by(genre)
> ```

See my earlier comment re select().

Also, I think this needs an example that contains actual aggregation. Someone coming from pandas would probably expect to see something coming out of group_by(), but that's not the case here.
___
*[`pixeltable/dataframe.py` line 785 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Wta87l8ct59D5f68:-OD8Wta87l8ct59D5f69:b-neh7uq) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L785)):*
> ```Python
> 
>         Examples:
>             Order a DataFrame book by two columns (price, pages) in descending order:
> ```

Add another example showing how to mix ascending and descending.
___
*[`pixeltable/dataframe.py` line 826 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8X4Wh8Js3A2kRI5ck:-OD8X4Wh8Js3A2kRI5cl:b-efmka5) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L826)):*
> ```Python
> 
>     def update(self, value_spec: dict[str, Any], cascade: bool = True) -> UpdateStatus:
>         ''' Update rows in the DataFrame.
> ```

This is misleading - dataframes don't contain rows.

Users coming from pandas will get tripped up by that. We need to be careful to make it clear that all data resides in tables.

anushas (commented):
Oh. I was going by https://peps.python.org/pep-0257/#multi-line-docstrings 
which says the first line is the summary similar to a single-line-docstring. 

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/394#-:-ODD16B83b9KAz6xWxIH:b6i15t7)* status: 0 of 1 files reviewed, 21 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/dataframe.py` line 303 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Ky--5lNWus0wzB8B:-ODD1vkY55QpECBLKDsI:ba2ch7x) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L303)):*

I would argue that "the first n rows of the DataFrame" is simply incorrect - there is no concept of first outside of an order-by.

Also, head() does not work with joins (because it needs to refer to a specific table), and this should also be pointed out here.
</blockquote></details>

I used pandas and polars documentation are reference, wherever possible. Both use the "first/last" qualifier for head/tail. We need to differentiate between the head and tail and count, can just say n rows. I don't think "inserted" applies to DataFrame, as we only allow inserts on Table.  So I have changed it to say Return instead of Get, and used "based on position" like pandas, and added comment regarding join, and a Raises section. 
___
*[`pixeltable/dataframe.py` line 320 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8L7hA115jbFfOj2sj:-ODDB4MbASmJu2JzJ65w:bnn3nud) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L320)):*

cf. `Table.tail()`
</blockquote></details>

same as head.
___
*[`pixeltable/dataframe.py` line 340 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8LwbB25uPN0Hb5yM_:-ODDBLRX1lOepb7a_1-b:blit9nd) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L340)):*

I don't think this should be exposed. ColumnType is not a user-facing class anymore.
</blockquote></details>

removed the docstring.
___
*[`pixeltable/dataframe.py` line 418 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8MA598OCSTIGpvKxn:-ODDBTW05Rk1q3SnVp-k:bnn3nud) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L418)):*

cf. `Table.count()`
</blockquote></details>

same as head.
___
*[`pixeltable/dataframe.py` line 492 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8MDi96SlJL8jwY6sV:-ODDjR3l33TgRD8mlgNH:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L492)):*

cf `Table.select()`.

In this case I think the new wording is preferred and we should update `Table.select()` accordingly.
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 494 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8MMEd221E3kn0tvHv:-ODDl7D414U4rYUZhTBR:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L494)):*

Suggestions:

"Error" -> "An error"

"the expression is" -> "any of the specified expressions are"

Also, this should be moved to a "Raises:" block in the docstring (typically between "Returns:" and "Examples:")
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 494 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Uz-aEb2YVcuZTbng:-ODDr2Wr55S8ElLr17b6:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L494)):*

The Raises section is also mandated by the Google Python style guide.
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 506 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8N11lA0bdSr1nyEHk:-ODDr7ex4_TlzkZheaOE:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L506)):*

Give an example using kwargs syntax
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 506 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Vndl0ths2BFL4si4:-ODDr85d5jp2AlIJb7HN:b-gv48sg) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L506)):*

This will return the string literals 'name' and 'age'.
</blockquote></details>

ah. corrected. it looks like we only accept column reference from the table from which we create the dataframe. it was not very intuitive .
___
*[`pixeltable/dataframe.py` line 557 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8NLysAHR_1GfbfqDO:-ODDrRIa36o25mvHvYxm:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L557)):*

cf. `Table.where()` (the new one is better)
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 559 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8NIVbDwaLY3yNI_fl:-ODDrRj68TfmMY2EkMD6:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L559)):*

Error -> An error, and move to Raises:
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 565 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8NQ9FCio7WXp6gc9_:-ODDsLomEk65hhPydZvL:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L565)):*

typo: specified

Also, the where clause won't be added, it will replace any existing one. We should probably revisit whether that is the correct behavior.
</blockquote></details>

done
___
*[`pixeltable/dataframe.py` line 570 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Nr5y1_eE3gmVhj78:-ODDsqeHBca7ljuJSRUE:b-4for5i) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L570)):*

This won't work (you can't dereference columns of a dataframe with `getattr` the way you can with a table)
</blockquote></details>

corrected.
___
*[`pixeltable/dataframe.py` line 724 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8NeRuCn5Fo6Pp1Zrd:-ODDt1QD3X4KqmO-9eTN:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L724)):*

Error -> An error and move to Raises:
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 739 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8O5B16Q69RQ0gPDFq:-ODDuqgj9kuH5B4EEQDM:b-ovzcc7) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L739)):*

Did you mean to put `genre` in quotes?
</blockquote></details>

yes, but i realize that doesnt work. corrected to use reference.
___
*[`pixeltable/dataframe.py` line 739 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8WFU-2FhPCj2LpfHC:-ODE7X_A5fgZE8HNZKac:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L739)):*

See my earlier comment re select().

Also, I think this needs an example that contains actual aggregation. Someone coming from pandas would probably expect to see something coming out of group_by(), but that's not the case here.
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 785 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Wta87l8ct59D5f68:-ODEAii7EG94BL4e1Aht:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L785)):*

Add another example showing how to mix ascending and descending.
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 787 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8OBCx1ZTVSTeWtw-i:-ODEAl1T4yzYOvdrGJBx:b-4for5i) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L787)):*

Did you mean to quote the column names?
</blockquote></details>

corrected.
___
*[`pixeltable/dataframe.py` line 826 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8X4Wh8Js3A2kRI5ck:-ODEBPUw5_FPKSa1s9rf:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L826)):*

This is misleading - dataframes don't contain rows.

Users coming from pandas will get tripped up by that. We need to be careful to make it clear that all data resides in tables.
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 829 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8OJO8AekUlIJIS8ST:-ODEBOKC5CUYhE445lec:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L829)):*

Should we mention that the underlying table must be a base table (not a view or snapshot)?
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 844 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8OkDk3txGz8mPWed4:-ODEBx5hDn4jiiuvQpjl:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L844)):*

Same comment as for update()
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 900 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8OoGB1plJWBJBtaY3:-ODED74L1YUOygsMVsMJ:blit9nd) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L900)):*

This does not show up in the API docs, because it doesn't have a docstring.

It is *not* meant to be user-facing.
</blockquote></details>

removed the docstring.

anushas:
Think I addressed all the comments, PTAL @aaron-siegel @mkornacker when you get the chance.

aaron-siegel (changes requested):
Interesting. I hadn't read PEP-0257. It suggests either is ok. I ran a quick test with `ruff` and `pylint` and both of them accept either variant. So maybe there isn't a clear convention here. We may want to pick a standard but probably not a priority.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/394#-:-ODId9kzEY1i-Pgv83OH:b-mybibx)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @anushas and @mkornacker)

___
*[`pixeltable/dataframe.py` line 303 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Ky--5lNWus0wzB8B:-ODIc-mjDeRAymgWL7g2:b9zj7hk) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L303)):*

I used pandas and polars documentation are reference, wherever possible. Both use the "first/last" qualifier for head/tail. We need to differentiate between the head and tail and count, can just say n rows. I don't think "inserted" applies to DataFrame, as we only allow inserts on Table.  So I have changed it to say Return instead of Get, and used "based on position" like pandas, and added comment regarding join, and a Raises section.
</blockquote></details>

Right, but a Pixeltable DataFrame is not the same thing as a Pandas DataFrame. We definitely need to state that it's in insertion order (it's not clear what position means in this case). I understand what you're saying about DataFrames not allowing inserts - maybe it would be clearest to say something like "Return the first n rows of the DataFrame, in insertion order of the underlying Table."
___
*[`pixeltable/dataframe.py` line 506 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Vndl0ths2BFL4si4:-ODIeCFm6Ffoht9unYO3:b-5fwmzf) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L506)):*

ah. corrected. it looks like we only accept column reference from the table from which we create the dataframe. it was not very intuitive .
</blockquote></details>

For future discussion: We might think about allowing `df.name` to dereference columns of a DataFrame the same way we do with tables. I agree with @anushas that this pattern is not intuitive.
___
*[`pixeltable/dataframe.py` line 519 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/394#-ODIf31t1bynElikukNR:-ODIf31t1bynElikukNS:b-mjujr8) ([raw file](https://github.com/pixeltable/pixeltable/blob/156f2517d4bd4abcf412950607dd1dd04ec78ee7/pixeltable/dataframe.py#L519)):*
> ```Python
> 
>             Select the columns 'name' (referenced in table t) from the DataFrame person,
>             and a named column 'is_adult' from the expression 'age' >= 18 where 'age' is
> ```

the quotes around 'age' don't belong in the expression. Also: enclose the expression in backticks (it's Python code)
___
*[`pixeltable/dataframe.py` line 522 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/394#-ODIepWGDDhdhUUPKe_Y:-ODIepWGDDhdhUUPKe_Z:bfx3lrg) ([raw file](https://github.com/pixeltable/pixeltable/blob/156f2517d4bd4abcf412950607dd1dd04ec78ee7/pixeltable/dataframe.py#L522)):*
> ```Python
>             another column in table t:
> 
>             >>> df = person.select(t.name, is_adult=t.age>=18)
> ```

Python style: surround `>=` with spaces (either `is_adult=t.age >= 18` or `is_adult=(t.age >= 18)`)
___
*[`pixeltable/dataframe.py` line 583 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/394#-ODIfL529FDdIBLGgfLi:-ODIfL529FDdIBLGgfLj:b89de6s) ([raw file](https://github.com/pixeltable/pixeltable/blob/156f2517d4bd4abcf412950607dd1dd04ec78ee7/pixeltable/dataframe.py#L583)):*
> ```Python
> 
>         Raises:
>             Error: If the predicate is not a Pixeltable expression
> ```

The linebreaks won't be preserved when the docstring is compiled, so we need commas after the clauses or this will be unreadable. (Or convert it into a bulleted list.)

Note: if you want to preview the output, you can run `mkdocs build` from the commandline; this will generate output to target/site that you can open locally in your browser.
___
*[`pixeltable/dataframe.py` line 760 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/394#-ODIfhUT3JaTh04vUjvL:-ODIfhUT3JaTh04vUjvM:b2oex9o) ([raw file](https://github.com/pixeltable/pixeltable/blob/156f2517d4bd4abcf412950607dd1dd04ec78ee7/pixeltable/dataframe.py#L760)):*
> ```Python
>         Raises:
>             Error: If the group-by clause is already specified,
>                 or if the specified expression is invalid
> ```

add commas after the clauses?

anushas (commented):
sounds good.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/394#-:-ODNG1RfC8zQ6mfWdCO0:bp7vbga)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/dataframe.py` line 303 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Ky--5lNWus0wzB8B:-ODNDS1S2ILHh5hOwb6D:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L303)):*

Right, but a Pixeltable DataFrame is not the same thing as a Pandas DataFrame. We definitely need to state that it's in insertion order (it's not clear what position means in this case). I understand what you're saying about DataFrames not allowing inserts - maybe it would be clearest to say something like "Return the first n rows of the DataFrame, in insertion order of the underlying Table."
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 506 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/394#-OD8Vndl0ths2BFL4si4:-ODNErJZEZV8P5lqf0fM:bpn82hp) ([raw file](https://github.com/pixeltable/pixeltable/blob/eed545bc2fca0a10e6900463dd1783b7ff4cfb8c/pixeltable/dataframe.py#L506)):*

For future discussion: We might think about allowing `df.name` to dereference columns of a DataFrame the same way we do with tables. I agree with @anushas that this pattern is not intuitive.
</blockquote></details>

logged a low priority ticket PXT-369 to track this.
___
*[`pixeltable/dataframe.py` line 519 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/394#-ODIf31t1bynElikukNR:-ODNG7pMAcxbk3YQX-40:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/156f2517d4bd4abcf412950607dd1dd04ec78ee7/pixeltable/dataframe.py#L519)):*

the quotes around 'age' don't belong in the expression. Also: enclose the expression in backticks (it's Python code)
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 522 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/394#-ODIepWGDDhdhUUPKe_Y:-ODNF6or6unuOJ6anu4i:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/156f2517d4bd4abcf412950607dd1dd04ec78ee7/pixeltable/dataframe.py#L522)):*

Python style: surround `>=` with spaces (either `is_adult=t.age >= 18` or `is_adult=(t.age >= 18)`)
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 583 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/394#-ODIfL529FDdIBLGgfLi:-ODNFYqJ-6jG48acD13R:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/156f2517d4bd4abcf412950607dd1dd04ec78ee7/pixeltable/dataframe.py#L583)):*

The linebreaks won't be preserved when the docstring is compiled, so we need commas after the clauses or this will be unreadable. (Or convert it into a bulleted list.)

Note: if you want to preview the output, you can run `mkdocs build` from the commandline; this will generate output to target/site that you can open locally in your browser.
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 760 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/394#-ODIfhUT3JaTh04vUjvL:-ODNFpo02TFcx36QNL6d:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/156f2517d4bd4abcf412950607dd1dd04ec78ee7/pixeltable/dataframe.py#L760)):*

add commas after the clauses?
</blockquote></details>

Done.

pixeltable/catalog/table.py
```diff
@@ -184,26 +184,26 @@ def queries(self) -> 'Table.QueryScope':
         return self.__query_scope
 
     def select(self, *items: Any, **named_items: Any) -> 'pxt.DataFrame':
-        """Return a [`DataFrame`][pixeltable.DataFrame] for this table."""
+        """ Select columns or expressions from this table."""
```
mkornacker: Probably best to refer to DataFrame.select() via a link.

pixeltable/dataframe.py
```diff
@@ -728,17 +869,37 @@ def limit(self, n: int) -> DataFrame:
         )
 
     def update(self, value_spec: dict[str, Any], cascade: bool = True) -> UpdateStatus:
+        """ Update rows in the underlying table of the DataFrame.
+
+        Update rows in the table with the specified value_spec.
+        The update operation is only allowed for DataFrames on base tables.
```
mkornacker: That's not true, you can update view columns.

pixeltable/dataframe.py
```diff
@@ -728,17 +869,37 @@ def limit(self, n: int) -> DataFrame:
...
+
+        Args:
+            value_spec: a dict of column names to update and the new value to update it to.
+            cascade: if True, also update all computed columns that transitively depend
+                    on the updated columns, including within views. Default is True.
+
+        Returns:
+            UpdateStatus: the status of the update operation.
```
mkornacker, line 898: Missing examples

pixeltable/dataframe.py
```diff
@@ -728,17 +869,37 @@ def limit(self, n: int) -> DataFrame:
...
         self._validate_mutable('update')
         return self._first_tbl.tbl_version.update(value_spec, where=self.where_clause, cascade=cascade)
 
     def delete(self) -> UpdateStatus:
+        """ Delete rows form the underlying table of the DataFrame.
+
+        The delete operation is only allowed for DataFrames on base tables.
+
+        Returns:
+            UpdateStatus: the status of the delete operation.
```
mkornacker, line 922: Missing examples

pixeltable/dataframe.py
anushas: Done. Removed.

pixeltable/dataframe.py
anushas: Done. Added.

pixeltable/catalog/table.py
anushas: Done

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/394)* status: 0 of 2 files reviewed, 16 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

pixeltable/dataframe.py
anushas: Done. Added.

aaron-siegel (approved):

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/394)* status: 0 of 2 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

mkornacker (approved):

PR #395, opened 2024-11-28
PXT-354: Bug: Indexed column for a view may not load correctly from persistence

anushas:
Bug RCA: 
    When an embedding is created on a column, the `indexed_col_id` field of the
    index metadata tracks the column on which this embedding was created.
    For a view that can be ambiguous: it may be id of a column in the view itself,
    or id of a column of any base table(s) that the view is derived from.

Fix:
    Persisted the indexed column's table id in the index metadata to resolve the
    ambiguity. Fixed the code in `TableVerison._init_schema()→_init_idxs()`  to
    lookup the right Column based on both column id and table id combination,
    so that when it is loaded back, it can map to the correct column for a view.
    Since this changed the persistent metadata, bumped up the pixeltable metadata 
    version and added converters to upgrade existing persistence to include the
    new information (indexed column's table id) in the index metadata, if it is missing. 
    Note: Since we cannot know the correct value of the new field for past views, 
    upgrade assigns it the table id of the view on which the index is created. 
    Thus, we can still run into the bug.

Testing:
    New test_embedding_index_reload_view that reproduced the bug for
    a fresh database works correctly with this commit. 
    Existing test_db_migration. that would fail to load an older database 
    dump due to the missing new field in the index metadata, passes with
    this commit, it passes.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @anushas)

___
*[`tests/test_index.py` line 429 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDNE07ph9LY3QbGez:-ODJDNE07ph9LY3QbGf-:b-hjn7ob) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L429)):*
> ```Python
>         self.run_btree_test(data, pxt.Timestamp)
> 
>     def test_embedding_index_reload_table(self, reset_db, reload_tester: ReloadTester) -> None:
> ```

Do we really need this as a separate test? Eventually (hopefully soon) we'll upgrade test_embedding_basic with reload_tester capabilities. (i.e., the goal isn't to view "reload tests" as a separate category of test, but to incorporate reloading as a routine part of our test design).
___
*[`tests/test_index.py` line 439 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDeac9hVmQ_Rs81B7:-ODJDeac9hVmQ_Rs81B8:b-quxs22) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L439)):*
> ```Python
>         _ = reload_tester.run_reload_test(df)
> 
>     def test_embedding_computedcol_reload_table(self, reset_db, reload_tester: ReloadTester) -> None:
> ```

likewise, this seems redundant.
___
*[`tests/test_index.py` line 449 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDkGt0jLajDVprU3b:-ODJDkGt0jLajDVprU3c:bganpe0) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L449)):*
> ```Python
>         _ = reload_tester.run_reload_test(df)
> 
>     def test_embedding_index_reload_view(self, reset_db, reload_tester: ReloadTester) -> None:
> ```

maybe call this `test_embedding_index_on_view` (again, reloading isn't something we need to call out specially)
___
*[`tests/test_index.py` line 476 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDqiSAXPyG5RWzI4r:-ODJDqiSAXPyG5RWzI4s:b-9o8nk2) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L476)):*
> ```Python
> 
> 
>     def test_embedding_computedcol_reload_view(self, reset_db, reload_tester: ReloadTester) -> None:
> ```

this test seems redundant (computed columns on views are tested elsewhere) and doesn't really have anything to do with embedding functionality.

aaron-siegel (changes requested):
We should also add an embedding index on a view to the DB dump tool. (look in `create_test_db_dump.py` and look for the call to `t.add_embedding_index`. In addition to adding an index on a column in `t`, we should add one on a column of a base table of `t` as well.)

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395#-:-ODJFYiS2NSuZ181Gjal:b-1f5j3a)* status: 0 of 12 files reviewed, 5 unresolved discussions (waiting on @anushas)

___
*[`tests/test_index.py` line 458 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJGD3B2maHmvj2MkMC:-ODJGD3B2maHmvj2MkMD:b-ooem8c) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L458)):*
> ```Python
> 
>         # ... restart python session ...
>         _ = reload_tester.run_reload_test(df)
> ```

This only needs to run once, at the end, correct? @mkornacker to confirm.

We also shouldn't need the intermediate queries. Adding the 3 indices and then doing one round of query-reload should be sufficient

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395)* status: 0 of 12 files reviewed, 11 unresolved discussions (waiting on @aaron-siegel and @anushas)

___
*[`tests/test_index.py` line 429 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDNE07ph9LY3QbGez:-ODNCae09X51BFpcuGcl:b-x7wogy) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L429)):*

Do we really need this as a separate test? Eventually (hopefully soon) we'll upgrade test_embedding_basic with reload_tester capabilities. (i.e., the goal isn't to view "reload tests" as a separate category of test, but to incorporate reloading as a routine part of our test design).
</blockquote></details>

We definitely want to use the reload_tester fixture and have that be part of any test that does metadata changes and runs queries.

I agree that all of these test cases should be part of test_embedding_basic.
___
*[`tests/test_index.py` line 458 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJGD3B2maHmvj2MkMC:-ODND-7U5P8jQLwq3bsW:bncnyqc) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L458)):*

This only needs to run once, at the end, correct? @mkornacker to confirm.

We also shouldn't need the intermediate queries. Adding the 3 indices and then doing one round of query-reload should be sufficient
</blockquote></details>

Confirmed.
___
*[`pixeltable/metadata/schema.py` line 116 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNDSdJ8RpOAxDGa5pw:-ODNDSdJ8RpOAxDGa5px:blzke38) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/schema.py#L116)):*
> ```Python
>     name: str
>     indexed_col_id: int  # column being indexed
>     indexed_col_tbl_id: str  # table containing the indexed column
> ```

What is in this string?

Also, move this above col_id. We usually deal with the qualified column id as (tbl id, col id). See MediaStore.prepare_media_path().
___
*[`tests/test_index.py` line 436 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNCnnH9W6UFLs8ZKNl:-ODNCnnH9W6UFLs8ZKNm:b-5klrmb) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L436)):*
> ```Python
>         _ = reload_tester.run_query(df)
> 
>         # ... restart python session ...
> ```

No need to add this explanation, that's already understood.
___
*[`pixeltable/metadata/notes.py` line 5 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNDXe92Gl2RY0R_XdG:-ODNDXe92Gl2RY0R_XdH:b-cfpoj7) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/notes.py#L5)):*
> ```Python
> # the unit tests when new versions are added.
> VERSION_NOTES = {
>     24: 'Added indexed column table id to the TableMd/IndexMd',
> ```

Mention IndexMd.indexed_col_tbl_id directly
___
*[`pixeltable/catalog/table_version.py` line 176 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNF1cGEq_--L4eIiyb:-ODNF1cGEq_--L4eIiyc:boxrfyy) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/catalog/table_version.py#L176)):*
> ```Python
>         return hash(self.id)
> 
>     def _get_column(self, col_id: int, tbl_id: UUID) -> Column:
> ```

Reorder parameters.
___
*[`pixeltable/metadata/converters/util.py` line 15 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNFsWi7SvEjLNBsfXO:-ODNFsWi7SvEjLNBsfXP:b-k3o0h3) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/converters/util.py#L15)):*
> ```Python
> def convert_table_md(
>     engine: sql.engine.Engine,
>     table_md_updater: Optional[Callable[[dict, UUID], None]] = None,
> ```

Requires explanation what this uuid is.
___
*[`pixeltable/metadata/converters/convert_23.py` line 23 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNG4BX-_5gerBGvU0E:-ODNG4BX-_5gerBGvU0F:b3kre) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/converters/convert_23.py#L23)):*
> ```Python
> 
>     Args:
>         table_md (dict): copt if the original table metadata. this gets updated in place.
> ```

?

anushas (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395#-:-ODQ04EB5J3CJGfbWBZX:b-896fix)* status: 0 of 12 files reviewed, 11 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/catalog/table_version.py` line 176 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNF1cGEq_--L4eIiyb:-ODO2UFl1-5sWa94JNVg:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/catalog/table_version.py#L176)):*

Reorder parameters.
</blockquote></details>

Done.
___
*[`pixeltable/metadata/notes.py` line 5 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNDXe92Gl2RY0R_XdG:-ODOReo183XF_pzvKTgW:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/notes.py#L5)):*

Mention IndexMd.indexed_col_tbl_id directly
</blockquote></details>

Done.
___
*[`pixeltable/metadata/schema.py` line 116 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNDSdJ8RpOAxDGa5pw:-ODOSLW85RI0XSCDVCNR:b-228o8j) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/schema.py#L116)):*

What is in this string?

Also, move this above col_id. We usually deal with the qualified column id as (tbl id, col id). See MediaStore.prepare_media_path().
</blockquote></details>

Moved it. I am saving the uuid of the table in the string. Do you want to combine them into one string? It would need some parsing logic. I prefer them in separate field (may be just a bias from db/c world).
___
*[`pixeltable/metadata/converters/convert_23.py` line 23 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNG4BX-_5gerBGvU0E:-ODOTynL5ZL2qJJ1m3No:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/converters/convert_23.py#L23)):*

?
</blockquote></details>

Done.
___
*[`pixeltable/metadata/converters/util.py` line 15 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNFsWi7SvEjLNBsfXO:-ODOUTsz1aRvZffW5iU0:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/converters/util.py#L15)):*

Requires explanation what this uuid is.
</blockquote></details>

Done.
___
*[`tests/test_index.py` line 429 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDNE07ph9LY3QbGez:-ODPpS9CAhPTnm9yPhe3:b-inylks) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L429)):*

We definitely want to use the reload_tester fixture and have that be part of any test that does metadata changes and runs queries.

I agree that all of these test cases should be part of test_embedding_basic.
</blockquote></details>

Done - folded the new testcases into test_embedding_basic.
FWIW, I added the different flavors during the initial repro experiment to understand the issue and its scope. The bug happens to be about reload, and so named the testcase this way. I understand reload_tester is a tool for any testcase that wants to verify persistence. Also, multiple small testcases, one for each flavor, instead of a big testcase with all flavors, makes it convenient to run the relevant ones selectively. 
___
*[`tests/test_index.py` line 436 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNCnnH9W6UFLs8ZKNl:-ODPpOoNE5a3yZbYNPWh:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L436)):*

No need to add this explanation, that's already understood.
</blockquote></details>

Done.
___
*[`tests/test_index.py` line 439 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDeac9hVmQ_Rs81B7:-ODPsuSY1Xdz2RshGZcw:bz20chq) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L439)):*

likewise, this seems redundant.
</blockquote></details>

I added this to during repro experiment because I thought even a computed column in a view that references the base table column may suffer from the same bug. But it does not (notes in [notion doc](https://www.notion.so/pixeltable/Bug-Indexed-column-for-a-view-may-not-load-correctly-from-persistence-14a54609e49080bda8c2cc7f96a8b1c3?pvs=4)). I kept it in the end as it verifies reload works. I have moved them to test_table.py
___
*[`tests/test_index.py` line 449 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDkGt0jLajDVprU3b:-ODPt0fV8yOmV9XnWcx1:b-gbveog) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L449)):*

maybe call this `test_embedding_index_on_view` (again, reloading isn't something we need to call out specially)
</blockquote></details>

Folded into test_embedding_basic. 
___
*[`tests/test_index.py` line 458 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJGD3B2maHmvj2MkMC:-ODPtaib3-KJHm2TwADi:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L458)):*

Confirmed.
</blockquote></details>

Done.
___
*[`tests/test_index.py` line 476 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDqiSAXPyG5RWzI4r:-ODPvXGN13azNTGQzhNp:bz20chq) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L476)):*

this test seems redundant (computed columns on views are tested elsewhere) and doesn't really have anything to do with embedding functionality.
</blockquote></details>

I added this to during repro experiment because I thought even a computed column in a view that references the base table column may suffer from the same bug. But it does not (notes in [notion doc](https://www.notion.so/pixeltable/Bug-Indexed-column-for-a-view-may-not-load-correctly-from-persistence-14a54609e49080bda8c2cc7f96a8b1c3?pvs=4)). I kept it in the end as it verifies reload works. I have moved them to test_table.py

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395)* status: 0 of 14 files reviewed, 12 unresolved discussions (waiting on @anushas and @mkornacker)

___
*[`pixeltable/metadata/schema.py` line 116 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNDSdJ8RpOAxDGa5pw:-ODgCNtP93E5cUxOrn1-:b-ih3hkb) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/schema.py#L116)):*

Moved it. I am saving the uuid of the table in the string. Do you want to combine them into one string? It would need some parsing logic. I prefer them in separate field (may be just a bias from db/c world).
</blockquote></details>

I think Marcel was just commenting on the order in which the fields appear in the MD struct. It definitely makes sense to me to keep them as separate fields.
___
*[`pixeltable/tool/create_test_db_dump.py` line 278 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODgCuoK-lR52tyBzX9K:-ODgCuoK-lR52tyBzX9L:bvatlum) ([raw file](https://github.com/pixeltable/pixeltable/blob/c288164683514983920a5953fe87008a25f8c281/pixeltable/tool/create_test_db_dump.py#L278)):*
> ```Python
>         )
> 
>         if col_prefix == 'view':
> ```

it would be better to condition on the structure of the table rather than the string content of the column names. This works, for example:

`if t.get_metadata()['is_view']`

In particular, it's best to add this index to all view instances in the dump for maximum coverage, not just the single one called 'view'. 

Also generate a new test db artifact to make sure this addition gets picked up (the one in the PR is still dated 11/27).

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395)* status: 0 of 14 files reviewed, 12 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/tool/create_test_db_dump.py` line 278 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODgCuoK-lR52tyBzX9K:-ODgcWinA-92kkMjoQ5W:bs7wumx) ([raw file](https://github.com/pixeltable/pixeltable/blob/c288164683514983920a5953fe87008a25f8c281/pixeltable/tool/create_test_db_dump.py#L278)):*

it would be better to condition on the structure of the table rather than the string content of the column names. This works, for example:

`if t.get_metadata()['is_view']`

In particular, it's best to add this index to all view instances in the dump for maximum coverage, not just the single one called 'view'. 

Also generate a new test db artifact to make sure this addition gets picked up (the one in the PR is still dated 11/27).
</blockquote></details>

Done.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395)* status: 0 of 14 files reviewed, 15 unresolved discussions (waiting on @aaron-siegel and @anushas)

___
*[`tests/test_index.py` line 279 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODhLEXECLgdiWE8NkGs:-ODhLEXECLgdiWE8NkGt:b-mxviwr) ([raw file](https://github.com/pixeltable/pixeltable/blob/01dc699ec6a3c94efafbb1f72ebb0ddb0452c9f7/tests/test_index.py#L279)):*
> ```Python
>         assert 'does not exist' in str(exc_info.value).lower()
> 
>         df = img_t.select()
> ```

Also add a similarity query to test the path with the index.
___
*[`tests/test_index.py` line 281 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODhJbVs8wsh2sah88sq:-ODhJbVs8wsh2sah88sr:b-60cpv6) ([raw file](https://github.com/pixeltable/pixeltable/blob/01dc699ec6a3c94efafbb1f72ebb0ddb0452c9f7/tests/test_index.py#L281)):*
> ```Python
>         df = img_t.select()
>         _ = reload_tester.run_query(df)
>         _ = reload_tester.run_reload_test(clear=True)
> ```

This returns None. Also, this should only be called once at the end
___
*[`tests/test_table.py` line 1682 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODhLkYVBat5AgsZ96-9:-ODhLkYVBat5AgsZ96-A:bmpgrn3) ([raw file](https://github.com/pixeltable/pixeltable/blob/01dc699ec6a3c94efafbb1f72ebb0ddb0452c9f7/tests/test_table.py#L1682)):*
> ```Python
>         df = t.select()
>         _ = reload_tester.run_query(df)
>         _ = reload_tester.run_reload_test(df)
> ```

Call this just once at the end

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395)* status: 0 of 15 files reviewed, 16 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/metadata/schema.py` line 116 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODNDSdJ8RpOAxDGa5pw:-ODmBDmD0MwFfcPbsfQT:b-shu5z1) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/pixeltable/metadata/schema.py#L116)):*

I think Marcel was just commenting on the order in which the fields appear in the MD struct. It definitely makes sense to me to keep them as separate fields.
</blockquote></details>

Ok, good then.
___
*[`tests/test_index.py` line 429 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDNE07ph9LY3QbGez:-ODmB38m3KCtR8B5-zNh:bwygxsx) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L429)):*

To be clear @mkornacker and @anushas , I wasn't proposing we put the new tests in test_embedding_basic! I was simply advocating against creating new tests are redundant with things already in test_embedding_basic. Having too many tests that do similar things makes our testing code harder to navigate.

I'm 100% in favor of breaking things down into smaller, more specific tests.
</blockquote></details>

Noted.
___
*[`tests/test_index.py` line 476 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODJDqiSAXPyG5RWzI4r:-ODmB6J08EuHDUdzRYfJ:bwygxsx) ([raw file](https://github.com/pixeltable/pixeltable/blob/0b3b0a6aad8aa44cf4601c09137cf265d8bb185c/tests/test_index.py#L476)):*

Ok. It's important to verify that reloading works, but I think it'd be better to augment the existing test clauses with reloading, rather than adding new clauses for it (same reasoning as above).
</blockquote></details>

Noted.
___
*[`tests/test_index.py` line 279 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODhLEXECLgdiWE8NkGs:-ODmAsk56cudEAFKgdBy:b-80bwnz) ([raw file](https://github.com/pixeltable/pixeltable/blob/01dc699ec6a3c94efafbb1f72ebb0ddb0452c9f7/tests/test_index.py#L279)):*

Also add a similarity query to test the path with the index.
</blockquote></details>

Added it. Found a bug. PTAL at the comment.
___
*[`tests/test_index.py` line 281 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODhJbVs8wsh2sah88sq:-ODmAs0z528MNHw6jKji:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/01dc699ec6a3c94efafbb1f72ebb0ddb0452c9f7/tests/test_index.py#L281)):*

This returns None. Also, this should only be called once at the end
</blockquote></details>

Done.
___
*[`tests/test_index.py` line 283 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODi6FP62_3xTU_BmRtf:-ODmB-cO9OJugZUu9RcL:bwygxsx) ([raw file](https://github.com/pixeltable/pixeltable/blob/01dc699ec6a3c94efafbb1f72ebb0ddb0452c9f7/tests/test_index.py#L283)):*

FWIW I thought it was better to have this one broken out as a separate test (see my other comment on this below)
</blockquote></details>

Noted.
___
*[`tests/test_table.py` line 1676 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODi3n2E9LA6nCrBxTwZ:-ODmU9FBDsOS1dVs8t0w:bufl8zz) ([raw file](https://github.com/pixeltable/pixeltable/blob/01dc699ec6a3c94efafbb1f72ebb0ddb0452c9f7/tests/test_table.py#L1676)):*

i still don't understand what this test is doing that isn't covered by other tests.

if the goal is to test the `sentence_transformer` UDF, then `test_sentence_transformer` checks for the exact same thing. (If the only difference is the presence of reload tests, the way to fix that is by putting reload tests in `test_sentence_transformer`, not by introducing a separate test for it.)

if the goal is to test a `FunctionCall` on a view that references a column of the base table (which seems like a good test to have in general), that can go in `test_add_computed_column`, and we might as well use a lighter weight function than `sentence_transformer`, one that doesn't have external dependencies.
</blockquote></details>

I added the reload to test_sentence_transformer and the testcase on view to the test_add_computed_column and removed this new function. Discovered an existing bug that needs to be debugged and fixed.
___
*[`tests/test_table.py` line 1682 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/395#-ODhLkYVBat5AgsZ96-9:-ODmB9okAoW7hMqGe3l3:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/01dc699ec6a3c94efafbb1f72ebb0ddb0452c9f7/tests/test_table.py#L1682)):*

Call this just once at the end
</blockquote></details>

Done.

anushas:
The CI failure looks unrelated to my changes. 
-

> ImportError while loading conftest '/Users/runner/work/pixeltable/pixeltable/tests/conftest.py'.
tests/conftest.py:19: in <module>
    from .utils import (
tests/utils.py:540: in <module>
    e5_embed = sentence_transformer.using(model_id='intfloat/e5-large-v2')
pixeltable/func/function.py:130: in using
    call = exprs.FunctionCall(self, bindings)
pixeltable/exprs/function_call.py:53: in __init__
    return_type = fn.call_return_type(bound_args)
pixeltable/func/function.py:94: in call_return_type
    return self._conditional_return_type(**kw_args)
pixeltable/functions/huggingface.py:69: in _
    model = _lookup_model(model_id, SentenceTransformer)
pixeltable/functions/huggingface.py:513: in _lookup_model
    model = create(model_id)
../../../miniconda3/envs/pxt/lib/python3.9/site-packages/sentence_transformers/SentenceTransformer.py:194: in __init__
    modules = self._load_sbert_model(
../../../miniconda3/envs/pxt/lib/python3.9/site-packages/sentence_transformers/SentenceTransformer.py:1073: in _load_sbert_model
    module = module_class.load(module_path)
../../../miniconda3/envs/pxt/lib/python3.9/site-packages/sentence_transformers/models/Pooling.py:195: in load
    with open(os.path.join(input_path, "config.json")) as fIn:
E   FileNotFoundError: [Errno 2] No such file or directory: '/Users/runner/.cache/huggingface/hub/models--intfloat--e5-large-v2/snapshots/b322e09026e4ea05f42beadf4d661fb4e101d311/1_Pooling/config.json'

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/395)* status: 0 of 15 files reviewed, 16 unresolved discussions (waiting on @mkornacker)

mkornacker (approved):

PR #396, opened 2024-11-28
fix: use != instead of 'is not' for string comparison

jacobweiss2305:
The get_txt function was using 'is not' to compare strings, which checks for object identity rather than value equality. This could lead to unreliable behavior due to Python's string interning. Changed to use != operator for proper string value comparison.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/396)

anushas (commented):
Thanks. Would be nice opportunity to add a testcase for an empty txt doc.

aaron-siegel (approved):

PR #397, opened 2024-11-28
Enhance error handling in Table.add_index method

jacobweiss2305:
This commit introduces a try-except block in the Table.add_index method to gracefully handle exceptions when adding an index. If an error occurs, the index is dropped if it exists, and the exception is re-raised for further handling. This change improves the robustness of the indexing process and ensures that eviction warnings are emitted only when the index addition is successful.

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -842,9 +842,14 @@ class Table(SchemaObject):
 
         # create the EmbeddingIndex instance to verify args
         idx = EmbeddingIndex(col, metric=metric, string_embed=string_embed, image_embed=image_embed)
-        status = self._tbl_version.add_index(col, idx_name=idx_name, idx=idx)
-        # TODO: how to deal with exceptions here? drop the index and raise?
-        FileCache.get().emit_eviction_warnings()
+        try:
+            status = self._tbl_version.add_index(col, idx_name=idx_name, idx=idx)
+            FileCache.get().emit_eviction_warnings()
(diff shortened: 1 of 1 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/397)

pixeltable/catalog/table.py
```diff
@@ -842,9 +842,14 @@ def add_embedding_index(
...
-        status = self._tbl_version.add_index(col, idx_name=idx_name, idx=idx)
-        # TODO: how to deal with exceptions here? drop the index and raise?
-        FileCache.get().emit_eviction_warnings()
+        try:
+            status = self._tbl_version.add_index(col, idx_name=idx_name, idx=idx)
+            FileCache.get().emit_eviction_warnings()
+        except Exception as e:
+            # If anything goes wrong, drop the index and re-raise the exception
+            if idx_name is not None and idx_name in self._tbl_version.idxs_by_name:
+                self._drop_index(col=col, idx_name=idx_name, _idx_class=EmbeddingIndex)
```
mkornacker, line 851: This is the wrong place to handle execution errors - Table is the user-facing wrapper, the actual operations are implemented in TableVersion.

In this case, TableVersion._add_index() needs to be handling that error.

Regarding test cases: I'm not sure how to manufacture a condition where index creation fails. I would say that's out of scope.

mkornacker (changes requested):

PR #398, opened 2024-11-28
Enhance HTML escaping in Formatter class

jacobweiss2305:
Updated the __escape method to improve handling of dollar signs. The new implementation uses a regular expression to replace unescaped dollar signs with escaped ones, while preserving already escaped dollar signs. This change ensures better compatibility with MathJax formatting.

Testing: pytest tests/test_formatter.py

### DIFF
--- a/pixeltable/utils/formatter.py
+++ b/pixeltable/utils/formatter.py
@@ -1,6 +1,7 @@
 import base64
 import html
 import io
+import re
 import json
 import logging
 import mimetypes
@@ -75,9 +76,10 @@ class Formatter:
 
     @classmethod
     def __escape(cls, val: str) -> str:
-        # HTML-escape the specified string, then escape $ signs to suppress MathJax formatting
-        # TODO(aaron-siegel): The '$' escaping isn't perfect; it will fail on '$' that are already escaped
-        return html.escape(val).replace('$', r'\$')
+        # First HTML-escape the string
+        escaped = html.escape(val)
+        # Then handle $ signs: replace unescaped $ with \$ while preserving already escaped ones
+        return re.sub(r'(?<!\\)\$', r'\\$', escaped)
 
     @classmethod
     def format_float(cls, val: float) -> str:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/398)

aaron-siegel (changes requested):
This still isn't correct in all cases unfortunately. For example, if the string is `\\$` (i.e., the length-3 string with two backslashes, which in Python would be input by `'\\\\$'`), then the `$` is unescaped, so we DO want to escape it. But (unless I've missed something) this regular expression will leave it unescaped. I believe the correct behavior would be:

`$` -> `\$`
`\$` -> `\$`
`\\$` -> `\\\$`

etc. It's probably still an improvement, since we're more likely to encounter an escaped `$` than an escaped `\` followed by a `$`, but I wonder if there's a way to get it working for all cases.

Either way it'd be good to update our tests to validate the new behavior. There's a "Test HTML and MathJax escaping" section in `test_formatter.py` that you can work off of (maybe just incorporate an escaped `$` into the `test_string`).

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/398#-:-OD8HV3PD8hc6Q3SH-6g:bjp7i4n)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jacobweiss2305)

___
*[`pixeltable/utils/formatter.py` line 81 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/398#-OD8Jru1ERQhdRD3dt4A:-OD8Jru1ERQhdRD3dt4B:bi9lsyg) ([raw file](https://github.com/pixeltable/pixeltable/blob/3925b91417514b12e74eac6ddb49e4615a939c6a/pixeltable/utils/formatter.py#L81)):*
> ```Python
>         # First HTML-escape the string
>         escaped = html.escape(val)
>         # Then handle $ signs: replace unescaped $ with \$ while preserving already escaped ones
> ```

It'd be good to preserve the existing comment re: MathJax so that we keep a record of why it's necessary to do this escaping.

Maybe say "Then escape $ signs in order to suppress MathJax formatting: replace unescaped $ with ..."

PR #401, opened 2024-12-05
PXT-179 Add a if_exists parameter to create_* APIs

anushas:
This commit adds a new parameter - if_exists - for the create_dir, create_table, create_view 
and create_snapshot APIs which can be used to handle scenarios  where a dir/table/view/snapshot 
may already exist. Based on this new parameter, user can ask the API to raise it as an error, 
or do nothing and return the existing dir/table/view/snapshot, or replace the existing dir/table/view/snapshot 
with a new one.

Testing:  improved existing testcases and added new testcases for the new parameters.

### DIFF
--- a/pixeltable/catalog/globals.py
+++ b/pixeltable/catalog/globals.py
@@ -52,23 +52,23 @@ class MediaValidation(enum.Enum):
             raise excs.Error(f'{error_prefix} must be one of: [{val_strs}]')
 
 class IfExistsParam(enum.Enum):
+    ERROR = 0
+    IGNORE = 1
+    REPLACE = 2
+    REPLACE_FORCE = 3
+    '''
     ERROR = 'error'
     IGNORE = 'ignore'
     REPLACE = 'replace'
     REPLACE_FORCE = 'replace_force'
-
+'''
     @classmethod
     def validated(cls, param_val: str, param_name: str) -> IfExistsParam:
         try:
             return cls[param_val.upper()]
         except KeyError:
-            raise excs.Error(f'{param_name} must be one of: {[e.value for e in cls]}')
-
-class SchemaObjectType(enum.Enum):
-    DIR = 'directory'
-    TABLE = 'table'
-    VIEW = 'view'
-    SNAPSHOT = 'snapshot'
+            val_strs = ', '.join(f'{s.lower()!r}' for s in cls.__members__.keys())
+            raise excs.Error(f'{param_name} must be one of: [{val_strs}]')
 
 def is_valid_identifier(name: str) -> bool:
     return name.isidentifier() and not name.startswith('_')
--- a/pixeltable/catalog/globals.py
+++ b/pixeltable/catalog/globals.py
@@ -64,7 +64,8 @@ class IfExistsParam(enum.Enum):
         except KeyError:
             raise excs.Error(f'{param_name} must be one of: {[e.value for e in cls]}')
 
-class CreateTableType(enum.Enum):
+class CreateType(enum.Enum):
+    DIR = 'directory'
     TABLE = 'table'
     VIEW = 'view'
     SNAPSHOT = 'snapshot'
--- a/pixeltable/catalog/globals.py
+++ b/pixeltable/catalog/globals.py
@@ -51,6 +51,23 @@ class MediaValidation(enum.Enum):
             val_strs = ', '.join(f'{s.lower()!r}' for s in cls.__members__.keys())
             raise excs.Error(f'{error_prefix} must be one of: [{val_strs}]')
 
+class IfExistsParam(enum.Enum):
+    ERROR = 'error'
+    IGNORE = 'ignore'
+    REPLACE = 'replace'
+    REPLACE_FORCE = 'replace_force'
+
+    @classmethod
+    def validated(cls, param_val: str, param_name: str) -> 'IfExistsParam':
+        try:
+            return cls[param_val.upper()]
+        except KeyError:
+            raise excs.Error(f'{param_name} must be one of: {[e.value for e in cls]}')
+
+class CreateTableType(enum.Enum):
+    TABLE = 'table'
+    VIEW = 'view'
+    SNAPSHOT = 'snapshot'
 
 def is_valid_identifier(name: str) -> bool:
     return name.isidentifier() and not name.startswith('_')
--- a/pixeltable/catalog/globals.py
+++ b/pixeltable/catalog/globals.py
@@ -64,7 +64,7 @@ class IfExistsParam(enum.Enum):
         except KeyError:
             raise excs.Error(f'{param_name} must be one of: {[e.value for e in cls]}')
 
-class CreateType(enum.Enum):
+class SchemaObjectType(enum.Enum):
     DIR = 'directory'
     TABLE = 'table'
     VIEW = 'view'
--- a/pixeltable/catalog/globals.py
+++ b/pixeltable/catalog/globals.py
@@ -58,7 +58,7 @@ class IfExistsParam(enum.Enum):
     REPLACE_FORCE = 'replace_force'
 
     @classmethod
-    def validated(cls, param_val: str, param_name: str) -> 'IfExistsParam':
+    def validated(cls, param_val: str, param_name: str) -> IfExistsParam:
         try:
             return cls[param_val.upper()]
         except KeyError:
--- a/pixeltable/catalog/globals.py
+++ b/pixeltable/catalog/globals.py
@@ -56,12 +56,7 @@ class IfExistsParam(enum.Enum):
     IGNORE = 1
     REPLACE = 2
     REPLACE_FORCE = 3
-    '''
-    ERROR = 'error'
-    IGNORE = 'ignore'
-    REPLACE = 'replace'
-    REPLACE_FORCE = 'replace_force'
-'''
+
     @classmethod
     def validated(cls, param_val: str, param_name: str) -> IfExistsParam:
         try:
--- a/pixeltable/globals.py
+++ b/pixeltable/globals.py
@@ -26,7 +26,7 @@ def init() -> None:
 
 def _get_or_drop_existing_path(
     path_str: str,
-    caller_context: catalog.CreateType,
+    caller_context: catalog.SchemaObjectType,
     if_exists: catalog.IfExistsParam
 ) -> Optional[catalog.SchemaObject]:
     """Handle existing dir, table, view, or snapshot, during create call as per user directive.
@@ -51,19 +51,19 @@ def _get_or_drop_existing_path(
     """
     cat = Catalog.get()
     path = catalog.Path(path_str)
-    assert cat.paths.check_if_exists(path) is not None
+    assert cat.paths.get_object(path) is not None
 
     if if_exists == catalog.IfExistsParam.ERROR:
         raise excs.Error(f'Path `{path_str}` already exists.')
 
     existing_path = cat.paths[path]
-    expected_obj_type = catalog.Dir if caller_context == catalog.CreateType.DIR else \
-        catalog.Table if caller_context == catalog.CreateType.TABLE else catalog.View
+    expected_obj_type = catalog.Dir if caller_context == catalog.SchemaObjectType.DIR else \
+        catalog.Table if caller_context == catalog.SchemaObjectType.TABLE else catalog.View
     obj_type_str = caller_context.value.capitalize()
     existing_path_is_snapshot = expected_obj_type == catalog.View and existing_path.get_metadata()['is_snapshot']
     # Check if the existing path is of expected type.
     if (not isinstance(existing_path, expected_obj_type)
-        or (caller_context == catalog.CreateType.SNAPSHOT and not existing_path_is_snapshot)):
+        or (caller_context == catalog.SchemaObjectType.SNAPSHOT and not existing_path_is_snapshot)):
             raise excs.Error(f'Path `{path_str}` already exists but is not a {obj_type_str}. Cannot {if_exists.value} it.')
 
     # if_exists='ignore' return the handle to the existing object.
@@ -85,7 +85,7 @@ def _get_or_drop_existing_path(
             drop_dir(path_str, force=True, ignore_errors=False)
         else:
             drop_table(path_str, force=True, ignore_errors=False)
-        assert cat.paths.check_if_exists(path) is None
+        assert cat.paths.get_object(path) is None
 
     return None
 
@@ -151,10 +151,10 @@ def create_table(
     path = catalog.Path(path_str)
     cat = Catalog.get()
 
-    if cat.paths.check_if_exists(path) is not None:
+    if cat.paths.get_object(path) is not None:
         # The table already exists. Handle it as per user directive.
         _if_exists = catalog.IfExistsParam.validated(if_exists, 'if_exists')
-        existing_table = _get_or_drop_existing_path(path_str, catalog.CreateType.TABLE, _if_exists)
+        existing_table = _get_or_drop_existing_path(path_str, catalog.SchemaObjectType.TABLE, _if_exists)
         if existing_table is not None:
             assert isinstance(existing_table, catalog.Table)
             return existing_table
@@ -275,10 +275,10 @@ def create_view(
     path = catalog.Path(path_str)
     cat = Catalog.get()
 
-    if cat.paths.check_if_exists(path) is not None:
+    if cat.paths.get_object(path) is not None:
         # The view already exists. Handle it as per user directive.
         _if_exists = catalog.IfExistsParam.validated(if_exists, 'if_exists')
-        curr_context = catalog.CreateType.SNAPSHOT if is_snapshot else catalog.CreateType.VIEW
+        curr_context = catalog.SchemaObjectType.SNAPSHOT if is_snapshot else catalog.SchemaObjectType.VIEW
         existing_path = _get_or_drop_existing_path(path_str, curr_context, _if_exists)
         if existing_path is not None:
             assert isinstance(existing_path, catalog.View)
@@ -557,10 +557,10 @@ def create_dir(path_str: str, if_exists: Literal['error', 'ignore', 'replace', '
     path = catalog.Path(path_str)
     cat = Catalog.get()
 
-    if cat.paths.check_if_exists(path):
+    if cat.paths.get_object(path):
         # The directory already exists. Handle it as per user directive.
         _if_exists = catalog.IfExistsParam.validated(if_exists, 'if_exists')
-        existing_path = _get_or_drop_existing_path(path_str, catalog.CreateType.DIR, _if_exists)
+        existing_path = _get_or_drop_existing_path(path_str, catalog.SchemaObjectType.DIR, _if_exists)
         if existing_path is not None:
             assert isinstance(existing_path, catalog.Dir)
             return existing_path
--- a/pixeltable/globals.py
+++ b/pixeltable/globals.py
@@ -106,11 +106,8 @@ def create_table(
         if if_exists == IfExistsParam.ERROR:
             raise excs.Error(f'Path `{path_str}` already exists.')
         existing_table = cat.paths[path]
-        is_table = isinstance(existing_table, catalog.Table)
-        if not is_table:
+        if not isinstance(existing_table, catalog.Table):
             raise excs.Error(f'Path `{path_str}` already exists but is not a Table. Cannot {if_exists} it.')
-
-        assert is_table
 
         if if_exists == IfExistsParam.IGNORE:
             return existing_table
@@ -119,7 +116,7 @@ def create_table(
         if if_exists == IfExistsParam.REPLACE and has_dependents:
             raise excs.Error(f'Table `{path_str}` already exists and has dependents. Use if_exists="replace_force" to replace it.')
         else:
-            assert if_exists == IfExistsParam.REPLACE_FORCE or (is_table and not has_dependents)
+            assert if_exists == IfExistsParam.REPLACE_FORCE or not has_dependents
             # Drop the existing table so it can be replaced.
             # Any error from drop_table will not be ignored.
             _logger.info(f'Dropping and recreating table `{path_str}`.')
@@ -512,11 +509,9 @@ def create_dir(path_str: str, if_exists: Literal['error', 'ignore', 'replace', '
         if if_exists == IfExistsParam.ERROR:
             raise excs.Error(f'Path `{path_str}` already exists.')
         existing_dir = cat.paths[path]
-        is_dir = isinstance(existing_dir, catalog.Dir)
-        if not is_dir:
+        if not isinstance(existing_dir, catalog.Dir):
             raise excs.Error(f'Path `{path_str}` already exists but is not a Dir. Cannot {if_exists} it.')
 
-        assert is_dir
         if if_exists == IfExistsParam.IGNORE:
             return existing_dir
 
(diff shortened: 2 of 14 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/401)

pixeltable/globals.py
```diff
@@ -33,7 +47,8 @@ def create_table(
     primary_key: Optional[Union[str, list[str]]] = None,
     num_retained_versions: int = 10,
     comment: str = '',
-    media_validation: Literal['on_read', 'on_write'] = 'on_write'
+    media_validation: Literal['on_read', 'on_write'] = 'on_write',
+    if_exists: str = 'error'
```
mkornacker: Follow implementation of media_validation parameter. See MediaValidation.

pixeltable/globals.py
```diff
@@ -66,10 +89,46 @@ def create_table(
 
         >>> tbl1 = pxt.get_table('orig_table')
         ... tbl2 = pxt.create_table('new_table', tbl1.where(tbl1.col1 < 10).select(tbl1.col2))
+
+        Create a table if does not already exist, otherwise get the existing table:
+
+        >>> tbl = pxt.create_table('my_table', schema={'col1': pxt.Int, 'col2': pxt.String}, if_exists='ignore')
+
+        Create a table with an int and a float column, and drop and replace, if it already exists:
```
mkornacker: 'drop and replace' -> 'replace'

pixeltable/globals.py
```diff
@@ -66,10 +89,46 @@ def create_table(
...
     """
+    if_exists = if_exists.lower()
+    _verify_param_if_exists(if_exists)
     path = catalog.Path(path_str)
-    Catalog.get().paths.check_is_valid(path, expected=None)
-    dir = Catalog.get().paths[path.parent]
+    cat = Catalog.get()
+    try:
+        cat.paths.check_is_valid(path, expected=None)
+    except excs.Error as check_error:
```
mkornacker: Error is only meant for user-facing interactions, and we don't want to rely on substrings to indicate status.

Use PathDict.get_schema_obj().

pixeltable/globals.py
```diff
@@ -46,15 +61,23 @@ def create_table(
         num_retained_versions: Number of versions of the table to retain.
         comment: An optional comment; its meaning is user-defined.
         media_validation: Media validation policy for the table.
-
             - `'on_read'`: validate media files at query time
             - `'on_write'`: validate media files during insert/update operations
+        if_exists: Directive regarding how to handle the path already exists.
+            Must be one of the following:
+            - 'error': raise an error
```
mkornacker: Following formatting of media_validation options.

tests/test_dirs.py
```diff
@@ -15,38 +15,49 @@ def test_create(self, reset_db) -> None:
             assert dir._parent._path == '.'.join(name.split('.')[:-1])
 
         # invalid names
-        with pytest.raises(excs.Error):
+        with pytest.raises(excs.Error) as exc:
```
mkornacker: This is uniformly 'as exc_info' (it doesn't return an exception).

tests/test_table.py
```diff
@@ -116,6 +125,92 @@ def test_create(self, reset_db: None) -> None:
...
+        assert_resultset_eq(res_before, res_after)
+        # if_exists='replace' should drop the existing table
+        tbl3 = pxt.create_table('test', schema, if_exists='replace')
+        assert tbl3 != tbl
+        assert tbl3._id != id_before
+        res_after = tbl3.select().collect()
+        assert len(res_after) == 0
+        id_before = tbl3._id
+
+        reload_catalog()
```
mkornacker: reload_tester instead

mkornacker (changes requested):

pixeltable/globals.py
anushas: Done.

pixeltable/globals.py
anushas: Done.

pixeltable/globals.py
anushas: Done.

pixeltable/globals.py
anushas: Sorry did not fully understand. Can you elaborate?

tests/test_dirs.py
anushas: Done

tests/test_table.py
anushas: Done.

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/401)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @mkornacker)

pixeltable/globals.py
aaron-siegel: Relying on the string content of an exception for control flow is a fragile design pattern (the error messages are intended to communicate human-readable information back to the user, not to drive control flow).

Marcel is suggesting to find a more robust way to check whether the table exists. However `PathDict.get_schema_obj()` requires a UUID parameter, not a path. Another idea is to factor the existence check out of `check_is_valid()`, and then selectively check for existence as a separate step (for example, by enhancing resolve_path(), or by introducing a new path_exists() method, or some such).

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/401)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @mkornacker)

pixeltable/globals.py
anushas: I extracted the exists check into a separate function. I think that matches what you had in mind.

anushas:
Latest update is to add the if_exists parameter to the create_view API.

anushas:
Latest update resolved type check errors and added the if_exists parameter to the create_snapshot API and consolidated the common code.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/401)* status: 0 of 12 files reviewed, 10 unresolved discussions (waiting on @anushas and @mkornacker)

___
*[`pixeltable/catalog/globals.py` line 67 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEQaWoc7Ms-KlDJdgh9:-OEQaWoc7Ms-KlDJdghA:b-7kcjg1) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/globals.py#L67)):*
> ```Python
>             raise excs.Error(f'{param_name} must be one of: {[e.value for e in cls]}')
> 
> class CreateType(enum.Enum):
> ```

SchemaObjectType or some such would be a better name (the enum values themselves don't have anything specifically to do with creation)
___
*[`pixeltable/catalog/path_dict.py` line 104 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OESfRm0E-D3qpKnE8hx:-OESfRm0E-D3qpKnE8hy:bwreqi) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/path_dict.py#L104)):*
> ```Python
>         self.dir_contents[to_dir._id][to_path.name] = obj
> 
>     def check_if_exists(self, path: Path) -> SchemaObject:
> ```

this isn't really checking whether something exists, it's looking up a SchemaObject by path. let's change the name to something more appropriate (lookup_schema_obj?)

this also duplicates a bunch of logic with _resolve_path. a less duplicative approach might be: rename _resolve_path to lookup_schema_obj and change its behavior to return None rather than throw an exception if the path is not found. then reimplement _resolve_path to just call lookup_schema_obj and throw an exception if the result is None.
___
*[`pixeltable/catalog/path_dict.py` line 111 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OESf9D17BEXO4XQmC_u:-OESf9D17BEXO4XQmC_v:bn3ojpb) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/path_dict.py#L111)):*
> ```Python
> 
>         Returns:
>             SchemaObject at the path if it exists, None otherwise.
> ```

return type should be `Optional[SchemaObject]` if this is the case
___
*[`pixeltable/catalog/table.py` line 67 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEQcBnd4Sj8oWmDgIYo:-OEQcBnd4Sj8oWmDgIYp:b-38ys5b) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/table.py#L67)):*
> ```Python
> 
>     @property
>     def _has_dependents(self) -> bool:
> ```

This does essentially the same thing as `Table._get_views()`. I'd suggest we don't need a separate `_has_dependents` property since it isn't any more efficient. An alternative would be to rename `Table._get_views()` to `Table._get_dependents()` and add a corresponding `Dir._get_dependents()`. Then in the calling function, instead of `has_dependents = obj._has_dependents`, just do `has_dependents = len(obj._get_dependents()) > 0`. Less code duplication + more flexibility.

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/401)* status: 0 of 12 files reviewed, 10 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/catalog/globals.py` line 67 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEQaWoc7Ms-KlDJdgh9:-OEWAWcOBwiq2LzJte0C:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/globals.py#L67)):*

SchemaObjectType or some such would be a better name (the enum values themselves don't have anything specifically to do with creation)
</blockquote></details>

Done.
___
*[`pixeltable/catalog/path_dict.py` line 104 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OESfRm0E-D3qpKnE8hx:-OEWSDu_E-6tC_Noiv9J:b-3zm4nn) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/path_dict.py#L104)):*

this isn't really checking whether something exists, it's looking up a SchemaObject by path. let's change the name to something more appropriate (lookup_schema_obj?)

this also duplicates a bunch of logic with _resolve_path. a less duplicative approach might be: rename _resolve_path to lookup_schema_obj and change its behavior to return None rather than throw an exception if the path is not found. then reimplement _resolve_path to just call lookup_schema_obj and throw an exception if the result is None.
</blockquote></details>

Done. I named it get_object() (along the lines of get_children()). Thanks!
___
*[`pixeltable/catalog/path_dict.py` line 111 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OESf9D17BEXO4XQmC_u:-OEWDfPJ6EhHR4P0VnJd:b-pean5h) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/path_dict.py#L111)):*

return type should be `Optional[SchemaObject]` if this is the case
</blockquote></details>

Done. Interestingly, typecheck did not catch this.
___
*[`pixeltable/catalog/table.py` line 67 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEQcBnd4Sj8oWmDgIYo:-OEWYWKI76m2YjZHJ-O2:b-w3foti) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/table.py#L67)):*

This does essentially the same thing as `Table._get_views()`. I'd suggest we don't need a separate `_has_dependents` property since it isn't any more efficient. An alternative would be to rename `Table._get_views()` to `Table._get_dependents()` and add a corresponding `Dir._get_dependents()`. Then in the calling function, instead of `has_dependents = obj._has_dependents`, just do `has_dependents = len(obj._get_dependents()) > 0`. Less code duplication + more flexibility.
</blockquote></details>

I have changed it to call _get_views() to remove code. I like the convenience of a boolean abstraction, so left that. Do you feel very strongly about removing that property?

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/401)* status: 0 of 12 files reviewed, 10 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/catalog/table.py` line 67 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEQcBnd4Sj8oWmDgIYo:-OEWZoYq-IoRHpz6u3_S:b-f31x0v) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/table.py#L67)):*

I have changed it to call _get_views() to remove code. I like the convenience of a boolean abstraction, so left that. Do you feel very strongly about removing that property?
</blockquote></details>

By "left that" I mean I continue to use _has_dependents property.

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/401)* status: 0 of 12 files reviewed, 14 unresolved discussions (waiting on @anushas and @mkornacker)

___
*[`pixeltable/catalog/path_dict.py` line 104 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OESfRm0E-D3qpKnE8hx:-OEYLHlT17L8HNP_deaZ:b-w8pvk7) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/path_dict.py#L104)):*

Done. I named it get_object() (along the lines of get_children()). Thanks!
</blockquote></details>

Sounds good. I like get_object().
___
*[`pixeltable/catalog/table.py` line 67 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEQcBnd4Sj8oWmDgIYo:-OEYLZQAExbtBzzTs7yc:be1zi1p) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/table.py#L67)):*

By "left that" I mean I continue to use _has_dependents property.
</blockquote></details>

No I don't feel that strongly about it. The solution you went with is fine. A case could be made for having _get_dependents be the hierarchical (abstract+overridden) function, with _has_dependents defined in SchemaObject (as a concrete property that calls _get_dependents) the way you've done here. The main advantage of this is it would also expose Dir._get_dependents(), which could be useful to have around.
___
*[`pixeltable/globals.py` line 242 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEYJIV0A331rBT76HGi:-OEYJIV0A331rBT76HGj:b4du914) ([raw file](https://github.com/pixeltable/pixeltable/blob/264bed2ec72016141be97edc7d306bebe45c6108/pixeltable/globals.py#L242)):*
> ```Python
>         Error: if the path already exists and `if_exists='error'`,
>             or if the path already exists and is not a view,
>             or other error conditions like invalid path, connection error etc.
> ```

A partial list of error conditions followed by "etc." isn't super helpful here. I'd just say "or an error occurs while attempting to create the view".
___
*[`pixeltable/globals.py` line 350 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEYK76i2Jrj_00VDp7Q:-OEYK76i2Jrj_00VDp7R:b-kb9px3) ([raw file](https://github.com/pixeltable/pixeltable/blob/264bed2ec72016141be97edc7d306bebe45c6108/pixeltable/globals.py#L350)):*
> ```Python
>         Error: if the path already exists and `if_exists='error'`,
>             or if the path already exists and is not a snapshot,
>             or other error conditions like invalid path, connection error etc.
> ```

Same as preceding comment
___
*[`pixeltable/globals.py` line 540 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEYKKti9bJX00N3rWxp:-OEYKKti9bJX00N3rWxq:b-duc46) ([raw file](https://github.com/pixeltable/pixeltable/blob/264bed2ec72016141be97edc7d306bebe45c6108/pixeltable/globals.py#L540)):*
> ```Python
>         Error: If the path already exists and `if_exists='error'`,
>             or if the path already exists and is not a directory,
>             or other error conditions like invalid path, store errors etc.
> ```

same comment
___
*[`pixeltable/catalog/schema_object.py` line 73 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEYMKDR2d3vsIpowgSt:-OEYMKDR2d3vsIpowgSu:b-35m9dv) ([raw file](https://github.com/pixeltable/pixeltable/blob/264bed2ec72016141be97edc7d306bebe45c6108/pixeltable/catalog/schema_object.py#L73)):*
> ```Python
>     def _has_dependents(self) -> bool:
>         """Returns True if this object has dependents (e.g., children, views)"""
>         pass
> ```

Python style: omit `pass` in abstract methods

pixeltable/globals.py
```diff
@@ -20,11 +20,74 @@
...
 _logger = logging.getLogger('pixeltable')
 
-
 def init() -> None:
     """Initializes the Pixeltable environment."""
     _ = Catalog.get()
 
+def _get_or_drop_existing_path(
+    path_str: str,
+    caller_context: catalog.SchemaObjectType,
```
mkornacker: instead: schema_obj_cls: Type[SchemaObject] or maybe even created_cls or something like that

and get rid of SchemaObjectType

pixeltable/catalog/globals.py
```diff
@@ -51,6 +51,24 @@ def validated(cls, name: str, error_prefix: str) -> MediaValidation:
             val_strs = ', '.join(f'{s.lower()!r}' for s in cls.__members__.keys())
             raise excs.Error(f'{error_prefix} must be one of: [{val_strs}]')
 
+class IfExistsParam(enum.Enum):
+    ERROR = 'error'
```
mkornacker: Follow MediaValidation pattern

pixeltable/globals.py
```diff
@@ -20,11 +20,74 @@
...
+    if_exists: catalog.IfExistsParam
+) -> Optional[catalog.SchemaObject]:
+    """Handle existing dir, table, view, or snapshot, during create call as per user directive.
+
+    Args:
+        path_str: An existing and valid path to the dir, table, view, or snapshot.
+        caller_context: Whether the caller of this function is creating a dir, table, view, or snapshot, at the existing path.
+        if_exists: Directive regarding how to handle the existing path.
+            Must be one of the following:
```
mkornacker: Not needed, this is not a user-facing API function. Also, if_exists is not a string parameter.

IfExistsParam is self-explanatory.

pixeltable/globals.py
```diff
@@ -20,11 +20,74 @@
 
 _logger = logging.getLogger('pixeltable')
 
-
 def init() -> None:
     """Initializes the Pixeltable environment."""
     _ = Catalog.get()
 
+def _get_or_drop_existing_path(
```
mkornacker, line 27: If I had to summarize what this is trying to accomplish, I'd probably say something like "handle SchemaObject path collision".

mkornacker (changes requested):

pixeltable/globals.py
anushas: Done. Changed the summary.

pixeltable/globals.py
anushas: Done. FWIW, I introduced SchemaObjectType mainly for Snapshot as it is not a class derived from SchemaObject.

pixeltable/globals.py
anushas: Done.

pixeltable/catalog/globals.py
anushas: Done.

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/401)* status: 0 of 12 files reviewed, 18 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/globals.py` line 242 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEYJIV0A331rBT76HGi:-OEm61qY3o2S0_6kPNHi:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/264bed2ec72016141be97edc7d306bebe45c6108/pixeltable/globals.py#L242)):*

A partial list of error conditions followed by "etc." isn't super helpful here. I'd just say "or an error occurs while attempting to create the view".
</blockquote></details>

Done.
___
*[`pixeltable/globals.py` line 350 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEYK76i2Jrj_00VDp7Q:-OEm62XH8Gx8c3zHhd3R:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/264bed2ec72016141be97edc7d306bebe45c6108/pixeltable/globals.py#L350)):*

Same as preceding comment
</blockquote></details>

Done.
___
*[`pixeltable/globals.py` line 540 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEYKKti9bJX00N3rWxp:-OEm638c2CRpeP1u2gCL:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/264bed2ec72016141be97edc7d306bebe45c6108/pixeltable/globals.py#L540)):*

same comment
</blockquote></details>

Done.
___
*[`pixeltable/catalog/schema_object.py` line 73 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEYMKDR2d3vsIpowgSt:-OEmAYanAY2dQP7tG5YO:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/264bed2ec72016141be97edc7d306bebe45c6108/pixeltable/catalog/schema_object.py#L73)):*

Python style: omit `pass` in abstract methods
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 67 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/401#-OEQcBnd4Sj8oWmDgIYo:-OEmKI0tBWuaTPTPPAIk:b-hlj6l1) ([raw file](https://github.com/pixeltable/pixeltable/blob/85aca92035473c4dee8a54d95e0808056ca1502b/pixeltable/catalog/table.py#L67)):*

No I don't feel that strongly about it. The solution you went with is fine. A case could be made for having _get_dependents be the hierarchical (abstract+overridden) function, with _has_dependents defined in SchemaObject (as a concrete property that calls _get_dependents) the way you've done here. The main advantage of this is it would also expose Dir._get_dependents(), which could be useful to have around.
</blockquote></details>

I think this is a good idea, but Dir dependents are Paths while for other SchemaObject the dependents are Tables. May be we can discuss and do this separately.

pixeltable/catalog/globals.py
```diff
@@ -52,23 +52,23 @@ def validated(cls, name: str, error_prefix: str) -> MediaValidation:
             raise excs.Error(f'{error_prefix} must be one of: [{val_strs}]')
 
 class IfExistsParam(enum.Enum):
+    ERROR = 0
+    IGNORE = 1
+    REPLACE = 2
+    REPLACE_FORCE = 3
+    '''
     ERROR = 'error'
```
mkornacker: This is not a useful comment.

pixeltable/globals.py
```diff
@@ -26,21 +26,17 @@ def init() -> None:
 
 def _get_or_drop_existing_path(
     path_str: str,
-    caller_context: catalog.SchemaObjectType,
+    expected_obj_type: Type[catalog.SchemaObject],
+    expected_snapshot: bool,
     if_exists: catalog.IfExistsParam
 ) -> Optional[catalog.SchemaObject]:
-    """Handle existing dir, table, view, or snapshot, during create call as per user directive.
+    """Handle schema object path collision create call as per user directive.
```
mkornacker: Grammar

mkornacker (approved):

pixeltable/globals.py
anushas: Corrected.

pixeltable/catalog/globals.py
anushas: Yes, was left by mistake. Removed.

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/401)* status: 0 of 12 files reviewed, 20 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

PR #402, opened 2024-12-09
Update Expr.id after a call to substitute()

aaron-siegel:
Expr.substitute() changes the composition of an Expr, but isn't updating the id. As a result, it's possible to end up with two expressions with the same id but different outputs. This causes Pixeltable to give the wrong answer to statements such as

`t.select(f(t.c1), f(t.c2))`

where `f` is an `@expr_udf`. Here `f(t.c1)` and `f(t.c2)` are distinct substitutions of the same variable expression; before this fix, they would have the same id, and hence be equated in the cache. The `select` statement thus would yield two identical columns, both equal to whichever subexpression was mentioned first.

### DIFF
--- a/pixeltable/exprs/expr.py
+++ b/pixeltable/exprs/expr.py
@@ -190,6 +190,7 @@ class Expr(abc.ABC):
                 return new.copy()
         for i in range(len(self.components)):
             self.components[i] = self.components[i].substitute(spec)
+        self.id = self._create_id()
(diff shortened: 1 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/402)

mkornacker (approved):

PR #403, opened 2024-12-09
Allow Function instances, UDFs, and UDAs to have multiple signatures

aaron-siegel:
Upgrades `Function` to allow multiple signatures with differently typed parameters. Instead of a single `Function.signature` instance, we now have a list `Function.signatures`. `f(...)` resolves uniquely to a `FunctionCall` by selecting the earliest signature in declaration order that matches the provided arguments.

A `CallableFunction` may have one Python function for all signatures or multiple Python functions (in any combination); likewise an `AggregateFunction` may have one or many corresponding `Aggregator` classes. `ExprTemplateFunction` is also upgraded to support multiple templates, ensuring correct behavior if `Function.using(...)` is called on a polymorphic instance. (`@expr_udf` still always produces a single-template `ExprTemplateFunction`, although we could easily upgrade this in the future if there's a need for it.)

The `@uda` declaration pattern has been upgraded to infer its type signature from the type hints on the `Aggregator` class (to align with the solution for `@udf`s that we implemented some time ago).

The library aggregators have been upgraded with polymorphic inputs where relevant (count, sum, mean, min, max).

There are two ways to *declare* a polymorphic UDF or UDA. The most straightforward way is via the `@my_udf.overload` decorator:

```python
@pxt.udf
def my_udf(x: int) -> int:
    return x + 1

@my_udf.overload
def _(x: float) -> float:
    return x + 1.0
```

When the Python functions are identical, a more concise syntax can be used; this is particularly useful for UDAs such as `max` and `min` that have many signatures.

```python
T = typing.TypeVar('T')

@pxt.udf(type_substitutions=({T: int}, {T: float}))
def my_udf(x: T) -> T:
    return x + x
```

One signature will be generated for each dictionary in `type_substitutions`, all with the same underlying Python function.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/403)

mkornacker (changes requested):
Some preliminary comments

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/403#-:-ODnHxfN3Pb6DkLymIhC:b-k5xpob)* status: 0 of 29 files reviewed, 3 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/func/aggregate_function.py` line 50 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODn8BosAEb32oSz4mil:-ODn8BosAEb32oSz4mim:bcxesut) ([raw file](https://github.com/pixeltable/pixeltable/blob/60a6977c3c917140b9e042653ac74ee99f9ecbed/pixeltable/func/aggregate_function.py#L50)):*
> ```Python
>         allows_window: bool
>     ) -> None:
>         if type_substitutions is None:
> ```

Normalize as type_substitutions = [None] here?
___
*[`pixeltable/func/expr_template_function.py` line 11 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODnD8O06TLgjWxdsdIf:-ODnD8O06TLgjWxdsdIg:bevpawz) ([raw file](https://github.com/pixeltable/pixeltable/blob/60a6977c3c917140b9e042653ac74ee99f9ecbed/pixeltable/func/expr_template_function.py#L11)):*
> ```Python
> 
> 
> class Template:
> ```

What's the meaning of this class? Isn't this really the monomorphic exprtemplatefunction?
___
*[`pixeltable/func/expr_template_function.py` line 64 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODnHY1w-d7xzlbgdB7w:-ODnHY1w-d7xzlbgdB7x:b-csqwrf) ([raw file](https://github.com/pixeltable/pixeltable/blob/60a6977c3c917140b9e042653ac74ee99f9ecbed/pixeltable/func/expr_template_function.py#L64)):*
> ```Python
>         super().__init__([t.signature for t in templates], self_path=self_path)
> 
>     def instantiate(self, signature_idx: int, args: Sequence[Any], kwargs: dict[str, Any]) -> 'pixeltable.exprs.Expr':
> ```

Why not *args, **kwargs?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/403)* status: 0 of 29 files reviewed, 3 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/func/aggregate_function.py` line 50 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODn8BosAEb32oSz4mil:-ODs3X4A7vvt_1ugJyD8:bhqe0t4) ([raw file](https://github.com/pixeltable/pixeltable/blob/60a6977c3c917140b9e042653ac74ee99f9ecbed/pixeltable/func/aggregate_function.py#L50)):*

Normalize as type_substitutions = [None] here?
</blockquote></details>

Done.

Technically it's inconsistent with the type hint of `type_substitutions` (which is `Optional[Sequence[dict]]`, not `Optional[Sequence[Optional[dict]]]`), but mypy seems content to let it slide.
___
*[`pixeltable/func/expr_template_function.py` line 11 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODnD8O06TLgjWxdsdIf:-ODs3gXv0IkYpbzgcFDj:bx7murt) ([raw file](https://github.com/pixeltable/pixeltable/blob/60a6977c3c917140b9e042653ac74ee99f9ecbed/pixeltable/func/expr_template_function.py#L11)):*

What's the meaning of this class? Isn't this really the monomorphic exprtemplatefunction?
</blockquote></details>

Not exactly.

A `CallableFunction` can have different Python functions for each signature. In `CallableFunction`, this is handled by having a separate `py_fns` list adjacent to the `signatures` list.

In the same way, an `ExprTemplateFunction` can have different `Expr`s for each signature. This is what you get when you call `.using()` on a polymorphic function (or `.overload` an `@expr_udf`, if we decide to allow that in the future). That could similarly be handled with an `exprs` list adjacent to `signatures`, but in this case, there's a whole bunch of precomputed metadata associated to each signature-expr pair. I found it easier to encapsulate that in a class, which is what `Template` is doing.

I added a code comment that hopefully helps clarify this
___
*[`pixeltable/func/expr_template_function.py` line 64 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODnHY1w-d7xzlbgdB7w:-ODs4wA67bHCBlsrCKk9:bur9977) ([raw file](https://github.com/pixeltable/pixeltable/blob/60a6977c3c917140b9e042653ac74ee99f9ecbed/pixeltable/func/expr_template_function.py#L64)):*

Why not *args, **kwargs?
</blockquote></details>

With the additional `signature_idx` parameter, it's safest to take explicit list and dict arguments. If `signature_idx` is going away it's less of an issue, but I'd still argue it's general best practice to avoid varargs in internal APIs, and especially when the typical call pattern is to send along a list and a dict. That is: use varargs only when they actually get used as such in the typical call pattern.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/403)* status: 0 of 32 files reviewed, 12 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/exprs/function_call.py` line 51 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODso2gU-EjluPOX0jVS:-ODso2gVDdnAg3KB6VLp:b-x49e1p) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/exprs/function_call.py#L51)):*
> ```Python
>         fn: func.Function,
>         bound_args: dict[str, Any],
>         return_type: ts.ColumnType,
> ```

Why do we need to pass it in, instead of using call_return_type()?
___
*[`pixeltable/exprs/function_call.py` line 447 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsoCKn5b15ooLSEEc1:-ODsoCKn5b15ooLSEEc2:b-9o2sgu) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/exprs/function_call.py#L447)):*
> ```Python
>         elif self.is_window_fn_call:
>             assert isinstance(self.fn, func.AggregateFunction)
>             agg_cls = self.fn.agg_classes[0]
> ```

I would prefer a property Function.agg_cls(self):
    assert not self.is_polymorphic
    return self.agg_classes[0]
___
*[`pixeltable/exprs/function_call.py` line 488 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsoZmr1ku90mlDyV99:-ODsoZmr1ku90mlDyV9A:b-r9bkks) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/exprs/function_call.py#L488)):*
> ```Python
> 
>         # Try to find a function signature that matches the types of the args and kwargs in the dict. We do it this
>         # way, rather than simply serializing signature_idx, in order to ensure we do the right thing after a
> ```

Is there something better, like assigning the particular signature a hash, like Expr.id? This seems like a lot of code for something that shouldn't need to be done.
___
*[`pixeltable/exprs/function_call.py` line 588 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODspNEt2-txhlKNvfSR:-ODspNEt2-txhlKNvfSS:buw9n6w) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/exprs/function_call.py#L588)):*
> ```Python
> 
>     @classmethod
>     def __unpack_bound_arg(cls, arg: Any) -> Any:
> ```

unpack?
___
*[`pixeltable/functions/globals.py` line 55 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODst-Po0jDdg9p-0Ptr:-ODst-Po0jDdg9p-0Pts:b-por99g) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/functions/globals.py#L55)):*
> ```Python
>     type_substitutions=tuple(
>         {T: Optional[t]}  # type: ignore[misc]
>         for t in (str, int, float, bool, ts.Timestamp, ts.Json, ts.Image, ts.Video, ts.Audio, ts.Document)
> ```

Why not use all pxt types?
___
*[`pixeltable/func/function.py` line 77 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsqAKJ6nQj-KNJ2UNK:-ODsqAKJ6nQj-KNJ2UNL:b-7ck2gv) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/func/function.py#L77)):*
> ```Python
> 
>     @property
>     def is_monomorphic(self) -> bool:
> ```

Can we have an is_polymorphic instead? I've never used the term monomorphic before (and doubt most engineers outside of the compilers community would have).
___
*[`pixeltable/func/function.py` line 90 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsqg7984pFmRlm5e8E:-ODsqg7984pFmRlm5e8F:b-3g8bzm) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/func/function.py#L90)):*
> ```Python
>         return len(self.signatures[0].parameters)
> 
>     def project(self, signature_idx: int) -> Function:
> ```

Can we find a better name? I don't have an intuitive sense of what projection means here.
___
*[`pixeltable/func/function.py` line 113 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsrR2sCnNk_RDv7lTC:-ODsrR2sCnNk_RDv7lTD:bjpjaa) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/func/function.py#L113)):*
> ```Python
>         return self.__projections[signature_idx]
> 
>     def _update_as_projection(self, signature_idx: int) -> None:
> ```

I have no idea what this should do.
___
*[`pixeltable/func/function.py` line 262 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsrubdCs8EzZSazIQV:-ODsrubdCs8EzZSazIQW:bg2lfb4) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/func/function.py#L262)):*
> ```Python
>         pass
> 
>     def to_sql(self, fn: Callable[..., Optional[sql.ColumnElement]]) -> Callable[..., Optional[sql.ColumnElement]]:
> ```

How does that work? These are expected to take the same sequence of parameter names as their signature.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/403)* status: 0 of 32 files reviewed, 12 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/function_call.py` line 51 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODso2gU-EjluPOX0jVS:-ODt0fIx0tRgeAYWUK7U:b-vs8hsr) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/exprs/function_call.py#L51)):*

Why do we need to pass it in, instead of using call_return_type()?
</blockquote></details>

For when we load the return type out of the dictionary
___
*[`pixeltable/exprs/function_call.py` line 447 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsoCKn5b15ooLSEEc1:-ODt11eL56pCNg6XlWZa:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/exprs/function_call.py#L447)):*

I would prefer a property Function.agg_cls(self):
    assert not self.is_polymorphic
    return self.agg_classes[0]
</blockquote></details>

Done.
___
*[`pixeltable/exprs/function_call.py` line 488 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsoZmr1ku90mlDyV99:-ODt13s_9KnRSbo_SNt9:b-5fj872) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/exprs/function_call.py#L488)):*

Is there something better, like assigning the particular signature a hash, like Expr.id? This seems like a lot of code for something that shouldn't need to be done.
</blockquote></details>

That won't work, because the signature doesn't have to be an exact match, it just has to be consistent. See the code comment immediately below this pointing out several legitimate ways this might happen
___
*[`pixeltable/exprs/function_call.py` line 588 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODspNEt2-txhlKNvfSR:-ODt1Q_I9UMmi0_6mhDz:b-85c0p3) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/exprs/function_call.py#L588)):*

unpack?
</blockquote></details>

Alternative suggestions welcome
___
*[`pixeltable/func/function.py` line 77 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsqAKJ6nQj-KNJ2UNK:-ODtUZYj3hIpWFBN8LOa:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/func/function.py#L77)):*

Can we have an is_polymorphic instead? I've never used the term monomorphic before (and doubt most engineers outside of the compilers community would have).
</blockquote></details>

Done.
___
*[`pixeltable/func/function.py` line 90 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsqg7984pFmRlm5e8E:-ODtUWAaAsboOVZAGJLe:b5m2up8) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/func/function.py#L90)):*

Can we find a better name? I don't have an intuitive sense of what projection means here.
</blockquote></details>

Suggestions welcome
___
*[`pixeltable/func/function.py` line 113 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsrR2sCnNk_RDv7lTC:-ODtUPaMCorui4IjEzBp:b62mnfa) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/func/function.py#L113)):*

I have no idea what this should do.
</blockquote></details>

This isn't an actionable comment
___
*[`pixeltable/func/function.py` line 262 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODsrubdCs8EzZSazIQV:-ODt1xs24IggPxTr7tQN:b-4qkgtr) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/func/function.py#L262)):*

How does that work? These are expected to take the same sequence of parameter names as their signature.
</blockquote></details>

The only existing use cases for `to_sql` on polymorphic functions involve functions defined with type_substitutions, where the sequence of parameter names is identical across all the signatures.

We can dream up various ways to specify distinct to_sql methods for different signatures but it seems like unnecessary effort until we actually end up with a use case for it (and right now I don't foresee any)
___
*[`pixeltable/functions/globals.py` line 55 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/403#-ODst-Po0jDdg9p-0Ptr:-ODt1prh2RmvqUJxzklA:bd8z9ka) ([raw file](https://github.com/pixeltable/pixeltable/blob/31998bea74c297e766cd9e2ce765c4fd4338fb3f/pixeltable/functions/globals.py#L55)):*

Why not use all pxt types?
</blockquote></details>

What do you mean? What's missing here?

pixeltable/exprs/function_call.py
```diff
@@ -429,30 +441,35 @@ def eval(self, data_row: DataRow, row_builder: RowBuilder) -> None:
 
         if isinstance(self.fn, func.CallableFunction) and not self.fn.is_batched:
             # optimization: avoid additional level of indirection we'd get from calling Function.exec()
-            data_row[self.slot_idx] = self.fn.py_fn(*args, **kwargs)
+            data_row[self.slot_idx] = self.fn.py_fns[0](*args, **kwargs)
```
mkornacker: Let's also add a py_fn property.

pixeltable/exprs/function_call.py
```diff
@@ -461,15 +478,108 @@ def _from_dict(cls, d: dict, components: list[Expr]) -> FunctionCall:
...
+        }
+
+        # `Function.from_dict()` does signature matching, so it is safe to assume that `args` and `kwargs` are
+        # consistent with its signature.
+
+        # Reassemble bound_args. Note that args and kwargs represent "already bound arguments": they are not bindable
+        # in the Python sense, because variable args (such as *args and **kwargs) have already been condensed.
+        param_names = list(fn.signature.parameters.keys())
+        bound_args = {param_names[i]: arg for i, arg in enumerate(args)}
+        bound_args.update(kwargs.items())
```
mkornacker, line 504: I prefer the previous version of the code, it's more concise (4 lines). Why change it?

pixeltable/exprs/function_call.py
```diff
@@ -43,17 +45,25 @@ class FunctionCall(Expr):
...
         if group_by_clause is None:
             group_by_clause = []
-        signature = fn.signature
-        return_type = fn.call_return_type(bound_args)
+
+        assert not fn.is_polymorphic
+
         self.fn = fn
         self.is_method_call = is_method_call
-        self.normalize_args(fn.name, signature, bound_args)
```
mkornacker, line 56: Is it now a requirement that this has been called prior to init()?

pixeltable/func/aggregate_function.py
```diff
@@ -67,23 +114,30 @@ def __init__(
...
+    @property
+    def agg_class(self) -> type[Aggregator]:
+        assert not self.is_polymorphic
+        return self.agg_classes[0]
 
-    def exec(self, *args: Any, **kwargs: Any) -> Any:
+    def exec(self, args: Sequence[Any], kwargs: dict[str, Any]) -> Any:
         raise NotImplementedError
 
+    def overload(self, cls: Callable) -> AggregateFunction:
```
mkornacker: cls: Callable?

I would have expected that to be Type[Aggregator].

pixeltable/func/aggregate_function.py
```diff
@@ -67,23 +114,30 @@ def __init__(
...
+        sig, init_param_names = self.__cls_to_signature(cls)
+        self.signatures.append(sig)
+        self.agg_classes.append(cls)
+        self.init_param_names.append(init_param_names)
+        return self
+
     def help_str(self) -> str:
         res = super().help_str()
-        res += '\n\n' + inspect.getdoc(self.agg_cls.update)
+        res += '\n\n' + inspect.getdoc(self.agg_classes[0].update)
```
mkornacker, line 146: Why not agg_cls?

If that's intentional, leave comment, it's not obvious why that would be wrong.

pixeltable/func/aggregate_function.py
```diff
@@ -121,18 +175,24 @@ def __call__(self, *args: object, **kwargs: object) -> 'pixeltable.exprs.Functio
                     f'{self.display_name}(): group_by invalid with an aggregate function that does not allow windows')
             group_by_clause = kwargs.pop(self.GROUP_BY_PARAM)
 
-        bound_args = self.signature.py_signature.bind(*args, **kwargs)
-        self.validate_call(bound_args.arguments)
+        mono_fn, bound_args = self._bind_to_matching_signature(args, kwargs)
```
mkornacker: mono_fn sounds cryptic. bound_fn? resolved_fn? just fn?

pixeltable/func/callable_function.py
```diff
@@ -20,50 +22,63 @@ class CallableFunction(Function):
...
+        self.py_fns = [self.py_fns[signature_idx]]
 
     @property
     def is_batched(self) -> bool:
         return self.batch_size is not None
 
-    def exec(self, *args: Any, **kwargs: Any) -> Any:
+    def exec(self, args: Sequence[Any], kwargs: dict[str, Any]) -> Any:
+        assert not self.is_polymorphic
+        signature = self.signatures[0]
```
mkornacker: Let's add signature and py_fn as properties.

pixeltable/func/expr_template_function.py
```diff
@@ -35,18 +41,33 @@ def __init__(
                 msg = str(e)
                 raise excs.Error(f"Default value for parameter '{param.name}': {msg[0].lower() + msg[1:]}")
 
-        super().__init__(signature, self_path=self_path)
 
-    def instantiate(self, *args: object, **kwargs: object) -> 'pixeltable.exprs.Expr':
-        bound_args = self.signature.py_signature.bind(*args, **kwargs).arguments
+class ExprTemplateFunction(Function):
+    """A parameterized expression from which an executable Expr is created with a function call."""
+
```
mkornacker, line 50: Declare instance vars

pixeltable/func/expr_template_function.py
```diff
@@ -8,15 +8,21 @@
...
 
-    def __init__(
-            self, expr: 'pixeltable.exprs.Expr', signature: Signature, self_path: Optional[str] = None,
-            name: Optional[str] = None):
-        import pixeltable.exprs as exprs
         self.expr = expr
-        self.self_name = name
+        self.signature = signature
+
         self.param_exprs = list(set(expr.subexprs(expr_class=exprs.Variable)))
```
mkornacker, line 27: Missing in instance var declarations

pixeltable/func/expr_template_function.py
```diff
@@ -35,18 +41,33 @@ def __init__(
...
+
+    def _update_as_overload_resolution(self, signature_idx: int) -> None:
+        self.templates = [self.templates[signature_idx]]
+
+    def instantiate(self, args: Sequence[Any], kwargs: dict[str, Any]) -> 'pixeltable.exprs.Expr':
+        from pixeltable import exprs
+
+        assert not self.is_polymorphic
+        template = self.templates[0]
+        signature = self.signatures[0]
```
mkornacker: Doesn't Function have a signature property?

pixeltable/func/expr_template_function.py
```diff
@@ -35,18 +41,33 @@ def __init__(
...
+        super().__init__([t.signature for t in templates], self_path=self_path)
+
+    def _update_as_overload_resolution(self, signature_idx: int) -> None:
+        self.templates = [self.templates[signature_idx]]
+
+    def instantiate(self, args: Sequence[Any], kwargs: dict[str, Any]) -> 'pixeltable.exprs.Expr':
+        from pixeltable import exprs
+
+        assert not self.is_polymorphic
+        template = self.templates[0]
```
mkornacker: template property?

pixeltable/func/function.py
```diff
@@ -64,79 +73,189 @@ def display_name(self) -> str:
...
+        assert 0 <= signature_idx < len(self.signatures)
+
+        # We cache the resolutions in self.__resolutions. This ensures that each resolution is represented globally
+        # by a single Python object. We do this dynamically rather than pre-constructing them in order to avoid
+        # circular complexity in the `Function` initialization logic.
+
+        # Note that it is necessary to create a resolution even if this Function has only one signature, because
+        # there is no guarantee we have seen all the Function's @overload declarations at the time this is called.
+
+        while len(self.__resolutions) <= signature_idx:
```
mkornacker: Is there a specific reason why this goes through all prior idxs? Or could you just make resolutions a dict[idx, Function]?

pixeltable/func/function.py
```diff
@@ -64,79 +73,189 @@ def display_name(self) -> str:
...
+            # Only one signature: call _bind_to_template() and surface any errors directly
+            template = self._bind_to_template(kwargs)
+            return ExprTemplateFunction([template])
+        else:
+            # Multiple signatures: iterate over each signature and generate a template for each
+            # successful binding. If there are no successful bindings, raise a generic error.
+            # (Note that the resulting ExprTemplateFunction may have strictly fewer signatures than
+            # this Function, in the event that only some of the signatures are successfully bound.)
+            templates: list[Template] = []
+            for mono_fn in self.__resolutions:
```
mkornacker: resolutions gets populated dynamically. How do we know we have seen the relevant resolutions here? Maybe the way that this is populated can be simplified, I'm finding it hard to follow (and why the current code is correct is not obvious to me).

pixeltable/func/function.py
```diff
@@ -64,79 +73,189 @@ def display_name(self) -> str:
...
+                try:
+                    template = mono_fn._bind_to_template(kwargs)
+                    templates.append(template)
+                except (TypeError, excs.Error):
+                    continue
+            if len(templates) == 0:
+                raise excs.Error(f'Function {self.name!r} has no matching signature for arguments')
+            return ExprTemplateFunction(templates)
+
+    def _bind_to_template(self, kwargs: dict[str, Any]) -> 'Template':
```
mkornacker: create_template()? when I saw bind, I was expecting it's dealing with something that already exists.

pixeltable/func/udf.py
```diff
@@ -104,25 +108,35 @@ def make_function(
...
+        if is_property and len(sig.parameters) != 1:
+            raise excs.Error(
+                f"`is_property=True` expects a UDF with exactly 1 parameter, but `{function_name}` has {len(sig.parameters)}"
+            )
+        if (is_method or is_property) and function_path is None:
+            raise excs.Error('Stored functions cannot be declared using `is_method` or `is_property`')
+
+        signatures = [sig]
+    else:
+        assert batch_size is None and is_method is None and is_property is None
```
mkornacker: Should these be error messages?

pixeltable/func/udf.py
```diff
@@ -132,8 +146,8 @@ def make_function(
         py_fn = substitute_fn
 
     result = CallableFunction(
-        signature=sig,
-        py_fn=py_fn,
+        signatures=signatures,
+        py_fns=[py_fn] * len(signatures),
```
mkornacker: comment: all signatures stem from the same py fn

pixeltable/functions/globals.py
```diff
@@ -43,12 +45,21 @@ def _(val: sql.ColumnElement) -> Optional[sql.ColumnElement]:
...
-@func.uda(update_types=[ts.IntType(nullable=True)], value_type=ts.IntType(), allows_window=True, requires_order_by=False)
-class count(func.Aggregator):
+@func.uda(
+    allows_window=True,
+    # Allow counting non-null values of any type
+    # TODO: I couldn't include "Array" because we don't have a way to represent a generic array (of arbitrary dimension).
+    # TODO: should we have an "Any" type that can be used here?
+    type_substitutions=tuple(
+        {T: Optional[t]}  # type: ignore[misc]
+        for t in (str, int, float, bool, ts.Timestamp, ts.Json, ts.Image, ts.Video, ts.Audio, ts.Document)
```
mkornacker: Why not use ts.String, etc?

pixeltable/func/expr_template_function.py
```diff
@@ -8,15 +8,21 @@
 from .signature import Signature
 
 
-class ExprTemplateFunction(Function):
-    """A parameterized expression from which an executable Expr is created with a function call."""
+class Template:
```
mkornacker: ExprTemplate then?

pixeltable/exprs/function_call.py
aaron-siegel: Done.

pixeltable/exprs/function_call.py
aaron-siegel: I originally changed it because I had to: in a previous version of the code, signature matching was done in FunctionCall. args and kwargs are needed for signature matching, and the matched signature is needed for arg binding, so it had to be a 2-step process with something else in between.

We moved signature matching to Function (at least for now), so that intermediate step went away. But I left the code like this, because I actually find it a lot clearer. As someone who recently read the code, I can tell you that it's clearer like this to someone reading the code. There are two separate operations here (reassembling args/kwargs and then using them to reassemble bound_args) and it's easier to read if they're broken up. Brevity does not always imply clarity...

pixeltable/func/aggregate_function.py
aaron-siegel: Fixed. I'm surprised the type checker didn't catch this one. Good catch

pixeltable/func/aggregate_function.py
aaron-siegel: Because it needs to work for polymorphic aggregators. I added a comment

pixeltable/func/aggregate_function.py
aaron-siegel: Changed to resolved_fn (here and elsewhere)

pixeltable/func/callable_function.py
aaron-siegel: Done.

pixeltable/func/expr_template_function.py
aaron-siegel: Added

pixeltable/func/expr_template_function.py
aaron-siegel: Done.

pixeltable/func/expr_template_function.py
aaron-siegel: Added

pixeltable/func/expr_template_function.py
aaron-siegel: Yep this code precedes it and I overlooked this one when I added the property. Fixed.

pixeltable/func/function.py
aaron-siegel: it's not going through all prior resolutions, it's just extending the cache as necessary. sure we could use a dict but i don't see what it gains

pixeltable/func/function.py
aaron-siegel: The short answer is that we don't, and we can't. There's nothing stopping a user from doing this:

```
@udf
def my_udf(...) -> ...:

x = my_udf.using(...)

@my_udf.overload
def _(...) -> ...:
```

but as long as the user is not-insane and puts all the overload declarations adjacent in their module, we're safe.

Note that this is true of *anything* in Python that uses this pattern; a user could put random code in between a `@property` declaration and its `@setter`, for example. Or `@singledispatch` declarations or anything else.

pixeltable/func/function.py
aaron-siegel: Sure. The way I was thinking about it is that the `Template` is basically just an (augmented) tuple that holds the return value of the binding. But I changed this to `_bind_and_create_template` to be clearer

It's the `@expr_udf` analog of `_bind_to_signature` in some sense

pixeltable/func/udf.py
aaron-siegel: Done.

pixeltable/func/udf.py
aaron-siegel: Done.

pixeltable/functions/globals.py
aaron-siegel: Done.

pixeltable/func/function.py
```diff
@@ -64,37 +73,114 @@ def display_name(self) -> str:
...
         return len(self.signature.parameters)
 
+    def _overload_resolution(self, signature_idx: int) -> Function:
+        """
+        Return a new  `Function` that retains just the single signature at index `signature_idx`, and is otherwise
+        identical to this `Function`.
+        """
+        assert 0 <= signature_idx < len(self.signatures)
+
+        # We cache the resolutions in self.__resolutions. This ensures that each resolution is represented globally
```
mkornacker: move this to the class comment, it's a global invariant.

pixeltable/func/function.py
```diff
@@ -64,37 +73,114 @@ def display_name(self) -> str:
...
+        # by a single Python object. We do this dynamically rather than pre-constructing them in order to avoid
+        # circular complexity in the `Function` initialization logic.
+
+        # Note that it is necessary to create a resolution even if this Function has only one signature, because
+        # there is no guarantee we have seen all the Function's @overload declarations at the time this is called.
```
mkornacker: aren't we saying that we don't want that situation (interleaving creating overloads with creating FunctionCalls)? what else can be simplified if we don't have to deal with that situation?

pixeltable/func/function.py
```diff
@@ -64,37 +73,114 @@ def display_name(self) -> str:
...
         from pixeltable import exprs
-        bound_args = self.signature.py_signature.bind(*args, **kwargs)
-        self.validate_call(bound_args.arguments)
-        return exprs.FunctionCall(self, bound_args.arguments)
+
+        mono_fn, bound_args = self._bind_to_matching_signature(args, kwargs)
+        return_type = mono_fn.call_return_type(args, kwargs)
+        return exprs.FunctionCall(mono_fn, bound_args, return_type)
+
+    def _bind_to_matching_signature(self, args: Sequence[Any], kwargs: dict[str, Any]) -> tuple[Function, dict[str, Any]]:
```
mkornacker: this is basically doing the overload resolution (ie, determine which of the available overloads is a good match).

_overload_resolution() is simply giving you a Function for the resolved signature.

mkornacker (commented):
some more comments, as i was looking at the overload resolution process again.

pixeltable/func/callable_function.py
```diff
@@ -95,7 +95,15 @@ def name(self) -> str:
 
     def overload(self, fn: Callable) -> CallableFunction:
         if self.self_path is None:
-            raise excs.Error('@overload can only be used with module UDFs (not locally defined UDFs)')
+            raise excs.Error('`overload` can only be used with module UDFs (not locally defined UDFs)')
+        if self.is_batched:
+            raise excs.Error('`overload` cannot be used with batched functions')
+        if self.is_method or self.is_property:
+            raise excs.Error('`overload` cannot be used with cannot be used with `is_method` or `is_property`')
```
mkornacker: cannot be used with cannot be used with

pixeltable/func/function.py
```diff
@@ -32,7 +33,7 @@ class Function(abc.ABC):
     is_method: bool
     is_property: bool
     _conditional_return_type: Optional[Callable[..., ts.ColumnType]]
-    __resolutions: list[Function]  # Cached overload resolutions of this function
+    _resolutions: list[Self]  # Cached overload resolutions of this function
```
mkornacker: I don't think Self is available in python 3.9.

pixeltable/func/function.py
```diff
@@ -123,11 +124,11 @@ def help_str(self) -> str:
...
 
-        mono_fn, bound_args = self._bind_to_matching_signature(args, kwargs)
-        return_type = mono_fn.call_return_type(args, kwargs)
-        return exprs.FunctionCall(mono_fn, bound_args, return_type)
+        resolved_fn, bound_args = self._bind_to_matching_signature(args, kwargs)
+        return_type = resolved_fn.call_return_type(args, kwargs)
+        return exprs.FunctionCall(resolved_fn, bound_args, return_type)
 
-    def _bind_to_matching_signature(self, args: Sequence[Any], kwargs: dict[str, Any]) -> tuple[Function, dict[str, Any]]:
+    def _bind_to_matching_signature(self, args: Sequence[Any], kwargs: dict[str, Any]) -> tuple[Self, dict[str, Any]]:
```
mkornacker, line 142: I would still argue that this is the actual overload resolution, whereas _overload_resolution() is basically get_overload(idx). Instead of the current _overload_resolution(), you could even do a
@property
def overloads() -> list[Function]
    if len(_resolutions) == 0:
        ...
    return self._resolutions
and then fn.overloads[i] looks pretty compact and still doesn't obscure anything.

mkornacker (approved):

pixeltable/func/function.py
aaron-siegel: Done.

pixeltable/func/function.py
aaron-siegel: Good point. Fixed. Not sure that there's anything else. I did add some checks in `@overload` to ensure new overloads can't be added once either the resolutions have been cached, or a conditional_return_type has been set.

pixeltable/func/function.py
aaron-siegel: See comment below

pixeltable/func/callable_function.py
aaron-siegel: Fixed.

pixeltable/func/expr_template_function.py
aaron-siegel: Ok

pixeltable/func/function.py
aaron-siegel: Just adding a note: with the most recent changes, once `using` has been called, subsequent `@overload` decorators will raise an error, so I think that answers this point.

pixeltable/func/function.py
aaron-siegel: Yes it is, via the `typing_extensions` package, which is how I'm importing it here.

I do all my development on 3.9 (and we have CI matrix configs that run against 3.9) so rest assured I've thought of and tested this :)

pixeltable/func/function.py
aaron-siegel: Hmm. In my mind they're "resolutions", not "overloads": the subtle difference being that the "overloads" are the different signatures and bodies that belong to the polymorphic function, and the "resolutions" are the corresponding signature-specific (non-polymorphic) functions.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/403)* status: 0 of 32 files reviewed, 36 unresolved discussions (waiting on @mkornacker)

PR #406, opened 2024-12-12
PXT-374: Bug: cannot create an index with the same name as one that e…

anushas:
…xisted before

When a named index is dropped, and then added again, the add step fails with an assertion that the index is already present in the index metadata. This error persists even on reloading from persistence.

RCA: table_version.drop_index code does not remove the index entry from the index metadata on success. It only removes it from the idxs_by_name.

This commit fixes the bug by updating the index metadata after drop.

Testing: a testcase that reproduced the bug before, passes with fix.

### DIFF
--- a/pixeltable/catalog/table_version.py
+++ b/pixeltable/catalog/table_version.py
@@ -494,6 +494,8 @@ class TableVersion:
         idx_md.schema_version_drop = self.schema_version
         assert idx_md.name in self.idxs_by_name
         idx_info = self.idxs_by_name[idx_md.name]
+        # remove this index entry from the active indexes (in memory)
+        # and the index metadata (in persistent table metadata)
         del self.idxs_by_name[idx_md.name]
         del self.idx_md[idx_id]
 
--- a/pixeltable/catalog/table_version.py
+++ b/pixeltable/catalog/table_version.py
@@ -495,6 +495,7 @@ class TableVersion:
         assert idx_md.name in self.idxs_by_name
         idx_info = self.idxs_by_name[idx_md.name]
         del self.idxs_by_name[idx_md.name]
+        del self.idx_md[idx_id]
 
         with Env.get().engine.begin() as conn:
             self._drop_columns([idx_info.val_col, idx_info.undo_col])
--- a/tests/test_index.py
+++ b/tests/test_index.py
@@ -155,26 +155,31 @@ class TestIndex:
         t = small_img_tbl
         sample_img = t.select(t.img).head(1)[0, 'img']
         t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
+        orig_res = t.select(t.img.localpath).order_by(t.img.similarity(sample_img, idx='clip_idx'), asc=False).limit(3).collect()
         t.revert()
         # creating an index with the same name again after a revert should be successful
         t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
-        _ = t.select(t.img.localpath).order_by(t.img.similarity(sample_img), asc=False).limit(3).collect()
+        res = t.select(t.img.localpath).order_by(t.img.similarity(sample_img, idx='clip_idx'), asc=False).limit(3).collect()
+        assert_resultset_eq(orig_res, res, True)
         t.revert()
         # should be true even after reloading from persistence
-        _ = reload_tester.run_reload_test()
+        reload_tester.run_reload_test()
         t = pxt.get_table('small_img_tbl')
         t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
-        _ = t.select(t.img.localpath).order_by(t.img.similarity(sample_img), asc=False).limit(3).collect()
+        res = t.select(t.img.localpath).order_by(t.img.similarity(sample_img, idx='clip_idx'), asc=False).limit(3).collect()
+        assert_resultset_eq(orig_res, res, True)
 
         # same should hold after a drop.
         t.drop_embedding_index(column='img')
         t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
-        _ = t.select(t.img.localpath).order_by(t.img.similarity(sample_img), asc=False).limit(3).collect()
+        res = t.select(t.img.localpath).order_by(t.img.similarity(sample_img, idx='clip_idx'), asc=False).limit(3).collect()
+        assert_resultset_eq(orig_res, res, True)
         t.drop_embedding_index(idx_name='clip_idx')
-        _ = reload_tester.run_reload_test()
+        reload_tester.run_reload_test()
         t = pxt.get_table('small_img_tbl')
         t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
-        _ = t.select(t.img.localpath).order_by(t.img.similarity(sample_img), asc=False).limit(3).collect()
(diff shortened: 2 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/406)

aaron-siegel (changes requested):
lgtm with one comment, but it'd be good for @mkornacker to have a look too

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/406#-:-OE19ACF7kqTbGveu_8V:b-9r7gms)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @anushas)

___
*[`tests/test_index.py` line 167 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/406#-OE18nqEBifRMWBO6Wza:-OE18nqEBifRMWBO6Wzb:bqp9yxp) ([raw file](https://github.com/pixeltable/pixeltable/blob/7b58c05d68c03c654cb79feab427b99b12cddc1c/tests/test_index.py#L167)):*
> ```Python
>         t = pxt.get_table('small_img_tbl')
>         t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
>         _ = t.select(t.img.localpath).order_by(t.img.similarity(sample_img), asc=False).limit(3).collect()
> ```

it'd be great to validate the query results here (not just the fact that the query runs), e.g., check that they're the same after each revert/add and drop/add cycle

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/406)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @aaron-siegel)

___
*[`tests/test_index.py` line 167 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/406#-OE18nqEBifRMWBO6Wza:-OE9Syb18aDTdUlLks0n:b-hb9nnx) ([raw file](https://github.com/pixeltable/pixeltable/blob/7b58c05d68c03c654cb79feab427b99b12cddc1c/tests/test_index.py#L167)):*

it'd be great to validate the query results here (not just the fact that the query runs), e.g., check that they're the same after each revert/add and drop/add cycle
</blockquote></details>

Done. Improved the check, thanks.

tests/test_index.py
```diff
@@ -149,6 +149,38 @@ def test_similarity_errors(self, indexed_img_tbl: pxt.Table, small_img_tbl: pxt.
...
+        t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
+        res = t.select(t.img.localpath).order_by(t.img.similarity(sample_img, idx='clip_idx'), asc=False).limit(3).collect()
+        assert_resultset_eq(orig_res, res, True)
+        t.revert()
+        # should be true even after reloading from persistence
+        reload_tester.run_reload_test()
+        t = pxt.get_table('small_img_tbl')
+        t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
+        res = t.select(t.img.localpath).order_by(t.img.similarity(sample_img, idx='clip_idx'), asc=False).limit(3).collect()
+        assert_resultset_eq(orig_res, res, True)
```
mkornacker, line 170: This is what the reload tester is for.

tests/test_index.py
```diff
@@ -149,6 +149,38 @@ def test_similarity_errors(self, indexed_img_tbl: pxt.Table, small_img_tbl: pxt.
...
+        t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
+        orig_res = t.select(t.img.localpath).order_by(t.img.similarity(sample_img, idx='clip_idx'), asc=False).limit(3).collect()
+        t.revert()
+        # creating an index with the same name again after a revert should be successful
+        t.add_embedding_index('img', idx_name='clip_idx', image_embed=clip_img_embed, string_embed=clip_text_embed)
+        res = t.select(t.img.localpath).order_by(t.img.similarity(sample_img, idx='clip_idx'), asc=False).limit(3).collect()
+        assert_resultset_eq(orig_res, res, True)
+        t.revert()
+        # should be true even after reloading from persistence
+        reload_tester.run_reload_test()
```
mkornacker: This isn't doing anything other than calling reload_catalog().

pixeltable/catalog/table_version.py
```diff
@@ -495,6 +495,7 @@ def drop_index(self, idx_id: int) -> None:
         assert idx_md.name in self.idxs_by_name
         idx_info = self.idxs_by_name[idx_md.name]
         del self.idxs_by_name[idx_md.name]
+        del self.idx_md[idx_id]
```
mkornacker, line 500: missing explanatory comment

mkornacker (changes requested):

tests/test_index.py
anushas: Yes, that was my intent. That even after reloading, the state is clean and the another index by same name can be created. I am replacing reload_catalog() with reload_test.run_reload_test() in new tests (even when no queries are to be verified). Let me know if we prefer to keep using reload_catalog() in such cases.

FWIW, t.revert() did not have any issue. The bug was only in t.drop_index(). But I retained revert() also in the final testing to keep the code coverage.

tests/test_index.py
anushas: In in this case, I am trying to compare the result of the similarity query with original index, and then the new index recreated after revert/drop, without any reload in between. In this testcase, at the time of reload, there's no index. The purpose of reload is to make sure drop_index cleans up the state in persistence and allows creating that index again. Using the index in the query is a sanity check to make sure the creation was correct (which is tested elsewhere, but not in this particular sequence). HTH.

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/406)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

pixeltable/catalog/table_version.py
anushas: I have added a comment, please let me know if this is what you had in mind.

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/406)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

tests/test_index.py
aaron-siegel: @mkornacker Is there an explicit preference for calling `reload_catalog()` directly if there are no queries to test after reload? (In this case there are no queries to test since `revert()` was called before the reload.)

tests/test_index.py
aaron-siegel: @mkornacker I don't think so? Here we're validating that the in-memory query results are stable over the intervening operations. That's a different scenario from what the reload tester checks for.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/406)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @mkornacker)

aaron-siegel (approved):

tests/test_index.py
mkornacker: If you don't run any queries through reload_tester, there is nothing to test.

In that case, calling reload_catalog() directly is clearer.

tests/test_index.py
mkornacker: Yes, true here, my mistake.

mkornacker (approved):

tests/test_index.py
anushas: Done.

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/406)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

PR #407, opened 2024-12-12
Enhance Makefile for cross-platform compatibility

jacobweiss2305:
- Updated the Makefile to detect the operating system and set the shell accordingly (PowerShell for Windows, Bash for others).
- Modified installation commands to create necessary directories and files in a platform-agnostic manner.
- Adjusted commands for running tests and linting to ensure compatibility with both Windows and Unix-like systems.
- Implemented conditional logic for cleaning up files and directories based on the operating system.

This commit improves the usability of the project on Windows environments, ensuring that setup and testing processes work seamlessly across different platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,33 @@
 ifeq ($(OS),Windows_NT)
     SHELL := pwsh.exe
     # PowerShell command to get directory name
-    KERNEL_NAME := $(shell (Get-Item .).Name)
+    # Define Windows-specific commands
+    SHELL_PREFIX := pwsh.exe
+    MKDIR := powershell -Command New-Item -ItemType Directory -Path
+    TOUCH := powershell -Command New-Item -ItemType File -Path
+    RM := powershell -Command Remove-Item -Force
+    RMDIR := powershell -Command Remove-Item -Force -Recurse
+    SET_ENV := set
+    DIR_SEP := \\
+    KERNEL_NAME := $(shell powershell -Command "(Get-Item .).Name")
+    ULIMIT_CMD :=
 else
     SHELL := /bin/bash
+    # Define Unix-specific commands
+    SHELL_PREFIX :=
+    MKDIR := mkdir -p
+    TOUCH := touch
+    RM := rm -f
+    RMDIR := rm -rf
+    SET_ENV := export
+    DIR_SEP := /
     KERNEL_NAME := $(shell basename `pwd`)
+    ULIMIT_CMD := ulimit -n 4000;
 endif
+
+# Common test parameters
+PYTEST_COMMON_ARGS := -v -n auto --dist loadgroup --maxprocesses 6 tests
+NB_CELL_TIMEOUT := 3600
 
 .DEFAULT_GOAL := help
 
@@ -36,9 +58,9 @@ help:
 .PHONY: setup-install
 setup-install:
 ifeq ($(OS),Windows_NT)
-	@if (-not (Test-Path .make-install)) { New-Item -ItemType Directory -Path .make-install }
+	@powershell -Command "if (-not (Test-Path '.make-install')) { New-Item -ItemType Directory -Path '.make-install' }"
 else
-	@mkdir -p .make-install
+	@if [ ! -d ".make-install" ]; then $(MKDIR) .make-install; fi
 endif
 ifdef CONDA_DEFAULT_ENV
 ifeq ($(CONDA_DEFAULT_ENV),base)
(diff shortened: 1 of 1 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/407)

aaron-siegel (commented):
As written, this change adds a lot of independent conditionals that make the Makefile harder to read and maintain. I wonder if we can reduce them by setting variables approprately. e.g., set a SHELL_PREFIX variable that is `pwsh.exe` on Windows, and the empty string on Linux/Mac, and then do

`@$SHELL_PREFIX scripts/lint-changed-files.sh`

and so on. Likewise, there could be a TOUCH_CMD variable that's `touch` on Linux/Mac, and `@New-Item -ItemType File -Path -Force` on Windows. [aside: I've never used powershell before, but wow, that's a verbose way to say "touch".] Same for mkdir, set/export, etc.

If this works, we could get away with a single conditional at the top that defines variables for all the different commands. The main issue with the present approach is that every time we make a change (e.g. by adjusting the `pytest` parameters) we have to make the identical change in multiple places.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/407#-:-ODx0YLM19ikn1HrndPj:bn1ptdc)* status: 0 of 1 files reviewed, all discussions resolved

jacobweiss2305:
Great call. 

This refactoring addresses maintainability concerns by:
- Defining OS-specific command variables at the top (SHELL_PREFIX, MKDIR, TOUCH, etc.)
- Consolidating repeated commands into variables (PYTEST_COMMON_ARGS)
- Using proper PowerShell commands with full path escaping
- Making file operations OS-agnostic via DIR_SEP variable
- Adding directory existence checks for robustness

aaron-siegel (changes requested):
Looks much better, although the use of `$(DIR_SEP)` bothers me a bit. Is that the standard way of implementing portable makefiles?

I'm looking at the Postgres Makefile as a reference. It builds on Windows, and it uses forward slashes as path separators.

I have limited experience with Windows development but it'd be great to have confidence that we're following standard practices here.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/407#-:-OE15UZ3AgIRmjxFfa27:bjqp392)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @jacobweiss2305)

___
*[`Makefile` line 98 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/407#-OE16MDy7kgbBdG-pNpn:-OE16MDy7kgbBdG-pNpo:bzbxucr) ([raw file](https://github.com/pixeltable/pixeltable/blob/c61f069169087d15766abb3112445d6f6942eb8d/Makefile#L98)):*
> ```Makefile
> 
> NB_CELL_TIMEOUT := 3600
> # We ensure the TQDM progress bar is updated exactly once per cell execution, by setting the refresh rate equal
> ```

Why did you get rid of this comment? It's useful context.
___
*[`Makefile` line 63 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/407#-OE14osP4_CmW3SMk1tV:-OE14osP4_CmW3SMk1tW:b4nbr16) ([raw file](https://github.com/pixeltable/pixeltable/blob/7ba27d51aa0db832654ecaa5747ad22a3a839d8c/Makefile#L63)):*
> ```Makefile
> 	@powershell -Command "if (-not (Test-Path '.make-install')) { New-Item -ItemType Directory -Path '.make-install' }"
> else
> 	@if [ ! -d ".make-install" ]; then $(MKDIR) .make-install; fi
> ```

Why did this change?

Why aren't we just doing `$(MKDIR) .make_install` here?
___
*[`Makefile` line 129 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/407#-OE155PD9-GXjQOA6uam:-OE155PD9-GXjQOA6uan:b-t6z0za) ([raw file](https://github.com/pixeltable/pixeltable/blob/7ba27d51aa0db832654ecaa5747ad22a3a839d8c/Makefile#L129)):*
> ```Makefile
> fullpytest: install
> 	@echo "Running pytest, including expensive tests ..."
> 	@$(ULIMIT_CMD) pytest -v -m '' -n auto --dist loadgroup --maxprocesses 6 tests
> ```

use `$(PYTEST_COMMON_ARGS)` here?

`$(ULIMIT_CMD) pytest -m '' $(PYTEST_COMMON_ARGS)`

jacobweiss2305:
@aaron-siegel I added additional logic to use .make-install folder if it already exists otherwise create directory:

```
	@powershell -Command "if (-not (Test-Path '.make-install')) { New-Item -ItemType Directory -Path '.make-install' }"
else
	@if [ ! -d ".make-install" ]; then $(MKDIR) .make-install; fi
```

jacobweiss2305:
The reason I didn't think we should use `PYTEST_COMMON_ARGS := -v -n auto --dist loadgroup --maxprocesses 6 tests` is meant for tag specific markers. For example it could skip tests that are marked as slow or expensive

Where as the make command below will run tests on ALL tests (no marker filtering)

```
fullpytest: install
	@echo "Running pytest, including expensive tests ..."
	@$(ULIMIT_CMD) pytest -v -m '' -n auto --dist loadgroup --maxprocesses 6 tests
```

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/407)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @jacobweiss2305)

___
*[`Makefile` line 61 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/407#-OEXy3xkCD-Q_ETeqTME:-OEXy3xkCD-Q_ETeqTMF:bajv7ov) ([raw file](https://github.com/pixeltable/pixeltable/blob/cf7dc3ea74b809c64fc8e172c2f2a5418609348f/Makefile#L61)):*
> ```Makefile
> setup-install:
> ifeq ($(OS),Windows_NT)
> 	@powershell -Command "if (-not (Test-Path '.make-install')) { New-Item -ItemType Directory -Path '.make-install' }"
> ```

Why a separate conditional here instead of just using the $(MKDIR) command for powershell?
___
*[`Makefile` line 81 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/407#-OEXxwTLE6QRWcvahfux:-OEXxwTLE6QRWcvahfuy:b-luo9bd) ([raw file](https://github.com/pixeltable/pixeltable/blob/cf7dc3ea74b809c64fc8e172c2f2a5418609348f/Makefile#L81)):*
> ```Makefile
> 	@python -m pip install -q poetry==1.8.4
> 	@poetry self add "poetry-dynamic-versioning[plugin]"
> 	@$(TOUCH) .make-install/poetry -Force
> ```

`-Force` isn't UNIX-friendly; it should be moved inside the `$(TOUCH)` command for Windows.

PR #410, opened 2024-12-17
Fix for change in Ollama API behavior

aaron-siegel:
Ollama changed the behavior of their API, which breaks our CI. This should fix it.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/410)

mkornacker (approved):

PR #411, opened 2024-12-18
PXT-382: Bug: Similarity queries do not work correctly with ReloadTester

anushas:
When there are multiple embedding indexes on a column, the similarity query must specify the index to use. When such a similarity query with a particular index specified is executed after reloading metadata from persistence via the ReloadTester APIs, it fails to run.

RCA: the index name is passed to the SimilarityExpr as a runtime parameter and is not serialized to persistence as part of the SimilarityExpr. Therefore, on reloading from persistence, the index name information is lost and is set as None. Since there are multiple indexes, the SimilarityExpr fails with an error.

This commit fixes SimilarityExpr to include the index used.

Testing: added coverage for the changes. test repro that failed before, passes with the fix.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -146,7 +146,22 @@ def img_tbl_exprs(indexed_img_tbl: catalog.Table) -> list[exprs.Expr]:
         t.img.rotate(90).resize([224, 224]),
         t.img.fileurl,
         t.img.localpath,
+    ]
+
+@pytest.fixture(scope='function')
+def sim_exprs_same_idx(indexed_img_tbl: catalog.Table) -> list[exprs.Expr]:
+    t = indexed_img_tbl
+    return [
         t.img.similarity('red truck'),
+        t.img.similarity('red truck', idx='img_idx0')
+    ]
+
+@pytest.fixture(scope='function')
+def sim_exprs_diff_idx(multi_idx_img_tbl: catalog.Table) -> list[exprs.Expr]:
+    t = multi_idx_img_tbl
+    return [
+        t.img.similarity('red truck', idx='img_idx1'),
+        t.img.similarity('red truck', idx='img_idx2'),
     ]
 
 @pytest.fixture(scope='function')
@@ -158,5 +173,14 @@ def indexed_img_tbl(reset_db) -> pxt.Table:
     skip_test_if_not_installed('transformers')
     t = create_img_tbl('indexed_img_tbl', num_rows=40)
     from .utils import clip_img_embed, clip_text_embed
-    t.add_embedding_index('img', metric='cosine', image_embed=clip_img_embed, string_embed=clip_text_embed)
+    t.add_embedding_index('img', idx_name='img_idx0', metric='cosine', image_embed=clip_img_embed, string_embed=clip_text_embed)
+    return t
+
+@pytest.fixture(scope='function')
+def multi_idx_img_tbl(reset_db) -> pxt.Table:
+    skip_test_if_not_installed('transformers')
+    t = create_img_tbl('multi_idx_img_tbl', num_rows=4)
+    from .utils import clip_img_embed, clip_text_embed
+    t.add_embedding_index('img', idx_name='img_idx1', metric='cosine', image_embed=clip_img_embed, string_embed=clip_text_embed)
+    t.add_embedding_index('img', idx_name='img_idx2', metric='cosine', image_embed=clip_img_embed, string_embed=clip_text_embed)
     return t
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -146,18 +146,11 @@ def img_tbl_exprs(indexed_img_tbl: catalog.Table) -> list[exprs.Expr]:
         t.img.rotate(90).resize([224, 224]),
         t.img.fileurl,
         t.img.localpath,
-    ]
-
-@pytest.fixture(scope='function')
-def sim_exprs_same_idx(indexed_img_tbl: catalog.Table) -> list[exprs.Expr]:
-    t = indexed_img_tbl
-    return [
-        t.img.similarity('red truck'),
         t.img.similarity('red truck', idx='img_idx0')
     ]
 
 @pytest.fixture(scope='function')
-def sim_exprs_diff_idx(multi_idx_img_tbl: catalog.Table) -> list[exprs.Expr]:
+def multi_img_tbl_exprs(multi_idx_img_tbl: catalog.Table) -> list[exprs.Expr]:
     t = multi_idx_img_tbl
     return [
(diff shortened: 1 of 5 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/411)

tests/conftest.py
```diff
@@ -146,7 +146,22 @@ def img_tbl_exprs(indexed_img_tbl: catalog.Table) -> list[exprs.Expr]:
...
         t.img.fileurl,
         t.img.localpath,
+    ]
+
+@pytest.fixture(scope='function')
+def sim_exprs_same_idx(indexed_img_tbl: catalog.Table) -> list[exprs.Expr]:
+    t = indexed_img_tbl
+    return [
         t.img.similarity('red truck'),
+        t.img.similarity('red truck', idx='img_idx0')
```
mkornacker, line 149: why not just leave these in img_tbl_exprs?

tests/conftest.py
```diff
@@ -146,7 +146,22 @@ def img_tbl_exprs(indexed_img_tbl: catalog.Table) -> list[exprs.Expr]:
...
+    ]
+
+@pytest.fixture(scope='function')
+def sim_exprs_diff_idx(multi_idx_img_tbl: catalog.Table) -> list[exprs.Expr]:
+    t = multi_idx_img_tbl
+    return [
+        t.img.similarity('red truck', idx='img_idx1'),
```
mkornacker, line 156: why not multi_idx_img_tbl_exprs?

tests/test_exprs.py
```diff
@@ -865,41 +865,44 @@ def test_similarity2(self, img_tbl: catalog.Table) -> None:
...
             assert e.id is not None
             d[e.id] = e
-        for e in img_tbl_exprs:
+        assert len(d) == len(test_tbl_exprs) + len(img_tbl_exprs) + len(sim_exprs_diff_idx)
+        nexprs_before = len(d)
+        for e in sim_exprs_same_idx:
             assert e.id is not None
             d[e.id] = e
-        assert len(d) == len(test_tbl_exprs) + len(img_tbl_exprs)
+        assert len(d) == nexprs_before + 1
```
mkornacker: you're testing something here that wasn't part of the original test (its intention), which is that the two similarity exprs are identical. that should go in a different test case in test_index.

tests/test_exprs.py
```diff
@@ -865,41 +865,44 @@ def test_similarity2(self, img_tbl: catalog.Table) -> None:
...
             e_serialized = e.serialize()
             e_deserialized = Expr.deserialize(e_serialized)
             assert e.equals(e_deserialized)
 
-        for e in img_tbl_exprs:
-            e_serialized = e.serialize()
-            e_deserialized = Expr.deserialize(e_serialized)
-            assert e.equals(e_deserialized)
+        serialized_vals = set(e.serialize() for e in sim_exprs_same_idx)
+        assert len(serialized_vals) == 1
```
mkornacker: Also something that belongs in a different test case.

tests/conftest.py
anushas: because i wanted to also verify they have same id. I have reverted this now that i moved the extra checks to another testcase.

tests/conftest.py
anushas: because it only had similarity exprs. I have changed the name, it fits better. thanks!

tests/test_exprs.py
anushas: that's a good idea. since it is focused on the expression, i moved it to test_similarity inside test_expr.py. hope you agree.
i initially thought it was ok to extend this test case as it was already verifying each id was unique. but its better separate, thanks!

tests/test_exprs.py
anushas: Done.

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/411)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @mkornacker)

anushas:
@mkornacker addressed the comments, and also rebased to get the drop index bug fix and enable the commented test scenario for this one.  PTAL. Sorry had to force push to refresh PR.

tests/test_exprs.py
```diff
@@ -864,42 +864,57 @@ def test_similarity2(self, img_tbl: catalog.Table) -> None:
...
+        t1 = indexed_img_tbl
+        # for a table with a single embedding index, whether we
+        # specify the index or not, the similarity expression
+        # would use that index. So these exressions should be equivalent.
+        sim1 = t1.img.similarity('red truck')
+        sim2 = t1.img.similarity('red truck', idx='img_idx0')
+        assert sim1.id == sim2.id
+        assert sim1.serialize() == sim2.serialize()
+        # index name should be present in the serialized expression
+        assert 'img_idx0' in sim1.serialize()
```
mkornacker: That's internal to the implementation and shouldn't be part of the test.

tests/test_exprs.py
```diff
@@ -864,42 +864,57 @@ def test_similarity2(self, img_tbl: catalog.Table) -> None:
...
+        assert 'img_idx0' in sim2.serialize()
+
+        t2 = multi_idx_img_tbl
+        # for a table with multiple embedding index, the index
```
mkornacker: grammar

tests/test_index.py
```diff
@@ -27,28 +27,63 @@ def bad_embed(x: str) -> str:
...
+        # on reload, because the index is no longer available
+        t.drop_embedding_index(idx_name='img_idx1')
+        with pytest.raises(pxt.Error) as exc_info:
+            _ = reload_tester.run_reload_test(clear=False)
+        assert "index 'img_idx1' not found" in str(exc_info.value).lower()
+
+        # After the query is serialized, dropping and recreating the index should work
+        # on reload, because the index is available again even if it is not the exact
+        # same one.
+        # recreate the index
```
mkornacker: stray line?

mkornacker (approved):

tests/test_exprs.py
anushas: Okay, removed.

tests/test_exprs.py
anushas: Corrected. Also corrected a typo.

tests/test_index.py
anushas: Removed.

anushas (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/411)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

PR #417, opened 2024-12-21
PXT-388: Change the SQL Engine isolation level to 'REPEATABLE READ'

anushas:
This commit changes the SQL Engine default isolation level from AUTOCOMMIT to REPEATABLE READ. This exposed two bugs:

1. batch_update() API hangs when if_not_exists='insert'

RCA: In the "insert if not exists" mode, batch_update() translates to two steps: a) update existing rows in the batch to new values, and b) insert non-existing rows in the batch. The code was using two different connection for these two steps. This resulted in a self-deadlock when updating the table metadata row that is common for both steps.

This commit fixes the code to use the same connection for both steps.

2. make_timestamp() resulted in sqlalchemy.exc.InternalError and transcation abort.

RCA: make_timestamp() UDF to SQL conversion was producting wrong SQL query. It cast a value to sql.Double which generates "double" which is not a valid datatype in Postgres and results in an error. Even though the error is seen irrespective of isolation level, for repeated read the connection is lost and xact aborted.

This commit fixes the code to use sql.Float instead.

### DIFF
--- a/pixeltable/functions/timestamp.py
+++ b/pixeltable/functions/timestamp.py
@@ -232,7 +232,7 @@ def _(
         sql.cast(day, sql.Integer),
         sql.cast(hour, sql.Integer),
         sql.cast(minute, sql.Integer),
-        sql.cast(second + microsecond / 1000000.0, sql.Double))
+        sql.cast(second + microsecond / 1000000.0, sql.Float))
 
 # @pxt.udf
 # def date(self: datetime) -> datetime:
--- a/pixeltable/catalog/table_version.py
+++ b/pixeltable/catalog/table_version.py
@@ -831,7 +831,7 @@ class TableVersion:
                 if error_if_not_exists:
                     raise excs.Error(f'batch_update(): {len(unmatched_rows)} row(s) not found')
                 if insert_if_not_exists:
-                    insert_status = self.insert(unmatched_rows, None, print_stats=False, fail_on_exception=False)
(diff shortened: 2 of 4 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/417)

pixeltable/functions/timestamp.py
```diff
@@ -232,7 +232,7 @@ def _(
         sql.cast(day, sql.Integer),
         sql.cast(hour, sql.Integer),
         sql.cast(minute, sql.Integer),
-        sql.cast(second + microsecond / 1000000.0, sql.Double))
+        sql.cast(second + microsecond / 1000000.0, sql.Float))
```
mkornacker, line 235: What does this do?

anushas: It casts to "float" which is an alias for double precision in postgres (and mysql). HTH

anushas: SQL generated before: CAST(EXTRACT(second FROM tbl_86ac1840fd56429798fffecb6dc1204a.col_0) + 0 / CAST(1000000.0 AS FLOAT) AS **_DOUBLE_**)) 
SQL generated now: CAST(EXTRACT(second FROM tbl_86ac1840fd56429798fffecb6dc1204a.col_0) + 0 / CAST(1000000.0 AS FLOAT) AS **_FLOAT_**))

mkornacker (approved):

PR #420, opened 2024-12-24
Windows Support: Update Makefile to enforce file creation with -Force option

jacobweiss2305:
## Summary
     - Added -Force option to file creation commands in Makefile for Windows compatibility
     - Ensures .make-install/poetry, .make-install/deps, and .make-install/others files are created without errors
     - Improves installation process reliability on Windows systems

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ setup-install:
 ifeq ($(OS),Windows_NT)
 	@powershell -Command "if (-not (Test-Path '.make-install')) { New-Item -ItemType Directory -Path '.make-install' }"
 else
-	@if [ ! -d ".make-install" ]; then $(MKDIR) .make-install; fi
+	@$(MKDIR) .make-install
 endif
 ifdef CONDA_DEFAULT_ENV
 ifeq ($(CONDA_DEFAULT_ENV),base)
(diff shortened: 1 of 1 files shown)


aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/420)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jacobweiss2305 and @mkornacker)

___
*[`Makefile` line 63 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/420#-OEq7MrA0pfccm4GqAgZ:-OEq7MrA0pfccm4GqAg_:b-7f2nmx) ([raw file](https://github.com/pixeltable/pixeltable/blob/13eeb429473f224ed9ff6643f71d096c26dc0dc8/Makefile#L63)):*
> ```Makefile
> 	@powershell -Command "if (-not (Test-Path '.make-install')) { New-Item -ItemType Directory -Path '.make-install' }"
> else
> 	@if [ ! -d ".make-install" ]; then $(MKDIR) .make-install; fi
> ```

This shouldn't be necessary. `mkdir -p` should suffice.

aaron-siegel (approved):

Issue #412, opened 2024-12-19
Fresh installation of pixeltable in a docker container fails due to missing directories 


zhenzhongxu:
Repro environment:
- Docker (codespace, debian bookworm)
- conda create -n pxt python=3.12
- pip install pixeltable
- pxt.init() or pxt.drop_dir() throws below error:

```
PermissionError                           Traceback (most recent call last)
Cell In[4], [line 4](vscode-notebook-cell:?execution_count=4&line=4)
      [1](vscode-notebook-cell:?execution_count=4&line=1) import pixeltable as pxt
      [3](vscode-notebook-cell:?execution_count=4&line=3) # Create the directory `demo` (if it doesn't already exist)
----> [4](vscode-notebook-cell:?execution_count=4&line=4) pxt.drop_dir('demo1', force=True)  # First drop `demo` to ensure a clean environment
      [5](vscode-notebook-cell:?execution_count=4&line=5) # pxt.create_dir('demo')
      [6](vscode-notebook-cell:?execution_count=4&line=6) 
      [7](vscode-notebook-cell:?execution_count=4&line=7) # # Create the table `demo.first` with a single column `input_image`
      [8](vscode-notebook-cell:?execution_count=4&line=8) # t = pxt.create_table('demo.first', {'input_image': pxt.Image})

File /opt/conda/envs/pai-dev/lib/python3.12/site-packages/pixeltable/globals.py:442, in drop_dir(path_str, force, ignore_errors)
    [422](https://vscode-remote+codespaces-002bopulent-002dspace-002dcarnival-002dpgw7x7gwgr39j5.vscode-resource.vscode-cdn.net/opt/conda/envs/pai-dev/lib/python3.12/site-packages/pixeltable/globals.py:422) def drop_dir(path_str: str, force: bool = False, ignore_errors: bool = False) -> None:
    [423](https://vscode-remote+codespaces-002bopulent-002dspace-002dcarnival-002dpgw7x7gwgr39j5.vscode-resource.vscode-cdn.net/opt/conda/envs/pai-dev/lib/python3.12/site-packages/pixeltable/globals.py:423)     """Remove a directory.
    [424](https://vscode-remote+codespaces-002bopulent-002dspace-002dcarnival-002dpgw7x7gwgr39j5.vscode-resource.vscode-cdn.net/opt/conda/envs/pai-dev/lib/python3.12/site-packages/pixeltable/globals.py:424) 
[...]

PR #421, opened 2024-12-24
Fixes for pgserver

aaron-siegel:
Fixes for two issues with pgserver:
- pgserver fails to initialize on Linux installations where `run/user/$uid` does not exist (#412 / PXT-389)
- pgserver goes into an undefined state if its parent process is terminated on Windows (PXT-370)

Also updates the CI configuration for ubuntu-arm64 and ubuntu-x64-t4 configurations.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/421)

anushas (approved):

mkornacker (approved):

PR #435, opened 2025-01-13
PXT-371: add table context to ColumnRef

anushas:
A column in the base table can be referenced by the table
or a view/snapshot derived from it. This commit adds a table
context to the ColumnRef expression to differentiate the context
in which the column is being used/referenced. 
This is currently needed to select the right embedding index on 
a base column, and to produce the right result when a column is
queried directly (v.colname.show()).
The table context is not used anywhere else.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/435)

PR #436, opened 2025-01-14
Support tool choice in OpenAI and Anthropic

aaron-siegel:
Introduces a `ToolChoice` abstraction that controls tool choice for tool calling APIs. Implements tool choice for OpenAI and Anthropic. Includes unit tests for a variety of tool-calling scenarios, including multi-tool invocations with various tool choice combinations.

Enables notebook UDFs to be used as tools.

Also includes a stealth change to accept scalars in Json-typed columns or UDF parameters (this ended up being unnecessary for tool calling, but is a generally useful change, so I left it in).


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/436)

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/436)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @aaron-siegel)

___
*[`tests/utils.py` line 555 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/436#-OGkFljA9WLXgzjWaPtP:-OGkFljA9WLXgzjWaPtQ:bb6xb9l) ([raw file](https://github.com/pixeltable/pixeltable/blob/624eee24fed02f9c14df51d075b96f5c6c3f8659/tests/utils.py#L555)):*
> ```Python
>     if ticker == 'NVDA':
>         return 131.17
>     elif ticker == 'None':
> ```

Huh?
___
*[`tests/utils.py` line 558 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/436#-OGkFr113HilF1r189ak:-OGkFr113HilF1r189al:bcihp0y) ([raw file](https://github.com/pixeltable/pixeltable/blob/624eee24fed02f9c14df51d075b96f5c6c3f8659/tests/utils.py#L558)):*
> ```Python
>         return None
>     else:
>         # Return 0.0 instead of None, to distinguish between these two cases: the tool not being called, and the tool
> ```

Is this baking in Anthropic's 'not called' semantics?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/436)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

___
*[`tests/utils.py` line 558 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/436#-OGkFr113HilF1r189ak:-OGl8R5G7Smq7nOnnFoQ:bf6yisq) ([raw file](https://github.com/pixeltable/pixeltable/blob/624eee24fed02f9c14df51d075b96f5c6c3f8659/tests/utils.py#L558)):*

Is this baking in Anthropic's 'not called' semantics?
</blockquote></details>

This line isn't, but the preceding line (`elif ticker == 'None'`) kind of is. I wasn't sure how best to handle that one.

I'm not sure that it's part of their official semantics, versus being an artifact of how the model generates its responses. OpenAI seems to do the same thing btw.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/436)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @mkornacker)

___
*[`tests/utils.py` line 555 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/436#-OGkFljA9WLXgzjWaPtP:-OGlaK-Q4ATk1obFrpeV:bt2vr1b) ([raw file](https://github.com/pixeltable/pixeltable/blob/624eee24fed02f9c14df51d075b96f5c6c3f8659/tests/utils.py#L555)):*

Huh?
</blockquote></details>

Fixed.
___
*[`tests/utils.py` line 558 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/436#-OGkFr113HilF1r189ak:-OGlaJ3_CznfNdPdOfTG:b1b7bz3) ([raw file](https://github.com/pixeltable/pixeltable/blob/624eee24fed02f9c14df51d075b96f5c6c3f8659/tests/utils.py#L558)):*

This line isn't, but the preceding line (`elif ticker == 'None'`) kind of is. I wasn't sure how best to handle that one.

I'm not sure that it's part of their official semantics, versus being an artifact of how the model generates its responses. OpenAI seems to do the same thing btw.
</blockquote></details>

Fixed this.

pixeltable/func/tools.py
```diff
@@ -122,23 +122,32 @@ def choice(
...
-    return bool(_extract_arg(tool_calls, func_name, param_name))
+def _extract_bool_tool_arg(tool_calls: dict[str, Any], func_name: str, param_name: str) -> Optional[bool]:
+    return _extract_arg(bool, tool_calls, func_name, param_name)
+
+
+T = TypeVar('T')
+
 
-def _extract_arg(tool_calls: dict, func_name: str, param_name: str) -> Any:
+def _extract_arg(eval: Callable[[Any], T], tool_calls: dict[str, Any], func_name: str, param_name: str) -> Optional[T]:
```
mkornacker: eval: name conflicts with built-in

tests/functions/test_anthropic.py
```diff
@@ -34,7 +34,7 @@ def test_anthropic(self, reset_db) -> None:
         assert len(results['output'][0]['content'][0]['text']) > 0
         assert len(results['output2'][0]['content'][0]['text']) > 0
 
-    @pytest.mark.flaky(reruns=3)
+    @pytest.mark.flaky(reruns=4)
```
mkornacker, line 37: Why is this flaky? We really need to avoid flaky tests.

pixeltable/func/tools.py
aaron-siegel: Fixed.

tests/functions/test_anthropic.py
aaron-siegel: I agree in most cases - but in this case it's because an LLM is involved and LLM outputs are probabilistic. So if we want a test that validates that the LLM responses are what we'd expect, then it's going to fail some (low) fraction of the time. The right thing to do is build in enough retries that the probability they all fail is close to 0.

That is to say: in this case it's not that the test is flaky, it's that retries are a feature of the expected behavior of LLM tests. The decorator name `flaky` is misleading here (it's really just a retry mechanism).

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/436)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @mkornacker)

mkornacker (approved):

PR #437, opened 2025-01-14
Add a section on UDAs to the UDF feature guide

aaron-siegel:
It's not the world's most interesting example, but it demonstrates the concept


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/437)

PR #438, opened 2025-01-16
Terminology: use 'aggregate' in place of 'aggregator' in UDF docs

aaron-siegel:
Standardizing as follows: the operation is an "aggregate"; the class instance that carries out the operation is an "aggregator".


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/438)

mkornacker (approved):

PR #442, opened 2025-01-22
PXT 401/PXT-402  Add verbose system config option

amithadke:
Adding console logger.
     - For console output pixeltable now can support multiple levels INFO, ERROR, WARNING, DEBUG
     - Replaced a few print calls with new console logger call.


pixeltable/utils/console_output.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Args:
+        level_str: Level string (e.g., 'INFO', 'DEBUG')
+        default: Default level to use if conversion fails
+
+    Returns:
+        Logging level as integer
+    """
+    try:
+        return getattr(logging, level_str.upper())
```
amithadke: perhaps only support debug, info, warn anything above warn is always going to console?

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/442)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @amithadke and @mkornacker)

___
*[`pixeltable/catalog/table.py` line 365 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/442#-OHFtA8pEO3YMUzSNe8j:-OHFtA8pEO3YMUzSNe8k:b-3kmi7a) ([raw file](https://github.com/pixeltable/pixeltable/blob/d67a676c446493a5c349493c1244975e88f998ed/pixeltable/catalog/table.py#L365)):*
> ```Python
>             display(self._repr_html_())
>         else:
>             env.Env.get().console_logger.info(repr(self))
> ```

In this specific case, we really do want it to be a print statement.
___
*[`docs/api/config.md` line 38 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/442#-OHFsW_78TB8b6mdtqdD:-OHFsW_78TB8b6mdtqdE:b-r8zlbl) ([raw file](https://github.com/pixeltable/pixeltable/blob/d67a676c446493a5c349493c1244975e88f998ed/docs/api/config.md#L38)):*
> ```Markdown
> | PIXELTABLE_TIME_ZONE         | [pixeltable]<br>time_zone         | (string) Default time zone in [IANA format](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones); defaults to the system time zone |
> | PIXELTABLE_HIDE_WARNINGS     | [pixeltable]<br>hide_warnings     | (bool) Suppress warnings generated by various libraries used by Pixeltable; default is `false`                                              |
> | PIXELTABLE_VERBOSITY         | [pixeltable]<br>verbosity         | (string) Define verbosity for Pixeltable console logging DEBUG (very verbose), INFO, WARNING, ERROR (less verbose); default is `INFO`       |
> ```

Are we sure that aligning the verbosity levels with logging levels is the right thing to do (versus just 0, 1, 2, 3, say?) I'm not 100% convinced the logging levels map well to console output (and they might even confuse/mislead users into thinking they're adjusting the logging level).
___
*[`pixeltable/catalog/table_version.py` line 536 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/442#-OHFtIuj4meF969zYqzG:-OHFtIuj4meF969zYqzH:b1gzul3) ([raw file](https://github.com/pixeltable/pixeltable/blob/d67a676c446493a5c349493c1244975e88f998ed/pixeltable/catalog/table_version.py#L536)):*
> ```Python
>         )
>         Env.get().console_logger.info(msg)
>         _logger.info(f'Columns {[col.name for col in cols]}: {msg}')
> ```

There's some redundancy here in that `msg` will get put in the logs twice, but I'm not sure how best to address it (maybe we don't need to).

aaron-siegel (approved):
approved with one more comment

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/442#-:-OHKf4mJBvMhD3tZeUva:bn0z8hw)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @amithadke and @mkornacker)

___
*[`pixeltable/utils/filecache.py` line 240 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/442#-OHKemioDTOQolXIg0c9:-OHKemioDTOQolXIg0cA:b-cmikf3) ([raw file](https://github.com/pixeltable/pixeltable/blob/944283a3d7259eca086a0cadfe180ce3746f7e34/pixeltable/utils/filecache.py#L240)):*
> ```Python
>         return self.FileCacheStats(self.total_size, self.num_requests, self.num_hits, self.num_evictions, col_stats)
> 
>     def debug_print(self) -> None:
> ```

AFAIK this is not being used? I'm not sure when it was ever used. Maybe @mkornacker knows.

I'd argue that if we keep it around, these should be ordinary log debug messages, and not go to the console at all. (I'm guessing the print statement is there from a very long time ago)

amithadke:
> approved with one more comment
> 
> > _[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/442#-:-OHKf4mJBvMhD3tZeUva:bn0z8hw)_ status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @amithadke and @mkornacker)
> 
> _[`pixeltable/utils/filecache.py` line 240 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/442#-OHKemioDTOQolXIg0c9:-OHKemioDTOQolXIg0cA:b-cmikf3) ([raw file](https://github.com/pixeltable/pixeltable/blob/944283a3d7259eca086a0cadfe180ce3746f7e34/pixeltable/utils/filecache.py#L240)):_
> 
> > ```python
> >         return self.FileCacheStats(self.total_size, self.num_requests, self.num_hits, self.num_evictions, col_stats)
> > 
> >     def debug_print(self) -> None:
> > ```
> 
> AFAIK this is not being used? I'm not sure when it was ever used. Maybe @mkornacker knows.
> 
> I'd argue that if we keep it around, these should be ordinary log debug messages, and not go to the console at all. (I'm guessing the print statement is there from a very long time ago)

I will move these messages to log i think they were for our internal debugging

PR #443, opened 2025-01-22
PXT-0244: Correct validity of col error properties

WildCard932:
This change corrects the logic for determining when errormsg and errortype properties are valid.

### DIFF
--- a/pixeltable/exprs/column_ref.py
+++ b/pixeltable/exprs/column_ref.py
@@ -101,7 +101,8 @@ class ColumnRef(Expr):
         # resolve column properties
         if name == ColumnPropertyRef.Property.ERRORTYPE.name.lower() \
                 or name == ColumnPropertyRef.Property.ERRORMSG.name.lower():
-            if not (self.col.is_computed and self.col.is_stored) and not self.col.col_type.is_media_type():
+            property_is_present = self.col.is_stored and (self.col.is_computed or self.col_type.is_media_type())
+            if not property_is_present:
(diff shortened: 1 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/443)

mkornacker (approved):

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/443)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @WildCard932)

___
*[`tests/test_exprs.py` line 261 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/443#-OHFrJGSA_FZS38AvKmS:-OHFrJGT0KszsxUD5V3K:b-tllgxf) ([raw file](https://github.com/pixeltable/pixeltable/blob/6cb8c8ba891000ba8fbe33b80390df1eea355377/tests/test_exprs.py#L261)):*
> ```Python
>         with pytest.raises(excs.Error) as excinfo:
>             _ = img_t.select(img_t.c9.errortype).show()
>         assert 'only valid for' in str(excinfo.value)
> ```

Also add a test case for a stored column that is neither computed nor a media column?

WildCard932:
@aaron-siegel 
About 10 lines above the two tests I added are a set of tests that look at non-computed, non-media columns for all 4 properties.

PR #450, opened 2025-01-29
Support generalized arrays (unparameterized or with only a dtype)

aaron-siegel:
Allows `pxt.Array[Int]` types specifying only a dtype (and not a shape), as well as a bare `pxt.Array` that permits any array.

### DIFF
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -801,6 +801,8 @@ class ArrayType(ColumnType):
     def __init__(self, shape: Optional[tuple[Optional[int], ...]] = None, dtype: Optional[ColumnType] = None, nullable: bool = False):
         super().__init__(self.Type.ARRAY, nullable=nullable)
         assert shape is None or dtype is not None, (shape, dtype)  # cannot specify a shape without a dtype
+        assert dtype is None or dtype.is_int_type() or dtype.is_float_type() or dtype.is_bool_type() or dtype.is_string_type()
+
         self.shape = shape
         self.pxt_dtype = dtype
         self.dtype = None if dtype is None else dtype._type
@@ -840,8 +842,7 @@ class ArrayType(ColumnType):
             return 'Array'
         if self.shape is None:
             return f'Array[{self.pxt_dtype}]'
-        if self.dtype is None:
-            return f'Array[{self.shape}]'
+        assert self.dtype is not None
         return f'Array[{self.shape}, {self.pxt_dtype}]'
 
     @classmethod
@@ -888,11 +889,9 @@ class ArrayType(ColumnType):
                 return False
             # check that the shapes are compatible
             for n1, n2 in zip(val.shape, self.shape):
-                if n1 is None:
-                    return False
+                assert n1 is not None  # `val` must have a concrete shape
                 if n2 is None:
-                    # wildcard
-                    continue
+                    continue  # wildcard
                 if n1 != n2:
                     return False
 
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -793,12 +793,17 @@ class JsonType(ColumnType):
 
 
 class ArrayType(ColumnType):
-    def __init__(self, shape: tuple[Union[int, None], ...], dtype: ColumnType, nullable: bool = False):
+
+    shape: Optional[tuple[Optional[int], ...]]
+    pxt_dtype: Optional[ColumnType]
+    dtype: Optional[ColumnType.Type]
+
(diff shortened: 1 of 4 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/450)

pixeltable/type_system.py
```diff
@@ -793,12 +793,17 @@ def _to_base_str(self) -> str:
 
 
 class ArrayType(ColumnType):
-    def __init__(self, shape: tuple[Union[int, None], ...], dtype: ColumnType, nullable: bool = False):
+
+    shape: Optional[tuple[Optional[int], ...]]
+    pxt_dtype: Optional[ColumnType]
```
mkornacker, line 798: What do we need this for again?

pixeltable/type_system.py
```diff
@@ -855,18 +871,32 @@ def from_literal(cls, val: np.ndarray, nullable: bool = False) -> Optional[Array
...
+            return False
+
+        # If a shape is specified, check that there's a match
+        if self.shape is not None:
+            if len(val.shape) != len(self.shape):
                 return False
-        return np.issubdtype(val.dtype, self.numpy_dtype())
+            # check that the shapes are compatible
+            for n1, n2 in zip(val.shape, self.shape):
+                if n1 is None:
```
mkornacker: I don't think this is actually possible, an array instance's shape won't have Nones

tests/test_table.py
```diff
@@ -1053,7 +1065,24 @@ def test_insert(self, reset_db: None) -> None:
...
             t.insert(c5=np.ndarray((3, 2)))
-        assert 'expected ndarray((2, 3)' in str(exc_info.value)
+        assert 'expected numpy.ndarray((2, 3)' in str(exc_info.value)
+
+        # bad array literal
+        pxt.drop_table(tbl_name, if_not_exists='ignore')
+        t = pxt.create_table(tbl_name, {'c5': pxt.Array[pxt.Int]})  # type: ignore[misc]
+        with pytest.raises(excs.Error) as exc_info:
+            t.insert(c5=np.ndarray((3, 2), dtype=np.float32))
+        assert 'expected numpy.ndarray of dtype int64' in str(exc_info.value)
```
mkornacker: Start putting this into the raises()?

pixeltable/type_system.py
aaron-siegel: It's used in the `str()` implementation.

pixeltable/type_system.py
aaron-siegel: Yep I noticed the same thing and fixed it in a follow-up commit, but I think you opened the code review in between commits :)

I changed it to be an assertion

tests/test_table.py
aaron-siegel: Ok, sure. Done.

There are a whole bunch of things like this where our practices have evolved and there are inconsistent methods in the codebase. I wonder if it makes sense to make a list of them and have a techdebt blitz at some point where we standardize.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/450)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @mkornacker)

pixeltable/type_system.py
mkornacker: A brief explanatory comment would be useful. Is this just for nullability? It seems redundant, since we can regenerate the ColumnType from dtype.

(Outside the scope of this PR, but probably worth a discussion, while we're on the topic:
If it's about nulls/nans, we should probably follow numpy semantics (okay for floats, but not for ints).)

mkornacker (approved):

pixeltable/type_system.py
aaron-siegel: I added a comment.

It's also needed by copy() btw. I think the reason I put it in is that the initializer takes a ColumnType, and it seemed really silly to turn it into a ColumnType.Type only to re-cast it as a ColumnType when calling copy() or __str__(). Felt like an antipattern.

Now that ArrayType is no longer part of the public API, we could consider having it simply take a ColumnType.Type as the dtype in the initializer. I think this makes the most sense actually.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/450)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @mkornacker)

PR #451, opened 2025-01-29
PXT-405 AudioSplitter support

amithadke:
- Split audio based on presentation times, we initialize iterator with target chunk start &  end boundaries
 - Scan audio stream frame by frame which may change start & end of the chunk than what was targeted at the initialization.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 2 files reviewed, 20 unresolved discussions (waiting on @amithadke and @mkornacker)

___
*[`tests/test_audio.py` line 93 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoBv_600pF3UgySRyU:-OHoBv_600pF3UgySRyV:b-fcz616) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/tests/test_audio.py#L93)):*
> ```Python
> 
>     def test_audio_iterator_on_audio(self, reset_db) -> None:
>         pxt.drop_dir("audio_chunking", force=True)
> ```

This isn't necessary (`reset_db` will take care of it)
___
*[`tests/test_audio.py` line 94 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoBz5k3uWxZItHR0TY:-OHoBz5l7BQW-Ksg8Idy:bdsqroe) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/tests/test_audio.py#L94)):*
> ```Python
>     def test_audio_iterator_on_audio(self, reset_db) -> None:
>         pxt.drop_dir("audio_chunking", force=True)
>         pxt.create_dir("audio_chunking")
> ```

Why create this dir if you don't end up using it?
___
*[`tests/test_audio.py` line 101 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoC79U1UfNZAKCiFj1:-OHoC79U1UfNZAKCiFj2:b-rok8kx) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/tests/test_audio.py#L101)):*
> ```Python
>             "audio_chunking.audio_chunks",
>             base_t,
>             iterator = AudioIterator.create(
> ```

We need more tests for the edge cases in the arguments (overlap = 0.0, drop_incomplete_chunks=False, etc)
___
*[`tests/test_audio.py` line 108 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoCBxED-5xtofCsw_C:-OHoCBxED-5xtofCsw_D:b79mzx) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/tests/test_audio.py#L108)):*
> ```Python
>                 drop_incomplete_chunks=True
>             ))
>         print(audio_chunk_view.count())
> ```

Need to do some actual validation of the output here. (Verify we get the right number of chunks, they're the right length, etc.)
___
*[`pixeltable/iterators/audio.py` line 17 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo60e33eSbBIbUEdB8:-OHo60e33eSbBIbUEdB9:bpu242o) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L17)):*
> ```Python
> 
> 
> class AudioIterator(ComponentIterator):
> ```

I'd find `AudioSplitter` to be a more descriptive name (cf. `DocumentSplitter`)
___
*[`pixeltable/iterators/audio.py` line 18 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo64nBATu6DChrSVa-:-OHo64nBATu6DChrSVa0:brwroy3) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L18)):*
> ```Python
> 
> class AudioIterator(ComponentIterator):
>     """
> ```

Please add a full docstring explaining what the iterator does, and add usage example(s)
___
*[`pixeltable/iterators/audio.py` line 21 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo6I2K0RePYRjcNv8n:-OHo6I2K0RePYRjcNv8o:bnsr7tu) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L21)):*
> ```Python
>     Args:
>         audio: URL or path of the audio file
>         chunk_duration: Target duration for each chunk in seconds to match model requirements.
> ```

"to match model requirements" is unnecessary (and misleading - there isn't necessarily a model involved)
___
*[`pixeltable/iterators/audio.py` line 22 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo6TbDEqIbwfUcpLoq:-OHo6TbDEqIbwfUcpLor:b9wvnab) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L22)):*
> ```Python
>         audio: URL or path of the audio file
>         chunk_duration: Target duration for each chunk in seconds to match model requirements.
>         overlap: Overlap between chunks in seconds to prevent word splits at chunk boundaries.
> ```

Are we actually doing anything special in the iterator to prevent word splits? If not, we should just leave it as "Overlap between chunks, in seconds" (don't imply that the splitter does something it doesn't do)
___
*[`pixeltable/iterators/audio.py` line 23 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo6asI2j4A-xg5A8AS:-OHo6asI2j4A-xg5A8AT:b-ia41wb) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L23)):*
> ```Python
>         chunk_duration: Target duration for each chunk in seconds to match model requirements.
>         overlap: Overlap between chunks in seconds to prevent word splits at chunk boundaries.
>         min_chunk_duration: Minimum chunk duration in seconds to match model requirements.
> ```

same comment as above
___
*[`pixeltable/iterators/audio.py` line 38 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo6gzH85Rxr5WQ4DBw:-OHo6gzH85Rxr5WQ4DBx:b-wfvl30) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L38)):*
> ```Python
>     audio_start_time: int
> 
>     # List of chunks to extract each chunk is defined by start time and end time that includes overlap
> ```

specify units in the comment.
___
*[`pixeltable/iterators/audio.py` line 43 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo6m__8hNOYLIzQio8:-OHo6m__8hNOYLIzQio9:b-6cx3lk) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L43)):*
> ```Python
>     next_pos: int
> 
>     def __init__(self, audio: str, *, chunk_duration: Optional[float] = None, overlap: Optional[float] = 0.0,
> ```

Is overlap a float or an Optional[float]? If it's an Optional[float], I'd expect the default to be None, not 0.0. Same comment for the other params. [Note: I'd think the parameter type should be float, not Optional[float], with default 0.0]
___
*[`pixeltable/iterators/audio.py` line 46 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo72mM95OrreDa5OI3:-OHo72mM95OrreDa5OI4:b-mtv948) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L46)):*
> ```Python
>                  min_chunk_duration: Optional[float] = 0.0, drop_incomplete_chunks: Optional[bool] = False):
>         if chunk_duration is None:
>             raise excs.Error('chunk_duration may be specified')
> ```

Then why allow it at all? Why not simply make chunk_duration a required parameter?
___
*[`pixeltable/iterators/audio.py` line 72 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo8cfPBQP3xsulYXTB:-OHo8cfPBQP3xsulYXTC:b-41e33h) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L72)):*
> ```Python
>                 if pts_min_chunk_duration > 0 and duration < pts_min_chunk_duration and drop_incomplete_chunks:
>                     break
>                 else:
> ```

This can be simplified: instead of an if/else use a single if statement and put the break outside the conditional
___
*[`pixeltable/iterators/audio.py` line 108 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo9Lgl0n7_el12115O:-OHo9Lgl0n7_el12115P:b1ss369) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L108)):*
> ```Python
>         else:
>             target_chunk_start = self.chunks_to_extract[self.next_pos][0]
>             target_chunk_end = self.chunks_to_extract[self.next_pos][1]
> ```

Better:

```python
target_chunk_start, target_chunk_end = self.chunks_to_extract[self.next_pos]
```
___
*[`pixeltable/iterators/audio.py` line 114 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo9pzF6mgcEKmRxc0p:-OHo9pzF6mgcEKmRxc0q:bo3uhyv) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L114)):*
> ```Python
>             input_stream = self.container.streams.audio[0]
> 
>             # TODO fix this
> ```

Clarify: what's the actual TODO item? (what needs to be fixed?)
___
*[`pixeltable/iterators/audio.py` line 129 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoA068CY4yWVQlp5Yx:-OHoA068CY4yWVQlp5Yy:b-21oju5) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L129)):*
> ```Python
>                     frame = next(self.container.decode(audio=0))
>                 except EOFError:
>                     raise StopIteration
> ```

Under normal circumstances, this should never happen, right? Because we've already calculated the expected number of chunks.

If that's the case, is raising a StopIteration the right thing to do or might that just result in swallowed errors? Would it be better to raise an exception saying the file has corrupted metadata?
___
*[`pixeltable/iterators/audio.py` line 143 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoB5GABEQHFkknY3Qp:-OHoB5GABEQHFkknY3Qq:bs3fz56) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L143)):*
> ```Python
>                 # write frame
>                 packet = output_stream.encode(frame)
>                 if packet:
> ```

`if packet is not None` is the preferred style here.

What does it mean for it to be `None`? Are we sure we're handling that case correctly? Add a code comment explaining
___
*[`pixeltable/iterators/audio.py` line 148 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoBSr63Pi9qXGhEZGX:-OHoBSr63Pi9qXGhEZGY:b-lnuhf7) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L148)):*
> ```Python
>                     # we have more data to collect
>                     continue
>                 # record result
> ```

This and below should go outside the main loop.
___
*[`pixeltable/iterators/audio.py` line 151 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoBcv-5FRUVtMUCiS8:-OHoBcv-5FRUVtMUCiS9:bwqmrjo) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L151)):*
> ```Python
>                 chunk_end_pts = frame_end
>                 if frame_count > 0:
>                     packet = output_stream.encode(None)
> ```

What does this do? (Add code comment)
___
*[`pixeltable/iterators/audio.py` line 166 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo90hj6CDwcLJZ22-H:-OHo90hj6CDwcLJZ22-I:b-yek5jj) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L166)):*
> ```Python
>                 return result
> 
>     def _create_chunk_file(self) -> str:
> ```

Precede with a double underscore (`__create_chunk_file(self)`) since this is a private method

amithadke (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 2 files reviewed, 20 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`tests/test_audio.py` line 101 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoC79U1UfNZAKCiFj1:-OHoG8Et0VKeWXhBkWIB:b-gg5q8u) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/tests/test_audio.py#L101)):*

We need more tests for the edge cases in the arguments (overlap = 0.0, drop_incomplete_chunks=False, etc)
</blockquote></details>

im adding more tests 
___
*[`tests/test_audio.py` line 108 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoCBxED-5xtofCsw_C:-OHoGONn6ezWO4VEQoC-:b-3dunn9) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/tests/test_audio.py#L108)):*

Need to do some actual validation of the output here. (Verify we get the right number of chunks, they're the right length, etc.)
</blockquote></details>

yes dont want to hard code number of chunks since its just going over all files in a dir. adding seperate test for a single file

amithadke (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 2 files reviewed, 20 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/iterators/audio.py` line 46 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo72mM95OrreDa5OI3:-OHoIB4q6LEC15MGaFGO:bgubdlw) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L46)):*

Then why allow it at all? Why not simply make chunk_duration a required parameter?
</blockquote></details>

forgot about required parameter, will remove the check
___
*[`pixeltable/iterators/audio.py` line 114 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo9pzF6mgcEKmRxc0p:-OHoHdnJ7NWWR-bBgQBe:b-tst9jn) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L114)):*

Clarify: what's the actual TODO item? (what needs to be fixed?)
</blockquote></details>

one of the test example audio file has mp3float codec which throws exception codec not found but after searching online I found that just using mp3 instead may work. I need to find out if there is a better way
___
*[`pixeltable/iterators/audio.py` line 129 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoA068CY4yWVQlp5Yx:-OHoH5k_84bjDW6pT2eN:b5xfe7p) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L129)):*

Under normal circumstances, this should never happen, right? Because we've already calculated the expected number of chunks.

If that's the case, is raising a StopIteration the right thing to do or might that just result in swallowed errors? Would it be better to raise an exception saying the file has corrupted metadata?
</blockquote></details>

not sure specially if we are going over bunch of files, let me think about it

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 28 unresolved discussions (waiting on @aaron-siegel and @amithadke)

___
*[`pixeltable/catalog/table_version.py` line 476 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIQBYj309g8ATwy-Cw:-OIIQBYj309g8ATwy-Cx:b-f4kck6) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/catalog/table_version.py#L476)):*
> ```Python
>             # if the column is intentionally not stored, we want to avoid the overhead of an index
>             return None
>         # Skip index for stored media columns produced by an iterator
> ```

Turn this into a separate check, the if clause is now hard to read. Also, the comment conflicts with the comment 3 lines below.
___
*[`pixeltable/iterators/audio.py` line 20 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIQh7J6rt690-3NnPs:-OIIQh7J6rt690-3NnPt:b-dgj1ug) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L20)):*
> ```Python
> class AudioSplitter(ComponentIterator):
>     """
>     Iterator over audio data and split it in the chunks of audio clips duration of which is provided in chunk_duration_sec.
> ```

Garbled grammar. This is going to show up in our API docs, so it's worth making it consumable.
___
*[`pixeltable/iterators/audio.py` line 25 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRA3h30AXaMqZ8uCA:-OIIRA3h30AXaMqZ8uCB:b-anrjwh) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L25)):*
> ```Python
> 
>     Args:
>         audio: URL or path of the audio file
> ```

It is not an url.
___
*[`pixeltable/iterators/audio.py` line 29 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRN-PE9mgDXz1SWYY:-OIIRN-PE9mgDXz1SWYZ:bz7wj8p) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L29)):*
> ```Python
>         overlap_sec: Overlap between chunks in seconds.
>         min_chunk_duration_sec: Minimum chunk duration in seconds.
>         drop_incomplete_chunks: Drop chunks smaller than min_chunk_duration at the end of the audio.
> ```

There can only be one incomplete chunk, no? Switch to singular.
___
*[`pixeltable/iterators/audio.py` line 42 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRjEj-_fEp1w8JIyW:-OIIRjEj-_fEp1w8JIyX:btqps5d) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L42)):*
> ```Python
>     container: av.container.input.InputContainer
>     audio_time_base: Fraction
>     audio_start_time: int
> ```

What's the unit?
___
*[`pixeltable/iterators/audio.py` line 82 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIISeu28FCbzHqYPIef:-OIISeu28FCbzHqYPIeg:b-pi7png) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L82)):*
> ```Python
>         self.min_chunk_duration_sec = min_chunk_duration_sec
>         self.drop_incomplete_chunks = drop_incomplete_chunks
>         self.audio_time_base = self.container.streams.audio[0].time_base
> ```

Do we know for sure that there is an audio stream?
___
*[`pixeltable/iterators/audio.py` line 106 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIT04V6GfQgGaSmdMG:-OIIT04V6GfQgGaSmdMH:b-6ekhie) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L106)):*
> ```Python
>     @classmethod
>     def build_chunks(
>         cls, start_time, total_duration, chunk_duration, overlap, min_chunk_duration, drop_incomplete_chunks
> ```

missing types
___
*[`pixeltable/iterators/audio.py` line 117 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIITW6aBolICihiTvrg:-OIITW6aBolICihiTvrh:bc69kzk) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L117)):*
> ```Python
>                 break
>             else:
>                 chunks_to_extract.append((i, i + chunk_duration + overlap))
> ```

This looks wrong. The chunks need to be chunk_duration long. The overlap reduces the distance between two consecutive chunks.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 29 unresolved discussions (waiting on @aaron-siegel and @amithadke)

___
*[`pixeltable/iterators/audio.py` line 198 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIV6CwEKgDihsISBMZ:-OIIV6CwEKgDihsISBM_:b-vomwtq) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L198)):*
> ```Python
> 
>     def __create_chunk_file(self) -> str:
>         return str(env.Env.get().tmp_dir / f'{uuid.uuid4()}{self.audio_path.suffix}')
> ```

Probably be better to inline it - there's only one call, and you're just adding another level of indirection.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 20 unresolved discussions (waiting on @amithadke and @mkornacker)

___
*[`tests/test_audio.py` line 93 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIbHu80k0fKrpU5j2X:-OIIbHu80k0fKrpU5j2Y:b3v7i91) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/tests/test_audio.py#L93)):*
> ```Python
>         }
> 
>     def __get_chunk_counts(self, files, target_chunk_size_sec) -> dict[str, int]:
> ```

I'd find it slightly more logical for this function to take a single file as input (`__get_chunk_count` rather than `__get_chunk_counts`) and use a collection comprehension to do it for multiple files in the caller:
```python
file_to_chunks = {file: __get_chunk_count(file) for file in files}
```
___
*[`tests/test_audio.py` line 106 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIajjO7CLgnQ90dLhQ:-OIIajjO7CLgnQ90dLhR:b-lvkr1r) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/tests/test_audio.py#L106)):*
> ```Python
>         return file_to_chunk_count
> 
>     def test_audio_iterator_on_audio(self, reset_db) -> None:
> ```

Add a ReloadTester to test a persist/reload cycle for the AudioSplitter metadata
___
*[`tests/test_audio.py` line 126 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIdsIrB7OqjBQ64sj_:-OIIdsIrB7OqjBQ64sja:b-srj1sb) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/tests/test_audio.py#L126)):*
> ```Python
>         file_to_chunks_from_view: dict[str, int] = {}
>         for result in results:
>             file_to_chunks_from_view[result['audio']] = file_to_chunks_from_view.get(result['audio'], 0) + 1
> ```

Python tip - You can do this in 1-line (and more efficiently) with
```python
file_to_chunks_from_view = dict(Counter(result['audio'] for result in results))
```
(there is another instance of this below)
___
*[`docs/api/api-cheat-sheet.md` line 63 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OICdZ83A5K92rHizwn9:-OICdZ83A5K92rHizwnA:b-j0qmh2) ([raw file](https://github.com/pixeltable/pixeltable/blob/cb36e3cc48d2d675bd5084a0f7d7096bf023b04a/docs/api/api-cheat-sheet.md#L63)):*
> ```Markdown
> audio_chunk_view = pxt.create_view("audio_chunks", video_t, iterator=AudioSplitter.create(audio=video_t.audio, chunk_duration_sec=5.0))
> ```
> `chunk_duration_sec` is time duration of an audio chunk in seconds
> ```

`s/is time duration/is the time duration/`
___
*[`pixeltable/iterators/audio.py` line 72 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIWWiI8WCOg43gHtoC:-OIIWWiI8WCOg43gHtoD:b-mlyc2z) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L72)):*
> ```Python
>         if chunk_duration_sec <= 0.0:
>             raise excs.Error('chunk_duration_sec must be a positive number')
>         if min_chunk_duration_sec is not None and chunk_duration_sec < min_chunk_duration_sec:
> ```

We can get rid of the `is not None` check now (it's no longer Optional)
___
*[`pixeltable/iterators/audio.py` line 85 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIXRWRC9435uNYOVWP:-OIIXRWRC9435uNYOVWQ:b-jf3pcu) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L85)):*
> ```Python
>         self.audio_start_time = self.container.streams.audio[0].start_time or 0
> 
>         total_audio_duration = self.container.streams.audio[0].duration or 0
> ```

Is this a `pts` as well?
___
*[`pixeltable/iterators/audio.py` line 86 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIWn2h8tCq5Gs2w6V0:-OIIWn2h8tCq5Gs2w6V1:b-ndnurk) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L86)):*
> ```Python
> 
>         total_audio_duration = self.container.streams.audio[0].duration or 0
>         pts_chunk_duration = int(self.chunk_duration_sec / self.audio_time_base)
> ```

I'd find it more natural to call these `chunk_duration_pts` etc. (to be consistent with `chunk_duration_sec` etc)
___
*[`pixeltable/iterators/audio.py` line 106 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIXFDz3z5Eava4uBHJ:-OIIXFDz3z5Eava4uBHK:bxg24qk) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L106)):*
> ```Python
>     @classmethod
>     def build_chunks(
>         cls, start_time, total_duration, chunk_duration, overlap, min_chunk_duration, drop_incomplete_chunks
> ```

I like that you used units for the arguments; maybe do the same for the parameters? `total_duration_pts` etc.

Also, per our conversation today: right now, you're rounding to pts first, then computing intervals. It'll be slightly more accurate to compute intervals first (using floats), then round to pts. See the code in FrameIterator for an example.
___
*[`pixeltable/iterators/audio.py` line 113 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OII_J37BXhYJVSYUGqa:-OII_J37BXhYJVSYUGqb:bvvwav0) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L113)):*
> ```Python
>             if i + chunk_duration + overlap >= total_duration:
>                 duration = total_duration - i
>                 if not (drop_incomplete_chunks and min_chunk_duration > 0 and duration < min_chunk_duration):
> ```

I think the `min_chunk_duration > 0` check is superfluous (duration can never be negative)
___
*[`pixeltable/iterators/audio.py` line 141 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OII_sUS4NXfRn1yxRl_:-OII_sUS4NXfRn1yxRla:bc2v6mn) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L141)):*
> ```Python
>         if self.next_pos >= len(self.chunks_to_extract):
>             raise StopIteration
>         else:
> ```

Unnecessary to put this in an `else:` block, since the `if` clause raises an exception.
___
*[`mkdocs.yml` line 31 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIU1112jHj8GDFkoff:-OIIU1112jHj8GDFkofg:bjtu7qv) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/mkdocs.yml#L31)):*
> ```YAML
>       - FrameIterator: pixeltable/iterators/frame-iterator.md
>       - TileIterator: pixeltable/iterators/tile-iterator.md
>       - AudioSplitter: pixeltable/iterators/audio-splitter.md
> ```

put in alphabetical order for ease of navigation

amithadke (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 21 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`mkdocs.yml` line 31 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIU1112jHj8GDFkoff:-OIT50pM1kQwTutU3Qxm:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/mkdocs.yml#L31)):*

put in alphabetical order for ease of navigation
</blockquote></details>

Done.
___
*[`docs/api/api-cheat-sheet.md` line 63 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OICdZ83A5K92rHizwn9:-OIT51sX0CojEgfb0Fvr:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/cb36e3cc48d2d675bd5084a0f7d7096bf023b04a/docs/api/api-cheat-sheet.md#L63)):*

`s/is time duration/is the time duration/`
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table_version.py` line 476 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIQBYj309g8ATwy-Cw:-OIT4nGQB-J92CRaVjfy:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/catalog/table_version.py#L476)):*

Turn this into a separate check, the if clause is now hard to read. Also, the comment conflicts with the comment 3 lines below.
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 38 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo6gzH85Rxr5WQ4DBw:-OIT5R8D2wGyjBpNUtOk:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L38)):*

specify units in the comment.
</blockquote></details>

done
___
*[`pixeltable/iterators/audio.py` line 43 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo6m__8hNOYLIzQio8:-OIT5XobEuF0oNjPM_cU:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L43)):*

Is overlap a float or an Optional[float]? If it's an Optional[float], I'd expect the default to be None, not 0.0. Same comment for the other params. [Note: I'd think the parameter type should be float, not Optional[float], with default 0.0]
</blockquote></details>

done
___
*[`pixeltable/iterators/audio.py` line 151 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoBcv-5FRUVtMUCiS8:-OIT6Dd18WykiiSTQjnM:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L151)):*

What does this do? (Add code comment)
</blockquote></details>

done
___
*[`pixeltable/iterators/audio.py` line 166 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHo90hj6CDwcLJZ22-H:-OIT5th5-_5pRs5WA4ve:bxdal1b) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/pixeltable/iterators/audio.py#L166)):*

Precede with a double underscore (`__create_chunk_file(self)`) since this is a private method
</blockquote></details>

not need made the function inline
___
*[`pixeltable/iterators/audio.py` line 20 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIQh7J6rt690-3NnPs:-OIT53bkBs24bEuyjCAK:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L20)):*

Garbled grammar. This is going to show up in our API docs, so it's worth making it consumable.
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 25 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRA3h30AXaMqZ8uCA:-OIT56qxDirrXzmpQoj3:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L25)):*

It is not an url.
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 29 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRN-PE9mgDXz1SWYY:-OIT59AR-MvqukJTfscR:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L29)):*

There can only be one incomplete chunk, no? Switch to singular.
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 42 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRjEj-_fEp1w8JIyW:-OIT5C1t7hdDwPw_4aHh:bcgh0lb) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L42)):*

What's the unit?
</blockquote></details>

for fraction? 
___
*[`pixeltable/iterators/audio.py` line 72 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIWWiI8WCOg43gHtoC:-OIT5ZkV7RWZy-0cO3pr:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L72)):*

We can get rid of the `is not None` check now (it's no longer Optional)
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 82 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIISeu28FCbzHqYPIef:-OIT5_mI2Uwzl8gGN-CD:b-7ql1td) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L82)):*

Do we know for sure that there is an audio stream?
</blockquote></details>

fixed
___
*[`pixeltable/iterators/audio.py` line 85 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIXRWRC9435uNYOVWP:-OIT5e1r1a4YL3rGSfwQ:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L85)):*

Is this a `pts` as well?
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 86 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIWn2h8tCq5Gs2w6V0:-OIT5epWAQYb3TpxTiZO:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L86)):*

I'd find it more natural to call these `chunk_duration_pts` etc. (to be consistent with `chunk_duration_sec` etc)
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 106 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIT04V6GfQgGaSmdMG:-OIT5gedCbQwQG-8pKH2:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L106)):*

missing types
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 106 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIXFDz3z5Eava4uBHJ:-OIT5hmHEURKO6OHognU:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L106)):*

I like that you used units for the arguments; maybe do the same for the parameters? `total_duration_pts` etc.

Also, per our conversation today: right now, you're rounding to pts first, then computing intervals. It'll be slightly more accurate to compute intervals first (using floats), then round to pts. See the code in FrameIterator for an example.
</blockquote></details>

done
___
*[`pixeltable/iterators/audio.py` line 113 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OII_J37BXhYJVSYUGqa:-OIT5mSz969XPxq8eDG5:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L113)):*

I think the `min_chunk_duration > 0` check is superfluous (duration can never be negative)
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 117 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIITW6aBolICihiTvrg:-OIT5mu7DJbOVI4UoMpB:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L117)):*

This looks wrong. The chunks need to be chunk_duration long. The overlap reduces the distance between two consecutive chunks.
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 141 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OII_sUS4NXfRn1yxRl_:-OIT5ntR74XGM3rNkLMo:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L141)):*

Unnecessary to put this in an `else:` block, since the `if` clause raises an exception.
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 198 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIV6CwEKgDihsISBMZ:-OIT5zfMCVsOUFGXicsX:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L198)):*

Probably be better to inline it - there's only one call, and you're just adding another level of indirection.
</blockquote></details>

Done.
___
*[`tests/test_audio.py` line 93 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OHoBv_600pF3UgySRyU:-OIT66-iAxuwTtH-7RVV:b-z7hui9) ([raw file](https://github.com/pixeltable/pixeltable/blob/cfbaeb33c6b74daf8a26de49b9378ad6dec4e12c/tests/test_audio.py#L93)):*

This isn't necessary (`reset_db` will take care of it)
</blockquote></details>

removed create_dir
___
*[`tests/test_audio.py` line 93 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIbHu80k0fKrpU5j2X:-OIT601pEgqEBPZjYM3L:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/tests/test_audio.py#L93)):*

I'd find it slightly more logical for this function to take a single file as input (`__get_chunk_count` rather than `__get_chunk_counts`) and use a collection comprehension to do it for multiple files in the caller:
```python
file_to_chunks = {file: __get_chunk_count(file) for file in files}
```
</blockquote></details>

Done.
___
*[`tests/test_audio.py` line 106 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIajjO7CLgnQ90dLhQ:-OIT6Ab03rcK1mqaCh_t:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/tests/test_audio.py#L106)):*

Add a ReloadTester to test a persist/reload cycle for the AudioSplitter metadata
</blockquote></details>

Done.
___
*[`tests/test_audio.py` line 126 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIdsIrB7OqjBQ64sj_:-OIT6BD5570H4kNhiROt:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/tests/test_audio.py#L126)):*

Python tip - You can do this in 1-line (and more efficiently) with
```python
file_to_chunks_from_view = dict(Counter(result['audio'] for result in results))
```
(there is another instance of this below)
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 168 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIT5rPI7DJ7xUCOAA-F:-OIT5rPI7DJ7xUCOAA-G:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L168)):*
> ```Python
>                 chunk_start_pts = frame.pts
>             # Write frame to output container
>             frame_count += 1
> ```

done

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 22 unresolved discussions (waiting on @amithadke and @mkornacker)

___
*[`docs/api/api-cheat-sheet.md` line 63 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OICdZ83A5K92rHizwn9:-OIShylp-thhUkKuob-2:b-zdvn8r) ([raw file](https://github.com/pixeltable/pixeltable/blob/cb36e3cc48d2d675bd5084a0f7d7096bf023b04a/docs/api/api-cheat-sheet.md#L63)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

doesn't seem to be done; the article ("the") is still missing
___
*[`pixeltable/iterators/audio.py` line 42 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRjEj-_fEp1w8JIyW:-OIWax6z-YeMCUJDfW1U:b2yc3k0) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L42)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

for fraction?
</blockquote></details>

seconds per pts, right? i'm not sure that needs to be in the variable name, but maybe a comment would be useful.
___
*[`pixeltable/iterators/audio.py` line 62 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWbNqa2rmlg50O50C2:-OIWbNqa2rmlg50O50C3:b-fuxwpe) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L62)):*
> ```Python
>     ):
>         if chunk_duration_sec <= 0.0:
>             raise excs.Error('chunk_duration_sec must be a positive number')
> ```

Best practice is to have unit tests that capture these error conditions.
___
*[`pixeltable/iterators/audio.py` line 83 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWenj14xBiUq6BmNSV:-OIWenj14xBiUq6BmNSW:bcc5h65) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L83)):*
> ```Python
>         total_audio_duration_sec = float(total_audio_duration_pts * self.audio_time_base)
> 
>         self.chunks_to_extract = [
> ```

It's confusing to have two variables in different places both called `chunks_to_extract`, but with different units. Use different variable names or add units to clarify.
___
*[`pixeltable/iterators/audio.py` line 84 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWcFxv1Nj7pNrM-yCm:-OIWcFxv1Nj7pNrM-yCn:b4gw09z) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L84)):*
> ```Python
> 
>         self.chunks_to_extract = [
>             (math.floor(start / self.audio_time_base), math.ceil(end / self.audio_time_base))
> ```

Gotta be careful here. Using floor on one end and ceil on the other means we may get misaligned endpoints between chunks.

I recommend using round() on both ends.
___
*[`pixeltable/iterators/audio.py` line 105 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWco5XD7Qi3HriT4gk:-OIWco5XD7Qi3HriT4gl:bbbyted) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L105)):*
> ```Python
>         current_pos = start_time_sec
>         while current_pos < total_duration_sec:
>             chunk_start = max(start_time_sec, current_pos - overlap_sec)
> ```

Won't this result in a short initial segment? e.g. if chunk duration is 10 seconds and overlap is 1 second, the first chunk will be 9 seconds, not 10.

Better to just put `chunk_start = current_pos`, `chunk_end = current_pos + chunk_duration_sec` and then apply overlap in the _increment_: `current_pos = chunk_end - overlap_sec`.
___
*[`pixeltable/iterators/audio.py` line 106 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdUYXAzcVbc7dLHqP:-OIWdUYY4GCdr9YLavY0:b4atnbg) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L106)):*
> ```Python
>         while current_pos < total_duration_sec:
>             chunk_start = max(start_time_sec, current_pos - overlap_sec)
>             chunk_end = min(chunk_start + chunk_duration_sec, total_duration_sec)
> ```

Are we sure this is correct in cases where `start_time_sec > 0`? i.e., is `total_duration_sec` always the same as `end_time` or do we need to adjust by `start_time_sec`?
___
*[`pixeltable/iterators/audio.py` line 112 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdmOGCx7eoua7WtH-:-OIWdmOGCx7eoua7WtH0:bmj06z2) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L112)):*
> ```Python
>         if (
>             len(chunks_to_extract) > 0
>             and float(chunks_to_extract[-1][1] - chunks_to_extract[-1][0] - overlap_sec) < min_chunk_duration_sec
> ```

Why cast to a float here? Aren't they already floats?

Also, why is `overlap_sec` being subtracted? It doesn't affect the chunk size.
___
*[`pixeltable/iterators/audio.py` line 114 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdraBB6VJiAwwDzPo:-OIWdraBB6VJiAwwDzPp:b4lgdxx) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L114)):*
> ```Python
>             and float(chunks_to_extract[-1][1] - chunks_to_extract[-1][0] - overlap_sec) < min_chunk_duration_sec
>         ):
>             return chunks_to_extract[:-1]
> ```

Add code comment explaining what's happening

amithadke (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 22 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`docs/api/api-cheat-sheet.md` line 63 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OICdZ83A5K92rHizwn9:-OIWgzVv0F_l7Df3pQK6:b9w8yqd) ([raw file](https://github.com/pixeltable/pixeltable/blob/cb36e3cc48d2d675bd5084a0f7d7096bf023b04a/docs/api/api-cheat-sheet.md#L63)):*

doesn't seem to be done; the article ("the") is still missing
</blockquote></details>

will fix it, i thought i did may be got lost on my side.
___
*[`pixeltable/iterators/audio.py` line 42 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRjEj-_fEp1w8JIyW:-OIWh9JL4uA9-rfXol6W:bfir0t4) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L42)):*

seconds per pts, right? i'm not sure that needs to be in the variable name, but maybe a comment would be useful.
</blockquote></details>

will add a comment
___
*[`pixeltable/iterators/audio.py` line 84 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWcFxv1Nj7pNrM-yCm:-OIWhrJa0lVnrneSMVaS:b-2jwrs) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L84)):*

Gotta be careful here. Using floor on one end and ceil on the other means we may get misaligned endpoints between chunks.

I recommend using round() on both ends.
</blockquote></details>

i think floor and ceil was for last implementation since two chunks boundaries were close for example 
0-10 (+overlap) and 10-20 (+overlap) . Hence used floor and ceil to make sure end of last chunk and start of next chunk look correct. We don't need this in current implementation, will use round
___
*[`pixeltable/iterators/audio.py` line 105 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWco5XD7Qi3HriT4gk:-OIWiesMAbL5J9QHbyi4:b7v9xtz) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L105)):*

Won't this result in a short initial segment? e.g. if chunk duration is 10 seconds and overlap is 1 second, the first chunk will be 9 seconds, not 10.

Better to just put `chunk_start = current_pos`, `chunk_end = current_pos + chunk_duration_sec` and then apply overlap in the _increment_: `current_pos = chunk_end - overlap_sec`.
</blockquote></details>

not sure if this will happen we have tests to cover this. 
chunk start will be 0  with over lap of 1 max(0, 0 - 1) will be 0
chunk end will be 0 +10 = 10
___
*[`pixeltable/iterators/audio.py` line 106 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdUYXAzcVbc7dLHqP:-OIWl7-xCiElPKICwYEB:b3ll6ms) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L106)):*

Are we sure this is correct in cases where `start_time_sec > 0`? i.e., is `total_duration_sec` always the same as `end_time` or do we need to adjust by `start_time_sec`?
</blockquote></details>

total_duration_sec is same as end_time start time is where first audio frame starts at in terms of presentation time.
___
*[`pixeltable/iterators/audio.py` line 112 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdmOGCx7eoua7WtH-:-OIWlY5K8IofN2GjR7pj:bjxcz1p) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L112)):*

Why cast to a float here? Aren't they already floats?

Also, why is `overlap_sec` being subtracted? It doesn't affect the chunk size.
</blockquote></details>

Thanks need to remove float, this list used to be ints before.

amithadke (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 22 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`docs/api/api-cheat-sheet.md` line 63 at r4](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OICdZ83A5K92rHizwn9:-OIXfpdMDX_CfRE6oxPK:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/cb36e3cc48d2d675bd5084a0f7d7096bf023b04a/docs/api/api-cheat-sheet.md#L63)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

will fix it, i thought i did may be got lost on my side.
</blockquote></details>

done
___
*[`pixeltable/iterators/audio.py` line 42 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRjEj-_fEp1w8JIyW:-OIXftis7Ahv7g1hJjxz:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L42)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

will add a comment
</blockquote></details>

done
___
*[`pixeltable/iterators/audio.py` line 62 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWbNqa2rmlg50O50C2:-OIXfxf12u6OgbYKzhmr:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L62)):*

Best practice is to have unit tests that capture these error conditions.
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 83 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWenj14xBiUq6BmNSV:-OIgIcF01z3MCSeKsGs5:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L83)):*

It's confusing to have two variables in different places both called `chunks_to_extract`, but with different units. Use different variable names or add units to clarify.
</blockquote></details>

Done.
___
*[`pixeltable/iterators/audio.py` line 84 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWcFxv1Nj7pNrM-yCm:-OIXgyNa9DGn4UcvIDSU:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L84)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

i think floor and ceil was for last implementation since two chunks boundaries were close for example 
0-10 (+overlap) and 10-20 (+overlap) . Hence used floor and ceil to make sure end of last chunk and start of next chunk look correct. We don't need this in current implementation, will use round
</blockquote></details>

done
___
*[`pixeltable/iterators/audio.py` line 105 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWco5XD7Qi3HriT4gk:-OIXhNg4A_5VYJJsOHFb:b-kzdxa8) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L105)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

not sure if this will happen we have tests to cover this. 
chunk start will be 0  with over lap of 1 max(0, 0 - 1) will be 0
chunk end will be 0 +10 = 10
</blockquote></details>

made the change have to add check in while to stop when chunk_end crosses total time duration
___
*[`pixeltable/iterators/audio.py` line 106 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdUYXAzcVbc7dLHqP:-OIXhQegAC0j4HYwoygj:b5quv29) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L106)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

total_duration_sec is same as end_time start time is where first audio frame starts at in terms of presentation time.
</blockquote></details>

I take it back, fixed
___
*[`pixeltable/iterators/audio.py` line 112 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdmOGCx7eoua7WtH-:-OIXhbrV2E17Ix7lx3DO:bb7f2mf) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L112)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

Thanks need to remove float, this list used to be ints before.
</blockquote></details>

done
___
*[`pixeltable/iterators/audio.py` line 114 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdraBB6VJiAwwDzPo:-OIgIhYB9vqut04Jj7oD:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L114)):*

Add code comment explaining what's happening
</blockquote></details>

Done.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 17 unresolved discussions (waiting on @amithadke and @mkornacker)

___
*[`pixeltable/iterators/audio.py` line 42 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRjEj-_fEp1w8JIyW:-OIm-OlEArDoqZEn6xIu:by5pxoz) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L42)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

done
</blockquote></details>

it's not samples per second, it's seconds per pts, unless i'm missing something. but i see this below,
```python
audio_start_time_sec = float(audio_start_time_pts * self.audio_time_base)
```
which implies that seconds should be in the numerator.
___
*[`pixeltable/iterators/audio.py` line 112 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdmOGCx7eoua7WtH-:-OIm0Idf1ZZ9n5l3xV5R:b1v6mva) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L112)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

done
</blockquote></details>

You didn't address the second question. (Why is overlap_sec being subtracted?)
___
*[`tests/test_audio.py` line 151 at r7](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIm0fAB2Gi7trrwpTPw:-OIm0fAB2Gi7trrwpTPx:b-dk815t) ([raw file](https://github.com/pixeltable/pixeltable/blob/450de83e0c9e4b77299b39bbba5a98a8a0e4089a/tests/test_audio.py#L151)):*
> ```Python
>         for file, count in file_to_chunks.items():
>             assert count == file_to_chunks_from_view.get(file, 0)
> ```

You need to add `reload_tester.run_reload_tests()` at the end of the function or it won't do anything. You can use the other tests that use a reload_tester as a template

amithadke (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 17 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/iterators/audio.py` line 42 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRjEj-_fEp1w8JIyW:-OIm5QHyBiD68RjDqAJ-:bin1w6i) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L42)):*

it's not samples per second, it's seconds per pts, unless i'm missing something. but i see this below,
```python
audio_start_time_sec = float(audio_start_time_pts * self.audio_time_base)
```
which implies that seconds should be in the numerator.
</blockquote></details>

its 1/sample rate so numerator always seems to be 1 and denominator is samples per sec
will change it
___
*[`pixeltable/iterators/audio.py` line 112 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdmOGCx7eoua7WtH-:-OIm46JQ-3Sbf8kWuTZ9:b-5cgbkn) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L112)):*

You didn't address the second question. (Why is overlap_sec being subtracted?)
</blockquote></details>

chunks_to_extract[-1][0] is start of last chunk which includes overlap  We want to include last chunk only if its actual duration is greater than min duration. (for example if over lap is 2s, chunk size is 5s and min chunk is 2s then for 21s file we will have 
0-5, 3-8, 6-11, 9-14, 12-17, 15-20 and last chunk in the list will be 18-21. In this case true size of chunk is just 1s but with overlap its 3s , if we don't subtract overlap chunk 18-21s gets included)
___
*[`tests/test_audio.py` line 151 at r7](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIm0fAB2Gi7trrwpTPw:-OIm6wE--rMwh-DYMWU5:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/450de83e0c9e4b77299b39bbba5a98a8a0e4089a/tests/test_audio.py#L151)):*

You need to add `reload_tester.run_reload_tests()` at the end of the function or it won't do anything. You can use the other tests that use a reload_tester as a template
</blockquote></details>

Done.

amithadke (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 17 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/iterators/audio.py` line 112 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdmOGCx7eoua7WtH-:-OImXTn41eAJrEz5A1Jb:b-te528m) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L112)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

chunks_to_extract[-1][0] is start of last chunk which includes overlap  We want to include last chunk only if its actual duration is greater than min duration. (for example if over lap is 2s, chunk size is 5s and min chunk is 2s then for 21s file we will have 
0-5, 3-8, 6-11, 9-14, 12-17, 15-20 and last chunk in the list will be 18-21. In this case true size of chunk is just 1s but with overlap its 3s , if we don't subtract overlap chunk 18-21s gets included)
</blockquote></details>

let me now if this not desirable. @mkornacker

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 16 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/iterators/audio.py` line 112 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdmOGCx7eoua7WtH-:-OInQANGANs491HduTuq:b-u67hem) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L112)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

let me now if this not desirable. @mkornacker
</blockquote></details>

I would argue that in your example, the 18-21s chunk should be included: it's a 3-second chunk, not a 1-second chunk. Filtering it out in this case seems counterintuitive to me. If we call the 18-21s chunk a 1-second chunk, wouldn't we also have to call (say) the 9-14s chunk a 3-second chunk because of the overlap (which conflicts with our terminology)? I'd suggest we use "chunk size" consistently to mean the actual duration of the chunk.

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 15 unresolved discussions (waiting on @mkornacker)

mkornacker (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 15 unresolved discussions (waiting on @aaron-siegel and @amithadke)

___
*[`pixeltable/iterators/audio.py` line 25 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRA3h30AXaMqZ8uCA:-OIr6WoV2GAfjIAWs91o:bufvvlr) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L25)):*
<details><summary><i>Previously, amithadke (amit hadke) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Looking at the docstring for extract_audio() and DocumentSplitter, we actually don't list the main media input in args.

The reason is that the user can simply pass in a reference to an audio column, without having to be aware of the file behind it.
___
*[`pixeltable/iterators/audio.py` line 112 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIWdmOGCx7eoua7WtH-:-OIr7eFd8LuOL_Odc3qW:b-t77xc3) ([raw file](https://github.com/pixeltable/pixeltable/blob/33b540dd7c84db38a120c01fe35f62d870674bea/pixeltable/iterators/audio.py#L112)):*

I would argue that in your example, the 18-21s chunk should be included: it's a 3-second chunk, not a 1-second chunk. Filtering it out in this case seems counterintuitive to me. If we call the 18-21s chunk a 1-second chunk, wouldn't we also have to call (say) the 9-14s chunk a 3-second chunk because of the overlap (which conflicts with our terminology)? I'd suggest we use "chunk size" consistently to mean the actual duration of the chunk.
</blockquote></details>

agreed

amithadke (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/451)* status: 0 of 8 files reviewed, 15 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/iterators/audio.py` line 25 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/451#-OIIRA3h30AXaMqZ8uCA:-OIrPzID8i1gK84vPXl7:b-k1gz30) ([raw file](https://github.com/pixeltable/pixeltable/blob/b65340f9646e67a7627b54908473e888047a027c/pixeltable/iterators/audio.py#L25)):*

Looking at the docstring for extract_audio() and DocumentSplitter, we actually don't list the main media input in args.

The reason is that the user can simply pass in a reference to an audio column, without having to be aware of the file behind it.
</blockquote></details>

removed audio from args

PR #452, opened 2025-01-29
Add get_view and get_snapshot as aliases for get_table

jacobweiss2305:
As a user, I get confused when i want to pull a view or snapshot from storage with `get_table`. The functionality works but it causes confusion in my code.

I propose a minimally invasive yet readable solution by creating two alias for get_table. These aliases are modular function names: `get_view` and `get_snapshot`.

This extends the global functions to provide additional ways to retrieve views and snapshots using the existing get_table implementation without confusion the developer.

### DIFF
--- a/pixeltable/__init__.py
+++ b/pixeltable/__init__.py
@@ -4,7 +4,7 @@ from .exceptions import Error
 from .exprs import RELATIVE_PATH_ROOT
 from .func import Aggregator, Function, expr_udf, query, uda, udf
 from .globals import (array, configure_logging, create_dir, create_snapshot, create_table, create_view, drop_dir,
-                      drop_table, get_table, init, list_dirs, list_functions, list_tables, move, tool, tools)
+                      drop_table, get_table, get_view, get_snapshot, init, list_dirs, list_functions, list_tables, move, tool, tools)
 from .type_system import (Array, ArrayType, Audio, AudioType, Bool, BoolType, ColumnType, Document, DocumentType, Float,
                           FloatType, Image, ImageType, Int, IntType, Json, JsonType, Required, String, StringType,
                           Timestamp, TimestampType, Video, VideoType)
--- a/pixeltable/globals.py
+++ b/pixeltable/globals.py
@@ -420,6 +420,8 @@ def get_table(path: str) -> catalog.Table:
     assert isinstance(obj, catalog.Table)
     return obj
 
+get_view = get_table
+get_snapshot = get_table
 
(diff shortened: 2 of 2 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/452)

aaron-siegel (commented):
I'm on board with this but I'd suggest they should be functions that check that the requested object actually is a view/snapshot, e.g.,
```
def get_view(path: str) -> catalog.Table:
    tbl = get_table(str)
    if not tbl._tbl_version.is_view():
        raise excs.Error(...)
    return tbl
```
Also add tests (both for the intended usage pattern and for the new errors)

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/452#-:-OHjcDJs52Y5lbSTU38l:b-5om9sf)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @mkornacker and @pierrebrunelle)

aaron-siegel (changes requested):
See above comment

pierrebrunelle:
"but it causes confusion in my code."

What kind of confusion? 

I'm not sure I like this. What if a user list_tables? Do we need list_views and list_snapshots now as well? Here I'd say why not cause it could potentially filter the list to "just" the views or a snapshot and therefore adds some value per se.

Simply adding an alias to me that does nothing but creating redundancy I don't see why we would need this but actually adding more confusion. What if I try now: get_view on a table? Should that actually fail? We are just creating edge cases four ourselves for no obvious value.

jacobweiss2305:
@pierrebrunelle 

We have an API inconsistency that could confuse users:

- CREATE: `pxt.create_view("agent.pdf_chunks", documents_t)`
- READ: `pxt.get_table("agent.pdf_chunks")` 🤔

Key issues:
- Views are implemented as specialized tables (View extends Table), but this is an implementation detail leaking into our API
- Violates principle of least surprise - users expect views and tables to be distinct
- Creates cognitive dissonance between creation (view) and retrieval (table) interfaces 
- While pragmatic from implementation perspective, it adds confusion for users from database backgrounds

I do agree with you and @aaron-siegel  on the alias. It should actually be seperate functions and not alias.

mkornacker:
Maybe this warrants further discussion.

We don't want users to see views and tables as being distinct, because the way you query them is the same (and a lot of other things as well).

Maybe that's an indication that we haven't given the user a clear enough picture of what that actually means.

jacobweiss2305:
Closing for now.

PR #454, opened 2025-01-29
[SHOW] PXT-0416: Enable variable parameter in query limit() clause

WildCard932:
Accomplishes the feature noted in the Jira ticket.

There are no tests yet, this PR is for feedback.

There seems to be a problem with the substitute() method in expr.py. The problem is noted in comments in the method(). I will pend this change until a decision is made about the operation of this method.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)

WildCard932 (commented):
Tests added. linter problem fixed.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454#-:-OHsKk7OBkQmBIOStc0h:b-hwlpir)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @aaron-siegel and @mkornacker)

aaron-siegel (changes requested):
The overall approach looks right to me. It'd be good to get @mkornacker 's impressions too. Various comments inline.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454#-:-OHtLoaVB9VzuxnV29rn:b-6dpevk)* status: 0 of 7 files reviewed, 9 unresolved discussions (waiting on @mkornacker and @WildCard932)

___
*[`tests/test_dataframe.py` line 287 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtK4vM3Kne5ZLd5qGh:-OHtK4vM3Kne5ZLd5qGi:b-s6lsn) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/tests/test_dataframe.py#L287)):*
> ```Python
> 
>         @pxt.query
>         def get_lim(n : int):
> ```

Python style: no space before colon

Also add `DataFrame` as the return type
___
*[`tests/test_dataframe.py` line 290 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtL5cu2dEXQdgdb96r:-OHtKRAW26qdxQFmKC7s:bxmu5y8) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/tests/test_dataframe.py#L290)):*
> ```Python
>             return t.select(t.c4).limit(n)
> 
>         res = t.select(t.c4, get_lim(2)).collect()
> ```

Add some code to validate `res`. Also use a `ReloadTester` to ensure we're testing the round trip serialization
___
*[`pixeltable/dataframe.py` line 404 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtHI0v4M_xxKMIoa7t:-OHtH2cG56Pk-BNmG-gN:bo165ik) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L404)):*
> ```Python
>                 (expr, asc) for expr, asc in zip(order_by_exprs, [asc for _, asc in self.order_by_clause])
>             ]
>         if (limit_val) is not None:
> ```

No need for enclosing parens around `limit_val`
___
*[`pixeltable/dataframe.py` line 405 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtJ8QM7WcyMj0t2hfI:-OHtJ8QM7WcyMj0t2hfJ:bldxtjr) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L405)):*
> ```Python
>             ]
>         if (limit_val) is not None:
>             limit_val = limit_val.substitute(var_exprs)
> ```

I'd suggest putting a check somewhere that the limit expression is a constant after binding (it can't contain column references). I understand it will be caught later but I think it'll be clearer for the user to detect this at an earlier stage and give an informative error message.
___
*[`pixeltable/dataframe.py` line 898 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtLTa10KOlK7Y2Y37z:-OHtLTa10KOlK7Y2Y38-:bqlkujt) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L898)):*
> ```Python
>         )
> 
>     def limit(self, n: Union[int, exprs.Expr]) -> DataFrame:
> ```

I'm not sure we want to expose the `exprs.Expr` as a type option in the user-facing API. The reason is that in user code, the input is always an `int`; it's the `@pxt.query` decorator that munges it into an `Expr`. I'm not sure there's a use case for user code directly supplying an `Expr` to `limit()`.
___
*[`pixeltable/dataframe.py` line 909 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtItyCAV6Xl14u3RZI:-OHtHhFO78gkJtFTpPMp:b-87iuvv) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L909)):*
> ```Python
>         assert n is not None
>         if isinstance(n, int):
>             n = exprs.Literal(n, ts.IntType(nullable=False))
> ```

It's not actually necessary to specify `IntType` explicitly here (`n = exprs.Literal(n)` should work and the Pixeltable type will be inferred)
___
*[`pixeltable/dataframe.py` line 911 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtHrsZ7T7_gCYGetsT:-OHtHrsZ7T7_gCYGetsU:b-s4opxr) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L911)):*
> ```Python
>             n = exprs.Literal(n, ts.IntType(nullable=False))
>         if not isinstance(n, exprs.Expr):
>             raise excs.Error(f'Limit() requires a Pixeltable expression, but instead got {type(n)}')
> ```

You could replace the preceding four lines with this:
```python
expr = exprs.Expr.from_object(n)
```
which will do what you want: turn `n` into a `Literal` if it's an `int`; leave it alone if it's an Expression
___
*[`pixeltable/dataframe.py` line 1010 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtIG6997O9Jj9-rwjT:-OHtIG6997O9Jj9-rwjU:bjxp0yh) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L1010)):*
> ```Python
> 
>     @classmethod
>     def limit_val_from_dict(cls, lv : Any) -> exprs.Expr:
> ```

Python style: no space before the type hint colon (`lv: Any`, not `lv : Any`)
___
*[`pixeltable/dataframe.py` line 1013 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtIXED9ZTxTEL_Xvfu:-OHtIXED9ZTxTEL_Xvfv:b-s7khx4) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L1013)):*
> ```Python
>         if lv is None:
>             return None
>         if isinstance(lv, int):
> ```

We try to avoid proliferation of deserialization cases after these types of changes. When the dict structure changes, you can avoid this with a database migration; see code in `pixeltable.metadata.converters`.

WildCard932 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)* status: 0 of 9 files reviewed, 9 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/dataframe.py` line 404 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtHI0v4M_xxKMIoa7t:-OHxA-9OC8nOA8HuYSPn:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L404)):*

No need for enclosing parens around `limit_val`
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 405 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtJ8QM7WcyMj0t2hfI:-OHxAUcpB4JD1CMVXT9E:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L405)):*

I'd suggest putting a check somewhere that the limit expression is a constant after binding (it can't contain column references). I understand it will be caught later but I think it'll be clearer for the user to detect this at an earlier stage and give an informative error message.
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 909 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtItyCAV6Xl14u3RZI:-OHxB4BC5QFgX4Ipa-ab:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L909)):*

It's not actually necessary to specify `IntType` explicitly here (`n = exprs.Literal(n)` should work and the Pixeltable type will be inferred)
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 911 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtHrsZ7T7_gCYGetsT:-OHxB2r_EzAx1O7oOxKl:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L911)):*

You could replace the preceding four lines with this:
```python
expr = exprs.Expr.from_object(n)
```
which will do what you want: turn `n` into a `Literal` if it's an `int`; leave it alone if it's an Expression
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 1010 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtIG6997O9Jj9-rwjT:-OHxBRD11E7qQ82SXPa3:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L1010)):*

Python style: no space before the type hint colon (`lv: Any`, not `lv : Any`)
</blockquote></details>

Done.
___
*[`pixeltable/dataframe.py` line 1013 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtIXED9ZTxTEL_Xvfu:-OICuRSj1uHEsIHzTnT2:b-nv0ixs) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L1013)):*

We try to avoid proliferation of deserialization cases after these types of changes. When the dict structure changes, you can avoid this with a database migration; see code in `pixeltable.metadata.converters`.
</blockquote></details>

I still need to do this.
___
*[`tests/test_dataframe.py` line 287 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtK4vM3Kne5ZLd5qGh:-OICutNH6wlPs1T8GLx7:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/tests/test_dataframe.py#L287)):*

Python style: no space before colon

Also add `DataFrame` as the return type
</blockquote></details>

Done.

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)* status: 0 of 9 files reviewed, 8 unresolved discussions (waiting on @mkornacker and @WildCard932)

___
*[`pixeltable/dataframe.py` line 898 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtLTa10KOlK7Y2Y37z:-OIH8vKFEnEgCjpLmDKs:b-tuei2f) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L898)):*

I'm not sure we want to expose the `exprs.Expr` as a type option in the user-facing API. The reason is that in user code, the input is always an `int`; it's the `@pxt.query` decorator that munges it into an `Expr`. I'm not sure there's a use case for user code directly supplying an `Expr` to `limit()`.
</blockquote></details>

Bumping this comment to be sure it's not overlooked - let's discuss if we need to
___
*[`tests/test_dataframe.py` line 306 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIHB1753UrIyOZ29neV:-OIHB1753UrIyOZ29neW:bxv50rd) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/tests/test_dataframe.py#L306)):*
> ```Python
>         assert res[0]['get_lim'] == [{'c4': False}, {'c4': True}]
> 
>         with pytest.raises(excs.Error) as exc_info:
> ```

Note: we've been adopting the following leaner syntax for this (but still haven't ported all the legacy `pytest.raises` instances):

```python
with pytest.raises(excs.Error, match='must be of type int'):
    _ = t.limit(5.3).collect()
```
___
*[`pixeltable/dataframe.py` line 408 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIH8r0c-l6HAftMW9ur:-OIH8r0c-l6HAftMW9us:bfzqgbg) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/pixeltable/dataframe.py#L408)):*
> ```Python
>             limit_val = limit_val.substitute(var_exprs)
>             if limit_val is not None and not isinstance(limit_val, exprs.Literal):
>                 raise excs.Error(f'Limit value must be a constant, but got type{limit_val}')
> ```

Missing a space before `{limit_val}`
___
*[`pixeltable/catalog/table.py` line 215 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIH99blEHd-R1uACZAA:-OIH99blEHd-R1uACZAB:b3im2ya) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/pixeltable/catalog/table.py#L215)):*
> ```Python
>         return self._df().group_by(*items)
> 
>     def limit(self, n: Union[int, 'exprs.Expr']) -> 'pxt.DataFrame':
> ```

Same comment as above (we probably don't want to expose `exprs.Expr` to the user here)
___
*[`pixeltable/exprs/expr.py` line 196 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIHAZVq9axC2FG6Nuf9:-OIHAZVr1PD6NpmA7Frv:b-mlr5cc) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/pixeltable/exprs/expr.py#L196)):*
> ```Python
>         for i in range(len(self.components)):
>             self.components[i] = self.components[i].substitute(spec)
>         if self.is_foldable():
> ```

Oh I see, you're actually doing constant folding inside expressions. This is a bigger change! I'm not opposed to it but also don't think we need it to solve the present problem. If we're going to do this, we should use the existing `as_constant()` pattern to pick up other cases as well such as inline dict/list folding.
___
*[`pixeltable/exprs/arithmetic_expr.py` line 131 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIH9GVC6v9bzvwQdMEe:-OIH9GVC6v9bzvwQdMEf:b7fo0hd) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/pixeltable/exprs/arithmetic_expr.py#L131)):*
> ```Python
>             data_row[self.slot_idx] = op1_val // op2_val
> 
>     def is_foldable(self) -> bool:
> ```

Why not use the existing `is_constant`/`_as_constant` pattern here? (See `InlineDict` for an example of how it's implemented. I think you don't need to override `is_constant` at all; the default should work; just replace `folded` with `_as_constant` and return the primitive value rather than an Expr?)

WildCard932 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)* status: 0 of 9 files reviewed, 8 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/dataframe.py` line 898 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtLTa10KOlK7Y2Y37z:-OIN1gZD7VY8Wv9FRRuu:bangth1) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L898)):*

Bumping this comment to be sure it's not overlooked - let's discuss if we need to
</blockquote></details>

A thought: if the user is able to use the API as:
t.select().limit(n + 1)
Does the user not see "n + 1" as an expression?
___
*[`pixeltable/dataframe.py` line 408 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIH8r0c-l6HAftMW9ur:-OINM2DO7mePGnGdGLIG:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/pixeltable/dataframe.py#L408)):*

Missing a space before `{limit_val}`
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 215 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIH99blEHd-R1uACZAA:-OIN2Mgw98MK4vmrMBen:b-u4gr7r) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/pixeltable/catalog/table.py#L215)):*

Same comment as above (we probably don't want to expose `exprs.Expr` to the user here)
</blockquote></details>

See above reply.
___
*[`tests/test_dataframe.py` line 290 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtL5cu2dEXQdgdb96r:-OINKyGH1tOChEXK79wf:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/tests/test_dataframe.py#L290)):*

Add some code to validate `res`. Also use a `ReloadTester` to ensure we're testing the round trip serialization
</blockquote></details>

Done.
___
*[`tests/test_dataframe.py` line 306 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIHB1753UrIyOZ29neV:-OIN2aOG65qt00FmngAJ:b-896fjx) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/tests/test_dataframe.py#L306)):*

Note: we've been adopting the following leaner syntax for this (but still haven't ported all the legacy `pytest.raises` instances):

```python
with pytest.raises(excs.Error, match='must be of type int'):
    _ = t.limit(5.3).collect()
```
</blockquote></details>

Done

WildCard932 (commented):
@aaron-siegel : I believe I have addressed all of your concerns
@mkornacker : Do you want to look at this review?

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454#-:-OIlTJ9R80eA1rbBpN9y:b-sr1tm)* status: 0 of 20 files reviewed, 8 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/dataframe.py` line 898 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtLTa10KOlK7Y2Y37z:-OIlSjzr4VQw240MFDvg:bb6un1j) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L898)):*
<details><summary><i>Previously, WildCard932 wrote…</i></summary><blockquote>

A thought: if the user is able to use the API as:
t.select().limit(n + 1)
Does the user not see "n + 1" as an expression?
</blockquote></details>

Done
___
*[`pixeltable/dataframe.py` line 1013 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OHtIXED9ZTxTEL_Xvfu:-OIlSofBEX7bIJe4-u_y:b-896fjx) ([raw file](https://github.com/pixeltable/pixeltable/blob/dd5f7e437f0025e7fe36fe4ff4c36c84a7bd6779/pixeltable/dataframe.py#L1013)):*
<details><summary><i>Previously, WildCard932 wrote…</i></summary><blockquote>

I still need to do this.
</blockquote></details>

Done
___
*[`pixeltable/exprs/arithmetic_expr.py` line 131 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIH9GVC6v9bzvwQdMEe:-OIlSwZ684DBHr3psWK-:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/pixeltable/exprs/arithmetic_expr.py#L131)):*

Why not use the existing `is_constant`/`_as_constant` pattern here? (See `InlineDict` for an example of how it's implemented. I think you don't need to override `is_constant` at all; the default should work; just replace `folded` with `_as_constant` and return the primitive value rather than an Expr?)
</blockquote></details>

Done.
___
*[`pixeltable/exprs/expr.py` line 196 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIHAZVq9axC2FG6Nuf9:-OIlSzNn-19NpjCS9XRx:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/pixeltable/exprs/expr.py#L196)):*

Oh I see, you're actually doing constant folding inside expressions. This is a bigger change! I'm not opposed to it but also don't think we need it to solve the present problem. If we're going to do this, we should use the existing `as_constant()` pattern to pick up other cases as well such as inline dict/list folding.
</blockquote></details>

Done.
___
*[`pixeltable/catalog/table.py` line 215 at r3](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIH99blEHd-R1uACZAA:-OIlT9fU42T7LkLmW7dz:b-896fjx) ([raw file](https://github.com/pixeltable/pixeltable/blob/b578e4e0d699c42c9e88871e6e377c0ef1a29e17/pixeltable/catalog/table.py#L215)):*
<details><summary><i>Previously, WildCard932 wrote…</i></summary><blockquote>

See above reply.
</blockquote></details>

Done

aaron-siegel (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)* status: 0 of 20 files reviewed, 9 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/exprs/arithmetic_expr.py` line 107 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlwuwB4r0r0WB1XRDF:-OIlwuwB4r0r0WB1XRDG:bsecsng) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exprs/arithmetic_expr.py#L107)):*
> ```Python
>         data_row[self.slot_idx] = self.eval_nullable(op1_val, op2_val)
> 
>     def eval_nullable(self, op1_val: Optional[Union[int, float]], op2_val: Optional[Union[int, float]]) -> Optional[Union[int, float]]:
> ```

`Union[int, float, None]` is a more concise way of saying `Optional[Union[int, float]]`.
___
*[`pixeltable/exprs/arithmetic_expr.py` line 134 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlxkHe4-Y6Q22aJv5_:-OIlxkHe4-Y6Q22aJv5a:b-7q90t4) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exprs/arithmetic_expr.py#L134)):*
> ```Python
> 
>     def is_constant(self) -> bool:
>         return self.is_foldable()
> ```

Why have these as separate methods now? Why not simply fold the body of `is_foldable` into `is_constant` and so on?
___
*[`pixeltable/exprs/expr.py` line 210 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlxtLz0BElQ8oTFqoR:-OIlxtLz0BElQ8oTFqoS:b-tcrmhy) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exprs/expr.py#L210)):*
> ```Python
>         for i in range(len(self.components)):
>             self.components[i] = self.components[i].substitute(spec)
>         if self.is_constant():
> ```

I still have the point of view that it'd be slightly more logical for substitute to *just* do variable substitution, and for constant folding to be a separate operation. But I don't feel super strongly about it.
___
*[`pixeltable/metadata/converters/convert_26.py` line 15 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlyYLr666Xpdk9o_3p:-OIlyYLr666Xpdk9o_3q:bos5xs2) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/metadata/converters/convert_26.py#L15)):*
> ```Python
> 
> def __substitute_md(k: Optional[str], v: Any) -> Optional[tuple[Optional[str], Any]]:
>     from pixeltable.exprs.literal import Literal
> ```

unused import (you refer to it as `exprs.Literal` below)
___
*[`pixeltable/metadata/converters/convert_26.py` line 19 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlye6kE5JuJ4MSE3TU:-OIlye6kE5JuJ4MSE3TV:b7f3xob) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/metadata/converters/convert_26.py#L19)):*
> ```Python
>     import pixeltable.exprs as exprs
>     if k == 'limit_val' and v is not None:
>         if isinstance(v, int):
> ```

might as well say `assert isinstance(v, int)` since it *must* be an int on versions prior to 27.
___
*[`pixeltable/metadata/converters/convert_26.py` line 21 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlySh2542xNp-ivrtk:-OIlySh2542xNp-ivrtl:b6wlvkm) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/metadata/converters/convert_26.py#L21)):*
> ```Python
>         if isinstance(v, int):
>             newv = exprs.Literal(v, ts.IntType(nullable=False)).as_dict()
>             print("*************************************", newv)
> ```

remove debugging code
___
*[`pixeltable/type_system.py` line 866 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlvwTs1cZkre1TBIUF:-OIlvwTs1cZkre1TBIUG:bdk7j11) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/type_system.py#L866)):*
> ```Python
>         elif np.issubdtype(val.dtype, np.str_):
>             # Note that this includes NumPy types like '<U1' -- arrays of single Unicode characters
>             dtype: ColumnType = StringType()
> ```

no need to repeat the type hint here. best practice in Python is actually to factor it out of the if/elif block:
```python
dtype: ColumnType
if np.issubdtype(val.dtype, np.integer):
    dtype = IntType()
elif # etc.
```
I realize the old code didn't follow this pattern :) But we can fix it now
___
*[`pixeltable/exec/exec_node.py` line 119 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlwPTD0BIIxwAPhe8G:-OIlwPTD0BIIxwAPhe8H:b-vcjogk) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exec/exec_node.py#L119)):*
> ```Python
> 
>     def set_limit(self, limit: exprs.Expr) -> None:
>         """Default implementation propagates to input"""
> ```

Put an assertion here to enforce that `limit` is a constant?
___
*[`tests/test_dataframe.py` line 308 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlyw2z8nPglPYxwg78:-OIlyw2z8nPglPYxwg79:b-1cuo5a) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/tests/test_dataframe.py#L308)):*
> ```Python
>         )
>         print(results.schema)
>         reload_tester.run_reload_test()
> ```

This isn't actually doing anything since you're just using dynamic select statements in the test. You need to create a view in the test so that it gets serialized to persistent metadata.

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)* status: 0 of 20 files reviewed, 8 unresolved discussions (waiting on @mkornacker and @WildCard932)

___
*[`pixeltable/exprs/arithmetic_expr.py` line 141 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OInOgf31ZVrGONdV91m:-OInOgf31ZVrGONdV91n:buxt91t) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/arithmetic_expr.py#L141)):*
> ```Python
>         return op1_ok and op2_ok
> 
>     def folded(self) -> Union[int, float]:
> ```

Same comment applies to folded/as_constant.

mkornacker (changes requested):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)* status: 0 of 20 files reviewed, 19 unresolved discussions (waiting on @aaron-siegel and @WildCard932)

___
*[`pixeltable/exprs/inline_expr.py` line 64 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqh0yh1k5E59Nz_2Jx:-OIqh0yh1k5E59Nz_2Jy:bfpavxq) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/inline_expr.py#L64)):*
> ```Python
>         return True  # Always true if components match
> 
>     def is_constant(self) -> bool:
> ```

This duplicates the logic from the superclass. Remove (here and elsewhere).
___
*[`pixeltable/exprs/expr.py` line 141 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqgaxd-jXGbWedSDvM:-OIqgaxd-jXGbWedSDvN:bjdaw0p) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/expr.py#L141)):*
> ```Python
>         for attr, value in self._id_attrs():
>             hasher.update(attr.encode('utf-8'))
>             hasher.update(str(value).encode('utf-8'))
> ```

Use repr() here, not str().
___
*[`pixeltable/exprs/expr.py` line 189 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqfymn6SNVM0_-vpNJ:-OIqfymn6SNVM0_-vpNK:btknh6c) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/expr.py#L189)):*
> ```Python
>         return result
> 
>     @property
> ```

Move back to ArithmeticExpr.  An Expr isn't generally expected to have operands.
___
*[`pixeltable/exprs/expr.py` line 210 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqgnK6DOQYdRrXCp1e:-OIqgnK6DOQYdRrXCp1f:b-sh302x) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/expr.py#L210)):*
> ```Python
>         for i in range(len(self.components)):
>             self.components[i] = self.components[i].substitute(spec)
>         if self.is_constant():
> ```

Remove. This only does substitution.
___
*[`pixeltable/exprs/expr.py` line 400 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqgu9u-gTlcTmPHatm:-OIqgu9u-gTlcTmPHatn:ba3dfmt) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/expr.py#L400)):*
> ```Python
>         return None
> 
>     def is_foldable(self) ->bool:
> ```

Remove.
___
*[`pixeltable/exprs/arithmetic_expr.py` line 133 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqgGsT2oYMwB6_ygw9:-OIqgGsT2oYMwB6_ygwA:b-ue91dl) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/arithmetic_expr.py#L133)):*
> ```Python
>             return op1_val // op2_val
> 
>     def is_constant(self) -> bool:
> ```

Why does this class need a custom implementation?
___
*[`pixeltable/exprs/arithmetic_expr.py` line 135 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqfXDi7jFh-ZF9wHkk:-OIqfXDi7jFh-ZF9wHkl:bj8u0ix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/arithmetic_expr.py#L135)):*
> ```Python
>     def is_constant(self) -> bool:
>         '''
>         Return True if the operation may be folded over its operands.
> ```

Remove comment, the superclass already contains a comment, and this doesn't add anything.
___
*[`pixeltable/exprs/arithmetic_expr.py` line 141 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqgUI82-qlViHoyCMw:-OIqgUI82-qlViHoyCMx:bii4gy0) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/arithmetic_expr.py#L141)):*
> ```Python
>         return op1_ok and op2_ok
> 
>     def folded(self) -> Union[int, float]:
> ```

Merge into _as_constant().
___
*[`pixeltable/dataframe.py` line 922 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqe18ZDno3Wo84izh0:-OIqe18ZDno3Wo84izh1:b-d8cnvr) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/dataframe.py#L922)):*
> ```Python
>         n = exprs.Expr.from_object(n)
>         if not n.col_type.is_int_type():
>             raise excs.Error(f'Limit(): parameter must be of type int, instead of {n.col_type}')
> ```

'limit(): ...'
___
*[`tests/test_dataframe.py` line 286 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqiLJY22V-zkmXCXId:-OIqiLJY22V-zkmXCXIe:b-957nl5) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/tests/test_dataframe.py#L286)):*
> ```Python
>         assert len(res) == 4
> 
>     def test_limit(self, test_tbl: catalog.Table, reload_tester: ReloadTester) -> None:
> ```

Why have test_limit, test_limit2, test_limit4, test_limit5?
___
*[`pixeltable/exprs/type_cast.py` line 49 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqhm7rB1s0ApE6BZWf:-OIqhm7rB1s0ApE6BZWg:b6cb9dn) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/type_cast.py#L49)):*
> ```Python
>         return self.col_type.is_numeric_type() and (self._op1.col_type.is_numeric_type() or self._op1.col_type.is_bool_type()) and isinstance(self._op1, Literal)
> 
>     def folded(self) -> Union[int, float]:
> ```

Why special-case numerical types, and why now?

Merge into _as_constant()

WildCard932 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)* status: 0 of 20 files reviewed, 19 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/type_system.py` line 866 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlvwTs1cZkre1TBIUF:-OIm3NL_CtlMvWpCXvSb:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/type_system.py#L866)):*

no need to repeat the type hint here. best practice in Python is actually to factor it out of the if/elif block:
```python
dtype: ColumnType
if np.issubdtype(val.dtype, np.integer):
    dtype = IntType()
elif # etc.
```
I realize the old code didn't follow this pattern :) But we can fix it now
</blockquote></details>

Done.
___
*[`pixeltable/exprs/arithmetic_expr.py` line 107 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlwuwB4r0r0WB1XRDF:-OIm3tntCb_YBdJvsbM4:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exprs/arithmetic_expr.py#L107)):*

`Union[int, float, None]` is a more concise way of saying `Optional[Union[int, float]]`.
</blockquote></details>

Done.
___
*[`pixeltable/exprs/arithmetic_expr.py` line 134 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlxkHe4-Y6Q22aJv5_:-OImA1-m7Jefzq5p6oO7:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exprs/arithmetic_expr.py#L134)):*

Why have these as separate methods now? Why not simply fold the body of `is_foldable` into `is_constant` and so on?
</blockquote></details>

Done.
___
*[`pixeltable/exprs/expr.py` line 210 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlxtLz0BElQ8oTFqoR:-OIm8WnG9KEFdqwBlSqb:b-41k3i) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exprs/expr.py#L210)):*

I still have the point of view that it'd be slightly more logical for substitute to *just* do variable substitution, and for constant folding to be a separate operation. But I don't feel super strongly about it.
</blockquote></details>

I am sensitive to the cost of another copy and another visit to all nodes in the expression tree(s).
Codebases sometimes build infrastructure to minimize the cost of these operations. I haven't found that in the pxt codebase yet, and it seems overkill to build it for this PR. 
___
*[`pixeltable/metadata/converters/convert_26.py` line 15 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlyYLr666Xpdk9o_3p:-OIm4R9107CIrlQiTnH6:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/metadata/converters/convert_26.py#L15)):*

unused import (you refer to it as `exprs.Literal` below)
</blockquote></details>

Done.
___
*[`pixeltable/metadata/converters/convert_26.py` line 19 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlye6kE5JuJ4MSE3TU:-OIm4a_f6W07qha1c1tu:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/metadata/converters/convert_26.py#L19)):*

might as well say `assert isinstance(v, int)` since it *must* be an int on versions prior to 27.
</blockquote></details>

Done.
___
*[`pixeltable/metadata/converters/convert_26.py` line 21 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlySh2542xNp-ivrtk:-OIm4b761oVp1FsZM3qp:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/metadata/converters/convert_26.py#L21)):*

remove debugging code
</blockquote></details>

Done.
___
*[`tests/test_dataframe.py` line 308 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlyw2z8nPglPYxwg78:-OImCxvpEImiPrQqdSMx:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/tests/test_dataframe.py#L308)):*

This isn't actually doing anything since you're just using dynamic select statements in the test. You need to create a view in the test so that it gets serialized to persistent metadata.
</blockquote></details>

Done.
___
*[`pixeltable/exec/exec_node.py` line 119 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlwPTD0BIIxwAPhe8G:-OIm7QPt9ddp9spD5s_a:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exec/exec_node.py#L119)):*

Put an assertion here to enforce that `limit` is a constant?
</blockquote></details>

Done.

WildCard932 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)* status: 0 of 20 files reviewed, 19 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/exprs/expr.py` line 141 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqgaxd-jXGbWedSDvM:-OIvKcAFBXxf_FaUaisD:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/expr.py#L141)):*

Use repr() here, not str().
</blockquote></details>

Done.
___
*[`pixeltable/exprs/expr.py` line 189 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqfymn6SNVM0_-vpNJ:-OIvKBUO4XZUgZJlNsLD:bcz25ph) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/expr.py#L189)):*

Move back to ArithmeticExpr.  An Expr isn't generally expected to have operands.
</blockquote></details>

This is generic way to refer to the first component - either operand or parameter to function. 
___
*[`pixeltable/exprs/expr.py` line 210 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqgnK6DOQYdRrXCp1e:-OIvJfqZ57qJKld9_MN1:b52bro5) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/expr.py#L210)):*

Remove. This only does substitution.
</blockquote></details>

Not done per discussion yesterday.
___
*[`pixeltable/exprs/expr.py` line 400 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqgu9u-gTlcTmPHatm:-OIvJnUXEjgE4QbCM3mu:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/expr.py#L400)):*

Remove.
</blockquote></details>

Done.
___
*[`pixeltable/exprs/type_cast.py` line 49 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqhm7rB1s0ApE6BZWf:-OIvJpmJ49ckxOIKr1e9:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/pixeltable/exprs/type_cast.py#L49)):*

Why special-case numerical types, and why now?

Merge into _as_constant()
</blockquote></details>

Done.
___
*[`tests/test_dataframe.py` line 286 at r6](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIqiLJY22V-zkmXCXId:-OIvJt932h-DsqLsdsjU:b-9ox1rn) ([raw file](https://github.com/pixeltable/pixeltable/blob/d12a934573e5025336d551fa52436d1e81619cee/tests/test_dataframe.py#L286)):*

Why have test_limit, test_limit2, test_limit4, test_limit5?
</blockquote></details>

They test slightly different things.

WildCard932 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/454)* status: 0 of 20 files reviewed, 19 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`pixeltable/exprs/expr.py` line 210 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlxtLz0BElQ8oTFqoR:-OIwe6xNBxiRYTrBAhje:b-4slpva) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exprs/expr.py#L210)):*
<details><summary><i>Previously, WildCard932 wrote…</i></summary><blockquote>

I am sensitive to the cost of another copy and another visit to all nodes in the expression tree(s).
Codebases sometimes build infrastructure to minimize the cost of these operations. I haven't found that in the pxt codebase yet, and it seems overkill to build it for this PR.
</blockquote></details>

Not done per discussion Tuesday.
___
*[`pixeltable/exec/exec_node.py` line 119 at r5](https://reviewable.io/reviews/pixeltable/pixeltable/454#-OIlwPTD0BIIxwAPhe8G:-OIwemgV0Msot5XJnwd1:bv6qao6) ([raw file](https://github.com/pixeltable/pixeltable/blob/3663f5ada70fa589328b82503e3ce2399f35fc8c/pixeltable/exec/exec_node.py#L119)):*
<details><summary><i>Previously, WildCard932 wrote…</i></summary><blockquote>

Done.
</blockquote></details>

It is downgraded from Literal to an int at the call site.

pixeltable/exprs/arithmetic_expr.py
```diff
@@ -130,21 +141,17 @@ def eval_non_null(self, op1_val: Union[int, float], op2_val: Union[int, float])
...
-        op1_ok = self._op1.col_type.is_numeric_type() and isinstance(self._op1, Literal)
-        op2_ok = self._op2.col_type.is_numeric_type() and isinstance(self._op2, Literal)
-        return op1_ok and op2_ok
-
-    def folded(self) -> Union[int, float]:
-        op1_val = self._op1.as_constant()
-        op2_val = self._op2.as_constant()
-        return self.eval_non_null(op1_val, op2_val)
+    def as_literal(self) -> Optional[Literal]:
+        if not isinstance(self._op1, Literal):
```
mkornacker: as_literal() == None instead of not isinstance(Literal)

mkornacker (approved):

PR #459, opened 2025-01-30
Pxt 429 missing tiktoken

jacobweiss2305:
Missing tiktoken dependency in sample apps


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/459)

aaron-siegel (changes requested):
This PR appears to also include the get_view and get_snapshot changes; is that intentional?

jacobweiss2305:
Good catch. Sorry about that.

aaron-siegel (approved):

PR #460, opened 2025-01-30
Remove unused code from arrow.py

aaron-siegel:
Removes unused code from pixeltable/utils/arrow.py. This fixes debug logging in pytest (the unused code was causing a phantom Env to be instantiated as a side effect, overriding the log settings).

Also adds a guard to `Env.__init__()` to enforce that `Env` is a singleton.

### DIFF
--- a/pixeltable/env.py
+++ b/pixeltable/env.py
@@ -95,8 +95,7 @@ class Env:
         cls._instance = env
 
     def __init__(self):
-        if self._instance is not None:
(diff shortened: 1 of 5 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/460)

mkornacker (approved):

PR #462, opened 2025-01-31
Pxt 428 add reasoning model prefix

jacobweiss2305:
# API Parameter Changes for OpenAI Models

OpenAI has introduced a change in token limit parameters across their model series:

- Legacy models continue to support the `max_tokens` parameter
- The O1 and O3 series models only accept `max_completion_tokens`

This pull request implements a mapping system that correlates the first character of each model name with its corresponding token parameter, resolving compatibility issues when using reasoning models.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/462)

PR #470, opened 2025-02-11
PXT-0455: Enable computed columns of all types to be unstored.

WildCard932:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/470)

aaron-siegel (approved):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/470)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mkornacker)

___
*[`tests/test_view.py` line 489 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/470#-OIrZzVQ79NlHfoeu0Ee:-OIrZzVQ79NlHfoeu0Ef:b9pjar) ([raw file](https://github.com/pixeltable/pixeltable/blob/fdd0d6225064e70f5058752cd33755f484a153a7/tests/test_view.py#L489)):*
> ```Python
>         check_views()
> 
>     def test_unstored_columns_non_image(self, reset_db) -> None:
> ```

My only quibble is that this test doesn't actually distinguish between stored and unstored columns in its observed behavior.

One way to do this is via a UDF `f` whose behavior is time-dependent (for example, a UDF that returns the value of a global counter that gets incremented with each call). Then if `c2=f(t.c1)` is a computed column, the value for `c2` should be the same on successive queries if `c2` is stored, but distinct if it isn't.

tests/test_view.py
```diff
@@ -486,6 +486,54 @@ def check_views():
         assert v2._version == v2_version
         check_views()
 
+    def test_unstored_columns_non_image(self, reset_db) -> None:
```
mkornacker, line 501: Let's also test this outside of views, ie, in test_table.py.

jpeterson-pxt: Done

WildCard932 (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/470)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @aaron-siegel and @mkornacker)

___
*[`tests/test_view.py` line 489 at r1](https://reviewable.io/reviews/pixeltable/pixeltable/470#-OIrZzVQ79NlHfoeu0Ee:-OIwdQjG0rMtfqvpSz09:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/fdd0d6225064e70f5058752cd33755f484a153a7/tests/test_view.py#L489)):*

My only quibble is that this test doesn't actually distinguish between stored and unstored columns in its observed behavior.

One way to do this is via a UDF `f` whose behavior is time-dependent (for example, a UDF that returns the value of a global counter that gets incremented with each call). Then if `c2=f(t.c1)` is a computed column, the value for `c2` should be the same on successive queries if `c2` is stored, but distinct if it isn't.
</blockquote></details>

Done.

PR #471, opened 2025-02-12
A packager for table data

aaron-siegel:
Introduces a packager for table data, intended as a way to share data between Pixeltable instances.

The packager first exports table data to Iceberg tables (one per ancestor) backed by a sqlite catalog, then builds a tarball containing the Iceberg warehouse along with any local media files referenced by the table or its ancestors.

The tarball does not include metadata; it is assumed that metadata will be communicated separately in synchronous REST calls, though we could consider also including it in the archive as json files.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/471)

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/471)* status: 0 of 11 files reviewed, 5 unresolved discussions

___
*[`pixeltable/share/packager.py` line 14 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-ithN97TGzdCvLWcM:-OJ-ithN97TGzdCvLWcN:b-sk28b5) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L14)):*
> ```Python
> import numpy as np
> import pyarrow as pa
> from pyiceberg.catalog import Catalog
> ```

Might be a good idea to do a qualified import. We also have a Catalog class.
___
*[`pixeltable/share/packager.py` line 75 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-lLaYCsaGWMVFt1dw:-OJ-lLaYCsaGWMVFt1dx:btrm7u7) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L75)):*
> ```Python
>         # Select only those columns that are defined in this table (columns inherited from ancestor
>         # tables will be handled separately)
>         # TODO: This is selecting only named columns; do we also want to preserve system columns such as errortype?
> ```

I think so. Otherwise you can't run the same queries against those tables.
___
*[`pixeltable/share/packager.py` line 77 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-lpOp-1wjqlzxvF0j:-OJ-lpOp-1wjqlzxvF0k:bz26274) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L77)):*
> ```Python
>         # TODO: This is selecting only named columns; do we also want to preserve system columns such as errortype?
>         col_refs = [
>             t[col_name] for col_name, col in t._tbl_version.cols_by_name.items() if not col.col_type.is_media_type()
> ```

This should not include unstored columns.
___
*[`pixeltable/share/packager.py` line 81 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-mX3y04I3s_oCgub2:-OJ-mX3y04I3s_oCgub3:b-uagrnt) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L81)):*
> ```Python
>         # For media columns, we substitute `col.fileurl` so that we always get the URL (which may be a file:// URL;
>         # these will be specially handled later)
>         media_col_refs = {
> ```

These are not col refs. But maybe this should be divided into media_col_refs and from that media_file_urls, and then you can reuse media_col_refs below.
___
*[`pixeltable/share/packager.py` line 97 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-mFqGB4wnB9M4BAGE:-OJ-mFqGB4wnB9M4BAGF:bc829ii) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L97)):*
> ```Python
>         # Separately construct a list of actual column types in the same order of appearance as in the data frame.
>         actual_col_types = [
>             col.col_type for col in t._tbl_version.cols_by_name.values() if not col.col_type.is_media_type()
> ```

easier to do
[cf.col.col_type for cr in col_refs]

mkornacker (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/471)* status: 0 of 11 files reviewed, 8 unresolved discussions (waiting on @aaron-siegel)

___
*[`pixeltable/share/packager.py` line 19 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-pRaM4WuwnQJFVXMB:-OJ-pRaM4WuwnQJFVXMC:b6z44fv) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L19)):*
> ```Python
> import pixeltable.type_system as ts
> from pixeltable.env import Env
> from pixeltable.utils.arrow import _pt_to_pa
> ```

Why _pt_to_pa? (as opposed to pt_to_pa)
___
*[`pixeltable/share/packager.py` line 104 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-pj9096RvCAnztbwj:-OJ-pj9096RvCAnztbwk:bspl8qq) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L104)):*
> ```Python
> 
>         # Populate the Iceberg table with data.
>         for pa_table in self.__to_pa_tables(df, actual_col_types, iceberg_schema):
> ```

Why plural? I would have expected this to produce exactly one iceberg table.
___
*[`pixeltable/share/packager.py` line 145 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-qI2b-vcSr79VMAL9:-OJ-qI2b-vcSr79VMALA:b4eaedm) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L145)):*
> ```Python
>         actual_col_types: list[pxt.ColumnType],
>         arrow_schema: pa.Schema,
>         batch_size: int = 1_000,
> ```

What's the purpose of batching?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/471)* status: 0 of 11 files reviewed, 8 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/share/packager.py` line 14 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-ithN97TGzdCvLWcM:-OJ-qVpP6d4HAuG6-Ht5:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L14)):*

Might be a good idea to do a qualified import. We also have a Catalog class.
</blockquote></details>

Done.
___
*[`pixeltable/share/packager.py` line 77 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-lpOp-1wjqlzxvF0j:-OJ-qsPqBiHjtHYyjuJs:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L77)):*

This should not include unstored columns.
</blockquote></details>

Done.
___
*[`pixeltable/share/packager.py` line 81 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-mX3y04I3s_oCgub2:-OJ-r2g63RX9lTN-wbaz:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L81)):*

These are not col refs. But maybe this should be divided into media_col_refs and from that media_file_urls, and then you can reuse media_col_refs below.
</blockquote></details>

Done.
___
*[`pixeltable/share/packager.py` line 97 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-mFqGB4wnB9M4BAGE:-OJ-sErq6XluOI7Zgb-d:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L97)):*

easier to do
[cf.col.col_type for cr in col_refs]
</blockquote></details>

Done.

pixeltable/share/packager.py
```diff
@@ -70,37 +71,47 @@ def package(self) -> Path:
...
+
+        for col_name, col in t._tbl_version.cols_by_name.items():
+            if not col.is_stored:
+                continue
+            if col.col_type.is_media_type():
+                select_exprs[col_name] = t[col_name].fileurl
+            else:
+                select_exprs[col_name] = t[col_name]
+            actual_col_types.append(col.col_type)
+            if col.is_computed:
```
mkornacker: We also create errortype/-msg for media columns (because they can fail validation). See Column.records_errors().

tests/test_audio.py
```diff
@@ -42,15 +42,15 @@ def test_extract(self, reset_db) -> None:
         assert status.num_excs == 0
         assert MediaStore.count(video_t._id) == len(video_filepaths) - 1
         assert video_t.where(video_t.audio != None).count() == len(video_filepaths) - 1
-        assert env.Env.get().num_tmp_files() == 0
+        tmp_files_before = env.Env.get().num_tmp_files()
```
mkornacker, line 45: Is that because the tmp dir doesn't get reset between tests anymore?

pixeltable/share/packager.py
aaron-siegel: Oops. I knew that :( Good catch. Fixed.

tests/test_audio.py
aaron-siegel: It was never getting reset between tests. So it's kind of a miracle this worked in the first place.

I tried adding a `clear_tmp_dir()` call between tests, but it fails on Windows for some weird file lock reasons. This seems like a reasonable solution in any case.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/471)* status: 0 of 13 files reviewed, 10 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/share/packager.py` line 75 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/471#-OJ-lLaYCsaGWMVFt1dw:-OJAAvRh79aIPwgAvGhg:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/055fb25a5823273396c41e822839d177e428859b/pixeltable/share/packager.py#L75)):*

I think so. Otherwise you can't run the same queries against those tables.
</blockquote></details>

Done.

pixeltable/share/packager.py
```diff
@@ -95,7 +95,7 @@ def __export_table(self, t: pxt.Table) -> None:
             else:
                 select_exprs[col_name] = t[col_name]
             actual_col_types.append(col.col_type)
-            if col.is_computed:
+            if col.is_computed or col.col_type.is_media_type():
```
mkornacker: You need to use if col.records_errors():

aaron-siegel: Got it - yes that's better. Fixed.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/471)* status: 0 of 13 files reviewed, 11 unresolved discussions (waiting on @mkornacker)

PR #480, opened 2025-02-22
Update postgres to 16.8 and pgvector to 0.8.0

aaron-siegel:


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/480)

PR #481, opened 2025-02-23
Improvements to handling of UDF calls in computed columns

aaron-siegel:
Addresses two loose ends in Pixeltable's handling of UDFs in computed columns:
- Ensures that Pixeltable will accept backward-compatible code changes to UDF signatures (previously, Pixeltable would raise an error in many such cases);
- Properly handles timestamp and array literals when they appear as default values and/or constant arguments to UDF parameters (previously, Pixeltable would throw a "not JSON serializable" error).

Architecturally, there are two significant changes to `expr.FunctionCall`.
- Previously, `FunctionCall` used an unusual pattern of binding the user-provided `args` and `kwargs`, then later unrolling them, in a way that led to information loss about the original call pattern. With this change, `args` and `kwargs` are preserved in their original call pattern (in `FunctionCall` and in its dict). Now, if the UDF signature changes in code, backward-compatibility can be checked in a fully consistent way, by re-binding `args` and `kwargs` against the new signature.
- Constant values for `args` and `kwargs` are now stored and serialized as `Literal`s, rather than their primitive Python values. This aligns with the approach now used for `Inline*` expressions, ensures that datetime and array literals are serialized correctly, and permits code simplification in several places. Future support for image literals is now a possibility (which would be useful for things like mask application).

A few additional benefits of this change:
- Unification of the code paths for `call_return_type` and `get_resource_pool`
- Unification of the logic for expression normalization (conversion of primitive Python objects and mixed-type lists and dicts to Exprs)
- Consistent handling of return types
- Consistent handling of variable args and kwargs parameters (`VAR_POSITIONAL` and `VAR_KEYWORD` kinds)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/481)

mkornacker (changes requested):
I'll do function_call.py tomorrow.

> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/481#-:-OKZnyuE5NBIk48szTIY:b-1i6dhp)* status: 0 of 29 files reviewed, 6 unresolved discussions

___
*[`pixeltable/catalog/view.py` line 142 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKXCerH2neW2IYIB_BO:-OKXCerH2neW2IYIB_BP:b-1vph2o) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/catalog/view.py#L142)):*
> ```Python
>             bound_args = {k: exprs.Expr.from_object(v) for k, v in bound_args.items()}
>             sig.validate_args(bound_args, context=f'in iterator {iterator_cls.__name__!r}')
>             bound_args = {k: v.val if isinstance(v, exprs.Literal) else v for k, v in bound_args.items()}
> ```

bound_args changes meaning a few times here, might be better to introduce a new variable at this point.
___
*[`pixeltable/func/signature.py` line 155 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKXzFR5-SVcPj84K38-:-OKXzFR5-SVcPj84K380:bewyax7) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/func/signature.py#L155)):*
> ```Python
> 
>         for param_name, arg in bound_args.items():
>             param = self.parameters[param_name]
> ```

This function can generate user-facing errors. Should it also handle the 'invalid parameter name' case?
___
*[`pixeltable/func/signature.py` line 165 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKY-ksbA5dG5rjXWYMl:-OKY-ksbA5dG5rjXWYMm:bz45dcb) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/func/signature.py#L165)):*
> ```Python
> 
>             # Check that the argument is consistent with the expected parameter type, with the allowance that
>             # non-nullable parameters can still accept nullable arguments (since function calls with Nones
> ```

Clearer explanation: (since FunctionCall.eval() checks for that and skips the evaluation, if present)
___
*[`pixeltable/func/function.py` line 157 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKYoVbu6wbdKE2JLv6j:-OKYoVbu6wbdKE2JLv6k:b-pld4zb) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/func/function.py#L157)):*
> ```Python
>         from pixeltable import exprs
> 
>         args = [exprs.Expr.from_object(arg) for arg in args]
> ```

If the conversion isn't possible you get None back, which you can't pass on as-is.

Salvage the corresponding error-checking functionality of FunctionCall.normalize_args() and move it here?
___
*[`pixeltable/func/function.py` line 325 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKZhu0-3EAKHNVgYg5X:-OKZhu0-3EAKHNVgYg5Y:bp22kvg) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/func/function.py#L325)):*
> ```Python
>                 raise excs.Error(f'Unknown parameter: {k}')
>             param = self.signature.parameters[k]
>             expr = exprs.Expr.from_object(v)
> ```

This is actually incorrect. For using(), the args need to be constants, ie, only Literals should be allowed here. Other non-constant exprs will blow up at runtime.
___
*[`pixeltable/exprs/function_call.py` line 28 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKXweRQ8zvFowq7U5hK:-OKXweRQ8zvFowq7U5hL:b-tai944) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/exprs/function_call.py#L28)):*
> ```Python
>     resource_pool: Optional[str]
> 
>     arg_idxs: list[int]
> ```

Preface w/ comment

pixeltable/func/function.py
```diff
@@ -185,49 +191,86 @@ def _bind_to_signature(self, signature_idx: int, args: Sequence[Any], kwargs: di
...
-        return self._conditional_return_type(**kw_args)
+                arg = bound_args[param.name]
+            elif self.signature.parameters[param.name].has_default():
+                arg = self.signature.parameters[param.name].default
+            else:
+                return None
+
+            if isinstance(arg, exprs.Literal):
+                callable_args[param.name] = arg.val
+            elif isinstance(arg, exprs.Expr):
```
mkornacker: This is incorrect (and contradicts the docstring), it'll break some of the conditional_return_type functions in image.py (eg, for get_channel()).

Apparently that's not covered by tests, would be good to add some.

pixeltable/func/function.py
```diff
@@ -185,49 +191,86 @@ def _bind_to_signature(self, signature_idx: int, args: Sequence[Any], kwargs: di
...
+            else:
+                # A conditional return type is specified and all its arguments are constants; use the specific
+                # call return type
+                return_type = self._conditional_return_type(**crt_kwargs)
+
+        if return_type.nullable:
+            return return_type
+
+        # If `return_type` is non-nullable, but the function call has a nullable input to any of its non-nullable
+        # parameters, then we need to make it nullable. This is because Pixeltable defaults a function output to
```
mkornacker, line 242: I don't think this logic is needed. From the function's perspective, it's not returning a None. The call is simply skipped in FunctionCall.eval(). If the sole parameter is a required image and the arg is a non-computed colref, it could still be None (and your return type would not be nullable).

aaron-siegel: But in this case we need to set the column type to `Optional`, right? The `return_type` will drive the `col_type` of the corresponding computed column. So if I have

`f(a: str) -> str`

and I put

`t.add_computed_column(new_col=f(old_col))`

where `old_col` has type `Optional[String]`, then `new_col` also needs to have type `Optional[String]`. The eval could output a `None`, so it doesn't make sense for `new_col` to have type `Required[String]`.

[Note: This logic isn't new to this PR, I just moved it around.]

I suppose we could draw a distinction between the "return type" and the "implied column type" or some such, where in this case the return type is `Required[String]` and the column type is `Optional[String]`. But I'm not sure why that'd be important; it's the latter that's relevant from the perspective of the `FunctionCall` as an expression.

pixeltable/func/function.py
aaron-siegel: Wait, what? You can have a conditional_return_type that takes non-constant arguments? Ok, I see the `getchannel` example now.

This makes the semantics of conditional_return_type *really* mungy. What happens if the argument to getchannel() is a `Literal`? Then it gets unpacked, because that's what we do with `Literal`s, but conditional_return_type still treats it as an `Expr`, and I imagine the result is a gnarly exception.

I'm not sure how best to fix this. I think the old behavior was broken too, in the opposite direction (though I'd have to go back and check) - if you passed a non-Literal as a `model_id` to an embedding function, for example, then conditional_return_type would pretend it was a constant and blow up.

One solution could be to trigger on the type hints of the conditional_return_type function to determine which type of call pattern it expects. That seems like a separate problem and I agree it should be improved, but maybe we should factor that out of this PR? In the meantime, I prefer the new behavior since at least it doesn't cause anything to blow up; it just skips calling conditional_return_type in a few places where it should. Are you ok with shipping this and fixing it in a separate PR?

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/481)* status: 0 of 29 files reviewed, 8 unresolved discussions (waiting on @mkornacker)

___
*[`pixeltable/catalog/view.py` line 142 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKXCerH2neW2IYIB_BO:-OKbKydg6lgX6w2jg_sx:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/catalog/view.py#L142)):*

bound_args changes meaning a few times here, might be better to introduce a new variable at this point.
</blockquote></details>

Done.
___
*[`pixeltable/exprs/function_call.py` line 28 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKXweRQ8zvFowq7U5hK:-OKbM3aZCtAT3NWu9YG5:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/exprs/function_call.py#L28)):*

Preface w/ comment
</blockquote></details>

Done.
___
*[`pixeltable/func/function.py` line 157 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKYoVbu6wbdKE2JLv6j:-OKcUPb6Ao8LOaeyMhVY:b-kmr861) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/func/function.py#L157)):*

If the conversion isn't possible you get None back, which you can't pass on as-is.

Salvage the corresponding error-checking functionality of FunctionCall.normalize_args() and move it here?
</blockquote></details>

Good catch. I added a validation step for this.
___
*[`pixeltable/func/function.py` line 325 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKZhu0-3EAKHNVgYg5X:-OKcPflN16_J7ocI-M5z:b-4ibom2) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/func/function.py#L325)):*

This is actually incorrect. For using(), the args need to be constants, ie, only Literals should be allowed here. Other non-constant exprs will blow up at runtime.
</blockquote></details>

True. Not exactly incorrect, there's just a missing validation step. Added a handler for it (and a corresponding test).
___
*[`pixeltable/func/signature.py` line 155 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKXzFR5-SVcPj84K38-:-OKcR7Df5clZhzS0h1u6:b-v7pa00) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/func/signature.py#L155)):*

This function can generate user-facing errors. Should it also handle the 'invalid parameter name' case?
</blockquote></details>

It shouldn't be possible for an invalid parameter name to appear here, since `bound_args` have already been passed through `.bind()` against the corresponding Python signature.

I'll add an assertion though, to make this clear.
___
*[`pixeltable/func/signature.py` line 165 at r2](https://reviewable.io/reviews/pixeltable/pixeltable/481#-OKY-ksbA5dG5rjXWYMl:-OKcRcTuDFCSNHBE-uVF:b-896fix) ([raw file](https://github.com/pixeltable/pixeltable/blob/34a1cce8089e5ab5e4af7a54fa2044b4e4b0e7ec/pixeltable/func/signature.py#L165)):*

Clearer explanation: (since FunctionCall.eval() checks for that and skips the evaluation, if present)
</blockquote></details>

Done.

pixeltable/func/function.py
```diff
@@ -246,29 +256,58 @@ def _assemble_callable_args(
...
-            if isinstance(arg, exprs.Literal):
-                callable_args[param.name] = arg.val
-            elif isinstance(arg, exprs.Expr):
-                return None
+            assert isinstance(arg, exprs.Expr)
+
+            expects_expr: Optional[type[exprs.Expr]] = None
+            type_hint = callable_type_hints.get(param.name)
+            if typing.get_origin(type_hint) is not None:
+                type_hint = typing.get_origin(type_hint)
```
mkornacker: What was this again? Brief comment would be helpful to the reader.

pixeltable/exprs/function_call.py
```diff
@@ -27,135 +25,86 @@ class FunctionCall(Expr):
...
+        self.components.extend(arg.copy() for arg in kwargs.values())
+        self.kwarg_idxs = {name: i + len(args) for i, name in enumerate(kwargs.keys())}
+
+        # Now generate bound_idxs for the args and kwargs indices.
+        # This is guaranteed to work, because at this point the call has already been validated.
+        # These will be used later to dereference specific parameter values.
+        bindings = fn.signature.py_signature.bind(*self.arg_idxs, **self.kwarg_idxs)
+        self.bound_idxs = bindings.arguments
+
+        # Sepearately generate bound_args for purposes of determining the resource pool.
```
mkornacker: Separately

pixeltable/exprs/function_call.py
```diff
@@ -194,83 +142,13 @@ def _create_rowid_refs(self, tbl: catalog.Table) -> list[Expr]:
...
     def _equals(self, other: FunctionCall) -> bool:
         if self.fn != other.fn:
             return False
-        if len(self.args) != len(other.args):
+        if len(self.arg_idxs) != len(other.arg_idxs):
             return False
-        for i in range(len(self.args)):
-            if self.args[i] != other.args[i]:
+        for i in range(len(self.arg_idxs)):
+            if self.arg_idxs[i] != other.arg_idxs[i]:
```
mkornacker: kwargs?

pixeltable/exprs/function_call.py
```diff
@@ -409,53 +272,60 @@ def update(self, data_row: DataRow) -> None:
 
     def make_args(self, data_row: DataRow) -> Optional[tuple[list[Any], dict[str, Any]]]:
         """Return args and kwargs, constructed for data_row; returns None if any non-nullable arg is None."""
+        args: list[Any] = []
+        parameters_by_pos = self.fn.signature.parameters_by_pos
+        for idx in self.arg_idxs:
+            val = data_row[self.components[idx].slot_idx]
+            if (
+                val is None
+                and idx < len(parameters_by_pos)
```
mkornacker: How could it not be?

pixeltable/exprs/function_call.py
```diff
@@ -409,53 +272,60 @@ def update(self, data_row: DataRow) -> None:
...
-                assert param.kind != inspect.Parameter.VAR_POSITIONAL
-                if not param.col_type.nullable and val is None:
-                    return None
-                kwargs[param_name] = val
+        parameters = self.fn.signature.parameters
+        for param_name, idx in self.kwarg_idxs.items():
+            val = data_row[self.components[idx].slot_idx]
+            if (
+                val is None
+                and param_name in parameters
```
mkornacker: Same question here.

mkornacker (approved):

pixeltable/exprs/function_call.py
aaron-siegel: Done.

pixeltable/exprs/function_call.py
aaron-siegel: Fixed.

pixeltable/exprs/function_call.py
aaron-siegel: Done.

pixeltable/exprs/function_call.py
aaron-siegel: Done.

aaron-siegel (commented):
> *[Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/481)* status: 0 of 31 files reviewed, 13 unresolved discussions (waiting on @mkornacker)

PR #483, opened 2025-02-24
Test suite improvement: Use pytest fixtures for Hugging Face embedding models

aaron-siegel:
Factors out the HF embedding models as pytest fixtures and adds retries, which should harden another point of failure in CI.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,17 +2,19 @@ import json
 import logging
 import os
 import pathlib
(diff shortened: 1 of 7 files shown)


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/483)

mkornacker (approved):

PR #485, opened 2025-02-27
Add user and additional_md fields to metadata structs

aaron-siegel:
The new fields are currently unused; this PR introduces them so we can use them for storing metadata in the cloud.


pixeltable/metadata/converters/convert_28.py
```diff
@@ -0,0 +1,15 @@
...
+import sqlalchemy as sql
+
+from pixeltable.metadata import register_converter
+from pixeltable.metadata.schema import Dir, Table, TableSchemaVersion, TableVersion
+
+
+@register_converter(version=28)
+def _(engine: sql.engine.Engine) -> None:
+    with engine.begin() as conn:
+        conn.execute(sql.update(Dir).values(md=Dir.md.concat({'user': None, 'additional_md': {}})))
```
mkornacker, line 12: Do we need to put 'user': None in there?

mkornacker (approved):

pixeltable/metadata/converters/convert_28.py
aaron-siegel: Yes. Dataclass deserialization expects it to be present in the dict.

PR #490, opened 2025-03-04
Readme update with developer-focused content

pierrebrunelle:
Update readme to match docs and get people to understand what Pixeltable is faster and how they can use it rather than being descriptive.


aaron-siegel:
This change is [Reviewable](https://reviewable.io/reviews/pixeltable/pixeltable/490)

aaron-siegel (approved):

PR #494, opened 2025-03-07
Fail gracefully when a backwards incompatibility is detected in a computed column UDF call

aaron-siegel:
Improves error handling in the case where a computed column references a UDF, but the UDF is no longer compatible with its definition in the column, for one of the following reasons:
- The referenced UDF no longer exists (because it has been removed or renamed from the code)
- The UDF signature(s) are no longer compatible with the call pattern in the computed column
- The UDF return type is no longer compatible with the column's col_type

Previously, a table containing such a UDF would fail to load. Now, the table will load and generate a warning when it is loaded. It can be queried normally, and standard table operations (such as adding and removing columns) can be carried out normally, but attempting to insert data will result in an error.


pixeltable/exprs/expr.py
```diff
@@ -110,6 +107,18 @@ def default_column_name(self) -> Optional[str]:
         """
         return None
 
+    @property
+    def validation_error(self) -> Optional[str]:
+        for c in self.components:
+            error = c.validation_error
```
mkornacker, line 119: Missing in class-level type decls.

Also, what are the semantics of this? When does it get set?

pixeltable/func/function.py
```diff
@@ -490,3 +500,20 @@ def from_store(cls, name: Optional[str], md: dict, binary_obj: bytes) -> Functio
...
+    fn_dict: dict[str, Any]
+    errormsg: str
+
+    def __init__(self, self_path: str, fn_dict: dict[str, Any], errormsg: str):
+        super().__init__([], self_path)
+        self.fn_dict = fn_dict
+        self.errormsg = errormsg
+
+    def _as_dict(self) -> dict:
+        return self.fn_dict
```
mkornacker, line 520: Does this ever need to get serialized? I thought invalid functions are the result of a load-time resolution problem (ie, they shouldn't manifest when I want to create new stored metadata).

pixeltable/plan.py
```diff
@@ -374,6 +390,21 @@ def create_update_plan(
             recomputed_cols.update(idx_val_cols)
             # we only need to recompute stored columns (unstored ones are substituted away)
             recomputed_cols = {c for c in recomputed_cols if c.is_stored}
+
+        for col in recomputed_cols:
+            if col.value_expr is not None and not col.value_expr.is_valid:
```
mkornacker: Move this into a helper class method and call from both places?

pixeltable/exprs/expr.py
aaron-siegel: There's nothing to declare, it's not a member var. Nor is there anything to set. I'll add a docstring to clarify.

pixeltable/func/function.py
aaron-siegel: True, but we need to preserve the metadata from the original `Function`, so that if the invalid expression is restored to health in the future, we can still recover it.

So, all this is doing is writing the invalid data back to the dictionary, verbatim.

The motivating case is this:
- A function gets inadvertently moved around in code
- Catalog reloads; the enclosing `FunctionCall` is now invalid
- User says "oops" and moves the function back into place in code
- Catalog reloads again; everything should look fine, as if the whole round trip were a no-op

The mechanism illustrated here ensures that this happens: we're not writing out any reference to `InvalidFunction`, but rather the underlying metadata that failed to load.

pixeltable/plan.py
aaron-siegel: Done.

mkornacker (approved):

PR #496, opened 2025-03-08
[OpenAI/Anthropic] Support multiple invocations of the same tool in tool calling logic

aaron-siegel:
Correctly handles the case where an LLM invokes a tool multiple times in the same response. The tool result dicts now have the form:
```
{'tool_a': [r1, r2, r3], 'tool_b': [s1, s2]}
```
For each tool, the results will be in the same order in which they appear in the LLM's response dict.

Includes several (unrelated) bugfixes to `JsonMapper` that were necessary to get this working:
- Fixes a bug in which JsonMapper results were evaluated incorrectly
- Fixes a bug that prevented creation of a computed column whose expression contains a JsonMapper
- Fixes a bug in the evaluation logic for expressions containing more than one JsonMapper

Also:
- Fixes a bug in the `__repr__()` method of `JsonPath`
- Improves error reporting when an `ExprEvalError` occurs (the stack trace is now surfaced to the user).


jacobweiss2305:
Hey @aaron-siegel , anthropic is still calling one tool:

```
from typing import Dict, List
import pixeltable as pxt
from pixeltable.functions.anthropic import invoke_tools, messages

@pxt.udf
def get_stock_info(ticker: str) -> Dict:
    """Get basic information about a stock."""
    return "stonk is up"

tools = pxt.tools(get_stock_info)

t = pxt.create_table('test_tbl', {'prompt': pxt.String}, if_exists='replace_force')
msgs = [{'role': 'user', 'content': t.prompt}]
t.add_computed_column(
    response=messages(
        model='claude-3-7-sonnet-latest', messages=msgs, tools=tools
    )
)
t.add_computed_column(tool_calls=invoke_tools(tools, t.response))
t.insert(prompt='What is the price of FDS stock? What is the price of TSLA stock?')
res = t.select(t.response, t.tool_calls).head()

print(res)
```

My implementation of tool calling for anthropic is here and it works for the prompt above:

https://github.com/pixeltable/pixelagent/tree/main/pixelagent/anthropic

jacobweiss2305:
Openai is working as expected

aaron-siegel:
Anthropic works for me. I added a test to `test_anthropic.py` to confirm, and it passes.

Did you try running multiple times? Might be flakiness on Anthropic's end. Also try with 3.5? Hopefully the output format didn't change between 3.5 and 3.7... :/

> Hey @aaron-siegel , anthropic is still calling one tool:

jacobweiss2305:
Okay this is working for me. Thanks!

pixeltable/exprs/json_mapper.py
```diff
@@ -29,12 +33,18 @@ def __init__(self, src_expr: Expr, target_expr: Expr):
 
         from .object_ref import ObjectRef
 
-        scope_anchor = ObjectRef(self.target_expr_scope, self)
-        self.components = [src_expr, target_expr, scope_anchor]
+        self.components = [src_expr, target_expr]
         self.parent_mapper: Optional[JsonMapper] = None
```
mkornacker: Can you add class-level type decls? I've been retrofitting them whenever I touch a class.

pixeltable/func/tools.py
```diff
@@ -48,22 +48,27 @@ def ser_model(self) -> dict[str, Any]:
...
-        kwargs = {param.name: self.__extract_tool_arg(param, tool_calls) for param in self.parameters.values()}
-        return self.fn(**kwargs)
+    # The output of `tool_calls` must be a dict in standardized tool invocation format:
+    # {tool_name: [{'args': {name1: value1, name2: value2, ...}}, ...], ...}
+    def invoke(self, tool_calls: 'exprs.Expr') -> 'exprs.Expr':
+        from ..exprs import RELATIVE_PATH_ROOT as R
 
-    def __extract_tool_arg(self, param: Parameter, tool_calls: 'exprs.Expr') -> 'exprs.Expr':
         func_name = self.name or self.fn.name
+        return tool_calls[func_name]['*'] >> self.__invoke_kwargs(R.args)
```
mkornacker: Construct a JsonMapper instance here directly? Probably easier to read.

mkornacker (approved):

pixeltable/exprs/json_mapper.py
aaron-siegel: Done.

pixeltable/func/tools.py
aaron-siegel: Done.

PR #497, opened 2025-03-08
Catalog changes to prepare for concurrency

mkornacker:
- removes path caching from catalog: path resolution is done for each catalog lookup
- new class TableVersionHandle, which store references to TableVersions
- stored references to TableVersions are replaced with TableVersionHandles
- records the current SQLAlchemy connection/session in Env and adds Env.begin() to create transactions


pixeltable/catalog/catalog.py
```diff
@@ -2,37 +2,51 @@
...
 import dataclasses
 import logging
-from typing import Optional
+from typing import Optional, Type
 from uuid import UUID
 
 import sqlalchemy as sql
-import sqlalchemy.orm as orm
 
+# This import must go last to avoid circular imports.
```
aaron-siegel: Either this comment is wrong, or there's a problem

pixeltable/catalog/catalog.py
```diff
@@ -2,37 +2,51 @@
...
 from .table_version_path import TableVersionPath
 
-# This import must go last to avoid circular imports.
-import pixeltable.env as env  # isort: skip
+# from .. import InsertableTable
 
 _logger = logging.getLogger('pixeltable')
 
 
+def _join_path(path: str, name: str) -> str:
```
aaron-siegel, line 27: Comment on code organization. It's more logical to me to put utilities like this at the bottom of the file, rather than the top; I find them distracting here.

pixeltable/catalog/catalog.py
```diff
@@ -41,150 +55,343 @@ def clear(cls) -> None:
...
+        self._tbls = {}  # don't use a defaultdict here, it doesn't cooperate with the debugger
+        self._init_store()
 
-        self.tbls: dict[UUID, Table] = {}  # don't use a defaultdict here, it doesn't cooperate with the debugger
-        self.tbl_dependents: dict[UUID, list[Table]] = {}
+    def get_dir_path(self, dir_id: UUID) -> str:
+        """Return path for directory with given id"""
+        session = env.Env.get().session
+        names: list[str] = []
+        while True:
```
aaron-siegel, line 66: I feel like we should be caching more of this stuff. It'd make sense to me to have a whole bunch of local caches sitting around for these types of resolutions. Then somewhere in the DB would be a "global catalog version" that gets incremented in a transaction with any mutable catalog operation. Most of the time (in any non-concurrent session or in any context where the concurrent processes are only doing reads), we'd only have to confirm this one value from the DB, and everything else would be cached.

Possibly you're already thinking of this as a future enhancement?

pixeltable/catalog/catalog.py
```diff
@@ -41,150 +55,343 @@ def clear(cls) -> None:
...
-            dir_md = schema.DirMd(name='', user=None, additional_md={})
-            dir_record = schema.Dir(parent_id=None, md=dataclasses.asdict(dir_md))
-            session.add(dir_record)
-            session.flush()
-            session.commit()
-            _logger.info(f'Initialized catalog')
+    @dataclasses.dataclass
+    class DirEntry:
+        dir: Optional[schema.Dir]
+        dir_entries: dict[str, Catalog.DirEntry]
```
aaron-siegel, line 85: Call this `children`?

pixeltable/catalog/table.py
```diff
@@ -140,7 +145,12 @@ def _check_is_dropped(self) -> None:
 
     def __getattr__(self, name: str) -> 'pxt.exprs.ColumnRef':
         """Return a ColumnRef for the given name."""
-        return self._tbl_version_path.get_column_ref(name)
+        col = self._tbl_version_path.get_column(name)
+        if col is None:
+            raise AttributeError(f'Column {name!r} unknown')
+        from pixeltable.exprs import ColumnRef
```
aaron-siegel: ColumnRef is already imported at top of file, no?

pixeltable/catalog/table.py
```diff
@@ -958,26 +972,32 @@ def add_embedding_index(
...
+                assert _if_exists == IfExistsParam.REPLACE or _if_exists == IfExistsParam.REPLACE_FORCE
+                self.drop_index(idx_name=idx_name)
+                assert idx_name not in self._tbl_version.get().idxs_by_name
+            from pixeltable.index import EmbeddingIndex
+
+            # create the EmbeddingIndex instance to verify args
+            idx = EmbeddingIndex(
+                col, metric=metric, embed=embedding, string_embed=string_embed, image_embed=image_embed
+            )
+            with env.Env.get().begin():
```
aaron-siegel: A begin() inside a begin()?

pixeltable/catalog/table_version.py
```diff
@@ -305,35 +303,37 @@ def create(
...
+        self.delete_md(self.id)
+        self.store_tbl.drop()
 
         # de-register table version from catalog
         from .catalog import Catalog
 
         cat = Catalog.get()
-        del cat.tbl_versions[(self.id, self.effective_version)]
-        # TODO: remove from tbl_dependents
+        cat.clear_tbl_version(self)
```
aaron-siegel: remove_table_version() ? (cf. preceding comment abt this)

pixeltable/catalog/table_version.py
```diff
@@ -513,7 +508,7 @@ def _add_index(
             schema_version_drop=None,
             records_errors=idx.records_value_errors(),
         )
-        val_col.tbl = self
+        val_col.tbl = TableVersionHandle(self.id, self.effective_version, tbl_version=self)
```
aaron-siegel: This particular construct happens often enough that I wonder if it wouldn't benefit from a shorthand, something like
`val_col.tbl = self.effective_tbl_version()`.

pixeltable/catalog/table_version.py
```diff
@@ -1235,31 +1186,35 @@ def _init_external_stores(self, tbl_md: schema.TableMd) -> None:
...
-            del self.external_stores[store_name]
-            self._update_md(time.time(), conn, update_tbl_version=False)
+        store.unlink(self)  # May result in additional metadata changes
+        del self.external_stores[store_name]
+        self._update_md(time.time(), update_tbl_version=False)
 
         if delete_external_data and isinstance(store, pxt.io.external_store.Project):
             store.delete()
 
+    def is_snapshot(self) -> bool:
```
aaron-siegel, line 1200: Why not make this a property? The usual convention is that nullary methods without side effects should be properties. We haven't been 100% consistent in the existing code but seeing all the `is_snapshot`s replaced by `is_snapshot()`s makes me feel it's a step in the wrong direction.

pixeltable/env.py
```diff
@@ -166,6 +170,34 @@ def default_time_zone(self, tz: Optional[ZoneInfo]) -> None:
...
+        assert self._current_conn is not None
+        return self._current_conn
+
+    @property
+    def session(self) -> Optional[sql.orm.Session]:
+        assert self._current_session is not None
+        return self._current_session
+
+    @contextmanager
+    def begin(self) -> Iterator[sql.Connection]:
```
aaron-siegel: Call this `transaction()`? `begin()` makes sense when it's a method of a SQL engine object, not so much when it's `Env` (what is being begun?)

aaron-siegel (commented):
I at-least-skimmed the entire PR. Couple high level thoughts and a bunch of minor (mostly nomenclature) comments.

High level thoughts:
- It's not clear to me what the TableVersionHandle -> TableVersion indirection is gaining. If TableVersion is cached in memory inside the TableVersionHandle, what does it buy us? I think a whiteboard convo would be helpful here. (Or maybe this pertains to the planned "phase 2" of concurrency?)
- This happens a lot now: `_tbl_version.get().name`, etc., the code littered with `get()`s. It's a fairly awkward pattern. It'd make most sense to me for those methods to actually be part of `TableVersionHandle`, with the actual `TableVersion` being a minimalist interface. Alternatively, provide aliases in `TableVersionHandle` for the most common ones (e.g., `TableVersionHandle.is_snapshot` as an alias for `self.get().is_snapshot`).

pixeltable/catalog/catalog.py
mkornacker: This happens so infrequently, I don't think it's worth trying to optimize it. And the cost is minimal. I measured 17usec for a local single row retrieval.

pixeltable/catalog/catalog.py
mkornacker: Removed

pixeltable/catalog/catalog.py
mkornacker: It's specifically referring to 'dir' here. And dir_children sounds weird.

pixeltable/catalog/catalog.py
mkornacker: Hm, I personally prefer it here.

pixeltable/catalog/table.py

pixeltable/catalog/table.py

pixeltable/catalog/table_version.py

pixeltable/catalog/table_version.py
mkornacker: Added a create_handle().

pixeltable/catalog/table_version.py
mkornacker: Done.

pixeltable/env.py
mkornacker: Renamed to begin_xact().

pixeltable/catalog/catalog.py
aaron-siegel: Ok

aaron-siegel (approved):

PR #499, opened 2025-03-10
Properly handle relative pathnames and filenames with unusual characters

aaron-siegel:
Fixes a bug where Pixeltable failed to accept files whose filenames contain unusual characters such as `'#'`.

Also fixes a bug where Pixeltable failed to accept relative pathnames. With this change, any relative pathnames will be converted to absolute paths on insertion.


pixeltable/utils/__init__.py
```diff
@@ -35,3 +37,22 @@ def sha256sum(path: Union[Path, str]) -> str:
             h.update(chunk)
 
     return h.hexdigest()
+
+
+def parse_file_or_url(file_or_url: str) -> Union[Path, urllib.parse.ParseResult]:
```
mkornacker: How about just parse_local_path() or something like that and return an Optional[Path]? Unions are harder to follow and calling urlparse() again isn't a big deal.

pixeltable/exprs/data_row.py
```diff
@@ -206,9 +207,10 @@ def __setitem__(self, idx: int, val: Any) -> None:
                 # local file path
                 assert self.file_urls[idx] is None and self.file_paths[idx] is None
                 if len(parsed.scheme) <= 1:
-                    self.file_urls[idx] = urllib.parse.urljoin('file:', urllib.request.pathname2url(val))
```
jpeterson-pxt, line 209: The logic above this point seems duplicative; is there a reason you didn't refactor starting with urllib.parse.urlparse(val) above?

There are two other places I was curious about as well:
share/packager.py:__process_media_url()
store.py:_move_tmp_media_file()

pixeltable/utils/__init__.py
aaron-siegel: Ok yeah this is better. Done.

pixeltable/exprs/data_row.py
aaron-siegel: The logic is subtly different, because here we need to do something different if it's a `file://` URL versus a straight-up local path. In the other cases we're given something that's known to be a URL, so again there are subtle differences.

There are probably further improvements to be made here - I think you're right that we can unify more of this logic, but that wasn't my primary goal in this PR... it needs a little care and I'm trying to avoid getting ratholed... maybe we can find some time for further improvements as a separate problem.

mkornacker (approved):

PR #503, opened 2025-03-12
WIP Start of new create_table / insert / import_* methods

WildCard932:
Methods:
- create_table (from schema and from DataFrame)
- import_rows
- import_pandas
are all redirected to the new method "create_from_import()" with the new parameters.
All tests of the redirected methods are working.

Refactoring is going on every hour or so. I'm most interested in input about the way in which the superclass TableDataSource is specialized into the specific subclasses. 

Much work remains to be done.


WildCard932:
WIP done

PR #510, opened 2025-03-17
Concurrent directory operations

mkornacker:
This adds a locking protocol that will enable concurrent Pixeltable processes to execute directory updates (add/drop entry) without corrupting the stored metadata. This covers:
create_table, create_view, create_snapshot, create_dir, move, drop_table, drop_dir
    
Not supported yet: concurrent table updates (insert/update/delete)

Also: contains logic for dealing with Postgres 'serialization failure' errors, which require retrying the transaction that got force-aborted by Postgres due to concurrent updates.


pixeltable/catalog/catalog.py
```diff
@@ -1,32 +1,70 @@
...
+import psycopg
 import sqlalchemy as sql
 
-import pixeltable.env as env
 import pixeltable.exceptions as excs
 import pixeltable.metadata.schema as schema
 from pixeltable.env import Env
+from pixeltable.iterators import ComponentIterator
 
+# from .. import DataFrame
```
aaron-siegel: ?

pixeltable/catalog/catalog.py
```diff
@@ -1,32 +1,70 @@
...
 
-def _join_path(path: str, name: str) -> str:
-    """Append name to path, if path is not empty."""
-    return name if path == '' else f'{path}.{name}'
+def _lock_str(for_update: bool) -> str:
+    return 'X' if for_update else 'S'
+
+
+def _debug_print(for_update: bool, msg: str) -> None:
+    return
```
aaron-siegel: Is this temporary code?

pixeltable/catalog/catalog.py
```diff
@@ -1,32 +1,70 @@
...
+    print(f'{datetime.datetime.now()}: {_lock_str(for_update)}: {msg}')
```
aaron-siegel: Use console logger instead of print() statements

pixeltable/catalog/catalog.py
```diff
@@ -59,61 +99,201 @@ def __init__(self) -> None:
         self._tbls = {}  # don't use a defaultdict here, it doesn't cooperate with the debugger
         self._init_store()
 
-    def get_dir_path(self, dir_id: UUID) -> str:
-        """Return path for directory with given id"""
-        session = env.Env.get().session
-        names: list[str] = []
+    def _retry_loop(self, op: Callable[[], Any]) -> Any:
```
aaron-siegel: Make this typesafe please.

```
def _retry_loop(self, op: Callable[[], T]) -> T:
```

pixeltable/catalog/catalog.py
```diff
@@ -59,61 +99,201 @@ def __init__(self) -> None:
...
     @dataclasses.dataclass
     class DirEntry:
         dir: Optional[schema.Dir]
         dir_entries: dict[str, Catalog.DirEntry]
         table: Optional[schema.Table]
 
-    def get_dir_contents(self, dir_id: UUID, recursive: bool = False) -> dict[str, DirEntry]:
+    def get_dir_contents(self, dir_path: Path, recursive: bool = False) -> dict[str, DirEntry]:
+        def op() -> dict[str, Catalog.DirEntry]:
+            with Env.get().begin_xact():
```
aaron-siegel: Most of the `op()` implementations are wrapped in a `begin_xact()`. Would it make sense to bake the `begin_xact()` into `_retry_loop()`?

pixeltable/catalog/catalog.py
```diff
@@ -59,61 +99,201 @@ def __init__(self) -> None:
...
+                    row = conn.execute(q).one()
+                    dir = schema.Dir(**row._mapping)
+                    if dir.md['name'] == '':
+                        break
+                    names.insert(0, dir.md['name'])
+                    dir_id_ = dir.parent_id
+                    assert dir_id is not None
+                return Path('.'.join(names), empty_is_valid=True)
+
+        return self._retry_loop(op)
```
aaron-siegel: Proper Python style is to use `_retry_loop` as a decorator rather than explicitly declaring `op()` and wrapping it.
```python
@self._retry_loop
def get_dir_path(...) -> Path:
    # No need to declare op() separately
```

Then decorate `self._retry_loop` itself with `@functools.wraps` to ensure metadata of the decorated methods is preserved.

pixeltable/catalog/catalog.py
```diff
@@ -384,14 +759,46 @@ def _load_tbl_version(self, tbl_id: UUID, effective_version: Optional[int]) -> O
...
                 return
             # create a top-level directory, so that every schema object has a directory
             dir_md = schema.DirMd(name='', user=None, additional_md={})
             dir_record = schema.Dir(parent_id=None, md=dataclasses.asdict(dir_md))
             session.add(dir_record)
             session.flush()
-            session.commit()
             _logger.info(f'Initialized catalog')
+
+    def _handle_path_collision(
```
aaron-siegel, line 772: I don't love `_handle_...` as a method name; it's super vague (could mean anything). `_resolve_path_collision` is clearer to me.

pixeltable/env.py
```diff
@@ -175,14 +175,16 @@ def begin_xact(self) -> Iterator[sql.Connection]:
...
+                    self._current_conn = conn
+                    self._current_session = session
                     yield conn
-                finally:
-                    self._current_session = None
-                    self._current_conn = None
+            finally:
+                self._current_session = None
+                self._current_conn = None
+                print(f'{datetime.datetime.now()}: end xact')
```
aaron-siegel: Is this meant to be temporary?

pixeltable/catalog/catalog.py
```diff
@@ -168,29 +324,232 @@ def get_schema_object(
...
+                    comment=comment,
+                    media_validation=media_validation,
+                )
+                FileCache.get().emit_eviction_warnings()
+                self._tbls[view._id] = view
+                return view
+
+        return self._retry_loop(op)
+
+    def get_table(self, path: Path) -> Optional[Table]:
```
aaron-siegel: Why does this return `Optional`?

scripts/dir_ops.py
```diff
@@ -0,0 +1,66 @@
+# Script that runs an infinite sequence of random directory operations.
```
aaron-siegel: This should go in `tools/`, not `scripts/` (`scripts` is meant for bash scripts and isn't covered under Python QA in CI)

Also, I'd prefer a more descriptive filename (`run_random_dir_ops.py`?)

aaron-siegel (changes requested):

pixeltable/catalog/catalog.py
mkornacker: Commented out and added comment to remove it entirely after the next round of changes.

pixeltable/catalog/catalog.py
mkornacker: Good idea.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/env.py
mkornacker: Added comment

pixeltable/catalog/catalog.py
mkornacker: I don't think 'resolve' is any more descriptive, so I'll leave it as-is.

pixeltable/catalog/catalog.py

scripts/dir_ops.py
mkornacker: Done

pixeltable/catalog/catalog.py
```diff
@@ -1,32 +1,95 @@
...
+
+
+def _retry_loop(op: Callable[..., T]) -> Callable[..., T]:
+    @functools.wraps(op)
+    def loop(*args: Any, **kwargs: Any) -> T:
+        num_remaining_retries = _MAX_RETRIES
+        while True:
+            try:
+                # in order for retry to work, we need to make sure that there aren't any prior db updates
+                # that are part of an ongoing transaction
```
aaron-siegel, line 80: Is this going to cause a problem with user-defined transactions? (i.e., what if a transaction block surrounds two user-level catalog operations?)

pixeltable/catalog/catalog.py
```diff
@@ -59,61 +122,177 @@ def __init__(self) -> None:
...
-            dir = session.query(schema.Dir).filter(schema.Dir.id == dir_id).one()
+            q = sql.select(schema.Dir).where(schema.Dir.id == dir_id_)
+            # _debug_print(for_update=False, msg=f'dir id={dir_id}')
+            row = conn.execute(q).one()
+            dir = schema.Dir(**row._mapping)
             if dir.md['name'] == '':
                 break
             names.insert(0, dir.md['name'])
-            dir_id = dir.parent_id
+            dir_id_ = dir.parent_id
```
aaron-siegel: What's the relationship between `dir_id` and `dir_id_` here? Seems strange do be doing this and then asserting `dir_id` is not none?

aaron-siegel (approved):

pixeltable/catalog/catalog.py
mkornacker: It won't. User-level transactions will only be applicable to data operations.

pixeltable/catalog/catalog.py
mkornacker: Fixed.

PR #511, opened 2025-03-18
Refactor Config into a separate abstraction distinct from Env

aaron-siegel:
This makes it possible to access the Pixeltable Config before Env has initialized.

### DIFF
--- a/pixeltable/config.py
+++ b/pixeltable/config.py
@@ -25,13 +25,35 @@ class Config:
 
     __home: Path
     __config_file: Path
-    __config: dict[str, Any]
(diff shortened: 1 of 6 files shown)


pixeltable/config.py
```diff
@@ -0,0 +1,103 @@
...
+import toml
+
+from pixeltable import exceptions as excs
+
+_logger = logging.getLogger('pixeltable')
+
+T = TypeVar('T')
+
+
+class Config:
```
mkornacker, line 18: Move to utils?

mkornacker (approved):

PR #512, opened 2025-03-18
Bugfixes

aaron-siegel:
Fixes two Pixeltable bugs that occurred sporadically:
- Calls to `Env.console_logger` during metadata upgrades sometimes resulted in circular `Env` initialization
- Interrupting the `test_db_migration` unit test could lead to unresolvable failures on subsequent runs due to UUID collisions in table IDs

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -62,6 +62,9 @@ def init_env(tmp_path_factory, worker_id) -> None:
     # don't have several processes trying to initialize pgserver in parallel.
     root_tmp_dir = tmp_path_factory.getbasetemp().parent
     with FileLock(str(root_tmp_dir / 'pxt-init.lock')):
+        # We need to call `Env._init_env()` with `reinit_db=True`. This is because if a previous test run was
+        # interrupted (e.g., by an inopportune Ctrl-C), there may be residual DB artifacts that interfere with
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -61,6 +62,7 @@ def init_env(tmp_path_factory, worker_id) -> None:
     # don't have several processes trying to initialize pgserver in parallel.
     root_tmp_dir = tmp_path_factory.getbasetemp().parent
     with FileLock(str(root_tmp_dir / 'pxt-init.lock')):
+        Env._init_env(reinit_db=True)
```
mkornacker, line 68: Why do we need to call this explicitly now()?

aaron-siegel: In order to set `reinit_db=True` when it gets called. Otherwise there can be stale db artifacts lying around when the tests start running, if they were previously interrupted with Ctrl-C.

I think this is related to the issue Jim's been running into.

mkornacker: Leave that as a comment? It's not immediately obvious.

mkornacker (approved):

tests/conftest.py
aaron-siegel: Added.

PR #514, opened 2025-03-19
Introduce `pxtf.map()` as a replacement for >> to create JsonMappers

aaron-siegel:
This introduces a `pxtf.map()` operator that uses lambdas to construct JsonMappers. In particular, the following:

```python
pxtf.map(my_expr, lambda x: x + 1)
```
has the identical meaning to:
```python
from exprs import RELATIVE_PATH_ROOT as R
my_expr >> R + 1
```

With this change, `RELATIVE_PATH_ROOT` becomes a purely internal implementation detail, no longer exposed to the user.

### DIFF
--- a/pixeltable/functions/globals.py
+++ b/pixeltable/functions/globals.py
@@ -1,14 +1,13 @@
 import builtins
 import typing
-
-from typing import _GenericAlias  # type: ignore[attr-defined]  # isort: skip
 from typing import Any, Callable, Optional, Union
 
 import sqlalchemy as sql
 
-import pixeltable.type_system as ts
-from pixeltable import exprs, func
+from pixeltable import exceptions as excs, exprs, func, type_system as ts
 from pixeltable.utils.code import local_public_names
+
+from typing import _GenericAlias  # type: ignore[attr-defined]  # isort: skip
 
 
 # TODO: remove and replace calls with astype()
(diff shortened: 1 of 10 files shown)


pixeltable/functions/globals.py
```diff
@@ -168,6 +168,11 @@ def _(val: sql.ColumnElement) -> Optional[sql.ColumnElement]:
     return sql.sql.func.avg(val)
 
 
+def map(expr: exprs.Expr, fn: Callable[[exprs.Expr], Any]) -> exprs.Expr:
+    target_expr = exprs.Expr.from_object(fn(exprs.json_path.RELATIVE_PATH_ROOT))
```
mkornacker: This should generate a user-relatable error if the expression cannot be generated.

I'm a bit worried that a lambda looks very general, but really needs to be fairly specific.

aaron-siegel: Ok. Note that `@pxt.query` has the identical problem (arguably much worse in that case, since it's a always full `def`, rather than a little lambda). So does `@expr_udf`, of course, but we're already moving away from that one.

mkornacker: Agreed, and that's why expr_udf is mostly hidden away. In this case, I think you can create a somewhat user-friendly error message.

mkornacker: I don't think this needs to be elaborate, just somewhat more informative than some obscure stack trace.

Also, it would be good to add a test for that.

mkornacker (approved):

pixeltable/functions/globals.py
aaron-siegel: Done.

PR #522, opened 2025-03-20
Fold import operations into table_create and insert methods.

jpeterson-pxt:
New create_table / insert / import_* methods.
    
        Add support for all original table.insert() cases.
        Add support, tests for insert of json, pandas, csv, excel.
        Add support, tests for insert of HuggingFace and Parquet.
        Support new catalog interface. Add doc to methods.
        Redirect import_* methods to create_table.
        Add validator for source_format.
        Create datarows.py with rows and json importers.
        Remove direct imports of hg datasets.
        Setup new create_table signature and tests.
        Clean up
        Make io/globals.py lint-free.
    Update relevant signatures, overloads, add documentation. Expand linting.
    Disallow python keywords as PXT column names.


Makefile
```diff
@@ -140,7 +140,7 @@ docstest: install
 .PHONY: lint
 lint: install
 	@echo "Running ruff check ..."
-	@ruff check pixeltable/*.py pixeltable/functions pixeltable/index pixeltable/iterators pixeltable/metadata
+	@ruff check pixeltable/*.py pixeltable/functions pixeltable/index pixeltable/iterators pixeltable/metadata pixeltable/io/globals.py pixeltable/io/hf_datasets.py pixeltable/io/datarows.py pixeltable/io/pandas.py pixeltable/io/parquet.py pixeltable/io/utils.py pixeltable/io/external_store.py
```
aaron-siegel: Reference `pixeltable/io` as a dir instead of each file separately?

Also, add this to `pytest.yml` in the appropriate place to ensure it gets picked up in CI

pixeltable/catalog/column.py
```diff
@@ -201,6 +201,11 @@ def media_validation(self) -> MediaValidation:
         assert self.tbl is not None
         return self.tbl.get().media_validation
 
+    @property
+    def is_required_for_insert(self) -> bool:
+        """Returns True if column is required when inserting rows."""
+        return not self.col_type.nullable and self.is_stored and not self.is_computed
```
aaron-siegel: I think `self.is_stored` is redundant here (a column must be either stored or computed; it can be both, but it can't be neither - right?)

pixeltable/globals.py
```diff
@@ -57,22 +58,56 @@ def _handle_path_collision(
     return None
 
 
+if TYPE_CHECKING:
+    import os
+    from pathlib import Path
+
+    import datasets  # type: ignore[import-untyped]
+
+    RowData = list[dict[str, Any]]
```
aaron-siegel, line 21: I feel like these types should be defined somewhere other than `pixeltable/globals`. We generally shouldn't be importing from `globals` back into anything else. Maybe in `catalog/types.py` or some such? @mkornacker, any opinion here?

pixeltable/globals.py
```diff
@@ -57,22 +58,56 @@ def _handle_path_collision(
...
+    TableDataSourceType = Union[
```
aaron-siegel: General Python convention is to not suffix types with `Type`. Simply `TableDataSource` (or `PxtDataSource`?) is preferred.

pixeltable/globals.py
```diff
@@ -57,22 +58,56 @@ def _handle_path_collision(
     return None
 
 
+if TYPE_CHECKING:
+    import os
```
aaron-siegel: `os` and `pathlib` don't need to be inside a `TYPE_CHECKING` block.

pixeltable/globals.py
```diff
@@ -57,22 +58,56 @@ def _handle_path_collision(
...
+        pd.DataFrame,  # pandas DataFrame
+        'datasets.Dataset',
+        'datasets.DatasetDict',  # Huggingface datasets
+    ]
+
+
 def create_table(
     path_str: str,
-    schema_or_df: Union[dict[str, Any], DataFrame],
+    schema: Union[dict[str, Any], None] = None,
```
aaron-siegel: `s/Union[..., None]/Optional[...]/`

pixeltable/globals.py
```diff
@@ -57,22 +58,56 @@ def _handle_path_collision(
...
+    source: Optional[TableDataSourceType] = None,
+    source_format: Optional[Literal['csv', 'excel', 'parquet', 'json']] = None,
+    schema_overrides: Optional[dict[str, Any]] = None,
+    on_error: Literal['abort', 'ignore'] = 'abort',
     primary_key: Optional[Union[str, list[str]]] = None,
     num_retained_versions: int = 10,
     comment: str = '',
     media_validation: Literal['on_read', 'on_write'] = 'on_write',
     if_exists: Literal['error', 'ignore', 'replace', 'replace_force'] = 'error',
+    **kwargs: Any,  # Explicit arguments to source data provider
```
aaron-siegel: I wonder if it might make more sense for this parameter to be

`source_kwargs: Optional[dict[str, Any]] = None`

instead of

`**kwargs: Any`

There are so many kw parameters that it feels like cleaner semantics. (And what if a source provider has a `comment` kwarg?)

This is not an uncommon Python pattern, especially in complicated APIs.

pixeltable/globals.py
```diff
@@ -57,22 +58,56 @@ def _handle_path_collision(
...
 ) -> catalog.Table:
     """Create a new base table.
```
aaron-siegel, line 58: The docstring would benefit from more elaboration now. In particular, I'm envisioning a clear description of the semantics and allowable call patterns of schema/source/etc.

pixeltable/globals.py
```diff
@@ -119,7 +156,23 @@ def create_table(
         Create a table with an int and a float column, and replace any existing table:
 
         >>> tbl = pxt.create_table('my_table', schema={'col1': pxt.Int, 'col2': pxt.Float}, if_exists='replace')
+
+        Create a table from a CSV file:
```
aaron-siegel, line 122: More examples would be welcome.

pixeltable/globals.py
```diff
@@ -130,45 +183,54 @@ def create_table(
...
+                primary_key=primary_key,
+                num_retained_versions=num_retained_versions,
+                comment=comment,
+                media_validation=catalog.MediaValidation.validated(media_validation, 'media_validation'),
+            )
 
-        if len(schema) == 0:
-            raise excs.Error(f'Table schema is empty: `{path_str}`')
+        assert table is not None
+        assert isinstance(table, catalog.Table)
```
aaron-siegel: I'm ok keeping it if you think it's clarifying, but, the first assertion here is redundant (if `table is None` then it will fail the `isinstance` check)

pixeltable/io/globals.py
```diff
@@ -378,3 +244,570 @@ def export_images_as_fo_dataset(
     return fo.Dataset.from_importer(
         PxtImageDatasetImporter(tbl, images, image_format, classifications=classifications, detections=detections)
     )
+
+
+# ---------------------------------------------------------------------------------------------------------
+
+if TYPE_CHECKING:
```
aaron-siegel: I feel like everything from here on down is out of place in `io/globals`. One option is to rename `datarows.py` to `data_sources.py` and move these all there.

pixeltable/io/globals.py
```diff
@@ -378,3 +244,570 @@ def export_images_as_fo_dataset(
...
+    def valid_row_batch(self) -> Iterator[RowData]:
+        if self.batch_count > 0:
+            self.batch_count -= 1
+            yield self.valid_rows
+
+
+# ---------------------------------------------------------------------------------------------------------
+
+
+class TableDataSourcePandas(TableDataSource):
```
aaron-siegel: The usual subclass naming convention would be `PandasTableDataSource` etc. Per earlier comment, I feel like we can drop the `Table` in all of these (it's a bit of a mouthful, and `PandasDataSource` seems clear to me; or do we expect to have other kinds of data sources in the future?)

pixeltable/io/globals.py
```diff
@@ -378,3 +244,570 @@ def export_images_as_fo_dataset(
...
+    pd_df: pd.DataFrame = None
+    batch_count: int = 0
+
+    @classmethod
+    def from_tds(cls, tds: TableDataSource) -> TableDataSourcePandas:
+        foo_fields = {f.name for f in fields(tds)}
```
aaron-siegel: `foo_fields`?

pixeltable/io/globals.py
```diff
@@ -378,3 +244,570 @@ def export_images_as_fo_dataset(
...
+        if self.source_column_map is None:
+            if self.src_schema_overrides is None:
+                self.src_schema_overrides = {}
+            self.src_schema = df_infer_schema(self.pd_df, self.src_schema_overrides, self.src_pk)
+            inferred_schema, inferred_pk, self.source_column_map = normalize_schema_names(
+                self.src_schema, self.src_pk, self.src_schema_overrides, False
+            )
+            return inferred_schema, inferred_pk
+        else:
+            raise AssertionError()
```
aaron-siegel: Cleaner would be to put `assert self.source_column_map is not None` at top of method, and eliminate the `if/else` block.

pixeltable/io/globals.py
```diff
@@ -378,3 +244,570 @@ def export_images_as_fo_dataset(
...
+    hf_schema: dict[str, Any] = None
+    dataset_dict: dict[str, datasets.Dataset] = None
+    hf_schema_source: dict[str, Any] = None
+
+    @classmethod
+    def from_tds(cls, tds: TableDataSource) -> 'TableDataSourceHF':
+        foo_fields = {f.name for f in fields(tds)}
+        kwargs = {k: v for k, v in tds.__dict__.items() if k in foo_fields}
+        t = cls(**kwargs)
+        if TYPE_CHECKING:
```
aaron-siegel: This isn't the intended usage of `if TYPE_CHECKING`.

A better way to handle this is to `import datasets` within the method. This method should never be invoked in contexts where `datasets` is unavailable (and if it is, raising an `ImportError` is correct behavior).

aaron-siegel (changes requested):
Preliminary review focused on code organization and nomenclature, with a few other scattered things I noticed.

I don't see many tests exercising the new `create_table` code paths. Are there plans to add them?

pixeltable/globals.py
jpeterson-pxt: I'm open to suggestions here, it was unclear to me where they belonged. 
Also, where they should be visible.
Maybe utils/type.py

pixeltable/globals.py
jpeterson-pxt: Ok, but the class implementing it is called "TableDataSource"; what are the naming conventions here?

Perhaps tabledatasource should be tabledataconveyor

pixeltable/io/globals.py
jpeterson-pxt: datarows.py is not a top-level module. Perhaps new file is best.
Put in a new file.

Makefile
jpeterson-pxt: I have questions about some of the linting issues I found in files I did not intend to change in the /io directory.

pixeltable/catalog/column.py
jpeterson-pxt: assert ?

jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/io/globals.py
jpeterson-pxt: Done

pixeltable/io/globals.py
jpeterson-pxt: Done

pixeltable/io/globals.py
jpeterson-pxt: Done

pixeltable/io/globals.py
jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/globals.py
```diff
@@ -119,7 +156,23 @@ def create_table(
         Create a table with an int and a float column, and replace any existing table:
 
         >>> tbl = pxt.create_table('my_table', schema={'col1': pxt.Int, 'col2': pxt.Float}, if_exists='replace')
+
+        Create a table from a CSV file:
+
+        >>> tbl = pxt.create_table('my_table', source='data.csv')
     """
+    if schema is not None:
```
aaron-siegel: `schema`/`source` exactly-one validation should be done through an `excs.Error`, not through assertions. (It's a user-facing API, and we don't want assertions firing in response to a "normal" user error.)

Easiest way to do the validation:
```python
if (schema is None) == (source is None):
    raise excs.Error('Exactly one of `schema` or `source` must be specified.')
```
(something like that)

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/globals.py
```diff
@@ -119,7 +156,23 @@ def create_table(
...
+        assert source is None
+        if isinstance(schema, DataFrame):
```
aaron-siegel: I'd take this out; we explicitly don't want to support this call pattern any more.

pixeltable/globals.py
```diff
@@ -119,7 +156,23 @@ def create_table(
...
+            source = schema
+            schema = None
+    if source is not None:
+        assert schema is None
+    assert schema is not None or source is not None
+
+    from pixeltable.io.globals import OnErrorParameter, TableDataSourceDF, TableDataSourceUnk
```
aaron-siegel: Move these imports to top of file (if they're causing a circular dependency, moving the type declarations out of globals as discussed is a preferred way to resolve that)

pixeltable/globals.py
```diff
@@ -130,45 +183,54 @@ def create_table(
...
-        elif isinstance(schema_or_df, DataFrameResultSet):
-            raise excs.Error(
-                '`schema_or_df` must be either a schema dictionary or a Pixeltable DataFrame. '
-                '(Is there an extraneous call to `collect()`?)'
-            )
+        dest_dir = cat.get_schema_object(str(path.parent), expected=catalog.Dir, raise_if_not_exists=True)
+        assert dest_dir is not None
+
+        primary_key = normalize_primary_key_parameter(primary_key)
+        table = None
```
aaron-siegel: best to add type hints to these decls, since their types cannot be inferred from the declarations

pixeltable/globals.py
```diff
@@ -130,45 +183,54 @@ def create_table(
...
+            data_source.infer_schema()
+            schema = data_source.pxt_schema
+            primary_key = data_source.pxt_pk
+            is_direct_df = data_source.is_direct_df()
         else:
-            raise excs.Error('`schema_or_df` must be either a schema dictionary or a Pixeltable DataFrame.')
+            is_direct_df = False
+
+        if table is None and schema is None and data_source is None:
+            raise excs.Error('Table does not exist, at least one of `schema` or `source` must be provided')
```
aaron-siegel: Pet peeve: comma splice (the comma should be a semicolon)

But I think it's best to handle this at top of method? I'd argue it should be an error if there's no schema or source, regardless of whether the table exists or not?

pixeltable/globals.py
```diff
@@ -130,45 +183,54 @@ def create_table(
...
+        if table is None and (len(schema) == 0 or not isinstance(schema, dict)):
+            raise excs.Error('Unable to create a proper schema, please supply one')
```
aaron-siegel: Comma splice :)

A more informative message might be "The table schema cannot be inferred from the given data source; please specify one via the `schema_overrides` parameter" - something like that?

Validation of the `schema` input should probably be handled as a separate check with its own (informative) error message.

pixeltable/catalog/insertable_table.py
```diff
@@ -83,65 +83,86 @@ def get_metadata(self) -> dict[str, Any]:
...
-            cols_with_excs_str = ''
-        else:
-            cols_with_excs_str = (
-                f' across {len(status.cols_with_excs)} column{"" if len(status.cols_with_excs) == 1 else "s"}'
-            )
-            cols_with_excs_str += f' ({", ".join(status.cols_with_excs)})'
-        msg = (
-            f'Inserted {status.num_rows} row{"" if status.num_rows == 1 else "s"} '
-            f'with {status.num_excs} error{"" if status.num_excs == 1 else "s"}{cols_with_excs_str}.'
+        from pixeltable.io.globals import OnErrorParameter, TableDataSourceUnk
```
aaron-siegel: `pixeltable.io` is probably not the right place for `OnErrorParameter`?

pixeltable/catalog/insertable_table.py
```diff
@@ -83,65 +83,86 @@ def get_metadata(self) -> dict[str, Any]:
...
+        data_source.prepare_for_insert_into_table()
+
+        fail_on_exception = OnErrorParameter.fail_on_exception(on_error)
+        return table.insert_table_data_source(
+            data_source=data_source, fail_on_exception=fail_on_exception, print_stats=print_stats
         )
-        Env.get().console_logger.info(msg)
-        _logger.info(f'InsertableTable {self._name}: {msg}')
+
+    if TYPE_CHECKING:
```
aaron-siegel: Move to top of file

pixeltable/catalog/table.py
```diff
@@ -1174,29 +1175,39 @@ def _drop_index(
             idx_id = idx_info[0].id
         self._tbl_version.get().drop_index(idx_id)
 
+    if TYPE_CHECKING:
```
aaron-siegel: Move to top of file. (General comment: Preferred Python style is to always put `import` statements at the top of the file unless there's a specific reason why they can't go there. And I don't think such a reason can ever exist for imports that are protected by `if TYPE_CHECKING`.)

pixeltable/catalog/table_version.py
```diff
@@ -1235,6 +1235,10 @@ def primary_key_columns(self) -> list[Column]:
         """Return all non-system columns"""
         return [c for c in self.cols if c.is_pk]
 
+    def primary_key(self) -> list[str]:
```
aaron-siegel, line 1244: Make this a `@property`?

pixeltable/catalog/globals.py
```diff
@@ -40,6 +40,21 @@ def __iadd__(self, other: 'UpdateStatus') -> 'UpdateStatus':
         self.cols_with_excs = list(dict.fromkeys(self.cols_with_excs + other.cols_with_excs))
         return self
 
+    def build_insert_msg(self) -> str:
```
aaron-siegel: I like the idea of baking this into `UpdateStatus`. An alternative would be to call it simply

```python
@property
def insert_msg(self) -> str:
```
The general guideline I prefer is: nullary methods without side effects should be properties, and their names should be nouns.

aaron-siegel (changes requested):

aaron-siegel (approved):

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done.

pixeltable/catalog/insertable_table.py
jpeterson-pxt: Done

pixeltable/catalog/table.py
jpeterson-pxt: Done

pixeltable/catalog/globals.py
jpeterson-pxt: Done

pixeltable/catalog/table_version.py
jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/catalog/insertable_table.py
jpeterson-pxt: Done

pixeltable/io/globals.py
jpeterson-pxt: I used NotImplementedError, as we intend to eventually support mapping from source to destination.

PR #525, opened 2025-03-21
PXT-499 Add create_parents option to create_dir to create missing ele…

amithadke:
…ments in the path

### DIFF
--- a/pixeltable/catalog/path.py
+++ b/pixeltable/catalog/path.py
@@ -55,5 +55,12 @@ class Path:
         is_prefix = self.components == other.components[: self.len]
         return is_prefix and (self.len == (other.len - 1) or not is_parent)
 
+    def ancestors(self):
+        """
+        Return all non-root ancestors of this path.
+        """
+        for i in range(1, len(self.components)):
+            yield Path('.'.join(self.components[0:i]))
+
     def __str__(self) -> str:
         return '.'.join(self.components)
--- a/pixeltable/catalog/path.py
+++ b/pixeltable/catalog/path.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+from typing import Generator
 
 from pixeltable import exceptions as excs
 
@@ -55,12 +56,16 @@ class Path:
         is_prefix = self.components == other.components[: self.len]
         return is_prefix and (self.len == (other.len - 1) or not is_parent)
 
-    def ancestors(self):
+    def ancestors(self) -> Generator[Path, None, None]:
         """
-        Return all non-root ancestors of this path.
+        Return all ancestors of this path in top-down order including root.
+        If this path is for the root directory, which has no parent, then None is returned.
         """
-        for i in range(1, len(self.components)):
-            yield Path('.'.join(self.components[0:i]))
+        if self.is_root:
+            yield None
+        else:
+            for i in range(0, len(self.components)):
+                yield Path('.'.join(self.components[0:i]), empty_is_valid=True)
 
     def __str__(self) -> str:
         return '.'.join(self.components)
--- a/pixeltable/catalog/path.py
+++ b/pixeltable/catalog/path.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from typing import Generator
+from typing import Iterator
 
 from pixeltable import exceptions as excs
 
@@ -56,13 +56,13 @@ class Path:
         is_prefix = self.components == other.components[: self.len]
         return is_prefix and (self.len == (other.len - 1) or not is_parent)
 
-    def ancestors(self) -> Generator[Path, None, None]:
+    def ancestors(self) -> Iterator[Path]:
         """
         Return all ancestors of this path in top-down order including root.
         If this path is for the root directory, which has no parent, then None is returned.
         """
         if self.is_root:
-            yield None
+            return
         else:
             for i in range(0, len(self.components)):
                 yield Path('.'.join(self.components[0:i]), empty_is_valid=True)
--- a/pixeltable/globals.py
+++ b/pixeltable/globals.py
@@ -573,7 +573,7 @@ def create_dir(
             - `'ignore'`: do nothing and return the existing directory handle
             - `'replace'`: if the existing directory is empty, drop it and create a new one
             - `'replace_force'`: drop the existing directory and all its children, and create a new one
-        create_parents: Create missing parent directories (defaults to `False`).
+        create_parents: Create missing parent directories.
 
     Returns:
         A handle to the newly created directory, or to an already existing directory at the path when
@@ -602,7 +602,7 @@ def create_dir(
 
         >>> pxt.create_dir('my_dir', if_exists='replace_force')
 
-        Create a sub directory along with its ancestors:
+        Create a subdirectory along with its ancestors:
 
         >>> pxt.create_dir('parent1.parent2.sub_dir', create_parents=True)
     """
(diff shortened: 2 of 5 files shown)


pixeltable/globals.py
```diff
@@ -573,6 +573,7 @@ def create_dir(
             - `'ignore'`: do nothing and return the existing directory handle
             - `'replace'`: if the existing directory is empty, drop it and create a new one
             - `'replace_force'`: drop the existing directory and all its children, and create a new one
+        create_parents: Create missing parent directories (defaults to `False`).
```
mkornacker: Leave out default, that's generated automatically.

pixeltable/globals.py
```diff
@@ -600,18 +601,34 @@ def create_dir(
         Create a directory and replace if it already exists:
 
         >>> pxt.create_dir('my_dir', if_exists='replace_force')
+
+        Create a sub directory along with its ancestors:
```
mkornacker: Spelling

pixeltable/globals.py
```diff
@@ -600,18 +601,34 @@ def create_dir(
...
+
+        >>> pxt.create_dir('parent1.parent2.sub_dir', create_parents=True)
     """
     path_obj = catalog.Path(path)
     cat = Catalog.get()
 
     with env.Env.get().begin_xact():
+        parent = cat.get_schema_object(str(path_obj.parent), raise_if_parent_missing=(not create_parents))
+        if parent is None and create_parents:
```
mkornacker: You might as well do this first, meaning check each ancestor path. If you do that, the raise_if_parent_missing flag isn't needed.

pixeltable/globals.py
```diff
@@ -600,18 +601,34 @@ def create_dir(
...
+            # start walking from the root
+            current_dir_path = catalog.Path('', empty_is_valid=True)
+            current_dir_object = cat.get_schema_object(str(current_dir_path))
+            for component in path_obj.components[:-1]:
```
mkornacker: Instead of bolting Paths together manually, evolve the Path class:
- property num_ancestors
- ancestor(n: int) -> Path, where ancestor(0) return parent, ancestor(1) returns the parent's parent, etc.

tests/test_dirs.py
```diff
@@ -221,3 +221,39 @@ def test_move(self, reset_db) -> None:
...
+        md = dir3.get_metadata()
+        assert md['path'] == 'dir1.dir2.dir3'
+        assert md['name'] == 'dir3'
+        listing = pxt.list_dirs(recursive=True)
+        assert listing == all_dirs_3
+
+        pxt.drop_dir('dir1.dir2.dir3')
+        pxt.drop_dir('dir1.dir2')
+
+        dir3 = pxt.create_dir('dir1.dir2.dir3', create_parents=True)
```
mkornacker: Comment how that's a new test case.

tests/test_dirs.py
```diff
@@ -221,3 +221,39 @@ def test_move(self, reset_db) -> None:
...
+        md = dir3.get_metadata()
+        assert md['path'] == 'dir1.dir2.dir3'
+        assert md['name'] == 'dir3'
+        listing = pxt.list_dirs(recursive=True)
+        assert listing == all_dirs_3
+
+        pxt.drop_dir('dir1.dir2.dir3')
```
mkornacker: How is this different than the last scenario?

pixeltable/globals.py
amithadke: I added ancestors() that generates ancestors starting from root level.

tests/test_dirs.py
amithadke: changed it only 2 cases there create entire path,
create partial path

pixeltable/catalog/path.py
```diff
@@ -55,5 +55,12 @@ def is_ancestor(self, other: Path, is_parent: bool = False) -> bool:
         is_prefix = self.components == other.components[: self.len]
         return is_prefix and (self.len == (other.len - 1) or not is_parent)
 
+    def ancestors(self):
```
mkornacker: Missing type decl

pixeltable/catalog/path.py
```diff
@@ -55,5 +55,12 @@ def is_ancestor(self, other: Path, is_parent: bool = False) -> bool:
...
+        """
+        Return all non-root ancestors of this path.
```
mkornacker: missing in comment: in what order

Also: why not include the root? Excluding it makes the semantics trickier.

pixeltable/globals.py
```diff
@@ -600,18 +601,32 @@ def create_dir(
...
+            # start walking from the root
+            ancestors = path_obj.ancestors()
+            root_path = catalog.Path('', empty_is_valid=True)
+            last_parent = cat.get_schema_object(str(root_path))
+            for ancestor in ancestors:
+                ancestor_obj = cat.get_schema_object(str(ancestor), expected=catalog.Dir)
+                last_parent = (
+                    catalog.Dir._create(last_parent._id, ancestor.name) if ancestor_obj is None else ancestor_obj
+                )
+            # get parent again
```
mkornacker: Again? Also, isn't this last_parent?

Shouldn't there be an else branch?

pixeltable/globals.py
```diff
@@ -600,18 +601,32 @@ def create_dir(
...
+            parent = cat.get_schema_object(str(path_obj.parent))
         if_exists_ = catalog.IfExistsParam.validated(if_exists, 'if_exists')
         existing = _handle_path_collision(path, catalog.Dir, False, if_exists_)
         if existing is not None:
             assert isinstance(existing, catalog.Dir)
             return existing
-
-        parent = cat.get_schema_object(str(path_obj.parent))
         assert parent is not None
```
mkornacker, line 630: How does this work? It doesn't look like parent even gets set when !create_parents.

pixeltable/globals.py
```diff
@@ -600,18 +601,32 @@ def create_dir(
...
+        >>> pxt.create_dir('parent1.parent2.sub_dir', create_parents=True)
     """
     path_obj = catalog.Path(path)
     cat = Catalog.get()
 
     with env.Env.get().begin_xact():
+        if create_parents:
+            # start walking from the root
+            ancestors = path_obj.ancestors()
+            root_path = catalog.Path('', empty_is_valid=True)
```
mkornacker: If ancestors() included the root, you wouldn't need this special-casing.

pixeltable/catalog/path.py
amithadke: I started with including root but then was trying to think of what would be returned in ancestor() of root path.

pixeltable/globals.py
amithadke: good catch we are not fetching parent at the beginning like earlier version of the code, will remove it and yes last parent should be the parent.

pixeltable/globals.py
amithadke: yup interestingly all tests passed, adding coverage :)

pixeltable/globals.py
amithadke: let me know what you think we should do when ancestors() called on root. for this use case its simple but for any other uses of walking down from root perhaps we can none (which means users of ancestors need to handle it)

pixeltable/globals.py
amithadke: fixed the bug

pixeltable/catalog/path.py
```diff
@@ -55,5 +56,16 @@ def is_ancestor(self, other: Path, is_parent: bool = False) -> bool:
         is_prefix = self.components == other.components[: self.len]
         return is_prefix and (self.len == (other.len - 1) or not is_parent)
 
+    def ancestors(self) -> Generator[Path, None, None]:
```
mkornacker: Simpler: Iterator[Path]

pixeltable/catalog/path.py
```diff
@@ -55,5 +56,16 @@ def is_ancestor(self, other: Path, is_parent: bool = False) -> bool:
...
+        """
+        Return all ancestors of this path in top-down order including root.
+        If this path is for the root directory, which has no parent, then None is returned.
+        """
+        if self.is_root:
+            yield None
```
mkornacker: Simply return here? root doesn't have ancestors.

mkornacker (approved):

pixeltable/catalog/path.py
```diff
@@ -55,5 +56,16 @@ def is_ancestor(self, other: Path, is_parent: bool = False) -> bool:
         is_prefix = self.components == other.components[: self.len]
         return is_prefix and (self.len == (other.len - 1) or not is_parent)
 
+    def ancestors(self) -> Iterator[Path]:
+        """
+        Return all ancestors of this path in top-down order including root.
+        If this path is for the root directory, which has no parent, then None is returned.
+        """
+        if self.is_root:
```
aaron-siegel, line 64: Why is this necessary as a special case? I'd expect that if `self.is_root`, then `len(self.components) == 0`, in which case it's handled correctly by the regular logic. (Or is there a component with value `''`?)

pixeltable/globals.py
```diff
@@ -560,7 +560,7 @@ def list_tables(dir_path: str = '', recursive: bool = True) -> list[str]:
 
 
 def create_dir(
-    path: str, if_exists: Literal['error', 'ignore', 'replace', 'replace_force'] = 'error'
+    path: str, if_exists: Literal['error', 'ignore', 'replace', 'replace_force'] = 'error', create_parents: bool = False
```
aaron-siegel: In the Python API this is called simply `parents`, not `create_parents`. We might want to consider calling it simply `parents` for consistency. This kind of makes sense, since the `create_` is implied by the method name.

See:
https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

pixeltable/catalog/path.py
amithadke: @property
 def is_root(self) -> bool:
        return self.components[0] == ''
from above I assumed that there is possibility that len is 1 with component[0]=''

PR #526, opened 2025-03-21
PXT-336: Add access method for embedding indices.

jpeterson-pxt:
See the Jira for full description.


pixeltable/exprs/column_ref.py
```diff
@@ -131,6 +131,15 @@ def similarity(self, item: Any, *, idx: Optional[str] = None) -> Expr:
 
         return SimilarityExpr(self, item, idx_name=idx)
 
+    def embedding(self, idx: Optional[str] = None) -> Expr:
```
mkornacker: Let's also make that a required kwarg.

Also, you can use the specific return type here.

pixeltable/catalog/column.py
```diff
@@ -171,6 +171,33 @@ def get_idx_info(self) -> dict[str, 'TableVersion.IndexInfo']:
         assert self.tbl is not None
         return {name: info for name, info in self.tbl.get().idxs_by_name.items() if info.col == self}
 
+    @classmethod
+    def find_embedding_index(
```
mkornacker: This is basically checking usage errors for the specific callers, which shouldn't live here. Move it  into ColumnRef.

pixeltable/exprs/column_ref.py
```diff
@@ -131,6 +131,15 @@ def similarity(self, item: Any, *, idx: Optional[str] = None) -> Expr:
 
         return SimilarityExpr(self, item, idx_name=idx)
```
mkornacker, line 159: Move the index validation logic into a private method in ColumnRef (and out of SimilarityExpr). Use that same private method for embedding().

mkornacker (changes requested):

tests/test_index.py
```diff
@@ -521,6 +521,48 @@ def test_embedding_basic(
...
+    def test_embedding_access(
+        self,
+        img_tbl: pxt.Table,
+        test_tbl: pxt.Table,
+        clip_embed: func.Function,
+        e5_embed: func.Function,
+        all_mpnet_embed: func.Function,
+        reload_tester: ReloadTester,
+    ) -> None:
+        skip_test_if_not_installed('transformers')
```
mkornacker: Create utility method in this class instead of copy & paste.

tests/test_index.py
```diff
@@ -521,6 +521,48 @@ def test_embedding_basic(
...
+        dummy_img_t = pxt.create_table('dummy', schema)
+        dummy_img_t.insert(rows[:10])
+
+        img_t.add_embedding_index('img', embedding=clip_embed)
+
+        r = img_t.select(img_t.img.embedding()).limit(2).collect()
+        assert len(r) == 2
+        assert isinstance(r[0, 'img_embedding_'], np.ndarray)
+
+        # multiple indices
```
mkornacker: Also test that this doesn't work on a snapshot.

pixeltable/exprs/column_ref.py
jpeterson-pxt: Interesting, I built it in ColumnRef, but moved it to Column because it doesn't depend on ColumnRef at all. 
I could / should have made it a method of the column, rather than a class method.

pixeltable/exprs/column_ref.py
jpeterson-pxt: Why is a kwarg required here?
Yes, specific return.

pixeltable/exprs/column_ref.py
mkornacker: The logic itself reflects the semantics of similarity() and embedding(), so it's better to have it sit in ColumnRef.

pixeltable/exprs/column_ref.py
mkornacker: The index name in similarity() is also a kwarg.

pixeltable/exprs/column_ref.py
```diff
@@ -131,6 +131,15 @@ def similarity(self, item: Any, *, idx: Optional[str] = None) -> Expr:
...
 
+    def embedding(self, idx: Optional[str] = None) -> Expr:
+        import copy
```
aaron-siegel: stale import?

tests/test_index.py
jpeterson-pxt: Done

pixeltable/exprs/similarity_expr.py
```diff
@@ -23,26 +23,12 @@ def __init__(self, col_ref: ColumnRef, item: Any, idx_name: Optional[str] = None
...
-            if idx_name is None:
-                raise excs.Error(
-                    f'Column {col_ref.col.name!r} has multiple indices; use the index name to disambiguate: '
-                    f'`{col_ref.col.name}.similarity(..., idx=<name>)`'
-                )
-            self.idx_info = embedding_idx_info[idx_name]
-        else:
-            self.idx_info = next(iter(embedding_idx_info.values()))
+        # determine index to use
+        idx_dict = ColumnRef.find_embedding_index(col_ref.col, idx_name, 'similarity')
```
mkornacker, line 29: It doesn't really make sense to do this here. ColumnRef.similarity() constructs the instance, so it might as well do the verification first and then tell SimilarityExpr what index to use. Basically what you're doing for embedding().

jpeterson-pxt: I left it this way because similiary() uses the name directly as well as using it to find the column.

pixeltable/exprs/column_ref.py
```diff
@@ -125,11 +125,47 @@ def __getattr__(self, name: str) -> Expr:
...
+        """Return IndexInfo for a column, with an optional given name"""
+        # determine index to use
+        idx_info_dict = col.get_idx_info()
+        from pixeltable import index
+
+        embedding_idx_info = {
+            info: value for info, value in idx_info_dict.items() if isinstance(value.idx, index.EmbeddingIndex)
+        }
+        if len(embedding_idx_info) == 0:
+            raise excs.Error(f'No indices found for {method_name} on column {col.name!r}')
```
aaron-siegel: `{method_name!r}` to be consistent?

aaron-siegel (approved):

pixeltable/exprs/column_ref.py
jpeterson-pxt: Its used a few lines later.

pixeltable/exprs/column_ref.py
jpeterson-pxt: Done

jpeterson-pxt: Done

jpeterson-pxt:
@mkornacker This PR is blocked on your approval; is there something else you'd like me to do here?

mkornacker (approved):

PR #530, opened 2025-03-24
PXT-465 Add support to initialize Pixeltable with database connection strings

amithadke:


pixeltable/utils/dbms_helper.py
```diff
@@ -0,0 +1,16 @@
+import abc
```
aaron-siegel: Do we really need three different `.py` files for these classes? Maybe put all three classes in a single `.py` file.

pixeltable/utils/dbms_helper.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class DbmsHelperBase(abc.ABC):
```
aaron-siegel: Omit `Base` in the classname here (just `DbmsHelper` would be more conventional).

I also don't love "Helper"; how about just `Dbms` as the class name? (With implementations `CockroachDbms` and `PostgresqlDbms`, etc.)

pixeltable/utils/dbms_helper.py
```diff
@@ -0,0 +1,16 @@
...
+    def __init__(self, name: str, transaction_isolation_level: str, vmin_vmax_index_type: str) -> None:
+        self.name = name
+        self.transaction_isolation_level = transaction_isolation_level
+        self.vmin_vmax_index_type = vmin_vmax_index_type
```
aaron-siegel: `version_index_type`?

pixeltable/utils/dbms_helper.py
```diff
@@ -0,0 +1,16 @@
...
+
+    @abc.abstractmethod
+    def build_drop_db_stmt(self, database: str) -> str:
```
aaron-siegel: The "modern" Python convention is to use an ellipsis for abstract methods rather than `pass`:

```python
@abstractmethod
def build_drop_db_stmt(self, database: str) -> str: ...
```

pixeltable/env.py
```diff
@@ -339,17 +349,36 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
...
 
-        # cleanup_mode=None will leave the postgres process running after Python exits
-        # cleanup_mode='stop' will terminate the postgres process when Python exits
-        # On Windows, we need cleanup_mode='stop' because child processes are killed automatically when the parent
-        # process (such as Terminal or VSCode) exits, potentially leaving it in an unusable state.
-        cleanup_mode = 'stop' if platform.system() == 'Windows' else None
-        self._db_server = pixeltable_pgserver.get_server(self._pgdata_dir, cleanup_mode=cleanup_mode)
-        self._db_url = self._db_server.get_uri(database=self._db_name, driver='psycopg')
+        db_connect_str = config.get_string_value('DB_CONNECT_STR')
+        db_default_connect_str = config.get_string_value('DEFAULT_DB_CONNECT_STR')
```
aaron-siegel: What's the relationship between `DEFAULT_DB_CONNECT_STR` and `DB_CONNECT_STR`? Some code comments that explain what's going on would be helpful.

pixeltable/env.py
```diff
@@ -339,17 +349,36 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
...
+        if db_default_connect_str is not None:
```
aaron-siegel: What if `DB_CONNECT_STR` is set, but not `DEFAULT_DB_CONNECT_STR`?

pixeltable/env.py
```diff
@@ -388,12 +417,24 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
         self._set_up_runtime()
         self.log_to_stdout(False)
 
+    def get_defaultdb_url(self):
```
aaron-siegel:
```python
@property
def default_db_url
```

aaron-siegel (changes requested):

pixeltable/env.py
amithadke: our code currently checks if db exists using DEFAULT_DB_CONNECT_STR
for tests we only need to give DEFAULT_DB_CONNECT_STR and db name (which is f'test_{worker_id}')

for cloud we are going to manually create databases and set DB_CONNECT_STR per user.
There will be one db for public and specific ones for paid customers (different user names and password and perhaps certs for each of them)

pixeltable/env.py
amithadke: thanks will do

pixeltable/utils/dbms_helper.py
amithadke: sounds good

pixeltable/catalog/table_version.py
```diff
@@ -341,7 +341,8 @@ def _init_cols(self, tbl_md: schema.TableMd, schema_version_md: schema.TableSche
         self.cols = []
         self.cols_by_name = {}
         self.cols_by_id = {}
-        for col_md in tbl_md.column_md.values():
+        sorted_column_md = sorted(tbl_md.column_md.values(), key=lambda item: item.id)
```
aaron-siegel, line 345: Why are we sorting these here? I'd think we'd want them to be in table order.

pixeltable/env.py
```diff
@@ -346,16 +354,39 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
...
+                self._db_name = db_url.database  # use the dbname given in connect string
+                dialect = db_url.get_dialect().name
+                if dialect == 'cockroachdb':
+                    self._dbms = CockroachDbms(db_url)
+                elif dialect == 'postgresql':
+                    self._dbms = PostgresqlDbms(db_url)
+                else:
+                    raise excs.Error("Unsupported DBMS '%s'" % dialect)
+                if not self._store_db_exists():
+                    error = f'Database "{self._db_name}" does not exist'
```
aaron-siegel: Best / most consistent practice for quoting entities in string messages:

```python
error = f'Database {self._db_name!r} does not exist'
```

pixeltable/env.py
```diff
@@ -346,16 +354,39 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
...
+                db_url = sql.make_url(db_connect_str)
+                self._db_url = db_url.render_as_string(hide_password=False)
+                self._db_name = db_url.database  # use the dbname given in connect string
+                dialect = db_url.get_dialect().name
+                if dialect == 'cockroachdb':
+                    self._dbms = CockroachDbms(db_url)
+                elif dialect == 'postgresql':
+                    self._dbms = PostgresqlDbms(db_url)
+                else:
+                    raise excs.Error("Unsupported DBMS '%s'" % dialect)
```
aaron-siegel: Use f-specifier

```python
f'Unsupported DBMS: {dialect}'
```
The % syntax is super old-school python

tests/test_dbms.py
```diff
@@ -0,0 +1,47 @@
...
+
+import pixeltable as pxt
+from pixeltable.env import Env
+
+from .utils import reload_catalog
+
+
+class TestDbms:
+    def test_create_table(self):
+        if os.environ.get('PIXELTABLE_DB_CONNECT_STR') is None:
```
aaron-siegel: Get this from pixeltable config, not environ?

tests/test_dbms.py
```diff
@@ -0,0 +1,47 @@
...
+            logging.error('PIXELTABLE_DB_CONNECT_STR is not set, skipping test')
+            return
```
aaron-siegel: Use pytest.skip(); otherwise pytest will mark this test as PASSED. (There are several other tests that have internal skip conditions that you can refer to as examples)

tests/test_dbms.py
```diff
@@ -0,0 +1,47 @@
...
+import pixeltable as pxt
+from pixeltable.env import Env
+
+from .utils import reload_catalog
+
+
+class TestDbms:
+    def test_create_table(self):
+        if os.environ.get('PIXELTABLE_DB_CONNECT_STR') is None:
+            logging.error('PIXELTABLE_DB_CONNECT_STR is not set, skipping test')
```
aaron-siegel: No need for this; it isn't necessarily an error. If you use pytest.skip() (see next comment) then you can report all the relevant info that way.

tests/test_dbms.py
```diff
@@ -0,0 +1,47 @@
...
+
+import pytest
+
+import pixeltable as pxt
+from pixeltable.env import Env
+
+from .utils import reload_catalog
+
+
+class TestDbms:
```
aaron-siegel: I put some comments below, but now I'm really confused about when/in what context this test is meant to run. If DB_CONNECT_STR is set, then that will impact the entire test suite, correct? So this test doesn't really isolate/test anything independently. And it seems to be testing general table ops, not anything specific to DBMS config.

aaron-siegel (changes requested):

tests/test_dbms.py
amithadke: Removed the test

Since conftest was setting reinit_db to true it was blowing database before every run.
In the case when db_connect_str is used, we don't want it to blow database but we can always drop tables, so made that change to drop tables in db.

Now to run a specific test for cockroach all u need is set db_connect_str in config and pick which test you want to run
test_dirs.py  all tests will pass 
37 tests out of test_table.py will pass, 18 will fail
Out of 18 some((test_drop_column) are failing  because jsonb_path_query() not supported in v24 of cockroachdb. Once we get to upgrade our cluster to v25, they 'should' work.

pixeltable/catalog/table_version.py
mkornacker, line 345: missing comment: why this needs to be sorted.

pixeltable/env.py
```diff
@@ -25,11 +25,13 @@
 
 import pixeltable_pgserver
 import sqlalchemy as sql
+from sqlalchemy.exc import ArgumentError
```
mkornacker: just use sql.exc.ArgumentError instead, makes the code easier to follow

pixeltable/env.py
```diff
@@ -346,16 +354,39 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
...
-        self._pgdata_dir = Path(os.environ.get('PIXELTABLE_PGDATA', str(Config.get().home / 'pgdata')))
-
-        # cleanup_mode=None will leave the postgres process running after Python exits
-        # cleanup_mode='stop' will terminate the postgres process when Python exits
-        # On Windows, we need cleanup_mode='stop' because child processes are killed automatically when the parent
-        # process (such as Terminal or VSCode) exits, potentially leaving it in an unusable state.
-        cleanup_mode = 'stop' if platform.system() == 'Windows' else None
-        self._db_server = pixeltable_pgserver.get_server(self._pgdata_dir, cleanup_mode=cleanup_mode)
-        self._db_url = self._db_server.get_uri(database=self._db_name, driver='psycopg')
+        db_connect_str = config.get_string_value('DB_CONNECT_STR')
```
mkornacker: move into method

pixeltable/env.py
```diff
@@ -346,16 +354,39 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
...
+        if db_connect_str is not None:
+            try:
+                db_url = sql.make_url(db_connect_str)
+                self._db_url = db_url.render_as_string(hide_password=False)
+                self._db_name = db_url.database  # use the dbname given in connect string
+                dialect = db_url.get_dialect().name
+                if dialect == 'cockroachdb':
+                    self._dbms = CockroachDbms(db_url)
+                elif dialect == 'postgresql':
```
mkornacker: What is this usage scenario? This is clearly making assumptions that aren't valid in the local case.

pixeltable/env.py
```diff
@@ -346,16 +354,39 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
...
+                    self._dbms = PostgresqlDbms(db_url)
+                else:
+                    raise excs.Error(f'Unsupported DBMS {dialect}')
+                if not self._store_db_exists():
+                    error = f'Database {self._db_name!r} does not exist'
+                    self._logger.error(error)
+                    raise excs.Error(error)
+                self._logger.info(f'found database at: {self.db_url}')
+            except ArgumentError as e:
```
mkornacker: Which sqlalchemy call can raise this? A tighter try/except would make the code easier to follow.

pixeltable/env.py
```diff
@@ -369,7 +400,10 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
                     self._logger.error(f'Invalid time zone specified in configuration: {tz_name}')
 
         if reinit_db and self._store_db_exists():
-            self._drop_store_db()
+            if db_connect_str is not None:
+                self._delete_all_tables()
```
mkornacker: Again, there are assumptions here that aren't spelled out. Probably best to do in the class comment.

Also: this seems *extremely* dangerous in the cloud scenario.

pixeltable/env.py
```diff
@@ -394,12 +429,21 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
         self._set_up_runtime()
         self.log_to_stdout(False)
 
+    @property
+    def default_system_db_url(self):
+        if self._db_server is None:
+            return self._dbms.default_system_db_url()
+        else:
+            return self._db_server.get_uri(database='postgres', driver='psycopg')
```
mkornacker: Why not just use dbms.default_system_db_url here as well?

pixeltable/env.py
```diff
@@ -423,46 +467,50 @@ def _store_db_exists(self) -> bool:
...
-                    "ENCODING 'utf-8' LC_COLLATE 'C' LC_CTYPE 'C' TEMPLATE template0"
-                )
+                stmt = self._dbms.build_create_db_stmt(preparer.quote(self._db_name))
                 conn.execute(sql.text(stmt))
         finally:
             engine.dispose()
 
         # enable pgvector
-        store_db_url = self._db_server.get_uri(database=self._db_name, driver='psycopg')
+        store_db_url = self.db_url
```
mkornacker: remove store_db_url and simply use db_url?

pixeltable/metadata/__init__.py
```diff
@@ -24,9 +24,10 @@ def create_system_info(engine: sql.engine.Engine) -> None:
     system_md = SystemInfoMd(schema_version=VERSION)
     record = SystemInfo(md=dataclasses.asdict(system_md))
     with orm.Session(engine, future=True) as session:
-        session.add(record)
-        session.flush()
-        session.commit()
+        if session.query(SystemInfo).count() == 0:
```
mkornacker, line 28: Why is this needed?

pixeltable/utils/dbms.py
```diff
@@ -0,0 +1,48 @@
+import abc
+
+from sqlalchemy import URL
+
+
+class Dbms(abc.ABC):
```
mkornacker, line 6: Missing: class comment

pixeltable/utils/dbms.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, name: str, transaction_isolation_level: str, version_index_type: str, db_url: URL) -> None:
+        self.name = name
```
mkornacker, line 17: missing: class-level member decls

pixeltable/utils/dbms.py
```diff
@@ -0,0 +1,48 @@
...
+        self.transaction_isolation_level = transaction_isolation_level
+        self.version_index_type = version_index_type
+        self.db_url = db_url
+
+    @abc.abstractmethod
+    def build_drop_db_stmt(self, database: str) -> str: ...
```
mkornacker: "build" is redundant

mkornacker (changes requested):

pixeltable/metadata/__init__.py
amithadke: for idempotency in cloud version we dont have a way to figure out if db is initialized or not.
So we try to add SystemInfo and create all tables again in each pass 
    metadata.schema.base_metadata.create_all(self._sa_engine, checkfirst=True)
    metadata.create_system_info(self._sa_engine)

pixeltable/env.py
amithadke: we will not delete any tables or db in cloud case 
added a check for reinit_db can not be set when in cloud setup

pixeltable/env.py
amithadke: fixed dbms. default_system_db_url can work w/o need of dbserver just need to replace databsase in url that ie has with 'postgres'

pixeltable/env.py
```diff
@@ -354,39 +360,13 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
...
-            cleanup_mode = 'stop' if platform.system() == 'Windows' else None
-            self._db_server = pixeltable_pgserver.get_server(self._pgdata_dir, cleanup_mode=cleanup_mode)
-            self._db_url = self._db_server.get_uri(database=self._db_name, driver='psycopg')
-            self._dbms = PostgresqlDbms(sql.make_url(self._db_url))
+        # configure pixeltable database
+        self._init_db(config)
+
+        if reinit_db and not self.is_local:
+            raise excs.Error(
+                'Reinitializing pixeltable database is not supported when running in non local environment'
```
mkornacker: 'non-local'

pixeltable/env.py
```diff
@@ -416,25 +393,62 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
...
+            self._db_name = os.environ.get('PIXELTABLE_DB', 'pixeltable')
+            self._pgdata_dir = Path(os.environ.get('PIXELTABLE_PGDATA', str(Config.get().home / 'pgdata')))
+            # cleanup_mode=None will leave the postgres process running after Python exits
+            # cleanup_mode='stop' will terminate the postgres process when Python exits
+            # On Windows, we need cleanup_mode='stop' because child processes are killed automatically when the parent
+            # process (such as Terminal or VSCode) exits, potentially leaving it in an unusable state.
+            cleanup_mode = 'stop' if platform.system() == 'Windows' else None
+            self._db_server = pixeltable_pgserver.get_server(self._pgdata_dir, cleanup_mode=cleanup_mode)
+            self._db_url = self._db_server.get_uri(database=self._db_name, driver='psycopg')
+            self._dbms = PostgresqlDbms(sql.make_url(self._db_url))
```
mkornacker, line 440: Before you return, add a few asserts like self._dbms is not None. That'll also help the reader see what needs to get set here.

pixeltable/env.py
```diff
@@ -467,36 +480,27 @@ def _store_db_exists(self) -> bool:
     def _create_store_db(self) -> None:
         assert self._db_name is not None
         # create the db
-        pg_db_url = self.default_system_db_url
-        engine = sql.create_engine(pg_db_url, future=True, isolation_level='AUTOCOMMIT')
+        engine = sql.create_engine(self._dbms.default_system_db_url(), future=True, isolation_level='AUTOCOMMIT')
         preparer = engine.dialect.identifier_preparer
         try:
             with engine.begin() as conn:
                 # use C collation to get standard C/Python-style sorting
```
mkornacker: Isn't this comment now obsolete?

aaron-siegel (approved):

PR #531, opened 2025-03-24
Jw/fix syntax error

jacobweiss2305:
We have a syntax error with a few examples. I introduce schema_of_df as the second positional arguement.

I also replace the formatting udf with a pxt format functions

### DIFF
--- a/docs/mintlify/docs/cookbooks/agents/pixelagent.mdx
+++ b/docs/mintlify/docs/cookbooks/agents/pixelagent.mdx
@@ -1,7 +1,7 @@
 ---
-title: 'Pixelagent'
-description: 'Understanding and implementing AI agents with Pixeltable'
-mode: 'wide'
+title: "Pixelagent"
+description: "Understanding and implementing AI agents with Pixeltable"
+mode: "wide"
 ---
 
 ## What is Pixelagent?
@@ -30,7 +30,7 @@ An agent is not simply a zero-shot call to a large language model. Proper agents
     External functions and APIs that the model can use to interact with the world and gather information.
   </Card>
   <Card title="Memory" icon="database">
-    The agent's ability to maintain and track state across multiple interactions.
+    The agent"s ability to maintain and track state across multiple interactions.
   </Card>
   <Card title="Loop" icon="rotate">
     The capability to execute multiple steps and iterations to complete a task.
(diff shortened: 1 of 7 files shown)


docs/mintlify/docs/cookbooks/agents/pixelagent.mdx
```diff
@@ -122,11 +122,11 @@ def create_chat(name: str, openai_model: str, tools: Optional[pxt.tools] = None)
     # Table to store conversation history between AI and User.
     message_table = pxt.create_table(
         path_str=f'{name}.messages',
-        {'role': pxt.String, 'content': pxt.String}
+        schema_or_df={'role': pxt.String, 'content': pxt.String}
```
aaron-siegel: We should not be referring to the `schema_or_df` parameter by name; it is going to change soon.

docs/mintlify/docs/cookbooks/chat/tools.mdx
```diff
@@ -73,7 +74,7 @@ Pixeltable tool-calling apps work in two phases:
     # Create base table
     tool_agent = pxt.create_table(
         path_str="agents.tools", 
-        {"prompt": pxt.String}, 
+        schema_or_df={"prompt": pxt.String},
```
aaron-siegel: Do we use single-quote or double-quote convention for strings in the docs? It doesn't seem to be consistent. (Is that intentional?)

aaron-siegel (changes requested):

docs/mintlify/docs/cookbooks/agents/pixelagent.mdx
pierrebrunelle: I'd remove both path_str and schema_or_df so if conventions change it would always work?

docs/mintlify/docs/cookbooks/chat/tools.mdx
pierrebrunelle: let's go with single quote.

docs/mintlify/docs/cookbooks/agents/pixelagent.mdx
jacobweiss2305: Im assuming `schema_or_df` is a backward compatible change. If so we should still have it in the docs, because we need to show the top 3 args for our most use function. Its unnatural not to show args for our most common function.

docs/mintlify/docs/cookbooks/chat/tools.mdx
jacobweiss2305: Okay I updated to double quotes

pierrebrunelle (approved):

docs/mintlify/docs/cookbooks/agents/pixelagent.mdx
jacobweiss2305: okay i removed args from create_table

aaron-siegel (approved):

PR #534, opened 2025-03-26
Bugfix related to chained tool calling

aaron-siegel:
Fixes a bug in `JsonMapper` evaluation that caused relative expressions to fail to bind when they appeared in the anchor of a `JsonPath`.

Also introduces a _workaround_ (not a proper fix) for a bug that causes `JsonMapper` evaluation to fail when the target expression involves an async UDF. The workaround forces the async UDF to be evaluated synchronously, which will give the correct output, but introduces a performance cliff when large numbers of rows are involved.

### DIFF
--- a/pixeltable/exprs/expr.py
+++ b/pixeltable/exprs/expr.py
@@ -514,7 +514,7 @@ class Expr(abc.ABC):
 
     @classmethod
     def _from_dict(cls, d: dict, components: list[Expr]) -> Self:
-        raise AssertionError('not implemented')
+        raise AssertionError(f'not implemented: {cls.__name__}')
 
     def isin(self, value_set: Any) -> 'exprs.InPredicate':
         from .in_predicate import InPredicate
--- a/pixeltable/exprs/expr.py
+++ b/pixeltable/exprs/expr.py
@@ -90,14 +90,28 @@ class Expr(abc.ABC):
                 result = c_scope
         return result
 
-    def bind_rel_paths(self, mapper: Optional['exprs.JsonMapper'] = None) -> None:
+    def bind_rel_paths(self) -> None:
         """
         Binds relative JsonPaths to mapper.
         This needs to be done in a separate phase after __init__(), because RelativeJsonPath()(-1) cannot be resolved
         by the immediately containing JsonMapper during initialization.
         """
+        self._bind_rel_paths()
+        assert not self._has_relative_path, self._expr_tree()
+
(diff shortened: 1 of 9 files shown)


jacobweiss2305:
I tested the code in the ticket and its getting stuck (never completing):
image

pixeltable/exprs/expr.py
```diff
@@ -90,14 +90,28 @@ def scope(self) -> ExprScope:
...
         by the immediately containing JsonMapper during initialization.
         """
+        self._bind_rel_paths()
+        assert not self._has_relative_path, self._expr_tree()
+
+    def _bind_rel_paths(self, mapper: Optional['exprs.JsonMapper'] = None) -> None:
+        for c in self.components:
+            c._bind_rel_paths(mapper)
+
+    def _expr_tree(self) -> str:
```
mkornacker, line 106: Is this basically a debug string? Do we want to keep it even after the hack has been removed?

If not, leave comment. If so, call it something that indicates that it's a debug string?

pixeltable/utils/coroutine.py
```diff
@@ -0,0 +1,36 @@
+import asyncio
```
mkornacker, line 1: Leave todo to remove this once the workaround has been replaced.

pixeltable/func/callable_function.py
```diff
@@ -93,13 +94,13 @@ def exec(self, args: Sequence[Any], kwargs: dict[str, Any]) -> Any:
             batched_kwargs = {k: [v] for k, v in kwargs.items() if k not in constant_param_names}
             result: list[Any]
             if inspect.iscoroutinefunction(self.py_fn):
-                result = asyncio.run(self.py_fn(*batched_args, **constant_kwargs, **batched_kwargs))
+                result = run_coroutine_synchronously(self.py_fn(*batched_args, **constant_kwargs, **batched_kwargs))
```
mkornacker, line 98: Comment out the existing code instead of removing it and leave a todo to remove this.

mkornacker (approved):

pixeltable/exprs/expr.py
aaron-siegel: Yeah, it's very useful for debugging. I'll add a comment saying what it is

pixeltable/utils/coroutine.py
aaron-siegel: Done.

PR #545, opened 2025-03-31
PXT-319: Disallow updates to media columns.

jpeterson-pxt:


tests/test_index.py
```diff
@@ -353,6 +353,58 @@ def test_add_embedding_index_if_exists(
...
+        img_t.insert(new_rows)
+        print(img_t.head())
+
+        with pytest.raises(pxt.Error, match='computed column may not reference the errortype or errormsg'):
+            img_t.add_computed_column(emsg=img_t.img.errormsg)
+
+        with pytest.raises(pxt.Error, match='computed column may not reference the errortype or errormsg'):
+            img_t.add_computed_column(etype=img_t.img.errortype)
+
+        # Update the first row with a new image
```
aaron-siegel, line 386: It would also be good to validate that computed columns that depend on the image column are also updated.

pixeltable/catalog/table.py
```diff
@@ -625,6 +625,12 @@ def add_computed_column(
         if stored is not None:
             col_schema['stored'] = stored
 
+        # Raise an error if the column expression refers to a column error property
+        if isinstance(spec, exprs.Expr):
+            for e in spec.subexprs(expr_class=exprs.ColumnPropertyRef, traverse_matches=False):
+                if e.is_error_prop():
+                    raise excs.Error('A computed column may not reference the errortype or errormsg of another column')
```
aaron-siegel: Mention the specific expression that's causing the issue?
'A computed column ... of another column, but {col_name!r}.errormsg is referenced' or some such?

aaron-siegel (approved):

pixeltable/catalog/table.py
jpeterson-pxt: Fixed this one.

PR #547, opened 2025-04-01
Add table ids to pixeltable mds.

amithadke:


amithadke:
need to fix move table to change dir id in md

pixeltable/catalog/table_version.py
```diff
@@ -53,6 +53,7 @@ class TableVersion:
     """
 
     id: UUID
+    dir_id: UUID
```
amithadke: I dont think this is good place to cache it since it can change. May be we don't need dir id in md, it doesn't tell much about table itself

pixeltable/metadata/schema.py
```diff
@@ -154,6 +154,8 @@ class ViewMd:
 @dataclasses.dataclass
 class TableMd:
     name: str
+    tbl_id: str  # uuid.UUID
+    dir_id: str  # uuid.UUID
```
mkornacker: Let's leave the dir id out of this, because that just reflects the local namespace (ie, the same table is going to reside in a different dir in the cloud namespace).

We could make the same argument for the name, but that would probably cause more downstream changes.

pixeltable/catalog/table_version.py
mkornacker: Yeah, the dir id isn't relevant here. It only matters in the context of Table (and subclasses).

pixeltable/metadata/schema.py
```diff
@@ -154,6 +154,7 @@ class ViewMd:
 @dataclasses.dataclass
 class TableMd:
     name: str
+    tbl_id: str  # uuid.UUID
```
mkornacker, line 156: Let's move this above name, it seems more meaningful.

mkornacker (approved):

PR #549, opened 2025-04-01
Prefer file extension over puremagic format when it matches a known document type

aaron-siegel:


pixeltable/utils/documents.py
```diff
@@ -18,85 +19,78 @@ class DocumentHandle:
...
+        if doc_format == ts.DocumentType.DocumentFormat.MD:
+            return DocumentHandle(doc_format, md_ast=get_markdown_handle(path))
+        if doc_format == ts.DocumentType.DocumentFormat.PDF:
+            return DocumentHandle(doc_format, pdf_doc=get_pdf_handle(path))
+        if doc_format == ts.DocumentType.DocumentFormat.XML:
+            return DocumentHandle(doc_format, bs_doc=get_xml_handle(path))
+        if doc_format == ts.DocumentType.DocumentFormat.TXT:
+            return DocumentHandle(doc_format, txt_doc=get_txt(path))
+    except Exception as exc:
+        raise excs.Error(f'An error occurred processing a {doc_format} document: {path}') from exc
```
jpeterson-pxt, line 52: PDF, XML, and HTML all raise their own error exceptions, rather than relying on this try-catch. 
Is this to provide better error messages in some cases?

aaron-siegel: Well they're still relying on this try/catch, in a sense. They raise their own exceptions in situations where no exception is thrown outside of Pixeltable, but it's still an error from Pixeltable's perspective; they'll still be caught by this try/catch and wrapped.

There are other errors that PDF/XML/HTML might throw in their processing, beyond the explicit ones defined by Pixeltable.

pixeltable/utils/documents.py
```diff
@@ -18,85 +19,78 @@ class DocumentHandle:
     txt_doc: Optional[str] = None
 
 
-def get_document_handle(path: str) -> Optional[DocumentHandle]:
-    doc_format = puremagic.from_file(path)
+def get_document_handle(path: str) -> DocumentHandle:
+    _, extension = os.path.splitext(path)
+    handle = get_handle_by_extension(path, extension)
```
mkornacker, line 24: Should this call be made if ext == ''?

aaron-siegel: Yes I think it's fine, it will return `None` like with any unknown extension

mkornacker (approved):

PR #559, opened 2025-04-04
PXT-410 Handle keyboard interrupt gracefully, clean up in memory stru…

amithadke:
…ctures during add computed column

Context:
Add compute can be interrupted at different stages and leaving in memory columns and indices and temp table sa metadata.
On database side transaction was never committed so on restart everything worked.

But most of the time if you Ctrl+C during add compute, that table handle was not useable.

Also noticed that if interrupted in finally block rest of the code in finally wont run
We want to catch KeyboardInterrupt and run cleanup without putting except KeyboardInterrupt everywhere in code so introducing 
run_cleanup (should be called to run critical cleanup)
run_cleanup_on_exception (should be called from finally block for cleanup only when any exception is raised)


pixeltable/catalog/table_version.py
```diff
@@ -545,14 +554,25 @@ def _add_index(self, col: Column, idx_name: Optional[str], idx: index.IndexBase)
...
+        finally:
+            # Remove in memory index structure in the case of error
+            def remove_index() -> None:
+                del self.idxs_by_name[idx_name]
+                del self.idx_md[idx_id]
+                self.next_idx_id = idx_id
 
+            run_cleanup_on_exception(remove_index)
+
+    def _add_index(self, col: Column, idx_name: Optional[str], idx: index.IndexBase) -> UpdateStatus:
```
amithadke, line 566: refactored this into 3 parts 
create index columns
add columns
create index along with its md

pixeltable/catalog/table_version.py
```diff
@@ -593,9 +613,21 @@ def add_columns(
         self.version += 1
         preceding_schema_version = self.schema_version
         self.schema_version = self.version
-        status = self._add_columns(cols, print_stats=print_stats, on_error=on_error)
+        index_cols: dict[Column, tuple[index.IndexBase, Column, Column]] = {}
+        all_cols: list[Column] = []
         for col in cols:
-            _ = self._add_default_index(col)
+            all_cols.append(col)
```
amithadke, line 617: goal is to call add_columns with all the columns that we will create for default index, this way if there is any error/interruption _add_columns can cleanup correctly

pixeltable/catalog/table_version.py
```diff
@@ -645,29 +678,33 @@ def _add_columns(
...
             try:
                 plan.open()
                 try:
-                    num_excs = self.store_tbl.load_column(col, plan, value_expr_slot_idx, on_error)
+                    excs_per_col = self.store_tbl.load_column(col, plan, value_expr_slot_idx, on_error)
                 except sql.exc.DBAPIError as exc:
                     # Wrap the DBAPIError in an excs.Error to unify processing in the subsequent except block
                     raise excs.Error(f'SQL error during execution of computed column `{col.name}`:\n{exc}') from exc
-                if num_excs > 0:
+                if excs_per_col > 0:
```
amithadke, line 685: now we create computed column followed by index columns we will not return correct num_excs
 how should we represent it? should we track num_excs per column and return in update_status? or it should be total (index columns dont have any so its going to only represent for computed column)

pixeltable/utils/exception_handler.py
```diff
@@ -0,0 +1,47 @@
+import logging
+import sys
+from typing import Callable, TypeVar
+
+R = TypeVar('R')
+
+
+def _is_in_exception():
```
mkornacker: Missing return types everywhere.

pixeltable/utils/exception_handler.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    Returns true when running in exception context.
+    """
+    current_exception = sys.exc_info()[1]
+    return current_exception is not None
+
+
+def run_cleanup_on_exception(cleanup_func: Callable[..., R], *args, **kwargs):
```
mkornacker: What's the use of R?

tests/test_table.py
```diff
@@ -1523,6 +1531,22 @@ def test_computed_col_exceptions(self, reset_db: None, test_tbl: catalog.Table)
         msgs = t.select(msg=t.add1.errormsg).collect()['msg']
         assert sum('division by zero' in msg for msg in msgs if msg is not None) == 10
 
+    def test_computed_col_with_interrupts(self, reset_db: None) -> None:
+        schema = {'c1': pxt.Int}
+        t = pxt.create_table('test_interrupt', schema)
+        t.insert(({'c1': i} for i in range(0, 1000)))
+        try:
+            _ = t.add_computed_column(cc1=self.f2(self.function_with_error(t.c1, 245)), on_error='ignore')
```
mkornacker: Why also f2()?

tests/test_table.py
```diff
@@ -1523,6 +1531,22 @@ def test_computed_col_exceptions(self, reset_db: None, test_tbl: catalog.Table)
         msgs = t.select(msg=t.add1.errormsg).collect()['msg']
         assert sum('division by zero' in msg for msg in msgs if msg is not None) == 10
 
+    def test_computed_col_with_interrupts(self, reset_db: None) -> None:
+        schema = {'c1': pxt.Int}
+        t = pxt.create_table('test_interrupt', schema)
+        t.insert(({'c1': i} for i in range(0, 1000)))
+        try:
```
mkornacker: Can't you just use with pytest.raises() here?

tests/test_table.py
```diff
@@ -1523,6 +1531,22 @@ def test_computed_col_exceptions(self, reset_db: None, test_tbl: catalog.Table)
...
+            _ = t.add_computed_column(cc1=self.f2(self.function_with_error(t.c1, 245)), on_error='ignore')
+            assert False, 'No exception was raised'
+        except (KeyboardInterrupt, Exception) as exc:
```
mkornacker: Why both keyboardinterrupt and exception?

pixeltable/catalog/table_version.py
```diff
@@ -472,33 +473,30 @@ def add_index(self, col: Column, idx_name: Optional[str], idx: index.IndexBase)
         _logger.info(f'Added index {idx_name} on column {col.name} to table {self.name}')
         return status
 
-    def _add_default_index(self, col: Column) -> Optional[UpdateStatus]:
-        """Add a B-tree index on this column if it has a compatible type"""
+    def _can_create_index(self, col: Column) -> bool:
```
mkornacker: This is checking whether this column is compatible with a b-tree index.

Find a better name (can_create_index isn't very clear).

pixeltable/catalog/table_version.py
```diff
@@ -472,33 +473,30 @@ def add_index(self, col: Column, idx_name: Optional[str], idx: index.IndexBase)
...
+
+    def _add_default_index(self, col: Column) -> Optional[UpdateStatus]:
+        """Add a B-tree index on this column if it has a compatible type"""
+        if not self._can_create_index(col):
             return None
         status = self._add_index(col, idx_name=None, idx=index.BtreeIndex(col))
         return status
 
-    def _add_index(self, col: Column, idx_name: Optional[str], idx: index.IndexBase) -> UpdateStatus:
+    def _create_index_columns(self, idx: index.IndexBase) -> Tuple[Column, Column]:
```
mkornacker, line 487: comment: return value

pixeltable/catalog/table_version.py
```diff
@@ -593,9 +613,21 @@ def add_columns(
...
         preceding_schema_version = self.schema_version
         self.schema_version = self.version
-        status = self._add_columns(cols, print_stats=print_stats, on_error=on_error)
+        index_cols: dict[Column, tuple[index.IndexBase, Column, Column]] = {}
+        all_cols: list[Column] = []
         for col in cols:
-            _ = self._add_default_index(col)
+            all_cols.append(col)
+            if self._can_create_index(col):
+                idx: index.IndexBase = index.BtreeIndex(col)
```
mkornacker: Do you need the type decl here?

pixeltable/catalog/table_version.py
mkornacker: We should never see an exception when computing a b-tree index value.

pixeltable/catalog/table_version.py
```diff
@@ -645,29 +678,32 @@ def _add_columns(
...
-                    del self.cols_by_id[col.id]
-                # we need to re-initialize the sqlalchemy schema
-                self.store_tbl.create_sa_tbl()
+                    num_excs += excs_per_col
+            except Exception as exc:
                 raise exc
             finally:
+
+                def delete_columns():
+                    self.cols.pop()
```
mkornacker: Do you mean to do this inside the loop?

pixeltable/catalog/table_version.py
```diff
@@ -645,29 +678,32 @@ def _add_columns(
...
-                        del self.cols_by_name[col.name]
-                    del self.cols_by_id[col.id]
-                # we need to re-initialize the sqlalchemy schema
-                self.store_tbl.create_sa_tbl()
+                    num_excs += excs_per_col
+            except Exception as exc:
                 raise exc
             finally:
+
+                def delete_columns():
```
mkornacker: Missing comment: what is being deleted.

pixeltable/catalog/table_version.py
```diff
@@ -645,29 +678,32 @@ def _add_columns(
...
-                    # remove columns that we already added
-                    if col.id not in self.cols_by_id:
-                        continue
-                    if col.name is not None:
-                        del self.cols_by_name[col.name]
-                    del self.cols_by_id[col.id]
-                # we need to re-initialize the sqlalchemy schema
-                self.store_tbl.create_sa_tbl()
+                    num_excs += excs_per_col
+            except Exception as exc:
```
mkornacker: The control flow requires an explanation (eg, why catch Exception, etc.).

Probably best to put into exception_handler.py. It should give specific instructions on how it needs to be used.

mkornacker (approved):

tests/test_table.py
amithadke: forgot to remove was testing function within function, removing it

pixeltable/utils/exception_handler.py
amithadke: for return type for cleanup functions, fixed it, exception handler will return function's return value

pixeltable/catalog/table_version.py
amithadke: let me check if typecheck was complaining

pixeltable/catalog/table_version.py
amithadke: will rename function too

pixeltable/catalog/table_version.py
amithadke: this is tricky one you are right as cols.pop() is not idempotent but then we want to put all of the clean up together and hand it off to exception handler. If we split it in two parts and get interrupted in between, second part wont execute.

Cleanup logic needs to change and delete current column idempotently

pixeltable/catalog/table_version.py
amithadke: pixeltable/catalog/table_version.py:636: error: Incompatible types in assignment (expression has type "IndexBase", variable has type "BtreeIndex")  [assignment]

pixeltable/catalog/table_version.py
amithadke: I wonder if we even need to catch exception here, as when handler is called from finally it can detect if an exception is raised or not

PR #560, opened 2025-04-04
Linting for `catalog`, `exec`, `io`, and `tool`

aaron-siegel:


pixeltable/catalog/catalog.py
```diff
@@ -132,7 +132,7 @@ def get_dir_path(self, dir_id: UUID) -> Path:
             # _debug_print(for_update=False, msg=f'dir id={dir_id}')
             row = conn.execute(q).one()
             dir = schema.Dir(**row._mapping)
-            if dir.md['name'] == '':
```
mkornacker: I find this harder to read that the comparison to ''. Can we get rid of this rule?

pixeltable/catalog/catalog.py
```diff
@@ -551,20 +551,20 @@ def get_view_ids(self, tbl_id: UUID, for_update: bool = False) -> list[UUID]:
 
     def get_tbl_version(self, tbl_id: UUID, effective_version: Optional[int]) -> Optional[TableVersion]:
         if (tbl_id, effective_version) not in self._tbl_versions:
-            self._tbl_versions[(tbl_id, effective_version)] = self._load_tbl_version(tbl_id, effective_version)
```
mkornacker, line 567: Why is this needed? I'm also finding that harder to read.

pixeltable/catalog/catalog.py
```diff
@@ -794,7 +794,8 @@ def _handle_path_collision(
             dir_contents = self._get_dir_contents(obj._id)
             if len(dir_contents) > 0 and if_exists == IfExistsParam.REPLACE:
                 raise excs.Error(
-                    f'Directory {path!r} already exists and is not empty. Use `if_exists="replace_force"` to replace it.'
+                    f'Directory {path!r} already exists and is not empty. '
```
mkornacker: Why is this necessary?

pixeltable/catalog/globals.py
```diff
@@ -66,8 +68,8 @@ def validated(cls, name: str, error_prefix: str) -> MediaValidation:
         try:
             return cls[name.upper()]
         except KeyError:
-            val_strs = ', '.join(f'{s.lower()!r}' for s in cls.__members__.keys())
-            raise excs.Error(f'{error_prefix} must be one of: [{val_strs}]')
+            val_strs = ', '.join(f'{s.lower()!r}' for s in cls.__members__)
```
mkornacker, line 71: What's wrong with making keys() explicit?

pixeltable/catalog/table_version_handle.py
```diff
@@ -31,6 +31,9 @@ def __eq__(self, other: object) -> bool:
             return False
         return self.id == other.id and self.effective_version == other.effective_version
 
+    def __hash__(self) -> int:
```
mkornacker, line 34: Where is this being used?

pixeltable/exec/cache_prefetch_node.py
```diff
@@ -253,12 +250,12 @@ def __fetch_url(self, url: str) -> tuple[Optional[Path], Optional[Exception]]:
                         }
                         self.boto_client = get_client(**config)
                 self.boto_client.download_file(parsed.netloc, parsed.path.lstrip('/'), str(tmp_path))
-            elif parsed.scheme == 'http' or parsed.scheme == 'https':
+            elif parsed.scheme in {'http', 'https'}:
```
mkornacker: Is this actually required? My expectation is that this would be slightly less efficient. Is there any reason for this to be a set?

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -76,7 +76,7 @@ def _set_pool_info(self) -> None:
         assert hasattr(self.pool_info, 'get_request_resources')
         sig = inspect.signature(self.pool_info.get_request_resources)
         self.get_request_resources_param_names = [p.name for p in sig.parameters.values()]
-        self.est_usage = {r: 0 for r in self._resources}
+        self.est_usage = dict.fromkeys(self._resources, 0)
```
mkornacker, line 79: I'm finding this harder to read. I'd have to look up with that means, which wasn't the case before.

pixeltable/catalog/catalog.py
aaron-siegel: Of course we can deselect rules when warranted. But I'd argue that personal stylistic preference isn't a good reason to do so. Here's why:
- We're building this as a community open source project. What's harder to read for one person might be easier to read for someone else. The linting rules represent the style consensus of the Python community and I'd argue our default should be to defer to that consensus.
- Personal preferences change; idioms that I found awkward when I first started getting into Python now seem (with the benefit of experience) perfectly natural. It's another reason to defer to the collective preference of the Python community.
- Plus the usual benefit of code consistency. (Perhaps Marcel prefers one style, but Aaron prefers a different style; if each author follows his personal preference rather than established convention, we end up with a codebase that's globally harder to read because it's inconsistent.)

I've disabled a number of rules (which you can see in pyproject.toml) but they're almost all code *structure* rules - where they impose some structural principle that doesn't make sense for our project. This one is purely stylistic and I'd argue we should need a more compelling (= less subjective) reason to overrule the linter.

pixeltable/catalog/catalog.py
aaron-siegel: Python convention is to omit tuple parenthesis where they're implied.

See preceding comment. On that one I'm actually somewhat agnostic; in this case I feel considerably more strongly that the linter is correct.

pixeltable/catalog/catalog.py
aaron-siegel: Line length was too long.

pixeltable/catalog/globals.py
aaron-siegel: Here's what ruff has to say about it:
https://docs.astral.sh/ruff/rules/in-dict-keys/

pixeltable/catalog/table_version_handle.py
aaron-siegel: It's a standard programming practice to always implement `__hash__` any time `__eq__` is overridden. (Not just in Python, but in other languages such as Java too.)

This is to prevent the existence of objects that present as equal but have different hash codes, which can lead to _very_ strange bugs if they get used in collections. I suspect most linters in most languages will flag this.

I guess it's less of an issue in Python where non-hashable objects are a thing, but I also see no harm in following the linter and implementing it. (honestly I'm trained to automatically implement `__hash__` any time I implement `__eq__`, so I'm quite sympathetic to the linter here)

pixeltable/exec/cache_prefetch_node.py
aaron-siegel: I tried it as a tuple first and the linter corrected me.

I'm not an expert in Python performance optimizations, so I tend to trust the linter unless there's specific evidence it's misleading us.

Here's what ruff says.
https://docs.astral.sh/ruff/rules/repeated-equality-comparison/

pixeltable/exec/expr_eval/schedulers.py
aaron-siegel: See preceding comments, I just don't think "Marcel finds it harder to read" is a good justification for disabling a linting rule. The fact that you'd have to look it up speaks to your individual fluency with this Python pattern; it doesn't imply it's poor practice to use it.

Perhaps to someone else who's more familiar with this pattern, the conventional form is actually easier to read. (Personally, I rather prefer it.) I'd surmise there's a performance benefit too, since no generator is involved (though again I'm not an expert on Python performance, but I do default to trusting the linter).

pixeltable/exec/cache_prefetch_node.py
mkornacker: I read the ruff explanation and I think it's ridiculous to argue that constructing a set and then using hash() could possibly be faster than two string comparisons.

To me, this defies common sense. It would be preferable to be able to stick with common sense in the cases where it deviates from ruff.

pixeltable/catalog/catalog.py
mkornacker: if dir.md['name'] doesn't distinguish between '' and None. That's a problem in principle, even if the code works here. We should disable that rule. We need to be allowed to write code that expresses the intended logic clearly.

pixeltable/exec/cache_prefetch_node.py
aaron-siegel: A well designed programming language will be well optimized and not use hash functions to test for membership of small sets.

For example, scala uses direct comparison on sets of cardinality <= 4 and only uses hashes if the cardinality is > 4.

I'm not an expert on Python performance and I don't know what Python does and I'll reiterate that I think we should default to trusting the tools unless we know otherwise.

mkornacker (approved):

PR #568, opened 2025-04-09
PXT-461: Enable table.select(None).

jpeterson-pxt:


tests/test_dataframe.py
```diff
@@ -535,6 +535,21 @@ def test_update_delete_where(self, test_tbl: catalog.Table) -> None:
             snap.where(t.c2 < 10).delete()
         assert 'Cannot delete from view' in str(exc_info.value)
 
+    def check_query(self, df: pxt.DataFrame, v: Any) -> None:
```
aaron-siegel: Make this a private method, `def __check_query` for clarity? Maybe also give it a more descriptive name (check_query is super vague)?

tests/test_dataframe.py
```diff
@@ -535,6 +535,21 @@ def test_update_delete_where(self, test_tbl: catalog.Table) -> None:
...
+        r = df.limit(5).collect()
+        assert len(r) == 5
+        print(r)
+        print(r[0])
+        print(r[0, 0])
+        assert r[0, 0] == v
```
aaron-siegel: Might as well check that all 5 rows are equal to the constant value.

tests/test_dataframe.py
```diff
@@ -535,6 +535,21 @@ def test_update_delete_where(self, test_tbl: catalog.Table) -> None:
...
+
+    def test_select_none(self, all_datatypes_tbl: catalog.Table) -> None:
```
aaron-siegel: Call this `test_select_constant`?

aaron-siegel (approved):

tests/test_dataframe.py
jpeterson-pxt: Done

tests/test_dataframe.py
jpeterson-pxt: Done

tests/test_dataframe.py
jpeterson-pxt: Done

PR #569, opened 2025-04-10
Fix for add_embedding_index() when embedding function has arity > 1

aaron-siegel:
In particular, fixes the use of `openai.embeddings` as an embedding function.


pixeltable/func/expr_template_function.py
```diff
@@ -97,6 +96,9 @@ def exec(self, args: Sequence[Any], kwargs: dict[str, Any]) -> Any:
 
     @property
     def display_name(self) -> str:
+        if not self.self_name and isinstance(self.templates[0].expr, exprs.FunctionCall):
+            # In this common case, fall back on the display name of the underlying FunctionCall
```
mkornacker: Could you spell out what this case is? I'm having trouble recognizing it.

mkornacker (approved):

pixeltable/func/expr_template_function.py
aaron-siegel: Done.

pixeltable/exprs/function_call.py
```diff
@@ -276,6 +276,30 @@ def reset_agg(self) -> None:
         assert isinstance(self.fn, func.AggregateFunction)
         self.aggregator = self.fn.agg_class(**self.agg_init_args)
 
+    @property
```
mkornacker, line 279: We need to get rid of properties that can result in exceptions (here and everywhere else, but we should definitely start here). The reason is that it can crash the debugger, because properties are shown automatically in the inspection window.

pixeltable/exprs/function_call.py
```diff
@@ -276,6 +276,30 @@ def reset_agg(self) -> None:
...
+                bound_args[name] = self.components[idx]
+            elif isinstance(idx, Sequence):
+                bound_args[name] = Expr.from_object([self.components[i] for i in idx])
+            elif isinstance(idx, dict):
+                bound_args[name] = Expr.from_object({k: self.components[i] for k, i in idx.items()})
+            else:
+                raise AssertionError(f'{name}: {idx} (of type `{type(idx)}`)')
+        return bound_args
+
+    def substitute(self, spec: dict[Expr, Expr]) -> Expr:
```
mkornacker, line 296: Requires explanation why it needs to be special-cased.

pixeltable/func/expr_template_function.py
```diff
@@ -78,6 +78,18 @@ def instantiate(self, args: Sequence[Any], kwargs: dict[str, Any]) -> 'exprs.Exp
         result = result.substitute(arg_exprs)
         return result
 
+    def call_return_type(self, bound_args: dict[str, 'exprs.Expr']) -> ts.ColumnType:
```
mkornacker, line 81: Explain why this needs to be special-cased.

tests/test_function.py
```diff
@@ -415,6 +415,31 @@ def test_nested_partial_binding(self, reset_db) -> None:
...
 
+    @staticmethod
+    @pxt.udf
+    def crt_test_udf(a: int, b: int, c: int = 5) -> pxt.Array[pxt.Int]:
+        return np.ones((b, c)) * a
+
+    def test_conditional_return_type(self, reset_db: None) -> None:
+        f = self.crt_test_udf
+
+        @f.conditional_return_type
```
mkornacker, line 426: Why not move this into the same scope as crt_test_udf?

pixeltable/exprs/function_call.py
aaron-siegel: I don't know if I agree with that. It means we're making decisions about our codebase design based on some peculiarity of a particular debugger. That's *really* putting the cart before the horse. If you're counting assertions as exceptions, it's way too broad.

It's very unlikely that this one would throw an exception though, so I think it's safe. Generally, nullary functions with no side effects should always be properties, and I'd like to stick to that. I suppose a case could be made that bona fide exceptions (not assertions) could be counted as side effects.

pixeltable/exprs/function_call.py
aaron-siegel: Added.

pixeltable/func/expr_template_function.py
aaron-siegel: Done.

tests/test_function.py
aaron-siegel: Doesn't work due to some Python weirdness involving the `@staticmethod` declaration. Maybe there's a better way to handle this but I think this is fine for now.

pixeltable/exprs/function_call.py
mkornacker: What is the advantage of having it be a property and not a function?

PR #575, opened 2025-04-11
Enable linting of `tests`

aaron-siegel:
Figured I might as well just knock this one off.


tests/test_table.py
```diff
@@ -564,9 +563,9 @@ def test_drop_table_force_via_handle(self, test_tbl: pxt.Table) -> None:
         t = pxt.get_table('test_tbl')
         v1 = pxt.create_view('v1', t)
         v2 = pxt.create_view('v2', t)
-        v3 = pxt.create_view('v3', v1)
-        v4 = pxt.create_view('v4', v2)
-        v5 = pxt.create_view('v5', t)
+        _ = pxt.create_view('v3', v1)
```
mkornacker: Can we please disable this rule? Having separate symbols is important when running the test in a debugger.

mkornacker (changes requested):

tests/test_table.py
aaron-siegel: We don't need to disable the rule; ruff is configured to treat any variable prefixed with `_` as unused, so we can give them different names. I just changed these to `_v3`, `_v4`, `_v5` (in both tests that use this pattern) and we can use the same pattern anywhere else it's useful to have named unused variables for debugging purposes.

mkornacker (approved):

PR #581, opened 2025-04-16
Logic for reconstituting replica data from a published bundle; new `pxt.create_replica()` method

aaron-siegel:


pixeltable/exprs/rowid_ref.py
```diff
@@ -114,3 +117,45 @@ def _as_dict(self) -> dict:
...
         tbl_id, normalized_base_id, idx = UUID(d['tbl_id']), UUID(d['normalized_base_id']), d['idx']
         return cls(tbl=None, idx=idx, tbl_id=tbl_id, normalized_base_id=normalized_base_id)
+
+
+class VminRef(Expr):
+    """
+    This bears some similarities to RowidRef, but there are enough differences to justify making it a separate class.
+    In particular, VminRef needs to be able to refer to a base table's vmin, and the retargeting simplifications of
+    RowidRef that are used by the planner don't apply (since a base table's vmin is in general different from its
+    descendants' vmins).
```
jpeterson-pxt: It might be useful to explain the differences between constructions with and without til:
VminRef(tbl, don't.care)
VminRef(None, id)
If both are specified doe the method need to assert that tbl.id == tlb_Id?

pixeltable/functions/globals.py
```diff
@@ -45,26 +45,34 @@ def _(val: sql.ColumnElement) -> Optional[sql.ColumnElement]:
     return sql.sql.func.sum(val)
 
 
+@func.uda(
+    allows_window=True,
+    type_substitutions=tuple({T: Optional[t]} for t in ts.ALL_PIXELTABLE_TYPES),  # type: ignore[misc]
+)
+class any_value(func.Aggregator, typing.Generic[T]):
+    result: T
```
jpeterson-pxt: Should we have a comment explaining what this does and why we have it?

pixeltable/functions/globals.py
```diff
@@ -45,26 +45,34 @@ def _(val: sql.ColumnElement) -> Optional[sql.ColumnElement]:
...
-            ts.Image,
-            ts.Video,
-            ts.Audio,
-            ts.Document,
-        )
-    ),
+    type_substitutions=tuple({T: Optional[t]} for t in ts.ALL_PIXELTABLE_TYPES),  # type: ignore[misc]
 )
 class count(func.Aggregator, typing.Generic[T]):
     def __init__(self) -> None:
```
jpeterson-pxt, line 55: Documentation comments?

jpeterson-pxt (commented):
I don't see anything wrong, but I'm not an expert on table versioning.

mkornacker:
Let's discuss in person tomorrow, but I think the data extraction for this scenario is best handled at the store level (ie, a data dump) and not as a query. The code will be more concise, easier to follow and also have better performance.

aaron-siegel:
> Let's discuss in person tomorrow, but I think the data extraction for this scenario is best handled at the store level (ie, a data dump) and not as a query. The code will be more concise, easier to follow and also have better performance.

Happy to discuss it, but it shouldn't be a blocker for reviewing and shipping this PR. The extraction method was introduced in the preceding PR (which is already on main). This PR is mostly about reconstitution, with relatively minor changes to extraction. I'd advocate for focusing on getting the end-to-end workflow shipped, and then we can separately decide whether revisiting/optimizing the extraction step is the right thing to prioritize next.

pixeltable/store.py
```diff
@@ -432,6 +432,37 @@ def delete_rows(
...
+        """
+        When instantiating a replica, we can't rely on the usual insertion code path, which contains error handling
+        and other logic that doesn't apply.
+        """
+        conn = Env.get().conn
+
+        # First cache the column mappings (for efficiency)
+        # (col_name, store_col_name) pairs
+        col_map: list[tuple[str, str]] = []
+        for col_name, col in self.tbl_version.get().cols_by_name.items():
```
mkornacker: cols_by_name doesn't include index or other system cols - why is that correct?

pixeltable/store.py
```diff
@@ -432,6 +432,37 @@ def delete_rows(
...
+            if col.is_stored:
+                col_map.append((f'val_{col_name}', col.store_name()))
```
mkornacker: Why not simply use the actual store column names here (and in the caller), instead of creating new ones?

pixeltable/store.py
```diff
@@ -432,6 +432,37 @@ def delete_rows(
         status = conn.execute(stmt)
         return status.rowcount
 
+    def insert_replica_rows(self, rows: list[dict[str, Any]]) -> None:
```
mkornacker: It feels like this function makes some assumptions about what the caller is doing, rather than making sense on its own. The row format is different here, and it's not clear what you gain from that.

pixeltable/share/publish.py
```diff
@@ -21,6 +20,9 @@
 
 
 def publish_snapshot(dest_tbl_uri: str, src_tbl: pxt.Table) -> str:
```
mkornacker: If we agree on calling this replication and the artefacts replicas, let's switch to that naming convention uniformly. Keeping 'publish' and 'clone' around leads to more confusion.

Also, we already know that the next thing after the initial release will be replicating base tables. Baking 'snapshot' into the interface seems needlessly restrictive.

pixeltable/share/packager.py
```diff
@@ -104,86 +110,116 @@ def __export_table(self, t: catalog.Table) -> None:
...
+        # The parquet file naming scheme anticipates future support for partitioning.
+        parquet_dir = self.tables_dir / f'tbl_{tv.id.hex}'
+        parquet_dir.mkdir()
+        parquet_file = parquet_dir / f'tbl_{tv.id.hex}.00000.parquet'
+        _logger.info(f'Creating parquet table: {parquet_file}')
+
+        # Populate the Parquet table with data.
+        # The data is first loaded from the DataFrame into a sequence of pyarrow tables, batched in order to avoid
+        # excessive memory usage. The pyarrow tables are then amalgamated into the (single) Parquet table on disk.
+        # We use snappy compression for the Parquet tables; the entire bundle will be bzip2-compressed later, so
```
mkornacker, line 106: Did you check whether bzip2 makes a difference? Parquet already has various forms of compression baked in.

pixeltable/share/packager.py
```diff
@@ -104,86 +110,116 @@ def __export_table(self, t: catalog.Table) -> None:
         # to get the column types, since we'll be substituting `fileurl`s for media columns.
         actual_col_types: list[ts.ColumnType] = []
 
-        for col_name, col in t._tbl_version.get().cols_by_name.items():
+        tv = tvp.tbl_version.get()
```
mkornacker: This seems error-prone and a lot more complicated than necessary.

Why not simply have a
StoreBase.dump_rows(version, filter_view: StoreBase, filter_view_version: int) -> Iterator[dict]
that produces table rows in the same format as what's needed for sql.insert()

And a corresponding StoreBase.load_rows(rows: Iterator[dict]). That could then also do its own batching.

StoreBase already understands the physical structure of the table, and you're replicating the physical structure (ie, you shouldn't need any back & forth translation).

pixeltable/share/packager.py
aaron-siegel: The tarball contains content other than parquet files, so I'd presume we need to compress it at the end regardless? (For example, it might contain compressible pdf or html documents.)

pixeltable/store.py
```diff
@@ -427,6 +427,34 @@ def delete_rows(
         status = conn.execute(stmt)
         return status.rowcount
 
+    def dump_rows(self, version: int, filter_view: StoreBase, filter_view_version: int) -> Iterator[dict[str, Any]]:
```
mkornacker, line 431: Missing method comment.

Also, I would have expected the filter_view to be optional (what about base tables?).

pixeltable/store.py
```diff
@@ -427,6 +427,34 @@ def delete_rows(
...
+        for row in result:
+            yield dict(zip(result.keys(), row))
+
+    def load_rows(self, rows: list[dict[str, Any]]) -> None:
+        """
+        When instantiating a replica, we can't rely on the usual insertion code path, which contains error handling
+        and other logic that doesn't apply.
+        """
+        conn = Env.get().conn
+        conn.execute(sql.insert(self.sa_tbl), rows)
```
mkornacker: Should this do batching that is independent of the size of the input?

pixeltable/store.py
```diff
@@ -427,6 +427,34 @@ def delete_rows(
...
+            .where(sql.exists().where(filter_predicate))
+        )
+        conn = Env.get().conn
+        _logger.debug(stmt)
+        log_explain(_logger, stmt, conn)
+        result = conn.execute(stmt)
+        for row in result:
+            yield dict(zip(result.keys(), row))
+
+    def load_rows(self, rows: list[dict[str, Any]]) -> None:
```
mkornacker: For large replicas, wouldn't an iterator be safer here?

pixeltable/catalog/table.py
```diff
@@ -254,11 +254,15 @@ def _schema(self) -> dict[str, ts.ColumnType]:
         """Return the schema (column names and column types) of this table."""
         return {c.name: c.col_type for c in self._tbl_version_path.columns()}
 
+    @property
```
mkornacker, line 261: I am very opposed to making functions that do things like database transactions and interact with the concurrency control mechanisms into properties. I think this really obscures the code and is an invitation for subtle bugs.

pixeltable/catalog/schema_object.py
```diff
@@ -33,6 +33,7 @@ def _parent(self) -> Optional['catalog.Dir']:
                 return None
             return Catalog.get().get_dir(self._dir_id)
 
+    @property
```
mkornacker, line 36: See my comment for base_table()

pixeltable/share/packager.py
```diff
@@ -69,138 +70,121 @@ def __init__(self, table: catalog.Table, additional_md: Optional[dict[str, Any]]
...
+        if isinstance(col_type, sql.JSON):
+            return pa.string()  # JSON will be exported as strings
+        if isinstance(col_type, sql.LargeBinary):
+            return pa.binary()
+        raise AssertionError(f'Unrecognized SQL type: {col_type} (type {type(col_type)})')
 
     def __to_pa_tables(
-        self, df: DataFrame, actual_col_types: list[ts.ColumnType], arrow_schema: pa.Schema, batch_size: int = 1_000
+        self,
+        row_iter: Iterator[dict[str, Any]],
```
mkornacker, line 144: Can you comment on the content of the dict (ie, what are the keys)? That would help with following along.

mkornacker (changes requested):
First round of comments. I haven't looked at packager.py in detail yet.

pixeltable/store.py
aaron-siegel: Iterator doesn't feel right to me as an input (too restrictive as an API). Might as well be an Iterable though, I can make that change.

pixeltable/catalog/table.py
aaron-siegel: See preceding comment for _path()

pixeltable/catalog/schema_object.py
aaron-siegel: Ok. This was changed recently from a property to a method and the change broke all sorts of stuff (log messages like `updating table '<bound method _path of blah blah blah>'` and such). The safest thing to do was roll back the change until we figure out how to safely implement it.

Assuming we do want to implement it. All of the coding I've done for basically the past 12 years (in any language) has followed the guideline that nullary methods without side effects should be properties. To me that's more or less settled coding practice, so I'm surprised this is coming up as an issue at all! It's basically the first time I've encountered the contrary point of view.

I don't see why it should matter whether they're methods or properties; what's an example of a bug that could be introduced if it were a property, but would be avoided if it's a method?

Regardless, for purposes of this PR let's please just roll back the breaking change and factor out separately the discussion of what to do here.

pixeltable/store.py
aaron-siegel: It makes more sense for me for batching to be controlled by the client of this API

pixeltable/store.py
aaron-siegel: I can make it optional but it's not really necessary, base tables just have filter_view == self (and this works in the present implementation)

pixeltable/store.py
aaron-siegel: Whatever, I added batching, it can't hurt :)

pixeltable/share/packager.py
aaron-siegel: Done. Note I haven't updated all the code comments yet; I want to get feedback on whether you're ok with the overall approach before I invest more in it

pixeltable/share/packager.py
```diff
@@ -69,138 +70,122 @@ def __init__(self, table: catalog.Table, additional_md: Optional[dict[str, Any]]
...
-        parent_path = table._parent()._path()
-        if len(parent_path) == 0:
-            return 'pxt'
-        else:
-            return f'pxt.{parent_path}'
+        parquet_schema = self.__to_parquet_schema(tv.store_tbl.sa_tbl)
+        # The parquet file naming scheme anticipates future support for partitioning.
+        parquet_dir = self.tables_dir / f'tbl_{tv.id.hex}'
+        parquet_dir.mkdir()
+        parquet_file = parquet_dir / f'tbl_{tv.id.hex}.00000.parquet'
```
mkornacker, line 100: Is there a single file per table?

Might be good to break it up into smaller pieces and then do parallel upload. Leave todo?

pixeltable/share/packager.py
```diff
@@ -69,138 +70,122 @@ def __init__(self, table: catalog.Table, additional_md: Optional[dict[str, Any]]
...
-        self.iceberg_catalog.create_namespace_if_not_exists(namespace)
-        iceberg_schema = self.__to_iceberg_schema(df._schema)
-        iceberg_tbl = self.iceberg_catalog.create_table(f'{namespace}.{t._name}', schema=iceberg_schema)
-
-        # Populate the Iceberg table with data.
-        # The data is first loaded from the DataFrame into a sequence of pyarrow tables, batched in order to avoid
-        # excessive memory usage. The pyarrow tables are then amalgamated into the (single) Iceberg table on disk.
-        for pa_table in self.__to_pa_tables(df, actual_col_types, iceberg_schema):
-            iceberg_tbl.append(pa_table)
+        sql_types = {col.name: col.type for col in tv.store_tbl.sa_tbl.columns}
```
mkornacker, line 89: Assert somewhere that tv is in the the tableversionpath of self.table?

pixeltable/store.py
```diff
@@ -427,6 +428,35 @@ def delete_rows(
...
         return status.rowcount
 
+    def dump_rows(self, version: int, filter_view: StoreBase, filter_view_version: int) -> Iterator[dict[str, Any]]:
+        filter_predicate = sql.and_(
+            filter_view.v_min_col <= filter_view_version,
+            filter_view.v_max_col > filter_view_version,
+            *[c1 == c2 for c1, c2 in zip(self.rowid_columns(), filter_view.rowid_columns())],
+        )
+        stmt = (
+            sql.select('*')
```
mkornacker: Don't you want to retrieve only the columns that are visible in 'version'? (Saves on IO, plus you mentioned that the merge would be in both the horizontal and vertical axis.)

pixeltable/catalog/table_version_path.py
```diff
@@ -98,6 +98,13 @@ def find_tbl_version(self, id: UUID) -> Optional[TableVersionHandle]:
             return None
         return self.base.find_tbl_version(id)
 
+    @property
+    def ancestors(self) -> list[TableVersionPath]:
```
mkornacker: ancestor_paths? I intuitively thought these are individual TableVersionHandles.

pixeltable/share/packager.py
```diff
@@ -69,138 +70,122 @@ def __init__(self, table: catalog.Table, additional_md: Optional[dict[str, Any]]
...
             json.dump(self.md, fp)
-        self.iceberg_catalog = sqlite_catalog(self.tmp_dir / 'warehouse')
+        self.tables_dir = self.tmp_dir / 'tables'
+        self.tables_dir.mkdir()
         with Env.get().begin_xact():
-            ancestors = (self.table, *self.table._base_tables)
-            for t in ancestors:
-                _logger.info(f"Exporting table '{t._path}'.")
-                self.__export_table(t)
+            for tvp in self.table._tbl_version_path.ancestors:
```
mkornacker: Why not just call TVP.get_tbl_versions() here? It doesn't look like you're using the path.

pixeltable/share/packager.py
```diff
@@ -69,138 +70,122 @@ def __init__(self, table: catalog.Table, additional_md: Optional[dict[str, Any]]
...
+        # Populate the Parquet table with data.
+        # The data is first loaded from the DataFrame into a sequence of pyarrow tables, batched in order to avoid
+        # excessive memory usage. The pyarrow tables are then amalgamated into the (single) Parquet table on disk.
+        # We use snappy compression for the Parquet tables; the entire bundle will be bzip2-compressed later, so
+        # faster compression should provide good performance while still reducing temporary storage utilization.
+        parquet_writer = pq.ParquetWriter(parquet_file, parquet_schema, compression='SNAPPY')
+        filter_tv = self.table._tbl_version.get()
+        row_iter = tv.store_tbl.dump_rows(tv.version, filter_tv.store_tbl, filter_tv.version)
+        for pa_table in self.__to_pa_tables(row_iter, sql_types, media_cols, parquet_schema):
+            parquet_writer.write_table(pa_table)
```
mkornacker, line 112: How does this interact with Parquet compression? Does this buffer everything in memory before writing out a single parquet file? (It would need to do that for things like dictionary encoding, which parquet supports.)

pixeltable/share/packager.py
```diff
@@ -225,9 +210,137 @@ def __build_tarball(self) -> Path:
...
+            # Add the dir containing Parquet tables
+            tf.add(self.tables_dir, arcname='tables')
             # Add the media files
             for src_file, dest_name in self.media_files.items():
                 tf.add(src_file, arcname=f'media/{dest_name}')
         return bundle_path
+
+
+class TableRestorer:
+    tbl_path: str
```
mkornacker, line 216: comment missing (what kind of path?)

Also: class comment missing

pixeltable/share/packager.py
```diff
@@ -225,9 +210,137 @@ def __build_tarball(self) -> Path:
...
+    md: Optional[dict[str, Any]]
```
mkornacker, line 217: comment missing

pixeltable/share/packager.py
```diff
@@ -225,9 +210,137 @@ def __build_tarball(self) -> Path:
...
+        print(f'Extracting table data into: {self.tmp_dir}')
+        with tarfile.open(bundle_path, 'r:bz2') as tf:
+            tf.extractall(path=self.tmp_dir)
+
+        if self.md is None:
+            # No metadata supplied; read it from the archive
+            with open(self.tmp_dir / 'metadata.json', 'r', encoding='utf8') as fp:
+                self.md = json.load(fp)
+
+        pxt_md_version = self.md['pxt_md_version']
```
mkornacker, line 238: Hard to follow without understanding what md is.

pixeltable/share/packager.py
```diff
@@ -225,9 +210,137 @@ def __build_tarball(self) -> Path:
...
+
+        tbl_md = [schema.FullTableMd.from_dict(t) for t in self.md['md']['tables']]
+
+        # Create the replica table
+        replica_tbl = catalog.Catalog.get().create_replica(catalog.Path(self.tbl_path), tbl_md)
+        assert replica_tbl._tbl_version.get().is_snapshot
+
+        # Now we need to instantiate and load data for replica_tbl and its ancestors, except that we skip
+        # replica_tbl itself if it's a pure snapshot.
+        if replica_tbl._id != replica_tbl._tbl_version.id:
```
mkornacker, line 256: Better test that doesn't use internals:
isinstance(View) and replica_tbl.snapshot_only (requires adding a View.snapshot_only property)

pixeltable/share/packager.py
```diff
@@ -225,9 +210,137 @@ def __build_tarball(self) -> Path:
...
+        if pxt_md_version != metadata.VERSION:
+            raise excs.Error(
+                f'Pixeltable metadata version mismatch: {pxt_md_version} != {metadata.VERSION}.\n'
+                'Please upgrade Pixeltable to use this dataset: pip install -U pixeltable'
+            )
+
+        tbl_md = [schema.FullTableMd.from_dict(t) for t in self.md['md']['tables']]
+
+        # Create the replica table
+        replica_tbl = catalog.Catalog.get().create_replica(catalog.Path(self.tbl_path), tbl_md)
```
mkornacker, line 251: At this point, you create a visible schema object. The assumption is that it and its ancestors are functional, which requires that they have store tables backing them. Catalog.create_replica() needs to somehow invoke TableVersion.create_replica() and then also load the data.

pixeltable/share/packager.py
```diff
@@ -225,9 +210,137 @@ def __build_tarball(self) -> Path:
...
+
+    def __import_table(self, bundle_path: Path, tv: catalog.TableVersion, tbl_md: schema.FullTableMd) -> None:
+        """
+        Import the Parquet table into the Pixeltable catalog.
+        """
+        tbl_id = uuid.UUID(tbl_md.tbl_md.tbl_id)
+        parquet_dir = bundle_path / 'tables' / f'tbl_{tbl_id.hex}'
+        parquet_table = pq.read_table(str(parquet_dir))
+
+        for batch in parquet_table.to_batches():
```
mkornacker, line 280: How does the batch size get determined? (Can this overflow memory?)

pixeltable/share/packager.py
```diff
@@ -225,9 +210,137 @@ def __build_tarball(self) -> Path:
...
+            pydict = batch.to_pydict()
+            rows = self.__from_pa_pydict(tv, pydict)
+            tv.store_tbl.load_rows(rows)
+
+    def __from_pa_pydict(self, tv: catalog.TableVersion, pydict: dict[str, Any]) -> list[dict[str, Any]]:
+        # Data conversions from pyarrow to Pixeltable
+        sql_types: dict[str, sql.types.TypeEngine[Any]] = {}
```
mkornacker, line 287: comment: keys

pixeltable/share/publish.py
```diff
@@ -102,5 +92,61 @@ def _upload_bundle_to_s3(bundle: Path, parsed_location: urllib.parse.ParseResult
...
     )
     s3_client.upload_file(
-        Filename=str(bundle), Bucket=bucket, Key=str(remote_path), ExtraArgs=upload_args, Callback=progress_bar.update
+        Filename=str(bundle), Bucket=bucket, Key=remote_path, ExtraArgs=upload_args, Callback=progress_bar.update
+    )
+
+
+def pull_replica(dest_path: str, src_tbl_uri: str) -> pxt.Table:
+    headers_json = {'X-api-key': Env.get().pxt_api_key, 'Content-Type': 'application/json'}
+    clone_request_json = {'operation_type': 'clone_snapshot', 'table_uri': src_tbl_uri}
```
mkornacker, line 101: Leave todo to streamline terminology? We shouldn't have multiple ways to refer to the same thing.

pixeltable/share/publish.py
```diff
@@ -102,5 +92,61 @@ def _upload_bundle_to_s3(bundle: Path, parsed_location: urllib.parse.ParseResult
...
+    response = requests.post(PIXELTABLE_API_URL, json=clone_request_json, headers=headers_json)
+    if response.status_code != 200:
+        raise excs.Error(f'Error cloning snapshot: {response.text}')
+    response_json = response.json()
+    if not isinstance(response_json, dict) or 'table_uri' not in response_json:
+        raise excs.Error(f'Error cloning shapshot: unexpected response from server.\n{response_json}')
+
+    print(response_json)
```
mkornacker: remove

pixeltable/share/publish.py
```diff
@@ -102,5 +92,61 @@ def _upload_bundle_to_s3(bundle: Path, parsed_location: urllib.parse.ParseResult
...
+    primary_tbl_additional_md = response_json['md']['tables'][0]['table_md']['additional_md']
```
mkornacker, line 109: Does this have a declared structure somewhere? As it is, it's hard to follow.

pixeltable/share/publish.py
```diff
@@ -102,5 +92,61 @@ def _upload_bundle_to_s3(bundle: Path, parsed_location: urllib.parse.ParseResult
...
+    bundle_size = obj['ContentLength']
+
+    bundle_path = Path(Env.get().create_tmp_path())
+    progress_bar = tqdm(
+        desc='Downloading',
+        total=bundle_size,
+        unit='B',
+        unit_scale=True,
+        unit_divisor=1024,
+        miniters=1,  # Update every iteration (should be fine for an upload)
```
mkornacker: upload?

pixeltable/catalog/schema_object.py
mkornacker: I'm only seeing two ._path without the (), both in packager.py. You should simply add the ().

mkornacker (approved):

pixeltable/catalog/schema_object.py
aaron-siegel: It was more than two. I'm pretty sure I fixed it in two places, then kept finding more. At that point it made more sense to roll back rather than try to fix forward.

pixeltable/share/packager.py
aaron-siegel: There's a comment in there already about future partitioning? I'll make it a TODO. For parallel uploads, we'd have to chunk the tarball, right?

pixeltable/share/packager.py
aaron-siegel: Done.

pixeltable/store.py
aaron-siegel: Maybe? Though it seems like one advantage to using StoreTable is that we don't have to worry about making sure we have the right list of columns selected (if we introduce new types of system columns, for example). And any columns that are not present in the version being published will necessarily have values that are all null, so the storage overhead is bounded. Let's consider this as a possible future optimization.

pixeltable/catalog/table_version_path.py
aaron-siegel: Ok

pixeltable/share/packager.py
aaron-siegel: ok

pixeltable/share/packager.py
aaron-siegel: If I understand the docs right, it's partitioned into row groups of size `min(pyarrow_table_size, 2**20)`. I'm using 1000-row pyarrow tables; we can of course tweak this at any point

pixeltable/share/packager.py
aaron-siegel: done

pixeltable/share/packager.py
aaron-siegel: done

pixeltable/share/packager.py
aaron-siegel: Defaults to `131_072` per pyarrow docs.

pixeltable/share/publish.py
aaron-siegel: Only in the Notion doc.

We should probably make it into a Pydantic structure with appropriate validators that's shared between the local/cloud logic. I've had some discussions with Amit about this.

PR #582, opened 2025-04-18
Properly resolve `ColumnRef`s whose reference table differs from the column definition table

aaron-siegel:


pixeltable/catalog/column.py
```diff
@@ -170,9 +171,10 @@ def has_window_fn_call(self) -> bool:
         )
         return len(window_fn_calls) > 0
 
-    def get_idx_info(self) -> dict[str, 'TableVersion.IndexInfo']:
+    def get_idx_info(self, reference_tbl: Optional['TableVersionPath'] = None) -> dict[str, 'TableVersion.IndexInfo']:
```
mkornacker, line 176: This is out of place here. Column is a representation of the physical column, which is always attached to a particular table. The caller should have the context to decide where to look.

Leave a todo.

pixeltable/exprs/column_ref.py
```diff
@@ -37,6 +37,7 @@ class ColumnRef(Expr):
     """
 
     col: catalog.Column
+    reference_tbl: catalog.TableVersionPath
```
mkornacker: Add an explanation of why this is needed to the class comment.

pixeltable/catalog/column.py
aaron-siegel: Done.

pixeltable/exprs/column_ref.py
aaron-siegel: Done.

mkornacker (approved):

PR #589, opened 2025-04-26
Improvements and fixes for Table._descriptors()

mkornacker:
- output with less gratuitous vertical space
- now also prints view predicate
- correctly prints snapshot bases

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -2120,6 +2120,7 @@ class TestTable:
         v2._link_external_store(MockProject.create(v2, 'project', {}, {}))
         v2.describe()
 
+        # test case: view with additional columns
         r = repr(v2)
         assert strip_lines(r) == strip_lines(
             """View 'test_subview' (of 'test_view', 'test_tbl')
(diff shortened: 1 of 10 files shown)


tests/test_table.py
```diff
@@ -2149,6 +2147,23 @@ def test_repr(self, test_tbl: catalog.Table, all_mpnet_embed: func.Function) ->
         )
         _ = v2._repr_html_()  # TODO: Is there a good way to test this output?
 
+        s = pxt.create_snapshot('test_snap', test_tbl)
```
aaron-siegel: Also include an example of a snapshot over a view? (Or a mixed hierarchy of view -> snapshot -> view -> snapshot or some such?)

Also include an example of a non-pure snapshot since the code path for resolving ancestors is (slightly) different?

aaron-siegel (approved):

tests/test_table.py
mkornacker: Done.

PR #591, opened 2025-04-28
WIP: stratified sampling operation on DataFrame

jpeterson-pxt:
WIP
Stratified sampling is functional when used like this:

df = t.select(t.a, t.b).where(t.a < 100).sample(n=100, seed=1234, stratify_by=[t.cat1, t.cat2])
df.collect()

Not yet implemented: 
- Binding sampling operation to a view.


pixeltable/dataframe.py
```diff
@@ -236,6 +241,28 @@ def _vars(self) -> dict[str, exprs.Variable]:
                 raise excs.Error(f'Multiple definitions of parameter {var.name}')
         return unique_vars
 
+    @classmethod
+    def _convert_int_param(cls, v: Optional[Any], required: bool, name: str) -> Optional[exprs.Expr]:
```
aaron-siegel: `Optional[Any]` might as well just be `Any`

pixeltable/dataframe.py
```diff
@@ -902,18 +990,98 @@ def limit(self, n: int) -> DataFrame:
...
+            raise excs.Error('sample() cannot be used with group_by')
+        if self.order_by_clause is not None:
+            raise excs.Error('sample() cannot be used with order_by')
+        if self.limit_val is not None:
+            raise excs.Error('sample() cannot be used with limit')
+        if self._has_joins():
+            raise excs.Error('sample() cannot be used with join')
+
+        if (n is None) and (fraction is None):
+            raise excs.Error('At least one of `n` or `fraction` must be supplied.')
```
aaron-siegel: Add a note to the docstring explaining the behavior if both are specified?

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +502,121 @@ def _create_stmt(self) -> sql.Select:
                 full=join_clause == plan.JoinType.FULL_OUTER,
             )
         return stmt
+
+
+class SqlSampleNode(SqlNode):
+    """
+    Returns rows from a stratified sample with N samples per strata.
```
aaron-siegel, line 533: I thought we decided on "n total samples, divided as evenly as possible among the strata", not "n samples per strata"?

pixeltable/metadata/converters/convert_34.py
```diff
@@ -0,0 +1,19 @@
...
+from pixeltable.metadata.converters.util import convert_table_md
+
+
+@register_converter(version=34)
+def _(engine: sql.engine.Engine) -> None:
+    convert_table_md(engine, substitution_fn=__substitute_md)
+
+
+def __substitute_md(k: Optional[str], v: Any) -> Optional[tuple[Optional[str], Any]]:
+    if isinstance(v, dict) and '_classname' in v and v['_classname'] == 'DataFrame':
```
aaron-siegel: `'_classname' in v and v['_classname'] == 'DataFrame'` can be shortened:

`v.get('_classname') == 'DataFrame'`

pixeltable/utils/sample.py
```diff
@@ -0,0 +1,143 @@
...
+import sqlalchemy as sql
+
+from pixeltable import exprs, type_system as ts
+from pixeltable.exprs.expr import DataRow, Expr
+from pixeltable.exprs.literal import Literal
+from pixeltable.exprs.row_builder import RowBuilder
+from pixeltable.exprs.sql_element_cache import SqlElementCache
+
+
+class SampleClause(Expr):
```
aaron-siegel: I'd expect this to go in `pixeltable.exprs`

pixeltable/utils/sample.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    An `Expr` that computes a sample of the associated table.
+    """
+
+    CURRENT_VERSION = 1
+
+    def __init__(
+        self,
+        version_expr: Optional[Expr],
+        n_expr: Optional[Expr],
```
aaron-siegel: It's not clear to me why these are expressions. Do we (or do we intend to) support variable values for `n`, `fraction`, etc.?

aaron-siegel (commented):
Some initial thoughts; will take another look tomorrow

pixeltable/exec/sql_node.py
jpeterson-pxt: This was a result of a suggestion from Marcel during our conversation with Obvio. 
It guarantees exactly identical sample counts for each strata, rather than randomly shorting some of the strata by one sample.
Obvio observed that everyone using the method will understand their dataset well enough to use it this way; they won't necessarily need to have an exact number of samples.

pixeltable/utils/sample.py
jpeterson-pxt: Yes, these could go either way. 
I did it this way so it would simply be unambiguous, and extensible in the future if needed.

pixeltable/utils/sample.py
jpeterson-pxt: Yes, I was unclear exactly where to put this, it can be moved.

pixeltable/metadata/converters/convert_34.py
jpeterson-pxt: Thanks, an improvement. I modeled this after other conversions.

pixeltable/dataframe.py
jpeterson-pxt: I've gone back and forth on this one. Perhaps both is not a viable option.

pixeltable/dataframe.py
mkornacker: Can we merge these two? Isn't this just creating a Literal? Also, what does 'param' mean here? It sounds like its making some assumption about the context in which it's getting called (but I don't see that reflected in the logic).

pixeltable/dataframe.py
```diff
@@ -272,15 +299,53 @@ async def _aexec(self) -> AsyncIterator[exprs.DataRow]:
...
+            item.bind_rel_paths()
+
+        #        s_key = SampleKey(samplex._seed_expr, self.__rowid_columns()) # TODO: use this for stratified sampling
+
+        return plan.Planner.create_query_plan(
+            self._from_clause, self._select_list_exprs, where_clause=self.where_clause, sample_clause=self.sample_clause
+        )
+
+    def _create_sample_plan(self) -> exec.ExecNode:
+        samplex = self.sample_clause
```
mkornacker: What is samplex? Please use meaningful variable names.

pixeltable/dataframe.py
```diff
@@ -902,18 +990,98 @@ def limit(self, n: int) -> DataFrame:
...
+        n: Optional[int] = None,
+        fraction: Optional[float] = None,
+        seed: Optional[int] = None,
+        stratify_by: Optional[list[Any]] = None,
+    ) -> DataFrame:
+        """Choose a shuffled sample of rows in the DataFrame
+
+        Args:
+            n: Number of rows to produce as a sample.
+            fraction: Fraction of available rows to produce as a sample.
```
mkornacker: Also: mention that n and fraction are mutually exclusive.

pixeltable/dataframe.py
```diff
@@ -902,18 +990,98 @@ def limit(self, n: int) -> DataFrame:
...
+            seed: Random seed for reproducible shuffling
+
+        Returns:
+            A new DataFrame with the sampled rows
```
mkornacker: This is confusing: this makes it sound like this operation returns rows (and that the df contains rows).

pixeltable/dataframe.py
```diff
@@ -902,18 +990,98 @@ def limit(self, n: int) -> DataFrame:
...
+        """
+
+        if self.sample_clause is not None:
+            raise excs.Error('sample() cannot be used with sample')
+        if self.group_by_clause is not None:
+            raise excs.Error('sample() cannot be used with group_by')
```
mkornacker: "... with group_by()", etc.

See examples elsewhere in the same file.

pixeltable/dataframe.py
```diff
@@ -902,18 +990,98 @@ def limit(self, n: int) -> DataFrame:
...
+
+        # analyze stratify list
+        stratify_list: list[exprs.Expr] = []
+        if stratify_by is not None:
+            if isinstance(stratify_by, exprs.Expr):
+                stratify_by = [stratify_by]
+            if not isinstance(stratify_by, list):
+                raise excs.Error('`stratify_by` parameter must be composed of expressions')
+            for raw_expr in stratify_by:
+                expr = exprs.Expr.from_object(raw_expr)
```
mkornacker: Why call from_object() if you're looking for a ColumnRef?

The method comment for from_object() says "Try to turn a literal object into an Expr".

pixeltable/dataframe.py
```diff
@@ -902,18 +990,98 @@ def limit(self, n: int) -> DataFrame:
...
+                if raw_expr is None or expr is None or not isinstance(raw_expr, exprs.ColumnRef):
+                    raise excs.Error(f'Invalid expression: {raw_expr}')
+                if not (expr.col_type.is_int_type() or expr.col_type.is_string_type() or expr.col_type.is_bool_type()):
```
mkornacker: Why not allow other scalars? I can see the utility of sampling from every hour of timestamped data. It wouldn't add any implementation complexity.

pixeltable/utils/sample.py
mkornacker: This is not an expression. It's a table operator, like grouping aggregation. (Which is also why it doesn't have an eval().)

This is collecting syntactic information. It should be in plan.py. See JoinClause and FromClause.

pixeltable/utils/sample.py
mkornacker: It doesn't make sense to have these be things like column refs. They need to be constants throughout the sampling operation.

pixeltable/utils/sample.py
```diff
@@ -0,0 +1,143 @@
...
+        max_md5_value = (2**32) - 1
+
+        # Calculate the fraction of this value
+        threshold_int = max_md5_value * int(1_000_000_000 * fraction) // 1_000_000_000
+
+        # Convert to hexadecimal string with padding
+        return format(threshold_int, '08x') + 'ffffffffffffffffffffffff'
+
+
+class SampleKey(Expr):
```
mkornacker: Why can't this just be a specific udf?

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +502,121 @@ def _create_stmt(self) -> sql.Select:
...
+            return self._create_stmt_n(self.n_samples)
+        else:
+            return self._create_stmt_fraction(self.fraction_samples)
+
+    def _create_stmt_n(self, n_samples: int) -> sql.Select:
+        """Create a Select stmt that returns n_samples per strata"""
+        sql_group_by_items = [self.sql_elements.get(e) for e in self.group_by_items]
+
+        # Get all columns from the input CTE dynamically
+        srct_columns = [self.input_cte.c[i] for i in range(len(self.input_cte.c))]
```
mkornacker: Please use meaningful variable names. (srct?)

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +502,121 @@ def _create_stmt(self) -> sql.Select:
...
+        # Construct an expression for randomly ordering rows with a given seed
+        s_key = SampleKey(self.seed, self.__rowid_columns())
+        o_by = s_key.sql_expr(self.sql_elements)
+
+        # Create a list of all columns plus the rank
+        select_columns = srct_columns.copy()
+        select_columns.append(
+            sql.func.row_number().over(partition_by=sql_group_by_items, order_by=o_by).label('rank')
+        )
+        srcp = sql.select(*select_columns).select_from(self.input_cte).cte('srcp')
```
mkornacker: srcp?

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +502,121 @@ def _create_stmt(self) -> sql.Select:
...
+
+        final_columns = [srcp.c[i] for i in range(len(srcp.c))]
```
mkornacker: Isn't that just *srcp.c?

mkornacker (changes requested):

pixeltable/utils/sample.py
jpeterson-pxt: I started down this path. Is there a way to reference a list of all row-ids in the public UI, or is it ok to use the internals in a UDF?

pixeltable/dataframe.py
jpeterson-pxt: Yes, these are meant to be called from a particular context. 
They are for checking parameters of user-callable methods. 
It is an attempt to give meaningful errors messages for the error conditions. 
Note that the method is now used in the logic for the "limit()" clause as well.  
Merging them would be a good idea, with an additional parameter for the desired type.
Done

pixeltable/dataframe.py
jpeterson-pxt: Done.

jpeterson-pxt: Done.

pixeltable/dataframe.py
jpeterson-pxt: This is a good question. Sampling from every hour of timestamped data would be sampling from a computation on the timestamp. This should be discussed - I put it in a slack channel.

pixeltable/dataframe.py
jpeterson-pxt: Good catch. Removed.

pixeltable/dataframe.py
jpeterson-pxt: Disallowing both. Done.

pixeltable/exec/sql_node.py

pixeltable/exec/sql_node.py

pixeltable/exec/sql_node.py
jpeterson-pxt: Both supported. 
Done

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/dataframe.py
jpeterson-pxt: Done. In combination with allowing arbitrary computations in the category expressions, it means only that the result of the computations is scalar.

pixeltable/utils/sample.py
jpeterson-pxt: Done.

pixeltable/utils/sample.py
jpeterson-pxt: Done

pixeltable/utils/sample.py
mkornacker: Why do you need that? Your SqlSampleNode already constructs a sequence of Exprs that are rowid refs, you can simply use them to create a FunctionCall instance like so:
fn_call_expr = sample_key_udf(*rowid_refs)

pixeltable/utils/sample.py
```diff
@@ -1,88 +1,98 @@
...
 from pixeltable.exprs.row_builder import RowBuilder
 from pixeltable.exprs.sql_element_cache import SqlElementCache
 
 
-class SampleClause(Expr):
-    """
-    An `Expr` that computes a sample of the associated table.
-    """
+@dataclasses.dataclass
+class SampleClause:
```
mkornacker: This represents a syntactic element, similar to JoinClause. It should also live in plan.py.

pixeltable/utils/sample.py
```diff
@@ -1,88 +1,98 @@
...
-    def _equals(self, other: SampleClause) -> bool:
-        return True
+    @property
+    def _version(self) -> int:
+        return self.version
 
     @property
-    def _version_expr(self) -> Optional[Expr]:
-        return self.components[0]
+    def _n(self) -> Optional[int]:
```
mkornacker: These are redundant and only add boilerplate, it's simpler to access .n directly.

pixeltable/utils/sample.py
```diff
@@ -1,88 +1,98 @@
...
+class SampleClause:
+    """Defines a sampling clause for a table."""
 
-    CURRENT_VERSION = 1
+    version: Optional[int]
+    n: Optional[int]
+    n_per_stratum: Optional[int]
+    fract: Optional[float]
+    seed: Optional[int]
+    stratify_list: Optional[list[Expr]]
```
mkornacker: Elsewhere in the codebase, we generally name list[Expr] as '<something>_exprs'. It would be good to follow that convention.

pixeltable/exec/sql_node.py
```diff
@@ -512,111 +528,104 @@ class SqlSampleNode(SqlNode):
...
 
-        return stmt
+        final_columns = [*row_rank_cte.c[:-1]].copy()  # exclude the rank column
+        if n_per_stratum is not None:
+            return sql.select(*final_columns).filter(row_rank_cte.c.rank <= n_per_stratum)
+        else:
+            return sql.select(*final_columns).order_by(row_rank_cte.c.rank).limit(n)
 
     def _create_stmt_fraction(self, fraction_samples: float) -> sql.Select:
         """Create a Select stmt that returns a fraction of the rows per strata"""
```
mkornacker, line 612: Isn't that simply the same as taking a fraction of the rows? I'm not sure where grouping comes in here.

pixeltable/plan.py
```diff
@@ -692,6 +728,53 @@ def _insert_prefetch_node(
...
+        order_by_clause: Optional[list[tuple[exprs.Expr, bool]]],
+        limit: Optional[exprs.Expr],
+    ) -> tuple[
+        exprs.Expr,
+        Optional[list[exprs.Expr]],
+        Optional[list[tuple[exprs.Expr, bool]]],
+        Optional[exprs.Expr],
+        Optional[SampleClause],
+    ]:
+        """Implement the sample clause by creating required subclauses."""
```
mkornacker: I'm confused. I thought the SampleSqlNode is doing all of that. Is this in effect a query rewrite?

pixeltable/plan.py
```diff
@@ -692,6 +728,53 @@ def _insert_prefetch_node(
...
+
+        # If no sample clause, return the original clauses
+        if sample_clause is None:
+            return where_clause, group_by_clause, order_by_clause, limit, None
+
+        # If the sample clause is stratified, we need to create a group by clause
```
mkornacker: This is what I expected to happen in SampleSqlNode.

pixeltable/utils/sample.py
```diff
@@ -95,6 +105,21 @@ def fraction_to_md5_hex(cls, fraction: float) -> str:
         # Convert to hexadecimal string with padding
         return format(threshold_int, '08x') + 'ffffffffffffffffffffffff'
 
+    @classmethod
+    def key_sql_expr(
```
mkornacker: This logic should be in SampleSqlNode.

pixeltable/utils/sample.py
mkornacker: I still don't follow why this exists. Is there a reason why this can't be done with a udf?

mkornacker (changes requested):

pixeltable/utils/sample.py
jpeterson-pxt: Thanks, good to know

pixeltable/utils/sample.py
jpeterson-pxt: Yes, not necessary.

pixeltable/exec/sql_node.py
jpeterson-pxt: Logically, yes. But the point of stratification is to ensure that the same fraction of samples comes from each strata. So we:
- Count the number of rows in each stratum, and compute the number of those rows which below in the sample
- Order the rows within each stratum by the key and add a rank to each row
- Select the rows from each stratum where the rank <= number of rows from the stratum which belong in the sample. 
It is an attempt to ensure that the sample fractions are similar across strata if the overall ordering would favor some strata.

pixeltable/plan.py
jpeterson-pxt: There are currently 4 queries which can be generated for sampling:

1. Non-stratified sampling can be done with a simple query rewrite. This is done in the planner, and does not require the SampleSqlNode at all. 
2 and 3. Two forms of stratified "n" sampling, one overall n and one n_per_stratum. These are almost the same.
4. Stratified fractional sampling (discussed in réponse to one of your other questions above).

pixeltable/plan.py
jpeterson-pxt: See earlier reply about the 4 queries created. 
This logic is abstracted here so that it can be used in create_query_plan() for the simple rewrite, and in create_view_load_plan() for a similar rewrite.

pixeltable/utils/sample.py
jpeterson-pxt: ok

pixeltable/catalog/table.py
```diff
@@ -227,6 +227,22 @@ def distinct(self) -> 'pxt.DataFrame':
     def limit(self, n: int) -> 'pxt.DataFrame':
         return self._df().limit(n)
 
+    def sample(
```
aaron-siegel, line 243: Add to `ColumnRef` too?

pixeltable/catalog/view.py
```diff
@@ -84,12 +86,21 @@ def _create(
...
                 raise excs.Error(f'Filter cannot be computed in the context of the base {base.tbl_name()}')
             # create a copy that we can modify and store
             predicate = predicate.copy()
+        if sample_clause is not None:
+            # make sure that the sample clause can be computed in the context of the base
+            if sample_clause._stratify_list is not None and not all(
+                stratify_expr.is_bound_by([base]) for stratify_expr in sample_clause._stratify_list
+            ):
+                raise excs.Error(f'Sample clause cannot be computed in the context of the base {base.tbl_name()}')
+            # create a copy that we can modify and store
```
aaron-siegel, line 106: I don't love this pattern. IMO code of this sort is much easier to reason about if the abstractions are immutable. I'd find it much clearer if `SampleClause` were immutable (either explicitly or treated as such) and copies be made at the time we need to change it.

pixeltable/dataframe.py
```diff
@@ -210,8 +215,9 @@ def _normalize_select_list(
 
     @property
     def _first_tbl(self) -> catalog.TableVersionPath:
-        assert len(self._from_clause.tbls) == 1
-        return self._from_clause.tbls[0]
+        return self._from_clause._first_tbl
+        # assert len(self._from_clause.tbls) == 1
```
aaron-siegel: remove commented code?

pixeltable/dataframe.py
```diff
@@ -236,6 +242,36 @@ def _vars(self) -> dict[str, exprs.Variable]:
...
+            if not isinstance(v_expr, exprs.Literal):
+                raise excs.Error(f'{name!r} parameter must be a constant, not {v_expr}')
+            if range[0] is not None and not (v_expr.val >= range[0]):
+                raise excs.Error(f'{name!r} parameter must be >= {range[0]}')
+            if range[1] is not None and not (v_expr.val <= range[1]):
+                raise excs.Error(f'{name!r} parameter must be <= {range[1]}')
+        return v_expr
+
+    @classmethod
+    def validate_constant_type_range(
```
aaron-siegel, line 263: This doesn't just validate, it does some normalization, right?

pixeltable/dataframe.py
```diff
@@ -292,14 +328,21 @@ def _create_query_plan(self) -> exec.ExecNode:
...
+    def __rowid_columns(self, num_rowid_cols: Optional[int] = None) -> list[exprs.Expr]:
+        """Return list of RowidRef for the given number of associated rowids"""
+        return Planner.rowid_columns(self._first_tbl.tbl_version, num_rowid_cols)
+
     def _has_joins(self) -> bool:
         return len(self._from_clause.join_clauses) > 0
 
     def show(self, n: int = 20) -> DataFrameResultSet:
+        if self.sample_clause is not None:
+            raise excs.Error('show() cannot be used with sample()')
```
aaron-siegel, line 342: Why not?

pixeltable/dataframe.py
```diff
@@ -945,18 +1004,110 @@ def limit(self, n: int) -> DataFrame:
...
+
+    def sample(
+        self,
+        n: Optional[int] = None,
+        n_per_stratum: Optional[int] = None,
+        fraction: Optional[float] = None,
+        seed: Optional[int] = None,
+        stratify_by: Any = None,
+    ) -> DataFrame:
+        """Choose a shuffled sample of rows from the DataFrame
```
aaron-siegel: Mention somewhere that exactly one of n, n_per_stratum, or fraction must be specified. I realize one can derive that from the notes in the `Args` block but it'd be clearer to state it explicitly up front.

The docstring could use some more elaboration explaining the behavior of `sample` in more depth/clarity.

pixeltable/dataframe.py
```diff
@@ -945,18 +1004,110 @@ def limit(self, n: int) -> DataFrame:
...
+            n_per_stratum: Number of rows to produce per stratum as a sample. This parameter is only valid if
+                `stratify_by` is specified. Only one of `n` or `n_per_stratum` can be specified.
+            fraction: Fraction of available rows to produce as a sample. This parameter is not usable with `n` or
+                `n_per_stratum`. The fraction must be between 0.0 and 1.0.
+            seed: Random seed for reproducible shuffling
+            stratify_by: If specified, the sample will be stratified by these values.
+
+        Returns:
+            A new DataFrame which specifies the sampled rows
+        """
```
aaron-siegel, line 1074: Add usage examples

pixeltable/dataframe.py
```diff
@@ -945,18 +1004,110 @@ def limit(self, n: int) -> DataFrame:
...
+            raise excs.Error('sample() cannot be used with group_by()')
+        if self.order_by_clause is not None:
+            raise excs.Error('sample() cannot be used with order_by()')
+        if self.limit_val is not None:
+            raise excs.Error('sample() cannot be used with limit()')
+        if self._has_joins():
+            raise excs.Error('sample() cannot be used with join()')
+
+        # Check paramter combinations
+        if n is None and n_per_stratum is None and fraction is None:
```
aaron-siegel: Three of these checks could be combined into one with
```python
if (n is not None) + (n_per_stratum is not None) + (fraction is not None) != 1:
    raise excs.Error('Exactly one of `n`, `n_per_stratum`, or `fraction` must be specified.')
```
I also find that error message to be clearer and more precise.

pixeltable/dataframe.py
```diff
@@ -945,18 +1004,110 @@ def limit(self, n: int) -> DataFrame:
...
+        if (n is not None or n_per_stratum is not None) and fraction is not None:
+            raise excs.Error('Cannot specify both `n` or `n_per_stratum` with `fraction`')
+
+        # Check parameter types and values
+        n = self.validate_constant_type_range(n, ts.IntType(nullable=False), False, 'n', (1, None))
+        n_per_stratum = self.validate_constant_type_range(
+            n_per_stratum, ts.IntType(nullable=False), False, 'n_per_stratum', (1, None)
+        )
+        fraction = self.validate_constant_type_range(
+            fraction, ts.FloatType(nullable=False), False, 'fraction', (0.0000001, 0.9999999)
```
aaron-siegel: I don't see a reason not to just make this `(0.0, 1.0)`.

pixeltable/dataframe.py
```diff
@@ -945,18 +1004,110 @@ def limit(self, n: int) -> DataFrame:
...
+        )
+        seed = self.validate_constant_type_range(seed, ts.IntType(nullable=False), False, 'seed')
+
+        # analyze stratify list
+        stratify_list: list[exprs.Expr] = []
+        if stratify_by is not None:
+            if isinstance(stratify_by, exprs.Expr):
+                stratify_by = [stratify_by]
+            if not isinstance(stratify_by, list):
```
aaron-siegel: We should allow a `tuple` as well

pixeltable/dataframe.py
```diff
@@ -945,18 +1004,110 @@ def limit(self, n: int) -> DataFrame:
...
+                raise excs.Error('`stratify_by` must be a list of scalar expressions')
+            for expr in stratify_by:
+                if expr is None or not isinstance(expr, exprs.Expr):
+                    raise excs.Error(f'Invalid expression: {expr}')
+                if not expr.col_type.is_scalar_type():
+                    raise excs.Error(f'Invalid type: {expr} must be a scalar type')
```
aaron-siegel: Better to give the type here instead of printing the full expression (which could be a mess)

`Invalid type: expression must be a scalar type (not {expr.col_type})`

pixeltable/exec/sql_node.py
```diff
@@ -284,7 +298,10 @@ async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
                 stmt_str = str(stmt.compile(compile_kwargs={'literal_binds': True}))
                 _logger.debug(f'SqlLookupNode stmt:\n{stmt_str}')
             except Exception:
-                pass
+                # log something if we can't log the compiled stmt
+                stmt_str = repr(stmt)
+                _logger.debug(f'SqlLookupNode proto-stmt:\n{stmt_str}')
+                print(stmt)
```
aaron-siegel: Remove print statement

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +518,114 @@ def _create_stmt(self) -> sql.Select:
...
+        super().__init__(input.tbl, row_builder, select_list, sql_elements, set_pk=True)
+        self.group_by_items = group_by_items
+        self.n_samples = sample_clause._n
+        self.n_per_stratum = sample_clause._n_per_stratum
+        self.fraction_samples = sample_clause._fraction
+        self.seed = sample_clause._seed if sample_clause._seed is not None else 0
+
+    def _create_order_by(self) -> sql.ColumnElement:
+        """Create an expression for randomly ordering rows with a given seed"""
+        rowid_cols = [*self.input_cte.c[-self.pk_count : -1]].copy()  # exclude the version column
```
aaron-siegel: redundant `copy()` (the whole list was just created)

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +518,114 @@ def _create_stmt(self) -> sql.Select:
...
+        assert len(rowid_cols) > 0
+        return SampleClause.key_sql_expr(self.sql_elements, self.seed, rowid_cols)
+
+    def _create_stmt(self) -> sql.Select:
+        if self.fraction_samples is not None:
+            return self._create_stmt_fraction(self.fraction_samples)
+        return self._create_stmt_n(self.n_samples, self.n_per_stratum)
+
+    def _create_stmt_n(self, n: int | None, n_per_stratum: int | None) -> sql.Select:
```
aaron-siegel: I'd say let's try to be consistent in our notation for union types and stick with `Optional[int]` unless we make a principled decision to move to `int | None` throughout the codebase.

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +518,114 @@ def _create_stmt(self) -> sql.Select:
...
+        """Create a Select stmt that returns n samples across all strata"""
+        sql_group_by_items = [self.sql_elements.get(e) for e in self.group_by_items]
+        order_by = self._create_order_by()
+
+        # Create a list of all columns plus the rank
+        # Get all columns from the input CTE dynamically
+        select_columns = [*self.input_cte.c].copy()
```
aaron-siegel: Redundant `copy()`

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +518,114 @@ def _create_stmt(self) -> sql.Select:
...
+        select_columns.append(
+            sql.func.row_number().over(partition_by=sql_group_by_items, order_by=order_by).label('rank')
+        )
+        row_rank_cte = sql.select(*select_columns).select_from(self.input_cte).cte('row_rank_cte')
+
+        final_columns = [*row_rank_cte.c[:-1]].copy()  # exclude the rank column
```
aaron-siegel: redundant `copy()`

there are a bunch more of these below, i'm going to stop commenting on every one

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +518,114 @@ def _create_stmt(self) -> sql.Select:
...
+        if n_per_stratum is not None:
+            return sql.select(*final_columns).filter(row_rank_cte.c.rank <= n_per_stratum)
+        else:
+            return sql.select(*final_columns).order_by(row_rank_cte.c.rank).limit(n)
```
aaron-siegel: I think there needs to be a secondary ordering key to ensure the "tiebreakers" are chosen consistently from among the strata.

`.order_by(row_rank_cte.c.rank, my_invariant_of(sql_group_by_items))`

or some such?

Or even just let the pk be the tiebreaker

`.order_by(row_rank_cte.c.rank, *pk_cols)`

anything that gets us deterministic behavior.

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +518,114 @@ def _create_stmt(self) -> sql.Select:
...
+    seed: int
+    input_cte: Optional[sql.CTE]
+    pk_count: int
+
+    def __init__(
+        self,
+        row_builder: exprs.RowBuilder,
+        input: SqlNode,
+        select_list: Iterable[exprs.Expr],
+        group_by_items: Optional[list[exprs.Expr]] = None,
```
aaron-siegel: Are these the strata keys? Maybe just call them that (the class is sampling-specific anyway, so there's no reason not to use sampling-specific language, unless I'm missing something or we're intending for it to have greater generality)

pixeltable/exec/sql_node.py
aaron-siegel: @mkornacker In the statistical limit they are identical, but if there are many strata then statistical variation could cause measurable distortion (some strata being over/underrepresented at random). The point of the group_by is to enforce that the distribution across strata accurately resembles that of the population.

Political surveys do this as a matter of course, for example

pixeltable/exec/sql_node.py
```diff
@@ -501,3 +518,114 @@ def _create_stmt(self) -> sql.Select:
...
+        select_columns = [*self.input_cte.c].copy()
+        select_columns.append(
+            sql.func.row_number().over(partition_by=sql_group_by_items, order_by=order_by).label('rank')
+        )
+        row_rank_cte = sql.select(*select_columns).select_from(self.input_cte).cte('row_rank_cte')
+
+        # Build the join criterion dynamically to accommodate any number of group by columns
+        join_c = sql.true()
+        for i in range(len(per_strata_count_cte.c) - 1):
+            join_c &= row_rank_cte.c[per_strata_count_cte.c[i].name].isnot_distinct_from(per_strata_count_cte.c[i])
```
aaron-siegel: I'd find this clearer
```python
join_c = sql.and_(
    row_rank_cte.c[col.name].isnot_distinct_from(col)
    for col in per_strata_count_cte.c
)
```

pixeltable/globals.py
```diff
@@ -249,13 +249,17 @@ def create_view(
...
     elif isinstance(base, DataFrame):
         base._validate_mutable('create_view', allow_select=True)
         if len(base._from_clause.tbls) > 1:
             raise excs.Error('Cannot create a view of a join')
         tbl_version_path = base._from_clause.tbls[0]
         where = base.where_clause
+        sample_clause = base.sample_clause
         select_list = base.select_list
+        if sample_clause is not None and not is_snapshot and not sample_clause.is_repeatable:
+            raise excs.Error('Views cannot be created with non-fractional or stratified sampling')
```
aaron-siegel: Clarify: Non-snapshot views

pixeltable/plan.py
```diff
@@ -692,6 +728,53 @@ def _insert_prefetch_node(
...
+    @classmethod
+    def create_sample_clauses(
+        cls,
+        from_clause: FromClause,
+        sample_clause: SampleClause,
+        where_clause: Optional[exprs.Expr],
+        group_by_clause: Optional[list[exprs.Expr]],
+        order_by_clause: Optional[list[tuple[exprs.Expr, bool]]],
+        limit: Optional[exprs.Expr],
+    ) -> tuple[
```
aaron-siegel: This would be a good place to use a `NamedTuple` (it'd make it a lot clearer what the return type actually represents)

pixeltable/utils/sample.py
aaron-siegel: Except `exprs_` not `_exprs`. (Anything *preceded* by an underscore will be treated by the linter as an anonymous variable.)

But personally I prefer `stratify_list` here. The generic `exprs_` is fine when it's a list of expressions with no specific meaning (such as in `InlineList`), but here I think it makes the code clearer if the variable is named for what it is. I'd vote for keeping as is.

pixeltable/utils/sample.py
aaron-siegel: I agree with this, they don't serve a purpose esp. since this is a purely internal class.

pixeltable/utils/sample.py
```diff
@@ -0,0 +1,169 @@
...
+import sqlalchemy as sql
+
+from pixeltable import type_system as ts
+from pixeltable.exprs.expr import DataRow, Expr
+from pixeltable.exprs.literal import Literal
+from pixeltable.exprs.row_builder import RowBuilder
+from pixeltable.exprs.sql_element_cache import SqlElementCache
+
+
+@dataclasses.dataclass
```
aaron-siegel: Per my comments much earlier in this review, I have a fairly strong preference for making this a

```python
@dataclasses.dataclass(frozen=True)
```

and in general, trying to move towards immutable objects for this sort of thing.

pixeltable/utils/sample.py
```diff
@@ -0,0 +1,169 @@
...
+        s = cls(**d_cleaned)
+        if s.is_stratified:
+            s.stratify_list = [Expr.from_dict(e) for e in d_cleaned.get('stratify_list', [])]
+        return s
+
+    def __repr__(self) -> str:
+        s = ','.join(e.display_str(inline=True) for e in self._stratify_list)
+        return (
+            f'sample_{self._version}(n={self._n}, n_per_stratum={self._n_per_stratum}, '
+            + f'fraction={self._fraction}, seed={self._seed}, [{s}])'
```
aaron-siegel: The `+` is unnecessary here (Python will concatenate adjacent strings, and I believe that is marginally more performant)

aaron-siegel (changes requested):

pixeltable/catalog/view.py
jpeterson-pxt: What are you referring to, exactly? I don't think 'is_bound_by()' modifies anything...

pixeltable/dataframe.py
jpeterson-pxt: Show uses limit() which conflicts with the sampling rewrite use of limit. 
DataFrame's not nestable issue.

pixeltable/exec/sql_node.py
jpeterson-pxt: This doesn't exactly work, but I simplified the code a bit.

pixeltable/catalog/table.py
jpeterson-pxt: What would be the semantic meaning? 
`t.col.xxx()` is currently only workable for methods like show(), head(), tail() which produce results. For example, t.col.select() gives an error.

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/dataframe.py
jpeterson-pxt: It checks presence, type, and range. 
I'm not sure what you're thinking, as these seem like validation, not normalization to me.

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/exec/sql_node.py

pixeltable/exec/sql_node.py

pixeltable/exec/sql_node.py

pixeltable/exec/sql_node.py

pixeltable/exec/sql_node.py

pixeltable/utils/sample.py
jpeterson-pxt: Done

pixeltable/globals.py
jpeterson-pxt: Done

pixeltable/utils/sample.py
jpeterson-pxt: While it is logically correct, "frozen=True" does not allow __post_init__ to modify the class. 
Bogus feature of dataclasses, I guess. 
And ruff does not allow assignment to self in an instance method. 
Open to suggestions here.

pixeltable/exec/sql_node.py
jpeterson-pxt: I did it this way because the strata-exprs play the same role as the group_by_list.
But I will change it.

jpeterson-pxt: Done. I used the term "strata_exprs" as @mkornacker indicated that the post-fix ..exprs was the correct term. 
Of course, sometimes these things are not Expr, but sql.junk.

pixeltable/utils/sample.py
jpeterson-pxt: I used the term "stratify_exprs". Everywhere. Except in the interface, where it is called "stratify_by".

pixeltable/exec/sql_node.py
jpeterson-pxt: Good catch. I will use the hash as a secondary key.
Done.

pixeltable/dataframe.py
jpeterson-pxt: Done.

pixeltable/dataframe.py
jpeterson-pxt: Done

jpeterson-pxt:
@mkornacker @aaron-siegel : I have been experimenting with the UDF approach, and there are a few roadblocks.

pixeltable/dataframe.py
```diff
@@ -1026,7 +1023,16 @@ def sample(
         seed: Optional[int] = None,
         stratify_by: Any = None,
     ) -> DataFrame:
-        """Choose a shuffled sample of rows from the DataFrame
+        """
+        Return a new DataFrame specifying a sample of rows from the DataFrame, considered in a shuffled order.
+
+        The size of the sample can be specified in three ways:
+        - `n`: the total number of rows to produce as a sample
```
aaron-siegel, line 1031: I think markdown lists might need to be separated by blank lines from the surrounding text. Not 100% sure though. Might be worth looking at the generated docstring in html to be sure it's formatted the way you expect. You can do this with `mkdocs serve`.

pixeltable/dataframe.py
```diff
@@ -1039,6 +1045,27 @@ def sample(
 
         Returns:
             A new DataFrame which specifies the sampled rows
+
+        Examples:
+            Given the DataFrame person specifying selected rows from a table t with all its columns:
```
aaron-siegel: It'd be clearer to me for the examples to sample directly from a table.

`t.sample(n=100)`

Otherwise users with less pixeltable knowledge might mistakenly conclude you have to construct some dataframe first as a separate operation.

pixeltable/exec/sql_node.py
```diff
@@ -8,7 +8,10 @@
 
 from pixeltable import catalog, exprs
 from pixeltable.env import Env
-from pixeltable.utils.sample import SampleClause
+
+if TYPE_CHECKING:
```
aaron-siegel: Usually the `TYPE_CHECKING` blocks go at the end of the imports section.

pixeltable/plan.py
jpeterson-pxt: Done.

pixeltable/dataframe.py
jpeterson-pxt: I changed it as you requested, but note that some of the other dataframe methods use the "modify a given DataFrame" paradigm.

pixeltable/exec/sql_node.py
jpeterson-pxt: Done.

pixeltable/dataframe.py
jpeterson-pxt: mkdocs does not appear to work for me, even on "main", perhaps we can look at it when you get in.

pixeltable/utils/sample.py
jpeterson-pxt: Done

pixeltable/plan.py
```diff
@@ -82,6 +82,86 @@ def _first_tbl(self) -> catalog.TableVersionPath:
...
+
+    version: Optional[int]
+    n: Optional[int]
+    n_per_stratum: Optional[int]
+    fraction: Optional[float]
+    seed: Optional[int]
+    stratify_exprs: Optional[list[exprs.Expr]]
+
+    DEFAULT_SEED = 0
+    CURRENT_VERSION = 1
```
mkornacker, line 100: What does this mean?

pixeltable/plan.py
```diff
@@ -82,6 +82,86 @@ def _first_tbl(self) -> catalog.TableVersionPath:
...
+
+    def __repr__(self) -> str:
+        s = ','.join(e.display_str(inline=True) for e in self.stratify_exprs)
+        return (
+            f'sample_{self.version}(n={self.n}, n_per_stratum={self.n_per_stratum}, '
+            f'fraction={self.fraction}, seed={self.seed}, [{s}])'
+        )
+
+    @classmethod
+    def fraction_to_md5_hex(cls, fraction: float) -> str:
```
mkornacker, line 147: Not clear how a fraction is related to an md5 hex string; missing comment.

pixeltable/utils/sample.py
jpeterson-pxt: After some experiments: The udf should have a signature like this:
`@udf
def sample_key(seed: int, key_fields: List[Expr]) -> str:`
pixeltable types this signature internally as: 
`def sample_key(seed: int, key_fields: ts.Json) -> str:`

Which is incompatible with the usage. 
Can you suggest a way around this?

pixeltable/plan.py
```diff
@@ -737,43 +817,51 @@ def create_sample_clauses(
...
+            Optional[list[tuple[exprs.Expr, bool]]],
+            Optional[exprs.Expr],
+            Optional[SampleClause],
+        ]:"""
+        """Construct clauses required for sampling under various conditions.
+        If there is no sampling, then return the original clauses.
+        If the sample is stratified, then return onlyt the group by clause. The rest of the
+        mechanism for stratified sampling is provided by the SampleSqlNode.
+        If the sample is non-stratified, then rewrite the query to accommodate the supplied where clause,
+        and provide the other clauses required for sampling
```
mkornacker, line 843: Is there a reason to do this here, rather than in SampleSqlNode? Doing it here comes with a number of disadvantages, such as requiring more restructuring in the upcoming planner work and spreading the blast radius of this logic.

pixeltable/utils/sample.py
```diff
@@ -165,5 +54,30 @@ def sql_expr(self, sql_elements: SqlElementCache) -> Optional[sql.ColumnElement]
...
         """
+        from pixeltable.exec.sql_node import SqlSampleNode
+
         rowid_sql_expr = [e.sql_expr(sql_elements) for e in self.components[1:]]
-        return SampleClause.key_sql_expr(sql_elements, self._seed, rowid_sql_expr)
+        return SqlSampleNode.key_sql_expr(self._seed, rowid_sql_expr)
+
+
+@udf
+def sample_key(seed: int, rowid_sql_exprs: ts.Json) -> str:
```
mkornacker: How is this related to SampleKey?

pixeltable/utils/sample.py
mkornacker: The question is still unanswered: why can't this be done with a udf? (such as the following sample_key())

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/plan.py
jpeterson-pxt: Comments added.

pixeltable/plan.py
jpeterson-pxt: Done

pixeltable/plan.py
jpeterson-pxt: The main reason was to decrease the "blast radius" of this logic by allowing the reuse of the logic in _create_query_plan. 
The SqlSampleNode is not required if the simple query rewrite is applicable.

pixeltable/utils/sample.py
jpeterson-pxt: It is now a udf.

tests/test_sample.py
```diff
@@ -0,0 +1,355 @@
...
+                b = hashlib.md5(str(random.randint(0, 1000000000)).encode()).hexdigest() < threshold_hex
+                if b:
+                    #                print(i, b)
+                    k += 1
+            print(fract, count, k, k / count)
+            self._check_sample_count(fract * count, k)
+
+    @classmethod
+    def _check_sample_count(cls, expected: Union[int, float], actual: int) -> None:
+        assert abs(expected - actual) / actual < 0.25
```
mkornacker, line 155: 25% deviation?

pixeltable/exec/sql_node.py
```diff
@@ -510,3 +526,127 @@ def _create_stmt(self) -> sql.Select:
                 full=join_clause == plan.JoinType.FULL_OUTER,
             )
         return stmt
+
+
+class SqlSampleNode(SqlNode):
+    """
+    Returns rows from a stratified sample with N samples per strata.
```
mkornacker, line 533: per stratum

pixeltable/exec/sql_node.py
```diff
@@ -510,3 +526,127 @@ def _create_stmt(self) -> sql.Select:
...
+    """
+
+    stratify_exprs: Optional[list[exprs.Expr]]
+    n_samples: Optional[int]
+    fraction_samples: Optional[float]
```
mkornacker, line 538: Isn't this contradicting the docstring?

pixeltable/exec/sql_node.py
```diff
@@ -510,3 +526,127 @@ def _create_stmt(self) -> sql.Select:
...
+        self.fraction_samples = sample_clause.fraction
+        self.seed = sample_clause.seed if sample_clause.seed is not None else 0
+
+    @classmethod
+    def key_sql_expr(cls, seed: sql.ColumnElement, sql_cols: Iterable[sql.ColumnElement]) -> sql.ColumnElement:
+        """Construct expression which is the ordering key for rows to be sampled
+        General SQL form is:
+        - MD5(<seed::text> [ + '___' + <rowid_col_val>::text]+
+        """
+        sql_expr: sql.ColumnElement = sql.cast(seed, sql.Text)
```
mkornacker, line 574: type declaration not needed

pixeltable/exec/sql_node.py
```diff
@@ -510,3 +526,127 @@ def _create_stmt(self) -> sql.Select:
...
+        for e in sql_cols:
+            sql_expr = sql_expr + sql.literal_column("'___'") + sql.cast(e, sql.Text)
```
mkornacker, line 576: Why the quotes inside the string? I feel I'm missing something...

pixeltable/exec/sql_node.py
```diff
@@ -510,3 +526,127 @@ def _create_stmt(self) -> sql.Select:
...
+            sql.func.row_number().over(partition_by=sql_strata_exprs, order_by=order_by).label('rank')
+        )
+        row_rank_cte = sql.select(*select_columns).select_from(self.input_cte).cte('row_rank_cte')
+
+        # Build the join criterion dynamically to accommodate any number of group by columns
+        join_c = sql.true()
+        for col in per_strata_count_cte.c[:-1]:
+            join_c &= row_rank_cte.c[col.name].isnot_distinct_from(col)
+
+        # Join srcp with per_strata_count_cte to limit returns to the requested fraction of rows
```
mkornacker, line 643: srcp?

pixeltable/exec/sql_node.py
```diff
@@ -510,3 +526,127 @@ def _create_stmt(self) -> sql.Select:
...
+        # Build a CTE that ranks the rows within each stratum
+        # Include all columns from the input CTE dynamically
+        order_by = self._create_order_by(self.input_cte)
+        select_columns = [*self.input_cte.c]
+        select_columns.append(
+            sql.func.row_number().over(partition_by=sql_strata_exprs, order_by=order_by).label('rank')
+        )
+        row_rank_cte = sql.select(*select_columns).select_from(self.input_cte).cte('row_rank_cte')
+
+        # Build the join criterion dynamically to accommodate any number of group by columns
```
mkornacker, line 638: Columns or expressions? (My understanding was that this works for general exprs...)

pixeltable/plan.py
```diff
@@ -692,6 +818,62 @@ def _insert_prefetch_node(
         prefetch_node = exec.CachePrefetchNode(tbl_id, file_col_info, input_node)
         return prefetch_node
 
+    @classmethod
+    def create_sample_clauses(
```
mkornacker, line 822: This entire function duplicates similar logic in SampleSqlNode. Coalesce the sampling logic there. This would simplify the planner logic (all you need to do is to add a SampleSqlNode at the end).

pixeltable/plan.py
mkornacker: See comment above.

mkornacker (approved):

tests/test_sample.py
jpeterson-pxt: Yes, this happens. The hash as a fraction is not particularly good for small numbers of objects.

pixeltable/exec/sql_node.py
jpeterson-pxt: Done in next version.

pixeltable/exec/sql_node.py
jpeterson-pxt: yes, fixed.

pixeltable/exec/sql_node.py
jpeterson-pxt: Added comment.

pixeltable/exec/sql_node.py
jpeterson-pxt: Done

pixeltable/exec/sql_node.py
jpeterson-pxt: Fixed.

pixeltable/exec/sql_node.py
jpeterson-pxt: This type declaration is required.

PR #598, opened 2025-04-30
PXT-514 Add distinct to table, df and columnref

amithadke:


pixeltable/share/publish.py
```diff
@@ -72,6 +72,7 @@ def clone_snapshot(dest_tbl_uri: str) -> list[FullTableMd]:
     if response.status_code != 200:
         raise excs.Error(f'Error cloning snapshot: {response.text}')
     response_json = response.json()
+    print(response_json)
```
jpeterson-pxt: Is this debugging, or intentional?

tests/test_dataframe.py
```diff
@@ -770,3 +778,89 @@ def test_to_coco(self, reset_db: None) -> None:
...
+            _ = t.select(t.c1, t.c3).distinct().tail(2)
+        assert 'tail() cannot be used with group_by' in str(exc_info.value)
+        with pytest.raises(excs.Error) as exc_info:
+            t.select(t.c1, t.c3).group_by(t.c2).distinct()
+        assert 'Group-by already specified' in str(exc_info.value)
+
+        with pytest.raises(excs.Error) as exc_info:
+            t.distinct().distinct()
+        assert 'Group-by already specified' in str(exc_info.value)
+
```
jpeterson-pxt, line 858: Do expressions like:
t.select(t.city).distinct().count()
work correctly? 
It would be good to have a test of this here, this is a very common use case in SQL.

jpeterson-pxt (commented):
Looks good to me.

pixeltable/share/publish.py
amithadke: yeah was for data sharing will remove it

tests/test_dataframe.py
amithadke: no distinct().count() doesn't work correctly today

aaron-siegel (approved):

pixeltable/dataframe.py
```diff
@@ -848,6 +852,29 @@ def group_by(self, *grouping_items: Any) -> DataFrame:
...
+        """
+        Remove duplicate rows from this DataFrame.
+
+        Note that grouping will be applied to the rows based on the select clause of this Dataframe.
+        In the absence of a select clause, by default, all columns are selected in the grouping.
+
+        Examples:
+            Get all unqiue addresses from table `addresses`.
+
+            >>> results = addresses.distinct()
```
mkornacker, line 868: distinct() doesn't (and shouldn't) call collect(), so this examples is probably misleading.

tests/test_dataframe.py
mkornacker: add a check and test case?

amithadke: check that count() can not be used after distinct (groupby)?

amithadke: Added a check in count for group_by clause

PR #599, opened 2025-04-30
PXT-480: Add DATE type to pixeltable supported types.

jpeterson-pxt:


pixeltable/exprs/literal.py
```diff
@@ -77,9 +80,10 @@ def _as_dict(self) -> dict:
         # how to interpret them unambiguously
         if self.col_type.is_timestamp_type():
             assert isinstance(self.val, datetime.datetime)
-            assert self.val.tzinfo == datetime.timezone.utc  # Must be UTC in a literal
-            # Convert to ISO format in UTC (in keeping with the principle: all timestamps are
-            # stored as UTC in the database)
+            encoded_val = self.val.isoformat()
```
aaron-siegel, line 86: Wait, why are we getting rid of the timezone assertion check and code comment? They seem useful.

pixeltable/io/pandas.py
```diff
@@ -209,7 +209,7 @@ def _df_row_to_pxt_row(
             nval = bool(val)
         elif pxt_type.is_string_type():
             nval = str(val)
-        elif pxt_type.is_timestamp_type():
+        elif pxt_type.is_date_type() or pxt_type.is_timestamp_type():
```
aaron-siegel: I'm confused, won't this result in a Python datetime being returned when pxt_type is a date?

pixeltable/type_system.py
```diff
@@ -81,12 +82,13 @@ class DType(enum.Enum):
...
         FLOAT64 = 11
 
-    scalar_types: ClassVar[set[Type]] = {Type.STRING, Type.INT, Type.FLOAT, Type.BOOL, Type.TIMESTAMP}
+    scalar_types: ClassVar[set[Type]] = {Type.STRING, Type.INT, Type.FLOAT, Type.BOOL, Type.TIMESTAMP, Type.DATE}
     numeric_types: ClassVar[set[Type]] = {Type.INT, Type.FLOAT}
     common_supertypes: ClassVar[dict[tuple[Type, Type], Type]] = {
         (Type.BOOL, Type.INT): Type.INT,
         (Type.BOOL, Type.FLOAT): Type.FLOAT,
         (Type.INT, Type.FLOAT): Type.FLOAT,
+        (Type.DATE, Type.TIMESTAMP): Type.TIMESTAMP,
```
aaron-siegel: I'd argue it's safer to treat date and datetime as incompatible; you can't turn a date into a timestamp without additional information (the time zone). (Also, isn't this the opposite of what Python does? In Python date is a supertype of datetime. Although neither direction makes sense to me...)

pixeltable/type_system.py
```diff
@@ -903,7 +940,11 @@ def from_np_dtype(cls, dtype: np.dtype, nullable: bool) -> Optional[ColumnType]:
             return StringType(nullable=nullable)
 
         if np.issubdtype(dtype, np.datetime64):
-            return TimestampType(nullable=nullable)
+            unit = str(dtype).split('[')[-1].split(']')[0]
```
aaron-siegel: This is really the best way to do this? :/

tests/functions/test_timestamp.py
```diff
@@ -102,7 +107,7 @@ def debug_str() -> str:
             else:
                 raise AssertionError()
 
-    def test_time_zones(self, reset_db: None) -> None:
+    def test_timestamp_zones(self, reset_db: None) -> None:
```
aaron-siegel, line 110: Was this an errant search and replace? :) ("time zone" is pretty safe terminology i'd think)

tests/functions/test_timestamp.py
```diff
@@ -161,7 +166,7 @@ def test_time_zones(self, reset_db: None) -> None:
                         results[method + '_tz'][row_idx] == getattr(effective_dt.astimezone(query_time_zone), method)()
                     )
 
-    def test_time_zone_in_literals(self, reset_db: None) -> None:
+    def test_timestamp_zone_in_literals(self, reset_db: None) -> None:
```
aaron-siegel, line 169: same comment

pixeltable/functions/date.py
```diff
@@ -0,0 +1,175 @@
...
+import sqlalchemy as sql
+
+import pixeltable as pxt
+from pixeltable.utils.code import local_public_names
+
+_SQL_ZERO = sql.literal(0)
+
+# NOT YET SUPPORTED date +/- integer
+# NOT YET SUPPORTED date1 - date2 -> integer
+# NOT YET SUPPORTED timestamp(date)
```
aaron-siegel, line 24: These should probably be explicit conversion UDFs?

```
def midnight(self: date, tz: Optional[str] = None) -> datetime: ...
```
etc?

aaron-siegel (changes requested):

tests/functions/test_timestamp.py
jpeterson-pxt: No, I'd like to be able to select all timestamp tests when testing.

tests/functions/test_timestamp.py
jpeterson-pxt: Same response

pixeltable/utils/arrow.py
```diff
@@ -128,6 +129,11 @@ def _ar_val_to_pxt_val(val: Any, pxt_type: ts.ColumnType) -> Any:
         return bool(val)
     elif pxt_type.is_string_type():
         return str(val)
+    elif pxt_type.is_date_type():
+        if isinstance(val, str):
+            return datetime.datetime.fromisoformat(val).date()
```
jpeterson-pxt: This looks wrong.
Done

pixeltable/type_system.py
jpeterson-pxt: It is correct, but more digging shows another way.
Done

pixeltable/exprs/literal.py
jpeterson-pxt: Good catch, this seems to have been an editing twitch.
Done

pixeltable/functions/date.py
jpeterson-pxt: Not sure what you're getting at here?
I left these as out-of-scope for the current iteration.

pixeltable/io/pandas.py
jpeterson-pxt: There was more than one thing wrong here, both with the new code and the existing code. 
Fixed.

pixeltable/type_system.py
jpeterson-pxt: If a column of incoming data has both dates and timestamps, e.g. "2025-01-01" and "2025-01-01T00:00:00" should we type it as timestamp or string (the default)?
Perhaps string is better in some ways.

tests/functions/test_date.py
```diff
@@ -0,0 +1,114 @@
...
+            elif isinstance(mref, exprs.FunctionCall):
+                # property
+                assert mref.fn.name == pxt_fn.name, pxt_fn
+            else:
+                raise AssertionError()
+
+    def test_date_make(self, reset_db: None) -> None:
+        Env.get().default_time_zone = ZoneInfo('America/Anchorage')
+        test_dts, t = self.make_test_table()
+        from pixeltable.functions.date import make_date
```
aaron-siegel: this can probably go at top of file.

tests/functions/test_timestamp.py
aaron-siegel: You can do that via `pytest -v tests/functions/test_timestamp.py` (to select the Python file) or `pytest -vk TestTimestamp` (to select all tests in the class). Unless there's a reason why those methods aren't adequate here, I'd prefer to stick with more natural test names? If I were encountering this for the first time, `test_time_zones` would be self-explanatory, but `test_timestamp_zones` would have me asking myself, "What's a timestamp zone?"

aaron-siegel (approved):

tests/functions/test_date.py
jpeterson-pxt: Moved all pixeltable.functions imports to the top of file.

pixeltable/functions/date.py
```diff
@@ -0,0 +1,175 @@
...
+from datetime import date
+
+import sqlalchemy as sql
+
+import pixeltable as pxt
+from pixeltable.utils.code import local_public_names
+
+_SQL_ZERO = sql.literal(0)
+
+# NOT YET SUPPORTED date +/- integer
```
mkornacker, line 22: remove (there is no consensus we want this)

PR #611, opened 2025-05-06
Support for concurrent table operations

mkornacker:
Adds logic to validate and reload table metadata on transaction boundaries.
    
Not yet covered:
- concurrent updates of a base table and views; those might result in deadlocks or midway-aborted transactions
- concurrent drop_column()


mkornacker:
What's still missing:

The Labelstudio-related code is broken, because it caches metadata. All Labelstudio-related tests currently fail. However, the fixes won't affect the core catalog logic, which is ready to review now.

tool/random_dir_ops.py
```diff
@@ -20,37 +17,11 @@ def list_objects(dir: str) -> tuple[list[str], list[str]]:
     return tbls, subdirs
 
 
-def log_op(msg: str) -> None:
-    timestamp = datetime.fromtimestamp(time.time())
-    print(timestamp, ospid, msg)
```
jpeterson-pxt, line 25: Why not keep the timestamp and PID in this tool's log? 
They are useful when examining the exact timing of events across threads/processes.

pixeltable/io/table_data_conduit.py
```diff
@@ -23,7 +23,6 @@
 
 _logger = logging.getLogger('pixeltable')
 
-# ---------------------------------------------------------------------------------------------------------
```
jpeterson-pxt, line 26: I found these comments useful, why remove them?

tool/random_tbl_ops.py
```diff
@@ -0,0 +1,69 @@
...
+
+    if is_data_op:
+        # add or delete rows
+        max_val = t.select(pxtf.max(t.c1)).collect()[0, 0]
+        if max_val is None:
+            max_val = 0
+        is_update = random.uniform(0, 1) < 0.1
+        if is_update:
+            # update rows
+            print(f'updating rows: c1 > {max_val - 10}')
```
jpeterson-pxt, line 27: Should this tool use the "log_op" mechanism to add timestamps and PIDs to the per-process logs, so that the relative ordering of the operations across the threads can be examined?

jpeterson-pxt (commented):
A few quick comments, not an in-depth review.

pixeltable/io/table_data_conduit.py
mkornacker: What are they useful for? We don't have formatting like that in any other source file.

tool/random_dir_ops.py
mkornacker: There's a random_dir_ops2 that preserves everything. I didn't find the output useful (plus, the pid is a constant). If you need to examine the actual timing, you'll have to look in the logs. Every process starts a new log file.

tool/random_tbl_ops.py
mkornacker: I'll add a timestamp, but I don't see the utility of printing a constant (the pid).

pixeltable/catalog/catalog.py
```diff
@@ -114,17 +120,96 @@ def clear(cls) -> None:
...
+
+    @contextmanager
+    def begin_xact(self, *, tbl_id: Optional[UUID] = None, for_write: bool = False) -> Iterator[sql.Connection]:
+        """
+        Return a context manager that yields a connection to the database. Idempotent.
+
+        It is mandatory to call this method, not Env.begin_xact(), if the transaction accesses any table data
+        or metadata.
+        """
+        if not Env.get().in_xact:
```
aaron-siegel: I'd find this easier to read with the conditional inverted

```
if Env.get().in_xact:
    yield Env.get().conn
else:
    ...
```

Or even
```
if Env.get().in_xact:
    yield Env.get().conn
    return

...
```

Another question though, is there a risk that this will be called in the context of an existing `Env.begin_xact()` block, but one that did not involve acquiring an x-lock for the desired table? Would it be safer to build in some logic to ensure that we do have the requisite x-lock, independent of the context of the existing transaction?

```
if Env.get().in_xact:
    reacquire_xlock(Env.get().conn, tbl_id, for_write)
    yield Env.get().conn
```

pixeltable/catalog/catalog.py
```diff
@@ -114,17 +120,96 @@ def clear(cls) -> None:
...
+                    if _logger.isEnabledFor(logging.DEBUG):
+                        self.validate()
+        else:
+            yield Env.get().conn
+
+    @property
+    def in_write_xact(self) -> bool:
+        return self._in_write_xact
+
+    def _xlock_dir(self, parent_id: Optional[UUID], dir_id: Optional[UUID], dir_name: Optional[str]) -> None:
```
aaron-siegel: `_acquire_dir_xlock()` would be clearer to me

pixeltable/catalog/table.py
```diff
@@ -1146,7 +1180,7 @@ def _drop_index(
         else:
             if col.tbl.id != self._tbl_version.id:
                 raise excs.Error(
-                    f'Column {col.name!r}: cannot drop index from column that belongs to base ({col.tbl.get().name}!r)'
+                    f'Column {col.name!r}: cannot drop index from column that belongs to base ({col.tbl.name}!r)'
```
aaron-siegel: `!r` outside of brackets will print the literal string `!r` to the user

(I realize this was a preexisting issue but i just noticed it)

pixeltable/catalog/table_version.py
```diff
@@ -106,19 +108,12 @@ def __init__(
         mutable_views: list[TableVersionHandle],
         base_path: Optional[pxt.catalog.TableVersionPath] = None,
         base: Optional[TableVersionHandle] = None,
-        # base_store_tbl: Optional['store.StoreBase'] = None,
     ):
+        self.is_validated = True  # a freshly constructed instance is always valid
         self.id = id
-        self.name = tbl_md.name
-        self.user = tbl_md.user
+        self._tbl_md = copy.deepcopy(tbl_md)
```
aaron-siegel, line 114: I like this change a lot (metadata as ground truth, rather than re-created from the class instance contents when it's needed)

pixeltable/catalog/table_version.py
```diff
@@ -317,6 +304,14 @@ def create(
     def drop(self) -> None:
         from .catalog import Catalog
 
+        if self.is_view and self.is_mutable:
+            # update mutable_views
+            from .table_version_handle import TableVersionHandle
+
+            assert self.base is not None
+            if self.base.get().is_mutable:
```
aaron-siegel, line 337: Should this be an assertion rather than an `if`? (My understanding from the preceding code is that only a mutable `TableVersion` can have mutable views, but I may be missing something)

pixeltable/catalog/table_version.py
```diff
@@ -420,28 +415,32 @@ def _init_sa_schema(self) -> None:
         else:
             self.store_tbl = StoreTable(self)
 
-    def _update_md(
-        self, timestamp: float, update_tbl_version: bool = True, preceding_schema_version: Optional[int] = None
-    ) -> None:
+    def _store_md(self, new_version: bool, new_version_ts: float, new_schema_version: bool) -> None:
```
aaron-siegel: Partly because we use "store" as a noun so much, I get confused when it gets used as a verb; this could validly mean either "store some metadata in the database" or "tell me something about the metadata of the store table".

Would `_write_md` or `_commit_md` be clearer?

pixeltable/catalog/table_version.py
```diff
@@ -420,28 +415,32 @@ def _init_sa_schema(self) -> None:
...
-        version_md = self._create_version_md(timestamp) if update_tbl_version else None
-        schema_version_md = (
-            self._create_schema_version_md(preceding_schema_version) if preceding_schema_version is not None else None
+        version_md: Optional[schema.TableVersionMd] = (
+            schema.TableVersionMd(
+                tbl_id=str(self.id),
+                created_at=new_version_ts,
+                version=self.version,
+                schema_version=self.schema_version,
+                additional_md={},
```
aaron-siegel, line 460: Since we're switching to metadata as ground truth, it may make sense to track `additional_md` alongside `tbl_md` and `schema_version_md`. This will ensure it gets preserved version over version. (That has no functional impact now, but may become relevant when we start enabling server-side table operations in the cloud)

pixeltable/catalog/table_version_path.py
```diff
@@ -47,36 +54,61 @@ def as_md(self) -> schema.TableVersionPath:
...
 
+    def refresh_cached_md(self) -> None:
+        from pixeltable.catalog import Catalog
+
+        if Env.get().in_xact:
+            # when we're running inside a transaction, we need to make sure to supply current metadata
+            if self._cached_tbl_version is not None and self._cached_tbl_version.is_validated:
+                # nothing to refresh
+                return
+        elif self._cached_tbl_version is not None:
```
aaron-siegel, line 75: I'm confused about why it's ok for `is_validated` to be `False` in this case, would welcome a discussion on this to educate me here

pixeltable/catalog/table_version_path.py
```diff
@@ -47,36 +54,61 @@ def as_md(self) -> schema.TableVersionPath:
...
+            return
+
+        with Catalog.get().begin_xact(for_write=False):
+            self._cached_tbl_version = self.tbl_version.get()
+
+    def clear_cached_md(self) -> None:
+        self._cached_tbl_version = None
+        if self.base is not None:
+            self.base.clear_cached_md()
```
aaron-siegel, line 84: `clear_cached_md()` is recursive but `refresh_cached_md()` is not? That's a bit counterintuitive. Maybe put a note in a docstring clarifying that, or add `recursive: bool` as a parameter to `clear_cached_md()` to make things totally clear.

pixeltable/catalog/catalog.py
```diff
@@ -56,40 +58,43 @@ def _unpack_row(
     return result
 
 
-_MAX_RETRIES = 3
+_MAX_RETRIES = 0
```
aaron-siegel, line 63: Is this change temporary? i.e. is the intent to set this back to `3` or some such when we ship? If so, might be good to have it controlled by an environment variable or config setting rather than by changing code in place?

pixeltable/catalog/catalog.py
```diff
@@ -114,17 +120,96 @@ def clear(cls) -> None:
...
         self._tbl_versions = {}
         self._tbls = {}  # don't use a defaultdict here, it doesn't cooperate with the debugger
+        self._in_write_xact = False
         self._init_store()
 
-    @classmethod
-    def _lock_dir(cls, parent_id: Optional[UUID], dir_id: Optional[UUID], dir_name: Optional[str]) -> None:
-        """Update directory record(s) to sequentialize thread access. Lock is released when transaction commits.
+    def validate(self) -> None:
+        """Validate structural consistency of cached metadata"""
```
aaron-siegel, line 152: Add a note that it only gets called in debug mode

pixeltable/catalog/catalog.py
```diff
@@ -580,14 +671,23 @@ def __store_replica_md(self, path: Path, md: schema.FullTableMd) -> None:
...
+        return obj
+
+    def _get_table(self, path: Path) -> Table:
         obj = Catalog.get()._get_schema_object(path, expected=Table, raise_if_not_exists=True)
         assert isinstance(obj, Table)
-        obj._tbl_version.get().ensure_md_loaded()
+        tbl_version = obj._tbl_version.get()
+        tbl_version.ensure_md_loaded()
+        if not (isinstance(obj, View) and obj._snapshot_only):
+            # if this table has mutable views, we need to load those as well, in order to record column dependencies
```
aaron-siegel: add a note explaining why we can skip this if it's a pure snapshot? that wasn't immediately obvious to me

pixeltable/catalog/catalog.py
```diff
@@ -723,17 +823,31 @@ def get_view_ids(self, tbl_id: UUID, for_update: bool = False) -> list[UUID]:
...
+
+                # the stored version can be behind TableVersion.version, because we don't roll back the in-memory
+                # metadata changes after a failed update operation
+                if current_version != tv.version:
+                    # the cached metadata is invalid
+                    _logger.debug(
+                        f'reloading metadata for table {tbl_id} '
+                        f'(cached version: {tv.version}, current version: {current_version})'
+                    )
+                    tv = self._load_tbl_version(tbl_id, effective_version)
```
aaron-siegel: might was well say `_load_tbl_version(tbl_id, None)` here which makes it slightly clearer what's going on.

pixeltable/catalog/catalog.py
```diff
@@ -723,17 +823,31 @@ def get_view_ids(self, tbl_id: UUID, for_update: bool = False) -> list[UUID]:
...
-        if tbl_version.is_view and tbl_version.effective_version is None:
-            base = tbl_version.base.get()
-            base.mutable_views.append(TableVersionHandle(tbl_version.id, tbl_version.effective_version))
+        # we need a transaction here, if we're not already in one; if this starts a new transaction,
+        # the returned TableVersion instance will not be validated
+        with self.begin_xact(tbl_id=tbl_id, for_write=False) as conn:
+            tv = self._tbl_versions.get((tbl_id, effective_version))
+            if tv is None:
+                tv = self._load_tbl_version(tbl_id, effective_version)
+            elif not tv.is_validated and effective_version is None:
```
aaron-siegel: What if `not tv.is_validated` but `effective_version is not None`? This can happen after a `_revert`. Maybe it's not possible for this to happen here due to careful manipulation of the `_tbl_versions` dict, in which case an assertion to make that clear is probably warranted

pixeltable/catalog/catalog.py
```diff
@@ -723,17 +823,31 @@ def get_view_ids(self, tbl_id: UUID, for_update: bool = False) -> list[UUID]:
...
+                # metadata changes after a failed update operation
+                if current_version != tv.version:
+                    # the cached metadata is invalid
+                    _logger.debug(
+                        f'reloading metadata for table {tbl_id} '
+                        f'(cached version: {tv.version}, current version: {current_version})'
+                    )
+                    tv = self._load_tbl_version(tbl_id, effective_version)
+
+            tv.is_validated = True
```
aaron-siegel: It's not clear to me why this is necessary. Shouldn't it be the case that if `tv` was invalid, we created a new one, which is necessarily valid? Maybe I'm missing something here.

pixeltable/catalog/catalog.py
```diff
@@ -114,17 +120,96 @@ def clear(cls) -> None:
...
+                        _logger.debug(f'Retrying ({num_retries}) after {type(e.orig)}')
+                        time.sleep(random.uniform(0.1, 0.5))
+                    else:
+                        raise
+                finally:
+                    self._in_write_xact = False
+
+                    # invalidate cached current TableVersion instances
+                    for tv in self._tbl_versions.values():
+                        tv.is_validated = False
```
aaron-siegel, line 239: I'd be more comfortable with a `tv.invalidate()` method (instead of reaching inside the class to mess with its member variables)

pixeltable/catalog/catalog.py
```diff
@@ -114,17 +120,96 @@ def clear(cls) -> None:
...
+                        if tv.effective_version is None:
+                            _logger.debug(f'invalidating table version {tv.id}:{tv.version}')
+                            tv.is_validated = False
```
aaron-siegel: Wait, didn't we just do this 3 lines above?

pixeltable/catalog/catalog.py
mkornacker: Let's leave it at 0 for now. It's not clear what a good setting is (I've seen it do 10 retries in a row, but doesn't turn into a livelock). I'm not sure this should be an env var (how does the user come up with the correct setting?).

pixeltable/catalog/catalog.py
mkornacker: That's really up to the caller, I'm not sure that assumption should be reflected here.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: It's just a flag that would otherwise need to get stored inside Catalog itself. TableVersion doesn't participate in the invalidation. From Catalog's perspective, TableVersion is a white box.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: check not needed, removed.

pixeltable/catalog/catalog.py
mkornacker: is_validated only applies to live versions, ie, this can't happen. revert() doesn't change that.

Added a longer explanation in the class docstring.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: superfluous, removed.

pixeltable/catalog/table.py

pixeltable/catalog/table_version.py
mkornacker: Yeah, there'll be more changes around that.

pixeltable/catalog/table_version.py
mkornacker: You can have a mutable view on top of a snapshot.

pixeltable/catalog/table_version.py
mkornacker: Renamed to write_md.

pixeltable/catalog/table_version.py
mkornacker: What do you mean, track? additional_md is part of _tbl_md, because that's how it's stored.

pixeltable/catalog/table_version_path.py
mkornacker: Expanded on the class docstring.

pixeltable/catalog/table_version_path.py
mkornacker: Done

mkornacker:
Feel free to hold off on reviewing the latest changes, I'm planning on making some more changes needed for the tests to pass.

pixeltable/catalog/table_version.py
aaron-siegel: Aha of course, sorry I was confused. This code is for creating a new table version, which should always be initialized with empty additional_md. (Table versions have additional_md fields as well as tables)

pixeltable/catalog/catalog.py
aaron-siegel: The env var would be more for testing/dev purposes than for user tuning (we could set it to 0 for debugging if we needed to, for example).

pixeltable/catalog/catalog.py
aaron-siegel: Got it, so if it's a fixed version, then is_validated might or might not be set, but has no meaning either way?

pixeltable/catalog/catalog.py
aaron-siegel: The latest commit addressed the first comment (about code organization) but not the second one (what happens if we're already in a transaction, but it's a transaction that was opened in a different way and didn't involve acquiring the requisite table xlock?) I remember being concerned about it when I read the code. If you say it's not an issue then I believe you, but it'd be good to walk through it at some point to be sure I understand it.

aaron-siegel (approved):

pixeltable/catalog/catalog.py
mkornacker: I'd be more comfortable with 0 at this point, that would guarantee that it doesn't error out. I've seen ~10 retries in a row, but nothing that just keeps retrying. Since this doesn't degrade the current (= non-concurrent) user experience, I think it's okay to leave it there.

pixeltable/catalog/catalog.py
mkornacker: Good point, fixed. It needs to check for the presence of that lock, otherwise it's an error (you can't ask for it in the middle of the transaction).

pixeltable/catalog/catalog.py
mkornacker: Good point, I changed the condition to remove ambiguity.

pixeltable/catalog/catalog.py
```diff
@@ -171,9 +184,19 @@ def begin_xact(self, *, tbl_id: Optional[UUID] = None, for_write: bool = False)
         or metadata.
         """
         if Env.get().in_xact:
+            if tbl_id is not None and for_write:
+                # make sure that we requested the required table lock at the beginning of the transaction
+                assert tbl_id == self._x_locked_tbl_id, f'{tbl_id} != {self._x_locked_tbl_id}'
```
aaron-siegel, line 196: (this isn't a blocking comment for the PR but something to think about for the future) I wonder if there are scenarios where we might need xlocks for multiple tables in the same transaction? If possible, it'd be more flexible to acquire additional xlocks as needed, rather than assert as here

pixeltable/catalog/table_version.py
```diff
@@ -634,7 +638,7 @@ def add_columns(
         all_cols: list[Column] = []
         for col in cols:
             all_cols.append(col)
-            if self._is_btree_indexable(col):
+            if col.name is not None and self._is_btree_indexable(col):
```
aaron-siegel, line 643: Should we alias `col.name is not None` with something like `col.is_anonymous`?

pixeltable/share/packager.py
```diff
@@ -247,13 +248,26 @@ def restore(self, bundle_path: Path) -> pxt.Table:
...
+        cat.create_replica(catalog.Path(self.tbl_path), tbl_md)
+        # don't call get_table() until after the calls to create_replica() and __import_table() below;
+        # the TV instances created by get_table() would be replaced by create_replica(), which creates duplicate
+        # TV instances for the same replica version, which then leads to failures when constructing queries
 
         # Now we need to instantiate and load data for replica_tbl and its ancestors, except that we skip
         # replica_tbl itself if it's a pure snapshot.
-        if replica_tbl._id != replica_tbl._tbl_version.id:
+        target_md = tbl_md[0]
+        is_pure_snapshot = (
```
aaron-siegel, line 271: Is this really preferable to comparing the ids? :)

Maybe move this into a `FullTableMd.is_pure_snapshot` property?

pixeltable/share/packager.py
```diff
@@ -267,7 +281,8 @@ def restore(self, bundle_path: Path) -> pxt.Table:
                 _logger.info(f'Importing table {tv.name!r}.')
                 self.__import_table(self.tmp_dir, tv, md)
 
-        return replica_tbl
+        with cat.begin_xact(for_write=False):
```
aaron-siegel, line 290: I think this fixes a loose end in the packager implementation (I'll confirm when everything is merged) - thanks!

pixeltable/share/packager.py
```diff
@@ -65,16 +66,16 @@ def package(self) -> Path:
...
-        self.iceberg_catalog = sqlite_catalog(self.tmp_dir / 'warehouse')
+        self.tables_dir = self.tmp_dir / 'tables'
+        self.tables_dir.mkdir()
         with catalog.Catalog.get().begin_xact(for_write=False):
-            ancestors = (self.table, *self.table._base_tables)
-            for t in ancestors:
-                _logger.info(f"Exporting table '{t._path}'.")
-                self.__export_table(t)
+            for tv in self.table._tbl_version_path.get_tbl_versions():
+                _logger.info(f"Exporting table '{tv.get().name}:{tv.get().version}'.")
```
aaron-siegel: Note: There's a `TableVersion.versioned_name` now, so you can do

```
'Exporting table {tv.get().versioned_name!r}.'
```

aaron-siegel (approved):

pixeltable/catalog/catalog.py
mkornacker: Added explanation why that doesn't work.

pixeltable/catalog/table_version.py
mkornacker: We should really add a separate flag. The names are useful for debugging.

pixeltable/share/packager.py
mkornacker: This needs to be restructured anyway, as per prior comment.

pixeltable/share/packager.py
mkornacker: Done.

PR #612, opened 2025-05-06
PXT-368: group_by requires a select specification.

jpeterson-pxt:


jpeterson-pxt:
Tests are failing, more work needs to be done.

PR #620, opened 2025-05-13
Update yolox docs

jacobweiss2305:
update example to use native integration


pierrebrunelle:
What about the Readme as mentioned?

pierrebrunelle:
@aaron-siegel approve when you have a second. Thanks!

PR #625, opened 2025-05-16
Merge different versions of base tables consistently when pulling replicas

aaron-siegel:
Introduces logic for doing a consistent data merge in the case where two replicas have a common base table.


pixeltable/share/packager.py
```diff
@@ -276,11 +282,127 @@ def __import_table(self, bundle_path: Path, tv: catalog.TableVersion, tbl_md: sc
...
+        # that have never been observed.
+
+        pk_predicates = [col == temp_cols[col.name] for col in tv.store_tbl.pk_columns()]
+        pk_clause = sql.and_(*pk_predicates)
+
+        # If the same pk exists in both the temporary table and the existing table, then the corresponding row data
+        # must be identical; the rows can differ only in their v_max value. As a sanity check, we go through the
+        # motion of verifying this; a failure implies data corruption in either the replica being imported or in a
+        # previously imported replica.
+        # The pxtmedia:// URLs of local media files are constructed using the SHA256 hash of the media file contents,
```
mkornacker: Why do we need to compute the hash of generated media files? That sounds expensive.

pixeltable/share/packager.py
```diff
@@ -276,11 +282,127 @@ def __import_table(self, bundle_path: Path, tv: catalog.TableVersion, tbl_md: sc
...
+        # so this works for media files too, providing bytewise validation of data integrity.
+        non_vmax_temp_cols = [col for col_name, col in temp_cols.items() if col_name != 'v_max']
+        non_vmax_store_cols = [store_sa_tbl.c[col_name] for col_name in temp_cols if col_name != 'v_max']
+        q = sql.select(*non_vmax_temp_cols, *non_vmax_store_cols).where(pk_clause)
```
mkornacker: Also add a where clause that selects only non-matching rows. That should run a lot faster than retrieving all of the row data and doing the comparisons in python.

pixeltable/functions/math.py
```diff
@@ -100,6 +100,69 @@ def _(self: sql.ColumnElement, digits: Optional[sql.ColumnElement] = None) -> sq
         return sql.func.round(sql.cast(self, sql.Numeric), sql.cast(digits, sql.Integer))
 
 
+@pxt.udf(is_method=True)
```
mkornacker, line 103: How is this related to replica merging?

tests/share/test_packager.py
```diff
@@ -153,30 +154,53 @@ def __check_media(self, pxt_values: list, parquet_values: list, media_dir: Path)
...
+        result_set: DataFrameResultSet  # Resultset corresponding to the query `tbl.head(n=5000)`
+
+    def __package_table(self, tbl: pxt.Table) -> BundleInfo:
+        """
+        Runs the query `tbl.head(n=5000)`, packages the table into a bundle, and returns a BundleInfo.
+        """
+        assert tbl._tbl_version.get().is_snapshot
+
+        schema = tbl._schema
+        depth = len(tbl._tbl_version_path.ancestor_paths)
```
mkornacker: Use TVP.path_len() instead.

It doesn't look like there are any legitimate use cases for ancestor_paths.

tests/share/test_packager.py
```diff
@@ -224,3 +248,78 @@ def test_iterator_view_round_trip(self, reset_db: None) -> None:
...
         t_replica = v_replica.base_table
         assert t_replica.count() == 2
+
+    def test_multi_view_round_trip_1(self, reset_db: None) -> None:
+        """
+        Simplest multi-view test: two snapshots that are exported at the same time.
+        (All v_min/v_max values are consistent in the bundles.)
+        """
+
+        t = pxt.create_table('base_tbl', {'int_col': pxt.Int})
```
mkornacker, line 267: Why not use the standard test table schema, or something else with more columns and types? I'm concerned about missing corner cases.

pixeltable/functions/math.py
aaron-siegel: It's not. I needed it for one of the tests so I snuck it in.

pixeltable/share/packager.py
aaron-siegel: We don't "need" to but there are two reasons for it:
- When we do the comparison below to check for data corruption, it also validates that any media data are identical
- If there are duplicate media files, it will ensure we dedup them here, reducing the bundle size and saving on network and storage costs

I don't think SHA256 calculation is particularly expensive compared to the other things we do (building the bundle, compressing data, uploading data etc). The most expensive part is reading the data from disk, which we have to do anyway to build the bundle. If it becomes a problem we could reorganize things slightly to ensure we're only reading the file once for both hashing and archiving.

pixeltable/share/packager.py
aaron-siegel: Are you sure this will work correctly for long string columns and binary columns?

tests/share/test_packager.py

I got rid of ancestor_paths.

tests/share/test_packager.py
aaron-siegel: Just trying to isolate functionality in the test to make debugging easier if something goes wrong.

I'll add some more tests to cover additional datatypes and also complex sequences of row/column insert/delete operations.

pixeltable/share/packager.py
aaron-siegel: I ran a few tests that suggested yes it will work correctly, so I went ahead and made the change (to do the data integrity check in SQL).

pixeltable/share/packager.py
```diff
@@ -343,22 +343,35 @@ def __import_table(self, bundle_path: Path, tv: catalog.TableVersion, tbl_md: sc
         # previously imported replica.
         # The pxtmedia:// URLs of local media files are constructed using the SHA256 hash of the media file contents,
         # so this works for media files too, providing bytewise validation of data integrity.
-        non_vmax_temp_cols = [col for col_name, col in temp_cols.items() if col_name != 'v_max']
-        non_vmax_store_cols = [store_sa_tbl.c[col_name] for col_name in temp_cols if col_name != 'v_max']
-        q = sql.select(*non_vmax_temp_cols, *non_vmax_store_cols).where(pk_clause)
+
+        pk_col_names = {col.name for col in tv.store_tbl.pk_columns()}
```
mkornacker: Add a StoreBase.system_columns() that returns [*_pk_cols, v_min_col, v_max_col], analogous to _create_system_columns(). That way, you don't need to keep special-casing v_max.

pixeltable/share/packager.py
```diff
@@ -343,22 +343,35 @@ def __import_table(self, bundle_path: Path, tv: catalog.TableVersion, tbl_md: sc
...
+        ]
+        value_temp_cols = [
+            col for col_name, col in temp_cols.items() if col_name not in pk_col_names and col_name != 'v_max'
+        ]
+        mismatch_predicates = [store_col != temp_col for store_col, temp_col in zip(value_store_cols, value_temp_cols)]
+        mismatch_clause = sql.or_(*mismatch_predicates)
+
+        # This query looks for rows that have matching primary keys (rowid + v_min), but differ in at least one value
+        # column.
+        q = sql.select(*value_store_cols, *value_temp_cols).where(pk_clause).where(mismatch_clause)
```
mkornacker, line 378: Would be good to add the SQL as an explanatory comment, along the lines of what's in catalog.py:load_tbl_md().

pixeltable/share/packager.py
aaron-siegel: Done.

pixeltable/share/packager.py
aaron-siegel: Done.

PR #631, opened 2025-05-21
Include preview data in request when publishing a table

aaron-siegel:
Also incorporates a slight refactoring of `formatter.py` in order to recycle some of the functionality used when rendering notebooks.


pixeltable/share/packager.py
```diff
@@ -200,6 +220,92 @@ def __build_tarball(self) -> Path:
...
+        Extract a preview of the table data for display in the UI.
+
+        In order to bound the size of the output data, all "unbounded" data types are resized:
+        - Strings are abbreviated as per Formatter.abbreviate()
+        - Arrays and JSON are shortened and formatted as strings
+        - Images are resized to thumbnail size as a base64-encoded webp
+        - Videos are replaced by their first frame and resized as above
+        - Documents are replaced by a thumbnail as a base64-encoded webp
+        """
+        # First 8 columns
```
mkornacker, line 240: Why not include all columns? That gives the UI more flexibility (meaning we don't have to re-write bundles if we decide to show more columns or a different subset), and there isn't really a downside.

pixeltable/share/packager.py
```diff
@@ -200,6 +220,92 @@ def __build_tarball(self) -> Path:
...
+        preview_cols = dict(itertools.islice(self.table._schema.items(), 0, 8))
+        select_list = [self.table[col_name] for col_name in preview_cols]
+        # First 5 rows
```
mkornacker, line 243: 5 seems stingy. 20?

pixeltable/share/packager.py
```diff
@@ -200,6 +220,92 @@ def __build_tarball(self) -> Path:
...
+        rows = list(self.table.select(*select_list).head(n=5))
+
+        preview_header = {col_name: str(col_type._type) for col_name, col_type in preview_cols.items()}
+        preview = [
+            [self.__encode_preview_data(val, col_type)]
+            for row in rows
+            for val, col_type in zip(row.values(), preview_cols.values())
+        ]
+
+        return preview_header, preview
```
mkornacker, line 253: Is the header needed or redundant, given that the receiver already has the schema?

pixeltable/share/packager.py
```diff
@@ -200,6 +220,92 @@ def __build_tarball(self) -> Path:
...
+
+            case ts.ColumnType.Type.DOCUMENT:
+                assert isinstance(val, str)
+                return self.__encode_document(val)
+
+            case _:
+                raise AssertionError(f'Unrecognized column type: {col_type._type}')
+
+    def __encode_image(self, img: PIL.Image.Image) -> str:
+        if img.height > img.width * 1.5:
```
mkornacker, line 303: Explain rationale

pixeltable/share/packager.py
```diff
@@ -200,6 +220,92 @@ def __build_tarball(self) -> Path:
                 tf.add(src_file, arcname=f'media/{dest_name}')
         return bundle_path
 
+    def __extract_preview_data(self) -> tuple[dict[str, str], list[list[Any]]]:
+        """
+        Extract a preview of the table data for display in the UI.
+
+        In order to bound the size of the output data, all "unbounded" data types are resized:
+        - Strings are abbreviated as per Formatter.abbreviate()
```
mkornacker, line 234: Should we send unabbreviated string (and json) data and let the UI decide how much to abbreviate? A single column value that is stored in Postgres is bounded in size.

pixeltable/share/packager.py
aaron-siegel: This is all part of the synchronous API call, and it will be loaded when the table is browsed, so we're really trying to bound the data size.

If a table has 50 columns and we do 50 columns x 20 rows and there are a lot of images, the request could get pretty fat.

pixeltable/share/packager.py
aaron-siegel: I thought we were allowing large strings in postgres?

What is the limit (in bytes) for strings/JSON?

pixeltable/share/packager.py
aaron-siegel: see above

pixeltable/share/packager.py
aaron-siegel: It's a tiny amount of data and I'm just trying to make things easier on the service. I see no harm in including it. Otherwise it's an extra Pixeltable SDK call on the server side to look up the columns in the schema.

pixeltable/share/packager.py
aaron-siegel: Just a heuristic to make things look nice; I'll add a comment

pixeltable/share/packager.py
mkornacker: I just looked it up and was surprised: it's 1GB, basically unbounded. We could impose a much smaller limit (1K?) that would still go unnoticed (as a limit) in most use cases.

aaron-siegel: Ok, now I'm not sure what you're proposing. What needs to be resolved before we can merge this PR?

Note that everything being done here is very similar to what happens when it's rendered in a notebook.

PR #639, opened 2025-06-02
Allow HEIC images

aaron-siegel:


tests/test_table.py
```diff
@@ -916,6 +916,13 @@ def test_create_s3_image_table(self, reset_db: None) -> None:
         cache_stats = FileCache.get().stats()
         assert cache_stats.total_size == 0
 
+    def test_image_formats(self, reset_db: None) -> None:
+        tbl = pxt.create_table('test', {'img': pxt.Image})
+        files = [
+            'sewing-threads.heic'  # HEIC format
```
mkornacker, line 922: Roll this into get_image_files(), maybe with a new format parameter if we need to pick out something specific?

That way, it would be automatically included in other tests.

mkornacker (approved):

tests/test_table.py
aaron-siegel: the test image data is a bit of a mess right now, I was hoping to ship this without having to solve that problem.

I'll create a rough-edges ticket for it.

PR #640, opened 2025-06-02
Add table.history() method to return a user-readable list of known changes to a table. 

jpeterson-pxt:


history_demo.ipynb
```diff
@@ -0,0 +1,258 @@
+{
```
aaron-siegel: Did you intend to check this file in? It presumably doesn't belong in the top-level repo dir.

Note that if you have notebooks you want to keep accessible for local use, you can put them in a directory called `docs/local` and git will ignore them (that path is in our `.gitignore`).

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,131 @@ def __dir__(self) -> list[str]:
...
+        """Return a string reporting the difference between two schema versions"""
+        assert new_md is not None
+        if old_md is None:
+            return 'Initial Version'
+        if old_md == new_md:
+            return ''
+        r = ''
+        added = {k: v['name'] for k, v in new_md.items() if k not in old_md}
+        if len(added) > 0:
+            r += 'Added: ' + ' '.join(added.values())
```
aaron-siegel: Just FYI string concatenation in Python is slow. It won't matter in this case, but patterns of the form
```
r = ''
for obj in collection:
    r += 'a string for {obj}'
```
are a notorious source of performance bugs (that loop has quadratic running time). Using join is generally preferred:
```
c = []
for obj in collection:
    c.append('a string for {obj}')
r = ''.join(c)
```

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,131 @@ def __dir__(self) -> list[str]:
...
+        changed = {
+            k: old_md[k]['name'] + ' to ' + v['name']
```
aaron-siegel: I'd find this easier to read with an f-specifier rather than `+`

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,131 @@ def __dir__(self) -> list[str]:
...
+            for k, v in new_md.items()
+            if k in old_md and old_md[k]['name'] != v['name']
+        }
+        if len(changed) > 0:
+            r += 'Changed: ' + ' '.join(changed.values())
+        deleted = {k: v['name'] for k, v in old_md.items() if k not in new_md}
+        if len(deleted) > 0:
```
aaron-siegel: What happens if `len(changed) > 0` and `len(deleted) > 0`? Or is that not possible?

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,131 @@ def __dir__(self) -> list[str]:
...
+            r += 'Deleted: ' + ' '.join(deleted.values())
+        return r
+
+    @classmethod
+    def create_md_change_dict(cls, src_rows: list[list[Any]]) -> dict[int, str]:
+        """Return a dictionary of schema changes by version given a list of version, column md"""
```
aaron-siegel: Wait, does `src_rows` not contain table rows? The word `rows` is so hard-wired in pxt to mean "table rows" that I find that quite confusing.

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,131 @@ def __dir__(self) -> list[str]:
 
     def _ipython_key_completions_(self) -> list[str]:
         return list(self._schema.keys())
+
+    @classmethod
+    def diff_md(cls, old_md: Optional[dict[str, Any]], new_md: Optional[dict[str, Any]]) -> str:
```
aaron-siegel: I'd argue these maybe don't belong in `table.py`, but we should have a separate module somewhere for history-related functionality, particularly if the logic is likely to grow more complex over time... `table.py` is already fairly fat.

If they stay in `table.py`, then we definitely want to precede them with underscores, since it's a user-facing class.

pixeltable/catalog/table.py
aaron-siegel: Ok, now I see how this is getting called; please please use something other than a `list[Any]` for the intermediate data structure. (A named tuple or similar struct is one option; if that seems too unwieldy, a typed tuple with appropriate code comments would still be way clearer than a list[Any].) That would make this code substantially easier to read.

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,131 @@ def __dir__(self) -> list[str]:
...
+                continue
+            assert ver > ver_old
+            nfld = row[1]
+            tf = cls.diff_md(ofld, nfld)
+            ofld = nfld
+            if tf != '':
+                r[ver] = tf
+        return r
+
+    def history(self, max_versions: int = 1_000_000_000) -> Optional[Any]:
```
aaron-siegel: Must the return type be Any?

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,131 @@ def __dir__(self) -> list[str]:
...
+        md_dict = self.create_md_change_dict(md_rows)
+
+        # Construct report rows
+        if max_versions is not None and len(src_rows) > max_versions:
+            assert len(src_rows) == max_versions + 1
+            over_count = 1
+        else:
+            over_count = 0
+
+        rpt_rows: list[list[Any]] = []
```
aaron-siegel: cf. comment above: pretty confusing to me to call these "rows" given how often that term is used in the codebase, always with the same meaning. "entries" or "history_entries" or some such would make it crystal clear.

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,131 @@ def __dir__(self) -> list[str]:
...
+        for row in src_rows[0 : len(src_rows) - over_count]:
+            version = row.TableVersion.version
+            schema_change = md_dict.get(version, '')
+            change_type = 'schema' if schema_change != '' else 'data'
+            rpt_row = [
```
aaron-siegel: Just as above, better to use some form of a tuple here than a list[Any].

aaron-siegel (changes requested):

history_demo.ipynb
jpeterson-pxt: Yes, this is an oversight.

pixeltable/catalog/table.py
jpeterson-pxt: Got it.

pixeltable/catalog/table.py
jpeterson-pxt: Fixed

pixeltable/catalog/table.py
jpeterson-pxt: Changes to make this NA

pixeltable/catalog/table.py
jpeterson-pxt: Good catch.

pixeltable/catalog/table.py
jpeterson-pxt: These are both intentionally list[list[Any]] and list[Any], as they are passed to a method intended to accept exactly this structure.

pixeltable/catalog/table.py
jpeterson-pxt: Happy to change to something else, but there are many places where we use "rows" for things which do not refer "pixeltable rows".
The idea here is that these will become "pixeltable rows" eventually. But I can call them something else. 
A similar strategy has always been followed in the .import classes, where "rows" can come from anything.

pixeltable/catalog/table.py
jpeterson-pxt: Done

pixeltable/catalog/table.py
jpeterson-pxt: Perhaps we can talk about code structuring as some point. 
Both table.py and dataframe.py are starting to have this problem.

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,138 @@ def __dir__(self) -> list[str]:
...
+
+            >>> tbl.history(max_versions=5)
+
+        Returns:
+            A list of rows of data.
+        """
+        if not isinstance(max_versions, int) or max_versions < 1:
+            raise excs.Error(f'Invalid value for max_versions: {max_versions}')
+
+        with Env.get().begin_xact():
```
mkornacker: Call Catalog.begin_xact() instead (see Env.begin_xact()).

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,138 @@ def __dir__(self) -> list[str]:
...
+                continue
+            assert ver > ver_old
+            nfld = row[1]
+            tf = cls._diff_md(ofld, nfld)
+            ofld = nfld
+            if tf != '':
+                r[ver] = tf
+        return r
+
+    def history(self, max_versions: int = 1_000_000_000) -> pixeltable.dataframe.DataFrameResultSet:
```
aaron-siegel: Maybe make `max_versions` an `Optional[int]` with default `None`.

pixeltable/catalog/table.py
```diff
@@ -1550,3 +1552,138 @@ def __dir__(self) -> list[str]:
...
+        """
+        if not isinstance(max_versions, int) or max_versions < 1:
+            raise excs.Error(f'Invalid value for max_versions: {max_versions}')
+
+        with Env.get().begin_xact():
+            # tbl_id = self._tbl_version_path.tbl_id
+            tbl_id = self._id
+            if tbl_id is None:
+                raise excs.Error('Cannot get history of a table that has been dropped')
+            q = (
```
mkornacker: Table should not be running queries against the stored metadata itself. It's only a user-facing interface class. All metadata retrieval is handled by Catalog.

See Catalog.get_table() for the expected pattern (retry_loop(), etc.).

pixeltable/dataframe.py
```diff
@@ -96,6 +108,29 @@ def __eq__(self, other: object) -> bool:
     def __hash__(self) -> int:
         return hash(self.to_pandas())
 
+    class ReportSpec(NamedTuple):
```
mkornacker: What's a report and why do we need this? What does 'production of a column in a ... report' mean?

pixeltable/catalog/table.py
mkornacker: Can we call this 'n'? eg, git log -n 10

Also, optional[int] = None is used everywhere else for 'not set'.

mkornacker (changes requested):

pixeltable/catalog/table.py
jpeterson-pxt: Done

pixeltable/catalog/table.py
jpeterson-pxt: done

pixeltable/catalog/table.py
jpeterson-pxt: Done

pixeltable/catalog/table.py
jpeterson-pxt: git syntax is also: 
    'git log -10'
the n is redundant.
Done

pixeltable/dataframe.py
jpeterson-pxt: history() is producing a report.

mkornacker: I still don't follow why we need this now. What is a report? A query result?

pixeltable/catalog/table.py
```diff
@@ -1561,3 +1562,124 @@ def __dir__(self) -> list[str]:
 
     def _ipython_key_completions_(self) -> list[str]:
         return list(self._get_schema().keys())
+
+    @classmethod
```
mkornacker: This has basically nothing to do with Table, which is an interface class. Does this have anything to do with the -Md dataclasses in schema.py? Is this a set of utilities for those classes (in which case it should be in something like metadata/utils.py)? If so, why does it use generic dicts?

pixeltable/catalog/catalog.py
```diff
@@ -1228,6 +1228,30 @@ def _load_tbl(self, tbl_id: UUID) -> Optional[Table]:
         self._tbls[tbl_id] = view
         return view
 
+    @_retry_loop(for_write=False)
+    def collect_tbl_history(
+        self, tbl_id: UUID, n: int
+    ) -> Sequence[sql.Row[tuple[schema.TableVersion, schema.TableSchemaVersion]]]:
+        """
+        Returns the history of all versions of the table with the given UUID.
+        The history is returned as a list of FullTableMd objects, ordered by version number.
```
mkornacker: Contradicts the signature.

pixeltable/catalog/catalog.py
```diff
@@ -1228,6 +1228,30 @@ def _load_tbl(self, tbl_id: UUID) -> Optional[Table]:
         self._tbls[tbl_id] = view
         return view
 
+    @_retry_loop(for_write=False)
+    def collect_tbl_history(
+        self, tbl_id: UUID, n: int
+    ) -> Sequence[sql.Row[tuple[schema.TableVersion, schema.TableSchemaVersion]]]:
```
mkornacker: We don't return sqlalchemy rows anywhere, why here?

jpeterson-pxt: We can move the items into a newly created list of lists if is important to not produced sqlalchemy rows for some reason. 
What do you suggest?

pixeltable/catalog/catalog.py
jpeterson-pxt: Done

pixeltable/catalog/table.py
jpeterson-pxt: Moved to metadata/utils. The typing is now more exacting, but it is using the metadata in dict form, and I couldn't find a typing declaration for the metadata dictionary form; is there one?

pixeltable/catalog/catalog.py
jpeterson-pxt: Done

pixeltable/catalog/table.py
jpeterson-pxt: Done

pixeltable/catalog/catalog.py
```diff
@@ -1228,6 +1228,42 @@ def _load_tbl(self, tbl_id: UUID) -> Optional[Table]:
         self._tbls[tbl_id] = view
         return view
 
+    @_retry_loop(for_write=False)
+    def collect_tbl_history(self, tbl_id: UUID, n: int) -> list[schema.FullTableMd]:
```
mkornacker: Make n optional, then
if n is not None:
    q = q.limit(n)

pixeltable/metadata/utils.py
```diff
@@ -0,0 +1,72 @@
...
+    @classmethod
+    def _create_md_change_dict(cls, md_list: list[tuple[int, dict[int, schema.SchemaColumn]]]) -> dict[int, str]:
+        """Return a dictionary of schema changes by version
+        Args:
+            md_list: a list of tuples, each containing a version number and a metadata dictionary.
+        """
+        r: dict[int, str] = {}
+        if md_list is None or len(md_list) == 0:
+            return r
+        md_list.sort()
```
mkornacker, line 58: What is this sorting on? The version?

pixeltable/metadata/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        return r
+
+    @classmethod
+    def _create_md_change_dict(cls, md_list: list[tuple[int, dict[int, schema.SchemaColumn]]]) -> dict[int, str]:
+        """Return a dictionary of schema changes by version
+        Args:
+            md_list: a list of tuples, each containing a version number and a metadata dictionary.
+        """
+        r: dict[int, str] = {}
+        if md_list is None or len(md_list) == 0:
```
mkornacker, line 54: None is not part of the type decl.

pixeltable/metadata/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        - If there are additions, changes, or deletions, returns a string summarizing the changes.
+        """
+        assert new_md is not None
+        if old_md is None:
+            return 'Initial Version'
+        if old_md == new_md:
+            return ''
+        added = {k: v.name for k, v in new_md.items() if k not in old_md}
+        changed = {
+            k: f'{old_md[k].name} to {v.name}' for k, v in new_md.items() if k in old_md and old_md[k].name != v.name
```
mkornacker: Quote all column names, like so:
{v.name!r}

pixeltable/metadata/utils.py
```diff
@@ -0,0 +1,72 @@
...
+        }
+        deleted = {k: v.name for k, v in old_md.items() if k not in new_md}
+        if len(added) == 0 and len(changed) == 0 and len(deleted) == 0:
+            return ''
+        # Format the result
+        t = []
+        if len(added) > 0:
+            t.append('Added: ' + ', '.join(added.values()))
+        if len(changed) > 0:
+            t.append('Changed: ' + ', '.join(changed.values()))
```
mkornacker: "Renamed:"

mkornacker (approved):

pixeltable/catalog/catalog.py
jpeterson-pxt: Done

pixeltable/metadata/utils.py
jpeterson-pxt: commented

pixeltable/metadata/utils.py
jpeterson-pxt: Done

pixeltable/metadata/utils.py
jpeterson-pxt: Done

pixeltable/metadata/utils.py
jpeterson-pxt: Done

PR #642, opened 2025-06-03
PXT-595: Raise error if attempting to access metadata from a future v…

jpeterson-pxt:
…ersion of pixeltable.

### DIFF
--- a/pixeltable/metadata/__init__.py
+++ b/pixeltable/metadata/__init__.py
@@ -55,6 +55,11 @@ def upgrade_md(engine: sql.engine.Engine) -> None:
         system_info = session.query(SystemInfo).one().md
         md_version = system_info['schema_version']
         assert isinstance(md_version, int)
+        if md_version > VERSION:
+            raise RuntimeError(
+                f'Metadata version {md_version} is newer than current version {VERSION}. '
+                f'Please update pixeltable to the latest version.'
+            )
         if md_version == VERSION:
             return
         while md_version < VERSION:
--- a/pixeltable/metadata/__init__.py
+++ b/pixeltable/metadata/__init__.py
@@ -8,6 +8,7 @@ from typing import Callable
 import sqlalchemy as sql
 from sqlalchemy import orm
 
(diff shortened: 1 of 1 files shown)


pixeltable/metadata/__init__.py
```diff
@@ -55,6 +55,11 @@ def upgrade_md(engine: sql.engine.Engine) -> None:
         system_info = session.query(SystemInfo).one().md
         md_version = system_info['schema_version']
         assert isinstance(md_version, int)
+        if md_version > VERSION:
+            raise RuntimeError(
+                f'Metadata version {md_version} is newer than current version {VERSION}. '
```
aaron-siegel: I'd argue this is not going to be a meaningful error message for users, since the MD version is an internal recordkeeping device that most users won't be (and shouldn't need to be) aware of, and to a user it's easily confused with "pixeltable version".

Something like this would be clearer:

```
"This Pixeltable database was created with a newer Pixeltable version"
" than the one currently installed (v{pxt.__version__}).\n"
"Please update to the latest Pixeltable version by running: pip install --upgrade pixeltable"
```
It'd still be useful to write the MD version information to a `_logger.error()` call.

pixeltable/metadata/__init__.py
```diff
@@ -55,6 +55,11 @@ def upgrade_md(engine: sql.engine.Engine) -> None:
         system_info = session.query(SystemInfo).one().md
         md_version = system_info['schema_version']
         assert isinstance(md_version, int)
+        if md_version > VERSION:
+            raise RuntimeError(
```
aaron-siegel: We use `pixeltable.excs.Error` for all exceptions that carry user-facing error messages.

aaron-siegel (changes requested):

pixeltable/metadata/__init__.py
jpeterson-pxt: Fixed.

aaron-siegel (approved):

PR #648, opened 2025-06-04
Update RowBuilder.create_table_raw to save PIL image with the jpeg extension

Yann-CV:
fix https://github.com/pixeltable/pixeltable/issues/647

This pull request updates the `create_table_row` method in `row_builder.py` to include a file extension parameter when preparing media paths. This change ensures that image files are stored with a `.jpeg` extension.

Key change:

* [`pixeltable/exprs/row_builder.py`](diffhunk://#diff-78230fdcfd85f491b5cf246dca1dc8ecc4a19a037c1baf3f2110de5cdfad26e0L451-R458): Modified the `MediaStore.prepare_media_path` call to include an `ext` parameter, specifying `.jpeg` as the file extension for images. This helps standardize image file storage.

### DIFF
--- a/pixeltable/exprs/row_builder.py
+++ b/pixeltable/exprs/row_builder.py
@@ -448,7 +448,7 @@ class RowBuilder:
             else:
                 if col.col_type.is_image_type() and data_row.file_urls[slot_idx] is None:
                     # we have yet to store this image
-                    filepath = str(MediaStore.prepare_media_path(col.tbl.id, col.id, col.tbl.version,))
+                    filepath = str(MediaStore.prepare_media_path(col.tbl.id, col.id, col.tbl.version))
                     data_row.flush_img(slot_idx, filepath)
                 val = data_row.get_stored_val(slot_idx, col.get_sa_col_type())
                 table_row[col.store_name()] = val
--- a/pixeltable/exprs/row_builder.py
+++ b/pixeltable/exprs/row_builder.py
@@ -448,7 +448,14 @@ class RowBuilder:
             else:
                 if col.col_type.is_image_type() and data_row.file_urls[slot_idx] is None:
                     # we have yet to store this image
-                    filepath = str(MediaStore.prepare_media_path(col.tbl.id, col.id, col.tbl.version))
+                    filepath = str(
+                        MediaStore.prepare_media_path(
+                            tbl_id=col.tbl.id, 
+                            col_id=col.id, 
+                            version=col.tbl.version, 
+                            ext=".jpeg",
+                        )
+                    )
                     data_row.flush_img(slot_idx, filepath)
                 val = data_row.get_stored_val(slot_idx, col.get_sa_col_type())
                 table_row[col.store_name()] = val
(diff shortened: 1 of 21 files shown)


pixeltable/exprs/row_builder.py
```diff
@@ -448,7 +448,14 @@ def create_table_row(self, data_row: DataRow, exc_col_ids: set[int]) -> tuple[di
             else:
                 if col.col_type.is_image_type() and data_row.file_urls[slot_idx] is None:
                     # we have yet to store this image
-                    filepath = str(MediaStore.prepare_media_path(col.tbl.id, col.id, col.tbl.version))
+                    filepath = str(
+                        MediaStore.prepare_media_path(
+                            tbl_id=col.tbl.id, 
+                            col_id=col.id, 
+                            version=col.tbl.version, 
+                            ext=".jpeg",
```
aaron-siegel: It's not necessarily a jpeg; it'll be stored as webp if the image contains transparency data. (See code in flush_img which is called by this function.)

One solution would be to append the extension in `flush_img` instead, once we know the actual filetype. Another could be to set the extension here via

`ext=('webp' if image.has_transparency_data else 'jpeg')`

but that seems less desirable, since it duplicates the logic for determining the image format.

aaron-siegel (changes requested):
Thanks for the PR! This looks like the right idea. See comment below.

Also, it'd be great to update `test_fiftyone.py` to capture this edge case. (I think that's as simple as inserting a few additional rows in the existing `test_export_images` that contain PIL images rather than URLs?)

pixeltable/exprs/row_builder.py
mkornacker: To add to Aaron's comments:

DataRow.flush_img() is the correct place for a fix.

Yann-CV: I made a new proposal.

Yann-CV:
@aaron-siegel for the test I added a test about the extension

tests/io/test_fiftyone.py
```diff
@@ -75,6 +81,9 @@ def test_export_images(self, reset_db: None) -> None:
         assert ds.count('detections') == 3
         assert ds.count('other') == 2
 
+        for sample in ds:
```
aaron-siegel: Better:
```
assert all(sample.filepath.endswith('.jpeg') for sample in ds)
```

aaron-siegel (approved):
lgtm with one small style suggestion in the unit test.

You'll want to run
```
make format
```
in your conda environment to reformat your changes so they'll pass our CI; then you can run
```
make check
```
locally to verify that all the static checks will pass in CI.

Yann-CV:
thanks @aaron-siegel ! by the way sorry for not running the make commands before pushing (poetry install was too long to let me do it yesterday evening ;-) )

tests/io/test_fiftyone.py
Yann-CV: done

PR #649, opened 2025-06-04
Consolidate all SQL generation related to sampling inside of SqlSampleNode

jpeterson-pxt:
Enhance Analyzer to handle stratification expressions from sample clauses.
Add logic to accommodate the stratify_exprs output from the Analyzer.
Generate all sampling SQL with then SqlSampleNode.


pixeltable/exec/sql_node.py
```diff
@@ -573,25 +572,44 @@ def key_sql_expr(cls, seed: sql.ColumnElement, sql_cols: Iterable[sql.ColumnElem
...
+        return self.key_sql_expr(sql.literal_column(str(self.sample_clause.seed)), rowid_cols)
 
     def _create_stmt(self) -> sql.Select:
-        if self.fraction_samples is not None:
-            return self._create_stmt_fraction(self.fraction_samples)
-        return self._create_stmt_n(self.n_samples, self.n_per_stratum)
+        from pixeltable.plan import SampleClause
+
+        if self.sample_clause.fraction is not None:
+            if self.stratify_exprs is None or len(self.stratify_exprs) == 0:
```
mkornacker: Is there any reason to allow both None and [] (and treat them the same)? You could further simplify the code by normalizing to [].

pixeltable/plan.py
```diff
@@ -180,6 +169,8 @@ class Analyzer:
     group_by_clause: Optional[list[exprs.Expr]]  # None for non-aggregate queries; [] for agg query w/o grouping
     grouping_exprs: list[exprs.Expr]  # [] for non-aggregate queries or agg query w/o grouping
     order_by_clause: OrderByClause
+    stratify_exprs: Optional[list[exprs.Expr]]
```
mkornacker: Do you need to distinguish None and []? (see comment for group_by_clause)

pixeltable/plan.py
```diff
@@ -984,6 +903,10 @@ def _create_query_plan(
             candidates.extend(
                 exprs.Expr.list_subexprs(analyzer.group_by_clause, filter=sql_elements.contains, traverse_matches=False)
             )
+        if analyzer.stratify_exprs is not None:
```
mkornacker: Also update the comment starting on line 883.

pixeltable/exec/sql_node.py

pixeltable/plan.py

pixeltable/plan.py

pixeltable/exec/sql_node.py
```diff
@@ -308,7 +308,7 @@ async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
                 _logger.debug(f'SqlLookupNode stmt:\n{stmt_str}')
             except Exception:
                 # log something if we can't log the compiled stmt
-                stmt_str = repr(stmt)
+                stmt_str = str(stmt)
```
aaron-siegel: This is unnecessary now (can just put `{stmt}` in the debug call)

pixeltable/exec/sql_node.py
```diff
@@ -573,25 +572,44 @@ def key_sql_expr(cls, seed: sql.ColumnElement, sql_cols: Iterable[sql.ColumnElem
...
+                # If non-stratified sampling, construct a where clause, order_by, and limit clauses
+                s_key = self._create_key_sql(self.input_cte)
+
+                # Construct a suitable where clause
+                fraction_sql = sql.cast(SampleClause.fraction_to_md5_hex(float(self.sample_clause.fraction)), sql.Text)
+                order_by = self._create_key_sql(self.input_cte)
+                return sql.select(*self.input_cte.c).where(s_key < fraction_sql).order_by(order_by)
+
+            return self._create_stmt_fraction(self.sample_clause.fraction)
+        else:
```
aaron-siegel, line 599: `assert self.sample_clause.n is not None` ?

pixeltable/exec/sql_node.py
```diff
@@ -573,25 +572,44 @@ def key_sql_expr(cls, seed: sql.ColumnElement, sql_cols: Iterable[sql.ColumnElem
...
+            if len(self.stratify_exprs) == 0:
+                # No stratification, just return n samples from the input CTE
+                order_by = self._create_key_sql(self.input_cte)
+                return sql.select(*self.input_cte.c).order_by(order_by).limit(self.sample_clause.n)
+
+            return self._create_stmt_n(self.sample_clause.n, self.sample_clause.n_per_stratum)
 
     def _create_stmt_n(self, n: Optional[int], n_per_stratum: Optional[int]) -> sql.Select:
```
aaron-siegel: Maybe call these _create_stmt_n_stratified and _create_stmt_fraction_stratified to be clear, since that's what they're doing (the logic for non-stratified sampling is elsewhere)

pixeltable/exec/sql_node.py

pixeltable/exec/sql_node.py
jpeterson-pxt: It has been indicated to me that because the rest of the code guarantees certain things, I should rely on those things here rather than adding additional checks.

pixeltable/exec/sql_node.py

jpeterson-pxt:
@aaron-siegel : Good idea. 
The tests are structured to do something like what you described. It is unclear if the tests are meaningful, as there is no obvious way to force PG to produce rows in different orders internally ahead of that sort. The indices we put in the DB make this a more interesting problem.

PR #652, opened 2025-06-06
Concurrent view interactions

mkornacker:
Covers all possible view/base table interactions:
- propagating base table inserts/updates/deletes
- creating and dropping views
- dropping base table columns
    
Also:
- handles concurrent table deletions
- fixes bug in how begin_xact() does retries

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -107,18 +107,18 @@ class Table(SchemaObject):
 
     def _get_metadata(self) -> dict[str, Any]:
         md = super()._get_metadata()
-        base = self._base_table()
+        base = self._get_base_table()
         md['base'] = base._path() if base is not None else None
-        md['schema'] = self._schema()
+        md['schema'] = self._get_schema()
         md['is_replica'] = self._tbl_version_path.is_replica()
-        md['version'] = self._version()
+        md['version'] = self._get_version()
         md['schema_version'] = self._tbl_version_path.schema_version()
-        md['comment'] = self._comment()
-        md['num_retained_versions'] = self._num_retained_versions()
-        md['media_validation'] = self._media_validation().name.lower()
+        md['comment'] = self._get_comment()
+        md['num_retained_versions'] = self._get_num_retained_versions()
+        md['media_validation'] = self._get_media_validation().name.lower()
         return md
 
-    def _version(self) -> int:
+    def _get_version(self) -> int:
         """Return the version of this table. Used by tests to ascertain version changes."""
         return self._tbl_version_path.version()
 
@@ -272,27 +272,27 @@ class Table(SchemaObject):
         cols = self._tbl_version_path.columns()
         return [c.name for c in cols]
 
-    def _schema(self) -> dict[str, ts.ColumnType]:
+    def _get_schema(self) -> dict[str, ts.ColumnType]:
         """Return the schema (column names and column types) of this table."""
         return {c.name: c.col_type for c in self._tbl_version_path.columns()}
 
-    def base_table(self) -> Optional['Table']:
+    def get_base_table(self) -> Optional['Table']:
         from pixeltable.catalog import Catalog
 
         with Catalog.get().begin_xact(for_write=False):
-            return self._base_table()
+            return self._get_base_table()
 
     @abc.abstractmethod
-    def _base_table(self) -> Optional['Table']:
+    def _get_base_table(self) -> Optional['Table']:
         """The base's Table instance. Requires a transaction context"""
 
-    def _base_tables(self) -> list['Table']:
+    def _get_base_tables(self) -> list['Table']:
         """The ancestor list of bases of this table, starting with its immediate base. Requires a transaction context"""
         bases: list[Table] = []
-        base = self._base_table()
+        base = self._get_base_table()
         while base is not None:
             bases.append(base)
-            base = base._base_table()
+            base = base._get_base_table()
         return bases
 
     @property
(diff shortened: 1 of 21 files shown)


mkornacker:
This is missing logic for Expr subclasses like ColumnRef to check for dropped columns in sql_expr(). I'll add that later today, but wanted to get this out now. The sql_expr() changes won't affect anything else.

pixeltable/catalog/catalog.py
```diff
@@ -71,25 +73,31 @@ def _retry_loop(*, for_write: bool) -> Callable[[Callable[..., T]], Callable[...
...
+                    with Catalog.get().begin_xact(for_write=for_write, convert_db_excs=False):
                         return op(*args, **kwargs)
                 except sql.exc.DBAPIError as e:
                     # TODO: what other exceptions should we be looking for?
-                    if isinstance(e.orig, psycopg.errors.SerializationFailure):
-                        if num_remaining_retries > 0:
-                            num_remaining_retries -= 1
-                            _logger.debug(f'Serialization failure, retrying ({num_remaining_retries} retries left)')
+                    if isinstance(e.orig, (psycopg.errors.SerializationFailure, psycopg.errors.LockNotAvailable)):
+                        if num_retries < _MAX_RETRIES or _MAX_RETRIES == 0:
```
aaron-siegel: This is confusing to me. I'd expect `_MAX_RETRIES = 0` to mean no retries, not infinite retries. I suggest either using `-1` for infinite retries or just use a very large integer. A third possibility would be for `_MAX_RETRIES` to be an `Optional[int]`.

pixeltable/catalog/catalog.py
```diff
@@ -71,25 +73,31 @@ def _retry_loop(*, for_write: bool) -> Callable[[Callable[..., T]], Callable[...
...
+                            num_retries += 1
+                            # print(f'Retrying ({num_retries}) after {type(e.orig)}')
```
aaron-siegel: remove? seems easy enough to swap between `_logger` and `_console_logger` when needed for debugging rather than litter the code with duplicative comments?

pixeltable/catalog/globals.py
```diff
@@ -21,6 +22,17 @@
 _PREDEF_SYMBOLS: Optional[set[str]] = None
 
 
+@dataclasses.dataclass
```
aaron-siegel: should be a named tuple or a frozen dataclass (in which case you don't have to implement `__hash__` manually)

pixeltable/catalog/table.py
```diff
@@ -100,42 +102,27 @@ def get_metadata(self) -> dict[str, Any]:
                 }
                 ```
         """
+        return super().get_metadata()
```
aaron-siegel, line 106: So the only purpose of defining this function here (in the subclass) is for the docstring?

pixeltable/catalog/table.py
```diff
@@ -162,15 +149,18 @@ def list_views(self, *, recursive: bool = True) -> list[str]:
...
+    def _get_views(self, *, recursive: bool = True, include_snapshots: bool = True) -> list['Table']:
         cat = catalog.Catalog.get()
         view_ids = cat.get_view_ids(self._id)
         views = [cat.get_table_by_id(id) for id in view_ids]
+        if not include_snapshots:
+            views = [t for t in views if not t._tbl_version_path.is_snapshot()]
         if recursive:
-            views.extend([t for view in views for t in view._get_views(recursive=True)])
+            views.extend(
+                [t for view in views for t in view._get_views(recursive=True, include_snapshots=include_snapshots)]
```
aaron-siegel: unnecessary list creation

pixeltable/catalog/table.py
```diff
@@ -276,53 +266,47 @@ def count(self) -> int:
...
     def _effective_base_versions(self) -> list[Optional[int]]:
         """The effective versions of the ancestor bases, starting with its immediate base."""
 
-    @property
     def _comment(self) -> str:
-        return self._tbl_version.get().comment
+        return self._tbl_version_path.comment()
 
-    @property
     def _num_retained_versions(self) -> int:
```
aaron-siegel: Per discussion last week, a clearer name for these would be `_get_comment()`, `_get_num_retained_versions()`, etc.

pixeltable/catalog/table.py
```diff
@@ -276,53 +266,47 @@ def count(self) -> int:
...
         return self._df().count()
 
-    @property
     def columns(self) -> list[str]:
         """Return the names of the columns in this table."""
         cols = self._tbl_version_path.columns()
         return [c.name for c in cols]
 
-    @property
     def _schema(self) -> dict[str, ts.ColumnType]:
```
aaron-siegel: Per discussion last week, it'd be clearer to call this `_get_schema()` if it's no longer a property.

I wonder if we should just call it `get_schema()` and make it a public method. It's come up a few times now that users have wanted to refer to the schema.

Or have a separate `get_schema()` that turns the `ColumnType` references back into `type`s? (Although, there are still a few other places in the code where `ColumnType` references leak to the user, if I recall correctly)

aaron-siegel (approved):
Mostly minor comments; the overall architecture looks non-controversial.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/globals.py
mkornacker: Done

pixeltable/catalog/table.py
mkornacker: Left comment.

pixeltable/catalog/table.py

pixeltable/catalog/table.py
mkornacker: Let's leave get_schema() for another time (but I think it should not return ColumnType, and we should get rid of that in the public API).

pixeltable/catalog/table.py

PR #653, opened 2025-06-06
Changes to pixeltable shared client for R2 support.

amithadke:
- Add bucket and is_public parameters for create_replica.
 - Add delete_replica (not visible to globals.py yet).


pixeltable/share/publish.py
```diff
@@ -149,3 +161,37 @@ def _download_bundle_from_s3(parsed_location: urllib.parse.ParseResult, bundle_f
...
     s3_client.download_file(Bucket=bucket, Key=remote_path, Filename=str(bundle_path), Callback=progress_bar.update)
     return bundle_path
+
+
+def _upload_bundle_to_r2(bundle: Path, parsed_location: urllib.parse.ParseResult) -> None:
+    try:
+        with open(bundle, 'rb') as f:
+            file_size = os.path.getsize(bundle)
+            headers = {'Content-Length': str(file_size)}
+            response = requests.put(parsed_location.geturl(), data=f, headers=headers)
```
amithadke: TODO : better way to put/upload a large file in chunks

pixeltable/share/publish.py
```diff
@@ -16,38 +17,46 @@
 # These URLs are abstracted out for now, but will be replaced with actual (hard-coded) URLs once the
 # pixeltable.com URLs are available.
 
-PIXELTABLE_API_URL = 'https://internal-api.pixeltable.com'
+PIXELTABLE_API_URL = 'https://dev-internal-api.pixeltable.com'
```
amithadke: internal-api.pixeltable.com doesn't work today will move it to back to internal-api.pixeltable.com once new changes are deployed to it

pixeltable/share/publish.py
```diff
@@ -149,3 +163,54 @@ def _download_bundle_from_s3(parsed_location: urllib.parse.ParseResult, bundle_f
...
+            if attempt == max_retries:
+                raise excs.Error(f'Download failed after {max_retries} attempts: {e}') from e
+            time.sleep(1)  # wait before retrying
+            continue
+        except Exception as e:
+            raise excs.Error(f'Failed to download : {e}') from e
+    raise excs.Error(f'Download failed after {max_retries} attempts')
+
+
+def delete_replica(dest_path: str) -> None:
```
amithadke, line 268: this should eventually be replaced by drop_table with remote table path
I use it for now to delete replica

pixeltable/env.py
```diff
@@ -816,6 +818,91 @@ def __init_spacy(self) -> None:
...
+        Internal cleanup method that properly closes all resources and resets state.
+        This is called before destroying the singleton instance.
+        """
+        # Close database connections and sessions
+        if self._current_session is not None:
+            try:
+                self._current_session.close()
+            except Exception as e:
+                _logger.error(f'Error closing session: {e}')
+            self._current_session = None
```
aaron-siegel: What's the purpose in clearing the member variables like this? Won't they get garbage collected anyway once the singleton instance is discarded?

pixeltable/env.py
```diff
@@ -816,6 +818,91 @@ def __init_spacy(self) -> None:
...
+                _logger.error(f'Error closing connection: {e}')
+            self._current_conn = None
+
+        # Stop HTTP server
+        if self._httpd is not None:
+            try:
+                self._httpd.shutdown()
+                self._httpd.server_close()
+            except Exception as e:
+                _logger.error(f'Error stopping HTTP server: {e}')
```
aaron-siegel: Are these all really errors or should they be warnings?

pixeltable/env.py
```diff
@@ -816,6 +818,91 @@ def __init_spacy(self) -> None:
         except Exception as exc:
             raise excs.Error(f'Failed to load spaCy model: {spacy_model}') from exc
 
+    def _cleanup(self) -> None:
+        """
+        Internal cleanup method that properly closes all resources and resets state.
+        This is called before destroying the singleton instance.
+        """
+        # Close database connections and sessions
+        if self._current_session is not None:
```
aaron-siegel: Have you checked that all these conditions actually have a chance of being met? I'd expect that `_cleanup()` would only ever be called in a context where `_current_session` and `current_conn` must be `None`. Maybe these should be assertions instead? (I could be wrong about this, just not seeing it right now, but I think we should at least be able to articulate the conditions under which they might be non-`None`)

pixeltable/env.py
```diff
@@ -816,6 +818,91 @@ def __init_spacy(self) -> None:
...
+                                AND pid <> pg_backend_pid()
+                            """
+                            conn.execute(sql.text(stmt))
+                            _logger.info(f"Terminated all connections to database '{self._db_name}'")
+                    except Exception as e:
+                        _logger.warning(f'Error terminating database connections: {e}')
+                    finally:
+                        temp_engine.dispose()
+            except Exception as e:
+                _logger.error(f'Error stopping database server: {e}')
```
aaron-siegel: Wait, which line of code actually stops the server?

pixeltable/env.py
```diff
@@ -816,6 +818,91 @@ def __init_spacy(self) -> None:
...
+                self._httpd.server_close()
+            except Exception as e:
+                _logger.error(f'Error stopping HTTP server: {e}')
+            self._httpd = None
+
+        # Stop database server (for local environment)
+        if self._db_server is not None and self._dbms is not None:
+            try:
+                # First terminate all connections to the database
+                if self._db_name is not None:
```
aaron-siegel: Is it possible for `self._db_name` to be `None` even though `self._db_server is not None and self._dbms is not None`? If not, make this into an assertion.

pixeltable/env.py
```diff
@@ -816,6 +818,91 @@ def __init_spacy(self) -> None:
...
+        for handler in self._logger.handlers[:]:
+            try:
+                handler.close()
+                self._logger.removeHandler(handler)
+            except Exception as e:
+                _logger.error(f'Error removing handler: {e}')
+
+        # Clear temporary directory
+        try:
+            self.clear_tmp_dir()
```
aaron-siegel: Are we certain this is the right thing to do? What if there are other concurrent processes running that are using the same tmp dir? What if there are media files that need to be replayed via a multi phase commit journal?

pixeltable/globals.py
```diff
@@ -396,7 +396,9 @@ def create_snapshot(
     )
 
 
-def create_replica(destination: str, source: str | catalog.Table) -> Optional[catalog.Table]:
+def create_replica(
+    destination: str, source: Union[str, catalog.Table], bucket_name: str | None = None, is_public: bool | None = False
```
aaron-siegel: Use Python 3.10 unions (`str | catalog.Table`)

pixeltable/globals.py
```diff
@@ -405,6 +407,10 @@ def create_replica(destination: str, source: str | catalog.Table) -> Optional[ca
         destination: Path where the replica will be created. Can be either a local path such as `'my_dir.my_table'`, or
             a remote URI such as `'pxt://username/mydir.my_table'`.
         source: Path to the source table, or (if the source table is a local table) a handle to the source table.
+        bucket_name: The name of the pixeltable cloud-registered bucket to use to store replica's data.
+            If no `bucket_name` is provided, the default Pixeltable storage bucket will be used.
+        is_public: If True, replica will be shared publicly. If false, replica will remain private.
```
aaron-siegel: Maybe instead of `is_public: bool`, make this `access: Literal['public', 'internal']` or some such. Would be more future-proof as we plan to provide more nuanced access policies in the future.

pixeltable/share/packager.py
```diff
@@ -92,10 +92,10 @@ def package(self) -> Path:
         self.bundle_path = self.__build_tarball()
 
         _logger.info('Extracting preview data.')
-        self.md['count'] = self.table.count()
+        self.md['rows'] = self.table.count()
```
aaron-siegel: `'row_count'` would be clearer

pixeltable/share/publish.py
```diff
@@ -149,3 +163,54 @@ def _download_bundle_from_s3(parsed_location: urllib.parse.ParseResult, bundle_f
     )
     s3_client.download_file(Bucket=bucket, Key=remote_path, Filename=str(bundle_path), Callback=progress_bar.update)
     return bundle_path
+
+
+def _upload_bundle_with_presigned_url(bundle: Path, presigned_url: str, max_retries: int = 3) -> None:
```
aaron-siegel: Use a progress bar so that there's a consistent experience w/ S3 uploads. Best if this can be done in a way that limits code duplication.

pixeltable/share/publish.py
```diff
@@ -149,3 +163,54 @@ def _download_bundle_from_s3(parsed_location: urllib.parse.ParseResult, bundle_f
...
+    file_size = bundle.stat().st_size
+    headers = {'Content-Length': str(file_size), 'Content-Type': 'application/octet-stream'}
+    for attempt in range(1, max_retries + 1):
```
aaron-siegel: Does `requests` not provide a retry mechanism? I thought it did. If so, better to use that than roll our own retry system (which is likely to miss edge cases / require downstream maintenance)

pixeltable/share/publish.py
aaron-siegel: Add a code comment with a TODO to make this clear

aaron-siegel (changes requested):

pixeltable/env.py
amithadke: we can ignore these actually. This was added too defensively when we were switching from control plane db to tenant db  using pixeltable Env. Now control plane has a separate env and its own engine/connection/session, we can lower severity down.

pixeltable/env.py
amithadke: for single threaded yes, _cleanup() gets called only when worker is moving to another db and _current_session and current_conn will be None
We can move them to assertion but I'm thinking of not raising exceptions for assertions in cloud env.

pixeltable/env.py
amithadke: it doesn't stop server/process comment is misleading just terminates any connections (doesnt really apply for cloud env)

pixeltable/env.py
amithadke: self._dbms should always be set if env has initialized. 
_db_name and _db_server will always be set together.
I can add assertions for them

In test setup, we just call Env.init_env() even if its not yet initialized. Goal here was to not worry if its initialized correctly. There were cases due to bugs/issues where env was not initialized correctly for a db but then for another db it was successfully initialized, so in this case we want force clearing out any previous attempt, rather than restarting lamdba function.

pixeltable/share/publish.py
amithadke: i think there is in urllib3.util.retry.Retry with requests.adapters.HTTPAdapter
let me try it out

pixeltable/env.py
amithadke: we were using same Env object before(in Env.clear())
After moving to Env.init_env we don't need it as it creates new object will remove this now

pixeltable/env.py
amithadke: we do it in _setup_up anyway. I'll remove it from cleanup (for data sharing it shouldn't be an issue)
We have to think about sharing directories across multiple worker processes, may be shared directories shouldn't be in tmp dir?

pixeltable/globals.py
```diff
@@ -397,7 +397,10 @@ def create_snapshot(
 
 
 def create_replica(
-    destination: str, source: Union[str, catalog.Table], bucket_name: str | None = None, is_public: bool | None = False
+    destination: str,
+    source: (str | catalog.Table),
```
aaron-siegel: Unnecessary parens

pixeltable/share/publish.py
```diff
@@ -165,45 +168,87 @@ def _download_bundle_from_s3(parsed_location: urllib.parse.ParseResult, bundle_f
...
+        response = session.get(
+            url, headers=headers, stream=True, timeout=(60, 300)
+        )  # 60 seconds to connect and 300 seconds for server response
+        response.raise_for_status()
+
+        total_size = int(response.headers.get('content-length', 0))
+
+        with (
+            open(output_path, 'wb') as f,
+            tqdm(total=total_size, unit='B', unit_scale=True, desc='Downloading') as pbar,
```
aaron-siegel: Use the same tqdm configuration as for S3 transfers to ensure a consistent experience.

pixeltable/share/publish.py
```diff
@@ -165,45 +168,87 @@ def _download_bundle_from_s3(parsed_location: urllib.parse.ParseResult, bundle_f
...
-    for attempt in range(1, max_retries + 1):
-        try:
-            response = requests.get(presigned_url, stream=True, timeout=(60, 1800))
+
+    session = _create_retry_session(max_retries=max_retries)
+
+    try:
+        with (
+            open(file_path, 'rb') as f,
+            tqdm.wrapattr(f, 'read', total=file_size, desc='Uploading') as file_with_progress,
```
aaron-siegel: Use the same tqdm configuration as for S3 transfers to ensure a consistent experience.

pixeltable/share/publish.py
```diff
@@ -165,45 +168,87 @@ def _download_bundle_from_s3(parsed_location: urllib.parse.ParseResult, bundle_f
...
-            # Stream download to file
-            with open(bundle_path, 'wb') as f:
-                for chunk in response.iter_content(chunk_size=8192):
+            return response
+    finally:
+        session.close()
+
+
+def _download_from_presigned_url(
+    url: str, output_path: Path, headers: Optional[Dict[str, str]] = None, max_retries: int = 3
```
aaron-siegel: Capitalized `Dict` is obsolete in Python 3.9+; use lowercase `dict` instead

pixeltable/env.py
```diff
@@ -816,6 +820,63 @@ def __init_spacy(self) -> None:
         except Exception as exc:
             raise excs.Error(f'Failed to load spaCy model: {spacy_model}') from exc
 
+    def _cleanup(self) -> None:
```
aaron-siegel: Nitpick: `_clean_up()` (to align with `_set_up()`; "cleanup" is a noun, "clean up" is a verb)

pixeltable/globals.py
```diff
@@ -405,6 +410,11 @@ def create_replica(destination: str, source: str | catalog.Table) -> Optional[ca
         destination: Path where the replica will be created. Can be either a local path such as `'my_dir.my_table'`, or
             a remote URI such as `'pxt://username/mydir.my_table'`.
         source: Path to the source table, or (if the source table is a local table) a handle to the source table.
+        bucket_name: The name of the pixeltable cloud-registered bucket to use to store replica's data.
+            If no `bucket_name` is provided, the default Pixeltable storage bucket will be used.
+        access: Access control for the replica.
+            - `'public'`: Anyone can access this replica.
```
aaron-siegel, line 417: Please preview this with `mkdocs serve`. I think you need a blank line before the list in the docstring; check the output to be sure it looks like what you expect

tests/test_env.py
```diff
@@ -0,0 +1,152 @@
...
+    Catalog.clear()
+    # Reload configs
+    Config.init(config_overrides={}, reinit=True)
+    Env._init_env(reinit_db=reinit)
+
+
+@pytest.fixture(scope='function')
+def test_env_setup(init_env: None) -> Iterator[None]:
+    """Restore PIXELTABLE_DB set by conftest"""
+    test_db = os.environ['PIXELTABLE_DB']
```
aaron-siegel: It'd be safer to use `config_overrides` for this rather than setting and carefully resetting the environment variable - e.g., just take a `db: str` parameter in `_reset_env()`, and pass it along to `config_overrides`.

Right now `PIXELTABLE_DB` is only supported from the environment but I believe it'd be a 1-line change to route it through Config, so that `config_overrides` can be used.

aaron-siegel (approved):

pixeltable/globals.py
```diff
@@ -405,6 +410,12 @@ def create_replica(destination: str, source: str | catalog.Table) -> Optional[ca
         destination: Path where the replica will be created. Can be either a local path such as `'my_dir.my_table'`, or
             a remote URI such as `'pxt://username/mydir.my_table'`.
         source: Path to the source table, or (if the source table is a local table) a handle to the source table.
+        bucket_name: The name of the pixeltable cloud-registered bucket to use to store replica's data.
+            If no `bucket_name` is provided, the default Pixeltable storage bucket will be used.
+        access: Access control for the replica.
+
+            - `'public'`: Anyone can access this replica.
+            - `'private'`: Only the owner can access (default)
```
mkornacker: remove "(default)", that gets auto-generated

PR #654, opened 2025-06-06
First part of additional status collection for table.history reporting.

jpeterson-pxt:
Bottleneck all Table, TableVersion, and TableSchemaVersion writes through one method to ensure consistency.
Use UpdateStatus to return stats from more places.
Add functionality to UpdateStatus to track more things. 

Work to do: 
Collect the additional stats.
Add stats to TableVersion records
Add stats to table.history() reports.


pixeltable/catalog/globals.py
```diff
@@ -21,6 +19,44 @@
 _PREDEF_SYMBOLS: Optional[set[str]] = None
 
 
+@dataclasses.dataclass()
```
aaron-siegel: I have a fairly strong preference for using immutable objects for things like this (e.g., make this a frozen dataclass, and implement `__add__` rather than an "accumulate" method). It makes the code that uses it easier to reason about, and it's unlikely to make any meaningful performance difference.

The codebase is inconsistent today (`UpdateStatus` uses an accumulator, whereas `SyncStatus` is immutable) so if this is controversial, let's tee it up for the next standup/eng meeting.

pixeltable/catalog/globals.py
```diff
@@ -21,6 +19,44 @@
...
+        """
+        return (
+            self.ins_rows == 0
+            and self.del_rows == 0
+            and self.upd_rows == 0
+            and self.exc_rows == 0
+            and self.num_excs == 0
+        )
+
+    def accumulate(self, other: 'RowCountStats', as_update: bool = False) -> 'RowCountStats':
```
aaron-siegel: Do we need an `as_update` parameter? In my opinion it introduces confusion... it'd be clearer if that logic were moved out of `RowCountStats` into the client.

jpeterson-pxt: I think it is clearer this way; I find that it reduces confusion.

pixeltable/catalog/globals.py
jpeterson-pxt: I do not find the concept of "an object that can be accumulated into" at all confusing. 
I removed the implementation of __add__ as the accumulation operation is more complex than simple addition, especially now that it has two parameters.
It was unclear how to make __add_ have two parameters.

jpeterson-pxt: Done

pixeltable/catalog/globals.py
jpeterson-pxt: Parameter(s) removed.

PR #655, opened 2025-06-06
Fix bug in handling "nullary" JsonMapper expressions

aaron-siegel:
Fixes a bug involving the evaluator's handling of JsonMappers whose map expressions don't actually contain any references to `RELATIVE_PATH_ROOT`. In particular, this fixes the evaluation of nullary tool calls.

### DIFF
--- a/pixeltable/exec/expr_eval/evaluators.py
+++ b/pixeltable/exec/expr_eval/evaluators.py
@@ -316,7 +316,6 @@ class JsonMapperDispatcher(Evaluator):
                 )
                 for _ in src
             ]
-            print(src)
             for nested_row, anchor_val in zip(nested_rows, src):
                 assert self.scope_anchor.slot_idx is not None
                 nested_row[self.scope_anchor.slot_idx] = anchor_val
--- a/pixeltable/exec/expr_eval/evaluators.py
+++ b/pixeltable/exec/expr_eval/evaluators.py
@@ -316,7 +316,9 @@ class JsonMapperDispatcher(Evaluator):
                 )
                 for _ in src
             ]
+            print(src)
             for nested_row, anchor_val in zip(nested_rows, src):
+                assert self.scope_anchor.slot_idx is not None
                 nested_row[self.scope_anchor.slot_idx] = anchor_val
                 for slot_idx_, nested_slot_idx in self.external_slot_map.items():
                     nested_row[nested_slot_idx] = row[slot_idx_]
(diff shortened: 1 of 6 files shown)


pixeltable/exec/expr_eval/evaluators.py
```diff
@@ -316,7 +316,9 @@ def schedule(self, rows: list[exprs.DataRow], slot_idx: int) -> None:
                 )
                 for _ in src
             ]
+            print(src)
```
mkornacker: remove

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -285,7 +285,7 @@ async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
                     if isinstance(self.error, excs.ExprEvalError):
                         raise self.error from self.error.exc
                     else:
-                        raise self.error
+                        raise excs.Error('An error occurred during expression evaluation.') from self.error
```
mkornacker: This breaks the concurrency PR that's in review (it won't break the tests, but it breaks the behavior).

pixeltable/functions/openai.py
```diff
@@ -822,7 +822,7 @@ def _openai_response_to_pxt_tool_calls(response: dict) -> Optional[dict]:
         tool_name = tool_call['function']['name']
         if tool_name not in pxt_tool_calls:
             pxt_tool_calls[tool_name] = []
-        pxt_tool_calls[tool_name].append({'args': json.loads(tool_call['function']['arguments'])})
+        pxt_tool_calls[tool_name].append({'args': json.loads(tool_call['function']['arguments']) or {'dummy': 5}})
```
mkornacker: What's going on here?

pixeltable/exec/expr_eval/expr_eval_node.py
mkornacker: Take a look at the expr_eval_node.py changes in that PR. We need to be able to see DBAPIErrors.

pixeltable/exec/expr_eval/evaluators.py
aaron-siegel: done

pixeltable/exec/expr_eval/expr_eval_node.py
aaron-siegel: I backed it out. But why would this prevent us from seeing DBAPIErrors?

pixeltable/functions/openai.py
aaron-siegel: Was the residue of an earlier experiment, fixed (sorry)

mkornacker (approved):

PR #658, opened 2025-06-10
~40% improvement in insert performance

aaron-siegel:
Several optimizations on the insertion path:
1. Batch database writes;
2. Use more efficient data structures to marshal data for DB writes;
3. Minor optimization to DataRow exception handling.

The net improvement is around 40% reduction in insert time on a test scenario (insert 100,000 rows into a table with a string column and an int column), from ~9.0 seconds to ~5.25 seconds.

### DIFF
--- a/pixeltable/exprs/data_row.py
+++ b/pixeltable/exprs/data_row.py
@@ -42,6 +42,10 @@ class DataRow:
     has_val: np.ndarray  # of bool
     excs: np.ndarray  # of object
 
+    # If `may_have_exc` is False, then we guarantee that no slot has an exception set. This is used to optimize
+    # exception handling under normal operation.
+    _may_have_exc: bool
+
     # expr evaluation state; indexed by slot idx
     missing_slots: np.ndarray  # of bool; number of missing dependencies
     missing_dependents: np.ndarray  # of int16; number of missing dependents
@@ -90,6 +94,7 @@ class DataRow:
         self.vals = np.full(num_slots, None, dtype=object)
         self.has_val = np.zeros(num_slots, dtype=bool)
         self.excs = np.full(num_slots, None, dtype=object)
+        self._may_have_exc = False
         self.missing_slots = np.zeros(num_slots, dtype=bool)
         self.missing_dependents = np.zeros(num_slots, dtype=np.int16)
         self.is_scheduled = np.zeros(num_slots, dtype=bool)
@@ -136,9 +141,13 @@ class DataRow:
         """
         Returns True if an exception has been set for the given slot index, or for any slot index if slot_idx is None
         """
+        if not self._may_have_exc:
+            return False
+
         if slot_idx is not None:
             return self.excs[slot_idx] is not None
-        return (self.excs != None).any()
+        else:
+            return (self.excs != None).any()
 
     def get_exc(self, slot_idx: int) -> Optional[Exception]:
         exc = self.excs[slot_idx]
@@ -154,6 +163,7 @@ class DataRow:
     def set_exc(self, slot_idx: int, exc: Exception) -> None:
         assert self.excs[slot_idx] is None
         self.excs[slot_idx] = exc
+        self._may_have_exc = True
 
         # an exception means the value is None
         self.has_val[slot_idx] = True
--- a/pixeltable/exprs/row_builder.py
+++ b/pixeltable/exprs/row_builder.py
@@ -427,33 +427,31 @@ class RowBuilder:
                         expr, f'expression {expr}', data_row.get_exc(expr.slot_idx), exc_tb, input_vals, 0
                     ) from exc
 
-    def create_table_row(self, data_row: DataRow, exc_col_ids: set[int]) -> tuple[dict[str, Any], int]:
+    def create_table_row(self, data_row: DataRow, exc_col_ids: set[int]) -> tuple[list[Any], int]:
         """Create a table row from the slots that have an output column assigned
 
         Return tuple[dict that represents a stored row (can be passed to sql.insert()), # of exceptions]
             This excludes system columns.
         """
         num_excs = 0
-        table_row: dict[str, Any] = {}
+        table_row: list[Any] = []
         for info in self.table_columns:
             col, slot_idx = info.col, info.slot_idx
             if data_row.has_exc(slot_idx):
                 # exceptions get stored in the errortype/-msg columns
+                assert col._records_errors
                 exc = data_row.get_exc(slot_idx)
                 num_excs += 1
                 exc_col_ids.add(col.id)
-                table_row[col.store_name()] = None
-                table_row[col.errortype_store_name()] = type(exc).__name__
-                table_row[col.errormsg_store_name()] = str(exc)
+                table_row.extend((None, type(exc).__name__, str(exc)))
             else:
                 if col.col_type.is_image_type() and data_row.file_urls[slot_idx] is None:
                     # we have yet to store this image
                     filepath = str(MediaStore.prepare_media_path(col.tbl.id, col.id, col.tbl.version))
                     data_row.flush_img(slot_idx, filepath)
                 val = data_row.get_stored_val(slot_idx, col.get_sa_col_type())
-                table_row[col.store_name()] = val
-                # we unfortunately need to set these, even if there are no errors
-                table_row[col.errortype_store_name()] = None
-                table_row[col.errormsg_store_name()] = None
+                table_row.append(val)
+                if col._records_errors:
+                    table_row.extend((None, None))
 
         return table_row, num_excs
--- a/pixeltable/exprs/row_builder.py
+++ b/pixeltable/exprs/row_builder.py
@@ -438,12 +438,13 @@ class RowBuilder:
         for info in self.table_columns:
             col, slot_idx = info.col, info.slot_idx
             if data_row.has_exc(slot_idx):
-                # exceptions get stored in the errortype/-msg columns
-                assert col._records_errors
                 exc = data_row.get_exc(slot_idx)
                 num_excs += 1
                 exc_col_ids.add(col.id)
-                table_row.extend((None, type(exc).__name__, str(exc)))
+                table_row.append(None)
+                if col.records_errors:
+                    # exceptions get stored in the errortype/-msg columns
+                    table_row.extend((type(exc).__name__, str(exc)))
             else:
                 if col.col_type.is_image_type() and data_row.file_urls[slot_idx] is None:
                     # we have yet to store this image
@@ -451,7 +452,7 @@ class RowBuilder:
                     data_row.flush_img(slot_idx, filepath)
                 val = data_row.get_stored_val(slot_idx, col.get_sa_col_type())
                 table_row.append(val)
-                if col._records_errors:
+                if col.records_errors:
                     table_row.extend((None, None))
 
         return table_row, num_excs
--- a/pixeltable/store.py
+++ b/pixeltable/store.py
@@ -125,13 +125,14 @@ class StoreBase:
 
     def _move_tmp_media_file(self, file_url: Optional[str], col: catalog.Column, v_min: int) -> str:
         """Move tmp media file with given url to Env.media_dir and return new url, or given url if not a tmp_dir file"""
-        pxt_tmp_dir = str(Env.get().tmp_dir)
         if file_url is None:
             return None
+        assert isinstance(file_url, str)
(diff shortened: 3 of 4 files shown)


pixeltable/exprs/row_builder.py
```diff
@@ -427,33 +427,32 @@ def eval(
                         expr, f'expression {expr}', data_row.get_exc(expr.slot_idx), exc_tb, input_vals, 0
                     ) from exc
 
-    def create_table_row(self, data_row: DataRow, exc_col_ids: set[int]) -> tuple[dict[str, Any], int]:
+    def create_table_row(self, data_row: DataRow, exc_col_ids: set[int]) -> tuple[list[Any], int]:
         """Create a table row from the slots that have an output column assigned
 
         Return tuple[dict that represents a stored row (can be passed to sql.insert()), # of exceptions]
```
mkornacker: Update comment.

pixeltable/exprs/row_builder.py
```diff
@@ -427,33 +427,32 @@ def eval(
...
             This excludes system columns.
         """
         num_excs = 0
-        table_row: dict[str, Any] = {}
+        table_row: list[Any] = []
         for info in self.table_columns:
```
mkornacker, line 447: It looks like this assumes something specific about the order of columns in table_columns.

pixeltable/exprs/row_builder.py
```diff
@@ -427,33 +427,32 @@ def eval(
...
             col, slot_idx = info.col, info.slot_idx
             if data_row.has_exc(slot_idx):
-                # exceptions get stored in the errortype/-msg columns
                 exc = data_row.get_exc(slot_idx)
                 num_excs += 1
                 exc_col_ids.add(col.id)
-                table_row[col.store_name()] = None
-                table_row[col.errortype_store_name()] = type(exc).__name__
-                table_row[col.errormsg_store_name()] = str(exc)
+                table_row.append(None)
```
mkornacker, line 453: Given that we know the number of values in advance, can we presize the list?

pixeltable/store.py
```diff
@@ -341,50 +339,83 @@ def insert_rows(
         cols_with_excs: set[int] = set()
```
mkornacker, line 330: What about load_column()? Fine to leave a todo, but I think add_computed_column() perf is as important as insert().

pixeltable/store.py
```diff
@@ -341,50 +339,83 @@ def insert_rows(
...
         progress_bar: Optional[tqdm] = None  # create this only after we started executing
         row_builder = exec_plan.row_builder
-        media_cols = [info.col for info in row_builder.table_columns if info.col.col_type.is_media_type()]
-        conn = Env.get().conn
+
+        store_col_names: list[str] = [pk_col.name for pk_col in self._pk_cols]
```
mkornacker: Move this logic into RowBuilder.init()

pixeltable/store.py
```diff
@@ -341,50 +339,83 @@ def insert_rows(
...
+        assert len(table_rows) > 0
+        conn = Env.get().conn
+        conn.execute(sql.insert(self.sa_tbl), [dict(zip(store_col_names, table_row)) for table_row in table_rows])
+
+        # TODO: Inserting directly via psycopg delivers a small performance benefit, but is somewhat fraught due to
+        #     differences in the data representation that SQLAlchemy/psycopg expect. The below code will do the
+        #     insertion in psycopg and can be used if/when we decide to pursue that optimization.
+        # col_names_str = ", ".join(store_col_names)
+        # placeholders_str = ", ".join('%s' for _ in store_col_names)
+        # stmt_text = f'INSERT INTO {self.sa_tbl.name} ({col_names_str}) VALUES ({placeholders_str})'
```
mkornacker, line 404: If you do this via sqlalchemy, how does change the performance? Does it still have data representation problems?

pixeltable/store.py
```diff
@@ -341,50 +339,83 @@ def insert_rows(
...
+
             return num_rows, num_excs, cols_with_excs
+
         finally:
             exec_plan.close()
 
+    def sql_insert(self, store_col_names: list[str], table_rows: list[tuple[Any]]) -> None:
+        assert len(table_rows) > 0
+        conn = Env.get().conn
+        conn.execute(sql.insert(self.sa_tbl), [dict(zip(store_col_names, table_row)) for table_row in table_rows])
```
mkornacker, line 397: Isn't this effectively running the same insert logic as before? If you change the batching behavior to what it was before, is there still a perf gain?

pixeltable/exprs/data_row.py
```diff
@@ -154,6 +162,7 @@ def get_first_exc(self) -> Optional[Exception]:
     def set_exc(self, slot_idx: int, exc: Exception) -> None:
         assert self.excs[slot_idx] is None
         self.excs[slot_idx] = exc
+        self._may_have_exc = True
```
mkornacker, line 165: Isn't this really a 'has_any_exc' as opposed to 'may_have'?

mkornacker:
Some thoughts/questions:
- The previous logic for constructing the table row dicts was a bit easier to follow, because there were fewer dependencies (eg, on the order of columns) between the various pieces of code involved.
- At what point does the perf increase from increasing batch size level off? Would 64K or 128K row batches make sense if we're trying to insert 1M rows (which I think is a more realistic target than 100K rows).

pixeltable/exprs/data_row.py
aaron-siegel: No. Once it's `True`, it will never be set to `False` again unless `clear()` is called. (This is by design; maintaining a `has_any_exc` that stays valid on every update negates any performance gain - I tried it. And the goal is really to maximize the performance gain in the frequent case where a row produces no exceptions.)

pixeltable/exprs/row_builder.py
aaron-siegel: It doesn't; it only assumes that the order of columns in table_columns doesn't change between when the list is produced and when it is consumed (and remains stable across different rows that are produced in the same operation).

pixeltable/exprs/row_builder.py
aaron-siegel: Possibly, though I'm skeptical that will have a measurable performance impact. I can try it though.

pixeltable/store.py
aaron-siegel: Sure, I haven't done any optimization of add_computed_column() yet. Should probably just be another ticket.

pixeltable/store.py
aaron-siegel: Most of the perf gain is from batching; the reorganization of the insert data structures is a smaller contribution. But still significant; rolling up the dict as a one-shot operation is measurably more performant.

pixeltable/store.py
aaron-siegel: I don't understand the question. It's being done via sqlalchemy right now. The code comment shows how to do it via psycopg.

aaron-siegel: Oh, do you mean just run sql.text() via sqlalchemy?

pixeltable/exprs/row_builder.py
aaron-siegel: Done.

pixeltable/store.py
aaron-siegel: Where does RowBuilder get the pk column names and such?

aaron-siegel:
> * The previous logic for constructing the table row dicts was a bit easier to follow, because there were fewer dependencies (eg, on the order of columns) between the various pieces of code involved.

I agree that dicts are easier to follow.

> * At what point does the perf increase from increasing batch size level off? Would 64K or 128K row batches make sense if we're trying to insert 1M rows (which I think is a more realistic target than 100K rows).

In my tests, I found that 10k row batches gave a boost over 5k row batches, but beyond that it was diminishing returns. I haven't tried testing it with wider tables. We could certainly set a larger batch size; the only concern is hitting some memory ceiling if the tables are very wide or contain large text/json/array values.

Perhaps a follow-on could involve some sort of dynamic batch sizing.

pixeltable/exprs/data_row.py
mkornacker: I'm not following. How would it ever lose the exception, once it's set? I think it's still a 'has_any_exc', not a 'may have'.

pixeltable/exprs/row_builder.py
```diff
@@ -427,33 +430,52 @@ def eval(
                         expr, f'expression {expr}', data_row.get_exc(expr.slot_idx), exc_tb, input_vals, 0
                     ) from exc
 
-    def create_table_row(self, data_row: DataRow, exc_col_ids: set[int]) -> tuple[dict[str, Any], int]:
+    def create_table_row(self, data_row: DataRow, exc_col_ids: set[int]) -> tuple[list[Any], int]:
         """Create a table row from the slots that have an output column assigned
 
-        Return tuple[dict that represents a stored row (can be passed to sql.insert()), # of exceptions]
+        Return tuple[list of stored row in `self.table_columns` order, # of exceptions]
```
mkornacker: 'list of stored row'?

pixeltable/exprs/row_builder.py
mkornacker: Let's, or at least leave a to-do. Keep in mind that even if the impact is small right now, with more perf improvements elsewhere it'll also make this more significant. Saving on object creation seems like a generally useful strategy.

pixeltable/store.py
```diff
@@ -41,7 +41,7 @@ class StoreBase:
     v_max_col: sql.Column
     base: Optional[StoreBase]
 
-    __INSERT_BATCH_SIZE = 1000
+    __INSERT_BATCH_SIZE = 10_000
```
mkornacker, line 48: Leave comment about declining perf improvements beyond this? Otherwise it looks a bit unmotivated.

pixeltable/store.py
```diff
@@ -145,26 +146,23 @@ def _move_tmp_media_file(self, file_url: Optional[str], col: catalog.Column, v_m
...
         """Return Tuple[complete table row, # of exceptions] for insert()
         Creates a row that includes the PK columns, with the values from input_row.pk.
         Returns:
             Tuple[complete table row, # of exceptions]
         """
         table_row, num_excs = row_builder.create_table_row(input_row, exc_col_ids)
         assert len(pk) == len(self._pk_cols)
-        for pk_col, pk_val in zip(self._pk_cols, pk):
-            table_row[pk_col.name] = pk_val
+        table_row = [*pk, *table_row]
```
mkornacker: Looking at this, maybe RowBuilder.create_table_row() should add the pk as well, given that it now has access to it. And then we can get rid of this method.

pixeltable/store.py
mkornacker: Can we prioritize this? Otherwise you end up with a (to the user) unexplained perf gap.

pixeltable/exprs/data_row.py
aaron-siegel: Individual slots might get ejected via `clear(idxs)`. If those slots contained the only exceptions in the row, then it's possible for the row to revert to an exception-free state.

pixeltable/exprs/row_builder.py
aaron-siegel: fixed

pixeltable/exprs/row_builder.py
aaron-siegel: It's not so obvious to me. Python list resizing is pretty well optimized. Moreover, there's some work to do to determine the proper size, since it depends on how many columns record errors. And Python has no clear pattern for presizing a list; you have to presize *and* prepopulate it, which is another tax. I'm skeptical the net effect will be helpful and it might even add up to a performance penalty.

I've found it's generally dangerous to implement performance optimizations on the grounds that they "seem" generally useful or that they might be useful in the future. Often one ends up just complicating the code with no real benefit. I've made this mistake many times. That's why I strive to always measure first, and only optimize if we can quantify the benefit.

Update: To humor you I implemented this one as a trial; it added a net of 5 lines of code and performance was actually slightly worse, by an average of around 20 ms in my trial runs (though still within the margin of error). I think it's best to keep the code as-is.

pixeltable/store.py
aaron-siegel: Done.

pixeltable/store.py
aaron-siegel: Done.

PR #661, opened 2025-06-12
First cut at MCP integration

aaron-siegel:


pixeltable/func/mcp.py
```diff
@@ -0,0 +1,69 @@
...
+        invoke.__doc__ = mcp_tool.description
+
+    input_schema = mcp_tool.inputSchema
+    params = {
+        name: __mcp_param_to_pxt_type(mcp_tool.name, name, param) for name, param in input_schema['properties'].items()
+    }
+    required = input_schema.get('required', [])
+
+    # Ensure that any params not appearing in `required` are nullable.
+    # (A required param might or might not be nullable, since its type might be an 'anyOf' containing a null.)
```
mkornacker, line 58: Do we need to deal with that case here?

tests/test_mcp.py
```diff
@@ -0,0 +1,61 @@
...
+from .utils import skip_test_if_no_client, skip_test_if_not_installed
+
+_logger = logging.getLogger('pixeltable')
+
+
+class TestMcp:
+    def test_mcp_server(self, reset_db: None, init_mcp_server: None) -> None:
+        skip_test_if_not_installed('mcp')
+
+        udfs = asyncio.run(pxt.mcp_udfs('http://localhost:8000/mcp'))
```
mkornacker: As discussed, let's make that synchronous.

mkornacker (approved):

pixeltable/func/mcp.py
aaron-siegel: No; it'll get taken care of in `from_json_schema`.

tests/test_mcp.py
aaron-siegel: Done.

PR #664, opened 2025-06-17
Use ColumnHandles in external stores

aaron-siegel:
Updates ExternalStore, Project, and LabelStudioProject code to use `ColumnHandle` references to avoid retaining catalog objects.


pixeltable/catalog/table_version_handle.py
```diff
@@ -67,3 +68,19 @@ def as_dict(self) -> dict:
...
         return cls(UUID(d['id']), d['effective_version'])
+
+
+@dataclass(frozen=True)
+class ColumnHandle:
+    tbl_version: TableVersionHandle
+    col_id: int
+
+    def get(self) -> 'Column':
+        return self.tbl_version.get().cols_by_id[self.col_id]
```
mkornacker, line 87: This is not guaranteed to work if it's the current version and the column got dropped. This needs to raise Error in that case. See Catalog.get_tbl_version() for dropped tables.

pixeltable/io/label_studio.py
```diff
@@ -152,20 +157,22 @@ def __update_tasks(self, t: Table, existing_tasks: dict[tuple, dict]) -> SyncSta
...
         t_rl_cols = [
-            t_col for t_col, ext_col_name in self.col_mapping.items() if ext_col_name in config.rectangle_labels
+            t_col.get() for t_col, ext_col_name in self.col_mapping.items() if ext_col_name in config.rectangle_labels
         ]
 
         # Destinations for `rectanglelabels` preannotations
         rl_info = list(config.rectangle_labels.values())
 
-        _logger.debug('`t_data_cols`: %s', t_data_cols)
+        _logger.debug('`t_data_col_ids`: %s', t_data_cols)
```
mkornacker: Do we still need this?

pixeltable/io/label_studio.py
```diff
@@ -152,20 +157,22 @@ def __update_tasks(self, t: Table, existing_tasks: dict[tuple, dict]) -> SyncSta
         config = self.__project_config
 
         # Columns in `t` that map to Label Studio data keys
-        t_data_cols = [t_col for t_col, ext_col_name in self.col_mapping.items() if ext_col_name in config.data_keys]
+        t_data_cols = [
```
mkornacker: Why not pass ColHandles around everywhere instead? That would make it easier to follow.

pixeltable/catalog/table_version_handle.py
aaron-siegel: Added.

pixeltable/io/label_studio.py
aaron-siegel: I reverted the change but kept the debug messages (no harm in additional debug logging?)

pixeltable/io/label_studio.py
aaron-siegel: Done... I did it this way to avoid excessive calls to `.get()`, but I suppose it also increases the number of calls to `.handle`. And I agree this is clearer.

pixeltable/catalog/table_version_handle.py
```diff
@@ -76,6 +78,11 @@ class ColumnHandle:
     col_id: int
 
     def get(self) -> 'Column':
+        if self.col_id not in self.tbl_version.get().cols_by_id:
+            raise excs.Error(
+                f'Column has been dropped (no record for column ID {self.col_id} '
```
mkornacker: You could actually tell the user at which version it was dropped, that's available in _tbl_md.column_md[col_id].schema_version_drop.

mkornacker (approved):

pixeltable/catalog/table_version_handle.py
aaron-siegel: Done.

PR #667, opened 2025-06-19
Added Table.recompute_columns()

mkornacker:


pixeltable/catalog/table.py
```diff
@@ -1419,6 +1419,52 @@ def batch_update(
...
+        from pixeltable.catalog import Catalog
+
+        cat = Catalog.get()
+        # lock_mutable_tree=True: we need to be able to see whether any transitive view has column dependents
+        with cat.begin_xact(tbl=self._tbl_version_path, for_write=True, lock_mutable_tree=True):
+            if self._tbl_version_path.is_snapshot():
+                raise excs.Error('Cannot drop column from a snapshot.')
+            col_name: str
+            if isinstance(column, str):
+                col = self._tbl_version_path.get_column(column, include_bases=False)
```
aaron-siegel: I am thinking we should allow for `include_bases=True` here. I'd be pretty annoyed as a user if Pixeltable said "nope, I won't do that because you didn't call recompute_column() on the right base table," and I had to go remember which table the column was defined on etc.

One option would be for
```
view.recompute_column('base_table_col')
```
to recompute the base table column, but only for those rows that exist in the view.

pixeltable/catalog/table_version.py
```diff
@@ -1061,6 +1061,38 @@ def _validate_update_spec(
 
         return update_targets
 
+    def recompute_column(self, col_name: str, errors_only: bool = False, cascade: bool = True) -> UpdateStatus:
+        if self.is_snapshot:
```
aaron-siegel: Shouldn't these be assertions here (they were already checked in user-facing code?)

tests/test_table.py
```diff
@@ -1948,6 +1948,56 @@ def test_add_column_if_exists(self, test_tbl: catalog.Table, reload_tester: Relo
...
+    recompute_udf_error_val: Optional[int] = None
+
+    @staticmethod
+    @pxt.udf
+    def recompute_udf(i: int) -> int:
+        if TestTable.recompute_udf_error_val is not None and i % TestTable.recompute_udf_error_val == 0:
+            raise RuntimeError(f'Error in recompute_udf for value {i}')
+        return i + TestTable.recompute_udf_increment
+
+    def test_recompute_column(self, reset_db: None) -> None:
```
aaron-siegel, line 1969: Add tests for error messages?

aaron-siegel: Also add a test for `cascade=False`

aaron-siegel (changes requested):
I suggest we also add a convenience method
```
ColumnRef.recompute()
```

pixeltable/catalog/table.py
mkornacker: I don't think it's clear that this is desirable. Let's leave that as a follow-on feature.

pixeltable/catalog/table_version.py

tests/test_table.py
mkornacker: Done.

pixeltable/exprs/column_ref.py
```diff
@@ -170,6 +167,20 @@ def find_embedding_index(
             idx_info = embedding_idx_info
         return idx_info
 
+    def recompute(self, cascade: bool = True, errors_only: bool = False) -> catalog.UpdateStatus:
```
aaron-siegel: I'd suggest the signature
```
def recompute(self, *, cascade: bool = True, errors_only: bool = False) -> catalog.UpdateStatus:
```
i.e., make `cascade` and `errors_only` kwargs-only parameters to match `Table.recompute_columns()`

aaron-siegel (approved):

pixeltable/exprs/column_ref.py
mkornacker: Done

PR #670, opened 2025-06-20
Handle 429 errors in RateLimitScheduler

mkornacker:
I verified by hand that this picks up on 429 errors generated by the gemini endpoints and also records non-retryable errors.

We should figure out how to create tests for these scenarios.


pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -358,6 +371,127 @@ async def _exec(self, request: FnCallArgs, exec_ctx: ExecCtx, num_retries: int,
...
+        """Returns True if the exception indicates a rate limit error, and the retry delay in seconds."""
+        from http import HTTPStatus
+
+        # Check for HTTP status TOO_MANY_REQUESTS in various exception classes
+        is_rate_limit_error = False
+        retry_delay: Optional[float] = None
+
+        # requests.HTTPError
+        if (
+            hasattr(exc, 'response')
```
aaron-siegel, line 387: Why use `hasattr()` here and not `isinstance()` on the appropriate exception type?

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -358,6 +371,127 @@ async def _exec(self, request: FnCallArgs, exec_ctx: ExecCtx, num_retries: int,
...
+        Args:
+            retry_count: Number of retries attempted (0-based)
+            retry_after: Suggested delay from Retry-After header
+
+        Returns:
+            Delay in seconds
+        """
+        if retry_after is not None and retry_after > 0:
+            # Use server-suggested delay, but cap it at max_delay
+            return min(retry_after, self.MAX_RETRY_DELAY)
```
aaron-siegel: `max(min(retry_after, self.MAX_RETRY_DELAY), self.BASE_RETRY_DELAY)`

just in case?

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -250,8 +250,20 @@ class RequestRateScheduler(Scheduler):
     total_retried: int
 
     TIME_FORMAT = '%H:%M.%S %f'
-    MAX_RETRIES = 10
+    MAX_RETRIES = 3
```
aaron-siegel, line 254: I'd advocate for a larger number here. 10 is fine, now that we're appropriately identifying rate limit errors with backoff.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -358,6 +371,127 @@ async def _exec(self, request: FnCallArgs, exec_ctx: ExecCtx, num_retries: int,
...
+            header_dict = headers
+        else:
+            # headers are a list of tuples or other format
+            try:
+                header_dict = dict(headers)
+            except (TypeError, ValueError):
+                return None
+
+        # check Retry-After header
+        retry_after = header_dict.get('Retry-After') or header_dict.get('retry-after')
```
aaron-siegel: Might be simpler & more robust to normalize the dictionary first with something like

```
header_dict = {re.sub('-', '', k.lower()): v for k, v in header_dict.items()}
```
to turn `Retry-After`, `retry-after`, and any other variants etc. into `retryafter`

aaron-siegel:
Thanks for doing this; it's a huge improvement!

pixeltable/exec/expr_eval/schedulers.py
mkornacker: I saw problems with this when trying this out with gemini, which sent 429s when it meant 'quota exceeded' (= every retry will fail). With 10 retries it takes minutes to finish every cell.

pixeltable/exec/expr_eval/schedulers.py
mkornacker: This works for other exception classes that have a 'response' field. There seems to be a wide variety of possible exception classes to deal with here (gemini has its own xxxClientError).

pixeltable/exec/expr_eval/schedulers.py
mkornacker: Done.

pixeltable/exec/expr_eval/schedulers.py
mkornacker: Done.

pixeltable/exec/expr_eval/schedulers.py
aaron-siegel: Hmm interesting. I wonder if there's a way to be smarter and trap this. Note that with MAX_RETRIES = 3, we'll never come close to the max retry duration (the intervals will go 1 second - 2 seconds - 4 seconds - fail)

pixeltable/exec/expr_eval/schedulers.py
aaron-siegel: Ok, maybe add a code comment to say this? It's not obvious from reading the code why it's structured this way.

pixeltable/exec/expr_eval/schedulers.py
aaron-siegel: Hmm the change was only made in 1 of 2 places, and it's the other one from the one I suggested, and I think the place where it was actually made will never trigger because the minimum delay is already BASE_RETRY_DELAY?

aaron-siegel (approved):

pixeltable/exec/expr_eval/schedulers.py
mkornacker: The exception has a retry-after buried somewhere, but the code today doesn't pick up on it because it's a custom exception class. Gemini also has daily quotas, and if you exceed that you shouldn't retry at all, but this is also out of scope.

But if you're getting something else back, such as a 'invalid request' due to a bad api key, it correctly doesn't retry it, which is really the main point.

pixeltable/exec/expr_eval/schedulers.py
mkornacker: Done.

pixeltable/exec/expr_eval/schedulers.py
mkornacker: Done.

pixeltable/env.py
```diff
@@ -140,6 +142,33 @@ def __init__(self) -> None:
...
         self._current_session = None
         self._dbms = None
+        self._event_loop = None
+
+    def _init_event_loop(self) -> None:
+        try:
+            # check if we are already in an event loop (eg, Jupyter's); if so, patch it to allow
+            # multiple run_until_complete()
+            running_loop = asyncio.get_running_loop()
+            import nest_asyncio  # type: ignore[import-untyped]
```
aaron-siegel: Is there any reason not to put this at top of file?

pixeltable/env.py
```diff
@@ -140,6 +142,33 @@ def __init__(self) -> None:
...
+
+            nest_asyncio.apply()
+            self._event_loop = running_loop
+            print('Patched running loop')
```
aaron-siegel: Is this print statement intentional? If so, use console logger rather than print()

tests/test_table.py
```diff
@@ -1588,6 +1588,7 @@ def test_computed_col_exceptions(self, reset_db: None, test_tbl: catalog.Table)
         msgs = t.select(msg=t.add1.errormsg).collect()['msg']
         assert sum('division by zero' in msg for msg in msgs if msg is not None) == 10
 
+    @pytest.mark.skip('Crashes pytest')
```
aaron-siegel, line 1591: Please create a rough-edges ticket to investigate why this is happening

aaron-siegel (approved):
seems reasonable, just some minor suggestions

pixeltable/env.py

pixeltable/env.py

tests/test_table.py
mkornacker: https://pixeltable.atlassian.net/browse/PXT-638

PR #676, opened 2025-06-23
CI workflow for random table ops

aaron-siegel:
Sets up a CI workflow for random table ops. It runs nightly on 8 workers for 15 minutes (we'll increase this to 5 hours or longer once we have confidence in the script). The script fails if any errors are encountered, except for the following error messages, which are ignored:
- "That Pixeltable operation could not be completed" (concurrency conflict)
- "Table was dropped" (by one of the other workers)
- "Path.*does not exist"

### DIFF
--- a/tool/worker_harness.py
+++ b/tool/worker_harness.py
@@ -14,8 +14,20 @@ def run_workers(num_workers: int, duration: float, script: str) -> None:
 
     time.sleep(duration)
 
+    success = True
+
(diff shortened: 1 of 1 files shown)


mkornacker (approved):

mkornacker:
Regarding failure: the script should only fail if it encounters exceptions other than pxt.Error.

PR #681, opened 2025-06-25
`pxt.ls()` command for pretty-printing all contents of a Pixeltable dir

aaron-siegel:
`pxt.ls(path: str = '')` now generates output that looks like this:
```
       Name              Kind Version                    Base
   replica1  snapshot-replica          <anonymous base table>
  snapshot1          snapshot                         view1:2
  snapshot2          snapshot                test_dir.view2:0
     subdir               dir                                
        tbl             table       4                        
      view2              view       1            test_dir.tbl
```


jpeterson-pxt:
I haven't read the code, but should it have parameter to make it recursive to show the contents of the child directories as well?

pixeltable/globals.py
```diff
@@ -633,6 +633,69 @@ def drop_dir(path: str, force: bool = False, if_not_exists: Literal['error', 'ig
     Catalog.get().drop_dir(path_obj, if_not_exists=if_not_exists_, force=force)
 
 
+def ls(path: str = '') -> pd.DataFrame:
```
mkornacker, line 640: Docstring missing.

pixeltable/globals.py
```diff
@@ -633,6 +633,69 @@ def drop_dir(path: str, force: bool = False, if_not_exists: Literal['error', 'ig
...
+    from pixeltable.metadata import schema
+
+    cat = Catalog.get()
+    path_obj = catalog.Path(path, empty_is_valid=True)
+    dir_entries = cat.get_dir_contents(path_obj)
+    rows: list[list[str]] = []
+    with Env.get().begin_xact():
```
mkornacker: Catalog.begin_xact().

Never use Env.begin_xact(), unless there is a specific reason.

pixeltable/globals.py
```diff
@@ -633,6 +633,69 @@ def drop_dir(path: str, force: bool = False, if_not_exists: Literal['error', 'ig
...
+                continue
+            if entry.dir is not None:
+                kind = 'dir'
+                version = ''
+                base = ''
+            else:
+                assert entry.table is not None
+                assert isinstance(entry.table, schema.Table)
+                tbl = cat.get_table_by_id(entry.table.id)
+                tvp = tbl._tbl_version_path
```
mkornacker: Let's use tbl.get_metadata() below here, instead of reaching into the internals. If there's something missing in the dict, let's add it.

Might need to add kind and base_version.

aaron-siegel:
> I haven't read the code, but should it have parameter to make it recursive to show the contents of the child directories as well?

Good idea - let's make that a follow-up ticket. Needs a little thought on how to set up the display.

pixeltable/globals.py
aaron-siegel: Good catch. I'd written this code before Catalog.begin_xact() was introduced and then forgot to change it. Fixed.

pixeltable/globals.py
aaron-siegel: Done.

pixeltable/globals.py
aaron-siegel: Done.

pixeltable/catalog/view.py
```diff
@@ -272,9 +272,22 @@ def _get_snapshot_path(cls, tbl_version_path: TableVersionPath) -> TableVersionP
...
     def _get_metadata(self) -> dict[str, Any]:
+        from .catalog import Catalog
+
         md = super()._get_metadata()
         md['is_view'] = True
         md['is_snapshot'] = self._tbl_version_path.is_snapshot()
+        # TODO: We'll be able to clean the logic for md['base'] up a bit once we have Table objects that are handles to
+        #     specific versions. Right now, Table objects only represent an effective version of None, so there's no
+        #     single source of path/version information.
+        if self._id != self._tbl_version_path.tbl_id:
```
mkornacker: Call _get_base_table() and _effective_base_versions() instead.

mkornacker (approved):

PR #682, opened 2025-06-30
Prepare to report more status in table.history()

jpeterson-pxt:
Hoist UpdateStatus into its own file.
    Lower emission of counts from store_table.insert to RowCountStats.
    Add doc to test_history.
    Add comment field to UpdateStatus.

### DIFF
--- a/pixeltable/catalog/update_status.py
+++ b/pixeltable/catalog/update_status.py
@@ -1,0 +1,153 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from IPython.lib.pretty import RepresentationPrinter
+
+
+@dataclass(frozen=True)
+class RowCountStats:
+    """
+    Statistics about the counts of rows affected by a table operation.
+    """
+
+    ins_rows: int = 0  # rows inserted
+    del_rows: int = 0  # rows deleted
+    upd_rows: int = 0  # rows updated
+    num_excs: int = 0  # total number of exceptions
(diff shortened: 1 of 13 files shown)


pixeltable/catalog/update_status.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+
+
+@dataclass(frozen=True)
+class UpdateStatus:
+    """
+    Information about changes to table data or table schema
+    """
+
+    comment: str = ''  # Note about this operation
```
aaron-siegel, line 59: Question: Would it be better to use an enum here, so that there's some structure around it, rather than treating it as a free-form comment? I'm assuming there are a fixed finite number of update types, but correct me if that's not a safe assumption

jpeterson-pxt: I considered using something like "operation: enum", and left it as "comment: str" for the following reasons:

1. The field is intended for human consumption. There is no benefit from being restrictive, as there is no need for machinability.
2. The string is flexible wrt new uses or modifications of existing uses. Modifiers can be added to operations, for example.

aaron-siegel (approved):

pixeltable/catalog/update_status.py
mkornacker: I don't think there is any real use case for this and would prefer not to have it, simply to reduce complexity. Saying "it might be useful in the future" is not a use case.

re enum: that would make more sense, because then you can do something with it programmatically. Strings make that more problematic.

For changes to the API (which this is), I would prefer to see some preceding discussion.

PR #683, opened 2025-07-01
Support NULL JSON fields in test_packager.py

jpeterson-pxt:


aaron-siegel:
Please add or expand a test to cover this edge case. Thanks!

aaron-siegel (changes requested):
Please add or expand a test to cover this edge case. Thanks!

PR #690, opened 2025-07-07
Refactor create_replica to conform to concurrency protocol

aaron-siegel:
Refactors Catalog.create_replica() to conform to the concurrency protocol:
- Instantiation of TableVersion instances is moved inside __store_tbl_md
- Transaction boundary is no longer localized to create_replica(), so that the data import can take place within the same transaction

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -824,8 +824,9 @@ class Catalog:
             # If it's a new version, this will result in a new TableVersion record being created.
             self.__store_replica_md(replica_path, ancestor_md)
 
-            # Now we must clear cached metadata for the ancestor table, so that if descendants have computed columns
-            # that reference the new version of the ancestor, then they will be properly resolved.
+            # Now we must clear cached metadata for the ancestor table, to force the next table operation to pick up
+            # the new TableVersion instance. This is necessary because computed columns of descendant tables might
+            # reference columns of the ancestor table that only exist in the new version.
             replica = Catalog.get().get_table_by_id(ancestor_id)
             assert replica is not None  # If it didn't exist before, it must have been created by now.
             replica._tbl_version_path.clear_cached_md()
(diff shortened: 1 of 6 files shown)


pixeltable/catalog/catalog.py
```diff
@@ -827,11 +819,21 @@ def create_replica(
...
 
-            # Store the metadata; as before, it could be a new version or a known version.
+            # Store the metadata; it could be a new version (in which case a new record will be created), or a known
+            # version (in which case the newly received metadata will be validated as identical).
+            # If it's a new version, this will result in a new TableVersion record being created.
             self.__store_replica_md(replica_path, ancestor_md)
 
-        # don't create TableVersion instances at this point, they would be superseded by calls to TV.create_replica()
-        # in TableRestorer.restore()
+            # Now we must clear cached metadata for the ancestor table, so that if descendants have computed columns
```
mkornacker: More directly stated: clear_cached_md() will force the next Table operation to get the replica's just-created TableVersion instance

tests/share/test_packager.py
```diff
@@ -345,16 +345,18 @@ def test_multi_view_round_trip_3(self, reset_db: None, pure_snapshots: bool) ->
...
         """
         Snapshots that involve all the different column types.
         """
         t = all_datatypes_tbl
         snap1 = pxt.create_snapshot('snap1', t.where(t.row_id % 2 != 0))
         bundle1 = self.__package_table(snap1)
 
-        more_data = create_table_data(t, num_rows=22)
-        t.insert(more_data[11:])
+        if different_tv:
```
mkornacker: Why different tv?

pixeltable/catalog/catalog.py
aaron-siegel: I've updated the comment.

tests/share/test_packager.py
aaron-siegel: Changed variable name and added a comment per our discussion today.

PR #691, opened 2025-07-07
Consolidate media operations in the MediaStore module

jpeterson-pxt:


pixeltable/utils/media_store.py
```diff
@@ -34,6 +35,44 @@ def prepare_media_path(cls, tbl_id: UUID, col_id: int, version: int, ext: Option
         parent.mkdir(parents=True, exist_ok=True)
         return parent / f'{tbl_id.hex}_{col_id}_{version}_{id_hex}{ext or ""}'
 
+    @classmethod
+    def move_tmp_media_file(cls, file_url: Optional[str], tbl_id: UUID, col_id: int, v_min: int) -> Optional[str]:
```
aaron-siegel, line 39: Should we also move the relevant code from `flush_img` etc inside MediaStore, so that `prepare_media_path` is purely internal to MediaStore?

aaron-siegel (approved):

pixeltable/utils/media_store.py
jpeterson-pxt: Yes, I have other changesets to put into PRs

PR #692, opened 2025-07-08
Multi-phase transactions

mkornacker:
This is currently limited to Catalog.create_view().

Follow-up work: apply the same pattern to all other transactional updates that require schema changes:
- create_table
- create_replica
- add_embedding_index
- add_columns
- add_computed_columns


pixeltable/catalog/catalog.py
```diff
@@ -70,18 +71,41 @@ def _unpack_row(
...
+    for_write: bool,
+    lock_mutable_tree: bool = False,
+    finalize_pending_ops: bool = True,
+) -> Callable[[Callable[..., T]], Callable[..., T]]:
     def decorator(op: Callable[..., T]) -> Callable[..., T]:
         @functools.wraps(op)
         def loop(*args: Any, **kwargs: Any) -> T:
+            cat = Catalog.get()
+            # retry_loop() is reentrant
+            if cat._in_retry_loop:
```
aaron-siegel, line 85: What if we're in a retry loop, but one that was initially entered with weaker constraints than this one (lock_mutable tree etc)? Should we be checking for this (and, say, firing an assertion if we try to strengthen the constraints on an existing retry_loop)?

pixeltable/catalog/catalog.py
```diff
@@ -70,18 +71,41 @@ def _unpack_row(
...
+                        for_write=for_write,
+                        convert_db_excs=False,
+                        lock_mutable_tree=lock_mutable_tree,
+                        finalize_pending_ops=finalize_pending_ops,
+                    ):
                         return op(*args, **kwargs)
+                except PendingTableOpsError as e:
+                    assert finalize_pending_ops
+                    _logger.debug(f'retry_loop(): finalizing pending ops for {e.tbl_id}')
+                    print(f'retry_loop(): finalizing pending ops for {e.tbl_id}')
```
aaron-siegel: Spurious `print()`

pixeltable/catalog/catalog.py
```diff
@@ -97,16 +121,31 @@ def loop(*args: Any, **kwargs: Any) -> T:
...
                     raise
+                finally:
+                    cat._in_retry_loop = False
 
         return loop
 
     return decorator
 
 
+class PendingTableOpsError(Exception):
```
aaron-siegel, line 129: Move this to pixeltable.exceptions?

pixeltable/catalog/catalog.py
```diff
@@ -237,10 +280,14 @@ def begin_xact(
 
         If convert_db_excs == True, converts DBAPIErrors into excs.Errors.
         """
+        assert tbl is None or tbl_id is None
```
aaron-siegel: Is it possible to specify them both? If not, change this to `assert (tbl is None) != (tbl_id is None)`. If so, the succeeding logic would seem to be incomplete (the `elif` should be an `if`)

pixeltable/catalog/catalog.py
```diff
@@ -252,32 +299,57 @@ def begin_xact(
         # )
         # _logger.debug(f'begin_xact(): {tv_msg}')
         num_retries = 0
+        pending_ops_tbl_id: Optional[UUID] = None
         while True:
+            if pending_ops_tbl_id is not None:
+                _logger.debug(f'begin_xact(): finalizing pending ops for {pending_ops_tbl_id}')
+                print(f'begin_xact(): finalizing pending ops for {pending_ops_tbl_id}')
```
aaron-siegel: Spurious `print()`

pixeltable/catalog/catalog.py
```diff
@@ -332,42 +419,57 @@ def in_write_xact(self) -> bool:
...
+        For writes: force acquisition of an X-lock on a Table record via a blind update.
 
         Either tbl_id or dir_id/tbl_name need to be specified.
         Returns True if the table was locked, False if it was a snapshot or not found.
         If lock_mutable_tree, recursively locks all mutable views of the table.
 
-        Returns False if the table is a snapshot or not found and !raise_if_not_exists.
+        Returns a handle to what was locked.
         """
+        assert (tbl_id is not None) ^ (dir_id is not None and tbl_name is not None)
```
aaron-siegel: Really nitpicky comment, but I find `!=` clearer as the operator here, since it's a logical (not numeric) compound. Obviously they mean the same thing, on some level.

pixeltable/catalog/catalog.py
```diff
@@ -722,49 +839,61 @@ def create_view(
...
+                iterator_args=iterator_args,
+                num_retained_versions=num_retained_versions,
+                comment=comment,
+                media_validation=media_validation,
+            )
+            tbl_id = UUID(md.tbl_md.tbl_id)
+            self.store_tbl_md(tbl_id, dir._id, md.tbl_md, md.version_md, md.schema_version_md, ops)
+            return tbl_id
+
+        view_id = retry_loop(for_write=True, finalize_pending_ops=True)(create_fn)()
```
aaron-siegel: Why this syntax instead of decorating `create_fn`?

pixeltable/catalog/catalog.py
```diff
@@ -722,49 +839,61 @@ def create_view(
...
+        view_handle = TableVersionHandle(view_id, effective_version=0 if is_snapshot else None)
```
aaron-siegel: I don't quite understand what's happening here in the case of a pure snapshot (which doesn't have table versions).

Anyways this only gets used in the `if not is_snapshot` block; so can't the definition of `view_handle` be moved inside that block? That would be a lot clearer to me.

pixeltable/catalog/catalog.py
```diff
@@ -931,7 +1060,66 @@ def get_table(self, path: Path) -> Table:
...
+                tv = self.get_tbl_version(tbl_id, tbl_version, check_pending_ops=False)
+                tv.exec_op(op)
+                with self.begin_xact(tbl_id=tbl_id, for_write=True, convert_db_excs=False, finalize_pending_ops=False):
+                    Env.get().conn.execute(delete_next_op_q)
+            except sql.exc.DBAPIError as e:
+                # TODO: why are we still seeing these here, instead of them getting taken care of by the retry
+                # logic of begin_xact()?
+                if isinstance(e.orig, (psycopg.errors.SerializationFailure, psycopg.errors.LockNotAvailable)):
+                    num_retries += 1
+                    print(f'finalize_pending_ops(): retrying ({num_retries}) op {op!s} after {type(e.orig)}')
```
aaron-siegel: `print()`

Suggestion: turn these into debug-level messages on `console_logger` so we don't have to keep track of adding and removing print statements

pixeltable/catalog/catalog.py
```diff
@@ -931,7 +1060,66 @@ def get_table(self, path: Path) -> Table:
...
+                    _logger.debug(f'finalize_pending_ops(): retrying ({num_retries}) op {op!s} after {type(e.orig)}')
+                    time.sleep(random.uniform(0.1, 0.5))
+                    continue
+                else:
+                    raise
+            except Exception as e:
+                _logger.debug(f'finalize_pending_ops(): caught {e}')
+                print(f'finalize_pending_ops(): caught {e}')
```
aaron-siegel: `print()`

pixeltable/catalog/catalog.py
```diff
@@ -1219,7 +1416,12 @@ def _load_tbl(self, tbl_id: UUID) -> Optional[Table]:
         # this is a view; determine the sequence of TableVersions to load
         tbl_version_path: list[tuple[UUID, Optional[int]]] = []
         schema_version_md = schema.md_from_dict(schema.TableSchemaVersionMd, schema_version_record.md)
-        pure_snapshot = view_md.is_snapshot and view_md.predicate is None and len(schema_version_md.columns) == 0
+        pure_snapshot = (
```
aaron-siegel: There's an `is_pure_snapshot` property on `tbl_md` now; it'd be best to start using that so we don't miss clauses in the future... it probably needs to be updated to include `sample_clause` (I'd forgotten about that)

pixeltable/catalog/catalog.py
```diff
@@ -1492,7 +1725,12 @@ def _load_tbl_version(self, tbl_id: UUID, effective_version: Optional[int]) -> O
             )
         else:
             assert len(view_md.base_versions) > 0  # a view needs to have a base
-            pure_snapshot = view_md.is_snapshot and view_md.predicate is None and len(schema_version_md.columns) == 0
+            pure_snapshot = (
```
aaron-siegel, line 1820: See above - use `TableMd.is_pure_snapshot`

pixeltable/catalog/schema_object.py
```diff
@@ -42,10 +42,9 @@ def _path(self) -> str:
 
     def get_metadata(self) -> dict[str, Any]:
         """Returns metadata associated with this schema object."""
-        from pixeltable.catalog import Catalog
+        from pixeltable.catalog import retry_loop
 
-        with Catalog.get().begin_xact(for_write=False):
-            return self._get_metadata()
+        return retry_loop(for_write=False, finalize_pending_ops=True)(self._get_metadata)()
```
aaron-siegel: Why do this instead of decorating _get_metadata?

pixeltable/catalog/table_version.py
```diff
@@ -40,6 +41,19 @@
 _logger = logging.getLogger('pixeltable')
 
 
+@dataclasses.dataclass(frozen=True)
+class TableVersionMd:
```
aaron-siegel, line 45: I'm not enthralled with the name `FullTableMd` and I'm ok if we "subsume" it, but we should call this something other than `TableVersionMd`. It'll be way too confusing to have two different (closely related) abstractions in different modules with the same name.

pixeltable/catalog/table_version.py
```diff
@@ -190,6 +204,75 @@ def handle(self) -> 'TableVersionHandle':
 
         return TableVersionHandle(self.id, self.effective_version, self)
 
+    @classmethod
+    def create_md(
```
aaron-siegel: Maybe call this `create_initial_md`?

pixeltable/globals.py
```diff
@@ -679,6 +681,9 @@ def ls(path: str = '') -> pd.DataFrame:
                 if md['is_replica']:
                     kind = f'{kind}-replica'
             rows.append([name, kind, version, base])
+        return rows
+
+    rows = retry_loop(for_write=False)(op)()
```
aaron-siegel: Decorate `op`?

tests/test_migration.py
```diff
@@ -79,7 +79,9 @@ def test_db_migration(self, init_env: None) -> None:
             with orm.Session(env.engine) as session:
                 convert_table_md(env.engine, substitution_fn=self.__substitute_md)
 
-            env._upgrade_metadata()
+            # make sure we run the env db setup
+            Env.clear()
```
aaron-siegel: I don't think we need to introduce something new here; you should be able to just call `Env._init_env(reinit_db=True)`, unless I'm missing something?

pixeltable/catalog/catalog.py
mkornacker: This is checked in begin_xact().

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: This is really local to catalog.py and part of control flow, not so much an error condition. It's called Error because otherwise the linter complains.

This exception should never be seen outside of catalog.py.

pixeltable/catalog/catalog.py
mkornacker: It's at most one. Clarified with comment.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: I'll consolidate that later. It's not available for TableSchemaVersionMd.

pixeltable/catalog/catalog.py
mkornacker: I'll consolidate in the next round.

pixeltable/catalog/schema_object.py
mkornacker: Done.

pixeltable/catalog/table_version.py
mkornacker: We already have that situation: catalog.TableVersion and schema.TableVersion.

I'm open to renaming catalog.TableVersion (or schema.TableVersion, although the latter would require a schema migration and be even more annoying).

At least this way it's consistent (this dataclass only applies to catalog.TableVersion).

pixeltable/catalog/table_version.py
mkornacker: Isn't that implied? You give it a bunch of parameters such as the columns, and it gives you the md record for all of that back?

pixeltable/globals.py

tests/test_migration.py
mkornacker: Done.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/table_version.py
aaron-siegel: Well it's not *exactly* analogous; catalog.TableVersion and schema.TableVersion are largely disjoint, whereas schema.TableVersionMd is a component of this new thing ("X" is nested inside a different "X"), and to me that adds a new level of confusion.

pixeltable/catalog/catalog.py
aaron-siegel: Maybe add a comment here to clarify that?

pixeltable/catalog/catalog.py
aaron-siegel: Ok, I've always been a fan of keeping exception types together in one place in a code base but I don't feel too strongly about it.

We should clarify what `pixeltable.exceptions` is intended for then.

pixeltable/catalog/catalog.py
aaron-siegel: You should just be able to call it on your `tbl_md`, no?

pixeltable/catalog/table_version.py
aaron-siegel: It gives you a "version 0" md which isn't so obvious (we have a lot of things that build metadata structures and not all of them initialize at version 0) so that's why I suggested clarifying its purpose.

aaron-siegel (approved):

pixeltable/catalog/table_version.py

pixeltable/catalog/catalog.py
```diff
@@ -98,13 +97,12 @@ def loop(*args: Any, **kwargs: Any) -> T:
                         for_write=for_write,
                         convert_db_excs=False,
                         lock_mutable_tree=lock_mutable_tree,
-                        finalize_pending_ops=finalize_pending_ops,
+                        finalize_pending_ops=True,
                     ):
                         return op(*args, **kwargs)
                 except PendingTableOpsError as e:
-                    assert finalize_pending_ops
                     _logger.debug(f'retry_loop(): finalizing pending ops for {e.tbl_id}')
```
aaron-siegel: If you are using the console logger there is no reason to also put a separate `_logger.debug()` (this will just result in duplicate messages in the logs)

Same comment applies to other uses of console logger below.

pixeltable/catalog/catalog.py
```diff
@@ -434,20 +455,23 @@ def _acquire_path_locks(
...
-        self._acquire_tbl_lock(
+            return True  # nothing left to lock
+        handle = self._acquire_tbl_lock(
             tbl_id=tbl.tbl_id,
             for_write=True,
             lock_mutable_tree=lock_mutable_tree,
             raise_if_not_exists=True,
             check_pending_ops=check_pending_ops,
         )
+        _ = self.get_tbl_version(path_handles[0].id, path_handles[0].effective_version, validate_initialized=True)
```
aaron-siegel, line 473: This is counterintuitive. Are we relying on `get_tbl_version()` having side effects here?

pixeltable/catalog/catalog.py
```diff
@@ -513,9 +541,98 @@ def _acquire_tbl_lock(
                 )
         return TableVersionHandle(tbl_id, effective_version)
 
+    def _finalize_pending_ops(self, tbl_id: UUID) -> None:
+        """
+        Finalizes the next pending op for the given table. Returns True if there was a pending op, False otherwise.
```
aaron-siegel: Docstring talks about returning `True`/`False`, but return type is `None`

pixeltable/catalog/catalog.py
```diff
@@ -513,9 +541,98 @@ def _acquire_tbl_lock(
...
+                    raise
+            except Exception as e:
+                _logger.debug(f'finalize_pending_ops(): caught {e}')
+                Env.get().console_logger.debug(f'finalize_pending_ops(): caught {e}')
+                raise
+
+            num_retries = 0
+
+    def _debug_str(self) -> str:
+        tv_str = '\n'.join([str(k) for k in self._tbl_versions])
```
aaron-siegel: spurious list creation (here and next line)

pixeltable/catalog/table_version.py
```diff
@@ -377,13 +384,17 @@ def create(
...
+        tbl_version = cls(
+            tbl_id, table_md, table_version_md, effective_version, schema_version_md, [], base_path=base_path, base=base
+        )
         # TODO: break this up, so that Catalog.create_table() registers tbl_version
         cat._tbl_versions[tbl_id, effective_version] = tbl_version
         tbl_version.init()
         tbl_version.store_tbl.create()
         is_mutable = not is_snapshot and not table_md.is_replica
         if base is not None and base.get().is_mutable and is_mutable:
+            # TODO before check-in: remove this branch
```
aaron-siegel: just flagging this

tests/test_migration.py
```diff
@@ -80,7 +80,7 @@ def test_db_migration(self, init_env: None) -> None:
                 convert_table_md(env.engine, substitution_fn=self.__substitute_md)
 
             # make sure we run the env db setup
-            Env.clear()
+            Env._init_env()
```
aaron-siegel, line 83: Does this need to be called with `reinit_db=True`?

pixeltable/catalog/table_version.py
```diff
@@ -1508,6 +1520,16 @@ def version(self) -> int:
...
+        self._version_md.version = version
+
+    @property
+    def created_at(self) -> float:
+        return self._version_md.created_at
+
+    @created_at.setter
+    def created_at(self, ts: float) -> None:
+        assert self.effective_version is None
+        self._version_md.created_at = ts
```
aaron-siegel, line 1494: I don't love the in-place update here; I'd prefer these be frozen dataclasses and updated via

`self._version_md = self._version_md.replace(created_at=ts)`

pixeltable/metadata/schema.py
```diff
@@ -224,7 +235,7 @@ class Table(Base):
     lock_dummy: orm.Mapped[int] = orm.mapped_column(BigInteger, nullable=True)
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass
```
aaron-siegel, line 239: I substantially prefer keeping these as frozen dataclasses; if that's controversial then we should discuss.

aaron-siegel (changes requested):

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: It updates the cache. Left comment.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/table_version.py

pixeltable/catalog/table_version.py
mkornacker: Why?

pixeltable/metadata/schema.py
mkornacker: TableMd and TableSchemaVersionMd are not frozen.

tests/test_migration.py
mkornacker: It fails with reinit_db=True.

pixeltable/catalog/catalog.py
aaron-siegel: Ok, I wonder if it'd be clearer to have a separate `_update_tbl_version_cache()` that is called by `get_tbl_version()` and can also be called separately when needed.

It always makes me squint a bit when a `get_*()` is called for its side effects.

pixeltable/catalog/table_version.py
aaron-siegel: It makes their usage easier to reason about. This is a fairly standard pattern for dataclasses btw.

Note that we have to copy them around anyway; right now we're doing the copy in the `TableVersion` initializer. It's much easier to reason about (esp. for ppl new to the code) if the copying is done at the time the field values are changed & the data structures can be presumed immutable.

pixeltable/metadata/schema.py
aaron-siegel: I'd argue they should be and we should make that change at some point (I'd be happy to do it).

Reverting TableVersionMd to a non-frozen data class is a step backwards IMO.

pixeltable/catalog/catalog.py
```diff
@@ -1403,25 +1700,32 @@ def store_tbl_md(
...
 
-        session.execute(
-            sql.update(schema.TableVersion.__table__)
-            .values({schema.TableVersion.md: dataclasses.asdict(version_md)})
-            .where(schema.TableVersion.tbl_id == version_md.tbl_id, schema.TableVersion.version == version_md.version)
+        stmt = (
+            sql.update(schema.TableVersion)
+            .where(schema.TableVersion.tbl_id == tbl_id, schema.TableVersion.version == version)
+            .values(
+                md=schema.TableVersion.md.op('||')({'additional_md': {'update_status': dataclasses.asdict(status)}})
```
jpeterson-pxt, line 1723: This change replaces anything in the 'additional_md' with the 'update_status'. 
The update_status is a part of the TableVersion.md. 
Perhaps this change should be 
`md=schema.TableVersion.md.op('||')({'update_status': dataclasses.asdict(status)})`

PR #694, opened 2025-07-08
Consolidate MediaStore file operations, including temp file name creation

jpeterson-pxt:


pixeltable/exec/data_row_batch.py
```diff
@@ -90,7 +90,8 @@ def flush_imgs(
             idx_range = slice(0, len(self.rows))
         for row in self.rows[idx_range]:
             for info in stored_img_info:
-                filepath = str(MediaStore.prepare_media_path(self.tbl.id, info.col.id, self.tbl.get().version))
+                col = info.col
+                filepath = str(MediaStore.prepare_media_path(col.tbl.id, col.id, col.tbl.version))
```
aaron-siegel, line 95: What are the implications of using `col.tbl.id` instead of `self.tbl.id`?

I'd feel more comfortable with this change if we put in an

`assert col.tbl.id == self.tbl.id`

(and if that assertion fails I'd want to understand why).

pixeltable/exec/in_memory_data_node.py
```diff
@@ -63,13 +63,11 @@ def _open(self) -> None:
...
-                if col_info.col.col_type.is_image_type() and isinstance(val, bytes):
-                    # this is a literal image, ie, a sequence of bytes; we save this as a media file and store the path
-                    path = str(MediaStore.prepare_media_path(self.tbl.id, col_info.col.id, self.tbl.get().version))
-                    with open(path, 'wb') as fp:
-                        fp.write(val)
-                    self.output_rows[row_idx][col_info.slot_idx] = path
+                col = col_info.col
+                if col.col_type.is_image_type() and isinstance(val, bytes):
+                    # this is a literal media file, ie, a sequence of bytes; save it as a binary file and store the path
+                    path = MediaStore.save_media_file(val, col.tbl.id, col.id, col.tbl.version)
```
aaron-siegel, line 70: same comment as above (add assertion)

pixeltable/utils/media_store.py
```diff
@@ -73,6 +73,16 @@ def relocate_local_media_file(cls, src_path: Path, tbl_id: UUID, col_id: int, tb
...
         return urllib.parse.urljoin('file:', urllib.request.pathname2url(str(dest_path)))
 
+    @classmethod
+    def save_media_file(cls, file_data: bytes, tbl_id: UUID, col_id: int, tbl_version: int) -> Path:
+        """Save a media binary data to a file in the MediaStore."""
+        assert isinstance(file_data, bytes)
+        media_path = cls.prepare_media_path(tbl_id, col_id, tbl_version)
+        with open(media_path, 'wb') as f:
+            f.write(file_data)
+            f.flush()
```
aaron-siegel: Is this necessary? I'd expect the buffer to be flushed automatically once the `with open` context manager scope is exited.

jpeterson-pxt: Yes, it was really sort of a marker for me to do more research. 
What's actually required is this:
    f.flush()  # Ensures Python buffers are written to OS
    os.fsync(f.fileno())  # Forces OS to write to physical storage
Otherwise a system crash might result in data not being on media, but lost from an OS buffer cache.

pixeltable/exec/data_row_batch.py
jpeterson-pxt: I believe that col.tbl.id the right answer, even more than self.tbl.id, per this comment:
`    # we store a TableVersion here, not a TableVersionHandle, because this column is owned by that TableVersion instance
    # (re-resolving it later to a different instance doesn't make sense)
    tbl: Optional[TableVersion]
`
But perhaps the Column data structure comment is incorrect. 
I can add the assert if you still feel it is necessary.

jpeterson-pxt: Done

pixeltable/utils/media_store.py

aaron-siegel (approved):

PR #696, opened 2025-07-10
Load replicas as views

aaron-siegel:
Ensures that all replicas are loaded as instances of `View` (regardless of the kind of the original table). Safeguards replicas against all mutable operations, regardless of their kind.

Also rewrites a bunch of error messages to make them clearer / more consistent.

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -1544,7 +1544,7 @@ class Table(SchemaObject):
         """
         from pixeltable.catalog import Catalog
 
-        if self._tbl_version_path.is_snapshot() or self._tbl_version_path.is_replica():
+        if not self._tbl_version_path.is_mutable():
             return
         with Catalog.get().begin_xact(tbl=self._tbl_version_path, for_write=True, lock_mutable_tree=False):
             all_stores = self.external_stores()
@@ -1584,7 +1584,7 @@ class Table(SchemaObject):
         """
         from pixeltable.catalog import Catalog
 
-        if self._tbl_version_path.is_snapshot() or self._tbl_version_path.is_replica():
+        if not self._tbl_version_path.is_mutable():
             return UpdateStatus()
         # we lock the entire tree starting at the root base table in order to ensure that all synced columns can
         # have their updates propagated down the tree
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -492,8 +492,7 @@ class Table(SchemaObject):
 
         # lock_mutable_tree=True: we might end up having to drop existing columns, which requires locking the tree
         with Catalog.get().begin_xact(tbl=self._tbl_version_path, for_write=True, lock_mutable_tree=True):
-            if self._tbl_version_path.is_snapshot():
-                raise excs.Error('Cannot add column to a snapshot.')
+            self.__check_mutable('add columns to')
             col_schema = {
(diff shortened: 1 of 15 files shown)


pixeltable/catalog/table.py
```diff
@@ -1560,7 +1544,7 @@ def unlink_external_stores(
         """
         from pixeltable.catalog import Catalog
 
-        if self._tbl_version_path.is_snapshot():
+        if self._tbl_version_path.is_snapshot() or self._tbl_version_path.is_replica():
```
mkornacker: add a TVP.is_mutable() and use that?

pixeltable/catalog/table_version.py
```diff
@@ -664,6 +664,7 @@ def add_columns(
     ) -> UpdateStatus:
         """Adds columns to the table."""
         assert not self.is_snapshot
+        assert not self.is_replica
```
mkornacker: replace both with self.is_mutable

pixeltable/catalog/view.py
```diff
@@ -45,9 +45,18 @@ def __init__(self, id: UUID, dir_id: UUID, name: str, tbl_version_path: TableVer
...
+        name: str
+        if self._tbl_version_path.is_snapshot():
+            name = 'snapshot'
+        elif self._tbl_version_path.is_view():
+            name = 'view'
+        else:
+            assert self._tbl_version_path.is_replica()
+            name = 'table'
+        if self._tbl_version_path.is_replica():
+            name = f'replica-{name}'
```
mkornacker: How about '{name} replica' instead? Reads a little more nicely.

tests/share/test_packager.py
```diff
@@ -464,3 +464,55 @@ def test_multi_view_non_snapshot_round_trip(self, reset_db: None) -> None:
...
+        v_bundle = self.__package_table(v)
+
+        clean_db()
+        reload_catalog()
+
+        self.__restore_and_check_table(v_bundle, 'view_replica')
+        # Check that test_tbl was instantiated as a system table
+        assert pxt.list_tables() == ['view_replica']
+        system_path = pxt.catalog.Path('_system', allow_system_paths=True)
+        system_contents = pxt.catalog.Catalog.get().get_dir_contents(system_path)
```
mkornacker: I really want us to get away from using module internals in tests, which has created a fair amount of work for me in the past. How about we add a _include_sys_tbls or similar kwarg to list_tables()?

pixeltable/catalog/table.py
aaron-siegel: It looks like I already added it at some point, I just neglected to use it :)

pixeltable/catalog/table_version.py
aaron-siegel: Done. Also changed various other instances of `assert not self.is_snapshot` to `assert self.is_mutable`.

pixeltable/catalog/view.py
aaron-siegel: Sure, "replica table" sounds a little more natural to me than "table replica", but I don't have super strong feelings here.

tests/share/test_packager.py

pixeltable/catalog/view.py
mkornacker: Really? To me "table replica" means "replica of table", etc.

mkornacker (approved):

pixeltable/catalog/view.py
aaron-siegel: Changed it to "{name}-replica" in the latest rev

PR #700, opened 2025-07-10
schema_overrides bugfixes

aaron-siegel:
Fixes a bug in the schema_overrides parameter for data imports: schema_overrides was still expecting old-style ColumnType parameters (`pxt.FloatType(nullable=True)`) rather than new-style type parameters (`pxt.Float`); this fixes it so that `schema_overrides` semantics align with `schema`.

Also improves the documentation for `create_table()`.

### DIFF
--- a/pixeltable/globals.py
+++ b/pixeltable/globals.py
@@ -44,7 +44,7 @@ def init(config_overrides: Optional[dict[str, Any]] = None) -> None:
 
 
 def create_table(
-    path_str: str,
+    path: str,
     schema: Optional[dict[str, Any]] = None,
     *,
     source: Optional[TableDataSource] = None,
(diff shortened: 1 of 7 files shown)


pixeltable/globals.py
```diff
@@ -146,7 +157,16 @@ def create_table(
         tds = UnkTableDataConduit(source, source_format=source_format, extra_fields=extra_args)
         tds.check_source_format()
         data_source = tds.specialize()
-        data_source.src_schema_overrides = schema_overrides
+        src_schema_overrides: dict[str, ts.ColumnType] = {}
+        if schema_overrides is None:
```
mkornacker: It's already {}. Simplify by dropping the if branch?

aaron-siegel: Done.

PR #701, opened 2025-07-14
Consolidate more MediaStore operations - part 3

jpeterson-pxt:
* Consolidate MediaStore operations. {media_3}
* Improve debug logging in FileCache.  
* Fix random python typing errors from Pylance.  
* Correct typo: properly case Windows (the operating system)

### DIFF
--- a/pixeltable/utils/media_store.py
+++ b/pixeltable/utils/media_store.py
@@ -39,15 +39,15 @@ class MediaStore:
         return parent / f'{tbl_id.hex}_{col_id}_{version}_{id_hex}{ext or ""}'
 
     @classmethod
-    def move_tmp_media_file(cls, file_url: Optional[str], tbl_id: UUID, col_id: int, version: int) -> Optional[str]:
-        """Move a tmp media file with given url into the MediaStore, and return new url
+    def move_tmp_media_file(cls, file_url: Optional[str], tbl_id: UUID, col_id: int, v_min: int) -> Optional[str]:
+        """If the given url is a tmp file, move it into the MediaStore, and return new url
         If it is not a tmp file in the tmp_dir, return the original url.
 
         Args:
             file_url: URL of the tmp media file to move
             tbl_id: Table ID to associate with the media file
             col_id: Column ID to associate with the media file
-            version: Version number to associate with the media file
+            v_min: v_min number to associate with the media file
 
         Returns:
             URL of the media final location of the file
@@ -67,12 +67,12 @@ class MediaStore:
         if not src_path.startswith(pxt_tmp_dir):
             # not a tmp file
             return file_url
-        new_file_url = cls.relocate_local_media_file(Path(src_path), tbl_id, col_id, version)
+        new_file_url = cls.relocate_local_media_file(Path(src_path), tbl_id, col_id, v_min)
         return new_file_url
 
     @classmethod
     def relocate_local_media_file(cls, src_path: Path, tbl_id: UUID, col_id: int, tbl_version: int) -> str:
-        """Relocate a random local file to the MediaStore, and return its new URL"""
+        """Relocate a local file to the MediaStore, and return its new URL"""
         dest_path = MediaStore.prepare_media_path(tbl_id, col_id, tbl_version, ext=src_path.suffix)
         src_path.rename(dest_path)
         return urllib.parse.urljoin('file:', urllib.request.pathname2url(str(dest_path)))


pixeltable/utils/media_store.py
```diff
@@ -36,53 +39,68 @@ def prepare_media_path(cls, tbl_id: UUID, col_id: int, version: int, ext: Option
...
+    def move_tmp_media_file(cls, file_url: Optional[str], tbl_id: UUID, col_id: int, version: int) -> Optional[str]:
         """Move a tmp media file with given url into the MediaStore, and return new url
         If it is not a tmp file in the tmp_dir, return the original url.
 
         Args:
             file_url: URL of the tmp media file to move
             tbl_id: Table ID to associate with the media file
             col_id: Column ID to associate with the media file
-            v_min: Version number to associate with the media file
+            version: Version number to associate with the media file
```
aaron-siegel: Why not use `v_min`? Was it not actually the row `v_min` in all cases?

pixeltable/utils/media_store.py
```diff
@@ -36,53 +39,68 @@ def prepare_media_path(cls, tbl_id: UUID, col_id: int, version: int, ext: Option
...
+        if not src_path.startswith(pxt_tmp_dir):
             # not a tmp file
             return file_url
-        new_file_url = cls.relocate_local_media_file(Path(file_path), tbl_id, col_id, v_min)
+        new_file_url = cls.relocate_local_media_file(Path(src_path), tbl_id, col_id, version)
         return new_file_url
 
     @classmethod
     def relocate_local_media_file(cls, src_path: Path, tbl_id: UUID, col_id: int, tbl_version: int) -> str:
+        """Relocate a random local file to the MediaStore, and return its new URL"""
```
aaron-siegel: random? It's not a random file, it's the file given by `src_path` (unless I'm missing something)

pixeltable/utils/media_store.py
```diff
@@ -36,53 +39,68 @@ def prepare_media_path(cls, tbl_id: UUID, col_id: int, version: int, ext: Option
         return parent / f'{tbl_id.hex}_{col_id}_{version}_{id_hex}{ext or ""}'
 
     @classmethod
-    def move_tmp_media_file(cls, file_url: Optional[str], tbl_id: UUID, col_id: int, v_min: int) -> Optional[str]:
+    def move_tmp_media_file(cls, file_url: Optional[str], tbl_id: UUID, col_id: int, version: int) -> Optional[str]:
```
mkornacker: v_min is correct here, version is less clear.

Please make an effort to conform to existing naming conventions and avoid gratuitous terminology changes.

jpeterson-pxt: Sorry, there are places where the same thing was described as "version".
I will fix this.
Shall I change all the other uses of "version" or "tbl_version" in this file to "v_min"?

pixeltable/utils/media_store.py
jpeterson-pxt: Shall I change all the references to "version" and "tbl_version" in this file to v_min?

pixeltable/utils/media_store.py

pixeltable/utils/media_store.py

jpeterson-pxt:
I am unclear about where to use "v_min" vs "version", "tbl_version", or some other variant. 
For the sake of consistency, should we purge the codebase of "version", "tbl_version", and the other variants and replace them all with "v_min"?

aaron-siegel:
> I am unclear about where to use "v_min" vs "version", "tbl_version", or some other variant. For the sake of consistency, should we purge the codebase of "version", "tbl_version", and the other variants and replace them all with "v_min"?

`v_min` is a row version, not a table version. A particular rowid can exist in different forms over different version ranges (e.g., in response to an `update()`); `v_min` and `v_max` capture a version range over which the row is constant. We should definitely not replace references to table versions with `v_min`; it only makes sense for rows.

jpeterson-pxt:
@aaron-siegel 
> `v_min` is a row version, not a table version. A particular rowid can exist in different forms over different version ranges (e.g., in response to an `update()`); `v_min` and `v_max` capture a version range over which the row is constant. We should definitely not replace references to table versions with `v_min`; it only makes sense for rows.

Makes sense, v_min is a row property. But now I'm genuinely curious why, in this case, where we are working with media files associated with a table/column/version, we want to say they are associated with a table/column/v_min?

aaron-siegel (approved):

PR #702, opened 2025-07-14
Fix for "zero timestamp" issue in table versions

aaron-siegel:
Fixes an issue where the created_at field in TableVersion metadata was inadvertently set to 0 after metadata was updated in response to UpdateStatus changes.

With this change, TableVersion keeps track of the original created_at timestamp and uses it to repopulate any md updates.


jpeterson-pxt:
I didn't see this, and put up a slightly different approach, where the creation timestamp is available at a slightly higher level in the creation tree.
https://github.com/pixeltable/pixeltable/pull/706

PR #706, opened 2025-07-15
Correct updated timestamp of version 0 of view.

jpeterson-pxt:


aaron-siegel:
This has already been fixed and can be closed, I think?

jpeterson-pxt:
No longer applicable.

PR #707, opened 2025-07-15
Replace uses of sql.text() in catalog.py with idiomatic SQLAlchemy

aaron-siegel:
This makes the code much easier to read IMO.

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -875,7 +875,7 @@ class Catalog:
         q = (
             sql.select(schema.TableVersion.md)
             .where(schema.TableVersion.tbl_id == tbl_id)
-            .where(sql.text(f"({schema.TableVersion.__table__}.md->>'version')::int = {md.version_md.version}"))
+            .where(schema.TableVersion.md['version'].cast(sql.Integer) == md.version_md.version)
         )
         existing_version_md_row = conn.execute(q).one_or_none()
         if existing_version_md_row is None:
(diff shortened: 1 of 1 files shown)


pixeltable/catalog/catalog.py
```diff
@@ -1474,11 +1451,10 @@ def _load_tbl_version(self, tbl_id: UUID, effective_version: Optional[int]) -> O
...
         if effective_version is None and not tbl_md.is_replica:
-            q = sql.select(schema.Table.id).where(
-                sql.text(
-                    f"md->'view_md'->'base_versions'->0->>0 = {tbl_id.hex!r} "
-                    "AND md->'view_md'->'base_versions'->0->>1 IS NULL"
-                )
+            q = (
+                sql.select(schema.Table.id)
+                .where(schema.Table.md['view_md']['base_versions'][0][0].astext == tbl_id.hex)
+                .where(schema.Table.md['view_md']['base_versions'][0][1].astext == None)
```
mkornacker, line 1783: Is the second astext needed? (Might result in more runtime conversions.)

pixeltable/catalog/catalog.py
```diff
@@ -1315,18 +1302,8 @@ def load_tbl_md(self, tbl_id: UUID, effective_version: Optional[int]) -> schema.
...
-                    f'{schema.TableVersion.__table__}.{schema.TableVersion.version.name}'
-                )
-            ).where(
-                sql.text(
-                    (
-                        f"({schema.Table.__table__}.md->>'current_schema_version')::int = "
-                        f'{schema.TableSchemaVersion.__table__}.{schema.TableSchemaVersion.schema_version.name}'
-                    )
-                )
+            q = q.where(schema.Table.md['current_version'].cast(sql.Integer) == schema.TableVersion.version).where(
```
mkornacker: Could you format them like you did below, with one .where per line? A bit easier to read.

mkornacker (approved):

pixeltable/catalog/catalog.py
aaron-siegel: Yes. It might be a bug in sqlalchemy, but comparing a JSONB to NULL gives an error.

pixeltable/catalog/catalog.py
aaron-siegel: ruff made this choice. I could get around it with 

`# fmt: skip`

directives but I generally like to avoid those? Is it worth it here?

aaron-siegel: Ah, I found an improvement that makes it even clearer and without using `fmt: skip`.

PR #710, opened 2025-07-15
Time travel

aaron-siegel:
Allows retrieving a handle to an older version of a table via the notation

```python
t = pxt.get_table('my_table:722')
```

Specific-version handles will always be instances of `View` and treated as snapshots.

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -1264,7 +1264,15 @@ class Catalog:
         tbl_md = schema.md_from_dict(schema.TableMd, tbl_record.md)
         version_md = schema.md_from_dict(schema.TableVersionMd, version_record.md)
 
-        ancestors: list[tuple[UUID, Optional[int]]] = []
+        # Reconstruct the TableVersionPath for the specified TableVersion. We do this by examining the created_at
+        # timestamps of this table and all its ancestors.
+        # TODO: Store the relevant TableVersionPaths in the database, so that we don't need to rely on timestamps
+        #     (which might be nondeterministic in the future).
+
+        # Build the list of ancestor versions, starting with the given table and traversing back to the base table.
+        # For each proper ancestor, we use the version whose created_at timestamp equals or most nearly precedes the
+        # given TableVersion's created_at timestamp.
+        ancestors: list[tuple[UUID, Optional[int]]] = [(tbl_id, effective_version)]
         if tbl_md.view_md is not None:
             for ancestor_id, _ in tbl_md.view_md.base_versions:
                 q = (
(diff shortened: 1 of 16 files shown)


pixeltable/catalog/catalog.py
```diff
@@ -698,6 +698,7 @@ def _acquire_dir_xlock(
 
     def get_dir_path(self, dir_id: UUID) -> Path:
         """Return path for directory with given id"""
+        assert isinstance(dir_id, UUID)
```
mkornacker, line 701: Doesn't mypy do that for us?

pixeltable/catalog/catalog.py
```diff
@@ -854,15 +857,19 @@ def _get_dir_entry(self, dir_id: UUID, name: str, lock_entry: bool = False) -> O
         )
         tbl_id = conn.execute(q).scalar_one_or_none()
         if tbl_id is not None:
-            if tbl_id not in self._tbls:
-                _ = self._load_tbl(tbl_id)
-            return self._tbls[tbl_id]
+            if version is not None:
+                return self._load_tbl_with_specified_version(tbl_id, version)
```
mkornacker: Let's expand _tbls to be able to cache these as well, no need to treat them differently (it's still a valid Table).

pixeltable/catalog/catalog.py
```diff
@@ -1529,6 +1536,67 @@ def _load_tbl(self, tbl_id: UUID) -> Optional[Table]:
         self._tbls[tbl_id] = view
         return view
 
+    def _load_tbl_with_specified_version(self, tbl_id: UUID, effective_version: int) -> Optional[Table]:
```
mkornacker: (Chatty name. load_tbl_at_version?)

The parameter should be 'version', we only use effective_version if it's Optional[int].

pixeltable/catalog/schema_object.py
```diff
@@ -18,6 +18,7 @@ class SchemaObject:
 
     def __init__(self, obj_id: UUID, name: str, dir_id: Optional[UUID]):
         # make these private so they don't collide with column names (id and name are fairly common)
+        assert dir_id is None or isinstance(dir_id, UUID), type(dir_id)
```
mkornacker, line 21: Just curious: did something happen to prompt these asserts?

pixeltable/catalog/view.py
```diff
@@ -255,10 +255,20 @@ def _get_snapshot_path(cls, tbl_version_path: TableVersionPath) -> TableVersionP
             base=cls._get_snapshot_path(tbl_version_path.base) if tbl_version_path.base is not None else None,
         )
 
+    def _is_anonymous_snapshot(self) -> bool:
+        """
+        Returns True if this is an unnamed snapshot (i.e., a snapshot that is not a separate schema object).
+        """
+        return self._snapshot_only and self._id == self._tbl_version_path.tbl_id
```
mkornacker, line 262: Hm, a bit weird. Could we make _id = None?

pixeltable/catalog/view.py
```diff
@@ -284,21 +294,22 @@ def delete(self, where: Optional[exprs.Expr] = None) -> UpdateStatus:
...
     def _get_base_table(self) -> Optional['Table']:
-        if self._tbl_version_path.base is None and not self._snapshot_only:
-            return None  # this can happen for a replica of a base table
-        # if this is a pure snapshot, our tbl_version_path only reflects the base (there is no TableVersion instance
-        # for the snapshot itself)
-        from pixeltable.catalog import Catalog
-
-        base_id = self._tbl_version_path.tbl_id if self._snapshot_only else self._tbl_version_path.base.tbl_id
-        with Catalog.get().begin_xact(tbl_id=base_id, for_write=False):
+        if self._tbl_version_path.tbl_id != self._id:
```
mkornacker, line 297: Isn't that just self._snapshot_only?

pixeltable/catalog/view.py
```diff
@@ -284,21 +294,22 @@ def delete(self, where: Optional[exprs.Expr] = None) -> UpdateStatus:
...
+            base_id = self._tbl_version_path.base.tbl_id
+        with catalog.Catalog.get().begin_xact(tbl_id=base_id, for_write=False):
             return catalog.Catalog.get().get_table_by_id(base_id)
 
     @property
     def _effective_base_versions(self) -> list[Optional[int]]:
         effective_versions = [tv.effective_version for tv in self._tbl_version_path.get_tbl_versions()]
-        if self._snapshot_only:
-            return effective_versions
+        if self._snapshot_only and not self._is_anonymous_snapshot():
```
mkornacker, line 311: Can't we treat an anonymous snapshot as a pure snapshot (and then not special-case it here and elsewhere)?

pixeltable/catalog/catalog.py
aaron-siegel: Not always :(

I put this in because one slipped through. I can take it back out, but I usually leave them in to make future debugging easier

pixeltable/catalog/schema_object.py
aaron-siegel: Yeah. I forget what though, but part of my debugging process was to put these asserts in, and when that happens I usually leave them in (see above)

pixeltable/catalog/view.py
aaron-siegel: Huh? Why?

pixeltable/catalog/view.py
aaron-siegel: Not anymore, because `self._snapshot_only == True` for anonymous snapshots too.

pixeltable/catalog/view.py
aaron-siegel: They're different b/c pure snapshots introduce a new schema object and anonymous snapshots don't. So the treatment of base tables is different. If this is controversial or isn't making sense let's discuss at standup

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
aaron-siegel: Done. Also refactored a bit to improve code consistency

mkornacker (approved):

PR #711, opened 2025-07-16
Move some column summary information into RowBuilder. {media_4}

jpeterson-pxt:


pixeltable/exprs/row_builder.py
```diff
@@ -234,6 +238,30 @@ def refs_unstored_iter_col(col_ref: ColumnRef) -> bool:
         for e in self.output_exprs:
             self._record_output_expr_id(e, e.slot_idx)
 
+        self.img_slot_idxs = None
+        self.media_slot_idxs = None
+        self.array_slot_idxs = None
+
+    def get_img_slot_idxs(self) -> list[int]:
```
aaron-siegel: Make these properties?

pixeltable/exprs/row_builder.py
```diff
@@ -234,6 +238,30 @@ def refs_unstored_iter_col(col_ref: ColumnRef) -> bool:
         for e in self.output_exprs:
             self._record_output_expr_id(e, e.slot_idx)
 
+        self.img_slot_idxs = None
```
aaron-siegel: Make these private with double-underscores to clearly indicate that the intended access is through the below interfaces

`self.__img_slot_idxs` etc.

pixeltable/exec/data_row_batch.py
```diff
@@ -34,31 +31,25 @@ def __init__(
...
                 num_rows = 0
-            self.rows = [
-                exprs.DataRow(
-                    row_builder.num_materialized, self.img_slot_idxs, self.media_slot_idxs, self.array_slot_idxs
-                )
-                for _ in range(num_rows)
-            ]
+            self.rows = [self.factory_row() for _ in range(num_rows)]
+
+    def factory_row(self) -> exprs.DataRow:
```
aaron-siegel: `make_row()` ?  Would be more consistent with our nomenclature elsewhere (`make_function()`, etc)

pixeltable/exec/data_row_batch.py
```diff
@@ -16,14 +16,11 @@ class DataRowBatch:
...
     row_builder: exprs.RowBuilder
-    img_slot_idxs: list[int]
-    media_slot_idxs: list[int]  # non-image media slots
-    array_slot_idxs: list[int]
     rows: list[exprs.DataRow]
 
     def __init__(
         self,
-        tbl: Optional[catalog.TableVersionHandle],
+        tbl: catalog.TableVersionHandle,
```
aaron-siegel: Pls put `assert tbl is not None` in the init body just to convince me that this is correct

aaron-siegel (changes requested):

pixeltable/exec/data_row_batch.py
mkornacker: It's not correct. See Sqlnode. Please revert.

pixeltable/exprs/row_builder.py
mkornacker: Nothing else in this class works like this right now, and I think we should keep it that way.

pixeltable/exprs/row_builder.py
```diff
@@ -234,6 +238,30 @@ def refs_unstored_iter_col(col_ref: ColumnRef) -> bool:
...
+        self.media_slot_idxs = None
+        self.array_slot_idxs = None
+
+    def get_img_slot_idxs(self) -> list[int]:
+        """Return indices of image slots"""
+        if self.img_slot_idxs is None:
```
mkornacker: This is most easily accomplished in __init__(). Same for all the other members.

mkornacker (changes requested):

pixeltable/exprs/row_builder.py
aaron-siegel: It's necessary if the accessors are properties (which they should be according to our criteria, I think?)

aaron-siegel: If they become simple member variables that are populated in `init` then the question is obviated

pixeltable/exprs/row_builder.py
jpeterson-pxt: Thanks. Because this was being created at a different time in the lifecycle of the row builder, I was unclear if the necessary information was present in the row builder at init() time. I will move the logic to init().

pixeltable/exec/data_row_batch.py
jpeterson-pxt: I will change to make_row(). 
After looking at it more closely, I think I will also move this method to the row builder, as it depends on nothing in the data row batch.

pixeltable/exprs/row_builder.py
jpeterson-pxt: I will move them to `init`.

pixeltable/exec/data_row_batch.py
jpeterson-pxt: Will revert.

pixeltable/exec/data_row_batch.py
jpeterson-pxt: Done

pixeltable/exec/data_row_batch.py
jpeterson-pxt: Done

pixeltable/exprs/row_builder.py
jpeterson-pxt: All addressed.

pixeltable/exprs/row_builder.py
jpeterson-pxt: Done.

pixeltable/exprs/row_builder.py
jpeterson-pxt: They have been removed.

aaron-siegel (approved):

mkornacker (approved):

PR #713, opened 2025-07-17
DataFrameResultSet.to_pydantic()

mkornacker:


pixeltable/dataframe.py
```diff
@@ -66,6 +72,43 @@ def _reverse(self) -> None:
     def to_pandas(self) -> pd.DataFrame:
         return pd.DataFrame.from_records(self._rows, columns=self._col_names)
 
+    def to_pydantic(self, model: type[pydantic.BaseModel]) -> Iterator[pydantic.BaseModel]:
```
aaron-siegel: This could be made friendlier via a generic:

```python
T = TypeVar('T', bound=pydantic.BaseModel)

def to_pydantic(self, model: type[T]) -> Iterator[T]
```

pixeltable/dataframe.py
```diff
@@ -66,6 +72,43 @@ def _reverse(self) -> None:
...
+        """
+        Convert the DataFrameResultSet to a list of Pydantic model instances.
+
+        Args:
+            model: A Pydantic model class.
+
+        Returns:
+            A list of Pydantic model instances, one for each row in the result set.
```
aaron-siegel: "An `Iterator` over Pydantic model instances"

Although one could make a case that it should return a list; `DataFrameResultSet` is an in-memory structure, and `to_pandas()` returns an in-memory structure, so it's a minor inconsistency to be returning an iterator here.

pixeltable/dataframe.py
```diff
@@ -66,6 +72,43 @@ def _reverse(self) -> None:
...
+
+        Raises:
+            Error: If the row data doesn't match the model schema.
+        """
+        model_fields = model.model_fields
+        model_config = getattr(model, 'model_config', {})
+        forbid_extra_fields = model_config.get('extra', 'ignore') == 'forbid'
```
aaron-siegel: The 'ignore' serves no purpose here (`model_config.get('extra')` would have identical behavior here)

tests/test_dataframe.py
```diff
@@ -867,3 +869,58 @@ def test_distinct(self, reset_db: None, reload_tester: ReloadTester) -> None:
...
+
+        with pytest.raises(pxt.Error, match=r'Required model fields .* are missing') as exc_info:
+            _ = list(t.select(t.i).collect().to_pydantic(TestModel))
+        assert extract_fields(exc_info) == {'s', 'f', 'b', 'ts', 'd'}
+        # case-sensitive field names
+        with pytest.raises(pxt.Error, match=r'Required model fields .* are missing') as exc_info:
+            _ = list(t.select(t.i, t.s, t.f, t.b, t.ts, D=t.d).collect().to_pydantic(TestModel))
+        assert extract_fields(exc_info) == {'d'}
+
+        with pytest.raises(pxt.Error, match='Input should be a valid integer'):
```
aaron-siegel: I'd include "Validation error" in the match string too

aaron-siegel (approved):

PR #714, opened 2025-07-18
PXT-667: Write media files to MediaStore with correct version.

jpeterson-pxt:


pixeltable/catalog/table_version.py
```diff
@@ -742,6 +742,7 @@ def add_columns(
         index_cols: dict[Column, tuple[index.BtreeIndex, Column, Column]] = {}
         all_cols: list[Column] = []
         for col in cols:
+            assert col.tbl == self
```
mkornacker: Not particularly useful here (see line 734).

pixeltable/store.py
```diff
@@ -123,15 +123,15 @@ def _rowid_join_predicate(self) -> sql.ColumnElement[bool]:
     def _storage_name(self) -> str:
         """Return the name of the data store table"""
 
-    def _move_tmp_media_file(self, file_url: Optional[str], col: catalog.Column, v_min: int) -> str:
-        return MediaStore.move_tmp_media_file(file_url, self.tbl_version.id, col.id, v_min)
+    def _move_tmp_media_file(self, file_url: Optional[str], col: catalog.Column) -> str:
```
mkornacker, line 126: assert col.tbl.id == self.tbl_version.id would make sense here

mkornacker (approved):

pixeltable/catalog/table_version.py
jpeterson-pxt: Yes, for sure. Put it there because Aaron has been concerned about these sorts of things in previous PR's :-)

pixeltable/store.py
jpeterson-pxt: Done

pixeltable/catalog/table_version.py
jpeterson-pxt: Done

PR #715, opened 2025-07-19
Consolidate ColumnMd operations into from_md() and to_md(). {media_5}

jpeterson-pxt:


pixeltable/catalog/table_version.py
```diff
@@ -1631,17 +1604,7 @@ def num_rowid_columns(self) -> int:
...
-            column_md[col.id] = schema.ColumnMd(
-                id=col.id,
-                col_type=col.col_type.as_dict(),
-                is_pk=col.is_pk,
-                schema_version_add=col.schema_version_add,
-                schema_version_drop=col.schema_version_drop,
-                value_expr=value_expr_dict,
-                stored=col.stored,
-            )
+            column_md[col.id] = col.to_md()
```
mkornacker: Remove this function and inline it at call sites.

pixeltable/catalog/column.py
```diff
@@ -126,6 +127,44 @@ def __init__(
         # computed cols also have storage columns for the exception string and type
         self.sa_cellmd_col = None
 
+    def to_md(self) -> schema.ColumnMd:
```
mkornacker: This is leaving out SchemaColumn. Either leave a todo that this should return tuple[columnmd, optional[schemacolumn]] or incorporate it (which requires adding Column.pos).

jpeterson-pxt: I added this functionality.

pixeltable/catalog/table_version.py
jpeterson-pxt: I added the functionality requested in the other comment instead.

pixeltable/catalog/column.py
```diff
@@ -126,6 +127,53 @@ def __init__(
         # computed cols also have storage columns for the exception string and type
         self.sa_cellmd_col = None
 
+    def to_md(self, pos: int) -> tuple[schema.ColumnMd, Optional[schema.SchemaColumn]]:
```
mkornacker: This still feels a bit confused to me. A column has an assigned position when it's a user-visible column. I think we should record that in Column, since we're recording everything else there as well.

If pos == None, it shouldn't have a SchemaColumn.

pixeltable/catalog/table_version.py
```diff
@@ -1629,21 +1582,19 @@ def num_rowid_columns(self) -> int:
...
-                id=col.id,
-                col_type=col.col_type.as_dict(),
-                is_pk=col.is_pk,
-                schema_version_add=col.schema_version_add,
-                schema_version_drop=col.schema_version_drop,
-                value_expr=value_expr_dict,
-                stored=col.stored,
-            )
-        return column_md
+        schema_col_md: dict[int, schema.SchemaColumn] = {}
```
mkornacker: You can do this in 3 lines, without a loop (using zip()). Which would have the advantage of being done mostly in C.

pixeltable/catalog/column.py
jpeterson-pxt: I have recoded this as requested.

pixeltable/catalog/table_version.py
jpeterson-pxt: I inlined the _create_column_md method into an existing loop. 
I also used "create_initial_md()" to replace a bunch of duplicated code. 
Perhaps this was something you were planning to do in the future?

mkornacker (approved):

PR #716, opened 2025-07-19
Extend FrameIterator to output all available frame attributes

mkornacker:
This behavior is controlled by kwarg all_frame_attrs and disabled by default to preserve compatibility with existing views created with this iterator.

This is undesirable (by default it should be on) and we should change that in a follow-up, once we settle on a good approach to iterator output schema evolution.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,7 @@
 # Global owners
 * @mkornacker @aaron-siegel
+pixeltable/catalog/** @mkornacker
 
 # Docs owners
-/docs/mintlify/** @mkornacker @aaron-siegel @pierrebrunelle @jacobweiss2305
-/docs/sample-apps/** @mkornacker @aaron-siegel @pierrebrunelle @jacobweiss2305
+/docs/mintlify/** @mkornacker @aaron-siegel @pierrebrunelle @goodlux
+/docs/sample-apps/** @mkornacker @aaron-siegel @pierrebrunelle @goodlux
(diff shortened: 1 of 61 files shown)


pixeltable/iterators/video.py
```diff
@@ -115,16 +126,28 @@ def input_schema(cls) -> dict[str, ts.ColumnType]:
...
-        return {
+        attrs: dict[str, ts.ColumnType] = {
             'frame_idx': ts.IntType(),
             'pos_msec': ts.FloatType(),
             'pos_frame': ts.IntType(),
-            'frame': ts.ImageType(),
-        }, ['frame']
+        }
+        if kwargs.get('all_frame_attrs'):
+            attrs.update(
```
aaron-siegel: I'd suggest we make this a single Json-typed field (`frame_attrs` or some such) rather than introduce 6 new columns. Otherwise the output table shape becomes pretty unwieldy.

Maybe even move `pos_msec` and `pos_frame` inside the json if `all_frame_attrs` is enabled?

aaron-siegel (changes requested):

pixeltable/iterators/video.py
```diff
@@ -164,8 +190,22 @@ def __next__(self) -> dict[str, Any]:
...
+                    'index': video_idx,
+                    'pts': frame.pts,
+                    'dts': frame.dts,
+                    'time': frame.time,
+                    'is_corrupt': frame.is_corrupt,
+                    'key_frame': frame.key_frame,
+                    'pict_type': frame.pict_type,
+                    'interlaced_frame': frame.interlaced_frame,
+                }
+                result.update({'frame_attrs': attrs})
```
aaron-siegel, line 206: `result['frame_attrs'] = attrs` is usually preferred when updating just a single dict value

aaron-siegel (approved):

PR #717, opened 2025-07-21
added initial mintlify integrated sdk docs and converted jupyter notebooks

goodlux:
This branch is for reviewing integration of SDK docs and Jupyter notebooks into the mintlify site.

To review:
```
git checkout feat_sdk_docs
cd docs/mintlify
mint dev
visit localhost:3000
```

- Added SDK tab (to replace API reference link)
- SDK docs with about 80% coverage of public functions, sourced from repo docstrings
- Version selection dropdown, for previous documentation on versions of the SDK
- Added a Jupyter Notebooks page group on the main page, to display converted Jupyter Notebooks

> [!WARNING] 
> These changes are not intended to be final, they are just for review to show what is possible and get feedback. Text and layout are still rough and not finalized. About 50% of the notebooks do not load, due to unresolved conversion error. Do not commit this PR, intended for review only! 

Comments / thoughts / critique are welcome!

--- 

Screenshot 2025-07-21 at 1 57 25 AM

Screenshot 2025-07-21 at 2 02 08 AM


aaron-siegel:
I haven't looked through the full PR in detail but as a first thought - are the mdx files here auto-generated from the notebooks and API docs? If so, best practice is to avoid checking generated files into the repo - instead we should be generating them as part of the docs build process. Otherwise it becomes harder to keep the notebooks & mdx docs in sync.

Also let's avoid checking `.DS_Store` into the repo (it can be added to `.gitignore`)

PR #718, opened 2025-07-21
Communicate Column spec for all MediaStore save and move operations

jpeterson-pxt:


pixeltable/utils/media_store.py
```diff
@@ -7,114 +9,138 @@
...
+        parent = cls._media_dir() / col.tbl.id.hex / id_hex[:2] / id_hex[:4]
         parent.mkdir(parents=True, exist_ok=True)
-        return parent / f'{tbl_id.hex}_{col_id}_{version}_{id_hex}{ext or ""}'
+        return parent / f'{col.tbl.id.hex}_{col.id}_{col.tbl.version}_{id_hex}{ext or ""}'
 
     @classmethod
-    def move_tmp_media_file(cls, file_url: Optional[str], tbl_id: UUID, col_id: int, v_min: int) -> Optional[str]:
-        """If the given url is a tmp file, move it into the MediaStore, and return new url
-        If it is not a tmp file in the tmp_dir, return the original url.
+    def is_tmp_url(cls, file_url: Optional[str]) -> tuple[bool, Path]:
```
aaron-siegel: This would be clearer to me if it were `resolve_tmp_url` and returned an `Optional[Path]` rather than a `tuple[bool, Path]`.

It's doing more work than `is_*` methods usually do.

pixeltable/utils/media_store.py
```diff
@@ -7,114 +9,138 @@
...
-        assert isinstance(image, PIL.Image.Image)
-        # Default to JPEG unless the image has a transparency layer (which isn't supported by JPEG).
-        # In that case, use WebP instead.
-        format = 'webp' if image.has_transparency_data else 'jpeg'
-        if not dest_path.endswith(f'.{format}'):
-            dest_path += f'.{format}'
-        image.save(dest_path, format=format)
-        url = urllib.parse.urljoin('file:', urllib.request.pathname2url(dest_path))
-        return dest_path, url
+        t_path = str(dest_path)
```
aaron-siegel: This may be cleaner than using `t_path`:

```python
if dest_path.suffix != f'.{format}':
    dest_path = dest_path.with_name(f'{dest_path.name}.{format}')
```

aaron-siegel (changes requested):

pixeltable/utils/media_store.py
jpeterson-pxt: Thanks, good idea. It is too big for an is_* method, but I didn't want to reparse the url if possible.

pixeltable/utils/media_store.py
jpeterson-pxt: Nice rewrite.
Done

PR #721, opened 2025-07-22
Use the method plan._insert_prefetch_node everywhere. {media_7}

jpeterson-pxt:


pixeltable/plan.py
```diff
@@ -789,15 +782,13 @@ def _is_contained_in(cls, l1: Iterable[exprs.Expr], l2: Iterable[exprs.Expr]) ->
 
     @classmethod
     def _insert_prefetch_node(
-        cls, tbl_id: UUID, row_builder: exprs.RowBuilder, input_node: exec.ExecNode
+        cls, tbl_id: UUID, expr_set: Iterable[exprs.Expr], input_node: exec.ExecNode
```
aaron-siegel: Please just call this `exprs` since it isn't actually a `set`.

jpeterson-pxt: I didn't want to overload the "exprs" identifier, and it can be a set. How about if I call it "expressions"?

Done

pixeltable/plan.py
```diff
@@ -789,15 +782,13 @@ def _is_contained_in(cls, l1: Iterable[exprs.Expr], l2: Iterable[exprs.Expr]) ->
 
     @classmethod
     def _insert_prefetch_node(
-        cls, tbl_id: UUID, row_builder: exprs.RowBuilder, input_node: exec.ExecNode
+        cls, tbl_id: UUID, expressions: Iterable[exprs.Expr], input_node: exec.ExecNode
     ) -> exec.ExecNode:
-        """Returns a CachePrefetchNode into the plan if needed, otherwise returns input"""
+        """Inserts a CachePrefetchNode into the plan if needed, otherwise returns input"""
```
mkornacker: "Return a .. if needed, otherwise ..."

mkornacker (approved):

pixeltable/plan.py

PR #724, opened 2025-07-23
Further simplify DataRowBatch. {media_9}

jpeterson-pxt:


pixeltable/exec/data_row_batch.py
```diff
@@ -17,22 +17,14 @@ class DataRowBatch:
...
-            self.rows = rows
-        else:
-            if num_rows is None:
-                num_rows = 0
-            self.rows = [self.row_builder.make_row() for _ in range(num_rows)]
+        self.rows = [] if rows is None else rows
 
-    def add_row(self, row: Optional[exprs.DataRow] = None) -> exprs.DataRow:
+    def add_row(self, row: exprs.DataRow) -> exprs.DataRow:
         if row is None:
```
aaron-siegel, line 28: This can't happen any more, according to the function signature.

aaron-siegel: Also, no point in returning the `DataRow` any more, might as well just make the function signature

`add_row(self, row: exprs.DataRow) -> None`

jpeterson-pxt: Good catch. However, I will fix the typing. It turns out there are uses of this method which look something like this:
`self.stateful_row_variable = ...add_row(self.stateful_row_variable)`

jpeterson-pxt: Done

aaron-siegel (approved):

PR #739, opened 2025-07-29
Introduce `TypedDict`s for user-facing table, dir, column, and index metadata

aaron-siegel:


pierrebrunelle:
@cursor review

pixeltable/catalog/dir.py
```diff
@@ -59,3 +60,27 @@ def _move(self, new_name: str, new_dir_id: UUID) -> None:
             )
         )
         Env.get().conn.execute(stmt, {'new_dir_id': new_dir_id, 'new_name': json.dumps(new_name), 'id': self._id})
+
+    def get_metadata(self) -> 'DirMetadata':
```
mkornacker: Do we want this? Not clear what it's good for, given that we don't return Dir instances in our public API (and even then you get that same info from a pxt.list_xx() operation).

pixeltable/catalog/table.py
```diff
@@ -80,49 +80,67 @@ def _move(self, new_name: str, new_dir_id: UUID) -> None:
...
+                name=col.name,
+                type_=col.col_type._to_str(as_schema=True),
+                version_added=col.schema_version_add,
+                is_stored=col.is_stored,
+                is_primary_key=col.is_pk,
+                media_validation=col.media_validation.name.lower() if col.media_validation is not None else None,  # type: ignore[typeddict-item]
+                computed_with=col.value_expr.display_str(inline=False) if col.value_expr is not None else None,
+            )
+        # Pure snapshots have no indices
+        indices = self._tbl_version.get().idxs_by_name.values() if self._tbl_version is not None else {}
```
mkornacker, line 112: type hint?

pixeltable/catalog/table.py
```diff
@@ -1705,3 +1723,66 @@ def __check_mutable(self, op_descr: str) -> None:
             raise excs.Error(f'{self._display_str()}: Cannot {op_descr} a snapshot.')
         if self._tbl_version_path.is_replica():
             raise excs.Error(f'{self._display_str()}: Cannot {op_descr} a {self._display_name()}.')
+
+
+class ColumnMetadata(TypedDict):
```
mkornacker, line 1731: Move into catalog/globals.py?

pixeltable/catalog/table.py
```diff
@@ -1705,3 +1723,66 @@ def __check_mutable(self, op_descr: str) -> None:
...
+    """The type specifier of the column."""
+    version_added: int
+    """The table version when this column was added."""
+    is_stored: bool
+    """`True` if this is a stored column; `False` if it is dynamically computed."""
+    is_primary_key: bool
+    """`True` if this column is part of the table's primary key."""
+    media_validation: Optional[Literal['on_read', 'on_write']]
+    """The media validation policy for this column."""
+    computed_with: Optional[str]
```
mkornacker, line 1746: I never particularly liked 'computed_with'. Do we want to call it something else?

pixeltable/catalog/table.py
```diff
@@ -1705,3 +1723,66 @@ def __check_mutable(self, op_descr: str) -> None:
...
+
+
+class IndexMetadata(TypedDict):
+    """Metadata for a column of a Pixeltable table."""
+
+    name: str
+    """The name of the index."""
+    columns: list[str]
+    """The table columns that are indexed."""
+    metric: Literal['cosine', 'ip', 'l2']
```
mkornacker, line 1763: This is really specific to embedding indices. Should we move that into a generic 'parameters: dict' field, and then have typeddicts for something like EmbeddingIndexParameters?

mkornacker (approved):

pixeltable/catalog/dir.py
aaron-siegel: Removed.

pixeltable/catalog/table.py
aaron-siegel: shouldn't be necessary as it's a one-and-done assignment? and anyway its type is some weird iterator type (via `values()`)

pixeltable/catalog/table.py
aaron-siegel: Per conversation today, keeping where it is for now

pixeltable/catalog/table.py
aaron-siegel: Let's discuss this separately, since I'm not sure offhand what else to call it :)

pixeltable/catalog/table.py
aaron-siegel: Done.

PR #741, opened 2025-07-30
PXT-671: fixes to RateLimitsScheduler

mkornacker:
The current rate limits scheduler discards the header-provided resource utilization information when it receives 429 (rate limit error) exceptions.

Tests with an Openai account that has standard (low) rate limits of 200k TPM confirms that with the changes, the scheduler correctly holds off on submitting requests after a 429 error until the indicated reset time, even when subsequent (out-of-order) responses indicate that more resources are available. The results is far fewer retries.


pixeltable/func/signature.py
```diff
@@ -251,6 +252,8 @@ def create_parameters(
         for idx, param in enumerate(py_params):
             if is_cls_method and idx == 0:
                 continue  # skip 'self' or 'cls' parameter
+            if param.name.startswith('_'):
+                continue  # skip system parameters
```
aaron-siegel, line 277: I'd suggest we check this against a whitelist of allowed system parameters and raise an exception otherwise.

Two reasons:
- So we don't swallow the parameter if someone misspells "runtime_ctx"
- So the user will get a clear error message if they innocently try to define a UDF with a parameter name starting with `_`

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -97,7 +99,7 @@ async def _main_loop(self) -> None:
                 continue
 
             # check rate limits
-            _logger.debug(f'checking rate limits for {self.resource_pool}')
+            # _logger.debug(f'checking rate limits for {self.resource_pool}')
```
aaron-siegel: Commented out intentionally?

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -158,16 +170,22 @@ def _get_request_resources(self, request: FnCallArgs) -> dict[str, int]:
 
     def _check_resource_limits(self, request_resources: dict[str, int]) -> Optional[env.RateLimitInfo]:
         """Returns the most depleted resource, relative to its limit, or None if all resources are within limits"""
-        candidates: list[tuple[env.RateLimitInfo, float]] = []  # (info, relative usage)
+        candidates: list[tuple[env.RateLimitInfo, float]] = []  # (info, relative remaining)
         for resource, usage in request_resources.items():
             # 0.05: leave some headroom, we don't have perfect information
```
aaron-siegel: This comment is misplaced now?

pixeltable/func/callable_function.py
```diff
@@ -48,6 +49,13 @@ def __init__(
         self.__doc__ = self.py_fns[0].__doc__
         super().__init__(signatures, self_path=self_path, is_method=is_method, is_property=is_property)
 
+        if not self.is_polymorphic:
+            # check for _runtime_ctx parameter in our py_fn
+            py_sig = inspect.signature(self.py_fn)
+            self.accepts_runtime_ctx = '_runtime_ctx' in py_sig.parameters
+        else:
+            self.accepts_runtime_ctx = False
```
aaron-siegel: Why do we need this limitation? We could easily check whether all signatures have `_runtime_ctx` as a parameter.

Even better would be to make `accepts_runtime_ctx` a property of the Pixeltable `Signature`, rather than the `CallableFunction`. That's really what it is. Then it would carry over to a particular `FunctionCall` just if it were present in the relevant signature for that `FunctionCall`.

pixeltable/func/signature.py
mkornacker: Done

pixeltable/exec/expr_eval/schedulers.py
mkornacker: Done

pixeltable/exec/expr_eval/schedulers.py
mkornacker: Done

aaron-siegel (approved):

PR #746, opened 2025-08-01
Performance test for chat completion integrations

mkornacker:


tool/perftest_providers.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+class ProviderConfig(TypedDict):
+    """Configuration for a provider."""
+
+    function: pxt.Function
+    default_model: str
+    prompt_type: Literal['messages', 'string']
+    prompt_udf: pxt.Function
+    params: Dict[str, Any]
```
aaron-siegel: Use lowercase `Dict` and `List` (the uppercase variants are obsolete in Python 3.9+)

tool/perftest_providers.py
```diff
@@ -0,0 +1,261 @@
...
+        {
+            'role': 'user',
+            'content': f'Generate a single sentence that uses both of these words: {word1} and {word2}. '
+            f'The sentence should be natural and make sense.',
+        },
+    ]
+
+
+@pxt.udf
+def create_claude_prompt(word1: str, word2: str) -> List[Dict[str, str]]:
```
aaron-siegel: Isn't this identical to `create_simple_messages_prompt()`?

tool/perftest_providers.py
```diff
@@ -0,0 +1,261 @@
...
+    system_prompt: Optional[str]
+
+
+def get_random_words(wordlist: List[str], k: int = 2) -> List[str]:
+    """Get k random words from the wordlist."""
+    return random.sample(wordlist, k=k)
+
+
+@pxt.udf
+def create_chatgpt_prompt(word1: str, word2: str) -> List[Dict[str, str]]:
```
aaron-siegel: Instead of naming these by provider, maybe the following would be clearer (since all of them use multiple providers)?

```
completion_prompt()
chat_completion_prompt()
chat_completion_prompt_with_system()
```

tool/perftest_providers.py
```diff
@@ -0,0 +1,261 @@
...
+from typing import Any, Callable, Dict, List, Literal, Optional, TypedDict
+
+import pixeltable as pxt
+import pixeltable.functions as pxtf
+
+
+class ProviderConfig(TypedDict):
+    """Configuration for a provider."""
+
+    function: pxt.Function
```
aaron-siegel: `udf`? (or `api_udf` or some such) (since the below is called `prompt_udf`)

tool/perftest_providers.py
```diff
@@ -0,0 +1,261 @@
...
+    # Load wordlist
+    with open('/usr/share/dict/american-english', encoding='utf-8') as f:
+        wordlist = [word.strip() for word in f]
+
+    t.add_computed_column(prompt=provider_config['prompt_udf'](t.word1, t.word2))
+
+    # Create provider-specific computed column
+    if provider_config['prompt_type'] == 'messages':
+        # Handle system prompt for providers that support it
+        if 'system_prompt' in provider_config and args.provider == 'anthropic':
```
aaron-siegel: Why is the second condition necessary?

tool/perftest_providers.py
```diff
@@ -0,0 +1,261 @@
...
+import argparse
+import random
+from datetime import datetime
+from typing import Any, Callable, Dict, List, Literal, Optional, TypedDict
+
+import pixeltable as pxt
+import pixeltable.functions as pxtf
+
+
+class ProviderConfig(TypedDict):
```
aaron-siegel: NamedTuple might have been more logical here (and more convenient since then you can use `provider_config.params` instead of `provider_config['params']` etc.) but maybe not worth refactoring at this point.

tool/perftest_providers.py
```diff
@@ -0,0 +1,261 @@
...
+    ]
+
+
+def create_provider_configs() -> Dict[str, ProviderConfig]:
+    """Create configuration for each supported provider."""
+    return {
+        'openai': {
+            'function': pxtf.openai.chat_completions,
+            'default_model': 'gpt-4o-mini',
+            'prompt_type': 'messages',
```
aaron-siegel: Calling these `completion` and `chat_completion` would better align with LLM terminology.

tool/perftest_providers.py
```diff
@@ -0,0 +1,261 @@
...
+                    messages=t.prompt, model=model, model_kwargs=provider_config['params']
+                )
+            )
+    else:  # prompt_type == 'string'
+        t.add_computed_column(
+            response=provider_config['function'](t.prompt, model=model, model_kwargs=provider_config['params'])
+        )
+
+    # Generate rows
+    rows = ({'word1': w1, 'word2': w2} for _ in range(args.n) for w1, w2 in [random.sample(wordlist, k=2)])
```
aaron-siegel: I'm pretty confused here, won't this result in w1 and w2 being set to a list? Since `random.sample()` is itself wrapped in an outer collection?

tool/perftest_providers.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # Insert and time the operation
+    start = datetime.now()
+    status = t.insert(rows, on_error='ignore')
```
aaron-siegel, line 189: Since `on_error='ignore'`, check for error count at the end and fail the script if it exceeds a certain threshold. (I'm thinking of the case where the API is down or has introduced some breaking change, and every row generates an error and we swallow them all and report success in record time.)

tool/perftest_providers.py
mkornacker: Made it a dataclass.

tool/perftest_providers.py
mkornacker: Done.

tool/perftest_providers.py
mkornacker: Done

tool/perftest_providers.py
mkornacker: Removed prompt_type and simplified the logic.

tool/perftest_providers.py
mkornacker: Done.

tool/perftest_providers.py
mkornacker: Simplified the logic.

tool/perftest_providers.py
mkornacker: Done.

tool/perftest_providers.py
mkornacker: Done.

tool/perftest_providers.py
mkornacker: Done.

tool/perftest_providers.py
aaron-siegel: Note: I'm not sure this change made it in (I don't see it reflected in the diff, but I do see the other changes)

tool/perftest_providers.py
```diff
@@ -0,0 +1,199 @@
...
+
+    print(f'Using provider: {args.provider}')
+    print(f'Using model: {model}')
+    print(f'Generating {args.n} sentences...')
+
+    t = pxt.create_table('sentence_tbl', {'word1': pxt.String, 'word2': pxt.String}, if_exists='replace')
+    t.add_computed_column(prompt=provider_config.prompt_udf(t.word1, t.word2))
+    t.add_computed_column(response=provider_config.udf(t.prompt, model=model, **provider_config.kwargs))
+
+    rows = ({'word1': words[0], 'word2': words[1]} for words in (random.sample(wordlist, k=2) for _ in range(args.n)))
```
aaron-siegel: Just a cute little syntactic convenience you can use here,
```
rows = ({'word1': word1, 'word2': word2} for word1, word2 in (random.sample(wordlist, k=2) for _ in range(args.n)))
```
you avoid the awkward use of `0` and `1` indices which don't really mean anything in this (tuple) context

aaron-siegel (approved):

PR #747, opened 2025-08-01
Bugfixes related to missing dependencies

aaron-siegel:
Fixes several bugs related to Pixeltable's handling of missing dependencies.
- Allow `DocumentSplitter.output_schema()` to resolve even if necessary dependencies are missing
- Fail gracefully with a UDF warning if `call_return_type()` fails due to missing dependencies (previously we just swallowed the `ImportError`)
- If `call_return_type` gives a more general type than the existing `return_type` of a `FunctionCall`, prefer the more specialized type

### DIFF
--- a/pixeltable/exprs/function_call.py
+++ b/pixeltable/exprs/function_call.py
@@ -115,6 +115,7 @@ class FunctionCall(Expr):
         self._validation_error = validation_error
 
         if validation_error is not None:
+            self.bound_idxs = {}
             self.resource_pool = None
             return
 
@@ -300,8 +301,11 @@ class FunctionCall(Expr):
         """
         res = super().substitute(spec)
         assert res is self
-        self.return_type = self.fn.call_return_type(self.bound_args)
-        self.col_type = self.return_type
+        # See if we can update the return_type with something more specific.
+        crt = self.fn.call_return_type(self.bound_args)
+        if self.return_type.is_supertype_of(crt):
+            self.return_type = crt
+            self.col_type = self.return_type
         return self
 
     def update(self, data_row: DataRow) -> None:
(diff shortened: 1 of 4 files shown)


pixeltable/exprs/function_call.py
```diff
@@ -300,8 +301,11 @@ def substitute(self, spec: dict[Expr, Expr]) -> Expr:
         """
         res = super().substitute(spec)
         assert res is self
-        self.return_type = self.fn.call_return_type(self.bound_args)
-        self.col_type = self.return_type
+        # See if we can update the return_type with something more specific.
+        crt = self.fn.call_return_type(self.bound_args)
+        if self.return_type.is_supertype_of(crt):
```
mkornacker: Why is this needed? Shouldn't the result of call_return_type() be at least as specific as self.return_type?

pixeltable/iterators/document.py
```diff
@@ -213,12 +213,6 @@ def output_schema(cls, *args: Any, **kwargs: Any) -> tuple[dict[str, ColumnType]
             if kwargs.get('limit') is None:
                 raise Error('limit is required with "token_limit"/"char_limit" separators')
 
-        # check dependencies at the end
```
mkornacker, line 216: So where is the importerror showing up now?

mkornacker (approved):

pixeltable/iterators/document.py
aaron-siegel: When you insert new data or otherwise cause the iterator to be invoked.

pixeltable/exprs/function_call.py
aaron-siegel: You're right, it should be, *as long as the codebase is stable*. But if (say) something changes in the implementation of call_return_type() then that's not guaranteed.

It took me some time to reason through all this, but I think I now understand it. I changed the condition here to `if self.is_valid`, which better captures what's actually going on, but it still isn't perfect. More details are in the new code comment. Getting the edge cases exactly right would involve some changes to EmbeddingIndex that require a migration, which probably isn't worth doing right now, since the PR as written fixes the proximate issue (graceful loading of metadata in cases where libraries needed by call_return_type() are missing).

PR #748, opened 2025-08-01
Group local media file operations into a MediaStore or TempStore class

jpeterson-pxt:
The MediaStore interface is used to write, move, and delete files containing media data persisted in a local filesystem directory.
The TempStore interface provides similar functionality for temporary files and directories.
The TempStore acts as a stateless interface to the Env tmp_dir.

Both classes provide some interfaces for summarizing the contents of the stores. These methods are only intended to be used with Pixeltable unit tests in specific circumstances.

### DIFF
--- a/pixeltable/share/publish.py
+++ b/pixeltable/share/publish.py
@@ -128,7 +128,7 @@ def pull_replica(dest_path: str, src_tbl_uri: str) -> pxt.Table:
     if parsed_location.scheme == 's3':
         bundle_path = _download_bundle_from_s3(parsed_location, bundle_filename)
     elif parsed_location.scheme == 'https':
-        bundle_path = Path(Env.get().create_tmp_path())
+        bundle_path = TempStore.create_path()
         _download_from_presigned_url(url=parsed_location.geturl(), output_path=bundle_path)
     else:
         raise excs.Error(f'Unexpected response from server: unsupported bundle uri: {bundle_uri}')


jpeterson-pxt:
This PR was stuck in the test_fiftyone.py::test_export_images test on all platforms. 
I need to understand why this might be.

aaron-siegel (changes requested):
Main thought: I found the way the `MediaStoreFile` pattern to be a bit confusing. I suggest the following scheme, which is more logical to me: a base class called `FileStore` that holds the `__base_dir` and implements the shared operations, with `MediaStore` and `TempStore` as subclasses of `FileStore`. This would also, I believe, reduce the amount of boilerplate code.

While we're at it, I'd suggest we also make `MediaStore` and `TempStore` into singleton instances rather than  using `@classmethod`s for everything, to align with the pattern used elsewhere (`Env`, `Catalog`, `Config`).

pixeltable/utils/media_store.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
```
aaron-siegel, line 1: Rename the module to `file_store.py` or some such now, since it's more general?

pixeltable/utils/media_store.py
```diff
@@ -19,49 +20,88 @@
...
     @classmethod
-    def _media_dir(cls) -> Path:
+    def get(cls, base_uri: Optional[Path] = None) -> MediaStore:
+        """Get a MediaStore instance for the given base URI, or the environment's media_dir if None."""
+        if base_uri is None:
+            return MediaStore(env.Env.get().media_dir)
+        raise NotImplementedError
+
+    @property
+    def _media_dir(self) -> Path:
```
aaron-siegel: Since it's only accessed internally in the class, do we really need this? (vs. just referencing `__base_dir` directly)

pixeltable/utils/media_store.py
```diff
@@ -170,3 +193,50 @@ def stats(cls) -> list[tuple[UUID, int, int, int]]:
...
+    @classmethod
+    def _tmp_dir(cls) -> Path:
+        """Returns the path to the temporary directory where files are stored."""
+        from pixeltable import env
+
+        return env.Env.get().tmp_dir
+
+    @classmethod
+    def count(cls, tbl_id: Optional[UUID] = None) -> int:
+        return MediaStore(cls._tmp_dir()).count(tbl_id)
```
aaron-siegel, line 212: `MediaStore(cls._tmp_dir())` appears repeatedly in the code; it'd be cleaner to make this into a property

pixeltable/utils/media_store.py
```diff
@@ -170,3 +193,50 @@ def stats(cls) -> list[tuple[UUID, int, int, int]]:
...
         result.sort(key=lambda e: e[3], reverse=True)
         return result
+
+
+class TempStore:
+    """
+    A temporary store for files of data that are not yet persisted to their destination(s).
+    A destination is typically either a MediaStore (local persisted files) or a cloud object store.
+
+    The TempStore class has no internal state, it provides functionality to manage temporary files
```
aaron-siegel: Nitpick: comma splice (the comma should be a semicolon)

pixeltable/utils/media_store.py
```diff
@@ -19,49 +20,88 @@
...
+        assert isinstance(file_data, bytes)
+        with open(dest_path, 'wb') as f:
+            f.write(file_data)
+            f.flush()  # Ensures Python buffers are written to OS
+            os.fsync(f.fileno())  # Forces OS to write to physical storage
+        return dest_path
 
     @classmethod
-    def _prepare_media_path(cls, col: Column, ext: Optional[str] = None) -> Path:
+    def _save_pil_image_file(cls, image: PIL.Image.Image, dest_path: Path, format: Optional[str]) -> Path:
```
mkornacker, line 61: Is there really a need to special-case images here and not in the caller?

pixeltable/utils/media_store.py
```diff
@@ -170,3 +193,50 @@ def stats(cls) -> list[tuple[UUID, int, int, int]]:
...
+
+        return env.Env.get().tmp_dir
+
+    @classmethod
+    def count(cls, tbl_id: Optional[UUID] = None) -> int:
+        return MediaStore(cls._tmp_dir()).count(tbl_id)
+
+    @classmethod
+    def resolve_url(cls, file_url: Optional[str]) -> Optional[Path]:
+        return MediaStore(cls._tmp_dir()).resolve_url(file_url)
```
mkornacker, line 216: Very confusing, that it just wraps MediaStore. Why does this exist?

mkornacker (approved):

pixeltable/utils/media_store.py

pixeltable/utils/media_store.py

pixeltable/utils/media_store.py
jpeterson-pxt: This determines if a file_url is in the TempStore.
Common functionality for determining if something is in a store.

PR #749, opened 2025-08-01
Move stored_img_cols from ExecNode To RowBuilder, add stored_media_cols

jpeterson-pxt:
The member stored_img_cols is moved from ExecNode to RowBuilder, and filled during RowBuilder construction.
The member stored_media_cols is added, with a similar policy.

### DIFF
--- a/pixeltable/exprs/row_builder.py
+++ b/pixeltable/exprs/row_builder.py
@@ -86,6 +86,8 @@ class RowBuilder:
     img_slot_idxs: list[int]  # Indices of image slots
     media_slot_idxs: list[int]  # Indices of non-image media slots
     array_slot_idxs: list[int]  # Indices of array slots
+    stored_img_cols: list[exprs.ColumnSlotIdx]
+    stored_media_cols: list[exprs.ColumnSlotIdx]
 
     @dataclass
     class EvalCtx:
@@ -242,6 +244,10 @@ class RowBuilder:
             e.slot_idx for e in self.unique_exprs if e.col_type.is_media_type() and not e.col_type.is_image_type()
         ]
         self.array_slot_idxs = [e.slot_idx for e in self.unique_exprs if e.col_type.is_array_type()]
+
+        stored_col_info = self.output_slot_idxs()
+        self.stored_img_cols = [info for info in stored_col_info if info.col.col_type.is_image_type()]
+        self.stored_media_cols = [info for info in stored_col_info if info.col.col_type.is_media_type()]
 
     def add_table_column(self, col: catalog.Column, slot_idx: int) -> None:
         """Record a column that is part of the table row"""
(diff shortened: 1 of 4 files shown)


pixeltable/exprs/row_builder.py
```diff
@@ -499,3 +499,15 @@ def store_column_names(self) -> list[str]:
     def make_row(self) -> exprs.DataRow:
         """Creates a new DataRow with the current row_builder's configuration."""
         return exprs.DataRow(self.num_materialized, self.img_slot_idxs, self.media_slot_idxs, self.array_slot_idxs)
+
+    def stored_img_cols(self) -> list[exprs.ColumnSlotIdx]:
```
aaron-siegel: Make these properties?

jpeterson-pxt: I could, I left them as methods because I judged that they do entail some computation per Marcel's criteria.

pixeltable/plan.py
```diff
@@ -1054,6 +1045,5 @@ def create_add_column_plan(cls, tbl: catalog.TableVersionPath, col: catalog.Colu
         plan.ctx.num_computed_exprs = len(computed_exprs)  # we are adding a computed column, so we need to evaluate it
 
         # we want to flush images
-        if col.is_computed and col.is_stored and col.col_type.is_image_type():
-            plan.set_stored_img_cols(row_builder.output_slot_idxs())
+        plan.set_stored_img_cols(row_builder.stored_img_cols())
```
aaron-siegel: Are we sure this is correct? All the others are logical no-ops (straight refactorings) but the logic changes on this one.

pixeltable/exprs/row_builder.py
aaron-siegel: They're doing very little computation (a single list comprehension and no db access). In my view it's a clear cut property but happy to let @mkornacker weigh in.

pixeltable/plan.py
jpeterson-pxt: It is slightly different, but the only consequence is that we will check some additional non-computed stored columns and write their media to the MediaStore if necessary. This may be more correct than what is there.

pixeltable/exprs/row_builder.py
jpeterson-pxt: Converted all to properties.

aaron-siegel (approved):

pixeltable/exprs/row_builder.py
```diff
@@ -499,3 +499,17 @@ def store_column_names(self) -> list[str]:
     def make_row(self) -> exprs.DataRow:
         """Creates a new DataRow with the current row_builder's configuration."""
         return exprs.DataRow(self.num_materialized, self.img_slot_idxs, self.media_slot_idxs, self.array_slot_idxs)
+
+    @property
+    def stored_img_cols(self) -> list[exprs.ColumnSlotIdx]:
+        """Returns the list of stored image columns"""
+        stored_col_info = self.output_slot_idxs()
+        stored_img_col_info = [info for info in stored_col_info if info.col.col_type.is_image_type()]
```
mkornacker: Why not just make that a member var that's set in the c'tor.

pixeltable/exprs/row_builder.py
```diff
@@ -499,3 +499,17 @@ def store_column_names(self) -> list[str]:
...
+        return stored_img_col_info
+
+    @property
+    def stored_media_cols(self) -> list[exprs.ColumnSlotIdx]:
+        """Returns the list of stored media columns"""
+        stored_col_info = self.output_slot_idxs()
+        stored_media_col_info = [info for info in stored_col_info if info.col.col_type.is_media_type()]
```
mkornacker: Ditto.

pixeltable/plan.py
```diff
@@ -406,6 +403,7 @@ def create_insert_plan(
                 ignore_errors=ignore_errors,
             )
         )
+        plan.set_stored_img_cols(row_builder.stored_img_cols)
```
mkornacker: Why not remove ExecNode.stored_img_cols altogether, it already has a row_builder.

mkornacker (changes requested):

pixeltable/plan.py

pixeltable/exprs/row_builder.py
jpeterson-pxt: Done

pixeltable/exprs/row_builder.py
jpeterson-pxt: Done

PR #760, opened 2025-08-07
Support for pydantic models in Table.insert()

mkornacker:
Missing yet:
- handling of media data or ndarrays in model fields
- more extensive tests for nested models

### DIFF
--- a/pixeltable/catalog/insertable_table.py
+++ b/pixeltable/catalog/insertable_table.py
@@ -148,6 +148,7 @@ class InsertableTable(Table):
         with Catalog.get().begin_xact(tbl=self._tbl_version_path, for_write=True, lock_mutable_tree=True):
             table = self
 
+            # TODO: unify with TableDataConduit
             if source is not None and isinstance(source, Sequence) and isinstance(source[0], pydantic.BaseModel):
                 status = self._insert_pydantic(
                     cast(Sequence[pydantic.BaseModel], source),  # needed for mypy


pixeltable/catalog/insertable_table.py
```diff
@@ -123,9 +126,19 @@ def insert(
         self, /, *, on_error: Literal['abort', 'ignore'] = 'abort', print_stats: bool = False, **kwargs: Any
     ) -> UpdateStatus: ...
 
+    @overload
```
aaron-siegel: Why is this needed, versus simply modifying the signature of the first `@overload` in the appropriate way?

In fact, if you augment `TableDataSource` as noted below, you shouldn't need to make any changes to the overload signatures at all.

pixeltable/catalog/insertable_table.py
```diff
@@ -123,9 +126,19 @@ def insert(
...
+        rows: Sequence[pydantic.BaseModel],
+        /,
+        *,
+        on_error: Literal['abort', 'ignore'] = 'abort',
+        print_stats: bool = False,
+    ) -> UpdateStatus: ...
+
+    def insert(
+        self,
+        source: Optional[TableDataSource | Sequence[pydantic.BaseModel]] = None,
```
aaron-siegel: No need to add an additional `Union` here; just extend the `TableDataSource` type to take a `Sequence[pydantic.BaseModel]` as a union parameter.

pixeltable/catalog/insertable_table.py
```diff
@@ -137,8 +150,23 @@ def insert(
...
         from pixeltable.io.table_data_conduit import UnkTableDataConduit
 
+        if source is not None and isinstance(source, Sequence) and len(source) == 0:
+            raise excs.Error('Cannot insert an empty sequence')
+        fail_on_exception = OnErrorParameter.fail_on_exception(on_error)
+
         with Catalog.get().begin_xact(tbl=self._tbl_version_path, for_write=True, lock_mutable_tree=True):
             table = self
+
+            if source is not None and isinstance(source, Sequence) and isinstance(source[0], pydantic.BaseModel):
```
aaron-siegel, line 152: I'd have expected Pydantic to go through the existing mechanism for data importation. i.e., add a `PydanticTableDataConduit` class in `table_data_conduit.py`, analogous to `RowDataTableDataConduit`, and put the validation and importation logic there. Is there a particular reason for bypassing it here and special-casing Pydantic? In particular why is a sequence of Pydantic instances treated differently from a sequence of dictionaries (in terms of the code architecture)? I find it awkward to have the special case of Pydantic live directly in `insertable_table.py` when everything else is in `pixeltable.io`.

pixeltable/catalog/table.py
```diff
@@ -1252,10 +1253,20 @@ def insert(
         self, /, *, on_error: Literal['abort', 'ignore'] = 'abort', print_stats: bool = False, **kwargs: Any
     ) -> UpdateStatus: ...
 
+    @overload
```
aaron-siegel: per above comment, it shouldn't be necessary to have an additional `@overload` signature.

pixeltable/catalog/table.py
```diff
@@ -1290,6 +1312,7 @@ def insert(
 
         Args:
             source: A data source from which data can be imported.
+            rows: (if inserting Pydantic models) An iterable of Pydantic BaseModel instances to insert.
```
aaron-siegel: No need for an additional parameter here

pixeltable/catalog/insertable_table.py
```diff
@@ -184,32 +211,88 @@ def insert_table_data_source(
...
+        model_class = type(rows[0])
+        self._validate_pydantic_model(model_class)
+        pxt_rows = [row.model_dump() for row in rows]
+
+        # explicitly check that all required columns are present and non-None in the rows,
+        # because we ignore nullability when validating the pydantic model
+        reqd_col_names = [col.name for col in self._tbl_version_path.columns() if col.is_required_for_insert]
+        for i, pxt_row in enumerate(pxt_rows):
+            if type(rows[i]) is not model_class:
+                raise excs.Error(f'Expected {model_class.__name__} instance, got {rows[i]!r}')
```
aaron-siegel: Use `model_class.__name__!r` ? Also, do you mean `type(rows[i])!r` here?

pixeltable/catalog/insertable_table.py
```diff
@@ -184,32 +211,88 @@ def insert_table_data_source(
...
+            for col_name in reqd_col_names:
+                if pxt_row.get(col_name) is None:
+                    raise excs.Error(f'Missing required column {col_name!r} in {rows[i]!r}')
```
aaron-siegel: Printing the entire `rows[i]!r` could give a huge error message. Maybe better to say `in row {i}`? The input is necessarily indexed since it's a sequence.

pixeltable/catalog/insertable_table.py
```diff
@@ -184,32 +211,88 @@ def insert_table_data_source(
...
+        """
+        assert isinstance(model, type) and issubclass(model, pydantic.BaseModel)
+
+        schema = self._get_schema()
+        required_cols = set(self._tbl_version.get().get_required_col_names())
+        computed_cols = set(self._tbl_version.get().get_computed_col_names())
+        model_fields = model.model_fields
+        model_field_names = set(model_fields.keys())
+
+        missing_required = required_cols - model_field_names
```
aaron-siegel, line 253: Just an observation, since we're using sets, this will cause the error message to be nondeterministic. Not the end of the world, but, we could just as easily preserve the deterministic ordering of columns this way:
```
required_cols = self._tbl_version.get().get_required_col_names()
missing_required = [name for name in required_cols if name not in model_field_names]
```

pixeltable/catalog/insertable_table.py
```diff
@@ -184,32 +211,88 @@ def insert_table_data_source(
...
+        if missing_required:
+            raise excs.Error(
+                f'Pydantic model {model.__name__} is missing required columns: '
+                f'{", ".join(f"{col_name!r}" for col_name in missing_required)}'
+            )
+
+        computed_in_model = computed_cols & model_field_names
```
aaron-siegel, line 260: Same comment as above.

pixeltable/utils/pydantic.py
```diff
@@ -0,0 +1,62 @@
...
+    if type_hint is type(None):
+        return True
+    if type_hint is Any:
+        return False
+
+    if type_hint in (str, int, float, bool):
+        return True
+
+    # datetime types serialize to ISO strings
+    if type_hint is datetime.datetime or type_hint is datetime.date:
```
aaron-siegel: can't this be merged with the previous line?

`if type_hint in (str, int, float, bool, datetime.datetime, datetime.date)`

pixeltable/utils/pydantic.py
```diff
@@ -0,0 +1,62 @@
...
+    if isinstance(type_hint, type) and issubclass(type_hint, Enum):
+        return True
+
+    if isinstance(type_hint, type) and issubclass(type_hint, pydantic.BaseModel):
+        return is_json_convertible(type_hint)
+
+    origin = typing.get_origin(type_hint)
+    args = typing.get_args(type_hint)
+
+    if origin is Union:
```
aaron-siegel: I'm not certain this will work for a Python 3.10 union (`T | U`), let's make sure there's a test case for that (I believe `T | U` maps to `types.UnionType`, not `typing.Union`)

pixeltable/utils/pydantic.py
```diff
@@ -0,0 +1,62 @@
+import datetime
+import typing
+from enum import Enum
+from typing import Any, Union
+
+import pydantic
+
+
+def is_json_convertible(model: type[pydantic.BaseModel]) -> bool:
```
aaron-siegel, line 10: Add a TODO that we want to support PIL images and np ndarrays eventually?

pixeltable/utils/pydantic.py
```diff
@@ -0,0 +1,62 @@
...
+    if origin is dict or (hasattr(origin, '__name__') and origin.__name__ == 'Dict'):
+        if len(args) != 2:
+            # we can't tell what this is
+            return False
+        key_type, value_type = args
+        # keys must be strings, values must be json-convertible
+        return key_type is str and _type_is_json_convertible(value_type)
+
+    # Literal types are json-convertible if their values are
+    if hasattr(typing, 'get_origin') and origin is typing.Literal:
```
aaron-siegel: What's the purpose of the `hasattr(typing, 'get_origin')`? Is that ever false?

aaron-siegel (changes requested):

pixeltable/catalog/insertable_table.py
mkornacker: Done

pixeltable/catalog/insertable_table.py
mkornacker: Done.

pixeltable/catalog/insertable_table.py
mkornacker: Done.

Changed to printing the index.

pixeltable/catalog/insertable_table.py
mkornacker: Done.

pixeltable/catalog/table.py

pixeltable/utils/pydantic.py
mkornacker: Done.

pixeltable/catalog/insertable_table.py
mkornacker: I don't think this would matter to a user.

pixeltable/catalog/insertable_table.py
mkornacker: Same as above.

pixeltable/globals.py
```diff
@@ -25,6 +26,7 @@
         Path,  # OS paths, filenames, URLs
         Iterator[dict[str, Any]],  # iterator producing dictionaries of values
         RowData,  # list of dictionaries
+        Sequence[pydantic.BaseModel],  # list of Pydantic models
```
aaron-siegel, line 29: This might as well be an `Iterable[pydantic.BaseModel]`, no? I think all we do is iterate over it.

mkornacker: It makes _insert_pydantic() a little simpler. We can change it later if we find an example where it's not a sequence.

PR #761, opened 2025-08-08
Correct construction of two row_builder members.

jpeterson-pxt:
Table columns can be added to the RowBuilder after its initial construction.
This change ensures that the stored_media_cols and stored_img_cols summaries
are kept up to date.

### DIFF
--- a/pixeltable/exprs/row_builder.py
+++ b/pixeltable/exprs/row_builder.py
@@ -114,6 +114,8 @@ class RowBuilder:
         """
         self.unique_exprs: ExprSet[Expr] = ExprSet()  # dependencies precede their dependents
         self.next_slot_idx = 0
+        self.stored_img_cols = []
+        self.stored_media_cols = []
 
         # record input and output exprs; make copies to avoid reusing execution state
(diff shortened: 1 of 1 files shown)


pixeltable/exprs/row_builder.py
```diff
@@ -245,18 +247,15 @@ def refs_unstored_iter_col(col_ref: ColumnRef) -> bool:
...
-
     def add_table_column(self, col: catalog.Column, slot_idx: int) -> None:
         """Record a column that is part of the table row"""
         assert self.tbl is not None
-        self.table_columns.append(ColumnSlotIdx(col, slot_idx))
-
-    def output_slot_idxs(self) -> list[ColumnSlotIdx]:
-        """Return ColumnSlotIdx for output columns"""
-        return self.table_columns
+        info = ColumnSlotIdx(col, slot_idx)
```
aaron-siegel, line 254: Where is it checked that `col` is stored? If it's assumed, add an assertion to that effect.

jpeterson-pxt: This a pre-condition of calling "add_table_column".
I can add a check but I'm pretty sure the "table_column" part means that it is part of a DB table.
:-)

PR #762, opened 2025-08-08
[WiP] Implementation of storing media files in specific local filesystem directories

jpeterson-pxt:
Creates an ObjectStoreSaveNode which is added to plans as needed. 
Plumbs Column destination parameter through from user schema specification to implementation.
Moves all saving / moving of media files from RowBuilder.create_table_store_row into the ObjectStoreSaveNode.
Still some editing / renaming required, but all tests pass.


jpeterson-pxt:
No longer needed.

PR #765, opened 2025-08-14
Implement a 'destination' specification for the results of stored computed media columns.

jpeterson-pxt:
The destination specification on stored, computed columns can take the form of a string describing a file path, a python Path object, or a string specifying an S3 bucket with optional prefixes.
Syntax examples: 
    t.add_computed_column(img_rot2=t.img.rotate(180), destination='s3://bucket_name/prefix_1/prefix_2/')
    t.add_computed_column(img_rot2=t.img.rotate(180), destination='/Users/name/dir_1/dir_2/')
Computed data for the column is stored in the URI or local Path specified by the destination.
When a table is dropped, the computed values are deleted from the associated destination(s) for all columns in
the table. 
If no destination parameter or the value None is supplied, then the computed values are sent to the default location for pixeltable data: .pixeltable/media


jpeterson-pxt:
Wrong branch

PR #766, opened 2025-08-14
Enable a destination parameter on stored computed columns.

jpeterson-pxt:
The destination specification on stored, computed columns can take the form of a string describing a file path, a python Path object, or a string specifying an S3 bucket with optional prefixes.
Syntax examples:
t.add_computed_column(img_rot2=t.img.rotate(180), destination='s3://bucket_name/prefix_1/prefix_2/')
t.add_computed_column(img_rot2=t.img.rotate(180), destination='/Users/name/dir_1/dir_2/')
Computed data for the column is stored in the URI or local Path specified by the destination.
When a table is dropped, the computed values are deleted from the associated destination(s) for all columns in
the table.
If no destination parameter or the value None is supplied, then the computed values are sent to the default location for pixeltable data: .pixeltable/media

### DIFF
--- a/pixeltable/catalog/column.py
+++ b/pixeltable/catalog/column.py
@@ -34,6 +34,7 @@ class Column:
     col_type: ts.ColumnType
     stored: bool
     is_pk: bool
+    destination: Optional[str]
     _media_validation: Optional[MediaValidation]  # if not set, TableVersion.media_validation applies
     schema_version_add: Optional[int]
     schema_version_drop: Optional[int]
@@ -62,6 +63,7 @@ class Column:
         stores_cellmd: Optional[bool] = None,
         value_expr_dict: Optional[dict[str, Any]] = None,
         tbl: Optional[TableVersion] = None,
+        destination: Optional[str] = None,
     ):
         """Column constructor.
 
@@ -126,6 +128,7 @@ class Column:
 
         # computed cols also have storage columns for the exception string and type
         self.sa_cellmd_col = None
+        self.destination = destination
 
     def to_md(self, pos: Optional[int] = None) -> tuple[schema.ColumnMd, Optional[schema.SchemaColumn]]:
         """Returns the Column and optional SchemaColumn metadata for this Column."""
@@ -138,6 +141,7 @@ class Column:
             schema_version_drop=self.schema_version_drop,
             value_expr=self.value_expr.as_dict() if self.value_expr is not None else None,
             stored=self.stored,
+            destination=self.destination,
         )
         if pos is None:
             return col_md, None
@@ -172,6 +176,7 @@ class Column:
             schema_version_drop=col_md.schema_version_drop,
             value_expr_dict=col_md.value_expr,
             tbl=tbl,
+            destination=col_md.destination,
         )
         return col
 
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -6,7 +6,7 @@ import json
 import logging
 from keyword import iskeyword as is_python_keyword
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, ClassVar, Iterable, Literal, Optional, TypedDict, overload
+from typing import TYPE_CHECKING, Any, ClassVar, Iterable, Literal, Optional, TypedDict, Union, overload
 
 from typing import _GenericAlias  # type: ignore[attr-defined]  # isort: skip
 import datetime
@@ -42,6 +42,7 @@ if TYPE_CHECKING:
 
     import pixeltable.plan
     from pixeltable.globals import TableDataSource
+
 
 _logger = logging.getLogger('pixeltable')
 
@@ -483,8 +484,7 @@ class Table(SchemaObject):
         Adds multiple columns to the table. The columns must be concrete (non-computed) columns; to add computed
         columns, use [`add_computed_column()`][pixeltable.catalog.Table.add_computed_column] instead.
 
-        The format of the `schema` argument is identical to the format of the schema in a call to
-        [`create_table()`][pixeltable.globals.create_table].
+        The format of the `schema` argument is a dict mapping column names to their types.
 
         Args:
             schema: A dictionary mapping column names to types.
@@ -597,6 +597,7 @@ class Table(SchemaObject):
         self,
         *,
         stored: Optional[bool] = None,
+        destination: Optional[Union[str, Path]] = None,
         print_stats: bool = False,
         on_error: Literal['abort', 'ignore'] = 'abort',
         if_exists: Literal['error', 'ignore', 'replace'] = 'error',
@@ -658,6 +659,9 @@ class Table(SchemaObject):
             if stored is not None:
                 col_schema['stored'] = stored
 
+            if destination is not None:
+                col_schema['destination'] = destination
+
             # Raise an error if the column expression refers to a column error property
             if isinstance(spec, exprs.Expr):
                 for e in spec.subexprs(expr_class=exprs.ColumnPropertyRef, traverse_matches=False):
@@ -672,7 +676,7 @@ class Table(SchemaObject):
                 [col_name], IfExistsParam.validated(if_exists, 'if_exists')
             )
             # if the column to add already exists and user asked to ignore
-            # exiting column, there's nothing to do.
+            # existing column, there's nothing to do.
             result = UpdateStatus()
             if len(cols_to_ignore) != 0:
                 assert cols_to_ignore[0] == col_name
@@ -693,7 +697,7 @@ class Table(SchemaObject):
         (on account of containing Python Callables or Exprs).
         """
         assert isinstance(spec, dict)
-        valid_keys = {'type', 'value', 'stored', 'media_validation'}
+        valid_keys = {'type', 'value', 'stored', 'media_validation', 'destination'}
         for k in spec:
             if k not in valid_keys:
                 raise excs.Error(f'Column {name}: invalid key {k!r}')
@@ -717,6 +721,10 @@ class Table(SchemaObject):
         if 'stored' in spec and not isinstance(spec['stored'], bool):
             raise excs.Error(f'Column {name}: "stored" must be a bool, got {spec["stored"]}')
 
+        d = spec.get('destination')
+        if d is not None and not isinstance(d, (str, Path)):
+            raise excs.Error(f'Column {name}: "destination" must be a string or path, got {d}')
+
     @classmethod
     def _create_columns(cls, schema: dict[str, Any]) -> list[Column]:
         """Construct list of Columns, given schema"""
@@ -727,6 +735,7 @@ class Table(SchemaObject):
             primary_key: bool = False
             media_validation: Optional[catalog.MediaValidation] = None
             stored = True
+            destination: Optional[str] = None
 
             if isinstance(spec, (ts.ColumnType, type, _GenericAlias)):
                 col_type = ts.ColumnType.normalize_type(spec, nullable_default=True, allow_builtin_types=False)
@@ -751,6 +760,10 @@ class Table(SchemaObject):
                 media_validation = (
                     catalog.MediaValidation[media_validation_str.upper()] if media_validation_str is not None else None
                 )
+                if 'destination' in spec:
+                    from pixeltable.utils.media_destination import MediaDestination
+
+                    destination = MediaDestination.validate_destination(name, spec['destination'])
             else:
                 raise excs.Error(f'Invalid value for column {name!r}')
 
@@ -761,6 +774,7 @@ class Table(SchemaObject):
                 stored=stored,
                 is_pk=primary_key,
                 media_validation=media_validation,
+                destination=destination,
             )
             columns.append(column)
         return columns
@@ -786,14 +800,16 @@ class Table(SchemaObject):
                     f'streaming function'
                 )
             )
+        if col.destination is not None and not (col.stored and col.is_computed):
+            raise excs.Error(
+                f'Column {col.name!r}: destination={col.destination} only applies to stored computed columns'
+            )
 
     @classmethod
     def _verify_schema(cls, schema: list[Column]) -> None:
         """Check integrity of user-supplied schema and set defaults"""
-        column_names: set[str] = set()
         for col in schema:
             cls._verify_column(col)
-            column_names.add(col.name)
 
     def drop_column(self, column: str | ColumnRef, if_not_exists: Literal['error', 'ignore'] = 'error') -> None:
         """Drop a column from the table.
--- a/pixeltable/catalog/table_version.py
+++ b/pixeltable/catalog/table_version.py
@@ -20,7 +20,7 @@ from pixeltable.iterators import ComponentIterator
 from pixeltable.metadata import schema
 from pixeltable.utils.exception_handler import run_cleanup_on_exception
 from pixeltable.utils.filecache import FileCache
-from pixeltable.utils.media_store import MediaStore
+from pixeltable.utils.media_destination import MediaDestination
 
 from .tbl_ops import TableOp
 
@@ -327,7 +327,7 @@ class TableVersion:
             from .table_version_path import TableVersionPath
 
             # clear out any remaining media files from an aborted previous attempt
-            MediaStore.get().delete(self.id)
+            self.delete_media()
             view_path = TableVersionPath.from_dict(op.load_view_op.view_path)
             plan, _ = Planner.create_view_load_plan(view_path)
             _, row_counts = self.store_tbl.insert_rows(plan, v_min=self.version)
@@ -364,6 +364,12 @@ class TableVersion:
         tbl_version.store_tbl.ensure_columns_exist(col for col in tbl_version.cols if col.is_stored)
         return tbl_version
 
+    def delete_media(self, tbl_version: Optional[int] = None) -> None:
+        # Assemble a set of column destinations and delete media from all of them
+        destinations = {col.destination for col in self.cols if col.is_stored}
+        for dest in destinations:
+            MediaDestination.delete(dest, self.id, tbl_version=tbl_version)
+
     def drop(self) -> None:
         # if self.is_view and self.is_mutable:
         #     # update mutable_views
@@ -374,7 +380,7 @@ class TableVersion:
         #     if self.base.get().is_mutable:
         #         self.base.get().mutable_views.remove(TableVersionHandle.create(self))
 
-        MediaStore.get().delete(self.id)
+        self.delete_media()
         FileCache.get().clear(tbl_id=self.id)
         self.store_tbl.drop()
 
@@ -1235,7 +1241,7 @@ class TableVersion:
             )
 
         # delete newly-added data
-        MediaStore.get().delete(self.id, tbl_version=self.version)
+        self.delete_media(tbl_version=self.version)
         conn.execute(sql.delete(self.store_tbl.sa_tbl).where(self.store_tbl.sa_tbl.c.v_min == self.version))
 
         # revert new deletions
--- a/pixeltable/env.py
+++ b/pixeltable/env.py
@@ -234,6 +234,11 @@ class Env:
         assert self._db_url is not None  # is_local should be called only after db initialization
         return self._db_server is not None
 
+    @property
+    def cpu_count(self) -> int:
+        n_cpu = os.cpu_count()
+        return n_cpu if n_cpu is not None and n_cpu > 0 else 1
+
     @contextmanager
     def begin_xact(self, for_write: bool = False) -> Iterator[sql.Connection]:
         """
--- a/pixeltable/exec/cache_prefetch_node.py
+++ b/pixeltable/exec/cache_prefetch_node.py
@@ -9,12 +9,14 @@ import urllib.request
 from collections import deque
 from concurrent import futures
 from pathlib import Path
-from typing import Any, AsyncIterator, Iterator, Optional
+from typing import AsyncIterator, Iterator, Optional
 from uuid import UUID
 
 from pixeltable import exceptions as excs, exprs
+from pixeltable.env import Env
 from pixeltable.utils.filecache import FileCache
 from pixeltable.utils.media_store import TempStore
+from pixeltable.utils.s3 import S3ClientContainer
 
 from .data_row_batch import DataRowBatch
 from .exec_node import ExecNode
@@ -26,16 +28,16 @@ class CachePrefetchNode(ExecNode):
     """Brings files with external URLs into the cache
 
     TODO:
-    - adapting the number of download threads at runtime to maximize throughput
+    - Adapt the number of download threads at runtime to maximize throughput
+    - Process a row at a time and limit the number of in-flight rows to control memory usage
     """
 
+    QUEUE_DEPTH_HIGH_WATER = 50  # target number of in-flight requests
+    QUEUE_DEPTH_LOW_WATER = 20  # target number of in-flight requests
     BATCH_SIZE = 16
-    NUM_EXECUTOR_THREADS = 16
 
     retain_input_order: bool  # if True, return rows in the exact order they were received
     file_col_info: list[exprs.ColumnSlotIdx]
-    boto_client: Optional[Any]
-    boto_client_lock: threading.Lock
 
     # execution state
     num_returned_rows: int
@@ -49,6 +51,7 @@ class CachePrefetchNode(ExecNode):
     in_flight_urls: dict[str, list[tuple[exprs.DataRow, exprs.ColumnSlotIdx]]]  # URL -> [(row, info)]
     input_finished: bool
     row_idx: Iterator[Optional[int]]
+    num_threads: int
 
     @dataclasses.dataclass
     class RowState:
@@ -64,10 +67,6 @@ class CachePrefetchNode(ExecNode):
         self.retain_input_order = retain_input_order
         self.file_col_info = file_col_info
 
-        # clients for specific services are constructed as needed, because it's time-consuming
-        self.boto_client = None
-        self.boto_client_lock = threading.Lock()
-
         self.num_returned_rows = 0
         self.ready_rows = deque()
         self.in_flight_rows = {}
@@ -75,24 +74,45 @@ class CachePrefetchNode(ExecNode):
         self.in_flight_urls = {}
         self.input_finished = False
         self.row_idx = itertools.count() if retain_input_order else itertools.repeat(None)
+        self.num_threads = max(4, Env.get().cpu_count)
+        assert self.QUEUE_DEPTH_HIGH_WATER > self.QUEUE_DEPTH_LOW_WATER
+        # Ensure that the S3ClientContainer is initialized before using threading
+        S3ClientContainer.get()
+
+    @property
+    def queued_work(self) -> int:
+        return len(self.in_flight_requests)
+
+    async def get_input_batch(self, input_iter: AsyncIterator[DataRowBatch]) -> Optional[DataRowBatch]:
+        """Get the next batch of input rows, or None if there are no more rows"""
+        try:
+            input_batch = await anext(input_iter)
+            if input_batch is None:
+                self.input_finished = True
+            return input_batch
+        except StopAsyncIteration:
+            self.input_finished = True
+            return None
 
     async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
         input_iter = self.input.__aiter__()
-        with futures.ThreadPoolExecutor(max_workers=self.NUM_EXECUTOR_THREADS) as executor:
-            # we create enough in-flight requests to fill the first batch
-            while not self.input_finished and self.__num_pending_rows() < self.BATCH_SIZE:
-                await self.__submit_input_batch(input_iter, executor)
-
+        with futures.ThreadPoolExecutor(max_workers=self.num_threads) as executor:
             while True:
-                # try to assemble a full batch of output rows
-                if not self.__has_ready_batch() and len(self.in_flight_requests) > 0:
-                    self.__wait_for_requests()
+                # Create work to fill the queue to the high water mark ... ?without overrunning the in-flight row limit.
+                while not self.input_finished and self.queued_work < self.QUEUE_DEPTH_HIGH_WATER:
+                    input_batch = await self.get_input_batch(input_iter)
+                    if input_batch is not None:
+                        self.__process_input_batch(input_batch, executor)
 
-                # try to create enough in-flight requests to fill the next batch
-                while not self.input_finished and self.__num_pending_rows() < self.BATCH_SIZE:
-                    await self.__submit_input_batch(input_iter, executor)
+                # Wait for enough completions to enable more queueing or if we're done
+                while self.queued_work > self.QUEUE_DEPTH_LOW_WATER or (self.input_finished and self.queued_work > 0):
+                    done, _ = futures.wait(self.in_flight_requests, return_when=futures.FIRST_COMPLETED)
+                    self.__process_completions(done, ignore_errors=self.ctx.ignore_errors)
 
-                if len(self.ready_rows) > 0:
+                # Emit results to meet batch size requirements or empty the in-flight row queue
+                if self.__has_ready_batch() or (
+                    len(self.ready_rows) > 0 and self.input_finished and self.queued_work == 0
+                ):
                     # create DataRowBatch from the first BATCH_SIZE ready rows
                     batch = DataRowBatch(self.row_builder)
                     rows = [self.ready_rows.popleft() for _ in range(min(self.BATCH_SIZE, len(self.ready_rows)))]
@@ -103,21 +123,14 @@ class CachePrefetchNode(ExecNode):
                     _logger.debug(f'returning {len(rows)} rows')
                     yield batch
 
-                if self.input_finished and self.__num_pending_rows() == 0:
+                if self.input_finished and self.queued_work == 0 and len(self.ready_rows) == 0:
                     return
-
-    def __num_pending_rows(self) -> int:
-        return len(self.in_flight_rows) + len(self.ready_rows)
 
     def __has_ready_batch(self) -> bool:
         """True if there are >= BATCH_SIZES entries in ready_rows and the first BATCH_SIZE ones are all non-None"""
         return (
             sum(int(row is not None) for row in itertools.islice(self.ready_rows, self.BATCH_SIZE)) == self.BATCH_SIZE
         )
-
-    def __ready_prefix_len(self) -> int:
-        """Length of the non-None prefix of ready_rows (= what we can return right now)"""
-        return sum(1 for _ in itertools.takewhile(lambda x: x is not None, self.ready_rows))
 
     def __add_ready_row(self, row: exprs.DataRow, row_idx: Optional[int]) -> None:
         if row_idx is None:
@@ -129,50 +142,36 @@ class CachePrefetchNode(ExecNode):
                 self.ready_rows.extend([None] * (idx - len(self.ready_rows) + 1))
             self.ready_rows[idx] = row
 
-    def __wait_for_requests(self) -> None:
-        """Wait for in-flight requests to complete until we have a full batch of rows"""
+    def __process_completions(self, done: set[futures.Future], ignore_errors: bool) -> None:
         file_cache = FileCache.get()
-        _logger.debug(f'waiting for requests; ready_batch_size={self.__ready_prefix_len()}')
-        while not self.__has_ready_batch() and len(self.in_flight_requests) > 0:
-            done, _ = futures.wait(self.in_flight_requests, return_when=futures.FIRST_COMPLETED)
-            for f in done:
-                url = self.in_flight_requests.pop(f)
-                tmp_path, exc = f.result()
-                local_path: Optional[Path] = None
-                if tmp_path is not None:
-                    # register the file with the cache for the first column in which it's missing
-                    assert url in self.in_flight_urls
-                    _, info = self.in_flight_urls[url][0]
-                    local_path = file_cache.add(info.col.tbl.id, info.col.id, url, tmp_path)
-                    _logger.debug(f'cached {url} as {local_path}')
+        for f in done:
+            url = self.in_flight_requests.pop(f)
+            tmp_path, exc = f.result()
+            if exc is not None and not ignore_errors:
+                raise exc
+            local_path: Optional[Path] = None
+            if tmp_path is not None:
+                # register the file with the cache for the first column in which it's missing
+                assert url in self.in_flight_urls
+                _, info = self.in_flight_urls[url][0]
+                local_path = file_cache.add(info.col.tbl.id, info.col.id, url, tmp_path)
+                _logger.debug(f'cached {url} as {local_path}')
 
-                # add the local path/exception to the slots that reference the url
-                for row, info in self.in_flight_urls.pop(url):
-                    if exc is not None:
-                        self.row_builder.set_exc(row, info.slot_idx, exc)
-                    else:
-                        assert local_path is not None
-                        row.set_file_path(info.slot_idx, str(local_path))
-                    state = self.in_flight_rows[id(row)]
-                    state.num_missing -= 1
-                    if state.num_missing == 0:
-                        del self.in_flight_rows[id(row)]
-                        self.__add_ready_row(row, state.idx)
-                        _logger.debug(f'row {state.idx} is ready (ready_batch_size={self.__ready_prefix_len()})')
+            # add the local path/exception to the slots that reference the url
+            for row, info in self.in_flight_urls.pop(url):
+                if exc is not None:
+                    self.row_builder.set_exc(row, info.slot_idx, exc)
+                else:
+                    assert local_path is not None
+                    row.set_file_path(info.slot_idx, str(local_path))
+                state = self.in_flight_rows[id(row)]
+                state.num_missing -= 1
+                if state.num_missing == 0:
+                    del self.in_flight_rows[id(row)]
+                    self.__add_ready_row(row, state.idx)
 
-    async def __submit_input_batch(
-        self, input: AsyncIterator[DataRowBatch], executor: futures.ThreadPoolExecutor
-    ) -> None:
-        assert not self.input_finished
-        input_batch: Optional[DataRowBatch]
-        try:
-            input_batch = await anext(input)
-        except StopAsyncIteration:
-            input_batch = None
-        if input_batch is None:
-            self.input_finished = True
-            return
-
+    def __process_input_batch(self, input_batch: DataRowBatch, executor: futures.ThreadPoolExecutor) -> None:
+        """Process a batch of input rows, submitting URLs for download and adding ready rows to ready_rows"""
         file_cache = FileCache.get()
 
         # URLs from this input batch that aren't already in the file cache;
@@ -180,7 +179,7 @@ class CachePrefetchNode(ExecNode):
         # the time it takes to get the next batch together
         cache_misses: list[str] = []
 
-        url_pos: dict[str, int] = {}  # url -> row_idx; used for logging
+        url_pos: dict[str, Optional[int]] = {}  # url -> row_idx; used for logging
         for row in input_batch:
             # identify missing local files in input batch, or fill in their paths if they're already cached
             num_missing = 0
@@ -235,18 +234,8 @@ class CachePrefetchNode(ExecNode):
         try:
             _logger.debug(f'Downloading {url} to {tmp_path}')
             if parsed.scheme == 's3':
-                from pixeltable.utils.s3 import get_client
-
-                with self.boto_client_lock:
-                    if self.boto_client is None:
-                        config = {
-                            'max_pool_connections': self.NUM_EXECUTOR_THREADS + 4,  # +4: leave some headroom
-                            'connect_timeout': 5,
-                            'read_timeout': 30,
-                            'retries': {'max_attempts': 3, 'mode': 'adaptive'},
-                        }
-                        self.boto_client = get_client(**config)
-                self.boto_client.download_file(parsed.netloc, parsed.path.lstrip('/'), str(tmp_path))
+                client = S3ClientContainer.get().get_client(for_write=False)
+                client.download_file(parsed.netloc, parsed.path.lstrip('/'), str(tmp_path))
             elif parsed.scheme in ('http', 'https'):
                 with urllib.request.urlopen(url) as resp, open(tmp_path, 'wb') as f:
                     data = resp.read()
@@ -259,6 +248,4 @@ class CachePrefetchNode(ExecNode):
             # we want to add the file url to the exception message
             exc = excs.Error(f'Failed to download {url}: {e}')
             _logger.debug(f'Failed to download {url}: {e}', exc_info=e)
-            if not self.ctx.ignore_errors:
-                raise exc from None  # suppress original exception
             return None, exc
--- a/pixeltable/exec/object_store_save_node.py
+++ b/pixeltable/exec/object_store_save_node.py
@@ -1,0 +1,304 @@
+from __future__ import annotations
+
+import dataclasses
+import itertools
+import logging
+from collections import defaultdict, deque
+from concurrent import futures
+from pathlib import Path
+from typing import AsyncIterator, Iterator, NamedTuple, Optional
+from uuid import UUID
+
+from pixeltable import exprs
+from pixeltable.env import Env
+from pixeltable.utils.media_destination import MediaDestination
+from pixeltable.utils.media_store import MediaStore, TempStore
+from pixeltable.utils.s3 import S3ClientContainer
+
+from .data_row_batch import DataRowBatch
+from .exec_node import ExecNode
+
+_logger = logging.getLogger('pixeltable')
+
+
+class ObjectStoreSaveNode(ExecNode):
+    """Brings files with external URLs into the cache
+
+    Each row may have multipe files that need to be saved to a destination.
+    Each file may be referenced by more than one column in the row.
+    Each file may have multiple destinations, e.g., S3 bucket and local file system.
+    If there are multiple destinations, the file cannot be moved to any destination
+    until it has been copied to all of the other destinations.
+    Diagrammatically:
+        Row -> [src_path1, src_path2, ...]
+            src_path -> [dest1, dest2, ...]
+                dest1: [row_location1, row_location2, ...]
+    Paths with multiple destinations are removed from the TempStore only after all destination copies are complete.
+
+    TODO:
+    - Adapt the number of download threads at runtime to maximize throughput
+    - Process a row at a time and limit the number of in-flight rows to control memory usage
+    """
+
+    QUEUE_DEPTH_HIGH_WATER = 50  # target number of in-flight requests
+    QUEUE_DEPTH_LOW_WATER = 20  # target number of in-flight requests
+    BATCH_SIZE = 16
+
+    class WorkDesignator(NamedTuple):
+        """Specify the source and destination for a WorkItem"""
+
+        src_path: str  # source of the file to be processed
+        destination: Optional[str]  # destination URI for the file to be processed
+
+    class WorkItem(NamedTuple):
+        src_path: Path
+        destination: Optional[str]
+        info: exprs.ColumnSlotIdx  # column info for the file being processed
+        destination_count: int = 1  # number of unique destinations for this file
+
+        def pp(self) -> str:
+            return f'WorkItem(src_path={self.src_path}, slot_idx={self.info.slot_idx}, info={self.info})'
+
+    retain_input_order: bool  # if True, return rows in the exact order they were received
+    file_col_info: list[exprs.ColumnSlotIdx]
+
+    # execution state
+    num_returned_rows: int
+
+    # ready_rows: rows that are ready to be returned, ordered by row idx;
+    # the implied row idx of ready_rows[0] is num_returned_rows
+    ready_rows: deque[Optional[exprs.DataRow]]
+
+    in_flight_rows: dict[int, ObjectStoreSaveNode.RowState]  # rows with in-flight work; id(row) -> RowState
+    in_flight_requests: dict[
+        futures.Future, WorkDesignator
+    ]  # in-flight requests to save paths: Future -> WorkDesignator
+    in_flight_work: dict[
+        WorkDesignator, list[tuple[exprs.DataRow, exprs.ColumnSlotIdx]]
+    ]  # WorkDesignator -> [(row, info)]
+
+    input_finished: bool
+    row_idx: Iterator[Optional[int]]
+    num_threads: int
+
+    @dataclasses.dataclass
+    class RowState:
+        row: exprs.DataRow
+        idx: Optional[int]  # position in input stream; None if we don't retain input order
+        num_missing: int  # number of references to media files in this row
+        delete_destinations: list[Path]  # paths to delete after all copies are complete
+
+    def __init__(
+        self, tbl_id: UUID, file_col_info: list[exprs.ColumnSlotIdx], input: ExecNode, retain_input_order: bool = True
+    ):
+        # input_/output_exprs=[]: we don't have anything to evaluate
+        super().__init__(input.row_builder, [], [], input)
+        self.retain_input_order = retain_input_order
+        self.file_col_info = file_col_info
+
+        self.num_returned_rows = 0
+        self.ready_rows = deque()
+        self.in_flight_rows = {}
+        self.in_flight_requests = {}
+        self.in_flight_work = {}
+        self.input_finished = False
+        self.row_idx = itertools.count() if retain_input_order else itertools.repeat(None)
+        self.num_threads = max(4, Env.get().cpu_count)
+        assert self.QUEUE_DEPTH_HIGH_WATER > self.QUEUE_DEPTH_LOW_WATER
+        # Ensure that the S3ClientContainer is initialized before using threading
+        S3ClientContainer.get()
+
+    @property
+    def queued_work(self) -> int:
+        return len(self.in_flight_requests)
+
+    async def get_input_batch(self, input_iter: AsyncIterator[DataRowBatch]) -> Optional[DataRowBatch]:
+        """Get the next batch of input rows, or None if there are no more rows"""
+        try:
+            input_batch = await anext(input_iter)
+            if input_batch is None:
+                self.input_finished = True
+            return input_batch
+        except StopAsyncIteration:
+            self.input_finished = True
+            return None
+
+    async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
+        input_iter = self.input.__aiter__()
+        with futures.ThreadPoolExecutor(max_workers=self.num_threads) as executor:
+            while True:
+                # Create work to fill the queue to the high water mark ... ?without overrunning the in-flight row limit.
+                while not self.input_finished and self.queued_work < self.QUEUE_DEPTH_HIGH_WATER:
+                    input_batch = await self.get_input_batch(input_iter)
+                    if input_batch is not None:
+                        self.__process_input_batch(input_batch, executor)
+
+                # Wait for enough completions to enable more queueing or if we're done
+                while self.queued_work > self.QUEUE_DEPTH_LOW_WATER or (self.input_finished and self.queued_work > 0):
+                    done, _ = futures.wait(self.in_flight_requests, return_when=futures.FIRST_COMPLETED)
+                    self.__process_completions(done, ignore_errors=self.ctx.ignore_errors)
+
+                # Emit results to meet batch size requirements or empty the in-flight row queue
+                if self.__has_ready_batch() or (
+                    len(self.ready_rows) > 0 and self.input_finished and self.queued_work == 0
+                ):
+                    # create DataRowBatch from the first BATCH_SIZE ready rows
+                    batch = DataRowBatch(self.row_builder)
+                    rows = [self.ready_rows.popleft() for _ in range(min(self.BATCH_SIZE, len(self.ready_rows)))]
+                    for row in rows:
+                        assert row is not None
+                        batch.add_row(row)
+                    self.num_returned_rows += len(rows)
+                    _logger.debug(f'returning {len(rows)} rows')
+                    yield batch
+
+                if self.input_finished and self.queued_work == 0 and len(self.ready_rows) == 0:
+                    return
+
+    def __has_ready_batch(self) -> bool:
+        """True if there are >= BATCH_SIZES entries in ready_rows and the first BATCH_SIZE ones are all non-None"""
+        return (
+            sum(int(row is not None) for row in itertools.islice(self.ready_rows, self.BATCH_SIZE)) == self.BATCH_SIZE
+        )
+
+    def __add_ready_row(self, row: exprs.DataRow, row_idx: Optional[int]) -> None:
+        if row_idx is None:
+            self.ready_rows.append(row)
+        else:
+            # extend ready_rows to accommodate row_idx
+            idx = row_idx - self.num_returned_rows
+            if idx >= len(self.ready_rows):
+                self.ready_rows.extend([None] * (idx - len(self.ready_rows) + 1))
+            self.ready_rows[idx] = row
+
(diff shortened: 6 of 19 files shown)


pixeltable/exec/cache_prefetch_node.py
```diff
@@ -64,35 +67,52 @@ def __init__(
...
-        self.boto_client_lock = threading.Lock()
-
         self.num_returned_rows = 0
         self.ready_rows = deque()
         self.in_flight_rows = {}
         self.in_flight_requests = {}
         self.in_flight_urls = {}
         self.input_finished = False
         self.row_idx = itertools.count() if retain_input_order else itertools.repeat(None)
+        self.num_threads = max(4, Env.get().cpu_count)
```
mkornacker: Why do we care about the number of cores here? These are IO threads and are mostly just sitting around doing nothing.

jpeterson-pxt: True, but if they are not explicitly sleeping, time is spent checking to see if there is work to do. 
The number should scale with CPU cores, as threads are typically distributed across the cores, and more cores allow for more servicing.
Remember too that many queries will now have two thread pools, one for fetch to cache, and one for write to target, so whatever number is used here is more or less doubled. 
The numbers here are lower bounds for pure IO threads, we could go higher.

mkornacker: It looks like this needs a different approach, such as a thread pool in ExecContext.

jpeterson-pxt:
@mkornacker: In reference to your comment with the AWS Lambda reference:
I hadn't seen that reference in particular, but I think this code is relatively consistent with that reference, given that we're not writing an AWS Lambda function and we don't know which platform(s) will run this code. (Especially note the recommended max thread count of 30).
We can perhaps discuss tomorrow in person.

pixeltable/exec/cache_prefetch_node.py
jpeterson-pxt: We could do that. I think for now it's fine to have two thread pools (they're not that expensive), but if we plan to expand our thread usage for other execution tasks, I would definitely push for a single pool and a more general thread tasking mechanism.
In this case I would make the thread pool a singleton, and just keep the threads around all the time.

mkornacker:
Everything I have seen so far, AWS posts or from elsewhere, says that you can use a single boto3 client for multi-threaded operations, including concurrent uploads and downloads.

It appears that the best path forward is to have a single client in ExecContext.

pixeltable/catalog/column.py
```diff
@@ -34,6 +34,7 @@ class Column:
     col_type: ts.ColumnType
     stored: bool
     is_pk: bool
+    destination: Optional[str]
```
mkornacker: | None (and elsewhere)

pixeltable/catalog/table_version.py
```diff
@@ -364,6 +364,12 @@ def create_replica(cls, md: schema.FullTableMd) -> TableVersion:
         tbl_version.store_tbl.ensure_columns_exist(col for col in tbl_version.cols if col.is_stored)
         return tbl_version
 
+    def delete_media(self, tbl_version: Optional[int] = None) -> None:
+        # Assemble a set of column destinations and delete media from all of them
+        destinations = {col.destination for col in self.cols if col.is_stored}
```
mkornacker, line 372: This can contain Nones. Why leave them in there?

mkornacker (commented):
Is this ready for review? We decided to having one client per object store service, in Env, but I'm not seeing that being implemented yet.

pixeltable/catalog/table_version.py
jpeterson-pxt: It is a set, there will be only one or zero None's. 
The "None" value for the column destination indicates that the computed value was saved to the local file default media directory. 
We want to remove them from there as well as all of the external destinations. 
So, if Nones are present, one is kept in the set, so that we remove files from the local default media directory and don't leave them there.

pixeltable/catalog/column.py
aaron-siegel: Technically we haven't agreed to standardize on this (but we could - let's discuss at standup?)

pixeltable/catalog/table.py
```diff
@@ -597,6 +597,7 @@ def add_computed_column(
         self,
         *,
         stored: Optional[bool] = None,
+        destination: Optional[Union[str, Path]] = None,
```
aaron-siegel: However, we have definitely agreed to standardize on `T | U` in place of `Union[T, U]`.

pixeltable/catalog/table_version.py
aaron-siegel: Maybe add a code comment to clarify this (it'll be more durable than the github review comment)

pixeltable/exec/object_store_save_node.py
```diff
@@ -0,0 +1,306 @@
...
+from .data_row_batch import DataRowBatch
+from .exec_node import ExecNode
+
+_logger = logging.getLogger('pixeltable')
+
+
+class ObjectStoreSaveNode(ExecNode):
+    """Brings files with external URLs into the cache
+
+    Each row may have multipe files that need to be saved to a destination.
```
aaron-siegel: typo ("multiple")

pixeltable/exec/object_store_save_node.py
```diff
@@ -0,0 +1,306 @@
...
+        src_path: str  # source of the file to be processed
+        destination: Optional[str]  # destination URI for the file to be processed
+
+    class WorkItem(NamedTuple):
+        src_path: Path
+        destination: Optional[str]
+        info: exprs.ColumnSlotIdx  # column info for the file being processed
+        destination_count: int = 1  # number of unique destinations for this file
+
+        def pp(self) -> str:
```
aaron-siegel: Is there a reason not to just implement this as `__repr__()`? (For that matter, why do we need it? What's wrong with the default `__repr__()` of `NamedTuple`?)

pixeltable/exec/object_store_save_node.py
```diff
@@ -0,0 +1,306 @@
...
+        self.ready_rows = deque()
+        self.in_flight_rows = {}
+        self.in_flight_requests = {}
+        self.in_flight_work = {}
+        self.input_finished = False
+        self.row_idx = itertools.count() if retain_input_order else itertools.repeat(None)
+        self.num_threads = max(4, Env.get().cpu_count)
+        assert self.QUEUE_DEPTH_HIGH_WATER > self.QUEUE_DEPTH_LOW_WATER
+        # Ensure that the ClientContainer is initialized before using threading
+        ClientContainer.get()
```
aaron-siegel: Relying on a `get()` method to do stateful initialization seems like confusing semantics to me.

pixeltable/exec/object_store_save_node.py
```diff
@@ -0,0 +1,306 @@
...
+            input_batch = await anext(input_iter)
+            if input_batch is None:
+                self.input_finished = True
+            return input_batch
+        except StopAsyncIteration:
+            self.input_finished = True
+            return None
+
+    async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
+        input_iter = self.input.__aiter__()
```
aaron-siegel: I think we can use `aiter(self.input)` now since we're on Python 3.10.

pixeltable/exprs/data_row.py
```diff
@@ -257,42 +257,43 @@ def __setitem__(self, idx: int, val: Any) -> None:
...
         self.vals[index] = None
+        return False
 
-    def move_tmp_media_file(self, index: int, col: catalog.Column) -> None:
-        """If a media url refers to data in a temporary file, move the data to a MediaStore"""
-        if self.file_urls[index] is None:
-            return
-        assert self.excs[index] is None
+    def save_media_to_temp(self, index: int, col: catalog.Column) -> str:
+        """Save the media object in the column to the column MediaStore file destination or TempStore.
```
jpeterson-pxt: This comment is out of date, it should be "Save a media object in a column to the TempStore."

pixeltable/utils/client_container.py
```diff
@@ -0,0 +1,156 @@
...
+
+import os
+import threading
+from typing import Any, Optional
+
+from pixeltable.env import Env
+from pixeltable.utils.media_path import StorageObjectAddress
+
+
+class ClientContainer:
```
aaron-siegel: This seems to duplicate logic already present in `Env` for managing clients. We had talked about leveraging that existing patter for storage clients; is there some reason why it's not adequate?

pyproject.toml
```diff
@@ -69,6 +69,7 @@ dependencies = [
     "pydantic>=2.7.4",
     "pyarrow>=13.0.0",
     "pixeltable-pgserver==0.3.1",
+    "google-cloud-storage>=2.18.0",
```
aaron-siegel: This should definitely be an optional (dev) dependency, not a required dependency.

aaron-siegel (changes requested):
First round of comments

pixeltable/exec/object_store_save_node.py
jpeterson-pxt: The purpose of of this invocation is to ensure that the singleton is initialized from single-threaded code (the singleton init is not locked). 
If I put a lock on the singleton init, this can disappear. 
What idiom do we prefer to ensure that the singleton is initialized?

pixeltable/exec/object_store_save_node.py
jpeterson-pxt: I tried that and it didn't work. Perhaps I did it incorrectly.

pixeltable/utils/client_container.py
jpeterson-pxt: I think we discussed this. Each client has both explicit and implicit methods for finding credentials. 
It is not clear that these can be stated in the paradigm of the existing client registry.
I can look further into this refactorization. 

Also, the ClientContainer is managing two resources for boto destinations: a "client" and a "resource".

pyproject.toml
jpeterson-pxt: Will fix. 
Different questions:
- With our cloud data sharing service on the way, should we make the boto client a required dependency?
- Do we need to have full support for Cloudflare R2 out of the box for data sharing?

pixeltable/utils/client_container.py
mkornacker: Yes, this needs to utilize Env (and possibly extend if).

Also, we do not need more than a single client instance per destination, it can be shared by all reading/writing threads.

mkornacker (changes requested):

pixeltable/catalog/table.py
jpeterson-pxt: Fixed.

pixeltable/exec/object_store_save_node.py
jpeterson-pxt: Done

pyproject.toml
jpeterson-pxt: Done

pixeltable/catalog/table.py
```diff
@@ -751,6 +760,10 @@ def _create_columns(cls, schema: dict[str, Any]) -> list[Column]:
                 media_validation = (
                     catalog.MediaValidation[media_validation_str.upper()] if media_validation_str is not None else None
                 )
+                if 'destination' in spec:
+                    from pixeltable.utils.media_destination import MediaDestination
```
aaron-siegel: Standard convention is that imports belong at top of file unless they would cause a specific circular dependency or dependency resolution failure, in which case they belong at top of method. (There are occasional exceptions to this rule but a good general guideline is to put them in the most general possible scope, unless there's a specific reason to narrow it)

pixeltable/exprs/data_row.py
jpeterson-pxt: Fixed.

pixeltable/exprs/data_row.py
```diff
@@ -257,42 +257,43 @@ def __setitem__(self, idx: int, val: Any) -> None:
             self.vals[idx] = val
         self.has_val[idx] = True
 
-    def flush_img(self, index: int, col: Optional[catalog.Column] = None) -> None:
-        """Save or discard the in-memory value (required to be a PIL.Image.Image)"""
+    def check_must_save(self, index: int, col: Optional[catalog.Column] = None) -> bool:
+        """Return true if the media object in the column needs to be saved, discard unneeded values"""
```
aaron-siegel: This is doing more than "checking" something, so I'd prefer some clearer verb, such as "prepare_media_val()"?

pixeltable/plan.py
```diff
@@ -762,6 +772,17 @@ def _create_combined_ordering(cls, analyzer: Analyzer, verify_agg: bool) -> Opti
             combined_ordering = combined
         return combined_ordering
 
+    @classmethod
+    def _insert_save_node(
+        cls, tbl_id: UUID, media_cols: list[exprs.ColumnSlotIdx], input_node: exec.ExecNode
+    ) -> exec.ExecNode:
+        """Return an ObjectStoreSaveNode if media columns are present, otherwise return input"""
```
aaron-siegel: Clarify: "if stored media columns are present" (and call the parameter `stored_media_cols` for clarity - this tripped me up when I was reviewing the code)

pixeltable/utils/media_path.py
```diff
@@ -0,0 +1,246 @@
...
+    @property
+    def to_path(self) -> Path:
+        assert self.storage_target == 'os'
+        assert self.prefix
+        path_str = urllib.parse.unquote(urllib.request.url2pathname(self.prefix))
+        return Path(path_str)
+
+    def __str__(self) -> str:
+        return (
+            f'{self.scheme}://{self.account}.{self.account_extension}/{self.container}/{self.prefix}{self.object_name}'
```
aaron-siegel: I feel like we should be using urllib to build these strings rather than string them together manually. (I'm worried about edge cases or things not getting escaped properly, etc.)

Or if we're making assumptions about alphanumeric composition of the ingredients, bake those in as assertions in the initializer

pixeltable/utils/media_path.py
```diff
@@ -0,0 +1,246 @@
...
+    @property
+    def is_http_readable(self) -> bool:
+        return self.scheme.startswith('http') and self.has_object
+
+    @property
+    def is_azure_scheme(self) -> bool:
+        return self.scheme in ['wasb', 'wasbs', 'abfs', 'abfss']
+
+    @property
+    def has_valid_storage_target(self) -> bool:
```
aaron-siegel: A better way to do this is to make `storage_target` an enum or a `Literal` type.

pixeltable/utils/media_path.py
```diff
@@ -0,0 +1,246 @@
...
+if TYPE_CHECKING:
+    from pixeltable.catalog import Column
+
+
+class StorageObjectAddress(NamedTuple):
+    """Contains components of a media address.
+    Unused components are empty strings.
+    """
+
+    storage_target: str
```
aaron-siegel: Is this derivable from scheme? If so, do we really need it as a separate var?

pixeltable/utils/media_store_base.py
```diff
@@ -0,0 +1,29 @@
...
+
+from pathlib import Path
+from typing import TYPE_CHECKING, Optional
+from uuid import UUID
+
+if TYPE_CHECKING:
+    from pixeltable.catalog import Column
+
+
+class MediaStoreBase:
```
aaron-siegel: Decorate this as an `@abstractclass`, and then the `raise NotImplementedError` statements are not necessary.

The modern convention is:
```
class MyClass(ABC):
    @abstractmethod
    def my_abstract_method(self) -> None:
        """An abstract method."""
```
(nothing else is needed; the docstring serves as the method body)

pixeltable/utils/media_store_base.py
```diff
@@ -0,0 +1,29 @@
...
+    def count(self, tbl_id: UUID, tbl_version: Optional[int] = None) -> int:
+        """Count the number of media objects for a given table ID."""
+        raise NotImplementedError
+
+    def copy_local_media_file(self, col: Column, src_path: Path) -> str:
```
aaron-siegel: This makes it sound like you're copying a file from one local location to another. Why not `upload_media_object`?

pixeltable/utils/media_destination.py
```diff
@@ -0,0 +1,142 @@
...
+            f.write(data)
+            f.flush()  # Ensures Python buffers are written to OS
+            os.fsync(f.fileno())  # Forces OS to write to physical storage
+
+
+class MediaDestination:
+    @classmethod
+    def get_store(cls, dest: Optional[str], soa: Optional[StorageObjectAddress], col_name: Optional[str] = None) -> Any:
+        from pixeltable.utils.gcs_store import GCSStore
+        from pixeltable.utils.s3_store import S3Store
```
aaron-siegel: Is this going to work in a case where (for example) someone has botocore installed in their environment, but not google storage?

pixeltable/utils/media_destination.py
```diff
@@ -0,0 +1,142 @@
...
+            raise excs.Error(f'Column {col_name}: "destination" must be a string or path, got {dest!r}')
+        soa = MediaPath.parse_media_storage_addr(dest, may_contain_object_name=False)
+        if soa.storage_target == 'os':
+            return MediaStore.validate_destination(col_name, soa)
+        store = cls.get_store(dest, soa)
+        dest2 = store.validate_uri()
+        if dest2 is not None:
+            return dest2
+        raise excs.Error(
+            f'Column {col_name}: "destination" must be a valid URI to a supported destination, got {dest!r}'
```
aaron-siegel: Identifiers in backticks:
```
`destination`
```

pixeltable/utils/media_destination.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def download_media_object(self, src_path: str, dest_path: Path) -> None:
+        with urllib.request.urlopen(self.base_url + src_path) as resp, open(dest_path, 'wb') as f:
+            data = resp.read()
+            f.write(data)
+            f.flush()  # Ensures Python buffers are written to OS
+            os.fsync(f.fileno())  # Forces OS to write to physical storage
+
+
+class MediaDestination:
```
aaron-siegel: I'm not sure I see the need for a separate abstraction here. Everything inside here is a `@classmethod`, so why not (for example) just rename `MediaStoreBase` -> `MediaStore` and put these `@classmethod`s there?

pixeltable/utils/media_store.py
```diff
@@ -189,6 +261,17 @@ def stats(self) -> list[tuple[UUID, int, int, int]]:
...
 
+    def list_objects(self, return_uri: bool, n_max: int = 10) -> list[str]:
+        """Return a list of objects found with the specified location
+        Each returned object includes the full set of prefixes.
+        if return_uri is True, the full GCS URI is returned; otherwise, just the object key.
+        """
+        r = []
+        for root, _, files in os.walk(self.__base_dir):
+            for file in files:
+                r.append(os.path.join(root, file) if not return_uri else Path(root, file).as_uri())
```
aaron-siegel: Let's please try to avoid double negatives in conditionals (`if not ... else ...`)

pixeltable/utils/media_store.py
```diff
@@ -147,38 +207,50 @@ def save_media_object(self, data: bytes | PIL.Image.Image, col: Column, format:
         new_file_url = urllib.parse.urljoin('file:', urllib.request.pathname2url(str(dest_path)))
         return dest_path, new_file_url
 
-    def delete(self, tbl_id: UUID, tbl_version: Optional[int] = None) -> None:
+    def delete(self, tbl_id: UUID, tbl_version: Optional[int] = None) -> Optional[int]:
```
aaron-siegel, line 188: Update the docstring to explain the return value

pixeltable/utils/media_store.py
```diff
@@ -32,20 +34,78 @@ class MediaStore:
...
+            return None
+
+        path_str = urllib.parse.unquote(urllib.request.url2pathname(parsed.path))
+        return Path(path_str)
+
     @classmethod
-    def get(cls, base_uri: Optional[Path] = None) -> MediaStore:
-        """Get a MediaStore instance for the given base URI, or the environment's media_dir if None."""
+    def get(cls, base_uri: Optional[str] = None) -> MediaStore:
+        """Get a MediaStoreFile instance for the specified base URI, or the environment's media_dir if None."""
```
aaron-siegel: What's MediaStoreFile? I don't see other references to it in the code

pyproject.toml
aaron-siegel: Data sharing won't rely on boto since we're using presigned URLs.

pixeltable/utils/media_destination.py
jpeterson-pxt: Yes, these are our methods, not something external.

pixeltable/catalog/table_version.py
jpeterson-pxt: Comment added.

pixeltable/utils/media_store_base.py
jpeterson-pxt: In some cases, we are copying from one local file location to another (when the user has specified a local destination). In others we are uploading. Suggested verb then?
upload_or_copy_or_move_media_file()?

pixeltable/utils/media_store.py
jpeterson-pxt: This is a result of a change you requested in one of the other reviews (you wanted to remove the "MediaStoreFile" - which we are now going to call "FileSystemStore". This change was in another branch, and I just missed the renaming.

pixeltable/exec/object_store_save_node.py
jpeterson-pxt: Fixed

pixeltable/exec/object_store_save_node.py
jpeterson-pxt: @aaron-siegel  Suggestion?

jpeterson-pxt: This will be unnecessary after moving the clients into the global mechanism.

pixeltable/catalog/table.py
jpeterson-pxt: Of course. Other places in table.py import locally, but I didn't investigate why they do that.

pixeltable/exprs/data_row.py
jpeterson-pxt: Fixed

pixeltable/plan.py
jpeterson-pxt: Fixed

pixeltable/utils/media_path.py
jpeterson-pxt: fixed

pixeltable/utils/media_path.py
jpeterson-pxt: The derivation is a complex function of more than one value.
I'd prefer to compute it once and store it.
Not changed.

pixeltable/utils/media_destination.py
jpeterson-pxt: Done

pixeltable/utils/media_store.py

pixeltable/utils/media_store.py

pixeltable/utils/media_store.py
jpeterson-pxt: Fixed.

pixeltable/utils/media_destination.py
jpeterson-pxt: Is this good python, or even good object design?
This would make the director (which has different signatures for the methods to allow direction) the same class as the abstract representation of all the implementations (which are not redirectable).
I'd really like to keep the method names the same and the signatures different. It is much easier to understand.

pixeltable/utils/media_path.py
jpeterson-pxt: An enum is possible, but the name seems unambiguous for now. 
I expect to store clients in the Env client mechanism by composite name, e.g. r2_client, r2_resource, gs_client, etc

pixeltable/utils/client_container.py
jpeterson-pxt: Reading and writing clients have been shared. 
Parameter is now removed.

pixeltable/utils/media_destination.py
aaron-siegel: I'm really trying to avoid a proliferation of modules and class names with overlapping abstractions, which just makes it harder for someone else to follow what's going on in the code.

Another option is to keep `MediaStoreBase`, but rename `MediaDestination` to `MediaStore` and put the class methods there. (Or `ObjectStoreBase`/`ObjectStore` cf. our conversation this morning.) Then consolidate them into one module.

pixeltable/exec/object_store_save_node.py
aaron-siegel: What happened when you tried it?

pixeltable/utils/s3_store.py
```diff
@@ -0,0 +1,342 @@
...
+                # No client args supplibed, attempt to get default (s3) credentials
+                boto3.Session().get_credentials().get_frozen_credentials()
+                config = botocore.config.Config(**config_args)
+                return boto3.client('s3', config=config)  # credentials are available
+            else:
+                # If client args are provided, use them directly
+                config = botocore.config.Config(**config_args)
+                return boto3.client('s3', **client_args, config=config)
+        except AttributeError:
+            # No credentials available, use unsigned mode
```
aaron-siegel, line 343: This error handling confuses me. At what point will the `AttributeError` be raised? Is this the standard way of catching this?

If credentials are invalid, we want to be sure that the error gets raised back to the user and not swallowed (i.e., the behavior we want is credentials missing -> unsigned client; invalid credentials -> raise exception) - are we sure that's what will happen here?

tests/test_destination.py
```diff
@@ -0,0 +1,368 @@
...
+
+class TestDestination:
+    USE_S3 = False
+    USE_GS = False
+    USE_R2 = False
+
+    @classmethod
+    def base_dest(cls) -> Path:
+        """Return the base destination directory for tests"""
+        base_path = env.Env.get().tmp_dir / '..' / 'test_dest'
```
aaron-siegel: What path are you searching for here? Isn't `{tmp_dir}/..` the same thing as `Config.get().home` or am I missing something?

tests/test_destination.py
```diff
@@ -0,0 +1,368 @@
...
+
+import pixeltable as pxt
+import pixeltable.exceptions as excs
+from pixeltable import env
+from pixeltable.utils.media_destination import MediaDestination
+from pixeltable.utils.media_path import MediaPath
+
+
+class TestDestination:
+    USE_S3 = False
```
aaron-siegel: How is this testing pattern intended to work? Don't we want our tests to cover all three? Would we be better off using `pytest.mark.parametrize` here?

tests/test_destination.py
```diff
@@ -0,0 +1,368 @@
...
+        for s in [
+            's3://container',
+            f'wasb://container@{a_name}.blob.core.windows.net',
+            f'https://{a_name}.blob.core.windows.net/container',
+            f'https://{a_name}.r2.cloudflarestorage.com/container',
+            'https://raw.github.com',
+            'file://dir1/dir2/dir3',
+            'dir1/dir2/dir3',
+        ]:
+            r &= self.parse_two(s)
```
aaron-siegel: IMO this is flawed assertion design and will make things harder to debug in the event something fails. Better would be to have a separate assertion for each check.

I'd advocate for something like
```
def parse_uri(self, uri: str, consider_object: bool) -> None:
    try:
        ...
    except ValueError as e:
        raise AssertionError('Failed to parse URI: {url}') from e

def parse_uri_twice(self, uri: str) -> None:
    self.parse_uri(uri, True)
    self.parse_uri(uri, False)
```

And then
```
assert self.parse_uri_twice(...)
```
each time it's used. Note that I'm also using method and parameter names that are more likely to be meaningful to someone reading the code

tests/test_destination.py
```diff
@@ -0,0 +1,368 @@
...
+            'https://raw.github.com/pixeltable/pixeltable/main/docs/resources/images/000000000030.jpg', True
+        )
+
+        r &= self.parse_one('file://dir1/dir2/dir3', False)
+        r &= self.parse_one(f'file://dir1/dir2/dir3/{o_name}', True)
+        r &= self.parse_one(f'dir2/dir3/{o_name}', True)
+
+        assert r
+
+    #        assert False
```
aaron-siegel: spurious comment

tests/test_destination.py
```diff
@@ -0,0 +1,368 @@
...
+        print(r_dest)
+
+        assert len(r) == 2
+        assert len(r) == self.count(None, t._id)
+        assert len(r) == self.count(dest1_uri, t._id)
+
+        # The outcome of this test is unusual:
+        # When the column img_rot4 is ADDED, the computed result for existing rows is not identified
+        # as a duplicate, so it is double copied to the destination.
+        # When new rows are INSERTED, the results and destinations for img_rot3 and img_rot4 are identified
```
aaron-siegel, line 251: Is this the behavior that we want? Let's surface this and make a note of it.

tests/test_destination.py
```diff
@@ -0,0 +1,368 @@
...
+        assert len(r) == 2
+
+        # Copying a local file to the MediaStore is not allowed
+        assert self.count(None, t._id) == 0
+
+        # Ensure that local file is copied to a specified destination
+        assert len(r) == self.count(dest1_uri, t._id)
+
+    def test_dest_write_perf(self, reset_db: None) -> None:
+        """Test write performance with multiple concurrent requests"""
```
aaron-siegel: In what way does this test performance?

tests/test_destination.py
```diff
@@ -0,0 +1,368 @@
...
+
+        img_data = 'tests/data/imagenette2-160/ILSVRC2012_val_00000557.JPEG'  # ~10 KB
+        # img_data = 'tests/data/images/sewing-threads.heic'  # ~1.5 MB
```
aaron-siegel: Why are these commented out?

tests/test_destination.py
```diff
@@ -0,0 +1,368 @@
...
+            assert 3 * n == self.count(dest_uri, t._id)
+
+        # Ensure that all media is removed when the table is dropped
+        save_id = t._id
+        pxt.drop_table(t)
+        for i in range(1, 4):
+            _, dest_uri = self.dest(i)
+            assert self.count(dest_uri, save_id) == 0
+
+    @pytest.mark.skipif(os.getenv('TEST_DEST_ALL', '').lower() != 'true', reason='Requires credentials for all clouds')
```
aaron-siegel: Can we condition this on testing for the presence of actual credentials (i.e., do the clients successfully initialize?) rather than some bespoke environment variable (which then becomes an extra thing to have to manage in CI and so on?)

Once you're using `Env` to do client registration, you can use `skip_test_if_no_client()` for this; see how e.g. the OpenAI tests are conditioned on the presence of appropriate credentials.

tests/test_destination.py
```diff
@@ -0,0 +1,368 @@
...
+    def test_dest_all(self, reset_db: None) -> None:
+        """Test destination with all available storage targets"""
+        n = 1
+        dest_path = self.base_dest() / f'img_rot{n}'
+        if not dest_path.exists():
```
aaron-siegel: `dest_path.mkdir(exists_ok=True)`, please.

aaron-siegel (changes requested):

pixeltable/utils/media_store_base.py
jpeterson-pxt: Maybe throw something else? 
It is no longer an abstract class. 
Raise assertion error for methods that should not be called.

pixeltable/utils/media_store_base.py
jpeterson-pxt: split to move and copy

pixeltable/utils/media_destination.py
jpeterson-pxt: After renaming leave as is.

pixeltable/utils/s3_store.py
jpeterson-pxt: Straight up copy from s3.py
Does this occur when boto cannot find the credentials?
Research is required.

tests/test_destination.py
jpeterson-pxt: Until we figure out CI credentials. 
This was just for testing.
Write custom code to check the access for the URL's and conditionalize on each of them.

tests/test_destination.py
jpeterson-pxt: Remove from code altogether. 
Or put in tools. 
Done

tests/test_destination.py
jpeterson-pxt: Fix with cred checks.
If we cannot access the requested store, the test is skipped.

pixeltable/exec/cache_prefetch_node.py
mkornacker: I want to state this here again:
- the S3 thread pool needs to go into ExecContext and be shared across all plan nodes that make S3 requests
- the general recommendation for the appropriate number of threads seems to be ~30, not the # of cores

pixeltable/exec/cache_prefetch_node.py
```diff
@@ -26,16 +29,16 @@ class CachePrefetchNode(ExecNode):
     """Brings files with external URLs into the cache
 
     TODO:
-    - adapting the number of download threads at runtime to maximize throughput
+    - Adapt the number of download threads at runtime to maximize throughput
```
mkornacker: Remove this, a fixed number is fine.

Add instead: Create asyncio.Tasks to consume our input in order to increase concurrency.

mkornacker (changes requested):

pixeltable/plan.py
```diff
@@ -989,6 +1010,7 @@ def _create_query_plan(
                 if not agg_output.issuperset(exprs.ExprSet(eval_ctx.target_exprs)):
                     # we need an ExprEvalNode to evaluate the remaining output exprs
                     plan = exec.ExprEvalNode(row_builder, eval_ctx.target_exprs, agg_output, input=plan)
+                plan = cls._insert_save_node(tbl.tbl_version.id, row_builder.stored_media_cols, input_node=plan)
```
mkornacker, line 1013: Why would this show up in a query plan?

pixeltable/exec/object_store_save_node.py
```diff
@@ -0,0 +1,303 @@
...
+from pixeltable.utils.media_store import MediaStore, TempStore
+
+from .data_row_batch import DataRowBatch
+from .exec_node import ExecNode
+
+_logger = logging.getLogger('pixeltable')
+
+
+class ObjectStoreSaveNode(ExecNode):
+    """Brings files with external URLs into the cache
```
mkornacker: "into the cache"?

pixeltable/exec/object_store_save_node.py
```diff
@@ -0,0 +1,303 @@
...
+        self.file_col_info = file_col_info
+
+        self.num_returned_rows = 0
+        self.ready_rows = deque()
+        self.in_flight_rows = {}
+        self.in_flight_requests = {}
+        self.in_flight_work = {}
+        self.input_finished = False
+        self.row_idx = itertools.count() if retain_input_order else itertools.repeat(None)
+        self.num_threads = max(4, Env.get().cpu_count)
```
mkornacker: Same comment as for cacheprefetchnode.

pixeltable/exprs/data_row.py
```diff
@@ -257,42 +257,48 @@ def __setitem__(self, idx: int, val: Any) -> None:
             self.vals[idx] = val
         self.has_val[idx] = True
 
-    def flush_img(self, index: int, col: Optional[catalog.Column] = None) -> None:
-        """Save or discard the in-memory value (required to be a PIL.Image.Image)"""
+    def prepare_col_val_for_save(self, index: int, col: Optional[catalog.Column] = None) -> bool:
+        """
+        Prepare the column at the index for saving media val. Discard unneeded values.
```
mkornacker: What is 'index'?

pixeltable/utils/media_path.py
```diff
@@ -0,0 +1,249 @@
...
```
mkornacker: Is the meaning of the fields meant to be intuitively obvious? I don't really follow.

pixeltable/plan.py
jpeterson-pxt: I had exactly the same question, actually. 
It is here because there was a bit of code which wrote images to the old MediaStore in the AggregationNode.
That code (aggregation_node.py line 106) (a call to "flush_imgs") was removed and the SaveNode added. 
This was the only place in the code base where flush_imgs was called. 
The method is NOT called by any test in our test repo.

tests/test_destination.py
jpeterson-pxt: I can use Config.get().home. 
Anything which is outside of the tmp and media directories will be fine.
Done

tests/test_destination.py
jpeterson-pxt: Done

tests/test_destination.py
jpeterson-pxt: Fixed.

pixeltable/exec/cache_prefetch_node.py
jpeterson-pxt: All but sharing a single thread pool done.

pixeltable/exec/object_store_save_node.py
jpeterson-pxt: Found the correct solution.
Done.

pixeltable/utils/client_container.py
jpeterson-pxt: Done

pixeltable/utils/media_path.py
jpeterson-pxt: Done

pixeltable/utils/media_store_base.py
jpeterson-pxt: Done.

pixeltable/utils/s3_store.py
jpeterson-pxt: With a bit of research, this is indeed the error thrown when specific credentials are not available. 
Changed this a bit to attach to more errors, as we now also have ProfileNotFound.
Done

tests/test_destination.py
jpeterson-pxt: Done.

pixeltable/plan.py
jpeterson-pxt: @mkornacker Do you have an answer to this question?

pixeltable/exec/object_store_save_node.py
jpeterson-pxt: Done

pixeltable/utils/media_path.py
jpeterson-pxt: Commented.

pixeltable/exprs/data_row.py
jpeterson-pxt: It is the same as "idx" or "index" in the rest of the file. 
Changed.

pixeltable/exec/object_store_save_node.py
jpeterson-pxt: Done

pixeltable/exec/cache_prefetch_node.py
jpeterson-pxt: Done

tests/test_destination.py
jpeterson-pxt: Done

tests/test_destination.py
jpeterson-pxt: Done.

jpeterson-pxt:
Hopefully this is closer to the desired result. 
I have not combined the thread pools, though I believe that everything else is addressed.

pixeltable/utils/gcs_store.py
```diff
@@ -0,0 +1,287 @@
...
+from pixeltable import env, exceptions as excs
+from pixeltable.utils.object_stores import ObjectPath, ObjectStoreBase, StorageObjectAddress, StorageTarget
+
+if TYPE_CHECKING:
+    from pixeltable.catalog import Column
+
+_logger = logging.getLogger('pixeltable')
+
+
+@env.register_client('gs')
```
aaron-siegel: I'd prefer to call it `'gcs_store'`

pixeltable/utils/gcs_store.py
```diff
@@ -0,0 +1,287 @@
...
+        return self.__prefix_name
+
+    def validate(self, error_col_name: str) -> Optional[str]:
+        """
+        Checks if the URI exists.
+
+        Returns:
+            str: The base URI if the GCS bucket exists and is accessible, None otherwise.
+        """
+        env.Env.get().require_package('google.cloud.storage')
```
aaron-siegel: Should we also be doing `require_package()` when the column is created?

pixeltable/utils/local_store.py
```diff
@@ -0,0 +1,305 @@
...
+            return None
+
+        path_str = urllib.parse.unquote(urllib.request.url2pathname(parsed.path))
+        return Path(path_str)
+
+    @classmethod
+    def from_soa(cls, soa: StorageObjectAddress) -> LocalStore:
+        """Return an instance for the specified base URI. The destination is sourced from Env if None."""
+        r = LocalStore(soa.to_path)
+        r.soa = soa
```
aaron-siegel: Just make this an optional initialization parameter?

pixeltable/utils/object_stores.py
```diff
@@ -0,0 +1,482 @@
...
+from pixeltable import exceptions as excs
+
+if TYPE_CHECKING:
+    from pixeltable.catalog import Column
+
+
+class StorageTarget(enum.Enum):
+    """Enumeration of storage kinds."""
+
+    OS = 'os'  # Local file system
```
aaron-siegel: I prefer more descriptive names LOCAL, GCS, AZURE in place of OS, GS, AZ; this would be more consistent with our other enum patterns

pixeltable/utils/object_stores.py
```diff
@@ -0,0 +1,482 @@
...
+            https://raw.github.com/pixeltable/pixeltable/main/docs/resources/images/000000000030.jpg
+        """
+        soa = cls.parse_object_storage_addr1(src_addr)
+        prefix, object_name = cls.separate_prefix_object(soa.key, may_contain_object_name)
+        assert not object_name.endswith('/')
+        r = soa._replace(prefix=prefix, object_name=object_name)
+        return r
+
+
+class ObjectStoreBase:
```
aaron-siegel, line 272: Make this a subclass of `abc.ABC` and add `@abstractmethod` decorators where appropriate (with empty method body)

pixeltable/utils/s3_store.py
```diff
@@ -0,0 +1,337 @@
...
+    from botocore.exceptions import ClientError
+
+    from pixeltable.catalog import Column
+
+_logger = logging.getLogger('pixeltable')
+
+
+@env.register_client('r2')
+def _() -> Any:
+    profile_name = Config.get().get_string_value('r2_profile') or 'r2_profile'
```
aaron-siegel: I don't think we should be setting our own default profile name - just let the boto client decide if one isn't specified.

pixeltable/utils/s3_store.py
```diff
@@ -0,0 +1,337 @@
...
+
+@env.register_client('r2_resource')
+def _() -> Any:
+    profile_name = Config.get().get_string_value('r2_profile') or 'r2_profile'
+    return S3Store.create_boto_resource(profile_name=profile_name)
+
+
+@env.register_client('s3')
+def _() -> Any:
+    return S3Store.get_boto_client()
```
aaron-siegel: Maybe add an s3_profile for parity with r2_profile (an optional one, per the preceding comment)

Also, make sure to document any new config options in `configuration.mdx`.

pixeltable/utils/s3_store.py
```diff
@@ -0,0 +1,337 @@
...
+        except ClientError as e:
+            self.handle_s3_error(e, self.bucket_name, f'deleting with {self.prefix}')
+            raise
+
+    def list_objects(self, return_uri: bool, n_max: int = 10) -> list[str]:
+        """Return a list of objects found in the specified destination bucket.
+        Each returned object includes the full set of prefixes.
+        if return_uri is True, full URI's are returned; otherwise, just the object keys.
+        """
+        from botocore.exceptions import ClientError
```
aaron-siegel: I assume the reason we have to keep declaring this in each method is that the `s3_store` module is getting pulled in somewhere even if boto is not installed?

Would it be easier to manage the `s3_store` import? i.e., only pull in `s3_store` itself at the time we need to do something boto-related.

tests/test_destination.py
```diff
@@ -0,0 +1,359 @@
...
+            return False
+
+    USE_LOCAL_DEST = 'fs'
+    USE_GS_DEST = 'gs'
+    USE_S3_DEST = 's3'
+    USE_R2_DEST = 'r2'
+    USE_AZURE_DEST = 'az'
+
+    @staticmethod
+    def dest2(n: int, dest_id: int | str) -> tuple[Path | str | None, str | None]:
```
aaron-siegel: Can we please give this a more meaningful name than `dest2`?

tests/test_destination.py
```diff
@@ -0,0 +1,359 @@
...
+            base_path.mkdir(exist_ok=True)
+            dest_path = base_path / f'img_rot{n}'
+            dest_path.mkdir(exist_ok=True)
+            dest_uri = dest_path.resolve().as_uri()
+            return dest_path, dest_uri
+        if dest_id == 'gs':
+            gs_uri = f'gs://pixeltable/my_folder/img_rot{n}'
+            return gs_uri, gs_uri
+        elif dest_id == 's3':
+            s3_uri = f's3://jimpeterson-test/img_rot{n}'
```
aaron-siegel: We'll want to use a `pxttest` bucket or some such here

tests/test_destination.py
```diff
@@ -0,0 +1,359 @@
...
+        print(us.row_count_stats)
+        print(us.cascade_row_count_stats)
+        print('============================================================')
+
+    def test_dest_errors(self, reset_db: None) -> None:
+        t = pxt.create_table('test_dest_errors', schema={'img': pxt.Image})
+        valid_dest = 'tests/data/'
+
+        # Basic tests of the destination parameter: types and store / computed
+        with pytest.raises(excs.Error, match='must be a string or path'):
```
aaron-siegel: In testing code, use `pxt.Error`, not `excs.Error`, to match the public API.

tests/test_destination.py
```diff
@@ -0,0 +1,359 @@
...
+        for s in [
+            's3://container',
+            f'wasb://container@{a_name}.blob.core.windows.net',
+            f'https://{a_name}.blob.core.windows.net/container',
+            f'https://{a_name}.r2.cloudflarestorage.com/container',
+            'https://raw.github.com',
+            'file://dir1/dir2/dir3',
+            'dir1/dir2/dir3',
+        ]:
+            r &= self.parse_object_addr_two_ways(s)
```
aaron-siegel: There's no longer a need for the binary intersection. Just call `self.parse_object_addr_two_ways(s)` and that's enough. The binary intersection can never actually become `False` because one of the wrapped assertions will have already fired.

tests/test_destination.py
```diff
@@ -0,0 +1,359 @@
...
+        try:
+            r = ObjectPath.parse_object_storage_addr(s, consider_object)
+            print(f'Success: {r!r}')
+            return True
+        except ValueError as e:
+            print(f'Error: {e!r}')
+            return False
+
+    def parse_object_addr_two_ways(self, s: str) -> bool:
+        r = self.parse_object_addr(s, False)
```
aaron-siegel: There's no need for the `r` variable here or a binary intersection. Just the following should be sufficient.

```
assert self.parse_object_addr(s, False)
assert self.parse_object_addr(s, True)
```

tests/test_destination.py
```diff
@@ -0,0 +1,359 @@
...
+        print(f'Parsing: {s}')
+        try:
+            r = ObjectPath.parse_object_storage_addr(s, consider_object)
+            print(f'Success: {r!r}')
+            return True
+        except ValueError as e:
+            print(f'Error: {e!r}')
+            return False
+
+    def parse_object_addr_two_ways(self, s: str) -> bool:
```
aaron-siegel: Maybe call this `validate_object_addr`.

tests/test_destination.py
```diff
@@ -0,0 +1,359 @@
...
+    def test_dest_all(self, reset_db: None) -> None:
+        """Test destination with all available storage targets"""
+        n = 1
+        _, lc_uri = self.dest2(n, self.USE_LOCAL_DEST)
+        c2_uri = self.get_valid_dest(n, self.USE_GS_DEST, lc_uri)
+        c3_uri = self.get_valid_dest(n, self.USE_S3_DEST, lc_uri)
+        c4_uri = self.get_valid_dest(n, self.USE_R2_DEST, lc_uri)
+        t = pxt.create_table('test_dest', schema={'img': pxt.Image})
+        t.insert([{'img': 'tests/data/imagenette2-160/ILSVRC2012_val_00000557.JPEG'}])
+        t.add_computed_column(**{'img_rot_1': t.img.rotate(90)}, destination=lc_uri)
```
aaron-siegel: Why the `**` syntax? Just this should work

`t.add_computed_column(img_rot_1=t.img.rotate(90), destination=lc_uri)`

uv.lock
```diff
@@ -6660,8 +6785,8 @@ name = "soundfile"
 version = "0.13.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "cffi", marker = "python_full_version < '4'" },
-    { name = "numpy", marker = "python_full_version < '4'" },
+    { name = "cffi", marker = "python_full_version < '3.13'" },
```
aaron-siegel, line 6903: There are a large number of changes to `uv.lock` that don't seem related to the GCS changes. Sometimes this happens, but, can you try overwriting `uv.lock` with a copy from `main` and re-running `uv sync` just to be safe?

jpeterson-pxt: I will do this after I merge main.

pixeltable/utils/gcs_store.py
jpeterson-pxt: This method is called to validate the address when the column is created.

pixeltable/utils/gcs_store.py
jpeterson-pxt: Do you want to do the same for 's3' and 'r2' clients so they remain consistent?

pixeltable/utils/object_stores.py
jpeterson-pxt: I wanted to keep it consistent with the store names (s3, r2, gs, etc)

pixeltable/utils/local_store.py
jpeterson-pxt: Done - merged into init.

tests/test_destination.py
jpeterson-pxt: Done

tests/test_destination.py
jpeterson-pxt: done

tests/test_destination.py
jpeterson-pxt: done

tests/test_destination.py
jpeterson-pxt: done

tests/test_destination.py
jpeterson-pxt: Done

tests/test_destination.py
jpeterson-pxt: Yes when CI is setup.

tests/test_destination.py
jpeterson-pxt: Done

pixeltable/utils/gcs_store.py
jpeterson-pxt: Done

pixeltable/utils/object_stores.py
jpeterson-pxt: Done

pixeltable/utils/s3_store.py
jpeterson-pxt: This works. The code will be somewhat simpler. It has the side effect of checking if boto is installed every time we want to do an operation.
I will do the same for the other targets as well.

jpeterson-pxt: Done

tests/test_destination.py
jpeterson-pxt: Done

pixeltable/utils/object_stores.py
jpeterson-pxt: There are no methods which are implemented by all stores.
- HTTPStore implements only one method
- LocalStore does not implement that method.
:-(

pixeltable/utils/s3_store.py
jpeterson-pxt: Done

pixeltable/utils/s3_store.py
jpeterson-pxt: Done

tests/test_destination.py
jpeterson-pxt: Done.

jpeterson-pxt:
I need to do some more cleanup on the S3 credentials.
You can hold off on reviewing.

Done. Ok to review.

docs/mintlify/docs/overview/configuration.mdx
```diff
@@ -45,6 +45,8 @@ api_key = 'my-label-studio-api-key'
 | PIXELTABLE_TIME_ZONE | [pixeltable]<br/>time_zone | (string) Default time zone in [IANA format](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones); defaults to the system time zone |
 | PIXELTABLE_HIDE_WARNINGS | [pixeltable]<br/>hide_warnings | (bool) Suppress warnings generated by various libraries used by Pixeltable; default is false |
 | PIXELTABLE_VERBOSITY | [pixeltable]<br/>verbosity | (int) Verbosity for Pixeltable console logging (0: minimum, 1: normal, 2: maximum); default is 1 |
+| PIXELTABLE_R2_PROFILE | [pixeltable]<br/>r2_profile_name | (string) Name of AWS config profile to use when accessing Cloudflare R2 resources |
```
aaron-siegel: Maybe add a note that if it's not specified, it will default to $AWS_PROFILE or the default AWS credentials

pixeltable/config.py
```diff
@@ -99,7 +99,13 @@ def __create_default_config(cls, config_path: Path) -> dict[str, Any]:
         free_disk_space_bytes = shutil.disk_usage(config_path.parent).free
         # Default cache size is 1/5 of free disk space
         file_cache_size_g = free_disk_space_bytes / 5 / (1 << 30)
-        return {'pixeltable': {'file_cache_size_g': round(file_cache_size_g, 1), 'hide_warnings': False}}
+        return {
+            'pixeltable': {
+                'file_cache_size_g': round(file_cache_size_g, 1),
+                'hide_warnings': False,
+                'r2_profile': 'r2_profile',
```
aaron-siegel: I don't think we should be specifying this here. Just leave it undefined unless the user specifies it and it will fall back on $AWS_PROFILE. An _a priori_ imposition of some ad hoc (nonstandard) profile name is likely to confuse users.

pixeltable/env.py
```diff
@@ -615,15 +619,17 @@ def get_client(self, name: str) -> Any:
         Args:
             - name: The name of the client
         """
-        cl = _registered_clients[name]
-        if cl.client_obj is not None:
-            return cl.client_obj  # Already initialized
-
-        # Construct a client, retrieving each parameter from config.
+        # Return the existing client if itsy has already been constructed
```
aaron-siegel: typo (itsy)

pixeltable/utils/gcs_store.py
```diff
@@ -0,0 +1,282 @@
...
+    soa: StorageObjectAddress
+
+    def __init__(self, soa: StorageObjectAddress):
+        assert soa.storage_target == StorageTarget.GCS_STORE, f'Expected storage_target "gs", got {soa.storage_target}'
+        self.soa = soa
+        self.__base_uri = soa.prefix_free_uri + soa.prefix
+        self.__bucket_name = soa.container
+        self.__prefix_name = soa.prefix
+
+    def client(self) -> Any:
```
aaron-siegel: Better to give this a proper type signature. See how this is done in e.g. openai.py.

Also, shouldn't this be a `@classmethod`?

pixeltable/utils/gcs_store.py
```diff
@@ -0,0 +1,282 @@
...
+            if 'Precondition' in error_message:
+                raise excs.Error(f'Precondition failed for bucket {bucket_name} during {operation}: {error_message}')
+            else:
+                raise excs.Error(f'Error during {operation} in bucket {bucket_name}: {error_message}')
+        else:
+            # Generic error handling
+            raise excs.Error(f'Unexpected error during {operation} in bucket {bucket_name}: {str(e)!r}')
+
+    @classmethod
+    def create_client(cls) -> Any:
```
aaron-siegel: Why is this here instead of inlined in the register_client implementation? Do we also need to call it from somewhere else?

pixeltable/utils/local_store.py
```diff
@@ -0,0 +1,312 @@
...
+            self.soa = location
+
+    def validate(self, error_col_name: str) -> str:
+        """Convert a Column destination parameter to a URI, else raise errors."""
+        path_str = self.soa.to_path
+        dest_path = Path(path_str)
+
+        # Check if path exists and validate it's a directory
+        if not dest_path.exists():
+            raise excs.Error(f'{error_col_name}`destination` does not exist')
```
aaron-siegel, line 55: It looks like what will be printed here is:

```
my_col`destination` does not exist
```

Is that the intended formatting?

pixeltable/utils/local_store.py
```diff
@@ -0,0 +1,312 @@
...
+        if dest_path.is_absolute():
+            # Convert to file URI
+            return dest_path.as_uri()
+
+        # For relative paths, convert to absolute first
+        try:
+            absolute_path = dest_path.resolve()
+            return absolute_path.as_uri()
+        except (OSError, ValueError) as e:
+            raise excs.Error(f'{error_col_name}`destination` must be a valid path. Error: {str(e)!r}') from None
```
aaron-siegel: Usual formatting is `Error: {e}`; it looks out of place to quote the entire error message

pixeltable/utils/local_store.py
```diff
@@ -0,0 +1,312 @@
...
+
+    @staticmethod
+    def file_url_to_path(url: str) -> Optional[Path]:
+        """Convert a file:// URI to a Path object with support for Windows UNC paths."""
+        assert isinstance(url, str), type(url)
+        parsed = urllib.parse.urlparse(url)
+
+        # Verify it's a file scheme
```
aaron-siegel, line 77: Is this duplicating existing logic in `utils.parse_local_file_path()`?

pixeltable/utils/object_stores.py
```diff
@@ -0,0 +1,489 @@
...
+
+
+class ObjectPath:
+    PATTERN = re.compile(r'([0-9a-fA-F]+)_(\d+)_(\d+)_([0-9a-fA-F]+)')  # tbl_id, col_id, version, uuid
+
+    @classmethod
+    def table_prefix(cls, tbl_id: UUID) -> str:
+        """Construct a unique unix-style prefix for objects in a table (without leading/trailing slashes)."""
+        assert isinstance(tbl_id, uuid.UUID)
+        return f'{tbl_id.hex}'
```
aaron-siegel: `f'{tbl_id.hex}'` means exactly the same thing as `tbl_id.hex`

pixeltable/utils/object_stores.py
```diff
@@ -0,0 +1,489 @@
...
+
+    @classmethod
+    def prefix_raw(cls, tbl_id: UUID, col_id: int, tbl_version: int, ext: Optional[str] = None) -> tuple[str, str]:
```
aaron-siegel: I'd prefer a more active verb here like `create_prefix_raw` to make it clear we're constructing something new.

pixeltable/utils/s3_store.py
```diff
@@ -0,0 +1,356 @@
...
+            raise excs.Error(f'Bucket {bucket_name} not found during {operation}: {error_message}')
+        elif error_code == '403':
+            raise excs.Error(f'Access denied to bucket {bucket_name} during {operation}: {error_message}')
+        elif error_code == 'PreconditionFailed' or 'PreconditionFailed' in error_message:
+            raise excs.Error(f'Precondition failed for bucket {bucket_name} during {operation}: {error_message}')
+        else:
+            raise excs.Error(f'Error during {operation} in bucket {bucket_name}: {error_code} - {error_message}')
+
+    @classmethod
+    def get_boto_session(cls, profile_name: Optional[str] = None) -> Any:
```
aaron-siegel: create_boto_session (right now there's an inconsistency w/ create_boto_resource; we should make them consistent, and create is more descriptive to me)

pixeltable/utils/s3_store.py
```diff
@@ -0,0 +1,356 @@
...
+                print(f'Creating boto session with profile {profile_name}')
+                session = boto3.Session(profile_name=profile_name)
+                return session
+            except Exception as e:
+                print(f'Error occurred while creating boto session with profile {profile_name}: {e}')
+                _logger.info(f'Error occurred while creating boto session: {e}, fallback to default profile')
+        return boto3.Session()
+
+    @classmethod
+    def get_boto_client(cls, profile_name: Optional[str] = None, extra_args: Optional[dict[str, Any]] = None) -> Any:
```
aaron-siegel: create_boto_client

scripts/setup_ci_credentials.sh
```diff
@@ -0,0 +1,37 @@
...
+#    session = boto3.Session(profile_name='r2_profile')
+
+# Create AWS directory if needed
+mkdir -p ~/.aws
+
+# Remove any existing config or credentials files
+rm -rf ~/.aws/config
+rm -rf ~/.aws/credentials
+
+# Add default and r2_profile to AWS config
```
aaron-siegel: I don't see r2_profile being added? (Maybe it's not necessary; just pointing out that the comment and code are inconsistent)

aaron-siegel (changes requested):

pixeltable/env.py
jpeterson-pxt: Done

pixeltable/utils/gcs_store.py
jpeterson-pxt: Done

pixeltable/utils/gcs_store.py
jpeterson-pxt: Done

pixeltable/utils/local_store.py
jpeterson-pxt: Done

pixeltable/utils/local_store.py
jpeterson-pxt: Done

pixeltable/utils/local_store.py
jpeterson-pxt: Not exactly enough... :-)

pixeltable/utils/object_stores.py
jpeterson-pxt: Done.

pixeltable/utils/s3_store.py
jpeterson-pxt: Done

pixeltable/utils/s3_store.py
jpeterson-pxt: Done

scripts/setup_ci_credentials.sh
jpeterson-pxt: Done.

pixeltable/utils/object_stores.py
jpeterson-pxt: Done

docs/mintlify/docs/overview/configuration.mdx
jpeterson-pxt: Done

pixeltable/config.py
jpeterson-pxt: Done

aaron-siegel (approved):

PR #767, opened 2025-08-15
[WIP] GCS support, further cleanup of S3 support 

jpeterson-pxt:
Adds support for GCS destinations, as well as the direct inclusion of GCS media objects in a pixeltable table insert command.
Overlaid on an obsolete work version of the S3 media destination support.


jpeterson-pxt:
No longer needed.

PR #768, opened 2025-08-15
More video built-ins

mkornacker:
Contains:
- clip()
- get_frame()
- segment_video()
- concat_videos()
- overlay_text()
- VideoSplitter


pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
     return metadata
 
 
+def _get_video_duration(path: str) -> float | None:
```
aaron-siegel: This duplicates logic that currently lives in `__get_stream_metadata`; can it be consolidated?

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+            video_stream = container.streams.video[0]
+            if video_stream.duration is None or video_stream.time_base is None:
+                return None
+            return float(video_stream.duration * video_stream.time_base)
+    except Exception:
+        return None
+
+
+@pxt.udf(is_method=True)
+def get_frame(video: pxt.Video, *, timestamp: float) -> PIL.Image.Image | None:
```
aaron-siegel: Just a comment, it's definitely possible to implement this one using pyav, since it's identical to the seek-and-extract operation that already works in `FrameIterator`.

I don't love the explicit calls to `ffmpeg` as they're likely to lead to system-dependent "gotchas" that the pyav authors have already worked through over the years.

I think `extract_frame` would be a better name - it's not really a getter; it's a data transcoding operation.

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+            '-map',
+            '0',  # Copy all streams from input
+            output_path,
+        ]
+    )
+    return cmd
+
+
+@pxt.udf(is_method=True)
+def get_clip(
```
aaron-siegel: `clip` would be a better name; this isn't a getter. (We don't say `get_substring()` in the string library, for example)

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+    video: pxt.Video, *, start_time: float, end_time: float | None = None, duration: float | None = None
+) -> pxt.Video | None:
+    """
+    Extract a clip from a video, specified by start_time and either end_time or duration (in seconds).
```
aaron-siegel: Literals (start_time, end_time, None, etc) should be enclosed in backticks in docstrings

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+
+    Args:
+        video: Input video file
+        start_time: Start time in seconds
+        end_time: End time in seconds (if None, goes to end of video)
```
aaron-siegel: This is a false docstring; if `end_time` is None, it doesn't necessarily mean that it goes to the end of the video, since `duration` may be specified instead.

Clearer to just say "End time in seconds" here, and then clarify in the docstring body that at most one of end_time and duration may be specified (which isn't currently mentioned anywhere), and if neither is then the default behavior is end-of-video

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+    if end_time is not None:
+        duration = end_time - start_time
+    cmd = _ffmpeg_clip_cmd(str(video), output_path, start_time, duration)
+
+    try:
+        result = subprocess.run(
+            cmd,
+            capture_output=True,
+            text=True,
+            timeout=300,  # 5 minute timeout
```
aaron-siegel: Should this be parameterized? Just in case someone has an enormous video they're working with? I don't know how long these operations normally take

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+            if result.stderr is not None:
+                error_msg += f'. ffmpeg stderr: {result.stderr.strip()}'
+            raise pxt.Error(error_msg)
+
+        # check for indications of errors in stderr
+        stderr_output = result.stderr.strip() if result.stderr is not None else ''
+        if len(stderr_output) > 0 and any(
+            error_word in stderr_output.lower() for error_word in ['error', 'failed', 'invalid']
+        ):
+            raise pxt.Error(f'ffmpeg reported errors: {stderr_output}')
```
aaron-siegel: Again... it would be so much better if this could all be done with pyav, which has likely sifted through all the permutations of this sort of thing over the years

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+
+
+@pxt.udf(is_method=True)
+def get_segments(video: pxt.Video, *, duration: float) -> list[str]:
+    """
+    Split a video into fixed-size segments.
+
+    __Requirements:__
+
+    - `ffmpeg` needs to be installed and in PATH
```
aaron-siegel, line 370: Why is this requirement mentioned only here (and not in the other ffmpeg-dependent udfs)?

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+        if e.stderr is not None:
+            error_msg += f': {e.stderr.strip()}'
+        raise pxt.Error(error_msg) from e
+
+    except subprocess.TimeoutExpired:
+        raise pxt.Error('ffmpeg timed out for command line {" ".join(cmd)}') from None
+
+
+@pxt.udf(is_method=True)
+def get_segments(video: pxt.Video, *, duration: float) -> list[str]:
```
aaron-siegel: Same comment as above - not a getter

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+        pxt.Error: If the video is missing timing information.
+
+    Examples:
+        Split a video at 1 minute intervals
+
+        >>> tbl.select(segment_paths=tbl.video.get_segments(duration=60)).collect()
+
+        Split video into two parts at the midpoint:
+
+        >>> duration = tbl.video.get_metadata()streams[0].duration_seconds
```
aaron-siegel: Missing dot separator after get_metadata()

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+                text=True,
+                timeout=300,  # 5 minute timeout
+                check=True,
+            )
+            result.append(segment_path)
+            start_time += duration
+
+        return result
+
+    except subprocess.CalledProcessError as e:
```
aaron-siegel, line 418: There was a bunch of other error handling logic in `clip()` that isn't replicated here. Can the error handling logic be consolidated so we're sure we're catching all error conditions in all places?

pixeltable/functions/video.py
```diff
@@ -283,6 +287,237 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+    ]
+
+    try:
+        result = subprocess.run(cmd, capture_output=True, text=True, check=False)
+
+        if result.returncode != 0:
+            raise pxt.Error(f'get_frame(): ffmpeg failed with error: {result.stderr}')
+        if not output_path.exists():
+            return None
+        return PIL.Image.open(output_path)
```
aaron-siegel: Call `image.load()` before returning? I've seen weird errors in the past if `load()` is not called right away (for example, if the temp file gets moved/purged before the image is operated upon)

tests/test_video.py
```diff
@@ -282,3 +288,129 @@ def test_make_video(self, reset_db: None) -> None:
...
+            _ = t.select(invalid_clip=t.video.get_clip(start_time=10.0, end_time=5.0)).collect()
+
+        with pytest.raises(pxt.Error, match='duration must be positive'):
+            _ = t.select(invalid_clip=t.video.get_clip(start_time=10.0, duration=-1.0)).collect()
+
+        with pytest.raises(pxt.Error, match='end_time and duration cannot both be specified'):
+            _ = t.select(invalid_clip=t.video.get_clip(start_time=10.0, end_time=20.0, duration=10.0)).collect()
+
+    def test_get_frame(self, reset_db: None) -> None:
+        skip_test_if_not_in_path('ffmpeg')
```
aaron-siegel, line 342: I think this should be unnecessary (I think `ffmpeg` is installed by `make install` as well as in all CI contexts)

pixeltable/functions/video.py
mkornacker: Done.

pixeltable/functions/video.py

pixeltable/functions/video.py

pixeltable/functions/video.py
mkornacker: Reimplemented with av.

Moviepy calls it get_frame(). We can change it, but it's not clear that users care about the meaning of 'getter'.

pixeltable/functions/video.py
mkornacker: Changed in response to user feedback.

pixeltable/functions/video.py

pixeltable/functions/video.py

pixeltable/functions/video.py
mkornacker: I tried but haven't been able to do it (while preserving all streams).

pixeltable/functions/video.py
aaron-siegel: Well, I do think consistency in how we name our UDFs is important. These things are much harder to change after the fact. Every programming language has its share of hastily named functions that everyone has come to wish were different; it seems worth a little extra thought to minimize the risk of that in Pixeltable.

I'm less opinionated about `get_frame()` than the others. I can see a case for it, though I still don't really agree. Note we have very few UDFs that start with `get_` and the ones that do (such as `get_metadata()`) really do just pluck out attributes of the structure. Skimming over a list of Postgres functions, I don't see a single one that starts with `get_`.

All I'm saying is let's be thoughtful about naming our UDFs in a way that gives the Pixeltable UDF library a consistent and intuitive feel. Even if users don't seem to care about a particular naming decision such as get_frame vs whatever, I suspect they'll ultimately appreciate an API that's well thought out as a whole.

pixeltable/functions/video.py
aaron-siegel: Ok I believe this is a tough one. It's sad that pyav doesn't have an easy way to do this :(

pixeltable/functions/video.py
aaron-siegel: I do feel more strongly that this one should not start with `get_`.

Another option is `split()`, as in `str.split()`. The analogous iterator is `AudioSplitter` and we also have a more loosely analogous `DocumentSplitter`. I'd imagine if this were an iterator it'd be `VideoSplitter`, so `split` makes a lot of sense to me.

I'd suggest adding an `overlap: float = 0.0` parameter, and perhaps a `min_duration` parameter or some such, mirroring what we did for `AudioSplitter`.

.github/workflows/pytest.yml
```diff
@@ -129,7 +129,7 @@ jobs:
         # cmake is needed for certain CI environments where it is not preinstalled.
         # libiconv and ffmpeg are needed by WhisperX.
         if: ${{ matrix.uv-options == '' }}
-        run: ./scripts/retry.sh 3 30 conda install -c conda-forge 'cmake>=3.22' libiconv ffmpeg
+        run: ./scripts/retry.sh 3 30 conda install -c conda-forge 'cmake>=3.22' libiconv 'ffmpeg>=6,<7'
```
aaron-siegel, line 133: Also make this change in `nightly.yml` or the nightly tests will break.

I'm thinking we might want to hard-code this to a specific version (say `6.1.1`) both here and in the Makefile, to ensure parity between dev/CI (so we don't end up in another situation where something works in dev but then fails in CI)

aaron-siegel: ping (at minimum, add this change to `nightly.yml` or the nightly job will break)

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+            return None
+
+    except Exception as e:
+        raise pxt.Error(f'extract_frame(): failed to extract frame: {e}') from e
+
+
+def _handle_ffmpeg_error(e: subprocess.CalledProcessError) -> NoReturn:
+    error_msg = f'ffmpeg failed with return code {e.returncode}'
+    if e.stderr is not None:
+        error_msg += f': {e.stderr.strip()}'
```
aaron-siegel: Change to `f':\n{e.stderr.strip()}'` (i.e. separate via a newline?)

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+
+
+@pxt.udf(is_method=True)
+def clip(
+    video: pxt.Video, *, start_time: float, end_time: float | None = None, duration: float | None = None
+) -> pxt.Video | None:
+    """
+    Extract a clip from a video, specified by `start_time` and either `end_time` or `duration` (in seconds).
+
+    If `start_time` is beyond the end of the video, returns None. Can only specify one of `end_time` and `duration`.
```
aaron-siegel, line 315: Prefer complete sentences...

If `start_time` is beyond the end of the video, returns None.

At most one of `end_time` or `duration` may be specified; if neither is specified, the clip goes to the end of the video.

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+        pxt.Error: If the video is missing timing information.
+
+    Examples:
+        Split a video at 1 minute intervals
+
+        >>> tbl.select(segment_paths=tbl.video.segment_video(duration=60)).collect()
+
+        Split video into two parts at the midpoint:
+
+        >>> duration = tbl.video.get_metadata().streams[0].duration_seconds
```
aaron-siegel: Can we use the new `tbl.video.get_video_duration()` UDF to simplify this?

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+    for video in videos:
+        metadata = av_utils.get_metadata(str(video))
+        video_stream = next((stream for stream in metadata['streams'] if stream['type'] == 'video'), None)
+        if video_stream is None:
+            raise pxt.Error(f'concat_videos(): file {video} has no video stream')
+        resolutions.append((video_stream['width'], video_stream['height']))
+
+    # check for divergence
+    first_res = resolutions[0]
+    for i, res in enumerate(resolutions[1:], start=1):
```
aaron-siegel: Don't you want `enumerate(resolutions, start=1)`? As written I believe it will start with video 2 and shift the indices by one.

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+    if not shutil.which('ffmpeg'):
+        raise pxt.Error('ffmpeg is not installed or not in PATH. Please install ffmpeg to use concat_videos().')
+
+    # Check that all videos have the same resolution
+    resolutions: list[tuple[int, int]] = []
+    for video in videos:
+        metadata = av_utils.get_metadata(str(video))
+        video_stream = next((stream for stream in metadata['streams'] if stream['type'] == 'video'), None)
+        if video_stream is None:
+            raise pxt.Error(f'concat_videos(): file {video} has no video stream')
```
aaron-siegel: `file {video!r}` ?

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+        resolutions.append((video_stream['width'], video_stream['height']))
+
+    # check for divergence
+    first_res = resolutions[0]
+    for i, res in enumerate(resolutions[1:], start=1):
+        if res != first_res:
+            raise pxt.Error(
+                f'concat_videos(): requires that all videos have the same resolution, but '
+                f'video 0 ({videos[0]}) has resolution {first_res[0]}x{first_res[1]} and '
```
aaron-siegel: I'd break this error message onto multiple lines for readability, something like this:
```
f'concat_videos(): requires that all videos have the same resolution, but:\n'
f'  video 0 has resolution {x0}x{y0}: {videos[0]}\n'
f'  video {i} has resolution {x}x{y}: {videos[i]}'
```

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+    resolutions: list[tuple[int, int]] = []
+    for video in videos:
+        metadata = av_utils.get_metadata(str(video))
+        video_stream = next((stream for stream in metadata['streams'] if stream['type'] == 'video'), None)
+        if video_stream is None:
+            raise pxt.Error(f'concat_videos(): file {video} has no video stream')
+        resolutions.append((video_stream['width'], video_stream['height']))
+
+    # check for divergence
+    first_res = resolutions[0]
```
aaron-siegel: I'd do this for code readability (it's preferred for tuples since the "indices" don't really mean anything):
```
x0, y0 = resolutions[0]
for i, (x, y) in enumerate(...):
```

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+                f'concat_videos(): requires that all videos have the same resolution, but '
+                f'video 0 ({videos[0]}) has resolution {first_res[0]}x{first_res[1]} and '
+                f'video {i} ({videos[i]}) has resolution {res[0]}x{res[1]}.'
+            )
+
+    # ffmpeg -f concat needs an input file list
+    filelist_path = TempStore.create_path(extension='.txt')
+    with filelist_path.open('w', encoding='utf-8') as f:
+        for video in videos:
+            f.write(f"file '{video}'\n")
```
aaron-siegel: `'file {video!r}'` preferred

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+        cmd = ['ffmpeg', '-f', 'concat', '-safe', '0', '-i', str(filelist_path), '-c', 'copy', '-y', str(output_path)]
+        _logger.debug(f'concat_videos(): {" ".join(cmd)}')
+        try:
+            _ = subprocess.run(cmd, capture_output=True, text=True, check=True)
+            return str(output_path)
+        except subprocess.CalledProcessError:
+            # Expected for mixed formats - continue to fallback
+            pass
+
+        # we might some corrupted output
```
aaron-siegel: typo in comment

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+    first_res = resolutions[0]
+    for i, res in enumerate(resolutions[1:], start=1):
+        if res != first_res:
+            raise pxt.Error(
+                f'concat_videos(): requires that all videos have the same resolution, but '
+                f'video 0 ({videos[0]}) has resolution {first_res[0]}x{first_res[1]} and '
+                f'video {i} ({videos[i]}) has resolution {res[0]}x{res[1]}.'
+            )
+
+    # ffmpeg -f concat needs an input file list
```
aaron-siegel, line 464: seriously???? :/

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+
+        # general approach: re-encode with -f filter_complex
+        cmd = ['ffmpeg']
+        for video in videos:
+            cmd.extend(['-i', video])
+
+        all_have_audio = all(av_utils.has_audio_stream(str(video)) for video in videos)
+        video_inputs = ''.join([f'[{i}:v:0]' for i in range(len(videos))])
+        # concat video streams
+        filter_str = f'{video_inputs}concat=n={len(videos)}:v=1:a=0[outv]'
```
aaron-siegel, line 507: Some code comments, including an example fully baked command line (and description of what it's doing / what the parameters mean), would be helpful here. It's really hard to tell what's going on just by reading the code.

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+        if all_have_audio:
```
aaron-siegel, line 508: Hmmm, I'd imagine that if only some of the videos have audio, the user would want the audio preserved (and padded with silence for the videos without audio). I don't have a sense of how hard this may be to achieve.

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+            # also concat audio streams
+            audio_inputs = ''.join([f'[{i}:a:0]' for i in range(len(videos))])
+            filter_str += f';{audio_inputs}concat=n={len(videos)}:v=0:a=1[outa]'
+        cmd.extend(['-filter_complex', filter_str, '-map', '[outv]'])
+        if all_have_audio:
+            cmd.extend(['-map', '[outa]'])
+
+        video_encoder = Env.get().default_video_encoder
+        if video_encoder is not None:
```
aaron-siegel, line 517: What happens if it is `None`? We just don't get video? I'd imagine this would be a very confusing error for users - better to just throw an exception "no suitable video encoder found" or some such. (Or maybe we're relying on ffmpeg to do this?)

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+        for video in videos:
+            f.write(f"file '{video}'\n")
+
+    output_path = TempStore.create_path(extension='.mp4')
+
+    try:
+        # First attempt: fast copy without re-encoding (works for compatible formats)
+        cmd = ['ffmpeg', '-f', 'concat', '-safe', '0', '-i', str(filelist_path), '-c', 'copy', '-y', str(output_path)]
+        _logger.debug(f'concat_videos(): {" ".join(cmd)}')
+        try:
```
aaron-siegel, line 476: As an alternative to `try/except` here, one could call `run()` with `check=False` and check the result code after it completes.

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+
+    Args:
+        video: Input video to overlay text on.
+        text: The text string to overlay on the video.
+        font: Font family or path to font file. If None, uses the system default.
+        font_size: Size of the text in points.
+        color: Text color (e.g., `'white'`, `'red'`, `'#FF0000'`).
+        opacity: Text opacity from 0.0 (transparent) to 1.0 (opaque).
+        horizontal_align: Horizontal text alignment (`'left'`, `'center'`, `'right'`).
+        horizontal_margin: Horizontal margin in pixels from the alignment edge.
```
aaron-siegel, line 565: It's not clear to me what this means (I can hazard a guess, or read the code, but a more explicit description in the docstring might be worthwhile)

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+        return str(output_path)
+
+    except subprocess.CalledProcessError as e:
+        _handle_ffmpeg_error(e)
+    finally:
+        filelist_path.unlink()
+
+
+@pxt.udf(is_method=True)
+def overlay_text(
```
aaron-siegel, line 533: I wonder if, rather than inventing our own bespoke parameter list, it wouldn't be better to just take a string as input that gets passed to `drawtext`? By introducing our own bespoke API, we're abstracting away the ffmpeg tool, which somewhat goes against our ethos ("don't abstract away the tools")

Then the signature would be:
```
def overlay_text(
    video: pxt.Video,
    text: str,
    *,
    drawtext: str
)
```
The docstring could give lots of clear examples on how to map typical patterns onto ffmpeg language.

Even if we keep all the bespoke parameters, I'd argue we should provide a way for users to specify their own `drawtext` instead; otherwise we're masking the full power of the underlying API.

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+    Overlay text on a video with customizable positioning and styling.
+
+    __Requirements:__
+
+    - `ffmpeg` needs to be installed and in PATH
+
+    Args:
+        video: Input video to overlay text on.
+        text: The text string to overlay on the video.
+        font: Font family or path to font file. If None, uses the system default.
```
aaron-siegel, line 560: If these were separate parameters `font_family` and `font_path`, it'd be easier for the user to debug if something went wrong.

pixeltable/functions/video.py
```diff
@@ -221,66 +223,471 @@ def get_metadata(video: pxt.Video) -> dict:
...
+        font_size: Size of the text in points.
+        color: Text color (e.g., `'white'`, `'red'`, `'#FF0000'`).
+        opacity: Text opacity from 0.0 (transparent) to 1.0 (opaque).
```
aaron-siegel, line 563: Why not just bake these into the color strings ffmpeg style? `white@0.6` etc. It feels like going out of our way to abstract the API.

tests/test_video.py
```diff
@@ -282,3 +290,432 @@ def test_make_video(self, reset_db: None) -> None:
...
+                text,
+                color='black',
+                opacity=0.5,
+                horizontal_align='left',
+                horizontal_margin=10,
+                vertical_align='top',
+                box=False,
+            )
+        )
+        t.add_computed_column(
```
aaron-siegel, line 608: It's hard to test the output because the output is a video.

But we COULD break up the logic in `overlay_text()` via a `__overlay_params_to_drawtext()` subroutine, and test that thing, which would at least make sure that we're correctly mapping various combinations of "our" parameters onto the expected ffmpeg inputs.

aaron-siegel (changes requested):

tests/test_video.py
mkornacker: I really don't think that's a good use of my time.

tests/test_video.py
mkornacker: As I recall, it broke the no-dev test, which is why I added it.

pixeltable/functions/video.py

pixeltable/functions/video.py

pixeltable/functions/video.py

pixeltable/functions/video.py
mkornacker: The logic is as intended.

pixeltable/functions/video.py

pixeltable/functions/video.py
mkornacker: Then it's up to ffmpeg.

pixeltable/functions/video.py
mkornacker: I have a hard time believing that this would be very popular. Moviepy didn't take that path, and it's a popular package. I would suggest we use this and have users tell us where it breaks down in their workflow.

pixeltable/functions/video.py
mkornacker: How so?

pixeltable/functions/video.py
mkornacker: Again, this is emulating moviepy.

pixeltable/functions/video.py
mkornacker: I don't know whether that's a real use case. I'd rather not spend time on something where I have to speculate if it makes sense.

pixeltable/functions/video.py
mkornacker: That would in essence be a tutorial for the filter_complex syntax, which is beyond the scope of this code.

pixeltable/functions/video.py

pixeltable/functions/video.py

pixeltable/functions/video.py
mkornacker: Let's ask Denise what she things.

.github/workflows/pytest.yml
mkornacker: Done.

pixeltable/functions/video.py

pixeltable/functions/video.py
aaron-siegel: Did you see this comment?

pixeltable/functions/video.py
aaron-siegel: I don't think a "tutorial" is necessary, but I do think it's important to have an example to make the code readable.

pixeltable/functions/video.py
aaron-siegel: Got it, we just omit the explicit choice of encoding.

pixeltable/functions/video.py
aaron-siegel: Ok, I'm fine with following the precedent set by moviepy.

pixeltable/functions/video.py
aaron-siegel: ok

tests/test_video.py
aaron-siegel: Hmm, I think a better solution would be to include ffmpeg installation in the no-dev CI test. That would increase the surface of the no-dev test to include these methods (which don't rely on any dev dependencies).

tests/test_video.py
aaron-siegel: I'm not sure I agree with this. It's a large set of method parameters that interact in complicated ways and whose behavior currently isn't tested at all. That's just asking for something to break the next time someone refactors or adds a new parameter.

It also wouldn't be a lot of time, say 10 minutes to do the refactoring and 30 minutes to add a decent set of test cases. Passing up an opportunity to improve our test surface in order to save ~45 minutes doesn't seem like a good tradeoff to me?

PR #769, opened 2025-08-21
Basic support for concurrent pixeltable metadata creation/upgrade

amithadke:
There is a race condition when more than one pixeltable processes try to create schema tables at the same time.
Similar issue when upgrading metadata.

Adding retry with backoff for schema table creations
Adding exclusive lock on system info row for upgrade metadata 

Env._init_env() will only set the class instance only if env setup completes without error

### DIFF
--- a/pixeltable/env.py
+++ b/pixeltable/env.py
@@ -10,9 +10,11 @@ import inspect
 import logging
 import os
 import platform
+import random
 import shutil
 import sys
 import threading
+import time
 import types
 import typing
 import warnings
@@ -104,10 +106,14 @@ class Env:
             cls._instance._clean_up()
         cls._instance = None
         env = Env()
-        env._set_up(reinit_db=reinit_db)
-        env._upgrade_metadata()
-        cls._instance = env
-        cls.__initializing = False
+        try:
+            env._set_up(reinit_db=reinit_db)
+            env._upgrade_metadata()
+            cls._instance = env
+        except BaseException as e:  # Catch all exceptions
+            raise e
+        finally:
+            cls.__initializing = False
 
     def __init__(self) -> None:
         assert self._instance is None, 'Env is a singleton; use Env.get() to access the instance'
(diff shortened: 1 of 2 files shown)


pixeltable/env.py
```diff
@@ -508,8 +512,25 @@ def _init_metadata(self) -> None:
         assert self._sa_engine is not None
         from pixeltable import metadata
 
-        metadata.schema.base_metadata.create_all(self._sa_engine, checkfirst=True)
-        metadata.create_system_info(self._sa_engine)
+        max_retries = 3
```
amithadke: need better solution i guess im going to setup tables during db creation commit it toss that db connection
for all subsequent db connections they should see tables and skip them also systeminfo table wont update row 

we need to make sure upgrade metadata is not happening from 2 places at the same time may be we disable for now?

pixeltable/env.py
```diff
@@ -508,8 +512,25 @@ def _init_metadata(self) -> None:
...
+        base_delay = 0.1
+        for attempt in range(max_retries + 1):
```
aaron-siegel: I'd suggest using a tenacity decorator rather than hacking together our own ad hoc backoff logic.

pixeltable/env.py
```diff
@@ -508,8 +512,25 @@ def _init_metadata(self) -> None:
...
+            try:
+                metadata.schema.base_metadata.create_all(self._sa_engine, checkfirst=True)
+                metadata.create_system_info(self._sa_engine)
+                return
+            except Exception as e:
+                error_msg = str(e).lower()
+                # If it's a "table already exists" error, that's actually success
+                if any(phrase in error_msg for phrase in ['already exists', 'duplicate', 'relation already exists']):
```
aaron-siegel: This is fragile error handling - we shouldn't be relying on the stability of our error message text to ensure correct behavior.

Can we parameterize pxt.Error in a more structured way to flag the relevant error conditions?

amithadke: we will just do retries and not parse for special errors. After 3 attempts, throw pxt.Error so cloud service can fail the request without leaving an unusable Env,

pixeltable/env.py
amithadke: will do

pixeltable/env.py
```diff
@@ -500,6 +503,10 @@ def _init_db(self, config: Config) -> None:
         assert self._db_url is not None
         assert self._db_name is not None
 
+    @retry(
+        stop=stop_after_attempt(3),  # Stop after 3 attempts
+        wait=wait_exponential_jitter(initial=0.2, max=1.0, jitter=0.2),  # Exponential backoff with jitter
+    )
```
aaron-siegel, line 509: Maybe add a debug log message in the method body so the retries are apparent in the logs.

aaron-siegel (approved):

pixeltable/env.py
mkornacker, line 509: Comment why this is necessary here.

pixeltable/env.py
```diff
@@ -104,10 +105,12 @@ def _init_env(cls, reinit_db: bool = False) -> None:
...
-        env._set_up(reinit_db=reinit_db)
-        env._upgrade_metadata()
-        cls._instance = env
-        cls.__initializing = False
+        try:
+            env._set_up(reinit_db=reinit_db)
+            env._upgrade_metadata()
+            cls._instance = env
+        finally:
+            cls.__initializing = False
```
mkornacker, line 113: Comment why this is necessary here.

PR #778, opened 2025-08-26
Fix local_public_names() to properly exclude private functions

goodlux:
The function was incorrectly including private functions (prefixed with _) in the public API when they were Pixeltable Function objects. This caused private functions like _anthropic_response_to_pxt_tool_calls to appear in documentation.

Added underscore check for Function objects to match the existing behavior for Python functions.

### DIFF
--- a/pixeltable/utils/code.py
+++ b/pixeltable/utils/code.py
@@ -19,12 +19,18 @@ def local_public_names(mod_name: str, exclude: Optional[list[str]] = None) -> li
     mod = importlib.import_module(mod_name)
     names = []
     for obj in mod.__dict__.values():
+        # Check for private/protected names at the top
+        if hasattr(obj, '__name__') and obj.__name__.startswith('_'):
+            continue
+        
         if isinstance(obj, Function):
-            # Pixeltable function
+            # Pixeltable function - check name attribute for private functions
+            if obj.name.startswith('_'):
+                continue
             names.append(obj.name)
         elif isinstance(obj, types.FunctionType):
             # Python function
-            if obj.__module__ == mod.__name__ and not obj.__name__.startswith('_'):
+            if obj.__module__ == mod.__name__:
                 names.append(obj.__name__)
         elif isinstance(obj, types.ModuleType):
             # Module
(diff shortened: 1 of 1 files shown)


pixeltable/utils/code.py
```diff
@@ -19,12 +19,18 @@ def local_public_names(mod_name: str, exclude: Optional[list[str]] = None) -> li
     mod = importlib.import_module(mod_name)
     names = []
     for obj in mod.__dict__.values():
+        # Check for private/protected names at the top
+        if hasattr(obj, '__name__') and obj.__name__.startswith('_'):
```
aaron-siegel: This just adds to the confusion, I'd advocate for removing it and reverting the change to `FunctionType` (see below)

pixeltable/utils/code.py
```diff
@@ -19,12 +19,18 @@ def local_public_names(mod_name: str, exclude: Optional[list[str]] = None) -> li
...
         if isinstance(obj, Function):
-            # Pixeltable function
+            # Pixeltable function - check name attribute for private functions
+            if obj.name.startswith('_'):
+                continue
             names.append(obj.name)
         elif isinstance(obj, types.FunctionType):
             # Python function
-            if obj.__module__ == mod.__name__ and not obj.__name__.startswith('_'):
+            if obj.__module__ == mod.__name__:
```
aaron-siegel: revert this

pixeltable/utils/code.py
```diff
@@ -19,12 +19,18 @@ def local_public_names(mod_name: str, exclude: Optional[list[str]] = None) -> li
...
     names = []
     for obj in mod.__dict__.values():
+        # Check for private/protected names at the top
+        if hasattr(obj, '__name__') and obj.__name__.startswith('_'):
+            continue
+        
         if isinstance(obj, Function):
-            # Pixeltable function
+            # Pixeltable function - check name attribute for private functions
+            if obj.name.startswith('_'):
```
aaron-siegel: No need for a `continue` here, just a simple conditional suffices
```
if not obj.name.startswith('_'):
    names.append(obj.name)
```

aaron-siegel (changes requested):

aaron-siegel (approved):

PR #782, opened 2025-08-27
Allow images and arrays in JSON structures

aaron-siegel:
- Allows PIL images and numpy arrays to be stored in JSON structures
- Exports any non-standard data from JSON structures into a binary file, substituting a placeholder that references the file location of the data
- Also does the same for (ArrayType) numpy arrays exceeding a certain threshold size (currently set to `10_000`)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -98,10 +98,10 @@ jobs:
         run: |
           conda install -y python=${{ matrix.python-version }} pip=25.0
           conda info
-      - name: Install ffmpeg
-        # libiconv and ffmpeg are needed by whisper.
-        if: ${{ contains(matrix.package-configs, 'whisper') }}
-        run: ./scripts/retry.sh 3 30 conda install -c conda-forge libiconv ffmpeg
+      - name: Install tools and libraries
+        # cmake is needed for certain CI environments where it is not preinstalled.
+        # libiconv and ffmpeg are needed by WhisperX.
+        run: ./scripts/retry.sh 3 30 conda install -c conda-forge 'cmake>=3.22' libiconv 'ffmpeg==6.1.1'
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -126,11 +126,10 @@ jobs:
         run: |
           conda install -y python=${{ matrix.python-version }} pip=25.0
           conda info
-      - name: Install tools
+      - name: Install tools and libraries
         # cmake is needed for certain CI environments where it is not preinstalled.
         # libiconv and ffmpeg are needed by WhisperX.
-        if: ${{ matrix.uv-options == '' }}
-        run: ./scripts/retry.sh 3 30 conda install -c conda-forge 'cmake>=3.22' libiconv ffmpeg
+        run: ./scripts/retry.sh 3 30 conda install -c conda-forge 'cmake>=3.22' libiconv 'ffmpeg==6.1.1'
       - name: Install ollama
         if: ${{ matrix.uv-options == '' && matrix.os == 'ubuntu-22.04' }}
         run: |
--- a/Makefile
+++ b/Makefile
@@ -83,12 +83,12 @@ endif
 	@python -m pip install -qU pip
 	@python -m pip install -q uv==0.8.2
 	@echo 'Installing ffmpeg ...'
-	@conda install -q -y -c conda-forge libiconv ffmpeg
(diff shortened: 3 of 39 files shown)


pixeltable/exprs/data_row.py
```diff
@@ -56,14 +56,14 @@ class DataRow:
     img_slot_idxs: list[int]
```
jpeterson-pxt, line 56: Comments about the meaning and use of these lists of indices would be useful. There are now a lot of them...

pixeltable/exprs/data_row.py
```diff
@@ -252,42 +258,208 @@ def __setitem__(self, idx: int, val: Any) -> None:
...
+    @classmethod
+    def __has_media(cls, element: Any) -> bool:
+        if isinstance(element, list):
+            return any(cls.__has_media(v) for v in element)
+        if isinstance(element, dict):
+            return any(cls.__has_media(v) for v in element.values())
+        return isinstance(element, (np.ndarray, PIL.Image.Image))
+
+    @classmethod
+    def __rewrite_json(cls, element: Any, fp: io.BufferedWriter) -> Any:
```
jpeterson-pxt, line 363: Is the DataRow a good place for the methods splitting and reassembling the JSON structures? 
Should they be in a module of there own related to JSON processing?

pixeltable/exprs/row_builder.py
```diff
@@ -86,6 +86,7 @@ class RowBuilder:
     img_slot_idxs: list[int]  # Indices of image slots
     media_slot_idxs: list[int]  # Indices of non-image media slots
     array_slot_idxs: list[int]  # Indices of array slots
+    json_slot_idxs: list[int]  # Indices of JSON slots
```
jpeterson-pxt, line 89: Now that JSON columns can be treated like media columns, should they be part of the "stored_media_cols" list below?
This is what triggers all media processing during storage.

pixeltable/exprs/data_row.py
```diff
@@ -75,18 +75,22 @@ class DataRow:
     parent_row: Optional[DataRow]
     parent_slot_idx: Optional[int]
 
+    MAX_ARRAY_IN_DB = 10_000  # arrays larger than this get stored outside the DB
```
jpeterson-pxt, line 78: Should this have a unit associated with it? Perhaps bytes?
Should be moving anything (strings) larger than this number into the file?

pixeltable/exprs/data_row.py
```diff
@@ -252,42 +258,208 @@ def __setitem__(self, idx: int, val: Any) -> None:
...
+        if isinstance(element, PIL.Image.Image):
+            format = 'webp' if element.has_transparency_data else 'jpeg'
+            begin = fp.tell()
+            element.save(fp, format=format)
+            end = fp.tell()
+            return {'__pxttype__': ts.ColumnType.Type.IMAGE.name, '__pxtbegin__': begin, '__pxtend__': end}
+        return element
+
+    @classmethod
+    def __add_url_to_json(cls, element: Any, url: str) -> None:
```
jpeterson-pxt, line 383: I know that the spec says that all references should stand alone, but does it make sense to change it to have a single reserved field in the top of the json which gives us the name of the location so we don't have to revisit the entire json when relocating?

pixeltable/exprs/data_row.py
aaron-siegel: There's an existing comment about them - "control structures that are shared across all DataRows in a batch" - do you think that needs elaboration?

pixeltable/exprs/data_row.py
aaron-siegel: Right now it's just the number of elements in the array. We can always make the behavior more sophisticated but I went with something simple for now.

pixeltable/exprs/data_row.py
aaron-siegel: That wouldn't work in general because a top-level Json object could be a list, not necessarily a dict.

It would also complicate how we handle SQL queries of JSON substructures. For example, if I do this:
```
t.select(t.my_big_json_col.field1)
```
Then the SQL query will be set up so that only the contents of `field1` are retrieved. If the indirection were done via a single reserved toplevel field, we'd have to retrieve that separately; whereas the current method ensures that it will always be repeated in any substructure that needs it.

Another option would be to put it in cellmd, which complicates queries in the same way, but at least avoids the first issue (it doesn't matter if the json object is a list b/c cellmd is always a dict)

pixeltable/exprs/row_builder.py
aaron-siegel: Hmm, it doesn't look like `stored_media_cols` is being used for anything currently... :/ Does that change with your media destination PR?

PR #787, opened 2025-08-30
recompute_columns(): added where parameter

mkornacker:


pixeltable/dataframe.py
```diff
@@ -1210,17 +1210,42 @@ def update(self, value_spec: dict[str, Any], cascade: bool = True) -> UpdateStat
...
+        Returns:
+            UpdateStatus: the status of the operation.
+
+        Example:
+            For table `person` with column `age` and computed column `height`, recompute the value of `height` for all
+            rows where `age` is less than 18:
+
+            >>> df = person.where(t.age < 18).recompute_columns(person.height)
+        """
+        self._validate_mutable('recompute_columns', False)
```
aaron-siegel, line 1244: Check that `group_by`, `sample`, etc., are not specified, in order to avoid potentially unexpected behavior here.

aaron-siegel (changes requested):

PR #789, opened 2025-08-31
Disable TestWhisperx on Linux w/ GPU

mkornacker:


tests/functions/test_whisperx.py
```diff
@@ -6,13 +6,14 @@
...
 
-from ..utils import get_audio_files, skip_test_if_not_installed, validate_update_status
+from ..utils import get_audio_files, runs_linux_with_gpu, skip_test_if_not_installed, validate_update_status
 
 
 @pytest.mark.skipif(
     sysconfig.get_platform() == 'linux-aarch64', reason='libsndfile.so is missing on CI Linux ARM instances'
 )
 @pytest.mark.skipif(os.environ.get('PXTTEST_CI_OS') == 'ubuntu-x64-t4', reason='crashes on t4 CI instances')
+@pytest.mark.skipif(runs_linux_with_gpu(), reason='crashes on Linux with GPU')
```
aaron-siegel, line 14: Add a `not IN_CI` clause to the condition: this test does actually work in CI and we don't want to disable it there.

aaron-siegel (changes requested):

tests/functions/test_whisperx.py
mkornacker: Done.

PR #797, opened 2025-09-09
Proper implementation of package/restore for non-snapshot replicas

aaron-siegel:
- Package/restore for non-snapshot replicas works correctly
- Introduces a new concept of "fragment table versions", which cannot be queried nor retrieved via get_table()
- get_table('my_replica') retrieves a fixed-version handle to the most recent *complete* (i.e., non-fragment) table version
- a table with no complete table versions must be a `_system` (hidden) table and cannot be retrieved via get_table()
- drop_table() on a replica will revert it to a `_system` table if it has dependents
- if a replica is dropped (actually dropped, not reverted) and its immediate base table is a `_system` table with no other dependents, then the base table will be dropped as well (possibly recursively, if there are multiple `_system` tables in the ancestor hierarchy with no remaining dependents)


pixeltable/metadata/schema.py
```diff
@@ -244,6 +244,10 @@ class TableVersionMd:
     schema_version: int
     user: Optional[str] = None  # User that created this version
     update_status: Optional[UpdateStatus] = None  # UpdateStatus of the change that created this version
+    # A hidden table version cannot be queried or instantiated via get_table(). Typically, a hidden table version
+    # represents a version of a replica table that has incomplete data, and exists only to provide base table support
+    # for a dependent view.
+    is_hidden: bool = False
```
mkornacker: I'm finding 'hidden' a bit indirect - it describes how we treat it in the API, but not what it is.

is_view_projection? is_partial? something like that?

pixeltable/dataframe.py
```diff
@@ -1276,10 +1276,10 @@ def _validate_mutable(self, op_name: str, allow_select: bool) -> None:
 
         # TODO: Reconcile these with Table.__check_mutable()
         assert len(self._from_clause.tbls) == 1
-        if self._first_tbl.is_snapshot():
-            raise excs.Error(f'Cannot use `{op_name}` on a snapshot.')
         if self._first_tbl.is_replica():
             raise excs.Error(f'Cannot use `{op_name}` on a replica.')
+        if self._first_tbl.is_snapshot():
```
mkornacker, line 1282: If the reordering is meaningful/required, leave explanatory comment.

tests/share/test_packager.py
```diff
@@ -268,14 +269,14 @@ def test_iterator_view_round_trip(self, reset_db: None) -> None:
...
         snapshot_replica = pxt.get_table('new_replica')
         assert snapshot_replica._snapshot_only
         assert snapshot_replica.count() == snapshot_row_count
-        v_replica = snapshot_replica.get_base_table()
-        assert v_replica.count() == snapshot_row_count
-        t_replica = v_replica.get_base_table()
-        assert t_replica.count() == 2
+        # We can't query the base table directly via snapshot_replica.get_base_table(), because it doesn't exist as a
+        # visible catalog object (it's hidden in _system). But we can manually construct the DataFrame and check that.
+        t_replica_df = pxt.DataFrame(FromClause(tbls=[snapshot_replica._tbl_version_path.base]))
```
mkornacker, line 278: We should have tests limit themselves to the public API. You should be able to express that with something like
snapshot_replica.select(s_r.video).distinct()

tests/share/test_packager.py
```diff
@@ -428,6 +429,28 @@ def test_multi_view_round_trip_6(self, reset_db: None) -> None:
...
+        t.insert({'int_col': i} for i in range(512))
+
+        t_bundle = self.__package_table(t)
+
+        v = pxt.create_view('view', t.where(t.int_col % 3 == 0))
+        t.insert({'int_col': i} for i in range(512, 1024))
+
+        v_bundle = self.__package_table(v)
+
+        # v_bundle has more rows, but fewer columns.
```
mkornacker: What am I missing? I would expect v_bundle to have 1024/3 rows and 1 column.

tests/share/test_packager.py
```diff
@@ -456,14 +479,11 @@ def test_multi_view_non_snapshot_round_trip(self, reset_db: None) -> None:
...
-        # Restore the odd-numbered bundles (directly packaged table versions). This needs to be done in order
-        # currently, because we don't have a way to get a handle to an older version of a table.
-        # TODO: Randomize the order once we have such a feature.
-        for n in (1, 3, 7, 9):
-            self.__restore_and_check_table(bundles[n], 'replica')
-
-        for n in (4, 0, 8, 10, 6):
-            self.__restore_and_check_table(bundles[n], f'replica_{n}')
+        for n in (4, 1, 0, 3, 8, 10, 7, 9, 6):
+            # We need to use a consistent name for the non-snapshot bundles, and distinct names for the snapshots.
```
mkornacker: I'm not following. Expand on the comment?

tests/test_table.py
```diff
@@ -2824,8 +2825,8 @@ def test_table_api_on_dropped_table(self, reset_db: None) -> None:
             _ = t.count()
         with pytest.raises(pxt.Error, match=unknown_tbl_msg):
             _ = t.head()
-        with pytest.raises(pxt.Error, match=unknown_tbl_msg):
-            _ = t.limit(1)
+        # with pytest.raises(pxt.Error, match=unknown_tbl_msg):
```
mkornacker: Why commented out?

pixeltable/catalog/catalog.py
```diff
@@ -1084,14 +1079,29 @@ def create_replica(self, path: Path, md: list[schema.FullTableMd]) -> None:
...
 
+        # Finally, it's possible that the table already exists in the catalog, but as an anonymous system table that
+        # was hidden the last time we checked (and that just became visible when the replica was imported). In this
+        # case, we need to make the existing table visible by moving it to the specified path.
+        existing = self.get_table_by_id(tbl_id)
+        if existing is not None:
+            existing_path = Path.parse(existing._path(), allow_system_path=True)
+            if existing_path != path:
+                assert existing_path.is_system_path
+                self._move(existing_path, path)
```
mkornacker, line 1101: This used to be done earlier. Is it necessary to do here now? If so, explain.

mkornacker (commented):
First round of comments.

pixeltable/catalog/catalog.py
```diff
@@ -1206,41 +1225,72 @@ def drop_table(self, path: Path, if_not_exists: IfNotExistsParam, force: bool) -
...
-            if not force:
-                is_snapshot = tbl._tbl_version_path.is_snapshot()
-                obj_type_str = 'Snapshot' if is_snapshot else tbl._display_name().capitalize()
+            if force:
+                # recursively drop views first
+                for view_id in view_ids:
+                    view = self.get_table_by_id(view_id)
+                    self._drop_tbl(view, force=force, is_replace=is_replace)
+
+            elif is_replica:
```
mkornacker, line 1270: I think if we have a local replica and create a local view on it, dropping that replica should be an error. Otherwise you can make that same argument for non-replica local tables (ie, they should become system tables).

pixeltable/catalog/catalog.py
```diff
@@ -1251,26 +1301,45 @@ def _drop_tbl(self, tbl: Table, force: bool, is_replace: bool) -> None:
...
+            del self._tbls[tbl_id, version]
+
+        _logger.info(f'Dropped table {tbl_id if tbl is None else repr(tbl._path())}.')
+
+        if (
+            is_replica  # if this is a replica,
+            and do_drop  # and it was actually dropped (not just renamed),
+            and tvp.base is not None  # and it has a base table,
+        ):
+            base_tbl = self.get_table_by_id(tvp.base.tbl_id)
```
mkornacker, line 1337: How could this be None?

pixeltable/catalog/catalog.py
```diff
@@ -1470,7 +1539,7 @@ def _load_tbl(self, tbl_id: UUID) -> None:
         if has_pending_ops:
             raise PendingTableOpsError(tbl_id)
 
-        q = (
+        q: sql.Executable = (
```
mkornacker, line 1545: Why is that necessary/useful?

pixeltable/catalog/catalog.py
```diff
@@ -1724,10 +1816,26 @@ def store_tbl_md(
...
-            )
-            session.add(tbl_version_record)
+            if (
+                session.query(schema.TableVersion)
+                .filter(schema.TableVersion.tbl_id == tbl_id, schema.TableVersion.version == version_md.version)
+                .count()
+                > 0
+            ):
+                # This table version already exists; update it. (This typically happens when setting the is_hidden
+                # flag of an existing TableVersion from True to False.)
```
mkornacker: Validate that only the flag gets updated (and in particular that created_at and schema_version are identical).

mkornacker (changes requested):

pixeltable/metadata/schema.py
aaron-siegel: `virtual`?

pixeltable/catalog/catalog.py
aaron-siegel: Added a comment.

pixeltable/catalog/catalog.py
aaron-siegel: We'd already discussed this and agreed that this is the correct behavior. (See PXT-707 which captures the outcome of that discussion.)

The situation with replicas is _not_ analogous to local tables, and the reason is that with replicas (unlike with local tables) it's possible to instantiate a view _before_ instantiating its base table. The key scenario is this:

- I clone a cloud-hosted table V. Only V is visible to me locally; not any of its base tables. Let's call this "database state A".
- I now clone T, which is a base table of V. Let's call this "database state B".
- I now say, "oops, I don't actually need T," so I drop it. But I find that I can't, because V exists.

It seems weird that I can't drop the table I created moments ago, when all I'm trying to do is get back to the valid database state A. Whether I can drop T now shouldn't depend on whether or not I replicated V at some point in the past. If I can't drop T now, then my only recourse is to drop both T and V and then re-create V by re-downloading it from scratch.

The difference vs. local tables is that database state A is a valid state for replicas, but not for local tables. And it all makes sense if we remember the principle that "all replicas are effectively snapshots".

We could in theory have a separate `drop_replica()` that's used for replicas, just as `create_replica()` is distinct from `create_table()`, to capture the difference in behavior in the API.

pixeltable/catalog/catalog.py
aaron-siegel: If the table has no visible table versions, then it is not instantiable as a catalog object, and this will return None.

pixeltable/catalog/catalog.py
aaron-siegel: Because `q` is reassigned with different types, so the type checker will (correctly) complain unless it is explicitly declared with a common supertype.

pixeltable/dataframe.py
aaron-siegel: Done.

tests/share/test_packager.py
aaron-siegel: I don't agree with this principle. I'd argue the right principle is: Tests should limit themselves to the public API whenever feasible.

But sometimes it's important to validate that the internal representation of something is correct/efficient (for example), in scenarios where the less efficient representation would produce identical output.

That's the case here. I can do what you're suggesting, but it doesn't really test the same thing, because we'd get the same result if the replica were restored as a single flat table (with each video path repeated N times) rather than a proper table/view hierarchy. And the point of this test is to validate that the table/view structure was faithfully recreated, even though it's now (properly) invisible to the user.

tests/share/test_packager.py
aaron-siegel: You're right, this comment doesn't make sense. The correct comment is "v_bundle is missing some of the rows that were present in t_bundle, but has some new ones as well." Fixed.

tests/test_table.py
aaron-siegel: I'm not sure. It's pretty strange. Right now, on `main` (independent of this PR):

If I call `t.limit(1)`, I get an error.

But if I call `t._tbl_version_path.is_replica()` and then do `t.limit(1)`, then no error! Merely checking whether `t` is a replica alters the behavior.

With this PR, an `is_replica()` call occurs in the prior history of `t` in this test, so that's why the error goes away. I was hoping you might be able to shed some light on why this is happening.

tests/share/test_packager.py

pixeltable/catalog/catalog.py

pixeltable/metadata/schema.py
mkornacker: Isn't the defining characteristic that it contains only what is needed for the views above it? I'm not sure I understand virtual.

(I think it's worth finding a term here that reflects the semantics. Things are fairly complicated already, and intuitive names make the code more accessible.)

tests/test_table.py
mkornacker: Hm, nothing comes to mind off the top of my head. But it feels like something potentially serious. Let's look at it together tomorrow, I think this should get fixed.

pixeltable/metadata/schema.py
aaron-siegel: I find `is_view_projection` to be more confusing than explanatory. I could be ok with `is_partial`. Or `is_fragment`?

tests/test_table.py
aaron-siegel: I think I found it, there was a missing `clear_cached_md()` in `_tbl_drop()` (both on main and in this branch). Adding the `is_replica()` call triggered `refresh_cached_md()`, which caused the downstream logic to assume the md from before the drop was still valid.

mkornacker (approved):

pixeltable/metadata/schema.py
aaron-siegel: Closing the loop on this conversation:
- Metadata field name changed to `is_fragment`
- A table is "hidden" or "visible"
- A table _version_ is a "fragment" or "complete"
- A visible table must have at least one complete table version

PR #799, opened 2025-09-09
Support create and insert from polars dataframes. Add doc for DataFrameResultSet.

jpeterson-pxt:
Core Changes

Polars Integration for create_table() and insert() methods:
- The polars DataFrame "pl.DataFrame" is a valid source for create_table and insert from in-memory data
- Support all Polars data types: integers, floats, dates, arrays, structs, categorical
- Handle complex nested data structures (JSON, variable-length arrays)
- Support schema overrides and primary key constraints

Add DataFrameResultSet.to_polars() method with similar functionality to existing to_pandas() method

API Documentation Fix:
- Create missing docs/api/pixeltable/data-frame-result-set.md documentation
- Add DataFrameResultSet to mkdocs navigation (was completely undocumented)
- Document to_pandas(), to_polars(), and to_pydantic() methods
- Add proper docstrings and cross-references

### DIFF
--- a/docs/mintlify/docs/datastore/bringing-data.mdx
+++ b/docs/mintlify/docs/datastore/bringing-data.mdx
@@ -499,10 +499,11 @@ Polars offers high-performance DataFrame operations with Rust-powered columnar p
 | `pl.Float32/64` | `pxt.Float` | Floating point numbers |
 | `pl.Boolean` | `pxt.Bool` | Boolean values |
 | `pl.Utf8/String/Categorical` | `pxt.String` | Text data |
-| `pl.Date` | `pxt.Date` | Date only (no time) |
+| `pl.Date`     | `pxt.Date` | Date only (no time) |
 | `pl.Datetime` | `pxt.Timestamp` | Date and time with timezone handling |
-| `pl.List` | `pxt.Array` or `pxt.Json` | Arrays for numeric, JSON for complex |
-| `pl.Struct` | `pxt.Json` | Nested structures |
+| `pl.Array`    | `pxt.Array` or `pxt.Json` | Arrays for numeric, JSON for complex |
+| `pl.List`     | `pxt.Array` or `pxt.Json` | Arrays for numeric, JSON for complex |
+| `pl.Struct`   | `pxt.Json` | Nested structures |
 
 ### Complex Data Types
 


docs/api/pixeltable/data-frame-result-set.md
```diff
@@ -0,0 +1,29 @@
+# DataFrameResultSet
+
+`DataFrameResultSet` represents the result of executing a DataFrame query. It's returned by methods like 
+[`collect`][pixeltable.DataFrame.collect], [`show`][pixeltable.DataFrame.show], 
+[`head`][pixeltable.DataFrame.head], and [`tail`][pixeltable.DataFrame.tail].
+
+The result set provides methods to access the query results and convert them to different formats for further processing.
+
+## Overview
```
aaron-siegel, line 23: This really needs a couple of examples on how to get data out of a `DataFrameResultSet`.

docs/api/pixeltable/data-frame.md
```diff
@@ -19,12 +23,12 @@ Another difference to pandas is that query execution needs to be initiated expli
...
 
-| Query Execution                           |                                     |
-|-------------------------------------------|-------------------------------------|
-| [`collect`][pixeltable.DataFrame.collect] | Return all output rows              |
-| [`show`][pixeltable.DataFrame.show]       | Return a number of output rows      |
-| [`head`][pixeltable.DataFrame.head]       | Return the oldest rows              |
-| [`tail`][pixeltable.DataFrame.tail]       | Return the most recently added rows |
+| Query Execution                           |                                                  |
+|-------------------------------------------|--------------------------------------------------|
+| [`collect`][pixeltable.DataFrame.collect] | Return all output rows as DataFrameResultSet     |
```
aaron-siegel: "as a DataFrameResultSet"

docs/mintlify/docs/datastore/bringing-data.mdx
```diff
@@ -470,6 +486,78 @@ table.insert(df)
   </Accordion>
 </AccordionGroup>
 
+## Polars DataFrame Integration
+
+Polars offers high-performance DataFrame operations with Rust-powered columnar processing. Pixeltable provides full support for Polars DataFrames with the same API patterns as Pandas.
+
+### Performance Benefits
```
aaron-siegel: Do we care about this in Pixeltable docs? I'd think a competitive analysis of Pandas vs. Polars would be out of scope for our docs.

docs/mintlify/docs/datastore/bringing-data.mdx
```diff
@@ -470,6 +486,78 @@ table.insert(df)
...
+| Polars Type | Pixeltable Type | Notes |
+|-------------|----------------|--------|
+| `pl.Int8/16/32/64` | `pxt.Int` | All integer types unified |
+| `pl.UInt8/16/32/64` | `pxt.Int` | Unsigned integers supported |
+| `pl.Float32/64` | `pxt.Float` | Floating point numbers |
+| `pl.Boolean` | `pxt.Bool` | Boolean values |
+| `pl.Utf8/String` | `pxt.String` | Text data |
+| `pl.Date` | `pxt.Date` | Date only (no time) |
+| `pl.Datetime` | `pxt.Timestamp` | Date and time with timezone handling |
+| `pl.List` | `pxt.Array` or `pxt.Json` | Arrays for numeric, JSON for complex |
```
aaron-siegel: I don't think this is right. We shouldn't munge the conversions - pl.List should go to pxt.Json. There is a pl.Array type that seems to be missing here which I assume is the natural analog of pxt.Array.

docs/mintlify/docs/datastore/bringing-data.mdx
```diff
@@ -470,6 +486,78 @@ table.insert(df)
...
+| `pl.Categorical` | `pxt.String` | Categorical → String conversion |
+
+### Complex Data Types
+
+```python
+# Polars handles complex nested data elegantly
+complex_df = pl.DataFrame({
+    'id': [1, 2],
+    'metadata': [{'type': 'A', 'score': 0.9}, {'type': 'B', 'score': 0.7}],
+    'embeddings': [[0.1, 0.2, 0.3], [0.4, 0.5, 0.6]],
```
aaron-siegel, line 515: Use a pl.Array here?

mkdocs.yml
```diff
@@ -28,12 +28,14 @@ nav:
...
       - DocumentSplitter: pixeltable/iterators/document-splitter.md
       - FrameIterator: pixeltable/iterators/frame-iterator.md
+      - StringSplitter: pixeltable/iterators/string-splitter.md
       - TileIterator: pixeltable/iterators/tile-iterator.md
       - VideoSplitter: pixeltable/iterators/video-splitter.md
     - ColumnMetadata: pixeltable/column-metadata.md
     - DataFrame: pixeltable/data-frame.md
     - DirContents: pixeltable/dir-contents.md
     - IndexMetadata: pixeltable/index-metadata.md
+    - DataFrameResultSet: pixeltable/data-frame-result-set.md
```
aaron-siegel: Keep these in alphabetical order

pixeltable/dataframe.py
```diff
@@ -22,6 +22,8 @@
 )
 
 import pandas as pd
+
+# import polars as pl
```
aaron-siegel: ?

pixeltable/dataframe.py
```diff
@@ -81,8 +93,44 @@ def _reverse(self) -> None:
...
+
+        # Create the data dictionary, handling None values properly for Polars
+        data_dict = {}
+        for col_name, col_data in zip(self._col_names, zip(*self._rows)):
+            # Convert numpy arrays to lists for Polars compatibility
+            processed_data: list[Any] = []
+            for val in col_data:
+                if val is None:
+                    processed_data.append(None)
+                elif hasattr(val, 'tolist'):
```
aaron-siegel: I don't understand what's going on here. Please add code comments to explain why this is necessary and what types of conversions are indicated here.

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+    # String types
+    elif isinstance(pl_dtype, (pl_types.Utf8, pl_types.String)):
+        return ts.StringType(nullable=nullable)
+    # Date and time types
+    elif isinstance(pl_dtype, pl_types.Date):
+        return ts.DateType(nullable=nullable)
+    elif isinstance(pl_dtype, pl_types.Datetime):
+        return ts.TimestampType(nullable=nullable)
+    elif isinstance(pl_dtype, (pl_types.Time, pl_types.Duration)):
+        return ts.StringType(nullable=nullable)  # Pixeltable doesn't have Time type, use String
```
aaron-siegel: For Pandas we simply fail if we encounter these types. What's the justification for doing things differently here? (FWIW I think failing is the correct behavior; Duration -> String is a pretty grotesque munge to be doing b/c Pixeltable lacks the corresponding feature. Better to add the feature if we think it's important.)

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+
+    return None
+
+
+def _pl_check_primary_key_values(df: 'pl.DataFrame', primary_key: list[str]) -> None:
+    """Check that primary key columns don't contain null values."""
+    for col_name in primary_key:
+        null_count = df.get_column(col_name).null_count()
+        if null_count > 0:
+            raise excs.Error(f'Primary key column `{col_name}` cannot contain null values.')
```
aaron-siegel:
```python
{col_name!r}
```
in place of
```python
`{col_name}`
```

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def pl_infer_schema(
```
aaron-siegel: Does this just cargo-cult code from Pandas? If so, is there a way to generalize (and is it worth it to generalize) so we're not cargo-culting?

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+                pl_types.Int16,
+                pl_types.Int32,
+                pl_types.Int64,
+                pl_types.UInt8,
+                pl_types.UInt16,
+                pl_types.UInt32,
+                pl_types.UInt64,
+            ),
+        ):
+            # Try to infer array shape from actual data
```
aaron-siegel: I'd argue we shouldn't be turning lists into arrays in the first place.

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+                        else:
+                            first_non_null = val
+                        break
+
+                if first_non_null is not None and hasattr(first_non_null, '__len__') and len(first_non_null) > 0:
+                    return ts.ArrayType(shape=(None, len(first_non_null)), dtype=ts.FloatType(), nullable=nullable)
+
+            return ts.ArrayType(shape=(None, None), dtype=ts.FloatType(), nullable=nullable)
+        elif isinstance(inner_type, (pl_types.Utf8, pl_types.String)):
+            # Handle List(String) - could be string arrays or mixed type converted to strings
```
aaron-siegel: I don't understand this; why would we ever want to interpret a List as an array of strings?

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+            # Use JSON for flexibility since string arrays can have variable lengths
+            return ts.JsonType(nullable=nullable)
+        else:
```
aaron-siegel: What's the point in having the separate elif clause above that returns the exact same thing?

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+    if isinstance(pl_dtype, pl_types.Categorical):
+        return ts.StringType(nullable=nullable)
+
+    # Handle Object type by inspecting actual data
+    if isinstance(pl_dtype, pl_types.Object):
+        # Make a copy of the Series without null values for type inference
+        non_null_data = data_col.drop_nulls()
+
+        if len(non_null_data) == 0:
+            # No non-null values; default to StringType
```
aaron-siegel: I'd argue automatic type munging is not a good idea. Did you compare what we're doing here vs. what we do in similar scenarios with Pandas and Parquet (for example) to ensure we're processing things in a consistent way?

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+        inferred_type = ts.ColumnType.infer_common_literal_type(sample_values)
+        if inferred_type is None:
+            # Fallback on StringType if everything else fails
+            return ts.StringType(nullable=nullable)
+        else:
+            return inferred_type.copy(nullable=nullable)
+
+    # Handle Null type
+    if isinstance(pl_dtype, pl_types.Null):
+        return ts.StringType(nullable=True)  # Default to nullable string for null columns
```
aaron-siegel: We've been using ts.InvalidType(nullable=True) for this. StringType definitely isn't right.

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+            nval = float(val)
+        elif pxt_type.is_int_type():
+            nval = int(val)
+        elif pxt_type.is_bool_type():
+            nval = bool(val)
+        elif pxt_type.is_string_type():
+            nval = str(val)
+        elif pxt_type.is_date_type():
+            if hasattr(val, 'date'):
+                nval = val.date()  # Extract date from datetime
```
aaron-siegel: Not a good idea in general to munge timestamps into dates, and even more so without specifying an explicit time zone for naive timestamps. Best to just error out if we're given a datetime but expect a date.

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+            else:
+                nval = val
+        elif pxt_type.is_timestamp_type():
+            if hasattr(val, 'astimezone'):
```
aaron-siegel: I don't understand what's going on here. All datetimes share the same class, so this doesn't actually tell us anything about whether it's a naive or aware datetime.

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+                # Handle timezone-aware timestamps
+                nval = val.astimezone(Env.get().default_time_zone)
+            else:
+                # Handle naive timestamps - localize to default timezone
+                import datetime as dt
+
+                if isinstance(val, dt.datetime):
+                    nval = val.replace(tzinfo=Env.get().default_time_zone)
+                else:
```
aaron-siegel: What else can it be?

pixeltable/io/polars.py
```diff
@@ -0,0 +1,265 @@
...
+                    nval = val
+        elif pxt_type.is_json_type():
+            # For JSON types, keep the value as-is (lists, dicts, etc.)
+            nval = val
+        elif pxt_type.is_array_type():
+            # For array types, convert lists to numpy arrays as Pixeltable expects
```
aaron-siegel: I'd argue we shouldn't be converting lists to arrays at all

pixeltable/io/table_data_conduit.py
```diff
@@ -11,12 +11,16 @@
 from typing import TYPE_CHECKING, Any, Iterable, Iterator, Literal, Optional, cast
 
 import pandas as pd
+
+if TYPE_CHECKING:
```
aaron-siegel: TYPE_CHECKING block should go after regular imports

pixeltable/io/table_data_conduit.py
```diff
@@ -280,6 +284,72 @@ def valid_row_batch(self) -> Iterator[RowData]:
...
+        if self.source_column_map is None:
+            if self.src_schema_overrides is None:
+                self.src_schema_overrides = {}
+            self.src_schema = pl_infer_schema(self.pl_df, self.src_schema_overrides, self.src_pk)
+            inferred_schema, inferred_pk, self.source_column_map = normalize_schema_names(
+                self.src_schema, self.src_pk, self.src_schema_overrides, False
+            )
+            return inferred_schema, inferred_pk
+        else:
+            raise NotImplementedError()
```
aaron-siegel, line 324: ? Why this exception? Very unusual to conditionally return a NotImplementedError() (which is intended to indicate the absence of a method body)

tests/io/test_polars.py
```diff
@@ -0,0 +1,465 @@
...
+            'str_col': ['a', 'b'],
+            'dt_col': [datetime.datetime(2024, 1, 1, 1, 1, 1), datetime.datetime(2024, 1, 2, 1, 1, 1)],
+            'aware_dt_col': [
+                datetime.datetime(2024, 1, 1, 1, 1, 1, tzinfo=(ZoneInfo('Europe/Moscow'))),
+                datetime.datetime(2024, 1, 1, 1, 1, 1, tzinfo=datetime.timezone.utc),
+            ],
+            'date_col': [datetime.date(2024, 1, 1), datetime.date(2024, 1, 2)],
+            'json_col_1': [[1, 2], [3, 4]],
+            'json_col_2': [{'a': 1}, {'b': 2}],
+            'array_col_1': [[1, 2], [3, 4]],  # polars handles arrays as lists
```
aaron-siegel: What about pl.Array?

tests/io/test_polars.py
```diff
@@ -0,0 +1,465 @@
...
+            'int_col': ts.IntType(nullable=True),
+            'float_col': ts.FloatType(nullable=True),
+            'bool_col': ts.BoolType(nullable=True),
+            'str_col': ts.StringType(nullable=True),
+            'dt_col': ts.TimestampType(nullable=True),
+            'aware_dt_col': ts.TimestampType(nullable=True),
+            'date_col': ts.DateType(nullable=True),
+            'json_col_1': ts.ArrayType(
+                shape=(None, 2), dtype=ts.IntType(), nullable=True
+            ),  # polars infers consistent-length int lists as Arrays
```
aaron-siegel: Polars isn't doing this, we are, and I don't think it's the right thing to do

jpeterson-pxt: You can ask polars to do this, but it is not implicit during data insertion.
I'm reworking this entire section, and the tests.

tests/io/test_polars.py
jpeterson-pxt: Done

pixeltable/io/table_data_conduit.py
jpeterson-pxt: DONE

pixeltable/dataframe.py
jpeterson-pxt: Done

pixeltable/io/polars.py
jpeterson-pxt: Done

pixeltable/io/polars.py
jpeterson-pxt: Done

pixeltable/io/polars.py
jpeterson-pxt: Done

pixeltable/io/polars.py
jpeterson-pxt: Done

pixeltable/io/table_data_conduit.py
jpeterson-pxt: This pattern is identical to all the other TableDataConduit implementations. 
We discussed this when they were merged. The essence of that discussion: There are situations where we wish there was code here, and we agreed it was better to report NotImplemented than assert. We can change them all if you feel differently now.

pixeltable/io/polars.py
jpeterson-pxt: I'm not sure what you're referring to here, exactly. Each of the TableDataConduit implementations relies on a method like this (though they are not named with the same pattern, which I have learned to view as a mistake :-)).
It is a part of the work that is done in each TDC implementation that is deferred to the code for the source (in this case polars).
While this one looks a lot like the pandas one, that is because the needs in the two cases are similar. 
I could go back and refactor the TDC to change the way work is allocated in the hope of getting something more concise, but that seems wasteful at this point.

pixeltable/io/polars.py
jpeterson-pxt: Yes, this is bogus.
Fixed.

pixeltable/io/polars.py
jpeterson-pxt: Done

pixeltable/io/polars.py
jpeterson-pxt: When polars 'exports' data row by row, it converts numpy arrays to lists. 
Lame, but that's what I could decipher from the docs.
So the code may be bogus, but the concept is needed.

pixeltable/io/polars.py
jpeterson-pxt: Done

pixeltable/io/polars.py
jpeterson-pxt: This is the action on columns of NULLs. 
Removed this logic.

pixeltable/io/polars.py
jpeterson-pxt: Done

pixeltable/io/polars.py
jpeterson-pxt: I have removed this logic.

mkdocs.yml
jpeterson-pxt: Done

docs/mintlify/docs/datastore/bringing-data.mdx
jpeterson-pxt: Removed.

docs/api/pixeltable/data-frame.md
jpeterson-pxt: Done

docs/api/pixeltable/data-frame-result-set.md
jpeterson-pxt: Done

docs/mintlify/docs/datastore/bringing-data.mdx
jpeterson-pxt: We can discuss. 
Both pl.Array and pl.List refer to collections of a single sub-type.
pl.Array refers to a fixed length 1d array
pl.List refers to a variable length 1d array
They can be nested as needed:
    pl.Array(pl.List(pl.List(pl.Array(pl.Int64, 7))), 92)
We would think of this as an array type with shape(92, None, None, 6), dtype = pxt.Int

pixeltable/dataframe.py
jpeterson-pxt: Substantially revised.

docs/mintlify/docs/datastore/bringing-data.mdx
jpeterson-pxt: A full schema must be specified here to create the pl.Array type. 
We will consider the pl.List type which results from this creation to be a pxt.Array type.

PR #800, opened 2025-09-11
Table.get_versions()

aaron-siegel:


pixeltable/catalog/table.py
```diff
@@ -1744,32 +1736,57 @@ def history(self, n: Optional[int] = None) -> pixeltable.dataframe.DataFrameResu
...
             version = vers_md.version_md.version
-            schema_change = md_dict.get(version, '')
+            schema_change = md_dict.get(version, None)
             update_status = vers_md.version_md.update_status
             if update_status is None:
                 update_status = UpdateStatus()
-            change_type = 'schema' if schema_change != '' else ''
-            if change_type == '':
-                change_type = 'data'
+            change_type: Literal['schema', 'data'] = 'schema' if schema_change != '' else 'data'
```
mkornacker: Shouldn't that be schema_change is not None?

If so, shouldn't a test have picked that up? In that case, either expand test or leave todo.

pixeltable/catalog/table.py
```diff
@@ -1845,3 +1862,28 @@ class TableMetadata(TypedDict):
     """The media validation policy for this table."""
     base: Optional[str]
     """If this table is a view or snapshot, the full path of its base table; otherwise `None`."""
+
+
+class VersionMetadata(TypedDict):
```
mkornacker: We're accumulating more typeddicts in this already long file. Move these into globals?

aaron-siegel: I moved them into a new table_metadata.py.

pixeltable/catalog/table.py
aaron-siegel: Good catch. Fixed and added missing tests.

PR #803, opened 2025-09-15
PXT-638, PXT-675, PXT-682 Handle Keyboard exception

amithadke:
- PXT-675 When collecting async task's results in call back, if task is interrupted ignore KeyboardInterrupt as task never completed
- Remove rollback logic in table_version and handle keyboard interrupt in Catalog transaction error handling path (Invalidate all modified table versions and table's cached md)
- PXT-638 Enabling test_computed_col_with_interrupts
- PXT-682 Don't raise error during temp table cleanup
    Temp tables are cleaned up after column loading. We use run_cleanup() to ensure
    the DROP TABLE completes even if interrupted by KeyboardInterrupt.
    However, if the transaction is already failed/aborted, the DROP TABLE will also fail.
    Using raise_error=False suppresses these expected cleanup failures without masking  the original error.


pixeltable/catalog/table_version.py
```diff
@@ -595,18 +594,7 @@ def _create_index(
         idx_info = self.IndexInfo(id=idx_id, name=idx_name, idx=idx, col=col, val_col=val_col, undo_col=undo_col)
         self._tbl_md.index_md[idx_id] = idx_md
         self.idxs_by_name[idx_name] = idx_info
-        try:
-            idx.create_index(self._store_idx_name(idx_id), val_col)
-        finally:
-
-            def cleanup_index() -> None:
```
aaron-siegel, line 611: Are we sure it's the right thing to do to take this out? @mkornacker please take a look

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -390,6 +390,10 @@ def _done_cb(self, t: asyncio.Task) -> None:
         # end the main loop if we had an unhandled exception
         try:
             t.result()
+        except KeyboardInterrupt:
```
mkornacker, line 393: Doesn't this end up swallowing the exception?

mkornacker (commented):
Nice!

pixeltable/exec/expr_eval/expr_eval_node.py
amithadke: We ignore KeyboardInterrupt here since task from previous operation that was interrupted may come and set error for next operation. When user interrupts main loop, it does catch and shows error the terminal, but orphan task  for this operation can raise old KeyboardInterrupt during callback. I think this is because we have long running exec eval node which will reused even after being interrupted.

mkornacker: But if one of the scheduled tasks is running while the user hits ctrl-c, what happens? Wouldn't it show up here? _done_cb propagates a regular exception by setting exc_event, shouldn't it do the same for a ctrl-c?

amithadke: added comment there explaining why we dont record keyboard interrupt

aaron-siegel (approved):

pixeltable/exec/expr_eval/expr_eval_node.py
amithadke: KeyboardInterrupt is initiated by the OS, which sends a SIGINT signal to the Python process's main thread. In response, the main thread's event loop begins a shutdown sequence, canceling all awaiting or pending tasks.

A running task that hasn't yielded control back to the event loop won't see this cancellation immediately. Instead, it will be handled by python's interpreter's signal handler, which then will raise KeyboardInterrupt in the task's context.

If no one in the task's callstack handle this KeyboardInterrupt then its result is marked with the KeyboardInterrupt.

If task is not running python code but an extension/library then its up to that library to check for pending interrupts. If this extension/library doesn't check for pending interrupts, then task will continue to run until its finished and we will see unresponsive behavior. 

When the task's completion callback is finally triggered, likely in the future during a subsequent operation, it will raise this stored KeyboardInterrupt. This is why you need to absorb it: it's a "phantom" exception from an already-handled user cancellation, and it shouldn't be treated as a new, unhandled error.

PR #805, opened 2025-09-17
Set up pydoclint

aaron-siegel:


mkornacker (commented):
I'm confused by the new comments for purely internal classes. Does that have anything to do with pydoclint?

aaron-siegel:
> I'm confused by the new comments for purely internal classes. Does that have anything to do with pydoclint?

They're not new. pydoclint complains if an `__init__()` method has a docstring. (Preferred style is for only the class itself to have a docstring.) All I did was merge the existing `__init__()` method docstrings with their corresponding class docstrings in order for them to pass linting.

PR #811, opened 2025-09-23
Add support for running tests with cockroachdb as backend

amithadke:
Make sure conftest can work with external db
 Add support for vector index creation


tests/conftest.py
```diff
@@ -76,8 +76,19 @@ def init_env(tmp_path_factory: pytest.TempPathFactory, worker_id: int) -> None:
     os.environ['PIXELTABLE_DB'] = f'test_{worker_id}'
     os.environ['PIXELTABLE_PGDATA'] = str(shared_home / 'pgdata')
     os.environ['FIFTYONE_DATABASE_DIR'] = f'{home_dir}/.fiftyone'
-
-    for var in ('PIXELTABLE_HOME', 'PIXELTABLE_CONFIG', 'PIXELTABLE_DB', 'PIXELTABLE_PGDATA', 'FIFTYONE_DATABASE_DIR'):
+    reinit_db = True
+    if os.environ.get('PIXELTABLE_DB_CONNECT_STR') is not None:
```
aaron-siegel, line 80: If this is used only for testing, call it `PXTTEST_DB_CONNECT_STR` to highlight that it's not part of the Pixeltable config system.

amithadke: We use PIXELTABLE_DB_CONNECT_STR  in cloud tests/setting up tenant db in worker. Lets keep it for now, I'll check if we can set this in config.init then change the name to something internal

tests/conftest.py
```diff
@@ -76,8 +76,19 @@ def init_env(tmp_path_factory: pytest.TempPathFactory, worker_id: int) -> None:
...
+
+    for var in (
+        'PIXELTABLE_HOME',
+        'PIXELTABLE_CONFIG',
+        'PIXELTABLE_DB',
+        'PIXELTABLE_PGDATA',
+        'FIFTYONE_DATABASE_DIR',
+        'PIXELTABLE_DB_CONNECT_STR',
+    ):
         print(f'{var:21} = {os.environ[var]}')
```
aaron-siegel: change 21 to 25 to preserve alignment

PR #812, opened 2025-09-23
Images and arrays in json structures, plus improved storage of array columns

mkornacker:
- allows json structures with inlined PIL images and ndarrays, which are stored in an external file
- stores ndarrays in Array columns above a certain size in an external file
- extends DataRow to also store cell values/cell md, in addition to slot data
- introduces two new ExecNodes: CellMaterializationNode and CellReconstructionNode
- also includes the planner changes required for these nodes
- JsonPath also handles reconstruction of json with inlined objects

### DIFF
--- a/pixeltable/exec/globals.py
+++ b/pixeltable/exec/globals.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import dataclasses
 
 from pixeltable.exprs import ArrayMd
+from pixeltable.utils.misc import non_none_dict_factory
 
 INLINED_OBJECT_MD_KEY = '__pxtinlinedobjmd__'
 
@@ -25,7 +26,7 @@ class InlinedObjectMd:
             return cls(**d)
 
     def as_dict(self) -> dict:
-        result = dataclasses.asdict(self, dict_factory=lambda d: {k: v for (k, v) in d if v is not None})
+        result = dataclasses.asdict(self, dict_factory=non_none_dict_factory)
         if self.array_md is not None:
             result['array_md'] = self.array_md.as_dict()
         return result
--- a/pixeltable/exprs/data_row.py
+++ b/pixeltable/exprs/data_row.py
@@ -17,6 +17,7 @@ import sqlalchemy as sql
 import pixeltable.utils.image as image_utils
 from pixeltable import catalog, env
 from pixeltable.utils.local_store import TempStore
+from pixeltable.utils.misc import non_none_dict_factory
 
 
 @dataclasses.dataclass
@@ -34,7 +35,7 @@ class ArrayMd:
 
     def as_dict(self) -> dict:
         # dict_factory: suppress Nones
-        x = dataclasses.asdict(self, dict_factory=lambda d: {k: v for (k, v) in d if v is not None})
+        x = dataclasses.asdict(self, dict_factory=non_none_dict_factory)
         return x
 
 
@@ -68,7 +69,7 @@ class CellMd:
         return x
 
     def as_dict(self) -> dict:
-        x = dataclasses.asdict(self, dict_factory=lambda d: {k: v for (k, v) in d if v is not None})
+        x = dataclasses.asdict(self, dict_factory=non_none_dict_factory)
         return x
 
 
--- a/pixeltable/exprs/row_builder.py
+++ b/pixeltable/exprs/row_builder.py
@@ -11,6 +11,7 @@ import sqlalchemy as sql
 
 from pixeltable import catalog, exceptions as excs, exprs, utils
 from pixeltable.env import Env
+from pixeltable.utils.misc import non_none_dict_factory
 
 from .data_row import DataRow
 from .expr import Expr, ExprScope
@@ -479,9 +480,7 @@ class RowBuilder:
                         table_row.append(sql.sql.null())
                     else:
                         # we want to minimize the size of the stored dict and use dict_factory to remove Nones
-                        md = dataclasses.asdict(
-                            data_row.cell_md[col.id], dict_factory=lambda d: {k: v for (k, v) in d if v is not None}
-                        )
+                        md = dataclasses.asdict(data_row.cell_md[col.id], dict_factory=non_none_dict_factory)
                         assert len(md) > 0
                         table_row.append(md)
                 if slot_idx is not None and data_row.has_exc(slot_idx):
--- a/pixeltable/utils/misc.py
+++ b/pixeltable/utils/misc.py
@@ -1,0 +1,5 @@
+from typing import Any
+
+
+def non_none_dict_factory(d: list[tuple[str, Any]]) -> dict:
+    return {k: v for (k, v) in d if v is not None}


pixeltable/catalog/column.py
```diff
@@ -259,7 +259,12 @@ def stores_cellmd(self) -> bool:
         # default: record errors for computed and media columns
         if self._stores_cellmd is not None:
             return self._stores_cellmd
-        return self.is_stored and (self.is_computed or self.col_type.is_media_type())
+        return self.is_stored and (
+            self.is_computed
+            or self.col_type.is_media_type()
```
aaron-siegel, line 264: Would it make sense to introduce a  `ColumnType.stores_cell_md()`?

pixeltable/exec/cell_materialization_node.py
```diff
@@ -0,0 +1,227 @@
...
+    - subsume all cell materialization
+    """
+
+    output_col_info: list[exprs.ColumnSlotIdx]
+
+    # execution state
+    inlined_obj_files: list[Path]  # only [-1] is open for writing
+    buffered_writer: io.BufferedWriter | None  # BufferedWriter for inlined_obj_files[-1]
+
+    MIN_FILE_SIZE = 8 * 2**20  # 8MB
```
aaron-siegel, line 58: technically it's 8 MiB, not 8 MB :)

pixeltable/exec/cell_materialization_node.py
```diff
@@ -0,0 +1,227 @@
...
+    MAX_DB_ARRAY_SIZE = 512  # max size of array stored in table column; in bytes
```
aaron-siegel, line 59: This seems awfully small to me. I'd expected it to be on the order of a typical index vector (~ 3072 * 4 bytes, say) as arrays of that size are routinely stored in the DB anyway. But I confess I don't have a good feel for the use cases here.

I wonder if there's a future where this is tuneable in some way.

pixeltable/exec/cell_materialization_node.py
```diff
@@ -0,0 +1,227 @@
...
+        np.save(self.buffered_writer, ar, allow_pickle=False)
+        end = self.buffered_writer.tell()
+        self._flush_full_buffer()
+        return url_idx, start, end, is_bool_array, shape
+
+    def _write_inlined_image(self, img: PIL.Image.Image) -> tuple[int, int, int]:
+        """Write a PIL image to buffered_writer and return: index into inlined_obj_files, start offset, end offset"""
+        url_idx = len(self.inlined_obj_files) - 1
+        start = self.buffered_writer.tell()
+        format = 'webp' if img.has_transparency_data else 'jpeg'
```
aaron-siegel: This specific logic is fairly bespoke and recurs several times in the codebase now; maybe factor it out in utils somewhere?

pixeltable/exec/cell_materialization_node.py
```diff
@@ -0,0 +1,227 @@
...
+    def _reset_buffer(self) -> None:
+        local_path = LocalStore(Env.get().media_dir)._prepare_path_raw(
+            self.row_builder.tbl.id, 0, self.row_builder.tbl.version
+        )
+        self.inlined_obj_files.append(local_path)
+        fh = open(local_path, 'wb', buffering=self.MIN_FILE_SIZE * 2)  # noqa: SIM115
+        assert isinstance(fh, io.BufferedWriter)
+        self.buffered_writer = fh
+
+    def _flush_full_buffer(self, finalize: bool = False) -> None:
```
aaron-siegel: Maybe just call this just `_flush_buffer()` since it doesn't have to be full if `finalize=True`.

A 1-sentence docstring would be useful too

pixeltable/exec/cell_materialization_node.py
```diff
@@ -0,0 +1,227 @@
...
+            return {k: self._rewrite_json(v) for k, v in element.items()}
+        if isinstance(element, np.ndarray):
+            url_idx, start, end, is_bool_array, shape = self._write_inlined_array(element)
+            return {
+                '__pxttype__': ts.ColumnType.Type.ARRAY.name,
+                '__pxturlidx__': url_idx,
+                '__pxtstart__': start,
+                '__pxtend__': end,
+                '__pxtisboolarray__': is_bool_array,
+                '__pxtarrayshape__': shape,
```
aaron-siegel: We can save a few bytes here by only writing `__pxtarrayshape__` if `is_bool_array=True`. (I notice you're already doing this for bare arrays)

pixeltable/exec/cell_reconstruction_node.py
```diff
@@ -0,0 +1,140 @@
...
+        self.file_handles = {}
+
+    async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
+        async for batch in self.input:
+            for row in batch:
+                for col_ref in self.json_refs:
+                    val = row[col_ref.slot_idx]
+                    if val is None:
+                        continue
+                    cell_md = row.slot_cellmd.get(col_ref.slot_idx)
```
aaron-siegel: It's a super minor thing but the use of `cell_md` in some places and `cellmd` in others bugs me a little. Standardize on one or the other (I vote for `cell_md` since we use `tbl_md` consistently)?

pixeltable/exprs/data_row.py
```diff
@@ -17,6 +18,31 @@
...
+
+    errortype: str | None = None
+    errormsg: str | None = None
+
+    # a list of file urls that are used to store images and arrays; only set for json and array columns
+    # for json columns: a list of all urls referenced in the column value
+    # for array columns: a single url
+    file_urls: list[str] | None = None
+
+    # for array cells that are stored externally, the start and end offsets
```
aaron-siegel: Suggestion: move these into a separate `ArrayCellMd` dataclass or some such with fields `start`, `end`, etc. This will actually save bytes in the DB since we won't have to repeat the `array_` prefix four times. And I can see `CellMd` becoming pretty crowded if we don't introduce some substructure into it. (And we may at some point want to query for specific components of cell_md rather than the whole structure, and it'll be a lot more convenient to fetch `cell_md.array_info` or whatever rather than four separate values in the SELECT.)

The structure `ArrayCellMd` or whatever we call it could be reused inside JSON rewrites in the same way, also saving a few bytes (`__pxtarrayisbool__` etc are fairly verbose prefixes)

pixeltable/exprs/json_path.py
```diff
@@ -67,15 +81,15 @@ def _from_dict(cls, d: dict, components: list[Expr]) -> JsonPath:
         return cls(anchor, path_elements, d['scope_idx'])
 
     @property
-    def _anchor(self) -> Optional[Expr]:
+    def anchor(self) -> Optional[Expr]:
```
aaron-siegel, line 84: These expression types are publicly visible so we may actually want to keep this protected.

pixeltable/exprs/json_path.py
```diff
@@ -159,12 +173,30 @@ def _json_path(self) -> str:
...
+            return
+
+        # the origin of val is a json-typed column, which might stored inlined objects
+        if self.anchor.slot_idx not in row.slot_cellmd:
+            # we can infer that there aren't any inlined objects because our execution plan doesn't include
+            # materializing the cellmd (eg, insert plans)
+            # TODO: have the planner pass that fact into ExprEvalNode explicitly to streamline this path a bit more
+            return
+
+        from pixeltable.exec.cell_reconstruction_node import json_has_inlined_objs, reconstruct_json
```
aaron-siegel, line 193: Move to top of method

pixeltable/metadata/converters/convert_40.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # Get the list of column names that need _cellmd columns
+    _logger.info(f'Checking table {orig_table_md["name"]} ({store_name})')
+    col_ids = find_target_columns(orig_table_md)
+    if len(col_ids) == 0:
+        _logger.info(f'No Array or Json columns found in table {orig_table_md["name"]}. Skipping migration.')
+        return
+
+    # Check which columns already exist in the table
+    check_columns_sql = sql.text(f"""
```
aaron-siegel, line 29: Is there no way to reference the information schema with idiomatic SQLAlchemy?

tests/utils.py
```diff
@@ -443,19 +467,47 @@ def __file_comparer(x: str, y: str) -> bool:
     return sha256sum(x) == sha256sum(y)
 
 
+def __img_comparer(img1: PIL.Image.Image, img2: PIL.Image.Image) -> bool:
```
aaron-siegel: The existing tests just use `__equality_comparer` to compare images (in general `x == y` should do the right thing if `x` and `y` are PIL images). Is there some reason why equality comparison isn't adequate for the new tests? If so please add a comment to explain.

tests/utils.py
```diff
@@ -276,6 +277,24 @@ def create_scalars_tbl(num_rows: int, seed: int = 0, percent_nulls: int = 10) ->
     return tbl
 
 
+def create_arrays(
```
aaron-siegel: I'd all this `infinite_array_iterator` or some such to make it really clear what it does (it's an unusual pattern)

tests/utils.py
```diff
@@ -367,6 +386,11 @@ def get_image_files(include_bad_image: bool = False) -> list[str]:
     return __IMAGE_FILES_WITH_BAD_IMAGE if include_bad_image else __IMAGE_FILES
 
 
+def image_iterator(include_bad_image: bool = False) -> Iterator[PIL.Image.Image]:
```
aaron-siegel: Likewise `infinite_image_iterator`

pixeltable/catalog/column.py
mkornacker: I don't think so, because that concept only applies at the column level.

pixeltable/exec/cell_materialization_node.py
mkornacker: Nobody with a CS degree would think of MB not meaning 1024^2 :)

pixeltable/exec/cell_materialization_node.py
mkornacker: It's on the order of the size of a string. We should avoid putting large objects in here, since they need to be stored for every version (that's not true for vector columns used for an index).

pixeltable/exec/cell_materialization_node.py
mkornacker: I can leave a todo.

pixeltable/exec/cell_materialization_node.py
mkornacker: Done.

pixeltable/exprs/json_path.py
mkornacker: Why? It does some work, and it may be avoidable.

tests/utils.py
mkornacker: Done.

tests/utils.py
mkornacker: Done.

tests/utils.py
mkornacker: Done.

(Gemini thought this couldn't be done.)

pixeltable/metadata/converters/convert_40.py
mkornacker: Doesn't look that way.

pixeltable/exec/cell_materialization_node.py
aaron-siegel: Ok, that makes sense.

pixeltable/catalog/column.py
aaron-siegel: How about `can_store_cellmd()`? Really any time you have to say `is_x_type() or is_y_type() or is_z_type()` it suggests that there's a missing property on `ColumnType`. I don't care what we call it :)

pixeltable/exec/cell_materialization_node.py
aaron-siegel: Well calling it a "megabyte" is an ancient and unfortunate misnomer that has caused its share of confusion, and there's an easy fix for it that's gaining currency. "mebibyte" is the official prefix now: "Use of mebibyte and related units is strongly endorsed by IEEE and CIPM and is becoming more common." Per wikipedia. So I try to do my part.

But my degrees are in mathematics and computer engineering so I'm just barely not a counterexample to your claim.

pixeltable/exprs/json_path.py
aaron-siegel: Preferred style is to put imports at the top of the widest permissible scope. i.e., any time one narrows the scope of an import, there should be a specific reason to do so.

Imports in Python incur a one-time cost and after that are optimized away, so any performance impact is very small (I'd expect it to be on the order of microseconds per Python process lifecycle). If I understand things right, your reason for wanting to put the import here is to avoid doing that initialization work in cases where an application never encounters this code path during its lifecycle (i.e., it does contain JsonPath expressions, but none of its JsonPath expressions ever get checked for cellmd). If you think it's important to do that then ok, but please add a code comment explaining why the import needs to stay here.

pixeltable/exec/cell_materialization_node.py
mkornacker: Created a default_format() utility.

pixeltable/exprs/data_row.py
mkornacker: Done.

pixeltable/exprs/json_path.py
mkornacker: But so is eval().

pixeltable/exprs/json_path.py
mkornacker: Done.

pixeltable/catalog/column.py
mkornacker: But it's a column-level concept, not a type concept.

pixeltable/exec/cell_materialization_node.py
mkornacker: Done.

pixeltable/exec/cell_reconstruction_node.py
mkornacker: Done.

pixeltable/catalog/column.py
aaron-siegel: "cellmd is stored here" is a column-level concept. "this type always stores cellmd" is a type-level concept, one that you're implicitly invoking here by saying `self.col_type.is_x() or self.col_type.is_y() or self.col_type.is_z()`. There may be additional codepaths in the future where we need to invoke that concept, and/or additional types that store cellmd. All I'm saying is it feels cleaner to me, and easier to maintain in those cases, to factor out that type-level concept as a method of ColumnType. But I'm not going to insist on it

pixeltable/exprs/data_row.py
```diff
@@ -13,15 +14,71 @@
...
+        if 'array_md' in d:
+            d2 = d.copy()
+            del d2['array_md']
+            x = cls(**d2, array_md=ArrayMd(**d['array_md']))
+        else:
+            x = cls(**d)
+        return x
+
+    def as_dict(self) -> dict:
+        x = dataclasses.asdict(self, dict_factory=lambda d: {k: v for (k, v) in d if v is not None})
```
aaron-siegel: The expression `lambda d: {k: v for (k, v) in d if v is not None}` shows up a lot; maybe this should be defined via a resuable `def non_null_dict_factory()` somewhere.

pixeltable/exec/globals.py
```diff
@@ -0,0 +1,31 @@
...
+
+import dataclasses
+
+from pixeltable.exprs import ArrayMd
+
+INLINED_OBJECT_MD_KEY = '__pxtinlinedobjmd__'
+
+
+@dataclasses.dataclass
+class InlinedObjectMd:
```
aaron-siegel, line 12: I'm curious why this doesn't live in the same place as `ArrayMd`? (what's the reasoning behind putting them in different places?)

aaron-siegel (approved):

pixeltable/exec/globals.py
mkornacker: It's only read and written by code in exec.

pixeltable/exprs/data_row.py
mkornacker: Done.

PR #814, opened 2025-09-24
Concurrency fixes & random_tbl_ops v2

aaron-siegel:
Introduces a v2 script for random_tbl_ops with:
- Column operations
- New framework with more configurability
- Statistics reporting (failure rates by operation, error summary)

Fixes a number of associated concurrency issues that could lead to fatal errors or database corruption:
- Update `Catalog._modified_tvs` in all relevant places
- Unify error handling for DB errors (previously, computed columns were covered by a separate code path)
- Introduce a rollback_action() mechanism for safely rolling back changes that are not encompassed by _modified_tvs invalidation (see note on this below)

Also fixes a bug in `create_table()`: If both a `source` and `if_exists='ignore'` were specified, then data from `source` were still inserted even if the table already exists.

Rollback actions:
- Previously, rollbacks were handled by a try/finally block with run_cleanup_on_exception() in the finally clause. In both cases where it was being used, the try/finally block did not encompass all of the transaction's DB write operations. Occasionally a transaction would fail during a write operation *after* the try/finally block had concluded, leading to a dirty metadata state.
- This PR replaces run_cleanup_on_exception() with a rollback action registration mechanism: rollback actions can be registered with Env at any time during a transaction. If the transaction succeeds, they are discarded. If the transaction fails, then they are applied in reverse order during end-of-transaction cleanup. This ensures they will always be applied at the end of a failed transaction.


pixeltable/catalog/catalog.py
```diff
@@ -12,6 +12,7 @@
 
 import psycopg
 import sqlalchemy as sql
+import sqlalchemy.exc as sql_exc
```
mkornacker, line 15: Is this needed, as opposed to just sql.exc? Just wondering if I'm missing something here...

pixeltable/env.py
```diff
@@ -222,6 +226,10 @@ def default_time_zone(self, tz: Optional[ZoneInfo]) -> None:
         self.engine.dispose()
         self._create_engine(time_zone_name=tz_name)
 
+    @property
+    def verbosity(self) -> int:
```
mkornacker, line 227: Of what?

pixeltable/env.py
```diff
@@ -93,6 +95,7 @@ class Env:
     _current_conn: Optional[sql.Connection]
     _current_session: Optional[orm.Session]
     _current_isolation_level: Optional[Literal['REPEATABLE_READ', 'SERIALIZABLE']]
+    _current_tx_rollback_actions: list[Callable[[], None]]
```
mkornacker: Let's move that into Catalog. Also, xact is used elsewhere in the codebase to mean 'transaction' (that's the standard abbreviation in the dbms world), let's stay consistent.

Also, standard dbms terminology for this would be "undo action" (as in, "redo/undo log entries").

pixeltable/env.py
```diff
@@ -268,15 +277,24 @@ def begin_xact(self, for_write: bool = False) -> Iterator[sql.Connection]:
                     self._current_conn = conn
                     self._current_session = session
                     yield conn
+            except Exception:
+                for hook in reversed(self._current_tx_rollback_actions):
+                    run_cleanup(hook, raise_error=False)
```
mkornacker: Can this actually tolerate multiple ctrl-c's in a row, or will the undo action get aborted?

pixeltable/catalog/table_version_path.py
```diff
@@ -66,14 +65,7 @@ def as_md(self) -> schema.TableVersionPath:
...
         from pixeltable.catalog import Catalog
 
-        if Env.get().in_xact:
-            # when we're running inside a transaction, we need to make sure to supply current metadata;
-            # mixing stale metadata with current metadata leads to query construction failures
-            # (multiple sqlalchemy Table instances for the same underlying table create corrupted From clauses)
-            if self._cached_tbl_version is not None and self._cached_tbl_version.is_validated:
-                # nothing to refresh
-                return
-        elif self._cached_tbl_version is not None:
```
mkornacker, line 76: Doesn't this mean that every "t.<col>" ends up re-validating the metadata? We intentionally use a potentially outdated metadata cache here for better interactive performance. (It's revalidated on execution anyway.)

pixeltable/catalog/table_version.py
```diff
@@ -295,7 +291,14 @@ def create(
         cat = pxt.catalog.Catalog.get()
 
         tbl_id = UUID(hex=inital_md.tbl_md.tbl_id)
+        assert (tbl_id, None) not in cat._tbl_versions
         tbl_version = cls(tbl_id, inital_md.tbl_md, inital_md.version_md, None, inital_md.schema_version_md, [])
+
+        @env.register_rollback_action
```
mkornacker: Just fyi, this is going to be replaced with a _create(), analogous to View._create(). It will only return metadata to the catalog, not instantiate anything in the catalog itself.

pixeltable/catalog/catalog.py
aaron-siegel: `exc` isn't actually brought into scope by `sqlalchemy`, so it needs to be imported separately. mypy doesn't catch this but pylance does (in the vscode IDE).

pixeltable/catalog/table_version_path.py
aaron-siegel: Yeah, I wasn't 100% sure this was the best solution. But as things currently stand, you can end up in a situation where a handle remains invalid indefinitely and never recovers.

Example:
- Another process drops a column
- I do `t.collect()` and get a "dropped column" error
- I do `t.collect()` and get a "dropped column" error
- I do `t.collect()` and get a "dropped column" error
- .....

The indefinite invalidation persists even if I try to "reload" by table via `t = pxt.get_table('my_table')`. (As a Pixeltable developer I understand it doesn't force anything to "reload", but as a user I'd be surprised by this.)

So I think it's important to make some change now. There are less drastic things we could do than revalidating on every `t.`, but I do think there's an argument it's the right approach and I'm not convinced it will observably impact interactive performance.

pixeltable/env.py
aaron-siegel: Console output. It exposes the contents of the `PIXELTABLE_VERBOSITY` config parameter.

Since it's a global config parameter and "verbosity" has a globally defined meaning in Pixeltable context, I don't think this needs a qualifier.

pixeltable/env.py
aaron-siegel: Probably not. Do we want that to be a requirement? There's only so much we can do if the user spams the system with Ctrl-C's, right?

I mean, it wouldn't be hard to put this in a while loop and trap those Ctrl-C's but there's still some chance an ill-timed sequence of Ctrl-C's breaks out of it.

pixeltable/catalog/table_version.py
```diff
@@ -697,7 +684,24 @@ def _add_columns(
         self, cols: Iterable[Column], print_stats: bool, on_error: Literal['abort', 'ignore']
     ) -> UpdateStatus:
         """Add and populate columns within the current transaction"""
+        from pixeltable.catalog import Catalog
+        from pixeltable.plan import Planner
+
         cols_to_add = list(cols)
+
+        @env.register_rollback_action
```
mkornacker: This isn't necessary, you already mark this instance is modified.

But would be good to do this here explicitly, since mark_modified_tvs() is idempotent, and it's critical to have that here.

pixeltable/catalog/catalog.py
```diff
@@ -423,11 +403,49 @@ def begin_xact(
...
                 if has_exc:
-                    # purge all modified TableVersion instances, we can't guarantee they are still consistent with the
+                    # purge all modified TableVersion instances; we can't guarantee they are still consistent with the
                     # stored metadata
                     for handle in self._modified_tvs:
                         self._clear_tv_cache(handle.id, handle.effective_version)
-                self._modified_tvs = set()
+                self._modified_tvs.clear()
+
+    def raise_from_sql_exc(
```
mkornacker: convert_sql_exc()? I had to look this up a few times to remind myself that it's raising a pxt.Error.

pixeltable/catalog/table_version.py
```diff
@@ -1398,35 +1387,42 @@ def version(self) -> int:
...
-        assert self.effective_version is None
-        self._version_md.created_at = ts
-
     @property
     def schema_version(self) -> int:
         return self._schema_version_md.schema_version
 
-    @schema_version.setter
-    def schema_version(self, version: int) -> None:
+    def bump_version(self, timestamp: Optional[float] = None, *, bump_schema_version: bool) -> None:
```
mkornacker, line 1386: Nice!

(This deserves a comment spelling out the semantics.)

pixeltable/catalog/table_version_handle.py
```diff
@@ -79,10 +82,14 @@ class ColumnHandle:
 
     def get(self) -> 'Column':
         if self.col_id not in self.tbl_version.get().cols_by_id:
-            schema_version_drop = self.tbl_version.get()._tbl_md.column_md[self.col_id].schema_version_drop
+            if self.col_id in self.tbl_version.get()._tbl_md.column_md:
```
mkornacker: explanatory comment: why might this not be in column_md

pixeltable/env.py
aaron-siegel: Done. I also made register_undo_action into a method of `Catalog` rather than a global, in anticipation of multiple catalog instances.

pixeltable/catalog/table_version.py
aaron-siegel: Ok, so it isn't necessary to call `self.store_tbl.create_sa_tbl()` at the end? I thought that was the point of the undo action.

pixeltable/catalog/catalog.py

pixeltable/catalog/table_version.py
aaron-siegel: Done.

pixeltable/catalog/table_version_handle.py
aaron-siegel: Hmm, I don't think it can be anymore - I think this was a workaround I put in for a bug that I've since fixed. I reverted the change for now.

pixeltable/catalog/catalog.py
```diff
@@ -200,6 +202,7 @@ def __init__(self) -> None:
         self._in_write_xact = False
         self._x_locked_tbl_ids = set()
         self._modified_tvs = set()
+        self._current_tx_undo_actions = []
```
mkornacker: 'xact' is used elsewhere in the code to mean 'transaction' (following common practice in the dbms community), let's stick with that.

Also, the other transaction state (modified_tvs, x_locked_tbl_ids) doesn't mention 'current xact', so fine to omit.

pixeltable/catalog/catalog.py
```diff
@@ -409,16 +412,45 @@ def begin_xact(
                         self._clear_tv_cache(handle.id, handle.effective_version)
                 self._modified_tvs.clear()
 
-    def raise_from_sql_exc(
-        self, e: sql_exc.StatementError, tbl_id: UUID | None, tbl: TableVersionHandle | None, convert_db_excs: bool
+    @contextmanager
+    def __begin_guarded_xact(self, *, for_write: bool) -> Iterator[sql.Connection]:
```
mkornacker: Missing comment. When do I use this?

pixeltable/catalog/catalog.py
```diff
@@ -409,16 +412,45 @@ def begin_xact(
...
+        assert not self._current_tx_undo_actions
+        with Env.get().begin_xact(for_write=for_write) as conn:
+            try:
+                yield conn
+            except:
+                for hook in reversed(self._current_tx_undo_actions):
```
mkornacker: Why not do this in begin_xact()?

pixeltable/catalog/table_version.py
```diff
@@ -734,40 +725,25 @@ def _add_columns(
...
+                # If it wasn't converted, re-raise as a generic Pixeltable error
+                # (this means it's not a known concurrency error; it's something else)
+                raise excs.Error(
+                    f'Unexpected SQL error during execution of computed column {col.name!r}:\n{exc}'
+                ) from exc
+            if excs_per_col > 0:
+                cols_with_excs.append(col)
+                num_excs += excs_per_col
+            computed_values += plan.ctx.num_computed_exprs * row_count
+            plan.close()
```
mkornacker: This needs to happen in a finally block.

pixeltable/catalog/table_version_path.py
mkornacker: Isn't this just a problem with a concurrent drop_column() specifically? That should get fixed independently.

But checking for validated here is basically removing the cache.

pixeltable/catalog/catalog.py
aaron-siegel: Changed to `_undo_actions`.

pixeltable/catalog/catalog.py
aaron-siegel: Never - that's why it's marked private. It should only be called by begin_xact. I added a code comment to clarify this.

pixeltable/catalog/catalog.py
aaron-siegel: begin_xact() is getting awfully crowded. I started putting this in begin_xact() but to be correct it would have added a third level of try/finally nesting, which I felt made begin_xact() almost unreadable. That's why I broke it out.

I'm of the opinion that begin_xact() should be broken down further into more digestible chunks (it's already quite hard to read) but I wasn't aiming to solve that problem here, just trying to avoid making things worse.

aaron-siegel: Moved it into the existing `finally` block with assertion safeguards against any stray undo actions, per discussion earlier today

pixeltable/catalog/table_version.py
aaron-siegel: oops good catch. don't know how that happened

PR #816, opened 2025-09-25
Replace `create_replica()` with separate `publish()` and `replicate()` methods

aaron-siegel:


pixeltable/globals.py
```diff
@@ -541,13 +556,17 @@ def drop_table(
...
         tbl_path = table
 
+    if_not_exists_ = catalog.IfNotExistsParam.validated(if_not_exists, 'if_not_exists')
+
     if tbl_path.startswith('pxt://'):
         # Remote table
+        if force:
+            raise excs.Error('Cannot use `force=True` with a cloud replica URI.')
+        # TODO: Handle if_not_exists properly
         share.delete_replica(tbl_path)
```
mkornacker, line 566: The implementation still has a TODO that can probably be removed.

pixeltable/globals.py
```diff
@@ -541,13 +556,17 @@ def drop_table(
         assert isinstance(table, str)
         tbl_path = table
 
+    if_not_exists_ = catalog.IfNotExistsParam.validated(if_not_exists, 'if_not_exists')
+
     if tbl_path.startswith('pxt://'):
         # Remote table
+        if force:
+            raise excs.Error('Cannot use `force=True` with a cloud replica URI.')
```
mkornacker, line 564: Why not?

pixeltable/globals.py
aaron-siegel: Fixed.

pixeltable/globals.py
aaron-siegel: Because it hasn't been implemented yet (in the protocol). We can remove this check once it is.

PR #817, opened 2025-09-25
Add config parameter for global media destination

jpeterson-pxt:
The configuration parameter pixeltable, media_destination can be used setup a global media destination. 
The environment variable PIXELTABLE_MEDIA_DESTINATION can set this configuration, or the user can put it a value for media_destination in the pixeltable section of their config.toml file. 

The format of the string is the same as that used in the media_destination parameter to a table column. 
These destination parameters must at least identify the service and the bucket. Sometimes account numbers are required, or URI destinations. Optional prefixes are possible in all cases. Local file paths or strings are possible as well.

Access credentials are checked on the first attempt to read from or write to the designated destination. 
Prefixes / directories are created as needed.

For files:
PIXELTABLE_MEDIA_DESTINATION=/Users/<a name>/<some directories>

For S3:
PIXELTABLE_MEDIA_DESTINATION=s3://<bucket>/<optional prefixes>

For R2:
PIXELTABLE_MEDIA_DESTINATION=https://<account>.r2.cloudflarestorage.com/<bucket>/<optional prefixes>

For GCS:
PIXELTABLE_MEDIA_DESTINATION=gs://<bucket>/<optional prefixes>

### DIFF
--- a/pixeltable/config.py
+++ b/pixeltable/config.py
@@ -163,6 +163,7 @@ KNOWN_CONFIG_OPTIONS = {
         'api_key': 'API key for Pixeltable cloud',
         'r2_profile': 'AWS config profile name used to access R2 storage',
         's3_profile': 'AWS config profile name used to access S3 storage',
+        'media_destination': 'Storage location for generated / inserted media files',
     },
     'anthropic': {'api_key': 'Anthropic API key'},
     'bedrock': {'api_key': 'AWS Bedrock API key'},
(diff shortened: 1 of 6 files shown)


aaron-siegel:
This PR surprised me a little: I was expecting some simple logic for registering a global default in `env`, and having columns use that global default whenever `media_destination` is not specified. Can you give me some context on why we need a separate abstraction for `GlobalMediaDestination`?

jpeterson-pxt:
wrt adding the GlobalMediaDestination object:
- This new object encapsulates the state required to manage the global destination. The state includes:
-- A lock, as the create may happen from within a thread
-- The status (created / not created)
-- The StorageObjectAddress of the destination
-- The path of the destination (if there is one)
- There is no state (other than the global ApiClient) associated with a MediaDestination, so there is no other object to reuse

If you would prefer, I can just embed all these fields in the Env directly, but it seemed useful to encapsulate them into an object.

PR #819, opened 2025-09-26
Add protocol for replica operations.

amithadke:


pixeltable/share/protocol/table.py
```diff
@@ -0,0 +1,107 @@
...
+class TableMetadataEntry(BaseModel):
+    """Entry containing table metadata information."""
+
+    table_id: str
+    table_md: TableMd
+    table_version_md: TableVersionMd
+    table_schema_version_md: TableSchemaVersionMd
+
+
+class ListTableMetadataEntry(BaseModel):
```
aaron-siegel: Do we really need this as a separate model (versus just inlining `list[TableMetadataEntry]` where needed?)

pixeltable/share/protocol/table.py
```diff
@@ -0,0 +1,107 @@
...
+
+from pydantic import AnyUrl, BaseModel
+
+from pixeltable.metadata.schema import TableMd, TableSchemaVersionMd, TableVersionMd
+
+from .common import PathUri, PathUriRequestModel, StorageDestination
+from .operation_types import CoreOperationType
+
+
+class TableMetadataEntry(BaseModel):
```
aaron-siegel: This is basically identical to `schema.FullTableMd`. Can we just use that? There are no validators here, so no particular reason to desire a Pydantic model.

pixeltable/share/protocol/table.py
```diff
@@ -0,0 +1,107 @@
...
+    table_schema_version_md: TableSchemaVersionMd
+
+
+class ListTableMetadataEntry(BaseModel):
+    """List of table metadata entries."""
+
+    tables: list[TableMetadataEntry]
+
+
+class PublishSnapshotRequest(PathUriRequestModel):
```
aaron-siegel: `PublishRequest` (it's not necessarily a snapshot)

pixeltable/share/protocol/table.py
```diff
@@ -0,0 +1,107 @@
...
+class PublishSnapshotResponse(BaseModel):
+    """Response from publishing a table snapshot."""
+
+    upload_id: UUID
+    destination: StorageDestination
+    destination_uri: AnyUrl
+    max_size: Optional[int] = None  # Maximum size that can be used by this snapshot, used for R2 home buckets
+
+
+class FinalizeSnapshotRequest(PathUriRequestModel):
```
aaron-siegel: `FinalizeRequest`

pixeltable/share/protocol/table.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class DeleteSnapshotResponse(BaseModel):
+    """Response from deleting a table snapshot."""
+
+    table_uri: PathUri
+    version: Optional[int] = None
+
+
+class CloneSnapshotRequest(PathUriRequestModel):
```
aaron-siegel: Let's call this `ReplicateRequest` to align with the new terminology

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+        """Get the table identifier (either UUID or path)."""
+        components = self.get_components()
+        return components.path_or_uuid
+
+
+class PathUriParser:
+    """Helper class to parse PathUri strings into components."""
+
+    @staticmethod
+    def parse(uri: str) -> PathUriComponents:
```
aaron-siegel: Can we not use `urllib.parse.urlparse` as a starting point rather than doing all the work manually?

pixeltable/share/protocol/operation_types.py
```diff
@@ -0,0 +1,58 @@
...
+
+# Define the operation sets as module-level constants
+TABLE_OPERATIONS: Set[CoreOperationType] = {
+    CoreOperationType.PUBLISH_SNAPSHOT,
+    CoreOperationType.FINALIZE_SNAPSHOT,
+    CoreOperationType.CLONE_SNAPSHOT,
+    CoreOperationType.DELETE_SNAPSHOT,
+}
+
+NAMESPACE_OPERATIONS: Set[CoreOperationType] = {
```
aaron-siegel: Use lowercase `set` (camel cased collection types are Python 3.8 syntax)

pixeltable/share/packager.py
```diff
@@ -55,22 +56,23 @@ class TablePackager:
...
         with catalog.Catalog.get().begin_xact(for_write=False):
             tbl_md = catalog.Catalog.get().load_replica_md(table)
-            self.md = {
-                'pxt_version': pxt.__version__,
-                'pxt_md_version': metadata.VERSION,
-                'md': {'tables': [md.as_dict() for md in tbl_md]},
-            }
-        if additional_md is not None:
-            self.md.update(additional_md)
+            self.request = PublishSnapshotRequest(
```
aaron-siegel: I feel like this is munging abstractions a bit too much. `TablePackager` shouldn't know anything about "requests"; its role is purely to package up table data into tarballs. Can you comment on why this change is needed?

If the goal is to use this as a way to do validation on the inputs, then to me it'd be cleaner for the client of TablePackager to do that. The only thing that really needs to be validated at this stage is the URI, correct? Everything else is Pixeltable-generated. And TablePackager doesn't even need the URI.

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+        """Parse a PathUri string into its components."""
+        match = PATTERN_PXT_URI.match(uri)
+        if not match:
+            raise ValueError('Invalid URI format. Expected pxt://<org_slug>[:<db_slug>]/<table_path_or_uuid>')
```
aaron-siegel: Code duplication with validate_pxt_uri(); can that be cleaned up? (i.e., can our validation simply be to call parse() and see if it succeeds?)

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+
+        org_slug, db_slug, path_part = match.groups()
+        path_or_uuid = path_part.lstrip('/') if path_part else ''
+
+        # Safely strip whitespace, handling None values
```
aaron-siegel: Do we really want to be stripping whitespace?

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+    def get_identifier(self) -> str:
+        """Get the table identifier (either UUID or path)."""
+        components = self.get_components()
+        return components.path_or_uuid
+
+
+class PathUriParser:
+    """Helper class to parse PathUri strings into components."""
+
+    @staticmethod
```
aaron-siegel: `@staticmethod` is discouraged except in rare special cases. `@classmethod` is preferred unless there's a specific reason not to do so.

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+        """Check if this URI uses UUID format."""
+        return PathUriParser.parse(uri).is_uuid
+
+
+class PathUriRequestModel(BaseModel):
+    """Base model for requests that need path URI for slug identification."""
+
+    def get_org_slug(self) -> Optional[str]:
+        """Extract the org_slug from a table_uri if present."""
+        if hasattr(self, 'table_uri'):
```
aaron-siegel: Not ideal to rely on `hasattr` here. I'd advocate for one of the following approaches instead:
- Have an abstract method on PathUriRequestModel that returns the uri, or
- Make the individual methods `get_org_slug` etc abstract

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+        components = PathUriParser.parse(uri)
+        return components.path_or_uuid if components.is_uuid else None
+
+    @staticmethod
+    def is_uuid_uri(uri: str) -> bool:
+        """Check if this URI uses UUID format."""
+        return PathUriParser.parse(uri).is_uuid
+
+
+class PathUriRequestModel(BaseModel):
```
aaron-siegel: Maybe call this `RequestBaseModel` or some such? To me that's a more natural way to think about it (the fact that every request has an associated path URI is more incidental)

pixeltable/share/protocol/operation_types.py
```diff
@@ -0,0 +1,58 @@
...
+pixeltable core and cloud implementations.
+"""
+
+from __future__ import annotations
+
+from enum import Enum
+from typing import Set
+
+
+class CoreOperationType(str, Enum):
```
aaron-siegel: Are we going to have non-core operations?

pixeltable/share/publish.py
```diff
@@ -27,24 +36,22 @@
 def push_replica(
     dest_tbl_uri: str, src_tbl: pxt.Table, bucket: str | None = None, access: Literal['public', 'private'] = 'private'
 ) -> str:
-    packager = TablePackager(
-        src_tbl, additional_md={'table_uri': dest_tbl_uri, 'bucket_name': bucket, 'is_public': access == 'public'}
-    )
-    request_json = packager.md | {'operation_type': 'publish_snapshot'}
-    response = requests.post(PIXELTABLE_API_URL, json=request_json, headers=_api_headers())
+    packager = TablePackager(src_tbl, table_uri=dest_tbl_uri)
+    packager.request.bucket_name = bucket
```
aaron-siegel: See above comment - I'd prefer to separate the request out from the packager; they're two different things that are put together in order to publish.

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+
+class StorageDestination(str, Enum):
+    """Storage destination types for table snapshots."""
+
+    S3 = 's3'
+    R2 = 'r2'
+    GCS = 'gcs'
+
+
+class PathUriComponents(NamedTuple):
```
aaron-siegel: It'd be clearer to me (here and elsewhere) to call these `PxtUri`s rather than `PathUri`s.

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+    """Components of a parsed PathUri."""
+
+    org_slug: str
+    db_slug: Optional[str]
+    path_or_uuid: str
+    is_uuid: bool  # New field to indicate if path_or_uuid is a UUID
```
aaron-siegel: "New field" isn't really meaningful as a code comment since it assumes a temporal reference point that won't always be valid.

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+
+    if path_part:
+        # Remove leading slash and validate the path/UUID part
+        clean_path = path_part.lstrip('/')
+        if not clean_path:
+            raise ValueError('URI must have a path or UUID after the slash')
+    return v
+
+
+class PathUri(str):
```
aaron-siegel: I'm questioning why we need both this and `PathUriComponents`. Is there any reason not to just have one or the other? They seem to capture the exact same information content. Maybe `parse()` could just return the `PathUriComponents` represenation?

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,191 @@
...
+    """Custom URI type for pxt:// URIs with validation."""
+
+    def __new__(cls, value: str) -> 'PathUri':
```
aaron-siegel: What's the reason for implementing `__new__()` here rather than `__init__()`?

pixeltable/share/protocol/table.py
amithadke: TableMetadataEntry we can get rid of this.
FullTableMd doesn't surface table id at top level, may be we can change it or make these classes that will be on wire/disk derive from pydantic BaseModel
In general even with PXT-703 I think we should start adding some native pydantic support to our structures, makes it easy to maintain. Let me know if you prefer that.

pixeltable/share/publish.py
amithadke: yeah i tried to put something together to make it work i think im going to leave packager and wire protocol different and build request from packager's fields inside publish.sh

pixeltable/share/protocol/common.py
amithadke: let me see if its used but i see your point. We can use one and keep all parsing validation internal.

pixeltable/share/protocol/common.py
amithadke: its inherited from str, we have to validate value before str. I guess its also because till today client was sending strings and server was trying to map it to PathUri.

pixeltable/share/protocol/operation_types.py
amithadke: not here inside pixeltable core, they are defined in cloud side
bucket/org/db and now public dataset list operations

pixeltable/catalog/catalog.py
```diff
@@ -1076,7 +1076,7 @@ def _clear_tv_cache(self, tbl_id: UUID, effective_version: Optional[int]) -> Non
             tv.is_validated = False
             del self._tbl_versions[tbl_id, effective_version]
 
-    def create_replica(self, path: Path, md: list[schema.FullTableMd]) -> None:
+    def create_replica(self, path: Path, md: list[TableVersionMd]) -> None:
```
aaron-siegel: I really don't like the name "TableVersionMd" for this. We already have a schema.TableVersionMd so this creates horrid namespace confusion. Now we have schema.TableVersionMd that's an element of table_version.TableVersionMd!

I'm fine consolidating these classes but let's please call it something else. cc @mkornacker

pixeltable/share/packager.py
```diff
@@ -81,7 +79,9 @@ def package(self) -> Path:
         _logger.info(f'Packaging table {self.table._path()!r} and its ancestors in: {self.tmp_dir}')
         self.tmp_dir.mkdir()
         with open(self.tmp_dir / 'metadata.json', 'w', encoding='utf8') as fp:
-            json.dump(self.md, fp)
+            json_md = self.bundle_md.copy()
+            json_md['md'] = [dataclasses.asdict(md) for md in self.bundle_md['md']]
```
aaron-siegel: Why not do this conversion up front in the class initializer?

pixeltable/share/packager.py
```diff
@@ -345,21 +346,22 @@ def restore(self, bundle_path: Path) -> pxt.Table:
         with tarfile.open(bundle_path, 'r:bz2') as tf:
             tf.extractall(path=self.tmp_dir)
 
-        if self.md is None:
+        if self.bundle_md is None:
             # No metadata supplied; read it from the archive
             with open(self.tmp_dir / 'metadata.json', 'r', encoding='utf8') as fp:
-                self.md = json.load(fp)
+                self.bundle_md = json.load(fp)
```
aaron-siegel, line 389: The semantics are a little confusing b/c the `bundle_md` in TablePackager contains dataclasses, but the `bundle_md` in TableRestorer is full JSON. If we do the JSON conversion in the class initializer of TablePackager as noted above, it will also resolve that discrepancy.

pixeltable/share/protocol/common.py
```diff
@@ -0,0 +1,87 @@
+import uuid
+from abc import ABC, abstractmethod
+from enum import Enum
+from typing import Any, Optional
+from urllib.parse import urlparse
+
+from pydantic import BaseModel
+
+
+class StorageDestination(str, Enum):
```
aaron-siegel, line 12: Can we use the existing StorageTarget enum for this instead of creating something new/duplicative?

amithadke: I'm just moving protocol from cloud to core :)  StorageTarget came after StorageDestination 
every change here we need to change cloud code to reflect it. Cloud side doesnt support all types in StorageTarget (specially wont support os)

pixeltable/catalog/catalog.py
```diff
@@ -1792,15 +1792,15 @@ def _collect_tbl_history(self, tbl_id: UUID, n: int | None) -> list[schema.FullT
             q = q.limit(n)
         src_rows = Env.get().session.execute(q).fetchall()
         return [
-            schema.FullTableMd(
-                schema.md_from_dict(schema.TableMd, row.Table.md),
-                schema.md_from_dict(schema.TableVersionMd, row.TableVersion.md),
-                schema.md_from_dict(schema.TableSchemaVersionMd, row.TableSchemaVersion.md),
+            TableVersionCompleteMd(
+                tbl_md=None,
```
aaron-siegel: Why is this changed to set to None?

pixeltable/catalog/table_version.py
```diff
@@ -43,17 +43,25 @@
...
-class TableVersionMd:
+class TableVersionCompleteMd:
     """
     Complete set of md records for a specific TableVersion instance.
-
-    TODO: subsume schema.FullTableMd
     """
 
-    tbl_md: schema.TableMd
+    tbl_md: schema.TableMd | None
```
aaron-siegel: This is a merge bug too. (It shouldn't be None)

PR #823, opened 2025-09-30
cockroachDB: div, mod operations SQL changed. Timestamp propagated through client stack

jpeterson-pxt:
Notes:

- Use of None as a timezone during setup is converted to the local timestamp of the python instance.
- Timestamp types are converted to the correct timezone when retrieved from the database to work around psycopg2 limitations
- Math operations for div, mod and integer div are modified to work with cockroachDB
- The round(x, numdigits) operation uses different SQL with cockroachDB

### DIFF
--- a/tests/functions/test_timestamp.py
+++ b/tests/functions/test_timestamp.py
@@ -133,8 +133,6 @@ class TestTimestamp:
             print(f'Requesting client in time zone: {tz_str}')
             default_time_zone = None if tz_str is None else ZoneInfo(tz_str)
             Env.get().default_time_zone = default_time_zone
-            # Be sure we are testing with the timezone of the local DB client
-            default_time_zone = Env.get().default_time_zone
             print(f'  (effective default time zone is: {default_time_zone})')
 
             pxt.drop_table('test_tbl', force=True)


pixeltable/functions/math.py
```diff
@@ -96,8 +97,18 @@ def round(self: float, digits: Optional[int] = None) -> float:
 def _(self: sql.ColumnElement, digits: Optional[sql.ColumnElement] = None) -> sql.ColumnElement:
     if digits is None:
         return sql.func.round(self)
+    elif Env.get().is_using_cockroachdb:
+        # CockroachDB cannot represent infinity as a NUMERIC, so special-case them
+        sqlinf = sql.cast(sql.literal('Infinity'), sql.Float)
+        return sql.case(
```
mkornacker: We already decided on Slack we weren't going to do this.

mkornacker (changes requested):

pixeltable/env.py
```diff
@@ -534,13 +553,16 @@ def _init_metadata(self) -> None:
         metadata.schema.base_metadata.create_all(self._sa_engine, checkfirst=True)
         metadata.create_system_info(self._sa_engine)
 
-    def _create_engine(self, time_zone_name: Optional[str], echo: bool = False) -> None:
+    def _create_engine(self, time_zone_name: str, echo: bool = False) -> None:
         connect_args = {} if time_zone_name is None else {'options': f'-c timezone={time_zone_name}'}
```
aaron-siegel: This can be simplified now that `time_zone_name` can never be `None`.

pixeltable/exec/sql_node.py
```diff
@@ -328,14 +331,25 @@ async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
...
                         output_row[slot_idx] = float(sql_row[i])
                     else:
                         raise RuntimeError(f'Unexpected Decimal value for {e}')
+                elif is_using_cockroachdb and isinstance(sql_row[i], datetime.datetime):
+                    # Ensure that the datetime is timezone-aware and in the session time zone
+                    # cockroachDB returns timestamps in the session time zone, with numeric offset,
+                    # convert to the session time zone with correct tzinfo for DST handling
+                    if e.col_type.is_timestamp_type():
+                        if isinstance(sql_row[i].tzinfo, datetime.timezone):
+                            output_row[slot_idx] = sql_row[i].replace(tzinfo=tzinfo)
```
aaron-siegel: should we assert that the offset is correct, just to be safe?

if we do a `replace` on an incompatible timezone, we'll end up returning the wrong answer, which would be pretty bad.

pixeltable/functions/math.py
aaron-siegel: This is a different scenario from the one we discussed on slack.

tests/functions/test_date.py
```diff
@@ -40,6 +40,7 @@ def test_date_methods(self, reset_db: None) -> None:
         # Set a default time zone that's likely to be different from the system time zone of most test environments
         default_tz = ZoneInfo('America/Anchorage')
         Env.get().default_time_zone = default_tz
+        default_tz = Env.get().default_time_zone
```
aaron-siegel: ? what's this doing?

jpeterson-pxt: In case we fail to set the default time zone to the requested time zone for some reason, it allows the test to proceed with the current default time zone.

pixeltable/exec/sql_node.py

pixeltable/env.py
jpeterson-pxt: Done

tests/functions/test_date.py
aaron-siegel: If we fail to set the default time zone, shouldn't we get an exception?

Either way, I'd argue this should be an assertion, not a fallback (i.e., we shouldn't silently fail the timezone operation and then let the test pass anyway)

jpeterson-pxt: Done

pixeltable/exec/sql_node.py
aaron-siegel: Just use `sql_row[i].timestamp()` instead of `astimezone(datetime.timezone.utc)`. No reason to create a bunch of new objects just for assertion purposes.

Actually, wouldn't it be safest just to do:
```
output_row[slot_idx] = sql_row[i].astimezone(tzinfo)
```
instead of using `replace()`? Then we wouldn't need the assertion at all since we're doing the conversion safely - right?

tests/functions/test_timestamp.py
```diff
@@ -26,12 +26,19 @@ def make_test_table(self) -> tuple[list[datetime], pxt.Table]:
...
+            x = irow.astimezone(timezone.utc)
+            y = rrow.astimezone(timezone.utc)
+            assert x == y, (irow, rrow)
         return test_dts, t
 
     def test_timestamp_methods(self, reset_db: None) -> None:
         # Set a default time zone that's likely to be different from the system time zone of most test environments
         default_tz = ZoneInfo('America/Anchorage')
         Env.get().default_time_zone = default_tz
+        default_tz = Env.get().default_time_zone
```
aaron-siegel: Per earlier conversation - this masks potential bugs. If it's important to have this check in the test, I think it should be an assertion rather than silently swallowing any inconsistency.

aaron-siegel (changes requested):

tests/functions/test_timestamp.py
jpeterson-pxt: Fixed.

pixeltable/functions/math.py
jpeterson-pxt: Done

pixeltable/exec/sql_node.py
jpeterson-pxt: Done.

tests/functions/test_timestamp.py
```diff
@@ -123,9 +130,12 @@ def test_timestamp_zones(self, reset_db: None) -> None:
         # values, to ensure that we test some values where it's different from
         # the system local time where the test is running.
         for tz_str in [None, 'America/Anchorage', 'America/New_York', 'Asia/Kolkata', 'Asia/Dubai']:
-            print(f'Testing with default time zone equal to: {tz_str}')
+            print(f'Requesting client in time zone: {tz_str}')
             default_time_zone = None if tz_str is None else ZoneInfo(tz_str)
             Env.get().default_time_zone = default_time_zone
+            # Be sure we are testing with the timezone of the local DB client
+            default_time_zone = Env.get().default_time_zone
```
aaron-siegel: I think I've made this comment before, but please do not silently swallow discrepancies here. Make this an assertion if that's what's intended.

jpeterson-pxt: Sorry, I thought I found all of them.

aaron-siegel (approved):

mkornacker (approved):

PR #825, opened 2025-10-01
Openrouter Integration

aaron-siegel:
Contains everything except the Working with Openrouter notebook.


pixeltable/functions/openrouter.py
```diff
@@ -0,0 +1,139 @@
...
+if TYPE_CHECKING:
+    import openai
+
+
+@register_client('openrouter')
+def _(api_key: str, site_url: Optional[str] = None, app_name: Optional[str] = None) -> 'openai.AsyncOpenAI':
+    import openai
+
+    # Create default headers for OpenRouter
+    default_headers = {}
```
mkornacker: type decl

pixeltable/config.py
```diff
@@ -181,6 +181,11 @@ def get_bool_value(self, key: str, section: str = 'pixeltable') -> Optional[bool
         'api_version': 'API version if using Azure OpenAI',
         'rate_limits': 'Per-model rate limits for OpenAI API requests',
     },
+    'openrouter': {
```
mkornacker, line 184: missing rate limiting-related config.

mkornacker (commented):
Please also extend tool/perftest_providers.py.

pixeltable/config.py
aaron-siegel: I haven't implemented rate-limiting for openrouter. I don't think this should be a prerequisite for shipping a new integration. It will take me more time to research and test how openrouter handles it and the original PR (from Cursor) didn't contain anything to go on.

The default rate-limiting will provide an ok fallback in the meantime and we can add a rough edges ticket for it if ppl start using openrouter and it becomes a priority.

Curious what @pierrebrunelle thinks, if this should be a blocker for the initial integration.

pixeltable/functions/openrouter.py
aaron-siegel: Done

aaron-siegel:
> Please also extend tool/perftest_providers.py.

Done. If this is a standing requirement for all new integrations, please update the integrations checklist accordingly. Thanks!

pixeltable/config.py
pierrebrunelle: Works for me. We can ship this and learn from users once they use it or migrate to it instead of custom UDFs.

aaron-siegel:
> Could you please also add a version of this to the docstring:
> 
> ```
> Request throttling:
> Applies the rate limit set in the config (section `fireworks`, key `rate_limit`). If no rate
> limit is configured, uses a default of 600 RPM.
> ```

Done. We should consider updating those docstrings with a link to the "Configuration" section of our docs. The rate limits can also be set by environment variable (FIREWORKS_RATE_LIMIT etc) and some users might prefer that.

PR #826, opened 2025-10-01
with_audio() udf

mkornacker:


pixeltable/functions/video.py
```diff
@@ -528,6 +528,123 @@ def concat_videos(videos: list[pxt.Video]) -> pxt.Video:
         filelist_path.unlink()
 
 
+@pxt.udf
+def add_audio(
```
aaron-siegel: This doesn't really add an audio track: if the video already contains audio, I'd expect something called `add_audio` to add a new audio track side-by-side with the existing one(s).

`replace_audio` would be clearer (and `replace_audio_track` even better, in my opinion).

pixeltable/functions/video.py
```diff
@@ -528,6 +528,123 @@ def concat_videos(videos: list[pxt.Video]) -> pxt.Video:
...
+    Add an audio track to a video file with specified offsets and durations.
+
+    __Requirements:__
+
+    - `ffmpeg` needs to be installed and in PATH
+
+    Args:
+        video: Input video.
+        audio: Input audio.
+        video_start_time: Start time in the video file (in seconds).
```
aaron-siegel: What does this do? Truncate the video, or keep the video as-is, inject the audio track at the specified location, and pad the initial segment with silence? It'd be good to clarify.

pixeltable/functions/video.py
```diff
@@ -528,6 +528,123 @@ def concat_videos(videos: list[pxt.Video]) -> pxt.Video:
...
+    video: pxt.Video,
+    audio: pxt.Audio,
+    *,
+    video_start_time: float = 0.0,
+    video_duration: float | None = None,
+    audio_start_time: float = 0.0,
+    audio_duration: float | None = None,
+) -> pxt.Video:
+    """
+    Add an audio track to a video file with specified offsets and durations.
```
aaron-siegel: Clarify here what happens if the audio and video are different in duration.

pixeltable/functions/video.py
```diff
@@ -528,6 +528,123 @@ def concat_videos(videos: list[pxt.Video]) -> pxt.Video:
...
+        ...     tbl.video.add_audio(
+        ...         tbl.music_track,
+        ...         video_start_time=30.0,
+        ...         video_duration=10.0,
+        ...         audio_start_time=15.0,
+        ...         audio_duration=10.0
+        ...     )
+        ... ).collect()
+    """
+    if not shutil.which('ffmpeg'):
```
aaron-siegel: This has appeared several times and could probably be factored out into a utility method (`Env.require_ffmpeg()` say, or `Env.require_binary()` analogous to `Env.require_package()`)

pixeltable/functions/video.py
mkornacker: Renamed to set_audio() (same as MoviePy).

pixeltable/functions/video.py

pixeltable/functions/video.py
mkornacker: Renamed it to with_audio() (in reference to MoviePy).

pixeltable/functions/video.py
mkornacker: Clarified

pixeltable/functions/video.py

PR #828, opened 2025-10-01
Randomize the behavior of `sample()` in the case `seed=None`

aaron-siegel:


pixeltable/plan.py
```diff
@@ -92,18 +93,17 @@ class SampleClause:
...
     # The version of the hashing algorithm used for ordering and fractional sampling.
     CURRENT_VERSION = 1
 
     def __post_init__(self) -> None:
-        """If no version was provided, provide the default version"""
+        # If no version was provided, provide the default version
         if self.version is None:
             self.version = self.CURRENT_VERSION
+        # If no seed was specified, pick a random one. (This will ensure that separate calls to sample() will be
+        # nondeterministic, but once a DataFrame is constructed it will be internally consistent.)
```
mkornacker: I think this is still incorrect behavior. If you run the same query multiple times, you should get different samples, unless you explicitly specify a seed. That's what Postgres does, as well as numpy and pandas.

aaron-siegel: Huh? That's exactly what this does. Not sure where the disconnect is

aaron-siegel: Per our various discussions, changed so that `df.collect()` run multiple times will return distinct results.

pixeltable/exec/sql_node.py
```diff
@@ -648,6 +650,11 @@ def __init__(
         )
         self.stratify_exprs = stratify_exprs
         self.sample_clause = sample_clause
+        if sample_clause.seed is None:
```
mkornacker: Let's put the seed into ExecContext (and initialize it there). We want to be able to cache plans.

aaron-siegel: Done.

pixeltable/exec/exec_context.py
```diff
@@ -1,13 +1,27 @@
...
+    row_builder: exprs.RowBuilder
+    profile: exprs.ExecProfile
+    show_pbar: bool
+    batch_size: int
+    num_rows: Optional[int]
+    conn: Optional[sql.engine.Connection]
+    pk_clause: Optional[list[sql.ClauseElement]]
+    num_computed_exprs: int
+    ignore_errors: bool
+    sample_clause: Optional['SampleClause']
```
mkornacker: I meant: put the actual seed into ExecContext (not SampleClause, which is a syntactic element), then refer to that in SampleClause._create_stmt() (which has access to the context).

It doesn't matter at the moment, because we re-generate the entire plan every time we run a query. But the goal for our serving setup is to cache plans and simply initialize new ExecContexts.

aaron-siegel: Ah, that context is helpful (it wasn't fully clear to me why you wanted to do it this way, but that makes sense).

It feels strange to me for `seed` to be divorced from a `SampleClause` though. It's not really a property of the execution context, it's a property of *that operation*. How would this work in a more general context where, say, your plan could involve more than one sample operation? (That also doesn't matter at the moment, but it's useful to think about; I'd argue it's the right mental model, and we should be choosing abstractions that make sense in general, even if our existing planner implementation only supports a single sample operation.)

PR #838, opened 2025-10-07
cockroachDB: Set null_ordered_last on session start.

jpeterson-pxt:
The setting "null_ordered_last" is set on cockroach session start.

### DIFF
--- a/pixeltable/env.py
+++ b/pixeltable/env.py
@@ -577,8 +577,7 @@ class Env:
             self._logger.info(f'Database time zone is now: {tz_name}')
             self._default_time_zone = ZoneInfo(tz_name)
             if self.is_using_cockroachdb:
-                # This should have been set when the database was created,
-                # but if the db predates the fix or we are running as root, set it now
+                # This could be set when the database is created, but we set it now
                 conn.execute(sql.text('SET null_ordered_last = true;'))
                 null_ordered_last = conn.execute(sql.text('SHOW null_ordered_last')).scalar()
                 assert isinstance(null_ordered_last, str)


pixeltable/env.py
```diff
@@ -576,6 +576,13 @@ def _create_engine(self, time_zone_name: str, echo: bool = False) -> None:
             assert isinstance(tz_name, str)
             self._logger.info(f'Database time zone is now: {tz_name}')
             self._default_time_zone = ZoneInfo(tz_name)
+            if self.is_using_cockroachdb:
+                # This should have been set when the database was created,
+                # but if the db predates the fix or we are running as root, set it now
```
aaron-siegel: Better to do this as a migration? Or just manually issue the ALTER command in our existing DBs? (if that's possible?) Then this could become just an assertion.

pixeltable/env.py
```diff
@@ -599,6 +606,8 @@ def _create_store_db(self) -> None:
             with engine.begin() as conn:
                 stmt = self._dbms.create_db_stmt(preparer.quote(self._db_name))
                 conn.execute(sql.text(stmt))
+                if self.is_using_cockroachdb:
+                    conn.execute(sql.text('ALTER ROLE ALL IN DATABASE mydb SET null_ordered_last = true;'))
```
jpeterson-pxt: mydb -> s/b self._db_name...
Nuke this.

pixeltable/env.py
jpeterson-pxt: Leaving this as we discussed.

aaron-siegel (approved):

PR #840, opened 2025-10-07
feat(storage): add Backblaze B2 S3-compatible integration and tests

jeronimodeleon:
### Summary
Adds support for Backblaze B2 as an S3-compatible storage backend.

### Details
- Introduced `StorageTarget.B2_STORE` for S3-compatible B2 endpoints.
- Updated `S3Store` to handle B2 profiles and endpoints.
- Added user-agent and config updates for B2.
- Extended integration tests (`test_destination.py`) to cover B2 upload, copy, and cleanup flows.
- Verified using a live B2 bucket (`s3.us-east-005.backblazeb2.com/pxt-test`).

### Testing
- All B2 tests pass locally (`pytest -v tests/test_destination.py -k "b2"`).
- CI-safe skips for other clouds (R2, GCS, Azure) remain unchanged.

### Notes
This PR introduces no breaking changes and aligns with existing S3-compatible integration patterns (R2).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,23 @@ target
 tests/target
 .DS_Store
 **/.DS_Store
+
+# Virtual environments
+venv/
+env/
+ENV/
+
+# Python
+*.pyc
+*.pyo
+*.egg-info/
+.pytest_cache/
+.tox/
+
+# Compiled files
+*.so
+*.dylib
+
+# Build artifacts
+build/
+*.egg
--- a/pixeltable/utils/s3_store.py
+++ b/pixeltable/utils/s3_store.py
@@ -44,6 +44,18 @@ def _() -> Any:
     return R2ClientDict(profile=profile_name, clients={})
 
 
+@env.register_client('b2')
+def _() -> Any:
+    profile_name = Config.get().get_string_value('b2_profile')
+    return R2ClientDict(profile=profile_name, clients={})
+
+
+@env.register_client('b2_resource')
+def _() -> Any:
+    profile_name = Config.get().get_string_value('b2_profile')
+    return R2ClientDict(profile=profile_name, clients={})
+
+
 @env.register_client('s3')
 def _() -> Any:
     profile_name = Config.get().get_string_value('s3_profile')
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -15,3 +15,23 @@ target
...
+
+# Python
+*.pyc
+*.pyo
+*.egg-info/
+.pytest_cache/
+.tox/
+
+# Compiled files
+*.so
```
aaron-siegel: Are these `.gitignore` changes due to files that showed up in your source tree during development?

pixeltable/utils/s3_store.py
```diff
@@ -329,6 +360,8 @@ def create_boto_client(cls, profile_name: Optional[str] = None, extra_args: Opti
             'connect_timeout': 15,
             'read_timeout': 30,
             'retries': {'max_attempts': 3, 'mode': 'adaptive'},
+            's3': {'addressing_style': 'path'},  # Use path-style addressing for S3-compatible services
+            'user_agent_extra': 'pixeltable',
```
aaron-siegel: I'd welcome a code comment explaining the effect of this. Thanks!

jeronimodeleon: Set addressing_style='path' for compatibility with S3-compatible endpoints like Backblaze B2 and Cloudflare R2.
Added user_agent_extra='pixeltable' to include a consistent Pixeltable identifier in request headers, this is useful for logs, traceability, and debugging.

.gitignore
jeronimodeleon: Yes, I updated the .gitignore to filter out a few development files that were showing up locally. I might have reused part of another project’s .gitignore, so feel free to clean up or revert anything unnecessary.

aaron-siegel: I wouldn't know how to clean it up since I don't know which specific file patterns are affecting you. I'd prefer the extraneous ones not get checked in in the first place... any chance you could trim out the ones that are not relevant? (Or just remove the .gitignore changes from the PR if that's easiest?)

aaron-siegel:
Also, we'll need to set up a backblaze account that we can use for CI, can you point us to instructions on the best way to do this? Thanks!

.gitignore
jeronimodeleon: Sounds good, I’ve removed the .gitignore changes from the PR so it only includes the relevant updates.

aaron-siegel:
Note there is a formatting issue; you can run `make format` to fix it. I'll merge once the tests pass

jeronimodeleon:
> Also, we'll need to set up a backblaze account that we can use for CI, can you point us to instructions on the best way to do this? Thanks!

Here’s the quick setup:
Go to [Backblaze Cloud Storage](https://www.backblaze.com/cloud-storage)
 and sign up.
In the left menu, open B2 Cloud Storage → Application Keys.
Click Add a New Application Key and copy the Key ID and Application Key.
Add these as CI secrets:
B2_APPLICATION_KEY_ID=<your-key-id>
B2_APPLICATION_KEY=<your-key>

PR #841, opened 2025-10-07
Support packaging of tables with embedding indices for data sharing

aaron-siegel:
Implements package/restore support for tables with embedding indices:
- Rectifies embedding indices on restore by shuffling the val/undo columns as appropriate
- Allows TableVersion to initialize indices for the head version of a replica, while still forbidding indices for other snapshots (including non-head replica versions)
- Fixes several instances where the SQLAlchemy schema of a TableVersion's BaseTable was misaligned with the underlying physical schema
- Updates required version of pgvector and dev version of pyarrow, and makes various related typing fixes


pixeltable/catalog/table_version.py
```diff
@@ -201,6 +201,9 @@ def versioned_name(self) -> str:
         else:
             return f'{self.name}:{self.effective_version}'
 
+    def __repr__(self) -> str:
+        return f'TableVersion(id={self.id!r}, version={self.version}, name={self.name!r})'
```
mkornacker: Also add effective_version, we can actually have both effective version for the same underlying version instantiated.

pixeltable/catalog/table_version.py
```diff
@@ -409,15 +412,40 @@ def init(self) -> None:
...
+        self._init_idxs()
+
         # create the sa schema only after creating the columns and indices
         self._init_sa_schema()
 
         # created value_exprs after everything else has been initialized
         for col in self.cols_by_id.values():
             col.init_value_expr()
 
+    def _has_idxs(self) -> bool:
```
mkornacker: Pass that into the c'tor and have Catalog make that determination. TableVersion shouldn't reach back into the history. Catalog understands Tables.

pixeltable/catalog/table_version.py
```diff
@@ -448,39 +476,70 @@ def _init_cols(self) -> None:
             #     self._record_refd_columns(col)
 
     def _init_idxs(self) -> None:
-        # self.idx_md = tbl_md.index_md
-        self.idxs_by_name = {}
-        import pixeltable.index as index_module
+        self.idxs_by_name: dict[str, TableVersion.IndexInfo] = {}
```
mkornacker: Duplicate of line 100 and 183.

pixeltable/catalog/table_version.py
```diff
@@ -448,39 +476,70 @@ def _init_cols(self) -> None:
...
+        has_idxs = self._has_idxs()
```
mkornacker: Misleading because it may not have any indices. supports_idxs? is_indexable?

pixeltable/catalog/table_version.py
```diff
@@ -448,39 +476,70 @@ def _init_cols(self) -> None:
...
+                # Sanity-check this.
+                assert md.indexed_col_id in self.cols_by_id
+                assert md.index_val_col_id in self.cols_by_id
+                assert md.index_val_undo_col_id in self.cols_by_id
+                idx_info = self.IndexInfo(
+                    id=md.id, name=md.name, idx=idx, col=idx_col, val_col=val_col, undo_col=undo_col
+                )
+                self.idxs_by_name[md.name] = idx_info
+
+    def _lookup_column(self, tbl_id: UUID, col_id: int) -> Column | None:
```
mkornacker: We already have a QColumnId.

pixeltable/catalog/table_version.py
```diff
@@ -448,39 +476,70 @@ def _init_cols(self) -> None:
...
-                idx_col = self.path.base.get_column_by_id(UUID(md.indexed_col_tbl_id), md.indexed_col_id)
+            cls = getattr(index, cls_name)
+            idx_col = self._lookup_column(UUID(md.indexed_col_tbl_id), md.indexed_col_id)
+            assert idx_col is not None
             idx = cls.from_dict(idx_col, md.init_args)
 
             # fix up the sa column type of the index value and undo columns
-            val_col = self.cols_by_id[md.index_val_col_id]
+            # we need to do this for all indices, not just those that are active in this TableVersion, to ensure we get
+            # the correct SA schema in the StoreTable.
```
mkornacker, line 469: If the intention is to have store_tbl.sa_tbl reflect the actual physical table, doing this in StoreBase.create_sa_tbl() is easier and less convoluted. You simply need to reference the val/undo_col_ids in tbl_version.tbl_md.index_md.

Index handling/initialization is too convoluted and needs a more holistic restructuring. I think the easiest short-term fix is
- create the correct sa_tbl in create_sa_tbl()
- leave out init_idxs() if Catalog decides this TV instance doesn't support indices

pixeltable/share/packager.py
```diff
@@ -579,6 +592,66 @@ def __import_table(self, bundle_path: Path, tv: catalog.TableVersion, tbl_md: sc
         result = conn.execute(q)
         _logger.debug(f'Inserted {result.rowcount} row(s) from {temp_sa_tbl_name!r} into {store_sa_tbl_name!r}.')
 
+        # STEP 5: Rectify any index columns.
+
+        # Finally, rectify any index columns in the table. This involves shuffling data between the index's val and
+        # undo columns to ensure they appropriately reflect the most recent replicated version of the table.
```
mkornacker, line 598: Keep in mind that replicas don't support revert(). All undo cols should always be null.

pixeltable/catalog/table_version.py
aaron-siegel: Done.

pixeltable/catalog/table_version.py
aaron-siegel: Fixed.

pixeltable/catalog/table_version.py
aaron-siegel: I'm not sure what the ask is here - are you suggesting replacing the parameters with a single QColumnId parameter? (Fine with me and I'm happy to do that, just not clear if that's what your asking)

pixeltable/share/packager.py
aaron-siegel: Hmmm, not sure I agree with this.

I mean, obviously, replicas don't support revert(). But I can think of several scenarios where we might want to make use of the data in undo_col:

- If we add support for dropping specific replica versions. For example, I've replicated versions 202, 270, 328, and 406 of a table. Then I decide I only want to keep version 328. It might be nice to be able to reconstruct a valid index at that point. (Same principle if 406 is a version fragment and I drop the view that depends on it.)

- If we want to add support at some point for reconstructing indices for older versions of tables. Currently there's no way to do this, but I can see value in it. (Maybe we don't maintain active indices for *all* past versions, but we allow users to declare, for example, that they want a usable index for version 270.)

If we add these features in the future, they'll rely on having preserved data in undo_col. So I'd be most comfortable keeping this data around even if there's no immediate visible difference.

pixeltable/catalog/table_version.py
aaron-siegel: Done.

pixeltable/catalog/table_version.py
aaron-siegel: Changed to supports_idxs.

pixeltable/catalog/table_version.py
aaron-siegel: This won't work because the `Column` instances will then have incorrect `sa_col_type`s.

We also need to do in-place modification of `stores_cellmd` for all indices, even if `supports_idxs` is False, to be sure we generate the correct schemas.

I agree that index initialization is dangerously convoluted, and I'd suggest we leave this as-is and refactor separately when we clean up index initialization more generally.

pixeltable/catalog/catalog.py
```diff
@@ -2024,12 +2048,13 @@ def _load_tbl_version(
             mutable_view_ids = [r[0] for r in conn.execute(q).all()]
 
         mutable_views = [TableVersionHandle(id, None) for id in mutable_view_ids]
+        supports_idxs = self.tv_supports_idxs(tbl_md, effective_version)
```
mkornacker: Can't this be directly derived from tbl_md and version_md?

Eg, effective_version is None or effective_version == tbl_md.current_version?

pixeltable/catalog/table_version.py
mkornacker: Yes. Sorry.

mkornacker (approved):

pixeltable/catalog/catalog.py

PR #842, opened 2025-10-08
Help users fix the `publish` error message

apreshill:
### The Problem
When users try to publish a table but a local replica already exists, they get a confusing error message that tells them what's wrong but doesn't tell them exactly how to fix it:

**Before:**
```
❌ Error publishing datasets: Error finalizing snapshot: Internal Server Error : 
That table has already been replicated as 'vision_waddle'.
Drop the existing replica if you wish to re-create it.
```

Users may not know:
- What "replica" means in this context
- How to drop a table
- What command to run after that

### The Solution
The new error message gives users exactly what they need: copy-pasteable commands with their actual table names and URIs filled in.

**After:**
```
❌ Cannot publish to 'pxt://acme_corp/datasets/sales_data': A local replica 'vision_waddle' already exists.

💡 To resolve this:
    pxt.drop_table('vision_waddle')
    pxt.publish('vision_waddle', 'pxt://acme_corp/datasets/sales_data')
```

Users can now:
- Understand the problem immediately
- Copy both commands with one selection
- Paste and run without having to figure out the syntax

### What Changed
- Updated error handling in `pixeltable/share/publish.py` to parse the server error and reconstruct the exact commands users need
- Replaced terms with plain language
- Added a 💡 emoji to draw attention to the solution

### DIFF
--- a/pixeltable/share/publish.py
+++ b/pixeltable/share/publish.py
@@ -68,7 +68,22 @@ def push_replica(
     # TODO: Use Pydantic for validation
     finalize_response = requests.post(PIXELTABLE_API_URL, json=finalize_request_json, headers=_api_headers())
     if finalize_response.status_code != 200:
-        raise excs.Error(f'Error finalizing snapshot: {finalize_response.text}')
+        # Check if this is a "table already replicated" error and provide actionable guidance
+        error_text = finalize_response.text
+        if 'already been replicated as' in error_text:
+            # Extract the existing table name from the error message
+            # Format: "That table has already been replicated as 'table_name'."
+            import re
+            match = re.search(r"already been replicated as ['\"]([^'\"]+)['\"]", error_text)
+            if match:
+                existing_table = match.group(1)
+                raise excs.Error(
+                    f"Cannot publish to '{dest_tbl_uri}': A local replica '{existing_table}' already exists.\n\n"
+                    f"💡 To resolve this:\n"
+                    f"    pxt.drop_table('{existing_table}')\n"
+                    f"    pxt.publish('{src_tbl._path()}', '{dest_tbl_uri}')"
+                )
+        raise excs.Error(f'Error finalizing snapshot: {error_text}')
     finalize_response_json = finalize_response.json()
     if not isinstance(finalize_response_json, dict) or 'confirmed_table_uri' not in finalize_response_json:
         raise excs.Error(f'Error finalizing snapshot: unexpected response from server.\n{finalize_response_json}')


pixeltable/share/publish.py
```diff
@@ -68,7 +68,22 @@ def push_replica(
     # TODO: Use Pydantic for validation
     finalize_response = requests.post(PIXELTABLE_API_URL, json=finalize_request_json, headers=_api_headers())
     if finalize_response.status_code != 200:
-        raise excs.Error(f'Error finalizing snapshot: {finalize_response.text}')
+        # Check if this is a "table already replicated" error and provide actionable guidance
+        error_text = finalize_response.text
+        if 'already been replicated as' in error_text:
+            # Extract the existing table name from the error message
+            # Format: "That table has already been replicated as 'table_name'."
+            import re
```
aaron-siegel: Imports should go at top of file unless they need to be in narrower scope.

pixeltable/share/publish.py
```diff
@@ -68,7 +68,22 @@ def push_replica(
...
+            match = re.search(r"already been replicated as ['\"]([^'\"]+)['\"]", error_text)
```
aaron-siegel, line 86: This is a really fragile way to do error handling. Innocent changes to the message format in the server code will now break the client logic.

Let's discuss this at standup if it's a priority - a better solution would be to use error codes in the client/server protocol and send the info in a structured way for the client to format it; or simply generate the desired message format on the server side.

pixeltable/share/publish.py
```diff
@@ -68,7 +68,22 @@ def push_replica(
...
+            if match:
+                existing_table = match.group(1)
+                raise excs.Error(
```
aaron-siegel, line 89: Since error message content is an important product requirement, we should add a unit test for it.

aaron-siegel (changes requested):

pixeltable/share/publish.py
apreshill: Fixed

pixeltable/share/publish.py
apreshill: Got it- thanks!

pixeltable/share/publish.py
apreshill: Attempted (blame Claude)

aaron-siegel: We use live testing throughout our unit tests, not mocks. I don't see a compelling reason to deviate from that here, especially since we're testing the behavior of our own services.

In general I'm not thrilled with directly reviewing LLM generated code. It tends to contain a lot of slop and the effort of reviewing exceeds the effort needed for an engineer to just build it. We may be better off with this as a rough edges ticket, but let's discuss in standup.

apreshill: I agree

apreshill:
Closing in favor of systematic approach to error message formatting

PR #854, opened 2025-10-12
Multi-phase create_table()

mkornacker:
- changes Column to have a TableVersionHandle
- ColumnRefs can now be constructed without a TableVersion instance
- create_table() now constructs default indices as part of the initial schema
- restructures IndexBase to avoid circular initialization dependencies between IndexBase and Column instances
- streamlines handling of indices in TableVersion
- removes index lookup logic elsewhere
- exprs now resolve indices at query execution time (to account for concurrent drop_index())

### DIFF
--- a/pixeltable/store.py
+++ b/pixeltable/store.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import abc
 import logging
 import sys
+import time
 import warnings
 from typing import Any, Iterable, Iterator, Optional
 
@@ -141,61 +142,83 @@ class StoreBase:
         assert isinstance(result, int)
         return result
 
-    def _exec_if_not_exists(self, create_stmt: str) -> None:
+    def _exec_if_not_exists(self, stmt: str, wait_for_table: bool) -> None:
         """
-        Execute a statement containing 'IF NOT EXISTS' and ignore any related errors.
+        Execute a statement containing 'IF NOT EXISTS' and ignore any duplicate object-related errors.
+
+        The statement needs to run in a separate transaction, because the expected error conditions will abort the
+        enclosing transaction (and the ability to run additional statements in that same transaction).
         """
-        conn = Env.get().conn
-        # Postgres seems not to handle concurrent Create ... If Not Exists correctly, we need to ignore the various
-        # errors that can occur when two connections run the same Create Table statement.
-        try:
-            conn.execute(sql.text(create_stmt))
-        except (sql.exc.IntegrityError, sql.exc.ProgrammingError) as e:
-            Env.get().console_logger.info(f'StoreBase.create() failed with: {e}')
-            if (
-                isinstance(e.orig, psycopg.errors.UniqueViolation)
-                and 'duplicate key value violates unique constraint' in str(e.orig)
-            ) or (
-                isinstance(e.orig, (psycopg.errors.DuplicateObject, psycopg.errors.DuplicateTable))
-                and 'already exists' in str(e.orig)
-            ):
-                pass
-            else:
-                raise
+        while True:
+            with Env.get().begin_xact(for_write=True) as conn:
+                try:
+                    if wait_for_table:
+                        # Try to lock the table to make sure that it exists. This needs to run in the same transaction
+                        # as 'stmt' to avoid a race condition.
+                        # TODO: adapt this for CockroachDB
+                        lock_stmt = f'LOCK TABLE {self._storage_name()} IN ACCESS EXCLUSIVE MODE'
+                        conn.execute(sql.text(lock_stmt))
+                    conn.execute(sql.text(stmt))
+                    return
+                except (sql.exc.IntegrityError, sql.exc.ProgrammingError) as e:
+                    Env.get().console_logger.info(f'{stmt} failed with: {e}')
+                    if (
+                        isinstance(e.orig, psycopg.errors.UniqueViolation)
+                        and 'duplicate key value violates unique constraint' in str(e.orig)
+                    ) or (
+                        isinstance(e.orig, (psycopg.errors.DuplicateObject, psycopg.errors.DuplicateTable))
+                        and 'already exists' in str(e.orig)
+                    ):
+                        # table already exists
+                        return
+                    elif isinstance(e.orig, psycopg.errors.UndefinedTable):
+                        # the Lock Table failed because the table doesn't exist yet; try again
+                        time.sleep(1)
+                        continue
+                    else:
+                        raise
+
+    def _store_tbl_exists(self) -> bool:
+        """Returns True if the store table exists, False otherwise."""
+        with Env.get().begin_xact(for_write=False) as conn:
+            q = (
+                'SELECT COUNT(*) FROM pg_catalog.pg_tables '
+                f"WHERE schemaname = 'public' AND tablename = {self._storage_name()!r}"
+            )
+            res = conn.execute(sql.text(q)).scalar_one()
+            return res == 1
 
     def create(self) -> None:
         """
-        Create or update store table to bring it in sync with self.sa_tbl.
+        Create or update store table to bring it in sync with self.sa_tbl. Idempotent.
 
-        If this is run outside of a transaction, runs each DDL statement in its own transaction.
+        This runs a sequence of DDL statements (Create Table, Alter Table Add Column, Create Index), each of which
+        is run in its own transaction.
+
+        The exception to that are local replicas, for which TableRestorer creates an enclosing transaction. In theory,
+        this should avoid the potential for race conditions that motivate the error handling present in
+        _exec_if_not_exists() (meaning: we shouldn't see those errors when creating local replicas).
+        TODO: remove the special case for local replicas in order to make the logic easier to reason about.
         """
         postgres_dialect = sql.dialects.postgresql.dialect()
-        stmt = sql.schema.CreateTable(self.sa_tbl, if_not_exists=True).compile(dialect=postgres_dialect)
-        create_stmt = str(stmt)
-        # 'lock table': force an update of the metadata cache in order to avoid
-        # psycopg.errors.UndefinedTable) relation "..." does not exist
-        # errors
-        # TODO: This is Postgres-specific. Adapt for Cockroachdb.
-        lock_stmt = f'LOCK TABLE {self._storage_name()} IN ACCESS EXCLUSIVE MODE'
-        stmts = [create_stmt, lock_stmt]
 
-        for col in self.sa_tbl.columns:
-            stmts.append(self._add_column_stmt(col))
+        if not self._store_tbl_exists():
+            # run Create Table If Not Exists; we always need If Not Exists to avoid race conditions between concurrent
+            # Pixeltable processes
+            create_stmt = sql.schema.CreateTable(self.sa_tbl, if_not_exists=True).compile(dialect=postgres_dialect)
+            self._exec_if_not_exists(str(create_stmt), wait_for_table=False)
+        else:
+            # ensure that all columns exist by running Alter Table Add Column If Not Exists for all columns
+            for col in self.sa_tbl.columns:
+                stmt = self._add_column_stmt(col)
+                self._exec_if_not_exists(stmt, wait_for_table=True)
+            # TODO: do we also need to ensure that these columns are now visible (ie, is there another potential race
+            # condition here?)
 
+        # ensure that all visible indices exist by running Create Index If Not Exists
         for index in self.sa_tbl.indexes:
-            stmt = sql.schema.CreateIndex(index, if_not_exists=True).compile(dialect=postgres_dialect)
-            create_stmt = str(stmt)
-            stmts.append(create_stmt)
-
-        env = Env.get()
-        for stmt in stmts:
-            if env.in_xact:
-                env.conn.execute(sql.text(stmt))
-            else:
-                # each of these statements need to be run as separate transactions; if one fails, we still want to run
-                # the rest
-                with env.begin_xact(for_write=True):
-                    self._exec_if_not_exists(stmt)
+            create_stmt = sql.schema.CreateIndex(index, if_not_exists=True).compile(dialect=postgres_dialect)
+            self._exec_if_not_exists(str(create_stmt), wait_for_table=True)
 
     def create_index(self, idx_id: int) -> None:
         """Create If Not Exists for this index"""
@@ -204,7 +227,7 @@ class StoreBase:
         conn = Env.get().conn
         stmt = sql.schema.CreateIndex(sa_idx, if_not_exists=True).compile(conn)
         create_stmt = str(stmt)
-        self._exec_if_not_exists(create_stmt)
+        self._exec_if_not_exists(create_stmt, wait_for_table=True)
 
     def validate(self) -> None:
         """Validate store table against self.table_version"""
--- a/pixeltable/env.py
+++ b/pixeltable/env.py
@@ -21,7 +21,7 @@ from contextlib import contextmanager
 from dataclasses import dataclass, field
 from pathlib import Path
 from sys import stdout
-from typing import TYPE_CHECKING, Any, Callable, Iterator, Literal, Optional, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Iterator, Optional, TypeVar
 from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 import nest_asyncio  # type: ignore[import-untyped]
@@ -55,6 +55,8 @@ class Env:
     For a local environment, Pixeltable uses an embedded PostgreSQL server that runs locally in a separate process.
     For a non-local environment, Pixeltable uses a connection string to the externally managed database.
     """
+
+    SERIALIZABLE_ISOLATION_LEVEL = 'SERIALIZABLE'
 
     _instance: Optional[Env] = None
     __initializing: bool = False
@@ -94,7 +96,7 @@ class Env:
     _resource_pool_info: dict[str, Any]
     _current_conn: Optional[sql.Connection]
     _current_session: Optional[orm.Session]
-    _current_isolation_level: Optional[Literal['REPEATABLE_READ', 'SERIALIZABLE']]
+    _current_isolation_level: str | None
     _dbms: Optional[Dbms]
     _event_loop: Optional[asyncio.AbstractEventLoop]  # event loop for ExecNode
 
@@ -274,7 +276,7 @@ class Env:
         if self._current_conn is None:
             assert self._current_session is None
             try:
-                self._current_isolation_level = 'SERIALIZABLE'
+                self._current_isolation_level = self.SERIALIZABLE_ISOLATION_LEVEL
                 with (
                     self.engine.connect().execution_options(isolation_level=self._current_isolation_level) as conn,
                     orm.Session(conn) as session,
@@ -289,7 +291,7 @@ class Env:
                 self._current_isolation_level = None
         else:
             assert self._current_session is not None
-            assert for_write == (self._current_isolation_level == 'serializable')
+            assert self._current_isolation_level == self.SERIALIZABLE_ISOLATION_LEVEL or not for_write
             yield self._current_conn
 
     def configure_logging(


pixeltable/catalog/column.py
```diff
@@ -200,17 +203,21 @@ def init_value_expr(self) -> None:
             )
             warnings.warn(message, category=excs.PixeltableWarning, stacklevel=2)
 
+    def tbl(self) -> TableVersion:
```
aaron-siegel: Call this `get_tbl()` now?

pixeltable/catalog/table_version.py
```diff
@@ -215,41 +215,87 @@ def create_initial_md(
...
+        tbl_id_str = str(tbl_id)
+        tbl_handle = TableVersionHandle(tbl_id, None)
+
+        def id_gen() -> Iterator[int]:
+            id_ = 0
+            while True:
+                yield id_
+                id_ += 1
+
+        column_ids = id_gen()
```
aaron-siegel: `column_ids = itertools.count()` should do the same thing. (We shouldn't need the bespoke `id_gen()`)

pixeltable/catalog/column.py
```diff
@@ -115,21 +115,30 @@ def __init__(
         # self.dependent_cols = set()  # cols with value_exprs that reference us; set by TableVersion
         self.id = col_id
         self.is_pk = is_pk
+        self.is_iterator_col = is_iterator_col
         self._media_validation = media_validation
         self.schema_version_add = schema_version_add
         self.schema_version_drop = schema_version_drop
 
-        self._stores_cellmd = stores_cellmd
+        if stores_cellmd is not None:
```
aaron-siegel, line 123: Is there any way to get rid of the `stores_cellmd` override and make it a deterministic property of the other column attributes? (I find these kinds of overrides make the code more convoluted and invite bugs. That is to say, it's clearest if this is either a pure input parameter or a pure derived parameter, rather than a hybrid of both.)

pixeltable/catalog/table_version.py
```diff
@@ -28,11 +29,17 @@
 from .tbl_ops import TableOp
 from .update_status import RowCountStats, UpdateStatus
 
+# from .catalog import Catalog
```
aaron-siegel: ?

pixeltable/catalog/table_version.py
```diff
@@ -6,16 +6,17 @@
...
 from uuid import UUID
 
 import jsonschema.exceptions
 import sqlalchemy as sql
 from sqlalchemy import exc as sql_exc
 
-import pixeltable as pxt
 import pixeltable.exceptions as excs
-from pixeltable import exprs, index
+import pixeltable.exprs as exprs
```
aaron-siegel, line 18: Better to condense these four lines into:
```
from pixeltable import exceptions as excs, exprs, index, type_system as ts
```
That would mirror the pattern used throughout the codebase

pixeltable/catalog/table_version.py
```diff
@@ -149,6 +160,8 @@ def __init__(
         else:
             self_handle = TableVersionHandle(id, self.effective_version)
             if self.is_view:
+                if base_path is None:
```
aaron-siegel: What is this code doing?

pixeltable/catalog/table_version.py
```diff
@@ -180,7 +193,15 @@ def __init__(
         self.cols = []
         self.cols_by_name = {}
         self.cols_by_id = {}
+        self.idxs = {}
         self.idxs_by_name = {}
+        self.idxs_by_col = {}
+        # TODO: THIS IS A BUG: for a replica of a snapshot, effective_version == current_version, but if there were
```
aaron-siegel: Per my in person comments today, I don't believe this is a bug, since a consistent state (in the catalog of the replica) will be reconstructed during the index rectification step in TableRestorer.

pixeltable/catalog/table_version.py
```diff
@@ -219,41 +240,87 @@ def create_initial_md(
...
             assert sch_md is not None
             column_md[col.id] = col_md
             schema_col_md[col.id] = sch_md
 
-        tbl_id = uuid.uuid4()
-        tbl_id_str = str(tbl_id)
+        index_md: dict[int, schema.IndexMd] = {}
+        if create_default_idxs and (view_md is None or not view_md.is_snapshot):
+            index_cols: list[Column] = []
+            for col in (c for c in cols if cls._is_btree_indexable(c)):
```
aaron-siegel, line 272: Unusual pattern; why not just
```
for col in cols:
    if cls._is_btree_indexable(col):
```
which is more standard and IMO easier to follow?

pixeltable/catalog/table_version.py
```diff
@@ -414,97 +446,117 @@ def init(self) -> None:
...
-                assert md.indexed_col_id in self.cols_by_id
-                assert md.index_val_col_id in self.cols_by_id
-                assert md.index_val_undo_col_id in self.cols_by_id
-                idx_info = self.IndexInfo(
-                    id=md.id, name=md.name, idx=idx, col=idx_col, val_col=val_col, undo_col=undo_col
+                # for md in (md for md in self.tbl_md.index_md.values() if md.schema_version_drop is None):
+                idx = idxs[md.id]
+                indexed_col_id = QColumnId(UUID(md.indexed_col_tbl_id), md.indexed_col_id)
+                idx_col = self._lookup_column(indexed_col_id)
+                if md.index_val_col_id not in self.cols_by_id:
```
aaron-siegel: What is this code doing?

pixeltable/catalog/table_version.py
```diff
@@ -1502,6 +1571,7 @@ def next_row_id(self, id: int) -> None:
     @property
     def is_snapshot(self) -> bool:
         return self.effective_version is not None
+        # return self.view_md is not None and self.view_md.is_snapshot
```
aaron-siegel: ?

pixeltable/exprs/comparison.py
```diff
@@ -70,7 +70,7 @@ def _op2(self) -> Expr:
         return self.components[1]
 
     def sql_expr(self, sql_elements: SqlElementCache) -> Optional[sql.ColumnElement]:
-        from pixeltable import index
+        import pixeltable.index as index
```
aaron-siegel, line 73: Why change this? `from ... import ...` is standard and used throughout the codebase.

Usually one only uses `import ... as ...` when renaming the module/class.

pixeltable/exprs/expr.py
```diff
@@ -495,6 +495,18 @@ def eval(self, data_row: DataRow, row_builder: 'exprs.RowBuilder') -> None:
...
 
+    def prepare(self) -> None:
+        """
+        Create execution state. This is called before the first eval() call.
+        """
+        for c in self.components:
+            c.prepare()
+
+    @classmethod
+    def prepare_list(cls, expr_list: list[Expr]) -> None:
```
aaron-siegel: This seems a little silly to me to have as a separate method, but if we're going to have it, it should take an Iterable.

pixeltable/catalog/table_version.py
```diff
@@ -414,97 +446,117 @@ def init(self) -> None:
...
+            if col_md.schema_version_add <= self.schema_version and (
+                col_md.schema_version_drop is None or col_md.schema_version_drop > self.schema_version
+            ):
+                if col.name is not None:
+                    self.cols_by_name[col.name] = col
+                self.cols_by_id[col.id] = col
+
+        if self.supports_idxs:
+            # create IndexInfo
+            # TODO: this logic is incorrect, because indices are not versioned; only indices that exist at the time
```
aaron-siegel: cf. above comment, index rectification in TableRestorer should ensure that everything is valid after a restore operation.

If you've identified a specific instance in which a package/restore round-trip yields inconsistent results then please flag it, but otherwise let's remove these TODOs.

pixeltable/catalog/table_version.py
```diff
@@ -574,45 +627,50 @@ def _add_default_index(self, col: Column) -> Optional[UpdateStatus]:
...
-    def _create_index_columns(self, idx: index.IndexBase) -> Tuple[Column, Column]:
+    @classmethod
+    def _create_index_columns(
+        cls,
+        col: Column,
+        idx: index.IndexBase,
+        schema_version: int,
+        tbl_handle: TableVersionHandle,
+        id_cb: Callable[[], int],
+    ) -> Tuple[Column, Column]:
```
aaron-siegel: Use lowercase `tuple` (uppercase form is Python 3.8 syntax)

pixeltable/exprs/column_ref.py
```diff
@@ -1,13 +1,13 @@
...
-import copy
-from typing import Any, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Optional, Sequence, cast
 from uuid import UUID
 
 import sqlalchemy as sql
 
-import pixeltable as pxt
-from pixeltable import catalog, exceptions as excs, iterators as iters
+import pixeltable.catalog as catalog
```
aaron-siegel, line 8: Why change the `from pixeltable ...` pattern?

pixeltable/index/base.py
```diff
@@ -32,15 +36,20 @@ def records_value_errors(self) -> bool:
...
         pass
 
     @abc.abstractmethod
-    def create_index(self, index_name: str, index_value_col: catalog.Column) -> None:
-        """Create the index on the index value column"""
+    def sa_index(self, index_name: str, index_value_col: catalog.Column) -> sql.Index:
+        """Return a sqlalchemy Index instance"""
         pass
 
+    # @abc.abstractmethod
```
aaron-siegel: Strip out this cruft?

pyproject.toml
```diff
@@ -268,6 +268,7 @@ ignore = [
     "PLC1901",  # Allow empty string comparisons
     "PLC2701",  # Allow private name imports
     "PLE0605",  # Invalid format for __all__ (appears to generate false positives)
+    "PLR0402",  # Allow `import x.y as y` instead of `from x import y` (needed to avoid circular imports)
```
aaron-siegel, line 273: Ah, now I see. This really makes a difference wrt circular imports? Ok, you can ignore the preceding comments around this, but I may play around with it a bit at some point. I guess `from x import y` attempts to instantiate `x`, and `import x.y as y` doesn't?

tests/test_mcp.py
```diff
@@ -7,11 +7,12 @@
 
 import pixeltable as pxt
 
-from .utils import skip_test_if_no_client, skip_test_if_not_installed
+from .utils import IN_CI, skip_test_if_no_client, skip_test_if_not_installed
 
 _logger = logging.getLogger('pixeltable')
 
 
+@pytest.mark.skipif(not IN_CI, reason='Hangs')
```
aaron-siegel: Can you provide some more color here? It doesn't hang on my system... is this something that's happening on your Linux machine or are you seeing it on MacOS too? Has it always been the case?

pixeltable/index/base.py
mkornacker: Done.

pyproject.toml
mkornacker: I spent a good amount of time trying to get rid of circular imports, not fun and not a good use of time...

pixeltable/exprs/column_ref.py
mkornacker: Yep, see below.

pixeltable/catalog/table_version.py
mkornacker: See below.

pixeltable/catalog/table_version.py

pixeltable/catalog/table_version.py

pixeltable/catalog/column.py
mkornacker: Only TableVersion methods use this. Column doesn't have the full context to set it correctly on its own (and shouldn't have to reach back into TableVersion).

The alternative seems more convoluted (you can have a TableVersion.stores_cellmd_default(is_computed, col_type), but now you need to set this in TV.add_column()). Either way, not a significant change. Let's leave it for now.

pixeltable/catalog/column.py
mkornacker: Done.

pixeltable/catalog/table_version.py

pixeltable/catalog/table_version.py

pixeltable/catalog/table_version.py

pixeltable/catalog/table_version.py

pixeltable/exprs/expr.py
mkornacker: Done.

pixeltable/catalog/table_version.py
```diff
@@ -417,98 +435,110 @@ def init(self) -> None:
...
+            if col_md.schema_version_add <= self.schema_version and (
+                col_md.schema_version_drop is None or col_md.schema_version_drop > self.schema_version
+            ):
+                if col.name is not None:
+                    self.cols_by_name[col.name] = col
+                self.cols_by_id[col.id] = col
+
+        if self.supports_idxs:
+            # create IndexInfo for indices visible in current_version
+            for md in (
```
aaron-siegel: Again, unorthodox/confusing pattern. Why not this, which is more familiar and easier to read?
```
for md in self.tbl_md.index_md.values():
     if ...:
```

pixeltable/catalog/table_version.py
```diff
@@ -417,98 +435,110 @@ def init(self) -> None:
...
+                md
+                for md in self.tbl_md.index_md.values()
+                if md.schema_version_add <= self.schema_version
```
aaron-siegel, line 512: Add a code comment explaining what this relatively complicated condition represents. (There was one in there before.)

pixeltable/exprs/column_ref.py
```diff
@@ -61,19 +65,17 @@ def __init__(
         perform_validation: Optional[bool] = None,
     ):
         super().__init__(col.col_type)
-        assert col.tbl is not None
+        # assert col.tbl is not None
```
aaron-siegel: dead code

pixeltable/globals.py
```diff
@@ -179,21 +181,27 @@ def create_table(
...
     )
-    if was_created and data_source is not None and not is_direct_df:
+
+    if was_created:
         fail_on_exception = OnErrorParameter.fail_on_exception(on_error)
-        table.insert_table_data_source(data_source=data_source, fail_on_exception=fail_on_exception)
+        if isinstance(data_source, DFTableDataConduit):
+            df = data_source.pxt_df
+            with Catalog.get().begin_xact(tbl=tbl._tbl_version_path, for_write=True, lock_mutable_tree=True):
+                tbl._tbl_version.get().insert(None, df, fail_on_exception=fail_on_exception)
```
aaron-siegel, line 201: Is this going to happen outside of the table creation transaction now? Shouldn't it be threaded through as a PendingTableOp so that the table is populated from source atomically?

pixeltable/index/embedding_index.py
```diff
@@ -39,16 +41,15 @@ class Metric(enum.Enum):
     }
 
     metric: Metric
-    value_expr: exprs.FunctionCall
+    # value_expr: exprs.FunctionCall
```
aaron-siegel: dead code

pixeltable/store.py
```diff
@@ -161,11 +163,56 @@ def create(self) -> None:
...
+                env.conn.execute(sql.text(stmt))
+            else:
+                # each of these statements need to be run as separate transactions; if one fails, we still want to run
+                # the rest
+                with env.begin_xact(for_write=True):
+                    self._exec_if_not_exists(stmt)
+
+    def create_index(self, idx_id: int) -> None:
+        """Create If Not Exists for this index"""
+        idx_info = self.tbl_version.get().idxs[idx_id]
```
aaron-siegel, line 225: Keep track of `all_idxs` separately in TableVersion (a record of all indexes, including dropped indexes), and use that here to get a fully accurate representation of the store schema. We might as well, and it will solve the problem below (see additional later comment).

pixeltable/utils/dbms.py
```diff
@@ -82,11 +77,9 @@ def create_db_stmt(self, database: str) -> str:
...
-    def create_vector_index(
-        self, index_name: str, index_value_sa_col: sql.schema.Column, conn: sql.Connection, metric: str
-    ) -> None:
-        create_index_sql = sql.text(
-            f"""CREATE VECTOR INDEX {index_name} ON {index_value_sa_col.table.name}
-             ({index_value_sa_col.name} {metric})"""
-        )
-        conn.execute(create_index_sql)
+    def sa_vector_index(self, store_index_name: str, sa_value_col: sql.schema.Column, metric: str) -> sql.Index:
+        return None
```
aaron-siegel: What's happening here? Why's this commented out and is it going to break Cockroach integration?

tests/share/test_packager.py
```diff
@@ -218,8 +218,11 @@ def __check_table(self, bundle_info: BundleInfo, tbl_name: str) -> None:
         # Verify that the postgres schema subsumes the original.
         # (There may be additional columns in the restored table depending on the order in which different versions are
         # restored. But everything present in the original table must be present and identical in the restored table.)
+        # Note: This is actually not the case: the restored table doesn't contain indices that were dropped on the
```
aaron-siegel: But this was working before! The way I solved this was to separately maintain an `all_idxs` in TableVersion that includes all indexes, including dropped indexes (just as `cols` includes all columns, including dropped columns). See above comment. Then if we use that to rectify the schema in the Store table, we get a fully faithful representation of the original schema. It's working on `main` as we speak.

Is there some particular reason not to do it this way? Regardless, let's please not simply comment this out with a TODO; it's a crucial check against potentially severe performance bugs.

pixeltable/globals.py
mkornacker: Out of scope, I'll leave a todo.

pixeltable/index/embedding_index.py
mkornacker: Done

pixeltable/utils/dbms.py
mkornacker: Left comment.

tests/share/test_packager.py
mkornacker: As discussed, let's stick with only creating the visible indices.

pixeltable/exprs/column_ref.py
mkornacker: Done

pixeltable/catalog/table_version.py
mkornacker: Done

pixeltable/catalog/table_version.py
```diff
@@ -417,98 +435,111 @@ def init(self) -> None:
...
+        if self.supports_idxs:
+            # create IndexInfo for indices visible in current_version
+            visible_idxs = [
+                md
+                for md in self.tbl_md.index_md.values()
+                if md.schema_version_add <= self.schema_version
+                and (md.schema_version_drop is None or md.schema_version_drop > self.schema_version)
+            ]
+            for md in visible_idxs:
+                # for md in (md for md in self.tbl_md.index_md.values() if md.schema_version_drop is None):
```
aaron-siegel: extraneous?

pixeltable/globals.py
aaron-siegel: Maybe make a JIRA ticket for this too to be sure we don't lose track of it?

pixeltable/store.py
```diff
@@ -161,11 +163,77 @@ def create(self) -> None:
...
+
+        If this is run outside of a transaction, runs each DDL statement in its own transaction.
+        """
+        postgres_dialect = sql.dialects.postgresql.dialect()
+        stmt = sql.schema.CreateTable(self.sa_tbl, if_not_exists=True).compile(dialect=postgres_dialect)
+        create_stmt = str(stmt)
+        # 'lock table': force an update of the metadata cache in order to avoid
+        # psycopg.errors.UndefinedTable) relation "..." does not exist
+        # errors
+        lock_stmt = f'LOCK TABLE {self._storage_name()} IN ACCESS EXCLUSIVE MODE'
```
aaron-siegel: Is this going to work in cockroachdb?

tests/share/test_packager.py
aaron-siegel: Agreed. As long as the validation is in there, I'm happy.

tests/share/test_packager.py
```diff
@@ -218,8 +218,11 @@ def __check_table(self, bundle_info: BundleInfo, tbl_name: str) -> None:
...
+        # Table before the bundle was created.
+        # TODO: find a better way to validate the physical schema
```
aaron-siegel: Can this TODO be removed now?

mkornacker: Updated.

pixeltable/catalog/table_version.py

pixeltable/store.py
mkornacker: Added Todo.

PR #857, opened 2025-10-14
Add image recipes to cookbook

apreshill:
@pierrebrunelle updated based on our review earlier today. Key changes:

1. Each recipe follows a "query then commit" pattern
2. No subdirectories, but I did add a prefix to the nb filenames (`img-` and `iter-`)


pierrebrunelle (approved):

pierrebrunelle:
@apreshill 
1. Separate the insert statements from the create statement
2. "Using .head() is optional but recommended for expensive operations." - they still have to use .collect() if they don't use head, they can't just use head and nothing else
3. I'd do the same separation for the addition as a computed column and then the query as a different cell

apreshill:
> @apreshill
> 
>     1. Separate the insert statements from the create statement
> 
>     2. "Using .head() is optional but recommended for expensive operations." - they still have to use .collect() if they don't use head, they can't just use head and nothing else
> 
>     3. I'd do the same separation for the addition as a computed column and then the query as a different cell

Got it for 1 and 3.

For 2 - this is in our tables and operations nb (https://github.com/pixeltable/pixeltable/blob/main/docs/notebooks/fundamentals/tables-and-data-operations.ipynb)

> `head(n)` and `limit(n).collect()` appear similar in this example. But `head()` always returns the earliest rows in a table, whereas `limit()` makes no promises about the ordering of its results (unless you specify an `order_by()` clause - more on this below). 

And these two calls, `head(2)` and `limit(2).collect()` seem interchangeable (apart from the ordering). You don't have to collect before using head, as far as I can tell in all the recipes here.

What should be the preferred workflow here?

aaron-siegel (approved):

PR #859, opened 2025-10-14
Fixed incorrect documentation links

metadaddy:
Browsing the Pixeltable README in GitHub, I noticed many broken documentation links. It seems like the docs used to be at `https://docs.pixeltable.com/docs/`, but that `docs/` path no longer exists. It bugged me enough that I spent a few minutes fixing almost all of them. Most of the mappings were straightforward - simply removing `docs/` - but some were more involved, for example, it looks like some web pages are now notebooks in GitHub.

There is one remaining reference to `https://docs.pixeltable.com/docs/`, in `docs/sample-apps/multimodal-chat/README.md`. I couldn't figure out the correct link for `https://docs.pixeltable.com/docs/data-types-and-formats`.

I hope this is useful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ The only open source Python library providing declarative data infrastructure fo
 [![PyPI Package](https://img.shields.io/pypi/v/pixeltable?color=4D148C)](https://pypi.org/project/pixeltable/)
 [![My Discord (1306431018890166272)](https://img.shields.io/badge/💬-Discord-%235865F2.svg)](https://discord.gg/QPyqFYx2UN)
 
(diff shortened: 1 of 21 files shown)


pierrebrunelle (approved):

pierrebrunelle:
@metadaddy Thanks Pat - very useful. We updated the link last Friday and missed a couple as you have seen.

We have a staged V2 of our documentation site incoming that will be a good improvement to our current one. Let us know if you have feedback. The latest addition was to provide a unified view of the user guide and the SDK: https://docs.pixeltable.com/sdk/latest/pixeltable - that is normally available at: https://pixeltable.github.io/pixeltable/

PR #861, opened 2025-10-15
Fixes for load_replica_md() with non-snapshot tables

aaron-siegel:
Also fixes up `test_publish`, and sets `preprod` to be the default API URL for testing.


pixeltable/catalog/view.py
```diff
@@ -252,6 +252,12 @@ def _get_snapshot_path(cls, tbl_version_path: TableVersionPath) -> TableVersionP
             base=cls._get_snapshot_path(tbl_version_path.base) if tbl_version_path.base is not None else None,
         )
 
+    def _is_named_pure_snapshot(self) -> bool:
+        """
+        Returns True if this is a named pure snapshot (i.e., a pure snapshot that is a separate schema object).
+        """
+        return self._id != self._tbl_version_path.tbl_id
```
mkornacker, line 259: why not also self.__snapshot_only?

aaron-siegel: I can add it, but it's implied/redundant (the only way this inequality can hold is if `self._snapshot_only` is also true)

mkornacker: Okay, yeah.

PR #868, opened 2025-10-16
Improvements to random_tbl_ops script

aaron-siegel:
- Add move operations
- Add array, JSON, and image data
- Improved configurability (individual operations can be excluded, and various config parameters tuned, on the command line)

Also, adds `if_exists` and `if_not_exists` parameters to `pxt.move()`.

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -260,21 +260,21 @@ class TestTable:
         pxt.create_table('tbl3', {'c1': pxt.Int})
         assert sorted(pxt.list_tables()) == ['tbl2', 'tbl3']
 
-        with pytest.raises(pxt.Error, match="Path 'tbl3' already exists."):
+        with pytest.raises(pxt.Error, match=r"Path 'tbl3' already exists."):
             pxt.move('tbl2', 'tbl3')
         assert sorted(pxt.list_tables()) == ['tbl2', 'tbl3']
 
         pxt.move('tbl2', 'tbl3', if_exists='ignore')
         assert sorted(pxt.list_tables()) == ['tbl2', 'tbl3']
 
-        with pytest.raises(pxt.Error, match="Path 'tbl1' does not exist."):
+        with pytest.raises(pxt.Error, match=r"Path 'tbl1' does not exist."):
             pxt.move('tbl1', 'tbl4')
         assert sorted(pxt.list_tables()) == ['tbl2', 'tbl3']
 
         pxt.move('tbl1', 'tbl4', if_not_exists='ignore')
         assert sorted(pxt.list_tables()) == ['tbl2', 'tbl3']
 
-        with pytest.raises(pxt.Error, match='move\(\): source and destination cannot be identical'):
+        with pytest.raises(pxt.Error, match=r'move\(\): source and destination cannot be identical'):
             pxt.move('tbl1', 'tbl1')
 
     def test_columns(self, reset_db: None) -> None:


pixeltable/catalog/catalog.py
```diff
@@ -792,19 +792,25 @@ def _get_dir_contents(self, dir_id: UUID, recursive: bool = False) -> dict[str,
...
 
     @retry_loop(for_write=True)
-    def move(self, path: Path, new_path: Path) -> None:
-        self._move(path, new_path)
+    def move(self, path: Path, new_path: Path, if_exists: IfExistsParam, if_not_exists: IfNotExistsParam) -> None:
+        self._move(path, new_path, if_exists, if_not_exists)
 
-    def _move(self, path: Path, new_path: Path) -> None:
-        _, dest_dir, src_obj = self._prepare_dir_op(
+    def _move(self, path: Path, new_path: Path, if_exists: IfExistsParam, if_not_exists: IfNotExistsParam) -> None:
```
mkornacker: Does this mean it can now do a replace? (which would require a preceding drop)

pixeltable/catalog/table.py
```diff
@@ -77,6 +77,16 @@ def __init__(self, id: UUID, dir_id: UUID, name: str, tbl_version_path: TableVer
         self._tbl_version = None
 
     def _move(self, new_name: str, new_dir_id: UUID) -> None:
+        old_name = self._name
+        old_dir_id = self._dir_id
+
+        cat = catalog.Catalog.get()
+
+        @cat.register_undo_action
+        def _() -> None:
```
mkornacker, line 86: Leave a todo to deal with this in a more complete manner. move() should fail if the table got renamed concurrently, at which point we need to reload the Table instance. We shouldn't try to fix up Table instances, we need to invalidate them and force a reload.

tool/random_tbl_ops_2.py
```diff
@@ -122,37 +181,97 @@ def insert_rows(self) -> Iterator[str]:
...
+            return PIL.Image.fromarray(random_data, 'RGBA')
 
     def update_rows(self) -> Iterator[str]:
         t = self.get_random_tbl(allow_view=False)
-        p = random.choice(self.PRIMES)
-        yield f'Update rows in {self.tbl_descr(t)} where c0 % {p} == 0: '
-        us = t.where(t.c0 % p == 0).update({'c1': t.c1 + 1.9, 'c2': t.c2 + '_u'})
+        p = random.choice(PRIMES)
+        yield f'Update rows in {self.tbl_descr(t)} where bc_int % {p} == 0: '
+        us = t.where(t.bc_int % p == 0).update(
```
mkornacker, line 263: The where clause is problematic/can become problematic. It basically forces a full table scan, which affects the timing, and also the way concurrent operations will interact. At least some of the operations should use the available btree indices.

tool/random_tbl_ops_2.py
```diff
@@ -280,6 +408,8 @@ def make_parser() -> ArgumentParser:
     parser.add_argument(
         '-r', '--read-only-workers', type=int, default=0, help='Number of read-only workers (default: 0)'
     )
+    parser.add_argument('--exclude', nargs='+', type=str, help='List of operations to exclude')
```
mkornacker, line 428: Maybe also have an --ops parameter, so I can select what I want to run? Would also be nice to specify relative frequencies here.

mkornacker (approved):

pixeltable/catalog/table.py
aaron-siegel: Done.

tool/random_tbl_ops_2.py
aaron-siegel: I added a TODO for this. I assume something like this would leverage the btree indices?
```
t.where((t.bc_int >= a) & (t.bc_int < b)).update(...)
```

tool/random_tbl_ops_2.py
aaron-siegel: I added an `--include-only` parameter and left a TODO for adjusting the weights.

tool/random_tbl_ops_2.py
mkornacker: Yes.

PR #869, opened 2025-10-16
Update many error messages for consistency

aaron-siegel:
Our error message formatting and verbiage is all over the place, and it's been bugging me. On the plane I went through and audited many error messages for consistency (but not all: we have a lot of them!)

I suggest standardizing on the following conventions, though I'm less invested in the specific conventions than that there be a consistent standard.
- Entity names and string literals are enclosed in single quotes.
```
Column 'my_col' does not have an index
```
- Code elements (parameter names, type literals, code snippets, etc) are enclosed in backticks.
```
`on` expects an expression of type `Bool`
```
- However, a single literal that appears after a colon at the end of a message is unquoted.
```
Unknown column: my_col
```

Comments welcome.

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -1272,7 +1272,8 @@ class Table(SchemaObject):
         else:
             if col.get_tbl().id != self._tbl_version.id:
                 raise excs.Error(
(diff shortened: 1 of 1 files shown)


apreshill:
This is a great start - I like the 3 conventions you propose. We should use these conventions as a starting point for an error message style guide.

PR #870, opened 2025-10-17
UDFs for Hugging Face Auto model integrations

aaron-siegel:
- **Text**: `text_generation`, `text_classification`, `text_summarization`, `question_answering`, `named_entity_recognition`, `translation`
- **Audio**: `automatic_speech_recognition`, `text_to_speech` 
- **Vision**: `image_captioning`
- **Generative**: `text_to_image`, `image_to_image`, `image_to_video`


pixeltable/functions/huggingface.py
```diff
@@ -454,7 +459,1036 @@ def detr_to_coco(image: PIL.Image.Image, detr_info: dict[str, Any]) -> dict[str,
     return {'image': {'width': image.width, 'height': image.height}, 'annotations': annotations}
 
 
-T = TypeVar('T')
+@pxt.udf(batch_size=1)
```
mkornacker: What's the point of batching for batch_size=1?

pixeltable/functions/huggingface.py
```diff
@@ -454,7 +459,1036 @@ def detr_to_coco(image: PIL.Image.Image, detr_info: dict[str, Any]) -> dict[str,
...
+        if hasattr(pipeline, 'enable_memory_efficient_attention'):
+            pipeline.enable_memory_efficient_attention()
+    except Exception:
+        pass  # Ignore optimization failures
+
+    generator = None if seed is None else torch.Generator(device=device).manual_seed(seed)
+
+    with torch.no_grad():
+        result = pipeline(prompt, height=height, width=width, generator=generator, **model_kwargs)
+        return result.images[0]
```
mkornacker, line 1033: Return a list of images?

pixeltable/functions/huggingface.py
```diff
@@ -454,7 +459,1036 @@ def detr_to_coco(image: PIL.Image.Image, detr_info: dict[str, Any]) -> dict[str,
...
+        processor = _lookup_processor(model_id, AutoProcessor.from_pretrained)
+        vocoder_model = None
+
+    # Load speaker embeddings once for SpeechT5 (following speech2text pattern)
+    speaker_embeddings = None
+    if 'speecht5' in model_id.lower():
+        embeddings_dataset = load_dataset('Matthijs/cmu-arctic-xvectors', split='validation')
+        speaker_embeddings = torch.tensor(embeddings_dataset[speaker_id or 7306]['xvector']).unsqueeze(0).to(device)
+
+    # Process each text input - following pattern from other functions
```
mkornacker: each text input?

pixeltable/functions/huggingface.py
```diff
@@ -454,7 +459,1036 @@ def detr_to_coco(image: PIL.Image.Image, detr_info: dict[str, Any]) -> dict[str,
...
+    except Exception:
+        pass  # Ignore optimization failures
+
+    generator = None if seed is None else torch.Generator(device=device).manual_seed(seed)
+
+    processed_image = image.convert('RGB')
+
+    with torch.no_grad():
+        result = pipe(prompt=prompt, image=processed_image, generator=generator, **model_kwargs)
+        return result.images[0]
```
mkornacker, line 1212: Return the whole list?

mkornacker (approved):

pixeltable/functions/huggingface.py
aaron-siegel: There isn't one. Pure LLM slop. I thought I'd caught these all (there were a half dozen or so spurious "batched" UDFs) but evidently I missed one. Fixed.

pixeltable/functions/huggingface.py
aaron-siegel: I guess we could do this now, but let's factor it into our more general overhaul of multimodal UDFs now that we have images in JSON available. (For now, this is consistent with our other UDFs that do similar things)

pixeltable/functions/huggingface.py
aaron-siegel: This was one of those spurious "batched" UDFs and the comment is a vestige of that :) Removed.

pixeltable/functions/huggingface.py
aaron-siegel: See above comment

pixeltable/functions/huggingface.py
mkornacker: Is there any reason not to do this here, incrementally? Am I missing anything, or isn't it just
- change the return type to list
- return result

aaron-siegel: Obviously the technical implementation is trivial, but I'm more concerned about the product implications. A large fraction of the time I would expect users to be using the UDF in a way that returns a single image. If we return it as a single image, it gets materialized on disk with a usable localpath. If it's returned as a list, then the image will get wrapped in a one-element array and buried in our data block structures.

So there is a cost in terms of the user experience (less transparency in the data) with uncertain benefit (how often will users actually want to generate multiple images with identical prompt?) I think this needs more thought and my preference is to stick with the established patterns until we've thought through what we want the eventual product surface to be.

PR #873, opened 2025-10-18
Segmentation-related improvements

mkornacker:
- clip() with mode 'accurate'
- fix for ffmpeg_segment_cmd
- VideoSplitter with segment_times

### DIFF
--- a/tests/test_video.py
+++ b/tests/test_video.py
@@ -409,9 +409,10 @@ class TestVideo:
             # :-1: omit last row since it typically contains the remainder and may be shorter
             assert df.iloc[:-1]['duration'].between(duration - eps, duration + eps).all()
         if durations is not None:
+            # strict=False: we don't know how many segments got created
             assert all(
                 expected - eps <= actual and expected + eps >= actual
-                for expected, actual in zip(durations, result['duration'][:-1], strict=True)
+                for expected, actual in zip(durations, result['duration'][:-1], strict=False)
             )
         pxt.drop_table('validate_segments')
 


tests/test_video.py
```diff
@@ -406,9 +409,9 @@ def _validate_segments(
             # :-1: omit last row since it typically contains the remainder and may be shorter
             assert df.iloc[:-1]['duration'].between(duration - eps, duration + eps).all()
         if durations is not None:
-            assert len(durations) == len(result) - 1
+            _ = zip(durations, result['duration'][:-1])
```
aaron-siegel: adding `strict=True` will enforce that the two lists have the same length.

Also, set this to a variable rather than repeating the expression below? (I assume you broke this out for easier debugging?)

pixeltable/iterators/video.py
```diff
@@ -268,25 +269,31 @@ def __init__(
...
+            assert overlap is None or overlap < duration
 
         video_path = Path(video)
         assert video_path.exists() and video_path.is_file()
 
         self.video_path = video_path
         self.segment_duration = duration
-        self.overlap = overlap
-        self.min_segment_duration = min_segment_duration
+        self.overlap = overlap if overlap is not None else 0.0
```
aaron-siegel, line 294: `self.overlap = overlap or 0.0` would do the same thing (also below)

tests/test_video.py
```diff
@@ -635,40 +664,65 @@ def test_video_splitter(self, segment_duration: float, mode: str, reset_db: None
...
                 pxt.drop_table('videos', force=True)
 
+    @pytest.mark.parametrize('segment_times,mode', [([6.0, 11.0, 16.0], 'fast'), ([6.0, 11.0, 16.0], 'accurate')])
+    def test_video_splitter_segment_times(self, segment_times: list[float], mode: str, reset_db: None) -> None:
+        from pixeltable.iterators import VideoSplitter
+
+        eps = 0.1 if mode == 'fast' else 0.0
+        video_filepaths = get_video_files()
+        t = pxt.create_table('videos', {'video': pxt.Video})
+        t.insert([{'video': p} for p in video_filepaths])
```
aaron-siegel: spurious list creation

tests/test_video.py
```diff
@@ -635,40 +664,65 @@ def test_video_splitter(self, segment_duration: float, mode: str, reset_db: None
...
+        s = pxt.create_view(
+            'segments', t, iterator=VideoSplitter.create(video=t.video, segment_times=segment_times, mode=mode)
+        )
+        start_times = [0.0, *segment_times]
+        durations = [start_times[i + 1] - start_times[i] for i in range(len(start_times) - 1)]
+        self._validate_splitter_segments(t, s, 0.0, 0.0, expected_durations=durations, eps=eps)
+        pxt.drop_table('videos', force=True)
```
aaron-siegel: why? it'll get cleared out before the next test anyway, and sometimes it can be useful to leave them around for debugging.

tests/test_video.py
```diff
@@ -568,10 +574,18 @@ def test_concat_videos_mixed_formats(self, reset_db: None, tmp_path: Path) -> No
...
 
     def _validate_splitter_segments(
-        self, base: pxt.Table, segments_view: pxt.Table, overlap: float, min_segment_duration: float
+        self,
+        base: pxt.Table,
+        segments_view: pxt.Table,
+        overlap: float | None,
+        min_segment_duration: float,
+        expected_durations: list[float] | None = None,
+        eps: float = 0.0,
```
aaron-siegel: Document the meaning of `eps`? (I can infer it by reading the code, but it's not exactly self-documenting)

tests/test_video.py

tests/test_video.py

tests/test_video.py

tests/test_video.py

PR #881, opened 2025-10-22
[PXT-814] UDF to encode a numpy array to an audio file

sergey-mkhitaryan:
A UDF that uses pyav to encode an array representation of an audio to an audio type + tests.

### DIFF
--- a/pixeltable/functions/audio.py
+++ b/pixeltable/functions/audio.py
@@ -57,7 +57,7 @@ def get_metadata(audio: pxt.Audio) -> dict:
 
 @pxt.udf()
 def encode_audio(
-    audio_data: pxt.Array, *, input_sample_rate: int, format: str, output_sample_rate: int | None = None
+    audio_data: pxt.Array[pxt.Float], *, input_sample_rate: int, format: str, output_sample_rate: int | None = None
 ) -> pxt.Audio:
     """
     Encodes an audio clip represented as an array into a specified audio format.
--- a/tests/test_audio.py
+++ b/tests/test_audio.py
@@ -431,7 +431,9 @@ class TestAudio:
 
         update_status = t.add_computed_column(
             audio_file=encode_audio(
-                t.audio.array.astype(pxt.Array), input_sample_rate=t.audio.sampling_rate.astype(pxt.Int), format='flac'
+                t.audio.array.astype(pxt.Array[pxt.Float]),  # type: ignore[misc]
+                input_sample_rate=t.audio.sampling_rate.astype(pxt.Int),
+                format='flac',
             )
         )
         validate_update_status(update_status)


CONTRIBUTING.md
```diff
@@ -83,6 +83,8 @@ Once you've set up your dev environment, you're ready to start contributing PRs.
 
     * Don't worry about making small, incremental commits to your branch; they'll be squash-committed when it
         eventually gets merged to `main`.
+    * Use `make format` to format your Python code, and `make lint && make typecheck` to check style and type
```
aaron-siegel: Instead of `make lint && make typecheck` it's recommended to use simply `make check`, which also runs other static checks (docs checks and formatting checks).

pixeltable/functions/array.py
```diff
@@ -0,0 +1,85 @@
...
+audio_formats: dict[str, tuple[str, str]] = {
+    'wav': ('pcm_s16le', 'wav'),
+    'mp3': ('libmp3lame', 'mp3'),
+    'flac': ('flac', 'flac'),
+    'mp4': ('aac', 'm4a'),
+}
+
+
+@pxt.udf(is_method=True)
+def to_audio(
```
aaron-siegel: I'd argue that this belongs in `pxtf.audio`, not as a method of Array . I do think we should have a `pxtf.array`, but I'd expect it to contain primitive array operations, whereas audio conversion seems like something much higher-level.

A method name like `pxtf.audio.encode_audio()` or `pxtf.audio.from_array()` would make sense to me.

cc @mkornacker

pixeltable/functions/array.py
```diff
@@ -0,0 +1,85 @@
...
+    Parameters:
+        audio_data: An array representing the audio data. The shape should be (1, N) for mono audio or (2, N) for
+        stereo.
+        input_sample_rate: The sample rate of the input audio data.
+        format: The desired output audio format. The supported formats are "wav", "mp3", "flac", and "mp4".
+        output_sample_rate: The desired sample rate for the output audio. Defaults to the input sample rate if
+        unspecified.
+
+    Example:
+        update_status = t.add_computed_column(
```
aaron-siegel: Use proper makedocstrings formatting for examples (refer to docstrings elsewhere in the codebase for the formatting convention) so it will render properly.

Also, we usually leave out the `update_status = ` in our examples.

pixeltable/functions/array.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Parameters:
+        audio_data: An array representing the audio data. The shape should be (1, N) for mono audio or (2, N) for
+        stereo.
+        input_sample_rate: The sample rate of the input audio data.
+        format: The desired output audio format. The supported formats are "wav", "mp3", "flac", and "mp4".
+        output_sample_rate: The desired sample rate for the output audio. Defaults to the input sample rate if
+        unspecified.
+
+    Example:
```
aaron-siegel: I'd suggest including a verbal description of what the example is doing.

pixeltable/functions/array.py
```diff
@@ -0,0 +1,85 @@
...
+    audio_data: pxt.Array, *, input_sample_rate: int, format: str, output_sample_rate: int | None = None
+) -> pxt.Audio:
+    """
+    Encodes an audio clip represented as an array into a specified audio format.
+
+    Parameters:
+        audio_data: An array representing the audio data. The shape should be (1, N) for mono audio or (2, N) for
+        stereo.
+        input_sample_rate: The sample rate of the input audio data.
+        format: The desired output audio format. The supported formats are "wav", "mp3", "flac", and "mp4".
```
aaron-siegel: Single-quote string literals in docstrings

pixeltable/functions/array.py
```diff
@@ -0,0 +1,85 @@
...
+        output_sample_rate: The desired sample rate for the output audio. Defaults to the input sample rate if
+        unspecified.
+
+    Example:
+        update_status = t.add_computed_column(
+            audio_file=to_audio(
+                t.audio.array.astype(pxt.Array), input_sample_rate=t.audio.sampling_rate.astype(pxt.Int), format='flac'
```
aaron-siegel: `t.audio.array.astype(...)` - not clear what this is doing. I'd find a simpler example like this to be clearer:

```
to_audio(t.audio_array, input_sample_rate=44100, format='flac')
```

pixeltable/functions/array.py
```diff
@@ -0,0 +1,85 @@
...
+            ),
+        )
+    """
+    if format not in audio_formats:
+        raise pxt.Error(f'Only the following formats are supported: {audio_formats.keys}')
+    if output_sample_rate is None:
+        output_sample_rate = input_sample_rate
+
+    codec = audio_formats[format][0]
```
aaron-siegel: better style:
```
codec, ext = audio_formats[format]
```

pixeltable/functions/array.py
```diff
@@ -0,0 +1,85 @@
...
+        case 2:
+            # Stereo audio. Input layout: [[L0, L1, L2, ...],[R0, R1, R2, ...]],
+            # pyav expects: [L0, R0, L1, R1, L2, R2, ...]
+            layout = 'stereo'
+            audio_data_transformed = np.empty(audio_data.shape[1] * 2, dtype=audio_data.dtype)
+            audio_data_transformed[0::2] = audio_data[0]
+            audio_data_transformed[1::2] = audio_data[1]
+            audio_data_transformed = audio_data_transformed.reshape(1, -1)
+        case _:
+            raise pxt.Error(
```
aaron-siegel: You'll want to use `excs.Error` here. We'll want to do one of two things: either adopt my suggestion of moving this to `pxtf.audio`, or expose `pxtf.array` in `__init__.py`. Either way, keeping this as `pxt.Error` here is likely to create a circular dependency conflict.

pixeltable/functions/array.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    if format not in audio_formats:
+        raise pxt.Error(f'Only the following formats are supported: {audio_formats.keys}')
+    if output_sample_rate is None:
+        output_sample_rate = input_sample_rate
+
+    codec = audio_formats[format][0]
+    ext = audio_formats[format][1]
+    output_path = str(TempStore.create_path(extension=f'.{ext}'))
+
```
aaron-siegel: Add a guard that `audio_data` has correct dimensions (`len(audio_data.shape) == 2` or similar) or the user will get cryptic errors on malformed arrays.

pixeltable/functions/array.py
```diff
@@ -0,0 +1,85 @@
...
+@pxt.udf(is_method=True)
+def to_audio(
+    audio_data: pxt.Array, *, input_sample_rate: int, format: str, output_sample_rate: int | None = None
+) -> pxt.Audio:
+    """
+    Encodes an audio clip represented as an array into a specified audio format.
+
+    Parameters:
+        audio_data: An array representing the audio data. The shape should be (1, N) for mono audio or (2, N) for
```
aaron-siegel: Should we permit a 1D array in addition to a 2D (1 x N) array for mono data?

tests/test_array.py
```diff
@@ -0,0 +1,90 @@
...
+        # Use to_audio to encode it to an audio file
+        t = pxt.create_table('test_mp3_to_array_and_back', {'audio_array': pxt.Array[pxt.Float]})  # type: ignore[misc]
+        output_sample_rate = sample_rate // 2 if downsample else sample_rate
+        t.add_computed_column(
+            audio_file=to_audio(
+                t.audio_array, input_sample_rate=sample_rate, format=format, output_sample_rate=output_sample_rate
+            )
+        )
+        update_status = t.insert(audio_array=audio_data)
+        assert update_status.num_rows == 1
```
aaron-siegel: You can use the `validate_update_status` utility instead of coding these assertions separately.

tests/test_array.py
```diff
@@ -0,0 +1,90 @@
...
+        assert update_status.num_excs == 0
+
+        row = t.head(1)[0]
+        assert set(row.keys()) == {'audio_array', 'audio_file'}
+        encoded_path = row['audio_file']
+        if format == 'mp4':
```
aaron-siegel: This would replace 4 lines of code with 1:
```
assert encoded_path.endswith('.m4a' if format == 'mp4' else f'.{format}')
```
IMO that's clearer but it's up to you.

tests/test_array.py
```diff
@@ -0,0 +1,90 @@
...
+            assert encoded_path.endswith('.m4a')
+        else:
+            assert encoded_path.endswith(f'.{format}')
+        print(f'Encoded audio file: {row["audio_file"]}')
+
+        # Read back, decode, and validate the encoded file
+        with av.open(encoded_path) as container:
+            audio_stream = container.streams.audio[0]
+            duration_seconds = float(audio_stream.duration * audio_stream.time_base)
+            # the "10 minute tour" is actually about 5 minutes long
```
aaron-siegel: I wonder if we should use a smaller file for testing.

tests/test_array.py
```diff
@@ -0,0 +1,90 @@
...
+            assert len(container.streams.audio) == 1
+            audio_stream = container.streams.audio[0]
+            sample_rate = audio_stream.rate
+            audio_frames = [frame.to_ndarray() for frame in container.decode(audio_stream)]
+
+        audio_data = np.concatenate(audio_frames, axis=1)
+        assert len(audio_data) > 0
+        return audio_data, sample_rate
+
+    @pytest.mark.skipif(IN_CI, reason='Too much IO for CI')
```
aaron-siegel: How big is the dataset?

tests/test_array.py
```diff
@@ -0,0 +1,90 @@
...
+    @rerun(reruns=3, reruns_delay=15)  # Guard against connection errors downloading datasets
+    def test_encode_dataset_audio(self, reset_db: None) -> None:
+        """
+        The point of this test case is to validate to_audio UDF on a real-world dataset.
+
+        As an added bonus, it exercises the stereo codepath which the test above doesn't.
```
aaron-siegel: Best to add a stereo file to the above test as well.

tests/test_array.py
sergey-mkhitaryan: almost 4GB. which actually made me realize that I can just get the test split.

pixeltable/functions/array.py
amrosado: Yes, possibly you might want to encode several audio channels.

pixeltable/functions/array.py
amrosado: I prefer encode_audio based on the functionality defined here.

pixeltable/functions/array.py
amrosado: You might want to also consider spectrograms which are a more common form of this data used in AI/ML based on my experience in laryngology.

CONTRIBUTING.md
sergey-mkhitaryan: thanks!

tests/test_array.py
sergey-mkhitaryan: awesome

pixeltable/functions/array.py
sergey-mkhitaryan: as discussed offline, since `make test` passes, we are probably fine

tests/test_array.py
sergey-mkhitaryan: discussed offline, each test scenario takes 1-2 seconds, which is probably fine.

pixeltable/functions/audio.py
```diff
@@ -51,6 +55,78 @@ def get_metadata(audio: pxt.Audio) -> dict:
...
 
+@pxt.udf()
+def encode_audio(
+    audio_data: pxt.Array, *, input_sample_rate: int, format: str, output_sample_rate: int | None = None
+) -> pxt.Audio:
+    """
+    Encodes an audio clip represented as an array into a specified audio format.
+
+    Parameters:
+        audio_data: An array of sampled amplitudes. The accepted array shapes are (N) or (1, N) for mono audio or (2, N)
```
aaron-siegel: A 1-tuple in Python is written `(N,)`.

Also I'd suggest enclosing (N,) and (1, N) and (2, N) in backticks.

pixeltable/functions/audio.py
```diff
@@ -51,6 +55,78 @@ def get_metadata(audio: pxt.Audio) -> dict:
     return av_utils.get_metadata(audio)
 
 
+@pxt.udf()
+def encode_audio(
+    audio_data: pxt.Array, *, input_sample_rate: int, format: str, output_sample_rate: int | None = None
```
aaron-siegel: Should we make the parameter type `pxt.Array[pxt.Float]` ?

It's a question; I'm not sure of the answer.

pixeltable/functions/audio.py
```diff
@@ -51,6 +55,78 @@ def get_metadata(audio: pxt.Audio) -> dict:
...
+            # Stereo audio. Input layout: [[L0, L1, L2, ...],[R0, R1, R2, ...]],
+            # pyav expects: [L0, R0, L1, R1, L2, R2, ...]
+            layout = 'stereo'
+            audio_data_transformed = np.empty(audio_data.shape[1] * 2, dtype=audio_data.dtype)
+            audio_data_transformed[0::2] = audio_data[0]
+            audio_data_transformed[1::2] = audio_data[1]
+            audio_data_transformed = audio_data_transformed.reshape(1, -1)
+        case _:
+            raise pxt.Error(
+                f'Supported input array shapes are (N), (1, N) for mono and (2, N) for stereo, got {audio_data.shape}'
```
aaron-siegel: (N,)

pixeltable/utils/av.py
```diff
@@ -5,6 +5,14 @@
 
 from pixeltable.env import Env
 
+# format -> (codec, extension)
+audio_format_defaults: dict[str, tuple[str, str]] = {
```
aaron-siegel, line 9: Good idea to refactor this. As long as we're moving it here, maybe rename it `AUDIO_FORMATS`?

pixeltable/functions/audio.py
sergey-mkhitaryan: yes, I think so. I did not do that initially because static analysis fails it (`error: The type "type[Array]" is not generic and not indexable`) but looking at other similar code, I see that we typically just ignore these warnings.

PR #883, opened 2025-10-24
Default media destination as configuration parameter

aaron-siegel:
Introduces two new configuration parameters: `input_media_dest` and `output_media_dest`, which serve as the default media destinations for data columns and computed columns, respectively. They may be either local paths or cloud buckets. If either or both is not specified, the "default default" is the local media dir (usually `~/.pixeltable/media`).

Extends our CI to allow for specifying additional environment variables on a per-configuration basis, and adds a CI job that sets `PIXELTABLE_OUTPUT_MEDIA_DEST=` an S3 bucket. Other configurations (including `input_media_dest`) are not yet tested).

Includes some improvements to S3 error handling.

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -818,9 +818,7 @@ class Table(SchemaObject):
                 )
             )
         if col._explicit_destination is not None and not (col.stored and col.is_computed):
-            raise excs.Error(
-                f'Column {col.name!r}: `destination` property only applies to stored computed columns'
-            )
+            raise excs.Error(f'Column {col.name!r}: `destination` property only applies to stored computed columns')
 
     @classmethod
     def _verify_schema(cls, schema: list[Column]) -> None:


pixeltable/catalog/column.py
```diff
@@ -219,10 +234,10 @@ def check_value_expr(self) -> None:
             )
 
     def has_window_fn_call(self) -> bool:
-        if self.value_expr is None:
-            return False
         from pixeltable import exprs
 
+        if self.value_expr is None:
```
sergey-mkhitaryan: curious why this change?

pixeltable/env.py
```diff
@@ -423,6 +416,16 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
                 f'(either add a `file_cache_size_g` entry to the `pixeltable` section of {Config.get().config_file},\n'
                 'or set the PIXELTABLE_FILE_CACHE_SIZE_G environment variable)'
             )
+
+        self._default_input_media_dest = config.get_string_value('input_media_dest')
```
mkornacker, line 422: Missing from class-level decls and init().

pixeltable/utils/object_stores.py
```diff
@@ -472,6 +459,16 @@ def delete(cls, dest: str | None, tbl_id: UUID, tbl_version: int | None = None)
         store = cls.get_store(dest, False)
         return store.delete(tbl_id, tbl_version)
 
+    @classmethod
+    def count_default_input_dest(cls, tbl_id: UUID, tbl_version: int | None = None) -> int:
```
mkornacker: Roll this into count(default_input=True)/count(default_output=True) to cut down on overall complexity of this already complex class hierarchy?

pixeltable/catalog/column.py
```diff
@@ -186,6 +187,20 @@ def get_tbl(self) -> TableVersion:
         tv = self.tbl_handle.get()
         return tv
 
+    @property
+    def resolved_destination(self) -> str | None:
```
mkornacker: Simply destination, and move the field to _destination?

aaron-siegel: This is how I had it initially, but I thought it was too confusing with `destination` and `_destination` having fundamentally different meanings.

I'm open to something other than `resolved_destination` but I do think we should disambiguate the terminology.

pixeltable/env.py
aaron-siegel: Fixed in class decls.

Do we actually need to set them in `init()`? I'd argue it's better to get an error if an attempt is made to access them before calling `set_up()`, rather than silently (and often incorrectly) defaulting to `None`.

I guess we could load them in config during `init()` instead of `set_up()`.

pixeltable/utils/object_stores.py
aaron-siegel: Ok, done.

mkornacker (approved):

PR #885, opened 2025-10-25
Initial integration of TwelveLabs Embed API

mkornacker:


pixeltable/functions/twelvelabs.py
```diff
@@ -0,0 +1,89 @@
...
+    Examples:
+        Add a computed column `embed` for an embedding of a string column `input`:
+
+        >>> tbl.add_computed_column(
+        ...     embed=embed(model_name='Marengo-retrieval-2.7', text=tbl.input)
+        ... )
+    """
+    cl = _twelvelabs_client()
+    res = await cl.embed.create(
+        model_name=model_name, text=text, text_truncate=text_truncate, audio_file=audio, **kwargs
```
aaron-siegel, line 77: Seems to be missing the image param?

pixeltable/functions/twelvelabs.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    if text is not None:
+        assert res.text_embedding is not None
```
aaron-siegel: Not sure this should be an assertion since it's not internal to our code (it's validating what comes out of the SDK)? Maybe raise an exception to say "invalid response" instead?

pixeltable/functions/twelvelabs.py
```diff
@@ -0,0 +1,89 @@
...
+    Request throttling:
+    Applies the rate limit set in the config (section `twelvelabs`, key `rate_limit`). If no rate
+    limit is configured, uses a default of 600 RPM.
+
+    __Requirements:__
+
+    - `pip install twelvelabs`
+
+    Args:
+        model_name: The name of the model to use. Must be `'Marengo-retrieval-2.7'`.
```
aaron-siegel: Do we really want to hard-code this as the only option? This means we need to update our UDF any time their model changes.

mkornacker: Done.

pixeltable/functions/twelvelabs.py
mkornacker: Added Todo.

pixeltable/functions/twelvelabs.py
mkornacker: Done.

PR #886, opened 2025-10-25
Azure support

aaron-siegel:


pixeltable/utils/azure_store.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+client_lock = threading.Lock()
+
+
+class AzureClientDict(NamedTuple):
+    """Container for actual Azure access objects (clients, resources)
+    Thread-safe, protected by the module lock 'client_lock'"""
+
+    profile: Optional[str]  # profile used to find credentials
```
mkornacker: Are we not using '| None' in new code?

pixeltable/utils/azure_store.py
```diff
@@ -0,0 +1,297 @@
...
+    # Container name extracted from the URI
+    __container_name: str
+
+    # Prefix path within the container, either empty or ending with a slash
+    __prefix_name: str
+
+    # URI scheme (wasb, wasbs, abfs, abfss, https)
+    __scheme: str
+
+    soa: StorageObjectAddress
```
mkornacker, line 52: Isn't this just a parsed uri? Should we get rid of 'soa' as a variable name?

pixeltable/utils/azure_store.py
```diff
@@ -0,0 +1,297 @@
...
+
+client_lock = threading.Lock()
+
+
+class AzureClientDict(NamedTuple):
+    """Container for actual Azure access objects (clients, resources)
+    Thread-safe, protected by the module lock 'client_lock'"""
+
+    profile: Optional[str]  # profile used to find credentials
+    clients: dict[str, Any]  # Dictionary of URI to client object attached to the URI
```
mkornacker: Don't the clients have a type?

pixeltable/utils/azure_store.py
```diff
@@ -0,0 +1,297 @@
...
+    def __init__(self, soa: StorageObjectAddress):
+        self.soa = soa
+        self.__scheme = soa.scheme
+        self.__account_name = soa.account
+        self.__container_name = soa.container
+        self.__prefix_name = soa.prefix
+
+        # Reconstruct base URI in normalized format
+        self.__base_uri = self.soa.prefix_free_uri + self.__prefix_name
+        if 0:
```
mkornacker: Remove

pixeltable/utils/azure_store.py
```diff
@@ -0,0 +1,297 @@
...
+            else:
+                raise excs.Error(
+                    f'HTTP error during {operation} in container {container_name}: {e.status_code} - {str(e)!r}'
+                )
+        else:
+            raise excs.Error(f'Error during {operation} in container {container_name}: {str(e)!r}')
+
+    @classmethod
+    def create_az_raw(cls, endpoint_url: str, account_name: str | None = None, account_key: str | None = None) -> Any:
+        """Get a raw client without any locking"""
```
mkornacker: What does 'raw' mean? Should this be create_client()? Why 'az'? (Refers to availability zone in AWS.)

Is there a reason for having it return Any?

pixeltable/utils/azure_store.py
```diff
@@ -0,0 +1,297 @@
...
+
+        # Reconstruct base URI in normalized format
+        self.__base_uri = self.soa.prefix_free_uri + self.__prefix_name
+        if 0:
+            print(
+                f'Initialized AzureBlobStore with base URI: {self.__base_uri},',
+                f'account: {self.__account_name}, container: {self.__container_name}, prefix: {self.__prefix_name}',
+            )
+
+    def client(self) -> Any:
```
mkornacker: Any objection to declaring the actual type? Any makes the code harder to read.

pixeltable/utils/azure_store.py
```diff
@@ -0,0 +1,297 @@
...
+        try:
+            container_client = self.client().get_container_client(self.container_name)
+            # Check if container exists by trying to get its properties
+            container_client.get_container_properties()
+            return self.__base_uri
+        except AzureError as e:
+            self.handle_azure_error(e, self.container_name, f'validate container {error_col_name}')
+        return None
+
+    def copy_object_to_local_file(self, src_path: str, dest_path: Path) -> None:
```
mkornacker, line 120: (Just as an aside, I think 'download()' and 'upload()' instead of copy_...() would be more concise and also clearer.)

pixeltable/utils/azure_store.py
```diff
@@ -0,0 +1,297 @@
...
+        try:
+            blob_client = self.client().get_blob_client(container=self.container_name, blob=self.prefix + src_path)
+            with open(dest_path, 'wb') as download_file:
+                download_stream = blob_client.download_blob()
+                download_file.write(download_stream.readall())
+        except AzureError as e:
+            self.handle_azure_error(e, self.container_name, f'download file {src_path}')
+            raise
+
+    def copy_local_file(self, col: 'Column', src_path: Path) -> str:
```
mkornacker, line 132: (Another aside, Column shouldn't leak into utils.)

pixeltable/utils/azure_store.py
```diff
@@ -0,0 +1,297 @@
...
+                    f'HTTP error during {operation} in container {container_name}: {e.status_code} - {str(e)!r}'
+                )
+        else:
+            raise excs.Error(f'Error during {operation} in container {container_name}: {str(e)!r}')
+
+    @classmethod
+    def create_az_raw(cls, endpoint_url: str, account_name: str | None = None, account_key: str | None = None) -> Any:
+        """Get a raw client without any locking"""
+        from azure.core.credentials import AzureNamedKeyCredential
+        from azure.storage.blob import BlobServiceClient
```
mkornacker: There is a azure.storage.blob.aio, which is what we should be using.

Leave a class Todo that this needs to be redesigned for async io.

pixeltable/utils/azure_store.py
aaron-siegel: It's not new code, it's adapted from Jim's existing PR. I'm planning to do a global search-and-replace to switch to `| None` but I'm waiting for the create_table() PR to merge. That'll be the easiest way to update these.

pixeltable/utils/azure_store.py
aaron-siegel: Changed to `_logger.info()`

pixeltable/utils/azure_store.py
aaron-siegel: will fix in a later PR.

pixeltable/utils/azure_store.py
aaron-siegel: Fixed.

pixeltable/utils/azure_store.py
aaron-siegel: Fixed

pixeltable/utils/azure_store.py
aaron-siegel: Will fix in upcoming cleanup PR

pixeltable/utils/azure_store.py
aaron-siegel: Added TODO

pixeltable/utils/azure_store.py
aaron-siegel: `az` is an abbreviation for Azure; it might be a bespoke one. I changed the method names and will clean up other references later

pixeltable/utils/azure_store.py
aaron-siegel: Done.

mkornacker (approved):

PR #895, opened 2025-10-28
[PXT-822, PXT-674] Fix for querying snapshots of tables with unstored columns

aaron-siegel:
Fixes a bug involving snapshots of tables with unstored columns; previously, attempting to query them would trigger an `AssertionError`. The issue was that column `value_expr`s in snapshot table versions were not being retargeted.

The bug would also manifest in snapshots of views with explicit column selections, as well as (data-sharing) replicas of views with explicit column selections, both of which rely on snapshots over unstored columns.

Fixes another (less critical) bug in the process, in which the `str(computed_with)` display of a snapshot would use the incorrect column names, in the event that the columns had been renamed in a later version of the table. (See: PXT-674)

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -1722,9 +1722,7 @@ class Catalog:
         self._tbls[tbl_id, version] = view
         return view
 
-    def construct_tvp(
-        self, tbl_id: UUID, version: int, ancestor_ids: list[str], created_at: float
-    ) -> TableVersionPath:
+    def construct_tvp(self, tbl_id: UUID, version: int, ancestor_ids: list[str], created_at: float) -> TableVersionPath:
         # Construct the TableVersionPath for the specified TableVersion. We do this by examining the created_at
         # timestamps of this table and all its ancestors.
         # TODO: Store the relevant TableVersionPaths in the database, so that we don't need to rely on timestamps
@@ -1735,7 +1733,7 @@ class Catalog:
         # Build the list of ancestor versions, starting with the given table and traversing back to the base table.
         # For each proper ancestor, we use the version whose created_at timestamp equals or most nearly precedes the
         # given TableVersion's created_at timestamp.
-        ancestors: schema.TableVersionPath = [(tbl_id, version)]
+        ancestors: list[tuple[UUID, int]] = [(tbl_id, version)]
         for ancestor_id in ancestor_ids:
             q = (
                 sql.select(schema.TableVersion)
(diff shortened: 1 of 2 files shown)


pixeltable/catalog/catalog.py
```diff
@@ -1716,12 +1716,22 @@ def _load_tbl_at_version(self, tbl_id: UUID, version: int) -> Table | None:
...
         tbl_md = schema.md_from_dict(schema.TableMd, tbl_record.md)
         version_md = schema.md_from_dict(schema.TableVersionMd, version_record.md)
+        tvp = self.reconstruct_tvp(tbl_id, version, tbl_md, version_md)
 
+        view = View(tbl_id, tbl_record.dir_id, tbl_md.name, tvp, snapshot_only=True)
+        self._tbls[tbl_id, version] = view
+        return view
+
+    def reconstruct_tvp(
+        self, tbl_id: UUID, version: int, tbl_md: schema.TableMd, version_md: schema.TableVersionMd
```
mkornacker: construct_tvp(tbl_id, version, base_versions: schema.TableVersionPath | None, created_at: float) is more concise and easier to follow.

pixeltable/catalog/column.py
```diff
@@ -102,7 +103,7 @@ def __init__(
                     f'but it is a {type(computed_with)}'
                 )
             else:
-                self._value_expr = value_expr.copy()
+                self._value_expr = value_expr
```
mkornacker: Why not a copy?

pixeltable/catalog/table_version.py
```diff
@@ -529,8 +531,13 @@ def _init_schema(self) -> None:
                 self.idxs_by_col.setdefault(indexed_col_id, []).append(info)
 
         # create value exprs, now that we have all lookup structures in place
+        tvp = (
+            None
+            if self.effective_version is None
+            else Catalog.get().reconstruct_tvp(self.id, self.effective_version, self.tbl_md, self.version_md)
```
mkornacker: (self.id, self.version, ...)
is a bit clearer

pixeltable/catalog/table_version.py
```diff
@@ -529,8 +531,13 @@ def _init_schema(self) -> None:
                 self.idxs_by_col.setdefault(indexed_col_id, []).append(info)
 
         # create value exprs, now that we have all lookup structures in place
+        tvp = (
```
mkornacker: Leave comment that for snapshot TV instances, we need to retarget the value exprs, otherwise they refer to the live versions.

pixeltable/catalog/column.py
aaron-siegel: Fixed.

pixeltable/catalog/table_version.py
aaron-siegel: Done.

pixeltable/catalog/table_version.py
aaron-siegel: Done. (See above.)

pixeltable/catalog/catalog.py

I made it even slightly simpler,
```
construct_tvp(tbl_id, version, ancestor_ids: list[str], created_at)
```
We don't need the full existing `base_versions` since we're reconstructing the versions anyway.

PR #896, opened 2025-10-28
UV_OPTS in Makefile

aaron-siegel:
A very useful Makefile enhancement. You can now run
```
make test UV_OPTS=--no-dev
```
to run just the minimal Pixeltable tests. It's fast! Likewise,
```
make install UV_OPTS=--no-dev
```
and
```
make install
```
will switch between minimal/full dependencies in your environment. `uv` is awesome. It's so fast it only takes a few seconds to make the switch - much easier than swapping conda environments!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ help:
 	@echo '  clean         Remove generated files and temp files'
 	@echo '  pytest        Run `pytest`'
 	@echo '  fullpytest    Run `pytest`, including expensive tests'
-	@echo '  slimpytest    Run a minimal pytest configuration'
+	@echo '  slimpytest    Run `pytest` with a minimal set of tests'
 	@echo '  nbtest        Run `pytest` on notebooks'
 	@echo '  stresstest    Run stress tests such as random-tbl-ops'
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -65,6 +67,10 @@ help:
 	@echo '  docscheck     Run `mkdocs build --strict`'
 	@echo '  lint          Run `ruff check`'
 	@echo '  formatcheck   Run `ruff format --check` (check only, do not modify files)'
+	@echo ''
+	@echo 'Other useful commands:'
+	@echo '  make install UV_ARGS=--no-dev   Minimal Pixeltable installation (no dev packages)'
```
sergey-mkhitaryan: nit: we list other commands without `make ` and should probably keep it consistent

aaron-siegel: Right, this was an intentional choice - the others are lists of targets, whereas these are fully-baked command lines, so I thought it made sense to include `make` for clarity. But I don't feel strongly about it.

aaron-siegel: I added another section for "Parameters" and changed this one to "Example commands" to make things clear.

PR #897, opened 2025-10-28
Intelligible error message when replicating a view of an existing original base table

aaron-siegel:
Prints an intelligible error message when replicating a view of an existing original base table. This addresses the following scenario:
- I create a base table `t`
- I create a view `v` of `t`
- I publish `v`
- I drop the local copy of `v`
- Now I try to replicate the cloud version of `v`

Previously this resulted in an assertion.


pixeltable/catalog/catalog.py
```diff
@@ -1214,12 +1214,17 @@ def __store_replica_md(self, path: Path, md: schema.FullTableMd) -> None:
...
             conn.execute(q)
-        else:
-            assert existing_md_row.md['is_replica']
-            if md.tbl_md.current_version > existing_md_row.md['current_version']:
-                # New metadata is more recent than the metadata currently stored in the DB; we'll update the record
-                # in place in the DB.
-                new_tbl_md = dataclasses.replace(md.tbl_md, name=path.name, user=Env.get().user, is_replica=True)
+        elif not existing_md_row.md['is_replica']:
+            raise excs.Error(
+                'An attempt was made to replicate a view whose base table already exists in the local catalog '
```
mkornacker, line 1231: Is this making assumptions about the state of the caller? Maybe this should be checked in the caller?

Also, does that mean that the following won't work?
- I publish T
- someone else replicates T and creates a view V
- I replicate V

aaron-siegel: What's the argument for doing it in the caller versus here?

Yes, that's right - replicas on top of non-replica tables aren't currently supported. The critical scenario is this:

1. I publish T version 700
2. someone else replicates T and creates a view V that depends on T version 700
3. version 700 gets garbage collected on the original pxt instance
4. I replicate V

I can't instantiate V directly since version 700 no longer exists, so I need to reconstruct version 700 from the data in the V-bundle. But the data for T in the V-bundle might be a fragment, and we don't currently support fragment table versions for non-replica tables.

(Of course we don't have garbage collection right now, but I'd hesitate to implement a solution that we know is going to implode when we do.)

It might be doable but needs to be done carefully and would need more work. How important is it to support this scenario? cc @apreshill

mkornacker (approved):

PR #899, opened 2025-10-28
pyscenedetect udfs

mkornacker:


pixeltable/functions/video.py
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
     return drawtext_params
 
 
+@pxt.udf(is_method=True)
```
aaron-siegel, line 942: I believe these belong more logically grouped in a `pxtf.scene` package rather than as methods of `video`. Do we have any customer feedback on this yet?

One possible general principle for deciding when something is a primitive type method, versus a UDF in a separate package: built-in UDFs (i.e., those that depend only on required dependencies of Pixeltable) are methods; UDFs that depend on optional packages go in separate modules.

It's a simple and objective criterion. If it's controversial we should have that discussion, BUT I do think we should have *some* clear criterion for what goes where, and not be making a bunch of ad hoc decisions about it (which is likely to result in a disjointed API).

pixeltable/functions/video.py
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
...
+def scene_detect_adaptive(
+    video: pxt.Video,
+    *,
+    fps: float | None = None,
+    adaptive_threshold: float = 3.0,
```
aaron-siegel, line 947: Another option would be to group these options into `content_kwargs: dict[str, Any] | None` and `adaptive_kwargs` parameters, rather than spelling them out individually. The main advantage of doing it this way is it saves us from having to keep the Pixeltable UDF in version sync with the Pyscenedetect library, in the event that pyscene adds new features or (worse) introduces backward incompatible changes.

It would also align the API with the way we're now doing things in most other UDFs, such as the Hugging Face or OpenAI UDFs. That pattern was well received by customers.

(In fact the "backward incompatibility" issue has already surfaced at least once, which is what motivated using `model_kwargs` for the other APIs.)

pixeltable/functions/video.py
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
...
+
+        >>> tbl.add_computed_column(
+        ...     scene_cuts=tbl.video.scene_detect_adaptive(adaptive_threshold=2.0)
+        ... )
+
+        Analyze at a lower frame rate for faster processing:
+
+        >>> tbl.select(tbl.video.scene_detect_adaptive(fps=2.0)).collect()
+    """
+    from scenedetect.detectors import AdaptiveDetector, ContentDetector
```
aaron-siegel, line 1027: Add `require_package('scenedetect')` just above the imports on all UDFs, to generate more useful error messages.

pixeltable/utils/av.py
```diff
@@ -210,3 +217,83 @@ def ffmpeg_segment_cmd(
...
+
+    def __init__(self, path: Path, fps: float | None = None) -> None:
+        self.path = path
+        self.fps = 0.0 if fps is None else fps
+        self.container = None
+        self.video_framerate = None
+        self.video_time_base = None
+        self.video_start_time = None
+
+    def __enter__(self) -> VideoFrames:  # noqa: PYI034; importing Self breaks the mypy plugin
```
aaron-siegel:
```
from typing_extensions import Self
```
is what you want

tests/test_video.py
```diff
@@ -1020,6 +1019,25 @@ def test_with_audio(self, reset_db: None) -> None:
...
+        'udf',
+        [
+            pxtf.video.scene_detect_adaptive,
+            pxtf.video.scene_detect_content,
+            pxtf.video.scene_detect_threshold,
+            pxtf.video.scene_detect_histogram,
+            pxtf.video.scene_detect_hash,
+        ],
+    )
+    def test_scene_detect(self, udf: pxt.Function, reset_db: None) -> None:
```
aaron-siegel: Ideally there would be some test instances that exercise non-default parameters of the UDFs as well.

Perhaps something more like `test_all` in `TestTimestamp`, `TestString`, etc., would be appropriate here, rather than a large-ish `parametrize`?

pixeltable/utils/av.py
```diff
@@ -210,3 +217,83 @@ def ffmpeg_segment_cmd(
...
+        if self.container:
+            self.container.close()
+
+    def __iter__(self) -> Iterator[Item]:
+        num_returned = 0
+        frame_idx = -1
+        while True:
+            try:
+                frame = next(self.container.decode(video=0))
+            except StopIteration:
```
aaron-siegel:
```
except (StopIteration, EOFError):
    return
````

File: pixeltable/functions/video.py [hunk 7]
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
...
+    delta_sat: float = 1.0,
+    delta_lum: float = 1.0,
+    delta_edges: float = 0.0,
+    luma_only: bool = False,
+    kernel_size: int | None = None,
+    filter_mode: Literal['merge', 'suppress'] = 'merge',
+) -> list[dict]:
+    """
+    Detect scene cuts in a video using PySceneDetect's
+    [ContentDetector](https://www.scenedetect.com/docs/0.6.7/api/detectors.html#scenedetect.detectors.content_detector.ContentDetector).
```
aaron-siegel: Is there any sort of `latest` URL slug on `scenedetect.com` that would save us from having to hard-code `0.6.7` here? I imagine it will get out of date pretty quickly.

pixeltable/functions/video.py
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
...
+class _SceneDetectFrameInfo(NamedTuple):
+    frame_idx: int
+    frame_pts: int
+    frame_time: float
+
+
+def _scene_detect(video: str, fps: float, detector: 'SceneDetector') -> list[dict[str, int | float]]:
+    from scenedetect import FrameTimecode  # type: ignore[import-untyped]
+
+    with av_utils.VideoFrames(pathlib.Path(video), fps=fps) as frame_iter:
```
aaron-siegel, line 1414: Better for `VideoFrames` to take `Path | str` as its first argument?

pixeltable/functions/video.py
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
...
+            Edge detection can help detect cuts in scenes with similar colors but different content.
+        luma_only: If True, only analyzes changes in the luminance (brightness) channel of the video,
+            ignoring color information. This can be faster and may work better for grayscale content.
+        kernel_size: Size of kernel to use for post edge detection filtering. If None, automatically set based on video
+            resolution.
+
+    Returns:
+        A list of dictionaries, one for each detected scene cut, with the following keys:
+
+        - `frame_idx` (int): The frame index (0-based) where the scene cut was detected
```
aaron-siegel: Are these correct? It looks like `start_time`, `start_pts`, `duration` to me.

pixeltable/functions/video.py
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
...
+        delta_edges: Weight for edge detection changes. Higher values make edge changes more important.
+            Edge detection can help detect cuts in scenes with similar colors but different content.
+        luma_only: If True, only analyzes changes in the luminance (brightness) channel,
+            ignoring color information. This can be faster and may work better for grayscale content.
+        kernel_size: Size of kernel for expanding detected edges. Must be odd integer greater than or equal to 3. If
+            None, automatically set using video resolution.
+        filter_mode: How to handle fast cuts/flashes. 'merge' combines quick cuts, 'suppress' filters them out.
+
+    Returns:
+        A list of timestamps (in seconds) where scene cuts were detected, ordered chronologically.
```
aaron-siegel: The return value is described differently for different UDFs. Is that intentional?

pixeltable/functions/video.py
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
...
+        # in order to determine the cut frame times, we need to record frame times (chronologically) and look them
+        # up by index; trying to derive frame times from frame indices isn't possible due to variable frame rates
+        frame_info: list[_SceneDetectFrameInfo] = []
+
+        def process_cuts(cuts: list[FrameTimecode]) -> None:
+            nonlocal frame_info, start_time, start_pts
+            for cut_timecode in cuts:
+                cut_frame_idx = cut_timecode.get_frames()
+                info_offset = next((i for i, info in enumerate(frame_info) if info.frame_idx == cut_frame_idx), None)
+                assert info_offset is not None
```
aaron-siegel: So what's the purpose in having the `None` default on the above line?

pixeltable/functions/video.py
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
...
+                info = frame_info[info_offset]
```
aaron-siegel, line 1434: Add an
```
assert info.frame_time > start_time
```
?

This logic clearly assumes that pyscenedetect presents its scene boundaries in chronological order, and I'd certainly expect that to be the case, but perhaps it's worth adding this assertion just to be safe.

(Or skip over the cut instead of asserting, and log a warning, if we want to be resilient; I'd argue this is better than spitting out a negative duration)

pixeltable/functions/video.py
```diff
@@ -936,6 +939,495 @@ def _create_drawtext_params(
...
+                scene_cuts.append(
+                    {'start_time': start_time, 'start_pts': start_pts, 'duration': info.frame_time - start_time}
```
aaron-siegel, line 1436: Are these scene_cuts or scenes?

If they're scene_cuts, then `duration` doesn't really belong and `cut_time`, `cut_pts` would be more logical (to me) as key names.

If they're scenes, I'd prefer `start_time`, `start_pts`, `end_time`, `end_pts` as the keys. (I generally prefer `start`/`end` to `start`/`len` (it's more Pythonic), and it feels like an awkward asymmetry to include duration/end_time but not end_pts.). Also, there should be N+1 of them in that case.

pixeltable/functions/video.py
mkornacker: It doesn't look like pyscenedetect sees a lot of API changes. This way, you only have to consult one set of documentation to understand how to use the udf.

pixeltable/functions/video.py
mkornacker: Corrected.

pixeltable/functions/video.py
mkornacker: Added an assert.

pixeltable/functions/video.py
mkornacker: Corrected.

pixeltable/functions/video.py
mkornacker: Corrected.

pixeltable/functions/video.py
mkornacker: Seeing the assertion instead of a StopIteration.

pixeltable/utils/av.py
mkornacker: Done.

pixeltable/functions/video.py
mkornacker: Done.

apreshill:
Customer feedback on the API:

1. Since the parameters are different for the different scene detection algorithms (i.e., adaptive, content, etc.):
      - Slight preference for: separate functions per detector
      ```
      detect_adaptive(video, threshold=3.0, min_scene_len=15)
      detect_content(video, threshold=27.0)
      ```
      - Also works: single function + dicts for method-specific args
      ```
      detect(video, method='adaptive', adaptive_args={'threshold': 3.0, 'min_scene_len': 15})
      ```
 2. "Feels to me to be a video operation over its own domain — unless of course there could be another media type outside of the blanket “video” that could use scene detection, not aware of any though."

pixeltable/utils/av.py
mkornacker: Done.

tests/test_video.py

pixeltable/functions/video.py

pixeltable/functions/video.py
mkornacker: No particular downside to restricting to Path.

tests/test_video.py
```diff
@@ -1020,6 +1020,62 @@ def test_with_audio(self, reset_db: None) -> None:
         with pytest.raises(pxt.Error, match='audio_duration must be positive'):
             t.add_computed_column(invalid=with_audio(t.video, t.audio, audio_duration=-1.0))
 
+    def test_scene_detect(self, reset_db: None) -> None:
+        skip_test_if_not_installed('scenedetect')
+        video_filepaths = get_video_files()
+
+        test_params: list[tuple[pxt.Function, dict[str, Any]]] = [
```
aaron-siegel, line 1027: Wait, now it seems like we're missing the basic patterns (using default args only). To be clear, I wasn't suggesting getting rid of those, but complementing them with test instances that override the defaults.

I think it's important to test both ways for all APIs; we've seen bugs of both sorts crop up in the past (parameters intended as optional inadvertently marked as required; and optional parameters not being handled correctly).

pixeltable/functions/video.py
aaron-siegel: Doesn't look like anything changed here?

pixeltable/functions/video.py
aaron-siegel: But is there any downside to accepting both? And there's a definite upside; the main advantage is it reduces the number of independent conversions that need to be done, by localizing the conversion in one place in the codebase. It's a very common pattern in Python to accept `Path | str` for a pathname.

pixeltable/functions/video.py
mkornacker: I'm going to leave it as duration. You can convert that back to pts with the time base.

tests/test_video.py
mkornacker: Done.

pixeltable/functions/video.py
aaron-siegel: Ok, I now see that we're calling them `scenes` and that we always generate an initial scene (starting at the first observed frame in the video) so that we get N + 1 scenes in the number of cuts, including in the case N = 0. Assuming this understanding is correct, I'm good. I feel less strongly about duration vs. end_time, although I might add that pts <-> time conversion is annoying and one of the things I like about this API is that it takes care of it under the covers.

pixeltable/functions/video.py
```diff
@@ -936,6 +939,536 @@ def _create_drawtext_params(
...
+            cuts = detector.process_frame(timecode, frame_array)
+            process_cuts(cuts)
+
+        # Post-process to capture any final scene cuts
+        if frame_idx is not None:
+            final_timecode = FrameTimecode(frame_idx, video_fps)
+            final_cuts = detector.post_process(final_timecode)
+            process_cuts(final_cuts)
+
+            # if we didn't detect any cuts but the video has content, add the full video as a single scene
```
aaron-siegel, line 1459: Does one of our test instances cover this case?

pixeltable/utils/av.py
```diff
@@ -210,3 +218,81 @@ def ffmpeg_segment_cmd(
...
+
+    path: Path
+    fps: float
+    container: av.container.input.InputContainer | None
+    video_framerate: Fraction | None
+    video_time_base: Fraction | None
+    video_start_time: int | None
+
+    @dataclass
+    class Item:
```
aaron-siegel, line 240: maybe would make more sense as a TypedDict, but not a major issue ofc

tests/test_video.py
```diff
@@ -1019,25 +1019,61 @@ def test_with_audio(self, reset_db: None) -> None:
...
+            (pxtf.video.scene_detect_hash, {'threshold': 0.595, 'size': 24, 'lowpass': 3, 'min_scene_len': 15}),
+        ]
+        for udf, params in test_params:
+            t = pxt.create_table('videos', {'video': pxt.Video}, if_exists='replace')
+            t.insert({'video': p} for p in video_filepaths)
+            status = t.add_computed_column(scenes=udf(t.video, fps=2.0, **params))
+            assert status.num_excs == 0
+            res = t.select(t.scenes).collect()
+            assert len(res) == len(video_filepaths)
+            assert all(len(row['scenes']) > 0 for row in res)
```
aaron-siegel, line 1082: How predictable is scene resolution in pyscenedetect? Is it deterministic or is there a way to ensure determinism? If so, it might be worthwhile to validate a few outputs here (= exact number of scenes at minimum) as well as success.

pixeltable/functions/video.py
mkornacker: Yes, two of our three videos don't have scene cuts.

tests/test_video.py
mkornacker: The different algorithms all have strengths and weaknesses.

We generally have too few videos to test with, but that's beyond this PR.

PR #900, opened 2025-10-29
CockroachDB fixes + CI target

aaron-siegel:
- Fixes the use of vector embeddings with CockroachDB
- Adds a CI target for running just `test_table.py` against CockroachDB

### DIFF
--- a/tool/ci_tool.py
+++ b/tool/ci_tool.py
@@ -94,19 +94,22 @@ def generate_matrix(args: argparse.Namespace) -> None:
         configs.extend(MatrixConfig('minimal', 'py', os, '3.10', uv_options='--no-dev') for os in ALTERNATIVE_PLATFORMS)
 
         # tests_table.py only, against CockroachDB backend
-        cockroachdb_connect_str = os.environ.get('PXTTEST_COCKROACHDB_CONNECT_STR')
-        if cockroachdb_connect_str:
-            configs.append(
-                MatrixConfig(
-                    'cockroach',
-                    'py',
-                    'ubuntu-24.04',
-                    '3.10',
-                    uv_options='--no-dev',
-                    pytest_options='tests/test_table.py',
-                    extra_env=f'PIXELTABLE_DB_CONNECT_STR={cockroachdb_connect_str}',
+        if trigger != 'merge_group':
+            # TODO For now, skip this in merge queue, until we're confident we can run multiple concurrent instances.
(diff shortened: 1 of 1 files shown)


pixeltable/store.py
```diff
@@ -216,18 +216,30 @@ def create(self) -> None:
             # condition here?)
 
         # ensure that all visible indices exist by running Create Index If Not Exists
-        for index in self.sa_tbl.indexes:
-            create_stmt = sql.schema.CreateIndex(index, if_not_exists=True).compile(dialect=postgres_dialect)
-            self._exec_if_not_exists(str(create_stmt), wait_for_table=True)
+        if not Env.get().is_using_cockroachdb:
+            # TODO: This approach doesn't work on CockroachDB. (Do we need it to?)
```
mkornacker, line 220: I don't understand this todo, let's discuss tomorrow.

pixeltable/store.py
```diff
@@ -216,18 +216,30 @@ def create(self) -> None:
...
+
         idx_info = self.tbl_version.get().idxs[idx_id]
-        sa_idx = idx_info.idx.sa_index(self.tbl_version.get()._store_idx_name(idx_id), idx_info.val_col)
-        conn = Env.get().conn
-        stmt = sql.schema.CreateIndex(sa_idx, if_not_exists=True).compile(conn)
-        create_stmt = str(stmt)
-        self._exec_if_not_exists(create_stmt, wait_for_table=True)
+        stmt: sql.Compiled
+
+        if isinstance(idx_info.idx, index.EmbeddingIndex):
```
mkornacker, line 232: Let's find a solution that doesn't involve special-casing that here.

pixeltable/store.py
aaron-siegel: Ok. I don't understand your intent behind the new architecture for index creation (in particular how it's supposed to work with cockroach vector indices). Happy to discuss tomorrow

pixeltable/store.py
aaron-siegel: I 100% agree with this, but to be clear, that's not the goal of this PR.

The index refactor broke Pixeltable cloud (data sharing won't deploy off main right now). The goal of this PR is to get things back into a working state so we don't delay the launch of push() / pull() and have a stable baseline for cockroachdb testing. We'll have plenty of time to circle back and find a more elegant architecture here.

mkornacker (approved):

PR #901, opened 2025-10-29
[PXT-817] UDFs for reve.com

sergey-mkhitaryan:
Adds UDFs that wrap Reve create, edit, and remix APIs.

### DIFF
--- a/pixeltable/functions/reve.py
+++ b/pixeltable/functions/reve.py
@@ -83,6 +83,7 @@ class _ReveClient:
                     if len(img_data) == 0:
                         raise ReveUnexpectedError(f'Reve request {request_id} resulted in an empty image')
                     img = PIL.Image.open(BytesIO(img_data))
+                    img.load()
                     _logger.debug(
                         f'Reve request {request_id} successful. Image bytes: {len(img_data)}, size: {img.size}'
                         f', format: {img.format}, mode: {img.mode}'


pixeltable/utils/image.py
```diff
@@ -2,6 +2,8 @@
 
 
 def default_format(img: PIL.Image.Image) -> str:
+    if img.format == 'PNG':
```
sergey-mkhitaryan: I don't know if this is the best fix possible, but without it, PNG images would end up as jpeg on disk.

pixeltable/functions/reve.py
```diff
@@ -0,0 +1,259 @@
...
+                        f'{retry_after_header}'
+                    )
+                    # This error message is formatted specifically so that RequestRateScheduler can extract the retry
+                    # delay from it
+                    raise pxt.Error(
+                        f'Reve request {request_id} failed due to rate limiting (429). retry-after:'
+                        f'{retry_after_seconds}'
+                    )
+                case _:
+                    _logger.error(
```
aaron-siegel: I'd argue this is `info` log level, not `error`. Requests fail for all sorts of reasons and don't indicate a problem with Pixeltable.

pixeltable/functions/reve.py
```diff
@@ -0,0 +1,259 @@
...
+    prompt: str, *, format: str, aspect_ratio: str | None = None, version: str | None = None
+) -> PIL.Image.Image:
+    """
+    Creates an image from a text prompt.
+
+    This UDF wraps the `https://api.reve.com/v1/image/create` endpoint. For more information, refer to the official
+    [API documentation](https://api.reve.com/console/docs/create).
+
+    Args:
+        prompt (str): prompt describing the desired image
```
aaron-siegel: No need to specify the types here; they'll be filled in during docs generation.

pixeltable/functions/reve.py
```diff
@@ -0,0 +1,259 @@
...
+        format (str): desired output image format
+        aspect_ratio (str | None, optional): desired image aspect ratio, e.g. '3:2', '16:9', '1:1', etc.
+        version (str | None, optional): specific model version to use. Latest if not specified.
+
+    Returns:
+        A generated image
+
+    Examples:
+        Add a computed column with generated square images to a table with text prompts:
+        ```
```
aaron-siegel: Use Python code example docstring formatting within `Examples` blocks (see other Pixeltable docstrings e.g. in globals.py for examples)

pixeltable/functions/reve.py
```diff
@@ -0,0 +1,259 @@
...
+    return Env.get().get_client('reve')
+
+
+# TODO Regarding rate limiting: Reve appears to be going for a credits per minute rate limiting model, but does not
+# currently communicate rate limit information in responses. Therefore neither of the currently implemented limiting
+# strategies is a perfect match, but "request-rate" is the closest. Reve does not currently enforce the rate limits,
+# but when it does, we can revisit this choice.
+@pxt.udf(resource_pool='request-rate:reve')
+async def create(
+    prompt: str, *, format: str, aspect_ratio: str | None = None, version: str | None = None
```
aaron-siegel: Do we really want user-specified image formats as UDF parameters? None of our other UDFs have that, and since it's being returned initially as an in-memory PIL image it's not clear what it means. There's an argument that if we do want user-specified image formats, they should be controlled at the column definition level rather than as UDF parameters. I think this merits some discussion.

cc @mkornacker

pixeltable/functions/reve.py
```diff
@@ -0,0 +1,259 @@
...
+            )
+        )
+        ```
+    """
+    if len(images) == 0:
+        raise pxt.Error('Must include at least 1 reference image')
+
+    # Replace $0, $1, etc. with <img>0</img>, ...
+    for i in range(0, len(images)):
+        prompt = re.sub(f'\\${i}', f'<img>{i}</img>', prompt)
```
aaron-siegel: Does this mean we're introducing our own bespoke formatting `$0` etc. as a substitution for what Reve expects? I'd advise against that; we strive to be "hands off" in how we adapt third-party libraries.

pixeltable/functions/reve.py
```diff
@@ -0,0 +1,259 @@
...
+    payload = {'prompt': prompt, 'reference_images': [_to_base64(img) for img in images]}
+    if version is not None:
+        payload['version'] = version
+    if aspect_ratio is not None:
+        payload['aspect_ratio'] = aspect_ratio
+    result = await _client()._post('/v1/image/remix', payload=payload, format=output_format)
+    return result
+
+
+def _to_base64(image: PIL.Image.Image) -> str:
```
aaron-siegel: This exactly duplicates code in `pxtf.image`; maybe it's worth factoring it out into `utils` somewhere.

pixeltable/utils/image.py
aaron-siegel: That might be what we want. Until now Pixeltable `pxt.Image` instances have been treated as "formatless" bitmaps. Let's discuss

pixeltable/functions/reve.py
```diff
@@ -0,0 +1,259 @@
...
+
+    Examples:
+        Add a computed column with promotional collages to a table with original images:
+        ```
+        t.add_computed_column(
+            collage=(
+                reve.remix(
+                    'Create a promotional collage by combining the picture of a hotel room from $0'
+                    ' with city attractions from $1 and $2',
+                    images=[t.room_img, t.city_attraction, t.city_skyline],
```
aaron-siegel: This seems like an unusual example; having such specific columns seems unlikely. I wonder if we can come up with an example that uses a more realistic table structure?

pixeltable/functions/reve.py
sergey-mkhitaryan: not disagreeing with the overall sentiment, but there are some examples of UDFs taking a format argument: `audio.encode_audio`, `image.b64_encode`

https://github.com/search?q=org%3Apixeltable%20path%3Afunctions+%22format%3A+str%22&type=code

sergey-mkhitaryan: offline discussion: 
1. drop format as a UDF parameter, always use png for now
2. save to disk using the default format, which is webp or jpeg at the moment.

tests/functions/test_reve.py
```diff
@@ -0,0 +1,43 @@
...
+
+from ..utils import rerun, skip_test_if_no_client, validate_update_status
+
+_logger = logging.getLogger('pixeltable')
+
+
+@pytest.mark.remote_api
+@pytest.mark.expensive
+@rerun(reruns=3, reruns_delay=8)
+class TestReve:
```
aaron-siegel, line 16: Worth adding some test cases that exercise the `version` parameter? Also, it's always good practice to have test cases that use default parameters only

aaron-siegel (approved):

PR #902, opened 2025-10-29
Rename `DataFrame` to `Query` and `DataFrameResultSet` to `ResultSet`

aaron-siegel:
I tried to be as thorough as I could, renaming local variables and doc references in addition to classes.

It wasn't entirely trivial, since `df` can also (and often does) mean "Pandas DataFrame", and also appears as a substring of `udf`, `pdf`, and `soundfile`. But I did my best to rename carefully.


docs/api/pixeltable/query.md
```diff
@@ -0,0 +1,50 @@
+# Query
+
+`Query` represents a query against a specific table. Unlike containers such as Pandas DataFrames,
+Pixeltable Queries do not hold data or allow you to update data
```
mkornacker: We actually have Df.update()/delete()/recompute_columns(), which are missing below.

pixeltable/dataframe.py
```diff
@@ -27,12 +27,12 @@
     import torch
     import torch.utils.data
 
-__all__ = ['DataFrame']
```
mkornacker, line 30: why _query.py?

aaron-siegel: `query.py` collides with the `@pxt.query()` so can't be used unless we rename the decorator. So I figured this would be a good opportunity to start moving to using "protected" names for non-public modules. (The DataFrame and ResultSet classes, of course, are made public by their imports in `pixeltable/__init__.py`.)

docs/api/pixeltable/query.md
aaron-siegel: Good point. I just removed the verbiage about Pandas; there's no compelling need to draw a contrast with Pandas DataFrame now.

PR #907, opened 2025-10-31
[PXT-786] push() and pull() implementations

amithadke:
Integrate protocol for push pull
 Adding version to replicate and pull
 Adding list_table_versions for testing (should be removed once we have remote table.get_versions())
 
 This change is using https://github.com/pixeltable/pixeltable/pull/890

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ dependencies = [
     "pyarrow>=19",
     "pixeltable-pgserver==0.4.0",
     "tzlocal>=5.0",
+    "aiohttp>=3.9",
 ]
 
 [project.urls]
--- a/uv.lock
+++ b/uv.lock
@@ -4947,6 +4947,7 @@ test = [
 
 [package.metadata]
 requires-dist = [
+    { name = "aiohttp", specifier = ">=3.9" },
     { name = "av", specifier = ">=10.0.0" },
     { name = "beautifulsoup4", specifier = ">=4.10" },
     { name = "cloudpickle", specifier = ">=2.2.1" },


pixeltable/globals.py
```diff
@@ -449,14 +449,15 @@ def publish(
...
-def replicate(remote_uri: str, local_path: str) -> catalog.Table:
+def replicate(remote_uri: str, local_path: str, version: int | None = None) -> catalog.Table:
     """
     Retrieve a replica from Pixeltable cloud as a local table. This will create a full local copy of the replica in a
     way that preserves the table structure of the original source data. Once replicated, the local table can be
     queried offline just as any other Pixeltable table.
 
     Args:
         remote_uri: Remote URI of the table to be replicated, such as `'pxt://org_name/my_dir/my_table'`.
+        version: Remote version of the table to be replicated, such as `1`.
```
aaron-siegel: Not sure the "such as `1`" is very helpful here :)

pixeltable/catalog/path.py
```diff
@@ -9,14 +10,9 @@
 _logger = logging.getLogger('pixeltable')
 
 
-class Path:
+class Path(NamedTuple):
     components: list[str]
-    version: int | None
-
-    def __init__(self, components: list[str], version: int | None = None) -> None:
-        assert len(components) > 0
```
mkornacker, line 17: Can we have this assertion back?

pixeltable/catalog/table_version.py
```diff
@@ -1431,6 +1431,21 @@ def unlink_external_store(self, store: ExternalStore) -> None:
         self._tbl_md.external_stores.pop(idx)
         self._write_md(new_version=True, new_schema_version=True)
 
+    @property
+    def cloud_uri(self) -> str | None:
```
mkornacker: Why not just additional_md, and the caller can call get()?

pixeltable/catalog/table_version.py
```diff
@@ -1431,6 +1431,21 @@ def unlink_external_store(self, store: ExternalStore) -> None:
...
+        return self._tbl_md.additional_md.get('cloud_uri')
```
mkornacker: Do we have a defined schema for additional_md somewhere? (pxt_uri would be a better name, just like we have s3_uri, etc.)

pixeltable/share/protocol/common.py
```diff
@@ -86,6 +86,14 @@ def _parse_and_validate_uri(cls, uri: str) -> dict:
 
         return {'org_slug': org_slug, 'db_slug': db_slug, 'table_identifier': table_identifier, 'is_uuid': is_uuid}
 
+    @classmethod
+    def from_components(cls, org_slug: str, table_identifier: str, db_slug: Optional[str] = None) -> PxtUri:
```
mkornacker: table_id for the sake of brevity? We don't spell out identifier elsewhere.

pixeltable/share/publish.py
```diff
@@ -269,10 +283,27 @@ def _download_from_presigned_url(
...
+    """List versions for a remote table."""
+    request_json = {'operation_type': 'list_table_versions', 'table_uri': {'uri': table_uri}}
+    response = requests.post(PIXELTABLE_API_URL, data=json.dumps(request_json), headers=_api_headers())
+    if response.status_code != 200:
+        raise excs.Error(f'Error listing table versions: {response.text}')
+    response_data = response.json()
+    return response_data.get('versions', [])
+
+
+def _format_uri_with_version(uri: str, version: int | None = None) -> str:
```
aaron-siegel: Not a blocker to merge this PR but a suggestion for the future: we should just build the version into the uri the way we do with `catalog.Path`

amithadke: yup that will be next change to protocol but we can make it backward compatible

pixeltable/globals.py
```diff
@@ -449,14 +449,15 @@ def publish(
     share.push_replica(destination_uri, source, bucket_name, access)
 
 
-def replicate(remote_uri: str, local_path: str) -> catalog.Table:
+def replicate(remote_uri: str, local_path: str, version: int | None = None) -> catalog.Table:
```
mkornacker: Shouldn't the version also be part of the uri? (just like in the local case)

pixeltable/share/publish.py
```diff
@@ -41,23 +45,23 @@ def push_replica(
...
         bucket_name=bucket,
         is_public=access == 'public',
     )
 
-    response = requests.post(PIXELTABLE_API_URL, json=publish_request.model_dump_json(), headers=_api_headers())
+    response = requests.post(PIXELTABLE_API_URL, data=publish_request.model_dump_json(), headers=_api_headers())
     if response.status_code != 200:
-        raise excs.Error(f'Error publishing snapshot: {response.text}')
-    publish_response = PublishResponse.model_validate_json(response.json())
+        raise excs.Error(f'Error publishing replica: {response.text}')
```
mkornacker: Shouldn't this be 'error publishing {src_tbl._display_name()}'? I thought we use replica purely to refer to local replicas of a published table.

pixeltable/share/publish.py
```diff
@@ -81,22 +85,28 @@ def push_replica(
...
-        raise excs.Error(f'Error finalizing snapshot: {finalize_response.text}')
-    finalize_response = FinalizeResponse.model_validate_json(finalize_response.json())
-    Env.get().console_logger.info(f'The published snapshot is now available at:{finalize_response.confirmed_table_uri}')
-    return str(finalize_response.confirmed_table_uri)
+    if finalize_response_json.status_code != 200:
+        raise excs.Error(f'Error finalizing replica: {finalize_response_json.text}')
+
+    finalize_response = FinalizeResponse.model_validate(finalize_response_json.json())
+    confirmed_tbl_uri = finalize_response.confirmed_table_uri
+    Env.get().console_logger.info(f'The published replica is now available at: {confirmed_tbl_uri}')
```
mkornacker: Same comment here.

pixeltable/share/publish.py
```diff
@@ -134,13 +144,17 @@ def pull_replica(dest_path: str, src_tbl_uri: str, version: int | None = None) -
...
     restorer = TableRestorer(
-        dest_path,
-        {'pxt_version': pxt.__version__, 'pxt_md_version': clone_response.pxt_md_version, 'md': clone_response.md},
+        dest_path, {'pxt_version': pxt.__version__, 'pxt_md_version': clone_response.pxt_md_version, 'md': md_list}
     )
+
     tbl = restorer.restore(bundle_path)
-    Env.get().console_logger.info(f'Created local replica {tbl._path()!r} from URI: {src_tbl_uri}')
+    Env.get().console_logger.info(
+        f'Created local replica {tbl._path()!r} from URI: {_format_uri_with_version(src_tbl_uri, version)}'
```
mkornacker: Same comment here, why wouldn't the version already be part of the uri?

pixeltable/catalog/path.py
aaron-siegel: Yes. In a `NamedTuple` the `__init__()` method is still called, but after the fields are populated, so @amithadke I believe the relevant code should look like this:
```
def __init__(self) -> None:
    assert len(self.components) > 0
```
As an aside, we should also change `components` to a `tuple[str, ...]` instead of a `list[str]` (it's more performant and also more correct, since `components` is immutable). I *believe* we could even make it a `tuple[str, str, ...]` and doing so would give a type guarantee that `len(components) > 0`. (That doesn't need to be done as part of this PR)

pixeltable/share/protocol/common.py
aaron-siegel: I believe it's not the usual meaning of `tbl_id`. It's a path OR uuid. Maybe `table_identifier` is a confusing term; how about `tbl_path_or_id`?

pixeltable/catalog/table_version.py
amithadke: we can rename it to pxt_uri 
For pixeltable core, we dont have schema defined we add TypedDict that has one entry pxt_uri

for cloud we store it as dict but its managed from cloud side directly (that schema is not visible here)

aaron-siegel: We could add a defined schema, but maybe as a follow up?

pixeltable/share/protocol/common.py
```diff
@@ -33,9 +33,10 @@ class PxtUri(BaseModel):
 
     # Parsed components
     org_slug: str  # Organization slug from the URI
-    db_slug: Optional[str]  # Database slug from the URI (optional)
-    table_identifier: str  # The table identifier (path or UUID)
-    is_uuid: bool  # True if table_identifier is a UUID, False if it's a path
+    db_slug: str | None  # Database slug from the URI (optional)
+    tbl_path_or_id: str  # The table path or UUID
```
mkornacker: How about 
tbl_path: catalog.Path | None
id: UUID | None
and leave out is_uuid

That should be easier to work with.

re db_slug: isn't this just 'db' (what does 'slug' add here?)

amithadke: tbl_path could also point to dir, im thinking of moving it to path
db_slug is what we are using in schema and i think in UI too, as db/project can have a display name like 'my cloud project'

mkornacker: But is 'db' a string identifier or something that can contain spaces? (in which case it should be called db_display_name or something like that).

I'm just trying to minimize the amount of new jargon. db_slug seems to mean 'db'. Same for org_slug.

amithadke: org or db slug wont support spaces we can rename org_slug to org and db_slug to db here as far as users know not to use display names. Slugs only support hyphens and alphanumerics

aaron-siegel: @mkornacker This is a non-user facing API right? We're under severe time pressure to make the Nov. 11 launch date; how important is it to be having this debate now? It's easy enough to refactor later after we've de-risked the launch

amithadke: i already moved it to db and org but left 'slug' in comment, most of the time we will be constructing pxturi off the string

PR #912, opened 2025-11-04
[PXT-809] Improve OpenAI rate limiting

sergey-mkhitaryan:
Unfortunately the 429 that we see in the OpenAI notebook is because embeddings API is not allowed on free tier at all, and no retry would ever succeed. Still I made some improvements to our resource-based rate limiting (used by OpenAI and Anthropic UDFs) based on what I've learnt.

### DIFF
--- a/pixeltable/env.py
+++ b/pixeltable/env.py
@@ -1119,7 +1119,7 @@ class RateLimitsInfo:
         # Find the highest wait until at least 5% availability of all resources
         max_wait = 0.0
         for limit_info in self.resource_limits.values():
-            time_until = limit_info.estimated_time_until_balance(
+            time_until = limit_info.estimated_resource_refill_delay(
                 math.ceil(TARGET_RATE_LIMIT_RESOURCE_FRACT * limit_info.limit)
             )
             if time_until is not None:
@@ -1161,7 +1161,7 @@ class RateLimitInfo:
         self.remaining = remaining
         self.reset_at = reset_at
 
-    def estimated_time_until_balance(self, target_remaining: int) -> float | None:
+    def estimated_resource_refill_delay(self, target_remaining: int) -> float | None:
         """Estimate time in seconds until remaining resources reaches target_remaining.
         Assumes linear replenishment of resources over time.
         Returns None if unable to estimate.
--- a/pixeltable/exec/expr_eval/schedulers.py
+++ b/pixeltable/exec/expr_eval/schedulers.py
@@ -161,7 +161,7 @@ class RateLimitsScheduler(Scheduler):
             info = self.pool_info.resource_limits[resource]
             # Note: usage and est_usage are estimated costs of requests, and it may be way off (for example, if max
             # tokens is unspecified for an openAI request).
-            time_until = info.estimated_time_until_balance(
+            time_until = info.estimated_resource_refill_delay(
                 math.ceil(info.limit * env.TARGET_RATE_LIMIT_RESOURCE_FRACT + usage + self.est_usage.get(resource, 0))
             )
             if time_until is not None and highest_wait < time_until:


pixeltable/env.py
```diff
@@ -1099,23 +1106,27 @@ def record(self, reset_exc: bool = False, **kwargs: Any) -> None:
...
         for limit_info in self.resource_limits.values():
-            if limit_info.remaining < 0.05 * limit_info.limit:
-                max_delay = max(max_delay, (limit_info.reset_at - now).total_seconds())
-        return max_delay if max_delay > 0 else None
+            time_until = limit_info.estimated_time_until_balance(
+                math.ceil(TARGET_RATE_LIMIT_RESOURCE_FRACT * limit_info.limit)
+            )
+            if time_until is not None:
+                max_wait = max(max_wait, time_until)
+        return max_wait if max_wait > 0 else 2.0 ** min(attempt, 4)
```
aaron-siegel: Include some jitter in the exponential backoff?

Also maybe define constants for the "magic numbers" `2.0` and `4` instead of hardcoding them?

pixeltable/env.py
```diff
@@ -1047,7 +1048,10 @@ class PackageInfo:
     version: list[int] | None = None  # installed version, as a list of components (such as [3,0,2] for "3.0.2")
 
 
-TIME_FORMAT = '%H:%M.%S %f'
+TIME_FORMAT = '%m/%d/%Y, %H:%M:%S.%f %Z'
```
aaron-siegel: In that case, do we really need some bespoke format here (versus just the usual `str()` output of a `datetime`?)

pixeltable/env.py
```diff
@@ -1135,18 +1146,44 @@ def debug_str(self) -> str:
...
+        if self.remaining >= target_remaining:
+            return 0
+        if self.recorded_at >= self.reset_at:
+            return 0
+        if self.limit < target_remaining:
+            return None
+
+        # Estimate resource refill rate based on the recorded state and timestamps. Assumes linear refill.
+        refill_rate = (self.limit - self.remaining) / (self.reset_at - self.recorded_at).total_seconds()
+        if refill_rate <= 0:
```
aaron-siegel: Isn't this impossible since the numerator and denominator must both be positive if we reach this point?

pixeltable/env.py
sergey-mkhitaryan: true enough. reverted TIME_FORMAT and stopped using it

pixeltable/env.py
sergey-mkhitaryan: I think that is correct, replaced with an assertion

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -167,31 +152,30 @@ def _get_request_resources(self, request: FnCallArgs) -> dict[str, int]:
             constant_kwargs, batch_kwargs = request.pxt_fn.create_batch_kwargs(batch_kwargs)
             return self.pool_info.get_request_resources(**constant_kwargs, **batch_kwargs)
 
-    def _check_resource_limits(self, request_resources: dict[str, int]) -> env.RateLimitInfo | None:
-        """Returns the most depleted resource, relative to its limit, or None if all resources are within limits"""
-        candidates: list[tuple[env.RateLimitInfo, float]] = []  # (info, relative remaining)
+    def _sleep_duration_for_resources(self, request_resources: dict[str, int]) -> float:
```
mkornacker: Is that basically a resource delay (or resource refill delay)? (Ie, a good name would describe what it is, not what the caller does with it.)

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -167,31 +152,30 @@ def _get_request_resources(self, request: FnCallArgs) -> dict[str, int]:
...
-            f'check_resource_limits({request_resources}): '
-            f'most_depleted={most_depleted[0].resource}, rel_remaining={most_depleted[1]}'
-        )
-        # 0.05: leave some headroom, we don't have perfect information
-        if most_depleted[1] < 0.05:
-            return most_depleted[0]
-        return None
+            # Note: usage is an estimated cost of the new request before it is executed, and it may be way off (for
+            # example, if max tokens is unspecified for an openAI request).
+            time_until = info.estimated_time_until_balance(
```
mkornacker: Are we not using self.est_usage anymore?

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -226,13 +210,16 @@ async def _exec(self, request: FnCallArgs, exec_ctx: ExecCtx, num_retries: int,
...
 
             if num_retries < self.MAX_RETRIES:
-                retry_delay = self.pool_info.get_retry_delay(exc)
+                retry_delay = self.pool_info.get_retry_delay(exc, num_retries)
                 if retry_delay is not None:
                     self.total_retried += 1
-                    _logger.debug(f'scheduler {self.resource_pool}: retrying in {retry_delay} seconds')
+                    _logger.debug(
+                        f'scheduler {self.resource_pool}: sleeping {retry_delay}s before retrying attempt {num_retries}'
+                        ' based on the information in the error'
```
mkornacker, line 221: Remove. Let's not make the log output overly chatty. :)

pixeltable/env.py
```diff
@@ -1081,16 +1086,19 @@ def is_initialized(self) -> bool:
     def reset(self) -> None:
         self.resource_limits.clear()
 
-    def record(self, reset_exc: bool = False, **kwargs: Any) -> None:
-        now = datetime.datetime.now(tz=datetime.timezone.utc)
+    def record(self, request_timestamp: datetime.datetime, reset_exc: bool = False, **kwargs: Any) -> None:
```
mkornacker: For brevity, 'timestamp' is often abbreviated ts, ie, request_ts

pixeltable/env.py
```diff
@@ -1099,23 +1107,27 @@ def record(self, reset_exc: bool = False, **kwargs: Any) -> None:
...
         for limit_info in self.resource_limits.values():
-            if limit_info.remaining < 0.05 * limit_info.limit:
-                max_delay = max(max_delay, (limit_info.reset_at - now).total_seconds())
-        return max_delay if max_delay > 0 else None
+            time_until = limit_info.estimated_time_until_balance(
+                math.ceil(TARGET_RATE_LIMIT_RESOURCE_FRACT * limit_info.limit)
+            )
+            if time_until is not None:
+                max_wait = max(max_wait, time_until)
+        return max_wait if max_wait > 0 else exponential_backoff(attempt)
```
mkornacker: Why bake in exponential back-off here? The idea was for the subclass to make that call (and return None here).

pixeltable/env.py
```diff
@@ -1135,18 +1147,43 @@ def debug_str(self) -> str:
         )
 
     def update(self, recorded_at: datetime.datetime, limit: int, remaining: int, reset_at: datetime.datetime) -> None:
-        # we always update everything, even though responses may come back out-of-order: we can't use reset_at to
```
mkornacker, line 1138: Why remove that comment? Explain (in a comment) why this doesn't apply anymore/what assumptions you're making.

pixeltable/env.py
```diff
@@ -1135,18 +1147,43 @@ def debug_str(self) -> str:
...
-        # determine order, because it doesn't increase monotonically (the reset duration shortens as output_tokens
-        # are freed up - going from max to actual)
+        if self.recorded_at > recorded_at:
```
mkornacker: self.recorded_at/recorded_at has changed its meaning (so the name should change).

pixeltable/env.py
```diff
@@ -1135,18 +1147,43 @@ def debug_str(self) -> str:
...
         self.limit = limit
         self.remaining = remaining
-        reset_delta = reset_at - self.reset_at
         self.reset_at = reset_at
-        # TODO: remove
-        _logger.debug(
-            f'Update {self.resource} rate limit: rem={self.remaining} reset={self.reset_at.strftime(TIME_FORMAT)} '
-            f'reset_delta={reset_delta.total_seconds()} recorded_delta={(self.reset_at - recorded_at).total_seconds()}'
+
+    def estimated_time_until_balance(self, target_remaining: int) -> float | None:
```
mkornacker: Based on the comment below, it looks like you're returning the estimated resource refill delay (which might be a more intuitive name).

pixeltable/exec/expr_eval/schedulers.py
sergey-mkhitaryan: are you sure? while the scheduler logs exceptions, there's nothing else than this line that logs the time to sleep, which may be important to understand what's going on. besides, it's our goal to minimize retries, so we should not be getting here very often.

aaron-siegel (approved):

PR #913, opened 2025-11-04
Manage `additional_md` from Catalog, rather than TableVersion

aaron-siegel:
`additional_md` can't be managed from TableVersion because named snapshots don't have TableVersion instances.

With this change, it's now managed directly from `Catalog`. In addition, `additional_md` is moved to a separate column in the metadata tables (rather than a field of `md`) so that updates to ordinary, version-specific metadata in `md` don't interfere with it.

Also included is a working unit test for push/pull. (Hooray!)

The existing `md.additional_md` fields are kept in place for now so that Cloud can continue to use them until a migration can be orchestrated.

### DIFF
--- a/pixeltable/metadata/schema.py
+++ b/pixeltable/metadata/schema.py
@@ -68,6 +68,7 @@ class SystemInfo(Base):
     """A single-row table that contains system-wide metadata."""
 
     __tablename__ = 'systeminfo'
+
     dummy = sql.Column(Integer, primary_key=True, default=0, nullable=False)
     md = sql.Column(JSONB, nullable=False)  # SystemInfoMd
 
@@ -266,6 +267,7 @@ class TableVersionMd:
 
 class TableVersion(Base):
     __tablename__ = 'tableversions'
+
     tbl_id: orm.Mapped[uuid.UUID] = orm.mapped_column(
         UUID(as_uuid=True), ForeignKey('tables.id'), primary_key=True, nullable=False
     )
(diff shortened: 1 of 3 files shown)


pixeltable/metadata/schema.py
```diff
@@ -87,6 +87,7 @@ class Dir(Base):
     )
     parent_id: orm.Mapped[uuid.UUID] = orm.mapped_column(UUID(as_uuid=True), ForeignKey('dirs.id'), nullable=True)
     md: orm.Mapped[dict[str, Any]] = orm.mapped_column(JSONB, nullable=False)  # DirMd
+    additional_md: orm.Mapped[dict[str, Any]] = orm.mapped_column(JSONB, nullable=False, default=dict)
```
mkornacker, line 91: Can you also add a comment to the additional_md fields in the <>Md structs?
`# deprecated; TODO: remove`

pixeltable/share/packager.py
```diff
@@ -423,7 +423,11 @@ def restore(self, bundle_path: Path) -> pxt.Table:
                     _logger.info(f'Importing table {tv.name!r}.')
                     self.__import_table(self.tmp_dir, tv, md)
 
-            return cat.get_table_by_id(UUID(tbl_md[0].tbl_md.tbl_id))
+            tbl = cat.get_table_by_id(UUID(tbl_md[0].tbl_md.tbl_id))
+            if pxt_uri is not None:
+                # Set pxt_uri for the newly created table
```
mkornacker, line 428: assert that this is not None for a new replica?

pixeltable/metadata/schema.py
aaron-siegel: Done.

pixeltable/share/packager.py
aaron-siegel: It won't be defined if the replica is being restored locally in a unit test.

We could potentially mock up some URIs in the tests in the future but I don't see a lot of value in that (those tests are meant to validate the package/restore catalog functionality; there are other tests that validate the protocol, including the presence of pxt_uri)

PR #916, opened 2025-11-05
`create_store_tbls: bool` option in Catalog.create_replica()

aaron-siegel:
Pixeltable cloud needs to be able to create replica metadata instances without actually creating the underlying store table.

This commit adds a `create_store_tbls` option to `Catalog.create_replica()`. We need the option to be selective on a per-table basis, since in the future we'll have both hosted tables (which do have underlying store tables) and replicas (which don't) living side by side, potentially in the same catalog.

There are various other ways to achieve this (e.g., have a flag in table metadata that controls whether it has an underlying store table; or have a flag in the environment that disables store tables specifically for replicas). This PR seems like the most straightforward solution, but we could consider those alternatives also.

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -117,7 +117,9 @@ class Table(SchemaObject):
         return op()
 
     def _get_metadata(self) -> TableMetadata:
-        columns = self._tbl_version_path.columns()
+        tvp = self._tbl_version_path
+        tv = tvp.tbl_version.get()
+        columns = tvp.columns()
         column_info: dict[str, ColumnMetadata] = {}
         for col in columns:
             column_info[col.name] = ColumnMetadata(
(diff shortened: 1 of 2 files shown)


pixeltable/catalog/table.py
```diff
@@ -130,8 +132,7 @@ def _get_metadata(self) -> TableMetadata:
                 computed_with=col.value_expr.display_str(inline=False) if col.value_expr is not None else None,
                 defined_in=col.get_tbl().name,
             )
-        # Pure snapshots have no indices
-        indices = self._tbl_version.get().idxs_by_name.values() if self._tbl_version is not None else {}
+        indices = tv.idxs_by_name.values()
```
mkornacker, line 135: Is this intentional? This will show indices from the underlying base table of a pure snapshot.

PR #917, opened 2025-11-06
Introduce "anchor tables" in TableVersion(Handle) for live replicas; working pull()

aaron-siegel:
Introduces an `anchor_tbl_id` parameter to `TableVersion` and `TableVersionHandle`. There are now three types of `TableVersion` instances:
- live tables (effective_version == None, anchor_tbl_id == None)
- snapshots (effective_version != None, anchor_tbl_id == None)
- live replicas (effective_version == None, anchor_tbl_id != None)

Details on the semantics are in the `TableVersion` docstring.

`get_table('my_replica')` now returns a `Table` instance whose `TableVersionPath` is anchored to the latest non-fragment version of `'my_replica'`. This will correctly track changes after doing a `pull()`.

Snapshot references can still be retrieved by (say) `get_table('my_replica:720')`, just as for non-replica tables.

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -1135,8 +1135,8 @@ class Catalog:
         """
         assert self.in_write_xact
 
-        # Acquire locks for any tables in the ancestor hierarchy that might already exist (base table last).
-        for ancestor_md in md:
+        # Acquire locks for any tables in the ancestor hierarchy that might already exist (base table first).
+        for ancestor_md in md[::-1]:  # base table first
             self._acquire_tbl_lock(for_write=True, tbl_id=UUID(ancestor_md.tbl_md.tbl_id), raise_if_not_exists=False)
 
         tbl_id = UUID(md[0].tbl_md.tbl_id)


pixeltable/catalog/table_version.py
```diff
@@ -82,6 +82,16 @@ class TableVersion:
 
     Only TableVersion and Catalog interact directly with stored metadata. Everything else needs to go through these
     two classes.
+
+    TableVersions come in three "flavors" depending on the `effective_version` and `anchor_tbl_id` settings:
+    - if both are None, it's a live table that tracks the latest version
```
mkornacker: More directly: it's version is TableMd.current_version

pixeltable/catalog/table_version.py
```diff
@@ -82,6 +82,16 @@ class TableVersion:
...
+    - if `effective_version` is defined, it's a snapshot of a specific version
```
mkornacker: And the version is effective_version.

pixeltable/exprs/column_ref.py
```diff
@@ -314,7 +314,7 @@ def get_column_id(cls, d: dict) -> catalog.QColumnId:
     def get_column(cls, d: dict) -> catalog.Column:
         tbl_id, version, col_id = UUID(d['tbl_id']), d['tbl_version'], d['col_id']
         # validate_initialized=False: this gets called as part of TableVersion.init()
-        tbl_version = catalog.Catalog.get().get_tbl_version(tbl_id, version, validate_initialized=False)
+        tbl_version = catalog.Catalog.get().get_tbl_version(tbl_id, version, None, validate_initialized=False)
```
mkornacker: Leave todo to enable ColumnRefs to local replica columns. Maybe even put an assert into init() that makes sure we're not seeing that kind of column (otherwise you might end up with an incorrect query result, rather than a failure).

pixeltable/catalog/table_version.py
```diff
@@ -63,6 +63,12 @@ def is_pure_snapshot(self) -> bool:
         )
 
 
+class TableVersionKey(NamedTuple):
```
mkornacker: Move into globals.py to reduce chance of circular imports?

pixeltable/catalog/table_version.py
```diff
@@ -142,22 +157,23 @@ class IndexInfo:
...
+        key: TableVersionKey,
         tbl_md: schema.TableMd,
         version_md: schema.TableVersionMd,
-        effective_version: int | None,
         schema_version_md: schema.TableSchemaVersionMd,
         mutable_views: list[TableVersionHandle],
         base_path: 'TableVersionPath' | None = None,
         base: TableVersionHandle | None = None,
     ):
+        assert key.effective_version is None or key.anchor_tbl_id is None
```
mkornacker: Can we put these in the key c'tor?

pixeltable/catalog/table_version.py
```diff
@@ -256,9 +266,9 @@ def create_initial_md(
 
         tbl_id = uuid.uuid4()
         tbl_id_str = str(tbl_id)
-        tbl_handle = TableVersionHandle(tbl_id, None)
-        column_ids = itertools.count(0)
-        index_ids = itertools.count(0)
+        tbl_handle = TableVersionHandle(TableVersionKey(tbl_id, None, None))
+        column_ids = itertools.count()
```
mkornacker, line 293: Providing an explicit starting value was deliberate (it makes the code more readable and communicates an intention).

pixeltable/catalog/table_version.py
```diff
@@ -548,6 +551,11 @@ def _init_schema(self) -> None:
             tvp = Catalog.get().construct_tvp(
                 self.id, self.effective_version, self.tbl_md.ancestor_ids, self.version_md.created_at
             )
+        elif self.anchor_tbl_id is not None:
+            # for replica TableVersion instances, we also need to retarget the value_exprs, this time to the
+            # "anchored" TableVerisonPath.
```
mkornacker: TableVersionPath

pixeltable/catalog/table_version.py
```diff
@@ -82,16 +88,25 @@ class TableVersion:
 
     Only TableVersion and Catalog interact directly with stored metadata. Everything else needs to go through these
     two classes.
+
+    TableVersions come in three "flavors" depending on the `effective_version` and `anchor_tbl_id` settings:
+    - if both are None, it's a live table that tracks the latest version
+    - if `effective_version` is defined, it's a snapshot of a specific version
+    - if `anchor_tbl_id` is defined, it's a replica table that is "anchored" to the given table, in the following
```
mkornacker, line 117: This needs more commentary. See comment for path on line 110, which needs to be expanded.

pixeltable/catalog/table_version_handle.py
```diff
@@ -57,22 +68,32 @@ def get(self) -> TableVersion:
...
+                assert self._tbl_version.key == self.key
         if self.effective_version is None:
             tvs = list(Catalog.get()._tbl_versions.values())
-            assert self._tbl_version in tvs
+            assert self._tbl_version in tvs, self._tbl_version
         return self._tbl_version
 
     def as_dict(self) -> dict:
-        return {'id': str(self.id), 'effective_version': self.effective_version}
+        return {
```
mkornacker: Maybe just have a TableVersionKey.as_dict()?

pixeltable/catalog/catalog.py
```diff
@@ -527,7 +537,10 @@ def _acquire_path_locks(
             check_pending_ops=check_pending_ops,
         )
         # update cache
-        _ = self.get_tbl_version(path_handles[0].id, path_handles[0].effective_version, validate_initialized=True)
+        _ = self.get_tbl_version(
+            TableVersionKey(path_handles[0].id, path_handles[0].effective_version, path_handles[0].anchor_tbl_id),
```
mkornacker: why not just get_tbl_version(path_handles[0].key, ...)?

pixeltable/catalog/catalog.py
```diff
@@ -584,10 +597,10 @@ def _acquire_tbl_lock(
         if for_write and not tbl_md.is_mutable:
             return None  # nothing to lock
 
-        effective_version = tbl_md.current_version if tbl_md.is_snapshot else None
+        key = TableVersionKey(tbl_id, tbl_md.current_version if tbl_md.is_snapshot else None, None)
```
mkornacker, line 593: Why None? We could have two concurrent pull().

pixeltable/catalog/catalog.py
```diff
@@ -639,7 +652,9 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> None:
 
                     if op.needs_xact:
                         tv = self.get_tbl_version(
-                            tbl_id, tbl_version, check_pending_ops=False, validate_initialized=True
+                            TableVersionKey(tbl_id, tbl_version, None),
```
mkornacker, line 648: None?

pixeltable/exprs/rowid_ref.py
```diff
@@ -77,7 +78,11 @@ def __repr__(self) -> str:
         # check if this is the pos column of a component view
         from pixeltable import store
 
-        tbl = self.tbl.get() if self.tbl is not None else catalog.Catalog.get().get_tbl_version(self.tbl_id, None)
+        tbl = (
+            self.tbl.get()
+            if self.tbl is not None
+            else catalog.Catalog.get().get_tbl_version(TableVersionKey(self.tbl_id, None, None))
```
mkornacker, line 84: So this won't work on replicas?

pixeltable/catalog/table_version.py
aaron-siegel: `itertools.count()` is the intended style; otherwise there wouldn't be a default!

The explicit `0` is idiosyncratic to trained eyes: it's akin to typing (say) `range(0, 10)` or `enumerate(var, start=0)`. We don't include the explicit `0` in those other cases, anywhere in the code base; the intention is implied. Explain why this should be any different?

pixeltable/catalog/table_version.py
aaron-siegel: clarified

pixeltable/catalog/table_version.py
aaron-siegel: clarified

pixeltable/catalog/table_version.py
aaron-siegel: `NamedTuple` doesn't give user-level control over the constructor; it's literally just a tuple with types.

pixeltable/catalog/table_version.py
aaron-siegel: fixed

pixeltable/catalog/table_version_handle.py
aaron-siegel: Done.

pixeltable/exprs/rowid_ref.py
aaron-siegel: It should still work for replicas, though of course if we have only a `UUID` and not a full handle it is going to limit the accuracy of the repr.

The issue is that when RowidRef is serialized we only store the UUID, not the full handle. Do you happen to remember why that is? Deserialization is the only place we need to construct a RowidRef without a proper handle.

We could change this to serialize the TableVersionKey instead of just the UUID, which would fix this problem, but require a schema migration (or some in place logic to accept both serialized forms)

pixeltable/catalog/table_version.py
aaron-siegel: I changed the comment on line 110. If there's still something you think is missing, can you be a little more specific? Thanks.

pixeltable/catalog/table_version.py
aaron-siegel: I kind of think it belongs here (or maybe in TableVersionHandle), but I can move it if you feel strongly

pixeltable/catalog/catalog.py
aaron-siegel: Fixed.

pixeltable/catalog/catalog.py
aaron-siegel: That shouldn't matter? I don't think the return value of `_acquire_tbl_lock` ever gets used for anything that's sensitive to the value of `anchor_tbl_id`. Unless there's something I'm not understanding here.

pixeltable/catalog/catalog.py
aaron-siegel: Why not? I don't think we'll ever see a replica here?

pixeltable/catalog/catalog.py
mkornacker: The semantics become very difficult to follow if all of a sudden this TV instance is a valid stand-in for the one with the anchor set. There's also a chance of divergent metadata if you have both this instance and the instance with the anchor set both (accidentally?) get used in the same transaction and be present in _tbl_versions at the same time.

Weakening the semantics will make it easier for bugs to creep in in the future, even if this code happens to work in this context right now.

Also, I don't agree with your argument that due to the way the return value is used, this distinction doesn't matter. That would effectively mean that in order to understand the semantics of this function, you also need to understand the semantics of all of its call sites, ie, there are no abstractions.

pixeltable/catalog/catalog.py
```diff
@@ -232,22 +234,30 @@ def validate(self) -> None:
                 f'snapshot_id={tbl_version.id} mutable_views={tbl_version.mutable_views}'
             )
 
-            if tbl_version.is_view and tbl_version.is_mutable and tbl_version.is_validated:
+            assert anchor_tbl_id is None or tbl_version.is_replica
```
mkornacker, line 237: This should be: is_replica == (anchor_tbl_id is not None)

pixeltable/catalog/catalog.py
mkornacker: That is true right now, but won't be in the future.

Add an assertion that checks is_replica.

pixeltable/catalog/catalog.py
```diff
@@ -685,10 +699,9 @@ def _debug_str(self) -> str:
 
     def _get_mutable_tree(self, tbl_id: UUID) -> set[UUID]:
         """Returns ids of all tables that form the tree of mutable views starting at tbl_id; includes the root."""
-        assert (tbl_id, None) in self._tbl_versions, (
-            f'({tbl_id}, None) not in {self._tbl_versions.keys()}\n{self._debug_str()}'
-        )
-        tv = self.get_tbl_version(tbl_id, None, validate_initialized=True)
+        key = TableVersionKey(tbl_id, None, None)
```
mkornacker, line 701: Add assert !is_replica.

This will break for views on replicas.

pixeltable/catalog/catalog.py
```diff
@@ -1355,7 +1368,9 @@ def drop_table(self, path: Path, if_not_exists: IfNotExistsParam, force: bool) -
 
         self._drop_tbl(tbl, force=force, is_replace=False)
 
-    def _drop_tbl(self, tbl: Table | TableVersionPath, force: bool, is_replace: bool) -> None:
+    def _drop_tbl(
+        self, tbl: Table | TableVersionPath, force: bool, is_replace: bool, is_replica: bool | None = None
```
mkornacker: Why do you need pass that as a parameter?

pixeltable/catalog/catalog.py
```diff
@@ -1847,34 +1879,82 @@ def _collect_tbl_history(self, tbl_id: UUID, n: int | None) -> list[TableVersion
...
+        if row is None:
+            return None
+        assert isinstance(row[0], dict)
+        return schema.md_from_dict(schema.TableVersionMd, row[0])
+
+    def load_tbl_md(self, key: TableVersionKey) -> TableVersionCompleteMd:
         """
         Loads metadata from the store for a given table UUID and version.
         """
+        assert key.effective_version is None or key.anchor_tbl_id is None
```
mkornacker: Move these checks into TableVersionKey.__new__

pixeltable/catalog/table.py
```diff
@@ -132,7 +132,9 @@ def _get_metadata(self) -> TableMetadata:
                 computed_with=col.value_expr.display_str(inline=False) if col.value_expr is not None else None,
                 defined_in=col.get_tbl().name,
             )
-        indices = tv.idxs_by_name.values()
+
+        # Pure snapshots have no indices
+        indices = {} if tv.tbl_md.is_pure_snapshot else tv.idxs_by_name.values()
```
mkornacker: Why is this needed? idxs_by_name only includes visible indices.

pixeltable/catalog/catalog.py
aaron-siegel: Not true; sometimes we need to instantiate TableVersion instances with `anchor_tbl_id is None` and `tbl_version.is_replica`. They're just not queryable.

pixeltable/catalog/catalog.py
aaron-siegel: Ok, so what do you think should be done here? How do we know which anchor id to use?

pixeltable/catalog/catalog.py
aaron-siegel: Done. Also added a TODO.

pixeltable/catalog/catalog.py
aaron-siegel: Done.

pixeltable/catalog/catalog.py
aaron-siegel: Oops, good catch. It was an attempt to get `_drop_tbl()` working that ended up being unnecessary, and I forgot to remove it.

pixeltable/catalog/catalog.py
aaron-siegel: Python does not allow overriding `NamedTuple.__new__`.

pixeltable/exprs/rowid_ref.py
aaron-siegel: I added a TODO for this.

pixeltable/exprs/column_ref.py
aaron-siegel: I left a TODO. I'm not sure what the assertion would be. We do see this kind of column when metadata is loaded; but it gets retargeted before any querying is done (same as for snapshots)

pixeltable/catalog/catalog.py
aaron-siegel: (To be clear: I'm not claiming that this is the right long-term implementation, just that there isn't an actual bug that needs to block the launch. I'm also happy to make changes now, but we should discuss first to be sure we're on the same page about what they are.)

mkornacker (commented):
Flushing out a pending comment before looking at the latest changes.

pixeltable/catalog/catalog.py
mkornacker: I think it would be really valuable for catching bugs to have some validation for these keys. Either a TVK.create() -> TVK classmethod or a dataclass with slots=True.

(As an anecdotal data point, introducing Catalog.validate() made it much easier to find bugs.)

pixeltable/catalog/catalog.py
mkornacker: When is that the case? Can we avoid that? Having clearly defined semantics makes avoiding bugs much easier.

pixeltable/exprs/column_ref.py
```diff
@@ -315,6 +315,7 @@ def get_column_id(cls, d: dict) -> catalog.QColumnId:
     def get_column(cls, d: dict) -> catalog.Column:
         tbl_id, version, col_id = UUID(d['tbl_id']), d['tbl_version'], d['col_id']
         # validate_initialized=False: this gets called as part of TableVersion.init()
+        # TODO: When we have views on replicas, we will need to store anchor_tbl_id in metadata as well.
```
mkornacker, line 319: Also assert in _as_dict() that anchor_tbl_id is None.

pixeltable/catalog/catalog.py
```diff
@@ -168,10 +168,11 @@ class Catalog:
 
     _instance: Catalog | None = None
 
-    # cached TableVersion instances; key: [id, version]
+    # cached TableVersion instances; key: [id, version, anchor_tbl_id]
     # - mutable version of a table: version == None (even though TableVersion.version is set correctly)
     # - snapshot versions: records the version of the snapshot
-    _tbl_versions: dict[tuple[UUID, int | None], TableVersion]
+    # - anchored versions: records the tbl_id of the anchor table
```
mkornacker: Spell out that this is for replicas (it's unclear whether there are other applications).

pixeltable/catalog/catalog.py
```diff
@@ -232,22 +234,30 @@ def validate(self) -> None:
...
             )
 
-            if tbl_version.is_view and tbl_version.is_mutable and tbl_version.is_validated:
+            assert anchor_tbl_id is None or tbl_version.is_replica
+
+            if (
+                tbl_version.is_view
+                and tbl_version.is_mutable
+                and tbl_version.is_validated
+                and not tbl_version.is_replica
```
mkornacker: Why is the !is_replica needed?

pixeltable/catalog/catalog.py
aaron-siegel: For example, if a _system table that only has fragments becomes orphaned because its visible descendant was dropped. We need to be able to construct a TableVersion reference to that system table so that we can garbage collect it, but there is no longer anything it can anchor to.

The semantics are quite clear; if anchor_tbl_id=None then the TableVersion points to HEAD as always. Such a TableVersion will never result from the user doing a `get_table()`, but that doesn't imply there's any ambiguity.

pixeltable/catalog/catalog.py
aaron-siegel: Isn't that's what happening here (validation of the keys?) I'm really not sure what additional value would come from trying to fold it into `TableVersionKey` that justifies making it a heavier weight structure.

pixeltable/catalog/table.py
aaron-siegel: Hmm, ok. Fixed.

pixeltable/exprs/column_ref.py
aaron-siegel: Done. (Here and in RowidRef, as well.)

pixeltable/catalog/catalog.py
mkornacker: Leave a todo: fix replica concurrency

Also put that everywhere you use anchor_tbl_id=None that doesn't have an assert !is_replica.

pixeltable/catalog/catalog.py
```diff
@@ -1579,44 +1594,72 @@ def get_tbl_version(
...
+                        _logger.debug(
+                            f'reloading metadata for live table {key.tbl_id} '
+                            f'(cached/current version: {tv.version}/{current_version}, '
+                            f'cached/current view_sn: {tv.tbl_md.view_sn}/{view_sn})'
+                        )
+                        reload = True
+
+                else:
+                    # live replica table; use the anchored version
+                    anchor_tbl_md = self.head_version_md(tv.anchor_tbl_id)
```
mkornacker: anchor_tbl_md is confusing because this isn't a TableMd instance.

pixeltable/catalog/table_version.py
mkornacker: What are the semantics of 'path' for replicas?

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
aaron-siegel: Oh sure, removed. (You're right, is_mutable already implies this condition)

pixeltable/catalog/catalog.py
aaron-siegel: fixed

pixeltable/catalog/table_version.py
aaron-siegel: Added comment.

pixeltable/catalog/catalog.py
aaron-siegel: Added a TODO here. Note that right now, this line of code will never be called, because `not tbl_md.is_mutable` will cause the function to return in the immediately preceding line. This will need to change when we have views on top of replicas.

pixeltable/catalog/catalog.py
```diff
@@ -1139,6 +1135,10 @@ def create_replica(self, path: Path, md: list[TableVersionCompleteMd], create_st
         """
         assert self.in_write_xact
 
+        # Acquire locks for any tables in the ancestor hierarchy that might already exist (base table last).
+        for ancestor_md in md:
```
mkornacker: That needs to happen in reverse order, base table first (following the same protocol we use elsewhere).

pixeltable/catalog/catalog.py
mkornacker: I still think this isn't defined correctly, but let's leave that for later.

pixeltable/catalog/catalog.py

PR #926, opened 2025-11-10
Allow hyphens in table and dir names

aaron-siegel:


pixeltable/catalog/globals.py
```diff
@@ -68,14 +68,19 @@ def validated(cls, param_val: str, param_name: str) -> IfNotExistsParam:
             raise excs.Error(f'{param_name} must be one of: [{val_strs}]') from None
 
 
-def is_valid_identifier(name: str, allow_system_identifiers: bool = False) -> bool:
-    return name.isidentifier() and (allow_system_identifiers or not name.startswith('_'))
+def is_valid_identifier(name: str, *, allow_system_identifiers: bool = False, allow_hyphens: bool = False) -> bool:
```
mkornacker, line 71: Adding a separate is_valid_path_element() or something like that might be warranted here, since this isn't an identifier anymore.

aaron-siegel: But it *is* an identifier! It may not be a valid *Python* identifier, but it's certainly an identifier (which in the general sense is just a token that names an entity).

PR #930, opened 2025-11-11
PXT-862: bug fix for drop_table()

mkornacker:


tests/utils.py
```diff
@@ -663,7 +663,9 @@ def assert_img_eq(img1: PIL.Image.Image, img2: PIL.Image.Image, context: str) ->
     assert diff.getbbox() is None, context
 
 
-def reload_catalog() -> None:
+def reload_catalog(reload: bool = True) -> None:
+    if not reload:
```
aaron-siegel, line 667: This seems silly to me and unnecessarily confusing; clearer to just use a conditional in the client:
```
if do_reload_catalog:
    reload_catalog()
```

mkornacker: I find this more compact.

PR #934, opened 2025-11-13
PXT-846: FrameIterator(keyframes_only: bool)

mkornacker:


pixeltable/iterators/video.py
```diff
@@ -33,8 +33,12 @@ class FrameIterator(ComponentIterator):
         fps: Number of frames to extract per second of video. This may be a fractional value, such as 0.5.
             If omitted or set to 0.0, or if greater than the native framerate of the video,
             then the framerate of the video will be used (all frames will be extracted).
+            Cannot be used with `keyframes_only`.
```
aaron-siegel: This is a curiously incomplete condition given that it also cannot be used with `num_frames`.

Maybe it'd be clearer to put a comment in the main docstring body "At most one of `fps`, `num_frames`, or `keyframes_only` may be specified", rather than stating it separately in each arg.

pixeltable/iterators/video.py
```diff
@@ -175,15 +193,21 @@ def __next__(self) -> dict[str, Any]:
             # Compute the index of the current frame in the video based on the presentation timestamp (pts);
             # this ensures we have a canonical understanding of frame index, regardless of how we got here
             # (seek or iteration)
+            # TODO: this computation is incorrect for variable frame rate videos, we need to use the actual
```
aaron-siegel, line 193: do we have a rough-edges ticket for this?

also, nit: comma splice

tests/test_video.py
```diff
@@ -126,11 +126,42 @@ def test_fps(self, reset_db: None) -> None:
...
+        frames = pxt.create_view(
+            'frames', videos, iterator=FrameIterator.create(video=videos.video, fps=0, all_frame_attrs=True)
+        )
+
+        videos.insert(video=path)
+
+        # Verify keyframes were extracted
+        keyframes_count = keyframes.count()
+        res = frames.order_by(frames.pos).collect()
+        assert keyframes_count == sum(int(attrs['key_frame']) for attrs in res['frame_attrs'])
```
aaron-siegel: The `int()` is extraneous here (`sum(attrs['key_frame']) for attrs in ...` is sufficient)

pixeltable/iterators/video.py
mkornacker: I think this is beyond a rough edges ticket. We also need to change the signature of set_pos(). Let's discuss tomorrow, but I think this is urgent.

pixeltable/iterators/video.py
mkornacker: Done.

tests/test_video.py

PR #938, opened 2025-11-14
Changes to protocol to handle publishing existing version          

amithadke:
Adding validation for table.md to have at least one entry

### DIFF
--- a/pixeltable/share/protocol/replica.py
+++ b/pixeltable/share/protocol/replica.py
@@ -35,7 +35,7 @@ class PublishRequest(RequestBaseModel):
 
     @field_validator('md')
     @classmethod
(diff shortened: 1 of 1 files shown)


pixeltable/share/publish.py
```diff
@@ -61,6 +61,15 @@ def push_replica(
     _logger.debug(f'Sending PublishRequest: {publish_request}')
 
     response = requests.post(PIXELTABLE_API_URL, data=publish_request.model_dump_json(), headers=_api_headers())
+    if response.status_code == 204:
+        publish_response = PublishResponse.model_validate(response.json())
+        existing_table_uri = str(publish_response.table_uri)
+        _logger.info(
```
aaron-siegel: I'd argue this should go to console logger.

PR #939, opened 2025-11-14
Add Reve integration notebook

apreshill:


aaron-siegel (dismissed):
Great overview! Some comments (github won't let me do them inline with the notebook):

- "A Pixeltable runtime (local or remote) with network access to Reve" - it's not clear to me what this means.

- `# Remove the 'reve_demo' directory and its contents, if it exists`
Alternatively just do this in one step:
`pxt.create_dir('reve_demo', if_exists='replace_force')`

- `from pixeltable.functions import reve`
out of place in its own cell; move to the cell where it first gets used

- When you first add the square photo, maybe collect() on all columns to make it clear all the data from both runs is still in the table.

sergey-mkhitaryan:
Overall looks very good, some minor suggestions. Thanks!

> we will create an empty single Pixeltable table

this sounds slightly off, should be just "an empty Pixeltable table" or "a new Pixeltable table"?

> Feel free to change the prompt—here we use your solarpunk skyline description.

your?

> brighten the subject in our base street photo.

just "base photo" or "original image"?

The link at the bottom is https://docs.pixeltable.com/sdk/v0.4.24/reve#module-pixeltable-functions-reve, should probably be https://docs.pixeltable.com/sdk/latest/reve instead

mkornacker (approved):

pierrebrunelle (approved):

sergey-mkhitaryan:
A couple more comments, just my 2c, not necessarily blockers.

> A Reve account with an API key (https://app.reve.com/ → Settings → API Keys)

Looks like they've updated the website, now it's https://app.reve.com/ → API → API Keys

I got confused on step 18 because the `edit` result already has a solarpunk-style background, and I even thought that the images got mixed up, but that's indeed what Reve produces. This makes the further remixing a little bit less interesting. I liked the pipeline you used in the first version I reviewed a little bit more, because you went from the original background to no background (edit) to a specified background (remix).

On step 22 or 23, maybe select and show the original images as well? Otherwise we don't see the original with the tennis player until the recap (step 25).

apreshill:
> A couple more comments, just my 2c, not necessarily blockers.
> 
> > A Reve account with an API key (https://app.reve.com/ → Settings → API Keys)
> 
> Looks like they've updated the website, now it's https://app.reve.com/ → API → API Keys
> 
> I got confused on step 18 because the `edit` result already has a solarpunk-style background, and I even thought that the images got mixed up, but that's indeed what Reve produces. This makes the further remixing a little bit less interesting. I liked the pipeline you used in the first version I reviewed a little bit more, because you went from the original background to no background (edit) to a specified background (remix).
> 
> On step 22 or 23, maybe select and show the original images as well? Otherwise we don't see the original with the tennis player until the recap (step 25).

Just seeing this feedback @sergey-mkhitaryan - thank you! Fixing now.

PR #940, opened 2025-11-15
[PXT-645] Support more numpy dtypes for Array

sergey-mkhitaryan:
`ArrayType` still accepts pxt Types as dtype, but internally they are immediately converted to a numpy dtype. New schema version and migration that updates dtypes in the catalog.

### DIFF
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -1036,16 +1036,14 @@ class ArrayType(ColumnType):
             raise TypeError(f'Expected numpy.ndarray, got {val.__class__.__name__}')
 
         # If column type has a dtype, check if it matches
-        if self.dtype is None:
-            pass
-        elif self.dtype == np.str_:
+        if self.dtype == np.str_:
             if val.dtype.type != np.str_:
                 raise TypeError(f'Expected numpy.ndarray of dtype {self.dtype}, got numpy.ndarray of dtype {val.dtype}')
-        elif self.dtype != val.dtype:
+        elif self.dtype is not None and self.dtype != val.dtype:
             raise TypeError(f'Expected numpy.ndarray of dtype {self.dtype}, got numpy.ndarray of dtype {val.dtype}')
 
-        # If no dtype is specified, we still need to check that the dtype is one of the supported types
-        if self.dtype is None and not any(np.issubdtype(val.dtype, ndtype) for ndtype in ARRAY_SUPPORTED_NUMPY_DTYPES):
+        # Check that the dtype is one of the supported types
+        if val.dtype.type != np.str_ and val.dtype not in ARRAY_SUPPORTED_NUMPY_DTYPES:
             raise TypeError(f'Unsupported dtype {val.dtype}')
 
         # If a shape is specified, check that there's a match


tests/test_array_type.py
```diff
@@ -0,0 +1,183 @@
...
+from pixeltable.type_system import ColumnType
+
+from .utils import validate_update_status
+
+
+class TestArrayType:
+    # 2. dtypes correctly preserved on disk, i.e. loading a table restores dtype in columns correctly.
+    # TODO fix warnings like DeprecationWarning: Converting `np.generic` to a dtype is deprecated. The current result
+    # is `np.dtype(np.void)` which is not strictly correct
+    @pytest.mark.parametrize(
```
aaron-siegel: I'm a little concerned about the proliferation of test instances, which can make the pytest results harder to wrestle with. I wonder if this is a case that's better handled by a single test with a loop over the different cases, rather than `@pytest.mark.parameterize`. Just opening this up for discussion.

You can look at `test_from_python_type` in `test_types.py`, or `test_timestamp_methods` in `test_timestamp.py`, for precedents for what I have in mind.

pixeltable/type_system.py
```diff
@@ -887,27 +868,43 @@ def _to_base_str(self) -> str:
...
+    np.float32,
+    np.float64,
+]
+
+
 class ArrayType(ColumnType):
     shape: tuple[int | None, ...] | None
-    pxt_dtype: ColumnType | None
-    dtype: ColumnType.Type | None
+    pxt_dtype: ColumnType | np.dtype | None
```
aaron-siegel: There's an argument this should just be a `np.dtype | None`, and we should do a prompt conversion if given a `ColumnType` as a type argument.

pixeltable/type_system.py
```diff
@@ -918,6 +915,36 @@ def matches(self, other: ColumnType) -> bool:
     def __hash__(self) -> int:
         return hash((self._type, self.nullable, self.shape, self.dtype))
 
+    def is_supertype_of(self, other: ColumnType, ignore_nullable: bool = False) -> bool:
```
aaron-siegel: It shouldn't be necessary to implement this directly (if the default implementation isn't working, it suggests a bug in `supertype()` )

pixeltable/type_system.py
```diff
@@ -1058,12 +1104,14 @@ def to_sa_type(cls) -> sql.types.TypeEngine:
...
     def numpy_dtype(self) -> np.dtype | None:
+        if isinstance(self.dtype, np.dtype):
+            return self.dtype
         if self.dtype is None:
             return None
         if self.dtype == self.Type.INT:
             return np.dtype(np.int64)
         if self.dtype == self.Type.FLOAT:
-            return np.dtype(np.float32)
+            return np.dtype(np.float64)
```
aaron-siegel: This is a major change, which could cause backward-compatibility problems. Let's make sure our DB migration tests include some data of type `Array[pxt.Float]` (I'm not sure they do today) so that we have an appropriate backward compatibility test.

If we adopt my prior suggestion of prompt conversion of pxt types to np.dtypes, then this issue goes away; all we need then is a migration script that converts any *existing* (historical) instances of `Array[pxt.Float]` to the expected `Array[np.float32]`, and going forward we can assume dynamic resolution to `np.float64`

tests/test_array_type.py
```diff
@@ -0,0 +1,255 @@
...
+import pixeltable.exceptions as excs
+from pixeltable import type_system as ts
+from pixeltable.type_system import ArrayType, ColumnType, IntType
+
+from .utils import reload_catalog, validate_update_status
+
+
+class TestArrayType:
+    @pytest.mark.parametrize('do_reload_catalog', [False, True], ids=['no_reload_catalog', 'reload_catalog'])
+    def test_array_dtypes(self, do_reload_catalog: bool, init_env: None, reset_db: None) -> None:
```
aaron-siegel, line 16: Just put this in `test_types.py`? Do we really need to introduce a new `.py` test file just for this?

sergey-mkhitaryan: @mkornacker do you have an opinion? Aaron and I both have not very strong feelings about this. Aaron's motivation is that he wants to avoid sprawling in the codebase. My argument is that `test_array_type.py` is already significantly longer than `test_types.py` which suggests that Arrays do deserve a dedicated test suite.

pixeltable/type_system.py
```diff
@@ -927,149 +941,116 @@ def supertype(self, other: ColumnType) -> ArrayType | None:
...
-        super_dtype = self.Type.supertype(self.dtype, other.dtype, self.common_supertypes)
-        if super_dtype is None:
-            # if the dtypes are incompatible, then the supertype is a fully general array
+        # Determine the dtype of the supertype
+        super_dtype: np.dtype | None = None
+        if self.dtype is not None and other.dtype is not None:
+            # This list is ordered such that for any two supported dtypes, the first common supertype in the array
+            # will be their lowest common supertype.
+            for dtype in ARRAY_SUPPORTED_NUMPY_DTYPES:
+                if np.can_cast(self.dtype, dtype) and np.can_cast(other.dtype, dtype):
```
mkornacker: I don't think this works. 'cast' for numpy means data conversion (ie, reformatting the data). If I have a column with int32 arrays and I'm adding int64 arrays, the type of the column can't be Array(np.int64), because an application that expects to get np.int64 when querying the column would blow up. The type would need to be np.int_ (or leave out the dtype altogether).

pixeltable/type_system.py
```diff
@@ -927,149 +941,116 @@ def supertype(self, other: ColumnType) -> ArrayType | None:
...
         result = super()._as_dict()
         shape_as_list = None if self.shape is None else list(self.shape)
-        dtype_value = None if self.dtype is None else self.dtype.value
-        result.update(shape=shape_as_list, dtype=dtype_value)
+        result.update(shape=shape_as_list)
+
+        if self.dtype is None:
+            result.update(numpy_dtype=None)
+        elif self.dtype == np.str_:
+            # str(np.str_) would be something like '<U'
```
mkornacker: Why do we need to avoid that?

pixeltable/type_system.py
```diff
@@ -927,149 +941,116 @@ def supertype(self, other: ColumnType) -> ArrayType | None:
...
+    def _validate_literal(self, val: Any) -> None:
         if not isinstance(val, np.ndarray):
-            return False
+            raise TypeError(f'Expected numpy.ndarray, got {val.__class__.__name__}')
 
-        # If a dtype is specified, check that there's a match
-        if self.dtype is not None and not np.issubdtype(val.dtype, self.numpy_dtype()):
-            return False
+        # If column type has a dtype, check for compatibility. Note: the exact dtype match is not required. We allow
+        # literals whose dtypes can be cast to the column's.
```
mkornacker: See comment above regarding cast. I think we need to guarantee the physical format, not just whether data conversion is possible.

pixeltable/type_system.py
sergey-mkhitaryan: My understanding is that column types exist for validation on write, and that validation only looks at the data type and not the actual values. With that in mind, keeping the full np str qualifier like '<U128' would incorrectly imply that we also validate the string data (some combination of its endianness, encoding, length).

But we can also make a paradigm shift and start validating the row data, and in that case serializing the numpy str type would make total sense.

pixeltable/type_system.py
sergey-mkhitaryan: Just to be clear, you can't cast int64 to int32, but you can cast int32->int64. So, the following problematic scenario is possible:
- create table with an array[int64] column
- insert an ndarray[int32]
- read it back and get an ndarray[int32] still
So yeah, an application that expect ndarray[int64] may blow up.

The behavior of scalar columns today is that I can insert a bool into a float column. It'll get cast up, and when I read it back, I'll get `0.0` or `1.0`. With that in mind, should the array behavior also be that all values are cast, when possible, to the column type during write?

pixeltable/metadata/converters/convert_43.py
```diff
@@ -0,0 +1,46 @@
...
+    """Converts ArrayTypes by replacing legacy dtype (which was a pxt Type ID) to numpy dtype."""
+    convert_table_md(engine, substitution_fn=_substitution_fn)
+
+
+def _substitution_fn(key: str | None, value: Any) -> tuple[str | None, Any] | None:
+    if not isinstance(value, dict):
+        return None
+    if value.get('_classname', None) != 'ArrayType':
+        return None
+    if 'numpy_dtype' in value:
```
aaron-siegel, line 22: How can this be true in a v43 catalog?

pixeltable/metadata/converters/convert_43.py
```diff
@@ -0,0 +1,46 @@
...
+        return None
+    assert 'dtype' in value
+
+    updated_value = copy.deepcopy(value)
```
aaron-siegel: pretty sure it's safe to update in place.

pixeltable/type_system.py
```diff
@@ -1455,3 +1442,32 @@ def as_col_type(cls, nullable: bool) -> ColumnType:
 
 
 ALL_PIXELTABLE_TYPES = (String, Bool, Int, Float, Timestamp, Json, Array, Image, Video, Audio, Document)
+
+
+def column_type_from_np_dtype(dtype: np.dtype, nullable: bool) -> ColumnType | None:
```
aaron-siegel: I'd suggest making this a `@classmethod` of `ColumnType` (`ColumnType.from_np_dtype`)

pixeltable/type_system.py
```diff
@@ -1455,3 +1442,32 @@ def as_col_type(cls, nullable: bool) -> ColumnType:
...
+
+    if np.issubdtype(dtype, np.str_):
+        return StringType(nullable=nullable)
+
+    if np.issubdtype(dtype, np.character):
+        return StringType(nullable=nullable)
+
+    if np.issubdtype(dtype, np.datetime64):
+        unit, _ = np.datetime_data(dtype)
+        if unit in ['D', 'M', 'Y']:
```
aaron-siegel: style nit: use a tuple
`unit in ('D', 'M', 'Y'):`

tests/test_array_type.py
```diff
@@ -0,0 +1,388 @@
...
+
+import pixeltable as pxt
+import pixeltable.exceptions as excs
+from pixeltable import type_system as ts
+from pixeltable.type_system import ArrayType, ColumnType, IntType
+
+from .utils import reload_catalog, validate_update_status
+
+
+class TestArrayType:
```
aaron-siegel, line 14: I didn't read the test code super carefully so it's possible this is already happening, but let's make sure we test the output of `str(Array[...])` for each of the supported array dtypes.

pixeltable/metadata/converters/convert_43.py
sergey-mkhitaryan: `convert_24.py` has the following code:
```
        if 'path' in v:
            assert 'signature' not in v
            f = resolve_symbol(__substitute_path(v['path']))
            assert isinstance(f, func.Function)
            v['signature'] = f.signatures[0].as_dict()
        return k, v
```
It creates a substitution that already includes a `numpy_dtype`, i.e. `v['signature']` is: 
`{'return_type': {'_classname': 'ArrayType', 'nullable': False, 'shape': [...], 'numpy_dtype': 'float32'}, 'parameters': [{...}, {...}], 'is_batched': True}`

And that of course happens before `convert_43.py` runs.

tests/test_array_type.py
sergey-mkhitaryan: `test_repr` should be it

pixeltable/type_system.py
sergey-mkhitaryan: PTAL

aaron-siegel (approved):

PR #941, opened 2025-11-16
PXT-871: == None filter doesn't work correctly on an array column

mkornacker:


pixeltable/exprs/is_null.py
```diff
@@ -23,6 +24,11 @@ def _equals(self, other: IsNull) -> bool:
         return True
 
     def sql_expr(self, sql_elements: SqlElementCache) -> sql.ColumnElement | None:
+        c = self.components[0]
+        if isinstance(c, ColumnRef) and c.col.stores_external_array():
+            # we also need to check CellMd.file_urls for null
+            e = sql.and_(c.col.sa_cellmd_col.op('->>')('file_urls') == None, c.col.sa_col == None)
```
aaron-siegel: I think
```python
c.col.sa_cellmd_col['file_urls']
```
is the preferred syntax. It actually compiles to different SQL and the bracket notation is the "modern" (Postgres 14+) one. There are performance implications too (`->>` in particular converts everything to text - you may need to put a sql.Null() on the rhs in place of None if you use bracket notation - I'm not 100% sure)

mkornacker: Good catch. Done.

PR #942, opened 2025-11-17
Circularity detection in view creation with if_exists='replace'

aaron-siegel:
Checks for a specific scenario in which:
- A view `my_view` is created
- A subview of `my_view` is created with the identical name `my_view`, using if_exists='replace'

If this situation not detected, it leads to permanent catalog corruption.

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -1125,7 +1125,7 @@ class Catalog:
                 )
                 assert result.rowcount == 1, result.rowcount
 
-            existing = self._handle_path_collision(path, View, is_snapshot, if_exists)
+            existing = self._handle_path_collision(path, View, is_snapshot, if_exists, base=base)
             if existing is not None:
                 assert isinstance(existing, View)
                 return existing._id
@@ -2378,7 +2378,13 @@ class Catalog:
             _logger.info(f'Added root directory record for user: {user!r}')
 
     def _handle_path_collision(
-        self, path: Path, expected_obj_type: type[SchemaObject], expected_snapshot: bool, if_exists: IfExistsParam
+        self,
+        path: Path,
+        expected_obj_type: type[SchemaObject],
+        expected_snapshot: bool,
+        if_exists: IfExistsParam,
+        *,
+        base: TableVersionPath | None = None,
     ) -> SchemaObject | None:
         obj, _, _ = self._prepare_dir_op(add_dir_path=path.parent, add_name=path.name)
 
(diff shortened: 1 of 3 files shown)


tests/test_view.py
```diff
@@ -1226,3 +1226,36 @@ def test_update_base_column(self, reset_db: None) -> None:
...
+        with pytest.raises(
+            pxt.Error, match=r"Cannot use if_exists='replace' with the same name as one of the view's own ancestors."
+        ):
+            _ = pxt.create_view('my_view_1', v3, if_exists='replace')
+        v1.collect()
+        v2.collect()
+        v3.collect()
+        pxt.drop_table('my_view_3')
+        pxt.drop_table('my_view_2')
+        pxt.drop_table('my_view_1')
```
sergey-mkhitaryan, line 1260: do we need a case in which we try to replace an ancestor table (as opposed to a view) with a view?

pixeltable/catalog/catalog.py
```diff
@@ -2404,6 +2410,16 @@ def _handle_path_collision(
         if if_exists == IfExistsParam.IGNORE:
             return obj
 
+        assert if_exists in (IfExistsParam.REPLACE, IfExistsParam.REPLACE_FORCE)
+
+        # Check for circularity
```
mkornacker, line 2415: Is this applicable beyond create_view()? If not, it would be easier to follow if the logic lives in create_view().

tests/test_view.py
```diff
@@ -1226,3 +1226,36 @@ def test_update_base_column(self, reset_db: None) -> None:
...
+        # The same problem also exists if there are additional tables in between: if creating a view with
+        # if_exists='replace', the name of the view cannot match any existing name in its ancestor chain.
+        t = pxt.create_table('my_tbl', {'col': pxt.Int})
+        v1 = pxt.create_view('my_view_1', t)
+        v2 = pxt.create_view('my_view_2', v1)
+        v3 = pxt.create_view('my_view_3', v2)
+        with pytest.raises(
+            pxt.Error, match=r"Cannot use if_exists='replace' with the same name as one of the view's own ancestors."
+        ):
+            _ = pxt.create_view('my_view_1', v3, if_exists='replace')
```
mkornacker, line 1254: Additional test case where you're creating a view 'my_tbl'?

tests/test_view.py
aaron-siegel: That's not possible currently due to PXT-877. (In fact I tried to implement exactly the test you suggest, and that's how I discovered and filed PXT-877.)

tests/test_view.py
aaron-siegel: See my response to Sergey (who asked the identical question).

pixeltable/catalog/catalog.py
aaron-siegel: I agree, but that would need a larger refactoring, since we need to do the circularity check *after* resolving the name collision but *before* dropping the table that's being replaced, and both those operations happen in `_handle_path_collision()`.

I guess in theory we could safely drop the table first as long as the circularity exception is thrown inside the same transaction, but it feels wrong to rely on that (and it would increase the load on the database, invite additional concurrency conflicts etc., all unnecessarily).

PR #944, opened 2025-11-18
Add TableVersionMd.from_dict and update publish to use objects instead of dicts

amithadke:
Starting with TableVersionMd 
This way helps server to rely on pydantic ser/de.


pixeltable/utils/pydantic_serializable.py
```diff
@@ -0,0 +1,71 @@
+"""
```
amithadke: note that we need this for cloud orchestration, I pull this off PXT-703 PR

aaron-siegel: Maybe just put it inside `pydantic.py` rather than introduce a separate utils module

pixeltable/utils/pydantic_serializable.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    @abstractmethod
+    def as_dict(self) -> dict[str, Any]:
+        """Convert object to dictionary representation."""
+        pass
+
+    @classmethod
+    @abstractmethod
+    def from_dict(cls: type[T], data: dict[str, Any]) -> T:
```
aaron-siegel: Why the need for the `cls` parameter? Would the following signature be simpler?
```python
def from_dict(data: dict[str, Any]) -> Self:
```

pixeltable/utils/pydantic_serializable.py
```diff
@@ -0,0 +1,71 @@
...
+
+class PydanticSerializable(ABC):
+    """
+    Abstract base class that provides Pydantic core schema using as_dict() and from_dict() methods.
+    """
+
+    @abstractmethod
+    def as_dict(self) -> dict[str, Any]:
+        """Convert object to dictionary representation."""
+        pass
```
aaron-siegel: Unnecessary `pass`

pixeltable/utils/pydantic_serializable.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @classmethod
+    @abstractmethod
+    def from_dict(cls: type[T], data: dict[str, Any]) -> T:
+        """Create object from dictionary representation."""
+        pass
```
aaron-siegel: Unnecessary `pass`

pixeltable/utils/pydantic_serializable.py
```diff
@@ -0,0 +1,71 @@
...
+            """Deserialize object using from_dict() class method."""
+            if isinstance(data, cls):
+                # Already the correct type, return as-is
+                return data
+            if isinstance(data, dict):
+                # Deserialize from dict
+                return cls.from_dict(data)
+            raise TypeError(f'Expected {cls.__name__} or dict, got {type(data).__name__}')
+
+        def serialize_to_dict(obj: Any, handler: Any) -> dict[str, Any]:
```
aaron-siegel: Same question, is it possible to be more type-friendly here?
```python
def serialize_to_dict(obj: Self, handler: GetCoreSchemaHandler) -> dict[str, Any]:
```

pixeltable/utils/pydantic_serializable.py
```diff
@@ -0,0 +1,71 @@
...
+
+        Args:
+            source_type: The source type being processed
+            handler: Pydantic's schema handler
+
+        Returns:
+            CoreSchema that uses as_dict() and from_dict() for serialization
+        """
+
+        def deserialize_from_dict(data: Any) -> Any:
```
aaron-siegel: Is it possible to be more type-friendly here?
```python
def deserialize_from_dict(data: Self | dict) -> Self:
```
or will that cause problems with the Pydantic interface?

pixeltable/utils/pydantic_serializable.py
```diff
@@ -0,0 +1,71 @@
+"""
+Pydantic serialization for Pixeltable classes.
+
+This base class provides automatic Pydantic serialization/deserialization using the as_dict() and from_dict() methods.
```
aaron-siegel: So this class functions as a mixin that enables non-Pydantic objects to appear inside Pydantic models and be serialized along with them?

I wonder if it wouldn't be lighter weight to just implement a custom `@pydantic.model_serializer` in `PublishRequest` that knows what to do with the `TableVersionMd` instance. Is that a feasible alternative? Curious what are the reasons to prefer the present approach.

aaron-siegel (changes requested):

pixeltable/utils/pydantic_serializable.py
amithadke: we will need this for plenty of other pixeltable classes that will go over the wire in next phase. 
if we don't use this approach, in next phase PXT-703 we have to write custom code to detect as_dict/from_dict and insert them in dyanmically generated pydantic models.

In fact use of mixin came up as part of review of PXT-703, makes it easy to put pixeltable classes on the wire. We can use pydantic typeadaptor than writing custom code that recurses thr data structures.

pixeltable/utils/pydantic_serializable.py
amithadke: let me check

pixeltable/utils/pydantic_serializable.py
amithadke: moved pydantic serializer out of this PR

aaron-siegel (approved):

PR #946, opened 2025-11-18
Add ".quatro" to gitignore

sergey-mkhitaryan:


aaron-siegel (changes requested):
This should no longer be necessary. Merge the latest main and docs generation should not create a .quarto (if it does, that’s a bug).

PR #951, opened 2025-11-19
[PXT-886] Fix a bug in RateLimitsScheduler's error handling

sergey-mkhitaryan:


pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -207,23 +207,23 @@ async def _exec(self, request: FnCallArgs, exec_ctx: ExecCtx, num_retries: int,
...
-                    _logger.debug(
-                        f'scheduler {self.resource_pool}: sleeping {retry_delay}s before retrying attempt {num_retries}'
-                        ' based on the information in the error'
-                    )
-                    await asyncio.sleep(retry_delay)
-                    self.queue.put_nowait(self.QueueItem(request, num_retries + 1, exec_ctx))
-                    return
+
+            # If pool info is available, attempt to retry based on the resource information
+            # Pool info may not be available yet if the exception occurred before the UDF set it
```
mkornacker, line 212: Does this basically mean that the very first request gets an exception? Shouldn't the exception headers contain the rate limits info? Can we simplify this (by having some guarantee that pool_info is available)?

PR #952, opened 2025-11-19
Introduce internal `LocalTable` class and move catalog-dependent functionality there

aaron-siegel:
This change makes `pxt.Table` into an abstract class. Currently the only subclass is `LocalTable`. The documentation and interface has been moved into `pxt.Table` and the catalog-related functionality into `LocalTable`.

More refactoring will likely be needed as we implement `RemoteTable`; this change is just a first step.

### DIFF
--- a/pixeltable/_query.py
+++ b/pixeltable/_query.py
@@ -870,7 +870,9 @@ class Query:
 
             >>> query = t.join(d, on=(t.d1 == d.pk1) & (t.d2 == d.pk2), how='left')
         """
-        assert isinstance(other, catalog.table.LocalTable)  # TODO Print a real error message when we have RemoteTables
+        # TODO: we shouldn't be requiring a LocalTable here; we need to refactor the join() implementation to support
+        #     remote tables (abstract Table objects without a TableVersionPath)
+        assert isinstance(other, catalog.table.LocalTable)
 
         if self.sample_clause is not None:
             raise excs.Error('join() cannot be used with sample()')
@@ -902,6 +904,7 @@ class Query:
         """Add a group-by clause to this Query.
 
         Variants:
+
         - group_by(base_tbl): group a component view by their respective base table rows
         - group_by(expr1, expr2, expr3): group by the given expressions
 


pixeltable/catalog/table.py
```diff
@@ -1427,29 +720,974 @@ def update(
...
+            Report only the most recent 5 changes to the table:
+
+            >>> tbl.history(n=5)
+        """
+        versions = self.get_versions(n)
+        assert len(versions) > 0
+        return pd.DataFrame([list(v.values()) for v in versions], columns=list(versions[0].keys()))
+
+
+class LocalTable(Table):
```
sergey-mkhitaryan, line 951: We may as well use this opportunity to move `LocalTable` to a separate file, now that we are screwing with git history of this code anyway. `table.py` on `main` is almost 2k LOC.

pixeltable/catalog/table.py
```diff
@@ -60,146 +60,40 @@ class Table(SchemaObject):
...
-    def _get_version(self) -> int:
-        """Return the version of this table. Used by tests to ascertain version changes."""
-        return self._tbl_version_path.version()
-
-    def _get_pxt_uri(self) -> str | None:
-        with catalog.Catalog.get().begin_xact(tbl_id=self._id):
-            return catalog.Catalog.get().get_additional_md(self._id).get('pxt_uri')
-
-    def __hash__(self) -> int:
-        return hash(self._tbl_version_path.tbl_id)
```
mkornacker, line 179: Why not leave this here?

pixeltable/_query.py
```diff
@@ -874,6 +874,8 @@ def join(
 
             >>> query = t.join(d, on=(t.d1 == d.pk1) & (t.d2 == d.pk2), how='left')
         """
+        assert isinstance(other, catalog.table.LocalTable)  # TODO Print a real error message when we have RemoteTables
```
mkornacker: We need to check for this in collect()/_exec(). join() is just assembling the query.

pixeltable/exprs/function_call.py
```diff
@@ -146,6 +146,7 @@ def __init__(
         self.id = self._create_id()
 
     def _create_rowid_refs(self, tbl: catalog.Table) -> list[Expr]:
+        assert isinstance(tbl, catalog.table.LocalTable)
```
mkornacker, line 149: Why?

pixeltable/catalog/table.py
aaron-siegel: It can't stay here because a Table doesn't necessarily have a TableVersionPath any more. Since Table instances are unique (managed by the catalog) it shouldn't matter anyway - right?

More generally, if we're not implementing `eq` then what is the purpose in implementing `hash`? If the instances are unique it's a no-op, and if they're not, all it does is reduce hashing efficiency by increasing the chance of collision.

pixeltable/catalog/table.py
aaron-siegel: @mkornacker Do you have any opinion here?

pixeltable/exprs/function_call.py
aaron-siegel: Because it needs to access _tbl_version_path

pixeltable/_query.py
aaron-siegel: I understand that, but right now the implementation of `Query.join()` relies on calling `other._tbl_version_path`. More refactoring is needed to disentangle that (in particular the join implementation relies on accessing TableVersionPaths in a few places); I'd propose we leave that for round 2. I've added a TODO for this.

pixeltable/catalog/table.py
mkornacker: Probably good to break up a bit.

pixeltable/exprs/function_call.py
mkornacker: Leave another todo.

(I think the correct way to do this is to create a new abstraction for the metadata-only aspects of TableVersionPath.)

pixeltable/catalog/view.py
```diff
@@ -28,7 +28,7 @@
 _logger = logging.getLogger('pixeltable')
 
 
-class View(Table):
+class View(LocalTable):
```
mkornacker, line 31: Is this indicating that we're going to have a separate RemoteView?

PR #953, opened 2025-11-20
Use database schemas and search_path for test isolation in parallel runs

amithadke:


pixeltable/env.py
```diff
@@ -571,10 +571,28 @@ def _init_metadata(self) -> None:
...
+        # Parse the database URL
+        db_url = sql.make_url(self.db_url)
+
+        # Get existing options and parse them
+        # Query parameters can be strings or tuples (if multiple values exist)
+        existing_options_raw = db_url.query.get('options', '') if db_url.query else ''
+        existing_options = (
+            ' '.join(existing_options_raw) if isinstance(existing_options_raw, tuple) else existing_options_raw
+        )
+        option_parts = existing_options.split() if existing_options else []
```
aaron-siegel: Doesn't this just reverse the preceding operation (join the tuple only to split it again)?

Why not simply
```python
option_parts = list(existing_options_raw) if isinstance(existing_options_raw, tuple) else existing_options_raw.split()
```
If that doesn't work please add a code comment explaining why we need the 2-step parsing

tests/conftest.py
```diff
@@ -61,8 +62,41 @@ def pytest_runtest_teardown(item: pytest.Item) -> None:
     _logger.info(f'Finished Pixeltable test: {current_test}')
 
 
+def _setup_external_db_schema(worker_id: int | str) -> str:
+    schema_name = f'test_{worker_id}_{os.getpid()}'
```
aaron-siegel: pid() is not necessarily globally unique right?

Why not a uuid?

tests/conftest.py
```diff
@@ -61,8 +62,41 @@ def pytest_runtest_teardown(item: pytest.Item) -> None:
...
+    # Each worker process has its own os.environ, so this is process-safe
+    db_url = sql.make_url(original_connect_str)
+    search_path_option = f'-c search_path={schema_name},public'
+
+    query_dict = dict(db_url.query) if db_url.query else {}
+    # Query parameters can be strings or tuples (if multiple values exist)
+    existing_options_raw = query_dict.get('options', '')
+    existing_options = (
+        ' '.join(existing_options_raw) if isinstance(existing_options_raw, tuple) else existing_options_raw
+    )
```
aaron-siegel: See preceding comment.

Either way, there is code duplication here. Can it be consolidated?

tests/conftest.py
```diff
@@ -61,8 +62,41 @@ def pytest_runtest_teardown(item: pytest.Item) -> None:
     _logger.info(f'Finished Pixeltable test: {current_test}')
 
 
+def _setup_external_db_schema(worker_id: int | str) -> str:
```
aaron-siegel: Nit: `_set_up_external_db_schema` ("setup" is a noun, "set up" is a verb)

tests/conftest.py
```diff
@@ -61,8 +62,41 @@ def pytest_runtest_teardown(item: pytest.Item) -> None:
...
+
+    modified_url = db_url.set(query=query_dict)
+    os.environ['PIXELTABLE_DB_CONNECT_STR'] = modified_url.render_as_string(hide_password=False)
+    _logger.info(f'Created schema and configured connection string with search_path: {schema_name}')
+    return schema_name
+
+
 @pytest.fixture(scope='session')
-def init_env(tmp_path_factory: pytest.TempPathFactory, worker_id: int) -> None:
+def init_env(tmp_path_factory: pytest.TempPathFactory, worker_id: int) -> Iterator[str | None]:
```
aaron-siegel: I'm a little skittish about changing the signature of this fixture. It's used in multiple places. Another option would be to keep the return type as `None`, and set a global `SCHEMA_NAME` in conftest.py (a peer to `DO_RERUN`, etc) to keep track of it.

Although looking at the code below, it's not clear why we actually even need it?

tests/conftest.py
amithadke: schema_name = f'test_{worker_id}_{uuid.uuid4().hex[:16]}' using this now

tests/conftest.py
amithadke: i was passing it to clean_db earlier will remove it not needed

amithadke: do u want to keep yield in init_env and follow up deletion of schema?

amithadke: using pytest_sessionfinish

aaron-siegel: ok. deleting it after the yield is fine too

aaron-siegel: i kind of liked it better in init_env after the yield :)

it keeps the _SCHEMA_NAME localized since it isn't really needed anywhere else.

aaron-siegel (approved):

PR #955, opened 2025-11-20
PXT-872 Support count() with sample and group by clause.

amithadke:


amithadke:
# TODO: create an exec.CountNode and change this to create_count_plan()
just saw this, i will try this way

pixeltable/_query.py
```diff
@@ -554,14 +555,19 @@ def count(self) -> int:
...
-            raise excs.Error('count() cannot be used with group_by()')
-
-        from pixeltable.plan import Planner
-
         with Catalog.get().begin_xact(tbl=self._first_tbl, for_write=False) as conn:
-            stmt = Planner.create_count_stmt(self._first_tbl, self.where_clause)
-            result: int = conn.execute(stmt).scalar_one()
+            # Create the query plan
+            plan = self._create_query_plan()
+            sql_node = plan.get_node(exec.SqlNode)
```
aaron-siegel: Why not put this logic into `Planner.create_count_stmt()`?

tests/test_query.py
```diff
@@ -432,6 +432,96 @@ def test_count(self, test_tbl: pxt.Table, small_img_tbl: pxt.Table) -> None:
         with pytest.raises(pxt.Error):
             _ = t.where(t.img.width > 100).count()
 
+    def test_count_with_sample(self, test_tbl: pxt.Table) -> None:
```
aaron-siegel: Move to `test_sample.py`?

pixeltable/_query.py
amithadke: doing it adding CountNode to support python filter, it will be expensive but count should work in call cases

aaron-siegel: :white_check_mark:

pixeltable/exec/sql_node.py
```diff
@@ -760,3 +760,25 @@ def _create_stmt_stratified_fraction(self, fraction_samples: float) -> sql.Selec
...
+    def __init__(self, row_builder: exprs.RowBuilder, input: SqlNode):
+        assert isinstance(input, SqlNode)
+        self.input_cte, input_col_map = input.to_cte()
+        sql_elements = exprs.SqlElementCache(input_col_map)
+        super().__init__(input.tbl, row_builder, list(row_builder.output_exprs), columns=[], sql_elements=sql_elements)
+        self.ctx = input.ctx if input is not None else None
+
+    def _create_stmt(self) -> sql.Select:
+        """Create a SELECT COUNT(*) statement from the input CTE."""
+        return sql.select(sql.func.count().label('all_count')).select_from(self.input_cte)
```
mkornacker: How about we just roll that into create_count_stmt().

tests/test_sample.py
```diff
@@ -366,3 +368,80 @@ def test_sample_iterator(self, reset_db: None) -> None:
...
         v_rows = v.count()
         print(f'total rows: {v_rows}')
+
+    def test_count(self, test_tbl: pxt.Table) -> None:
+        """Test that count() correctly returns the number of sampled rows."""
+        t = test_tbl
+        # Add duplicate c1 values for stratified sampling tests
+        existing_c1_values = [row['c1'] for row in t.select(t.c1).distinct().collect()]
+
+        # Add 4 more rows per distinct c1 value to get 5 total per value (total 400 rows)
```
mkornacker: total 500 rows?

tests/test_sample.py
```diff
@@ -366,3 +368,80 @@ def test_sample_iterator(self, reset_db: None) -> None:
...
+        cnt = t.sample(n=10).count()
+        assert cnt == 10
+
+        # Sample more rows than total rows in the table
+        cnt = t.sample(n=1000).count()
+        assert cnt == total_rows
+
+        # Test count() with sample with fraction
+        cnt = t.sample(fraction=0.5).count()
+        assert 200 <= cnt <= 300  # Should be approximately 50% of total rows
```
mkornacker: I think we're using +/-25% somewhere as an appropriate range. You could phrase this assertion in terms of 0.5 * total_count +/- window.

tests/test_query.py
```diff
@@ -427,11 +427,26 @@ def test_count(self, test_tbl: pxt.Table, small_img_tbl: pxt.Table) -> None:
         cnt = t.where(t.c2 < 10).count()
         assert cnt == 10
 
-        # for now, count() doesn't work with non-SQL Where clauses
+        # count() does not support Python-only filters
         t = small_img_tbl
-        with pytest.raises(pxt.Error):
+        with pytest.raises(pxt.Error, match='count\\(\\) cannot be used with Python-only filters'):
```
aaron-siegel: Use a raw string to avoid excessive double-escaping
`match=r'count`...

aaron-siegel (approved):

PR #956, opened 2025-11-22
Progress tracker

mkornacker:
- replaces tqdm progress bars with more detailed progress indicators using rich
- removes tracking of 'computed expressions', since this doesn't really map onto anything the user cares about

### DIFF
--- a/docs/release/integrations/cloud-storage.mdx
+++ b/docs/release/integrations/cloud-storage.mdx
@@ -286,7 +286,7 @@ Destinations are resolved in this order:
 <Tabs>
   <Tab title="URI Format">
     ```
-    https://fly.storage.tigris.dev/bucket-name/prefix/
+    https://t3.storage.dev/bucket-name/prefix/
     ```
   </Tab>
   <Tab title="Authentication">
@@ -311,7 +311,7 @@ Destinations are resolved in this order:
 
     t.add_computed_column(
         thumbnail=t.file.resize((128, 128)),
-        destination='https://fly.storage.tigris.dev/my-bucket/thumbnails/'
+        destination='https://t3.storage.dev/my-bucket/thumbnails/'
     )
     ```
   </Tab>


pixeltable/_query.py
```diff
@@ -343,13 +340,10 @@ async def _aexec(self) -> AsyncIterator[exprs.DataRow]:
         This function must not modify the state of the Query, otherwise it breaks dataset caching.
         """
         plan = self._create_query_plan()
-        plan.open()
-        try:
+        with plan:
             async for row_batch in plan:
                 for row in row_batch:
```
aaron-siegel, line 347: `yield from row_batch` while we're looking at this part of the code?

pixeltable/env.py
```diff
@@ -265,6 +265,18 @@ def is_local(self) -> bool:
...
         return self._db_server is not None
 
+    def is_interactive(self) -> bool:
+        """Return True if running in an interactive environment."""
+        if getattr(builtins, '__IPYTHON__', False):
+            return True
+        # Python interactive shell
+        if hasattr(sys, 'ps1'):
+            return True
+        # for script execution, __main__ has __file__
```
aaron-siegel, line 280: Are we sure we want to treat this as "interactive"? It's not so clear to me.

pixeltable/env.py
```diff
@@ -930,12 +941,19 @@ def __init_spacy(self) -> None:
         dependency, we install it programmatically here. This should cause no problems, since the model packages
         have no sub-dependencies (in fact, this is how spaCy normally manages its model resources).
         """
+        import contextlib
```
aaron-siegel: Move these to top of file

pixeltable/env.py
```diff
@@ -930,12 +941,19 @@ def __init_spacy(self) -> None:
...
+        import io
+
         import spacy
         from spacy.cli.download import download
 
         spacy_model = 'en_core_web_sm'
         self._logger.info(f'Ensuring spaCy model is installed: {spacy_model}')
-        download(spacy_model)
+
+        # prevent download() from hanging due to its progress bar, which conflicts with our use of Rich Progress
```
aaron-siegel, line 1005: We should really just get rid of the spacy auto-downloader (I could take that on as a future rough-edges ticket)

pixeltable/exec/exec_context.py
```diff
@@ -1,42 +1,210 @@
...
+        def _get_display_unit(self) -> tuple[int, str]:
+            # scale to human-readable unit
+            scale: int
+            unit: str
+            if self.total < 2**10:
+                scale = 0
+                unit = 'B'
+            elif self.total < 2**20:
+                scale = 10
+                unit = 'KB'
```
aaron-siegel: `kB` not `KB`

pixeltable/exec/exec_context.py
```diff
@@ -1,42 +1,210 @@
...
+                self.task_id = self.ctx.progress.add_task(self.desc, rate='0/s', unit=self.unit)
+                self.last_update_ts = time.monotonic()  # start now
+
+        def _get_display_unit(self) -> tuple[int, str]:
+            # scale to human-readable unit
+            scale: int
+            unit: str
+            if self.total < 2**10:
+                scale = 0
+                unit = 'B'
```
aaron-siegel: Suggestion: use `'B '` (with a space at the end) for consistent alignment

pixeltable/exec/exec_context.py
```diff
@@ -1,42 +1,210 @@
...
+
+def _print_number(val: float | int) -> str:
+    if isinstance(val, int):
+        return str(val)
+    else:
+        assert isinstance(val, float)
+        if val < 1.0:
+            return f'{val:.3f}'
+        elif val < 10.0:
+            return f'{val:.2f}'
```
aaron-siegel: Suggestion: `f'{val:.2f} '` (with an extra space) to align the decimal point. Likewise add extra spaces below to keep the decimal point aligned in all cases. This will avoid distracting "blips" when the order of magnitude changes, and also avoid hard-to-read tables with misaligned decimal points.

pixeltable/exec/exec_context.py
```diff
@@ -1,42 +1,210 @@
...
+        elif val < 100.0:
+            return f'{val:.1f}'
+        else:
+            return f'{int(val)}'
```
aaron-siegel: This will round down and give inconsistent behavior with the preceding ones. For example, `val = 39.99` will print as `f'{val:.1f}' == '40.0'` (rounding up), but `val = 399.9` will print as `{int(val)} = '399'` (rounding down).

Use `f'{val:.0f}'` instead of `{int(val)}` for consistent behavior. (Even better, `f'{val:.0f}.   '` to print a clarifying decimal point and three spaces for alignment, as noted above.)

pixeltable/store.py
```diff
@@ -349,38 +346,52 @@ def load_column(self, col: catalog.Column, exec_plan: ExecNode, abort_on_exc: bo
 
         try:
             table_rows: list[tuple[Any]] = []
+            with exec_plan:
+                progress_reporter = (
+                    exec_plan.ctx.add_progress_reporter(
+                        desc=f'Column values written (table {self.tbl_version.get().name!r})', unit='rows'
```
aaron-siegel: Does this mean we'll get a separate progress line for each table written to? I feel like that could get pretty distracting in deep hierarchies where we end up writing to a lot of tables. Are we sure this is what we want?

pyproject.toml
```diff
@@ -126,6 +126,7 @@ dev = [
     "pixeltable-doctools @ git+https://github.com/pixeltable/pixeltable-doctools",
     "pycocotools>=2.0.10",
     "ipykernel>=6.27.1,<7",
+    "ipywidgets>=7.6,<9",
```
aaron-siegel: Why is this needed? What will the user see in an IPython shell if `ipywidgets` is not installed? I'm concerned b/c many users are likely to install `ipython` or `jupyter` without `ipywidgets`.

pixeltable/exec/exec_context.py
```diff
@@ -1,42 +1,210 @@
...
+        self.conn = None
         self.pk_clause = pk_clause
-        self.num_computed_exprs = num_computed_exprs
         self.ignore_errors = ignore_errors
         self.random_seed = random.randint(0, 1 << 63)
+
+    def add_progress_reporter(self, desc: str, unit: str) -> 'ExecContext.ProgressReporter':
+        """Records new ProgressReporter for the given desc/unit, or returns the existing one."""
+        assert self.progress is not None
+        key = f'{desc}_{unit}'
```
aaron-siegel: I thought we agreed on just one line per `desc` type?

pixeltable/exec/exec_context.py
```diff
@@ -1,42 +1,210 @@
...
-    batch_size: int
-    num_rows: int | None
-    conn: sql.engine.Connection | None
+    conn: sql.engine.Connection | None  # if present, use this to execute SQL queries
     pk_clause: list[sql.ClauseElement] | None
-    num_computed_exprs: int
     ignore_errors: bool
     random_seed: int  # general-purpose source of randomness with execution scope
 
+    class ProgressReporter:
```
sergey-mkhitaryan: Does it have to be a nested class? Seems like it would be more readable as a sister class in the same module. Unless nesting it gives it a special meaning.

pixeltable/exec/exec_context.py
```diff
@@ -1,42 +1,210 @@
...
+        """Create Progress object and start the timer. Idempotent."""
+        if not self.show_progress or self.progress is not None:
+            return
+
+        self.progress = Progress(
+            TextColumn('[progress.description]{task.description}', table_column=Column(min_width=40)),
+            AdaptiveNumericColumn(),
+            TextColumn('[progress.completed] {task.fields[unit]}', justify='left'),
+            # ' ',
+            # TextColumn('[progress.percentage]{task.fields[rate]}[/progress.percentage]', justify='right'),
```
sergey-mkhitaryan: please cleanup the commented out code

pixeltable/env.py
mkornacker: Which part isn't clear? Is there a better way to check for this?

pixeltable/_query.py
mkornacker: That's not possible, because it mixes sync and async.

pixeltable/env.py

pixeltable/env.py
mkornacker: Left todo.

pixeltable/exec/exec_context.py
mkornacker: Restructured.

pixeltable/exec/exec_context.py
mkornacker: Done.

pixeltable/exec/exec_context.py
mkornacker: KB is widely understood to mean 1024 bytes.

pixeltable/exec/exec_context.py
mkornacker: Restructured

pixeltable/exec/exec_context.py
mkornacker: Done.

pixeltable/exec/exec_context.py
mkornacker: Done.

pixeltable/exec/exec_context.py
mkornacker: Changed to round().

pixeltable/store.py
mkornacker: Discussed offline.

pyproject.toml
mkornacker: You'll get a warning and no Rich output.

pixeltable/env.py
aaron-siegel: Ah I think you misinterpreted my comment. I'm saying maybe we shouldn't be treating script execution as interactive.

pixeltable/exec/exec_context.py
aaron-siegel: `k` is the SI prefix, not `K`. Most unix utilities use `k` or `kB`, not `KB`.

What's the rationale for us using a different symbol than (for example) `curl`?

pixeltable/catalog/table_version.py
```diff
@@ -1033,10 +1035,13 @@ def rowids() -> Iterator[int]:
...
-        result = self._insert(
-            plan, time.time(), print_stats=print_stats, rowids=rowids(), abort_on_exc=fail_on_exception
-        )
-        return result
+        try:
+            result = self._insert(
+                plan, time.time(), print_stats=print_stats, rowids=rowids(), abort_on_exc=fail_on_exception
+            )
+            return result
+        finally:
```
aaron-siegel: This pattern is repeated a lot. Would it be better as a context manager?
```
with Env.get().record_progress():
```
or some such?

pixeltable/env.py
aaron-siegel: Closed after further discussion (now I understand...)

pixeltable/utils/progress_reporter.py
```diff
@@ -0,0 +1,106 @@
...
+    if isinstance(val, int):
+        return str(val)
+    if val < 1.0:
+        return f'{val:.3f}'
+    elif val < 10.0:
+        return f'{val:.2f} '  # trailing space for alignment
+    elif val < 100.0:
+        return f'{val:.1f}  '  # trailing spaces for alignment
+    else:
+        return f'{round(val)}'
```
aaron-siegel: Needs 4 trailing spaces also

mkornacker: Done.

PR #959, opened 2025-11-23
Fal.ai Integration

pierrebrunelle:
## Add fal.ai Integration

Adds support for fal.ai's fast inference models for image generation.

### DIFF
--- a/pixeltable/functions/fal.py
+++ b/pixeltable/functions/fal.py
@@ -12,7 +12,7 @@ from pixeltable.env import Env, register_client
 from pixeltable.utils.code import local_public_names
 
 if TYPE_CHECKING:
-    import fal_client  # type: ignore[import-untyped]
+    import fal_client
 
 
 @register_client('fal')


tests/functions/test_fal.py
```diff
@@ -0,0 +1,83 @@
...
+        )
+        t.add_computed_column(image=t.response['images'][0]['url'].astype(pxt.Image))
+        validate_update_status(t.insert(prompt='A futuristic cityscape with flying cars'), 1)
+        results = t.collect()
+        img = results['image'][0]
+        assert isinstance(img, PIL.Image.Image)
+        # fast-sdxl with square_hd should produce a 1024x1024 image
+        assert img.size == (1024, 1024)
+
+    @pytest.mark.expensive
```
aaron-siegel: Why this one marked expensive?

pierrebrunelle:
Removed the @pytest.mark.expensive marker. The test_subscribe function uses fal-ai/flux/dev which takes ~2.5 seconds to complete - not significantly more expensive than the other tests. The marker was added conservatively but testing showed it's not needed.

pierrebrunelle:
Additional commits: @aaron-siegel 

- Added missing API doc stubs for voyageai and twelvelabs to mkdocs
- Added FAL_API_KEY to mintlify configuration docs
- Added fal, voyageai, and twelvelabs to public_api.opml for SDK docs
- Fixed multimodal_embed usage in voyageai notebook (was incorrectly passing both image and text)
- Refactored voyageai notebook: split large cells for better output visibility
- Executed voyageai notebook to generate outputs

pixeltable/functions/fal.py
```diff
@@ -0,0 +1,113 @@
...
+    Env.get().require_package('fal_client')
+    client = _fal_client()
+    result = await client.run(app, arguments=input)
+    return result
+
+
+@pxt.udf(resource_pool='request-rate:fal')
+async def subscribe(input: dict[str, Any], *, app: str) -> pxt.Json:
+    """
+    Subscribe to and run a model on fal.ai with real-time updates.
```
mkornacker: Having separate run() and subscribe() makes no sense, the latter simply uses the recommended auto-polling mechanism with the queue. We only need one udf.

Also: the docstring isn't meaningful ('with real-time updates'?).

pierrebrunelle: Fixed to use subscribe under the cover but with one run udf()

pierrebrunelle:
Latest commits:
- Added missing UDFs to public_api.opml: image.thumbnail (from PR #967), audio.encode_audio, string.zfill
- Executed working-with-fal notebook with outputs

PR #960, opened 2025-11-23
[PXT-765] Support for Office Formats as part of Document Type through MarkdownIT

pierrebrunelle:
## Add PowerPoint, Word, and Excel support via MarkItDown

**Core Integration:**
- Added `markitdown[pptx,docx,xlsx]>=0.1.3` dependency
- Extended `DocumentType.DocumentFormat` enum with `PPTX`, `DOCX`, `XLSX`
- Created `get_office_handle()` that converts office files to markdown AST via MarkItDown
- Added `_markitdown_sections()` to `DocumentSplitter` for processing converted documents
- Registered `markitdown` as optional package in `env.py`

**Design Decision:**
Office documents are converted to markdown on load and processed through the existing markdown AST pipeline. This provides consistent behavior across all document types and leverages existing splitting/metadata logic.

**Bug Fix:**
Fixed edge case in `_token_chunks()` where Unicode decode failures could cause assertion errors on certain document content.

**Testing:**
- Added 3 tests in `test_document.py` covering insert, handle creation, and splitting
- Tested with 26,971-line CalPy presentation (splits into 2,369 token chunks)

**Documentation:**
- Created cookbook recipe: `doc-extract-text-from-office-files.ipynb`
- Updated cookbook README, CONTRIBUTING.md, and STYLE_GUIDE.md

### Supported Formats
- PowerPoint: `.pptx`, `.ppt`
- Word: `.docx`, `.doc`  
- Excel: `.xlsx`, `.xls`

### DIFF
--- a/pixeltable/functions/bedrock.py
+++ b/pixeltable/functions/bedrock.py
@@ -5,13 +5,19 @@ Provides integration with AWS Bedrock for accessing various foundation models
 including Anthropic Claude, Amazon Titan, and other providers.
 """
 
+import asyncio
+import json
 import logging
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
+
+import numpy as np
+import PIL.Image
 
 import pixeltable as pxt
-from pixeltable import env, exprs
+from pixeltable import env, exprs, type_system as ts
 from pixeltable.func import Tools
 from pixeltable.utils.code import local_public_names
+from pixeltable.utils.image import to_base64
 
 if TYPE_CHECKING:
     from botocore.client import BaseClient
@@ -31,8 +37,17 @@ def _bedrock_client() -> Any:
     return env.Env.get().get_client('bedrock')
 
 
+# Default embedding dimensions for models
+_embedding_dimensions: dict[str, int] = {
+    'amazon.titan-embed-text-v1': 1536,
+    'amazon.titan-embed-text-v2:0': 1024,
+    'amazon.titan-embed-image-v1': 1024,
+    'amazon.nova-2-multimodal-embeddings-v1:0': 3072,
+}
+
+
 @pxt.udf
-def converse(
+async def converse(
     messages: list[dict[str, Any]],
     *,
     model_id: str,
@@ -45,7 +60,8 @@ def converse(
     Generate a conversation response.
 
     Equivalent to the AWS Bedrock `converse` API endpoint.
-    For additional details, see: <https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html>
+    For additional details, see:
+    <https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html>
 
     __Requirements:__
 
@@ -59,7 +75,7 @@ def converse(
         additional_model_request_fields: Additional inference parameters to use.
 
     For details on the optional parameters, see:
-    <https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html>
+    <https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html>
 
     Returns:
         A dictionary containing the response and other metadata.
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -71,6 +71,7 @@ dependencies = [
     "pixeltable-pgserver==0.4.0",
     "tzlocal>=5.0",
     "aiohttp>=3.9",
+    "markitdown[pptx,docx,xlsx]>=0.1.3",
```
aaron-siegel: I'd presume we do not want this is a required dependency (especially since mistune is optional and that's needed too!)

tests/test_document.py
```diff
@@ -47,6 +47,17 @@ def valid_doc_paths(self) -> list[str]:
     def invalid_doc_paths(self) -> list[str]:
         return [get_video_files()[0], get_audio_files()[0], get_image_files()[0]]
 
+    def office_doc_paths(self) -> list[str]:
+        """Return paths to office format test documents."""
+        from pathlib import Path
+
+        docs_dir = Path(__file__).parent / 'data' / 'documents'
+        office_files = []
+        for ext in ['.pptx', '.docx', '.xlsx']:
```
aaron-siegel: style nit: use a tuple
```
for ext in ('.pptx', '.docx', '.xslx')
```

tests/test_document.py
```diff
@@ -47,6 +47,17 @@ def valid_doc_paths(self) -> list[str]:
...
+            for f in docs_dir.glob(f'*{ext}'):
```
aaron-siegel: Why the loop? Why not simply
```
office_files.extend(docs_dir.glob(...))
```
Or even more concise,
```
office_files = [f for ext in (...) for f in docs_dir.glob(...)]
```

pixeltable/iterators/document.py
```diff
@@ -472,6 +480,67 @@ def _txt_sections(self) -> Iterator[DocumentSection]:
         assert self._doc_handle.txt_doc is not None
         yield DocumentSection(text=ftfy.fix_text(self._doc_handle.txt_doc), metadata=DocumentSectionMetadata())
 
+    def _markitdown_sections(self) -> Iterator[DocumentSection]:
```
aaron-siegel: This looks like pure slop to me. Why does this need to be a separate method from `_markdown_sections()`? Either reuse `_markdown_sections()` and parameterize as necessary to minimize code duplication, or add a comment to explain the reasoning why this needs to be separate.

pixeltable/utils/documents.py
```diff
@@ -1,7 +1,7 @@
 import dataclasses
 import os
 
-import bs4
+import bs4  # type: ignore[import-untyped]
```
aaron-siegel: Why?

aaron-siegel (changes requested):

pyproject.toml
pierrebrunelle: Done. Moved to dev dependencies.

tests/test_document.py
pierrebrunelle: Done.

tests/test_document.py
pierrebrunelle: Done. Simplified to single list comprehension.

pixeltable/iterators/document.py
pierrebrunelle: Removed. Now reuses _markdown_sections() with an optional md_ast parameter for the AST source.

pixeltable/utils/documents.py
pierrebrunelle: Removed. Unnecessary since we have types-beautifulsoup4 installed.

pixeltable/iterators/document.py
```diff
@@ -365,9 +373,16 @@ def process_element(el: Tag | NavigableString) -> Iterator[DocumentSection]:
         yield from process_element(self._doc_handle.bs_doc)
         yield from emit()
 
-    def _markdown_sections(self) -> Iterator[DocumentSection]:
-        """Create DocumentSections reflecting the html-specific separators"""
-        assert self._doc_handle.md_ast is not None
+    def _markdown_sections(self, md_ast: list[dict] | None = None) -> Iterator[DocumentSection]:
```
aaron-siegel: If we're refactoring this way (which does make sense to me), better to have md_ast be a required parameter and use a consistent pattern (passing in md_ast as an argument) across both kinds of markdown input.

pixeltable/iterators/document.py
```diff
@@ -365,9 +373,16 @@ def process_element(el: Tag | NavigableString) -> Iterator[DocumentSection]:
...
+        """Create DocumentSections from a markdown AST.
+
+        Args:
+            md_ast: The markdown AST to process. If None, uses self._doc_handle.md_ast.
+                    This allows reuse for office formats converted to markdown via MarkItDown.
+        """
+        if md_ast is None:
```
aaron-siegel: With the above suggestion, these three lines of code would go away.

pixeltable/utils/documents.py
```diff
@@ -16,6 +16,8 @@ class DocumentHandle:
     md_ast: dict | None = None
     pdf_doc: fitz.Document | None = None
     txt_doc: str | None = None
+    # For office formats converted to markdown via markitdown
+    markitdown_md_ast: dict | None = None
```
aaron-siegel: Alternatively, is there any reason to keep this separate rather than just using md_ast in all cases?

tests/test_document.py
```diff
@@ -47,6 +47,13 @@ def valid_doc_paths(self) -> list[str]:
     def invalid_doc_paths(self) -> list[str]:
         return [get_video_files()[0], get_audio_files()[0], get_image_files()[0]]
 
+    def office_doc_paths(self) -> list[str]:
+        """Return paths to office format test documents."""
+        from pathlib import Path
```
aaron-siegel: Move to top of file (there is no justification for an inline import here).

tests/test_document.py
```diff
@@ -47,6 +47,13 @@ def valid_doc_paths(self) -> list[str]:
     def invalid_doc_paths(self) -> list[str]:
         return [get_video_files()[0], get_audio_files()[0], get_image_files()[0]]
 
+    def office_doc_paths(self) -> list[str]:
```
aaron-siegel: Missing are test documents from all supported formats.

pixeltable/iterators/document.py
pierrebrunelle: Done. md_ast is now a required parameter.

pixeltable/iterators/document.py
pierrebrunelle: Done.

pixeltable/utils/documents.py
pierrebrunelle: Done. Removed markitdown_md_ast, now using md_ast for all markdown AST cases.

tests/test_document.py
pierrebrunelle: Done.

tests/test_document.py
pierrebrunelle: Added test_presentation.pptx, test_document.docx, and test_spreadsheet.xlsx.

pixeltable/utils/documents.py
```diff
@@ -61,14 +67,14 @@ def get_html_handle(path: str) -> bs4.BeautifulSoup:
...
-def get_markdown_handle(path: str) -> dict:
+def get_markdown_handle(path: str) -> list:
     Env.get().require_package('mistune', [3, 0])
     import mistune
 
     with open(path, encoding='utf8') as file:
         text = file.read()
     md_ast = mistune.create_markdown(renderer=None)
-    return md_ast(text)
+    return md_ast(text)  # type: ignore[return-value]
```
aaron-siegel: What's going on here? We changed the return type of the function, and then added a `# type: ignore` because the return type is invalid? This needs an explanation at least

pixeltable/utils/documents.py
```diff
@@ -93,3 +99,21 @@ def get_txt(path: str) -> str:
...
+
+    # Convert office document to markdown using MarkItDown
+    md = MarkItDown(enable_plugins=False)
+    result = md.convert(path)
+    markdown_text = result.text_content
+
+    # Parse the markdown into an AST for consistent handling
+    md_parser = mistune.create_markdown(renderer=None)
+    # When renderer=None, mistune returns the AST as a list
+    return md_parser(markdown_text)  # type: ignore[return-value]
```
aaron-siegel: Same, why not just use the correct return type?

tests/test_document.py
```diff
@@ -372,3 +385,93 @@ def test_doc_splitter_images(self, reset_db: None) -> None:
...
+                assert handle.format == ts.DocumentType.DocumentFormat.PPTX, path
+                assert handle.md_ast is not None, path
+            elif extension in ('.docx', '.doc'):
+                assert handle.format == ts.DocumentType.DocumentFormat.DOCX, path
+                assert handle.md_ast is not None, path
+            elif extension in ('.xlsx', '.xls'):
+                assert handle.format == ts.DocumentType.DocumentFormat.XLSX, path
+                assert handle.md_ast is not None, path
+
+    def test_office_formats_splitter(self, reset_db: None) -> None:
```
aaron-siegel: Why is this necessary? Don't we already have a `test_document_splitter` that picks up everything in the `tests/documents` folder? Is there a particular reason why we need to special-case office documents with separate tests?

tests/test_document.py
```diff
@@ -47,6 +48,15 @@ def valid_doc_paths(self) -> list[str]:
     def invalid_doc_paths(self) -> list[str]:
         return [get_video_files()[0], get_audio_files()[0], get_image_files()[0]]
 
+    def office_doc_paths(self) -> list[str]:
```
aaron-siegel: Next question :) Why is this special-cased rather than just using the existing `utils.get_documents()` (which already picks up everything in tests/documents) and incorporating into the existing document tests? (Also see the below comment)

pierrebrunelle:
Addressed all review comments:

**documents.py type annotations**: Changed `# type: ignore[return-value]` to proper `cast(list, ...)` with explanatory comments. The issue is that mistune's `create_markdown(renderer=None)` is typed to return `str | list` but actually returns `list[dict]` when renderer is None.

**Redundant office tests**: Removed the separate `office_doc_paths()` method and the three office-specific test functions (`test_office_formats_insert`, `test_office_formats_get_handle`, `test_office_formats_splitter`). The office test documents are already in `tests/data/documents/` and are picked up by `get_documents()`, so `test_doc_splitter` now covers them. Added `skip_test_if_not_installed('markitdown')` to handle environments without the package.

pierrebrunelle (commented):
Done.

pixeltable/utils/documents.py
pierrebrunelle: Done. Changed to `cast(list, ...)` with a comment explaining that mistune types it as `str | list` but actually returns `list[dict]` when `renderer=None`.

pixeltable/utils/documents.py
pierrebrunelle: Done. Same fix - using `cast(list, ...)`.

tests/test_document.py
pierrebrunelle: Removed. The office test documents are in `tests/data/documents/` and picked up by `get_documents()`, so `test_doc_splitter` handles them. Added `skip_test_if_not_installed('markitdown')` to handle environments without the package.

tests/test_document.py
pierrebrunelle: Done. Removed `office_doc_paths()` - the office documents are picked up by `get_documents()` and tested through the existing setup.

pixeltable/utils/documents.py
```diff
@@ -93,3 +101,19 @@ def get_txt(path: str) -> str:
...
+    Env.get().require_package('markitdown')
+    import mistune
+    from markitdown import MarkItDown
+
+    md = MarkItDown(enable_plugins=False)
+    result = md.convert(path)
+    markdown_text = result.text_content
+
+    md_parser = mistune.create_markdown(renderer=None)
+    # renderer=None returns AST as list[dict], but mistune types it as str | list
```
aaron-siegel: You can also do it this way (usually preferred to using `cast()`):
```
result = md_parser(markdown_text)
assert isinstance(result, list)
return result
```
and mypy will understand how to interpret it.

aaron-siegel (approved):

pixeltable/utils/documents.py
pierrebrunelle: Done. Updated both functions to use `isinstance` assertion.

pierrebrunelle: done

PR #961, opened 2025-11-24
[PXT-892] Support variable framerate in FrameIterator

aaron-siegel:
Includes upgrades to iterator execution to include stored outputs of the row being sought in calls to `set_pos()`.

### DIFF
--- a/pixeltable/iterators/video.py
+++ b/pixeltable/iterators/video.py
@@ -25,17 +25,20 @@ _logger = logging.getLogger('pixeltable')
 
 class FrameIterator(ComponentIterator):
     """
-    Iterator over frames of a video. At most one of `fps`, `num_frames` or `keyframes_only` may be specified. If `fps`
+    Iterator over frames of a video. At most one of `fps`, `num_frames`, or `keyframes_only` may be specified. If `fps`
     is specified, then frames will be extracted at the specified rate (frames per second). If `num_frames` is specified,
-    then the exact number of frames will be extracted. If neither is specified, then all frames will be extracted. The
-    first frame of the video will always be extracted, and the remaining frames will be spaced as evenly as possible.
+    then the exact number of frames will be extracted. If neither is specified, then all frames will be extracted.
+
+    If `fps` or `num_frames` is large enough to exceed the native framerate of the video, then all frames will be
+    extracted. (Frames will never be duplicated; the maximum number of frames extracted is the total number of frames
+    in the video.)
 
     Args:
-        fps: Number of frames to extract per second of video. This may be a fractional value, such as 0.5.
-            If omitted or set to 0.0, or if greater than the native framerate of the video,
-            then the framerate of the video will be used (all frames will be extracted).
-        num_frames: Exact number of frames to extract. The frames will be spaced as evenly as possible. If
-            `num_frames` is greater than the number of frames in the video, all frames will be extracted.
+        fps: Number of frames to extract per second of video. This may be a fractional value, such as `0.5` (one frame
+            per two seconds). The first frame of the video will always be extracted.
+        num_frames: Exact number of frames to extract. The frames will be spaced as evenly as possible: the video will
+            be divided into `num_frames` evenly spaced intervals, and the midpoint of each interval will be used for
+            frame extraction.
         keyframes_only: If True, only extract keyframes.
         all_frame_attrs:
             If True, outputs a `pxt.Json` column `frame_attrs` with the following `pyav`-provided attributes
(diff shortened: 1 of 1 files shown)


pixeltable/exec/sql_node.py
```diff
@@ -128,11 +126,17 @@ def __init__(
...
+        # iter args and stored outputs
         for iter_arg in row_builder.unstored_iter_args.values():
             sql_subexprs = iter_arg.subexprs(filter=self.sql_elements.contains, traverse_matches=False)
-            for e in sql_subexprs:
-                self.select_list.add(e)
+            self.select_list.update(sql_subexprs)
+        # We query for unstored outputs only if we're not loading a view; when we're loading a view, we are populating
+        # those columns, so we need to keep them out of the select list. This isn't a problem, because view loads never
+        # need to call set_pos().
+        if not row_builder.for_view_load:
```
mkornacker, line 141: I don't understand this condition. A view load plan runs a query against the base table (and can never reference iterator outputs).

pixeltable/exprs/column_ref.py
```diff
@@ -94,8 +96,10 @@ def __init__(
             self.components = [non_validating_col_ref]
         self.id = self._create_id()
 
-    def set_iter_arg_ctx(self, iter_arg_ctx: RowBuilder.EvalCtx) -> None:
+    def set_iter_arg_ctx(self, iter_arg_ctx: RowBuilder.EvalCtx, iter_outputs: list[ColumnRef]) -> None:
         self.iter_arg_ctx = iter_arg_ctx
+        self.iter_outputs = iter_outputs
+        self.components.extend(iter_outputs)
```
mkornacker, line 105: Add explanatory comment (it's unusual for a columnref to have components).

pixeltable/exec/sql_node.py
aaron-siegel: That appears to be false, at least the way it's implemented today, according to the observed behavior of Pixeltable during a view load. (If I remove this condition, there are multiple test failures.)

pixeltable/exprs/column_ref.py
aaron-siegel: Done.

pixeltable/iterators/video.py
```diff
@@ -87,6 +88,12 @@ def __init__(
         if int(fps is not None) + int(num_frames is not None) + int(keyframes_only) > 1:
             raise excs.Error('At most one of `fps`, `num_frames` or `keyframes_only` may be specified')
 
+        if fps is not None and fps < 0.0:
```
mkornacker, line 95: Also add this to output_schema() for constant args (or at least add TODO, but ideally we'd fix this). It doesn't make sense to accept the view definition only to generate runtime errors for every input row.

pixeltable/iterators/video.py
```diff
@@ -96,54 +103,54 @@ def __init__(
...
-            # Extract all frames
-            self.frames_to_extract = None
+                # TODO: Anything we can do here? Other methods of determining the duration are expensive and
+                #     not so appropriate for an iterator initializer.
+                self.video_duration = None
+
+        if self.video_duration is None and (self.fps is not None or self.num_frames is not None):
+            raise excs.Error(f'Could not determine duration of video: {video}')
+
+        # If self.fps or self.num_frames is specified, we cannot rely on knowing in advance which frame positions will
```
mkornacker, line 133: Instead of precomputing the times here, can't we just do that in next()? That way, we don't need the duration, which is an artificial requirement.

pixeltable/iterators/video.py
```diff
@@ -164,86 +171,118 @@ def output_schema(cls, *args: Any, **kwargs: Any) -> tuple[dict[str, ts.ColumnTy
...
+                # - next_extraction_time is before cur_frame_time (never skips)
+                # - next_extraction_time is after next_frame_time (always skips)
+                # - next_extraction_time is between cur_frame_time and next_frame_time (depends on which is closer)
+                if next_frame_time - next_extraction_time < next_extraction_time - cur_frame_time:
+                    self.cur_frame = next_frame
+                    continue
+
+                # We didn't skip cur_frame, so we're definitely going to be materializing it. But we also need to
+                # ensure we get the *index* correct and consistent with future seek() operations. If fps or
+                # num_frames is large, then multiple extraction times may map onto cur_frame. So we need to
```
mkornacker: How can that be, if we cap fps and num_frames at their maxima?

pixeltable/iterators/video.py
aaron-siegel: Done.

pixeltable/iterators/video.py
aaron-siegel: Switched from using a list of times to a next_time + step. We still need duration with `num_frames` in order to calculate step, but this makes `fps` more resilient against videos with incomplete metadata.

PR #963, opened 2025-11-25
[PXT-875] performance test for openai endpoints

sergey-mkhitaryan:
Adds a recurring OpenAI performance test action that submits the results to Grafana.

Also includes a number of performance improvements.

### DIFF
--- a/pixeltable/functions/openai.py
+++ b/pixeltable/functions/openai.py
@@ -24,8 +24,8 @@ from pixeltable import env, exprs, type_system as ts
 from pixeltable.config import Config
 from pixeltable.func import Batch, Tools
 from pixeltable.utils.code import local_public_names
-from pixeltable.utils.http import set_file_descriptor_limit
 from pixeltable.utils.local_store import TempStore
+from pixeltable.utils.system import set_file_descriptor_limit
 
 if TYPE_CHECKING:
     import openai
--- a/pixeltable/utils/http.py
+++ b/pixeltable/utils/http.py
@@ -1,39 +1,8 @@
-import logging
-import os
 import re
 import time
 from http import HTTPStatus
 from random import random
 from typing import Any
-
-_logger = logging.getLogger('pixeltable')
-
-
-def set_file_descriptor_limit(preferred_limit: int) -> None:
-    """Checks and possibly updates the open file descriptor limit for the process.
-
-    Note that there may be an OS-enforced upper bound on this limit, so this function may not always succeed in setting
-    the preferred limit. It will log a warning and return normally in that case.
-    """
-    if os.name == 'nt':
-        _logger.info('Skipping FD limit adjustment for Windows')
-        return
-    import resource
-
-    soft_limit, hard_limit = resource.getrlimit(resource.RLIMIT_NOFILE)
-    _logger.info(f'Current RLIMIT_NOFILE soft limit: {soft_limit}, hard limit: {hard_limit}')
-    if soft_limit < preferred_limit and soft_limit < hard_limit:
-        new_limit = min(hard_limit, preferred_limit)
-        _logger.info(f'Setting RLIMIT_NOFILE soft limit to: {new_limit}')
-        resource.setrlimit(resource.RLIMIT_NOFILE, (new_limit, hard_limit))
-        soft_limit = new_limit
-
-    if soft_limit < preferred_limit:
-        _logger.warning(
-            f'RLIMIT_NOFILE soft limit is {soft_limit}, which is less than the preferred {preferred_limit}. '
-            'You may experience suboptimal network performance.'
-        )
-
 
 _RETRIABLE_ERROR_INDICATORS = (
     'rate limit',
--- a/pixeltable/utils/system.py
+++ b/pixeltable/utils/system.py
@@ -1,0 +1,30 @@
+import logging
+import os
+
+_logger = logging.getLogger('pixeltable')
+
+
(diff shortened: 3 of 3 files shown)


pixeltable/functions/openai.py
```diff
@@ -30,19 +30,60 @@
 
 _logger = logging.getLogger('pixeltable')
 
+# These are somewhat arbitrary numbers chosen to balance the throughput and the rate of errors on MacOS
```
mkornacker: Baking Macos/your personal environment into this file doesn't seem like a good approach.

pixeltable/functions/openai.py
```diff
@@ -30,19 +30,60 @@
...
+MAX_KEEPALIVE_CONNECTIONS = 500
+MAX_CONNECTIONS = 2000
+TIMEOUT = 60.0
+
+
+def _adjust_fd_limit() -> None:
```
mkornacker: This is a general utility and not specific to OpenAI udfs. Find a place in /utils.

Also, this should probably be done during env initialization. I would expect other clients to be affected in the same way (eg, Anthropic).

pyproject.toml
```diff
@@ -169,6 +169,7 @@ dev = [
     "azure-identity>=1.16.0",
     "twelvelabs>=1.0",
     "scenedetect>=0.6.0.1",
+    "opentelemetry-exporter-otlp>=1.38.0",
```
mkornacker, line 174: Why?

tests/test_schedulers.py
```diff
@@ -6,7 +6,7 @@
 
 from pixeltable import func
 from pixeltable.exec.expr_eval.globals import Dispatcher, ExecCtx, FnCallArgs
-from pixeltable.exec.expr_eval.schedulers import RateLimitsScheduler, RequestRateScheduler
+from pixeltable.exec.expr_eval.schedulers import RateLimitsScheduler, _is_retriable_error
```
mkornacker: It's not a good idea to test the internals of modules.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -19,6 +19,129 @@
...
+)
+RETRIABLE_HTTP_STATUSES: dict[str, int] = {
+    'TOO_MANY_REQUESTS': HTTPStatus.TOO_MANY_REQUESTS.value,
+    'SERVICE_UNAVAILABLE': HTTPStatus.SERVICE_UNAVAILABLE.value,
+    'REQUEST_TIMEOUT': HTTPStatus.REQUEST_TIMEOUT.value,
+    'GATEWAY_TIMEOUT': HTTPStatus.GATEWAY_TIMEOUT.value,
+}
+
+
+def _is_retriable_error(exc: Exception) -> tuple[bool, float | None]:
```
mkornacker: These look like general http-related utilities, maybe move into a new module in /utils?

tests/test_schedulers.py
sergey-mkhitaryan: do you have a specific suggestion here?

pyproject.toml
sergey-mkhitaryan: This is used to send the perftest results to Grafana. See `report_to_grafana` in `perftest_runner.py`

tests/test_schedulers.py
aaron-siegel: I'll chime in here (I've made this point before) that I think whitebox testing can sometimes be ok *as a supplement to* (not a replacement for) blackbox testing. Best to avoid it when feasible but sometimes it isn't; there are certain things whose user-visible impact is hard to measure (cache behavior is a classic example).

In this case I'd argue it's quite reasonable to have this as a separate check; it'd be difficult to engineer real use cases that reliably reproduce all the different edge cases.

.github/workflows/performance-tests.yml
```diff
@@ -0,0 +1,59 @@
+name: Performance Tests
```
aaron-siegel: I suggest this be a job in stress-tests rather than a separate workflow.

pixeltable/functions/openai.py
```diff
@@ -30,19 +30,60 @@
...
+    """Checks and possibly updates the file descriptor limits to accommodate the max connections.
+    Returns the max connections limit to use for the OpenAI client."""
+    try:
+        import resource
```
aaron-siegel: Maybe better to check if we're running Windows and fail-fast, so that we're not swallowing the error if something goes wrong with the import on other systems?

pixeltable/functions/openai.py
sergey-mkhitaryan: Removed the comment. I'll iterate on the knobs once I'm able to run perf test somewhere other than my laptop.

pixeltable/functions/openai.py
sergey-mkhitaryan: Moved to `utils/http.py`.

Any reason why it's a bad idea for OpenAI, Anthropic, etc. clients to update this limit lazily?

pixeltable/utils/http.py
```diff
@@ -0,0 +1,164 @@
...
+import re
+import time
+from http import HTTPStatus
+from random import random
+from typing import Any
+
+_logger = logging.getLogger('pixeltable')
+
+
+def set_file_descriptor_limit(preferred_limit: int) -> None:
```
aaron-siegel: Does this really belong in utils.http? Something like utils.system would be more logical.

aaron-siegel (approved):

PR #964, opened 2025-11-26
Docs Cleanup + Cookbooks + Versioning/Lineage + Production for Workshop

pierrebrunelle:
Lots of cookbooks, working with pydantic, custom iterator, custom UDA, model versions update to working with X


pierrebrunelle:
- @apreshill I'd love your 2cts on the cookbooks and we can discuss to scale it down or structure
- @aaron-siegel I'd love your review of the custom **iterator** and custom **UDA** ones - skip the rest

docs/release/docs.json
```diff
@@ -73,26 +71,57 @@
...
+                    "pages": [
+                      "howto/cookbooks/core/workflow-api-keys",
+                      "howto/cookbooks/core/dev-iterative-workflow",
+                      "howto/cookbooks/core/workflow-json-extraction",
+                      "howto/cookbooks/core/workflow-uuid-identity",
+                      "howto/cookbooks/core/query-join-tables",
+                      "howto/cookbooks/core/version-control-history",
+                      "howto/cookbooks/core/custom-aggregates-uda",
+                      "howto/cookbooks/core/custom-iterators",
+                      "howto/cookbooks/core/time-zones"
```
aaron-siegel, line 93: time-zones really isn't a cookbook, it's reference material on how time zones behave and are configured in Pixeltable.

docs/release/docs.json
```diff
@@ -300,5 +363,35 @@
     "body": {
       "family": "Inter"
     }
-  }
+  },
+  "redirects": [
```
aaron-siegel, line 367: Why do we need these?

docs/release/docs.json
```diff
@@ -23,39 +23,37 @@
             ]
           },
           {
-            "group": "Platform",
+            "group": "Core Concepts",
```
aaron-siegel, line 26: This is going to be a confusing heading if we also introduce a "Concepts" section. Is the idea that the other material we were thinking of putting in "Concepts" would also go here?

docs/release/docs.json
pierrebrunelle: For older links that were indexed by SEOs to make sure they are not 404 and redirects to existing pages - I'm still cleaning it up but that doesn't harm anything to have it, better safe than sorry on this.

docs/release/docs.json
pierrebrunelle: Yes my proposal is to merge them for now. We have 0 content for what is suggested as Concepts and Platform doesn't represents what is actually in this section. We can introduce a new heading: "Reference" once we have things that can go in there. @apreshill

docs/release/docs.json
pierrebrunelle: We can put it in configuration or others but at a top level "Concept or Reference", to me Time zone sticks out as an intruder and doesn't represent a core concept or element of Pixeltable, it's merely a "setting" or "configuration" or a "how-to". It doesn't belong next to Table, Views, Computed Columns or Embeddings. Open to suggestion.

apreshill:
This is looking so good. 

General feedback:

1. Really like this Cookbooks organization, and the "how to" section overall
2. General comment: for every core concept, we should make sure to repeatedly link to the SDK docs pages to avoid mystical functions. Not just at the end or beginning, but when a function is shown.


Specific questions:
1. For Core Concepts, what do you think about having a combined "Iterators & Views" section? Both of these pages seem very overlapping right now, and iterators are arguably "my first view" for most users. If we want to keep them separate, we may want to work on the iterators page so it is more distinct.
2. WDYT about having a core concepts section called "functions"- as a place to describe/link to:
    - built in functions (link to this section in the SDK: https://pixeltable-dev.mintlify.app/sdk/latest/audio)
    - integrations (in the how-to guides, but these might be worth highlighting)
    - UDFs/UDAs (current UDF page)

aaron-siegel:
I think both the Views and Iterators sections need to be rewritten and turned into notebooks like the Tables/Computed Columns/Queries & Expressions notebooks. (This was the original plan btw, around 1 year ago: see PXT-222 and PXT-226! But that plan was interrupted when we switched to mintlify and introduced the "Multimodal Datastore" section or whatever it was called, and buried the notebooks. Now we're coming full circle I guess.)

If we have a "Functions" section we should call it "UDFs" (we've long ago agreed to consistently use the term UDF to mean "function" in Pixeltable. "Function" already has 10000 meanings in Python/computing.)

> This is looking so good.
> 
> General feedback:
> 
> 1. Really like this Cookbooks organization, and the "how to" section overall
> 2. General comment: for every core concept, we should make sure to repeatedly link to the SDK docs pages to avoid mystical functions. Not just at the end or beginning, but when a function is shown.
> 
> Specific questions:
> 
> 1. For Core Concepts, what do you think about having a combined "Iterators & Views" section? Both of these pages seem very overlapping right now, and iterators are arguably "my first view" for most users. If we want to keep them separate, we may want to work on the iterators page so it is more distinct.
> 2. WDYT about having a core concepts section called "functions"- as a place to describe/link to:
>    
>    * built in functions (link to this section in the SDK: https://pixeltable-dev.mintlify.app/sdk/latest/audio)
>    * integrations (in the how-to guides, but these might be worth highlighting)
>    * UDFs/UDAs (current UDF page)

apreshill:
> I think both the Views and Iterators sections need to be rewritten and turned into notebooks like the Tables/Computed Columns/Queries & Expressions notebooks. (This was the original plan btw, around 1 year ago: see PXT-222 and PXT-226! But that plan was interrupted when we switched to mintlify and introduced the "Multimodal Datastore" section or whatever it was called, and buried the notebooks. Now we're coming full circle I guess.)
> 
> If we have a "Functions" section we should call it "UDFs" (we've long ago agreed to consistently use the term UDF to mean "function" in Pixeltable. "Function" already has 10000 meanings in Python/computing.)
> 
> > This is looking so good.
> > General feedback:
> > 
> > 1. Really like this Cookbooks organization, and the "how to" section overall
> > 2. General comment: for every core concept, we should make sure to repeatedly link to the SDK docs pages to avoid mystical functions. Not just at the end or beginning, but when a function is shown.
> > 
> > Specific questions:
> > 
> > 1. For Core Concepts, what do you think about having a combined "Iterators & Views" section? Both of these pages seem very overlapping right now, and iterators are arguably "my first view" for most users. If we want to keep them separate, we may want to work on the iterators page so it is more distinct.
> > 2. WDYT about having a core concepts section called "functions"- as a place to describe/link to:
> >    
> >    * built in functions (link to this section in the SDK: https://pixeltable-dev.mintlify.app/sdk/latest/audio)
> >    * integrations (in the how-to guides, but these might be worth highlighting)
> >    * UDFs/UDAs (current UDF page)

Do you have an alternative idea for how to surface built in functions, integrations, etc in addition to UDFs and UDAs? I suggested "functions" for precisely the reason that I want users to think "oh I need to know what functions I can use and write here." This information is currently findable, but you have to know how to already use Pixeltable. If I was a new user coming into this doc site (and I have been!), I would not know about built in functions or that integrations are already built- I would assume I need to roll my own UDFs.

For example:
https://datafusion.apache.org/library-user-guide/functions/index.html

docs/release/docs.json
pierrebrunelle: @aaron-siegel Lmk if you have a better place for it, but for now I don't see it. I'll reference that notebook in the new Multimodal Type System page wrt Timestamp.

apreshill (approved):
LGTM- this PR makes the casing more consistent across all the docs: https://github.com/pierrebrunelle/pixeltable/pull/3

aaron-siegel (approved):

PR #967, opened 2025-11-26
Some rough-edges features/improvements

mkornacker:
This is to support the upcoming workshop:
- thumbnail() udf for images
- VideoSplitter can now take an empty list of segment_times (useful for passing in the output of scene detection, which might not produce any scene breaks)
- added segment_times parameter to VideoSplitter docstring

### DIFF
--- a/pixeltable/iterators/video.py
+++ b/pixeltable/iterators/video.py
@@ -293,6 +293,9 @@ class VideoSplitter(ComponentIterator):
         video_encoder_args: dict[str, Any] | None = None,
     ):
         Env.get().require_binary('ffmpeg')
+        self._check_args(
+            duration, segment_times, overlap, min_segment_duration, mode, video_encoder, video_encoder_args
+        )
         assert (duration is not None) != (segment_times is not None)
         if duration is not None:
             assert duration > 0.0
@@ -334,6 +337,44 @@ class VideoSplitter(ComponentIterator):
         }
 
     @classmethod
+    def _check_args(
+        cls,
+        segment_duration: Any,
+        segment_times: Any,
+        overlap: Any,
+        min_segment_duration: Any,
+        mode: Any,
+        video_encoder: Any,
+        video_encoder_args: Any,
+    ) -> None:
+        if segment_duration is None and segment_times is None:
+            raise excs.Error('Must specify either duration or segment_times')
+        if segment_duration is not None and segment_times is not None:
+            raise excs.Error('duration and segment_times cannot both be specified')
+        if segment_times is not None and overlap is not None:
+            raise excs.Error('overlap cannot be specified with segment_times')
+        if segment_duration is not None and isinstance(segment_duration, (int, float)):
+            if segment_duration <= 0.0:
+                raise excs.Error(f'duration must be a positive number: {segment_duration}')
+            if (
+                min_segment_duration is not None
+                and isinstance(min_segment_duration, (int, float))
+                and segment_duration < min_segment_duration
+            ):
+                raise excs.Error(
+                    f'duration must be at least min_segment_duration: {segment_duration} < {min_segment_duration}'
+                )
+            if overlap is not None and isinstance(overlap, (int, float)) and overlap >= segment_duration:
+                raise excs.Error(f'overlap must be less than duration: {overlap} >= {segment_duration}')
+        if mode == 'accurate' and overlap is not None:
+            raise excs.Error("Cannot specify overlap for mode='accurate'")
+        if mode == 'fast':
+            if video_encoder is not None:
+                raise excs.Error("Cannot specify video_encoder for mode='fast'")
+            if video_encoder_args is not None:
+                raise excs.Error("Cannot specify video_encoder_args for mode='fast'")
+
+    @classmethod
     def output_schema(cls, *args: Any, **kwargs: Any) -> tuple[dict[str, ts.ColumnType], list[str]]:
         param_names = ['duration', 'overlap', 'min_segment_duration', 'segment_times']
         params = dict(zip(param_names, args))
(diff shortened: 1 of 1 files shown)


pixeltable/functions/image.py
```diff
@@ -400,6 +400,31 @@ def reduce(self: PIL.Image.Image, factor: int, box: tuple[int, int, int, int] |
...
+
+    Args:
+        size: The size of the thumbnail, as a tuple of (width, height).
+        resample: The resampling filter to use. See the
+            [Pillow documentation](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.thumbnail)
+            for a list of supported filters.
+        reducing_gap: The reducing gap to use.
+    """
+    result = self.copy()
+    result.thumbnail(size, resample, reducing_gap)  # type: ignore[arg-type]
```
aaron-siegel: Instead of `type: ignore`, why not just do this in a typesafe way?

`resample_ = PIL.Image.Resampling(resample)`
`result.thumbnail(size, resample_, reducing_gap)`

then we also get validation on the integer input for free.

pixeltable/iterators/video.py
```diff
@@ -254,6 +254,9 @@ class VideoSplitter(ComponentIterator):
         duration: Video segment duration in seconds
         overlap: Overlap between consecutive segments in seconds. Only available for `mode='fast'`.
         min_segment_duration: Drop the last segment if it is smaller than min_segment_duration.
+        segment_times: List of timestamps (in seconds) in video where segments should be split. Note that these are not
+            segment durations. If all segment times are less than the duration of the video, produces exactly
+            `len(segment_times) + 1` segments.
```
aaron-siegel: Is this true even if one of the segment_times is `0`? Would I then get a duration-0 segment? What happens if one of the segment times precedes `video_start_time`?

Also, explain the difference between `segment_times=None` and `segment_times=[]`

pixeltable/iterators/video.py
```diff
@@ -346,17 +348,18 @@ def output_schema(cls, *args: Any, **kwargs: Any) -> tuple[dict[str, ts.ColumnTy
...
             raise excs.Error('duration and segment_times cannot both be specified')
-        if segment_times is not None:
-            if len(segment_times) == 0:
-                raise excs.Error('segment_times cannot be empty')
-            if overlap is not None:
-                raise excs.Error('overlap cannot be specified with segment_times')
-        if segment_duration is not None:
+        if segment_times is not None and overlap is not None:
+            raise excs.Error('overlap cannot be specified with segment_times')
+        if segment_duration is not None and isinstance(segment_duration, (int, float)):
```
aaron-siegel, line 356: What's the purpose of `isinstance(segment_duration, (int, float))`? Isn't this implied by the type signature?

pixeltable/iterators/video.py
```diff
@@ -346,17 +348,18 @@ def output_schema(cls, *args: Any, **kwargs: Any) -> tuple[dict[str, ts.ColumnTy
...
             if segment_duration <= 0.0:
                 raise excs.Error('duration must be a positive number')
-            if min_segment_duration is not None and segment_duration < min_segment_duration:
+            if (
+                min_segment_duration is not None
+                and isinstance(min_segment_duration, (int, float))
```
aaron-siegel, line 361: Likewise, I don't understand the purpose of these conditionals

pixeltable/iterators/video.py
```diff
@@ -367,15 +370,47 @@ def output_schema(cls, *args: Any, **kwargs: Any) -> tuple[dict[str, ts.ColumnTy
...
-            'segment_end': ts.FloatType(nullable=False),
-            'segment_end_pts': ts.IntType(nullable=False),
+            'segment_start': ts.FloatType(nullable=True),
+            'segment_start_pts': ts.IntType(nullable=True),
+            'segment_end': ts.FloatType(nullable=True),
+            'segment_end_pts': ts.IntType(nullable=True),
             'video_segment': ts.VideoType(nullable=False),
         }, []
 
+    def complete_video_iter(self) -> Iterator[dict[str, Any]]:
```
aaron-siegel, line 402: Add docstring to explain what this is doing

pixeltable/iterators/video.py
```diff
@@ -367,15 +370,47 @@ def output_schema(cls, *args: Any, **kwargs: Any) -> tuple[dict[str, ts.ColumnTy
...
+
+        with av.open(str(self.video_path)) as container:
+            video_stream = container.streams.video[0]
+            start_ts = (
+                float(video_stream.start_time * video_stream.time_base)
+                if video_stream.start_time is not None and video_stream.time_base is not None
+                else 0.0
+            )
+            end_pts = (
+                video_stream.start_time + video_stream.duration
```
aaron-siegel, line 414: This is not always a reliable way to get the duration of the video stream. Make sure these code paths are tested with our "10-Second Video" test video

pixeltable/iterators/video.py
```diff
@@ -367,15 +370,47 @@ def output_schema(cls, *args: Any, **kwargs: Any) -> tuple[dict[str, ts.ColumnTy
...
+            )
+            result = {
+                'segment_start': start_ts,
+                'segment_start_pts': video_stream.start_time,
+                'segment_end': end_ts,
+                'segment_end_pts': end_pts,
+                'video_segment': str(self.video_path),
+            }
+            yield result
+            return
```
aaron-siegel: Extraneous `return`

aaron-siegel (changes requested):

pixeltable/functions/image.py
mkornacker: Done.

pixeltable/iterators/video.py
mkornacker: Clarified. I don't know if you'll get a playable video if you pass in 0 (but what could you do with that anyway).

pixeltable/iterators/video.py
mkornacker: No, this is checking for constants. You can also pass in exprs.

pixeltable/iterators/video.py
mkornacker: Same.

pixeltable/iterators/video.py
mkornacker: Done.

pixeltable/iterators/video.py
mkornacker: It's checking both inputs. What's missing?

pixeltable/iterators/video.py
mkornacker: Done.

pixeltable/iterators/video.py
aaron-siegel: Ah. Makes sense, *but* if we want to handle the case where segment_duration is an expr, then this is really not the right place to be doing this verification: if it *is* an expr, and the expr happens to evaluate to a negative duration for some row, then this check will never happen.

Rather, we should do the verification here, but do it again in `__init__()` (currently it's an assertion)

pixeltable/iterators/video.py
aaron-siegel: Sometimes `video_stream.duration` is missing - I believe that's true in the "10-Second Video" file, for example. The fallback is to look up the `'DURATION'` field in metadata; see FrameIterator for an example of how to do this.

Maybe we should break that logic out into a util method `get_video_stream_duration()` or some such.

aaron-siegel (approved):

pixeltable/iterators/video.py
mkornacker: I think we want both. If you pass in constants, you get the error when you create the view, rather then when you're trying to insert into the base table later, which is probably what you want.

pixeltable/iterators/video.py
mkornacker: As per the pyav docs, the duration attribute should always be there, but might be None.

PR #972, opened 2025-12-01
Documentation and config updates

aaron-siegel:
- Fix iterator docstring rendering
- Display `@overload`ed signatures correctly
- Downsample images in markdown output from notebook conversion to keep `.mdx` file sizes manageable
- Prepend `pxt.` to types in UDF signatures for clarity
- Reve notebook fixes
- Change VoyageAI config section to `voyage` (from `voyageai`)

### DIFF
--- a/pixeltable/functions/gemini.py
+++ b/pixeltable/functions/gemini.py
@@ -219,6 +219,9 @@ async def generate_videos(
         await asyncio.sleep(3)
         operation = await _genai_client().aio.operations.get(operation)
 
+    if operation.error:
+        raise Exception(f'Video generation failed: {operation.error}')
+
     video = operation.response.generated_videos[0]
 
     video_bytes = await _genai_client().aio.files.download(file=video.video)  # type: ignore[arg-type]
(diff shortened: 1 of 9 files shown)


pixeltable/func/signature.py
```diff
@@ -227,7 +227,7 @@ def __str__(self) -> str:
             elif p.kind == inspect.Parameter.VAR_KEYWORD:
                 param_strs.append(f'**{p.name}')
             else:
-                param_strs.append(f'{p.name}: {p.col_type}')
+                param_strs.append(f'{p.name}: pxt.{p.col_type}')
```
mkornacker, line 230: That feels a bit weird, not sure we should try to embed module abbreviations here.

aaron-siegel: I think we need them in the docs. So the alternatives are:
- Use `pixeltable.` as the prefix
- Remove the prefix from code, and munge the signatures in the docs to include it, either as `pxt.` or `pixeltable.` (this would be more work)

Note that in the legacy SDK docs we use `pxt.`

PR #979, opened 2025-12-05
PXT-903 Add UUID in pixeltable types 

amithadke:
- Add uuid4 function to generate uuids
- To create table that has auto generated uuid4 as primary key :
   pxt.create_table('test_uuid_pk_tbl', {'id': pxt.uuid4(), ...}, primary_key=['id', ...])

### DIFF
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -68,15 +68,8 @@ class ColumnType:
                 return t
             return None
 
-    scalar_types: ClassVar[set[Type]] = {
-        Type.STRING,
-        Type.INT,
-        Type.FLOAT,
-        Type.BOOL,
-        Type.TIMESTAMP,
-        Type.DATE,
-        Type.UUID,
-    }
+    scalar_json_types: ClassVar[set[Type]] = {Type.STRING, Type.INT, Type.FLOAT, Type.BOOL}
+    scalar_types: ClassVar[set[Type]] = scalar_json_types | {Type.TIMESTAMP, Type.DATE, Type.UUID}
     numeric_types: ClassVar[set[Type]] = {Type.INT, Type.FLOAT}
     common_supertypes: ClassVar[dict[tuple[Type, Type], Type]] = {
         (Type.BOOL, Type.INT): Type.INT,
@@ -210,7 +203,13 @@ class ColumnType:
         # Default: just compare base types (this works for all types whose only parameter is nullable)
         return self._type == other._type
 
-    def supertype(self, other: ColumnType) -> ColumnType | None:
+    def supertype(self, other: ColumnType, for_inference: bool = False) -> ColumnType | None:
+        """
+        Returns the most specific type that is a supertype of both `self` and `other`.
+
+        If `for_inference=True`, then we disallow certain type relationships that are technically correct, but may
+        be confusing for schema inference during data imports.
+        """
         if self == other:
             return self
         if self.matches(other):
@@ -225,7 +224,15 @@ class ColumnType:
             t = self.Type.supertype(self._type, other._type, self.common_supertypes)
             if t is not None:
                 return self.make_type(t).copy(nullable=(self.nullable or other.nullable))
-            return None
+
+        # If we see a mix of JSON and/or JSON-compatible scalar types, resolve to JSON.
+        # (For JSON+JSON, we return None to allow JsonType to handle merging the type schemas.)
+        if not for_inference and (
+            (self.is_json_type() and other.is_scalar_json_type())
+            or (self.is_scalar_json_type() and other.is_json_type())
+            or (self.is_scalar_json_type() and other.is_scalar_json_type())
+        ):
+            return JsonType(nullable=(self.nullable or other.nullable))
 
         return None
 
@@ -280,7 +287,7 @@ class ColumnType:
             if inferred_type is None:
                 inferred_type = val_type
             else:
-                inferred_type = inferred_type.supertype(val_type)
+                inferred_type = inferred_type.supertype(val_type, for_inference=True)
             if inferred_type is None:
                 return None
             if not inferred_type.has_supertype():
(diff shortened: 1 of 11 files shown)


pixeltable/functions/uuid.py
```diff
@@ -0,0 +1,32 @@
...
+import uuid
+
+import sqlalchemy as sql
+
+import pixeltable as pxt
+from pixeltable.utils.code import local_public_names
+
+
+@pxt.udf
+def make_uuid() -> uuid.UUID:
```
aaron-siegel: Just call it uuid4()

We should maybe support at least some of the others as well? Not everyone prefers uuid4(). I don't know how those would map onto postgres.

pixeltable/type_system.py
```diff
@@ -713,6 +733,38 @@ def _create_literal(self, val: Any) -> Any:
...
+
+    def _validate_literal(self, val: Any) -> None:
+        if not isinstance(val, uuid.UUID):
+            raise TypeError(f'Expected uuid.UUID, got {val.__class__.__name__}')
+
+    def _create_literal(self, val: Any) -> Any:
+        if isinstance(val, str):
+            return uuid.UUID(val)
+        if isinstance(val, uuid.UUID):
+            return val
```
aaron-siegel: What's the point of this if we just `return val` anyway?

amithadke: ohh i had bytes support (removed it), will just return val

pixeltable/functions/uuid.py
amithadke: will do, gen_random_uuid() will be using uuid4 anyway

amithadke: renamed it to make_uuid4

pierrebrunelle:
**Cookbook added**

I've created a cookbook demonstrating the UUID/Identity feature: [workflow-uuid-identity.ipynb](https://github.com/pierrebrunelle/pixeltable/blob/PXT-903/docs/cookbook/workflow-uuid-identity.ipynb)

It covers:
- Creating tables with `pxt.Identity` columns
- Adding UUID columns to existing tables via `add_column(col_name=pxt.Identity)`
- Using `make_uuid4()` in computed columns

**Suggestion: Remove `add_identity_column()`**

The `add_identity_column()` method appears redundant. Users could use already:
- Use `table.add_column(col_name=pxt.Identity)` to add a UUID column
- Use `table.drop_column('col_name')` to remove it

Since we treat pxt.identify as a type when defining the schema alongside other columns in the create table. It matches to me the pattern of our having a schema with default value with we know is a feature we want to provide in the future.

pierrebrunelle:
### Cookbook added

- Demo: Creating tables with pxt.Identity columns
- Adding UUID columns to existing tables via add_column(col_name=pxt.Identity)
- Demo: Using make_uuid4() in computed columns

Suggestion: Remove **add_identity_column()**
- The add_identity_column() method appears redundant. Users **could** already:
- Use table.add_column(col_name=pxt.Identity) to add a UUID column
- Use table.drop_column('col_name') to remove it

This would be consistent with [PXT-139](https://pixeltable.atlassian.net/browse/PXT-139) (Enable adding a new non-nullable column to a nonempty table by specifying a default value).`pxt.Identity` would just be a type with an implicit default, making `add_identity_column()` redundant.

[PXT-139]: https://pixeltable.atlassian.net/browse/PXT-139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pierrebrunelle (approved):
Updated cookbook - LGTM

tests/test_table.py
```diff
@@ -960,6 +968,77 @@ def test_schema_types(self, reset_db: None) -> None:
...
+        ]
+
+        # Test basic UUID column operations: insert and query
+        t = pxt.create_table('test_uuid_tbl', {'uuid_col': pxt.UUID})
+        validate_update_status(t.insert({'uuid_col': u} for u in test_uuids), expected_rows=len(test_uuids))
+
+        # Query all UUIDs
+        res = reload_tester.run_query(t.select(t.uuid_col))
+        assert len(res) == len(test_uuids)
+        assert res['uuid_col'] == test_uuids
```
mkornacker: This will break if the order changes randomly. Best to use a set.

tests/test_table.py
```diff
@@ -960,6 +968,77 @@ def test_schema_types(self, reset_db: None) -> None:
         df = t._col_descriptor()
         assert list(df['Type']) == expected_strings + expected_strings
 
+    def test_uuid_type(self, reset_db: None, reload_tester: ReloadTester) -> None:
```
mkornacker: Let's move this into a new test file, test_table.py is getting a bit hard to navigate.

tests/test_table.py
```diff
@@ -2710,6 +2791,30 @@ def test_add_computed_column(self, test_tbl: pxt.Table, reload_tester: ReloadTes
 
         reload_tester.run_reload_test()
 
+    def test_uuid_primary_key(self, reset_db: None, reload_tester: ReloadTester) -> None:
+        # Test creating a table with a UUID primary key using computed column
+        t = pxt.create_table('test_uuid_pk_tbl', {'id': {'value': uuid4()}, 'data': pxt.String}, primary_key=['id'])
```
mkornacker: Can you also add the abbreviated form
{'id': pxtf.uuid4(), ...}

That's easier to follow (and I don't think we have that specification dict documented anywhere).

tests/test_table.py
```diff
@@ -2710,6 +2791,30 @@ def test_add_computed_column(self, test_tbl: pxt.Table, reload_tester: ReloadTes
...
+
+        # Verify UUID column is created as primary key
+        pk_cols = t._tbl_version.get().primary_key_columns()
```
mkornacker: Don't use private members in tests.

get_metadata() is public.

aaron-siegel (approved):

PR #983, opened 2025-12-07
Multimodal support for Gemini `generate_content()`

aaron-siegel:
Adds multimodal support to Gemini `generate_content()` integration by enhancing the UDF signature to allow JSON input.

Enhances unit tests for `generate_content()` and `generate_videos()`.

Modifies the type hierarchy to make JsonType() a supertype of the JSON scalars (str, int, float, bool) to ensure that `generate_content()` can accept strings and is backward-compatible with the string-only signature.

Adds a `restrictive: bool` parameter to `supertype()` to allow for a more restrictive type hierarchy for schema inference during data imports.

### DIFF
--- a/tests/functions/test_gemini.py
+++ b/tests/functions/test_gemini.py
@@ -28,11 +28,9 @@ class TestGemini:
             top_p=0.95,
             top_k=40,
             response_mime_type='text/plain',
-            presence_penalty=0.6,
-            frequency_penalty=0.6,
         )
         t.add_computed_column(output2=generate_content(t.contents, model='gemini-2.5-flash', config=config))
-        validate_update_status(t.insert(contents='Write a story about a magic backpack.'), expected_rows=1)
+        validate_update_status(t.insert(contents='Write a sentence about a magic backpack.'), expected_rows=1)
         results = t.collect()
         text = results['output'][0]['candidates'][0]['content']['parts'][0]['text']
         text2 = results['output2'][0]['candidates'][0]['content']['parts'][0]['text']


pierrebrunelle:
Do we need to expand the working with gemini notebook?

pixeltable/functions/gemini.py
```diff
@@ -34,11 +34,11 @@ def _genai_client() -> 'genai.client.Client':
 
 @pxt.udf(resource_pool='request-rate:gemini')
 async def generate_content(
-    contents: str, *, model: str, config: dict | None = None, tools: list[dict] | None = None
+    contents: pxt.Json, *, model: str, config: dict | None = None, tools: list[dict] | None = None
 ) -> dict:
     """
-    Generate content from the specified model. For additional details, see:
-    <https://ai.google.dev/gemini-api/docs/text-generation>
+    Generate content from the specified model. `contents` can be a prompt, or a list containing images and text
```
mkornacker: Move to line 52?

pixeltable/type_system.py
```diff
@@ -213,7 +220,15 @@ def supertype(self, other: ColumnType) -> ColumnType | None:
...
             if t is not None:
                 return self.make_type(t).copy(nullable=(self.nullable or other.nullable))
-            return None
+
+        # If we see a mix of JSON and/or JSON-compatible scalar types, resolve to JSON.
+        # (For JSON+JSON, we return None to allow JsonType to handle merging the type schemas.)
+        if not restrictive and (
+            (self.is_json_type() and other.is_scalar_json_type())
+            or (self.is_scalar_json_type() and other.is_json_type())
+            or (self.is_scalar_json_type() and other.is_scalar_json_type())
```
mkornacker, line 229: Where do we need this case?

pixeltable/type_system.py
```diff
@@ -198,7 +199,13 @@ def matches(self, other: ColumnType) -> bool:
         # Default: just compare base types (this works for all types whose only parameter is nullable)
         return self._type == other._type
 
-    def supertype(self, other: ColumnType) -> ColumnType | None:
+    def supertype(self, other: ColumnType, restrictive: bool = False) -> ColumnType | None:
```
mkornacker: It sounds cryptic. Having an explicit "allow_json" (or something else, but meaning !restrictive) would be clearer.

tests/test_exprs.py
```diff
@@ -618,8 +618,8 @@ def test_inline_array(self, test_tbl: pxt.Table) -> None:
         assert col_type.dtype == np.dtype('int64')
 
         with pytest.raises(pxt.Error) as excinfo:
-            _ = t.select(pxt.array([t.c1, t.c2])).collect()
-        assert 'element of type `Int` at index 1 is not compatible with type `String` of preceding elements' in str(
+            _ = t.select(pxt.array([datetime(2025, 12, 5), t.c2])).collect()
+        assert 'element of type `Int` at index 1 is not compatible with type `Timestamp` of preceding elements' in str(
```
mkornacker: This should still flag int and string. To turn that into an array of json doesn't make sense.

tests/test_types.py
```diff
@@ -218,13 +217,17 @@ def test_supertype(self, init_env: None) -> None:
             (ImageType(height=100, width=200, mode='RGB'), ImageType(height=300, width=400, mode='RGBA'), ImageType()),
             (ImageType(height=100, width=200, mode='RGB'), ImageType(), ImageType()),
             (JsonType(), JsonType(), JsonType()),
+            (IntType(), StringType(), JsonType()),
```
mkornacker, line 220: This isn't useful.

pixeltable/functions/gemini.py
aaron-siegel: Sure. Done.

pixeltable/type_system.py
aaron-siegel: Ok. I was anticipating that we might in the future have other differences between the "type resolving" hierarchy and the "type inference" hierarchy. How about `for_inference` or some such?

pixeltable/type_system.py
aaron-siegel: It's implied by the other conditions; without it the type hierarchy wouldn't be a semilattice, e.g., we'd have supertype(json, supertype(int, float)) != supertype(supertype(json, int), float)

tests/test_exprs.py
aaron-siegel: Good point. It was still generating an error, but the wrong error (the error it was giving was "Element type is not a valid dtype for an array: Json")

Fixed so that it uses the type inference hierarchy (which is correct in this situation in any case)

tests/test_types.py
aaron-siegel: maybe not, but it's definitionally true

PR #986, opened 2025-12-10
[PXT-915] Gemini embedding UDFs

sergey-mkhitaryan:


pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+# The vector dimensionality limit imposed by pgvector (note: Pixeltable does not use halfvec yet)
+_MAX_EMBEDDING_DIMENSIONALITY = 2000
+# The highest embedding dimensionality recommended by Google that is within the above limit
+_DEFAULT_EMBEDDING_DIMENSIONALITY = 1536
+
+
+@pxt.udf(resource_pool='request-rate:gemini', batch_size=32)
+async def generate_embedding(
+    input: Batch[str], *, model: str = 'gemini-embedding-001', config: dict | None = None, async_: bool = False
+) -> Batch[pxt.Array[(None,), pxt.Float]]:
```
aaron-siegel: Might as well start using the "proper" dtypes (I assume np.float32 here)

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
 
+# The vector dimensionality limit imposed by pgvector (note: Pixeltable does not use halfvec yet)
+_MAX_EMBEDDING_DIMENSIONALITY = 2000
+# The highest embedding dimensionality recommended by Google that is within the above limit
+_DEFAULT_EMBEDDING_DIMENSIONALITY = 1536
+
+
+@pxt.udf(resource_pool='request-rate:gemini', batch_size=32)
+async def generate_embedding(
+    input: Batch[str], *, model: str = 'gemini-embedding-001', config: dict | None = None, async_: bool = False
```
aaron-siegel: `config: dict[str, Any]`?

Also, `async_` isn't what Google calls it. I know Google's terminology is confusing but IMO this makes it even more confusing. Maybe call it `use_gemini_batch_api` or some such.

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
     return f'request-rate:veo:{model}'
 
 
+# The vector dimensionality limit imposed by pgvector (note: Pixeltable does not use halfvec yet)
+_MAX_EMBEDDING_DIMENSIONALITY = 2000
```
aaron-siegel: I feel like this belongs somewhere other than in `gemini.py`. Isn't this a more general constraint? Probably we should have a check in the index creation code.

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+# The highest embedding dimensionality recommended by Google that is within the above limit
+_DEFAULT_EMBEDDING_DIMENSIONALITY = 1536
+
+
+@pxt.udf(resource_pool='request-rate:gemini', batch_size=32)
+async def generate_embedding(
+    input: Batch[str], *, model: str = 'gemini-embedding-001', config: dict | None = None, async_: bool = False
+) -> Batch[pxt.Array[(None,), pxt.Float]]:
+    """Generate embeddings for the input strings. For more information on Gemini embeddings API, see:
+    https://ai.google.dev/gemini-api/docs/embeddings
```
aaron-siegel: Bare URLs in markdown are discouraged - enclose in angle braces:
```
<https://ai.google.dev/gemini-api/docs/embeddings>
```

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+
+    This UDF can run in either synchronous or asynchronous mode. Synchronous mode is optimized for latency.
+    Asynchronous mode uses Gemini's Batch API described at https://ai.google.dev/gemini-api/docs/batch-api which
+    provides a better throughput at lower cost at the expense of higher latency.
+
+    The `output_dimensionality` in the config must not exceed 2000, if present.
```
aaron-siegel: Have you tested the specific limit of 2000 to make sure it's accurate?

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+        The generated embeddings.
+
+    Examples:
+        Add a computed column with embeddings to an existing table with a `text` column:
+
+        >>> t.add_computed_column(embedding=generate_embedding(t.text))
+
+        Add an embedding index on `text` column:
+
+        >>> t.add_embedding_index(t.text, embedding=generate_embedding.using(model='gemini-embedding-001', async_=True))
```
aaron-siegel: I find the inclusion of the `async_` parameter confusing here, especially with no additional context given. Unsuspecting users might presume it's somehow related to the fact that it's an embedding index, since that's the only noted difference with the preceding example. Clearest to just leave it out.

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+    env.Env.get().require_package('google.genai')
+
+    client = _genai_client()
+    config_ = _embedding_config(config)
+
+    if not async_:
+        requests: list[Any] = input  # makes mypy happy
+        result = await client.aio.models.embed_content(model=model, contents=requests, config=config_)
+        assert len(result.embeddings) == len(input)
+        return [np.array(emb.values, dtype=np.float64) for emb in result.embeddings]
```
aaron-siegel: Why `float64`?

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+
+    # Async embedding generation (Google calls it Batch API)
+    from google.genai import types
```
aaron-siegel: Move to top of function (immediately after `require_package()`)

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+        batch_job = client.batches.get(name=batch_job.name)
+        if batch_job.state in (
+            types.JobState.JOB_STATE_SUCCEEDED,
+            types.JobState.JOB_STATE_FAILED,
+            types.JobState.JOB_STATE_CANCELLED,
+            types.JobState.JOB_STATE_EXPIRED,
+        ):
+            break
+        _logger.debug(f'Waiting for embedding batch job {batch_job.name} to complete. Latest state: {batch_job.state}')
+        await asyncio.sleep(30)
```
aaron-siegel: seems like an awfully long sleep given how cheap it is to check each time we wake up.

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+    return f'request-rate:gemini:{model}'
+
+
+def _embedding_config(config: dict | None) -> 'genai.types.EmbedContentConfig':
+    from google.genai import types
+
+    config_ = types.EmbedContentConfig(**config) if config else types.EmbedContentConfig()
+    if config_.output_dimensionality is None:
+        config_.output_dimensionality = _DEFAULT_EMBEDDING_DIMENSIONALITY
+    elif config_.output_dimensionality > _MAX_EMBEDDING_DIMENSIONALITY:
```
aaron-siegel: see above - maybe we move this check to index creation.

After all, it's perfectly fine to use a large dimensionality if defining as a *computed column* (which doesn't need pgvector); just not for an embedding index.

tests/functions/test_gemini.py
```diff
@@ -129,3 +131,49 @@ def test_generate_videos(self, reset_db: None) -> None:
...
+        skip_test_if_no_client('gemini')
+        from pixeltable.functions.gemini import generate_embedding
+
+        t = pxt.create_table('test', {'rowid': pxt.Int, 'text': pxt.String})
+
+        # Test embeddings as a computed column
+        t.add_computed_column(
+            embedding=generate_embedding(t.text, async_=async_, config={'output_dimensionality': 768})
+        )
+        assert t.embedding.col.col_type.matches(ts.ArrayType((768,), np.dtype('float64'))), t.embedding.col.col_type
```
aaron-siegel: Why `float64`? Does Gemini actually give 64-bit precision?

tests/functions/test_gemini.py
```diff
@@ -129,3 +131,49 @@ def test_generate_videos(self, reset_db: None) -> None:
...
             assert audio_stream['duration_seconds'] == duration, metadata
+
+    def test_generate_embeddings(self, reset_db: None) -> None:
+        self._test_generate_embeddings(async_=False)
+
+    @pytest.mark.skip('Very slow')
+    def test_generate_embeddings_async(self, reset_db: None) -> None:
+        self._test_generate_embeddings(async_=True)
+
+    def _test_generate_embeddings(self, async_: bool) -> None:
```
aaron-siegel: Add tests for:
- No optional arguments specified
- Exactly 2000 dimensions specified
- Exactly 2001 dimensions specified

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+
+    # Async embedding generation (Google calls it Batch API)
+    from google.genai import types
+
+    batch_job = client.batches.create_embeddings(
+        model=model,
+        src=types.EmbeddingsBatchJobSource(inlined_requests=types.EmbedContentBatch(contents=input, config=config_)),
+    )
+
+    await asyncio.sleep(10)
```
mkornacker: 10? That's a long time to start out with.

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+
+@pxt.udf(resource_pool='request-rate:gemini', batch_size=32)
+async def generate_embedding(
+    input: Batch[str], *, model: str = 'gemini-embedding-001', config: dict | None = None, async_: bool = False
+) -> Batch[pxt.Array[(None,), pxt.Float]]:
+    """Generate embeddings for the input strings. For more information on Gemini embeddings API, see:
+    https://ai.google.dev/gemini-api/docs/embeddings
+
+    This UDF can run in either synchronous or asynchronous mode. Synchronous mode is optimized for latency.
+    Asynchronous mode uses Gemini's Batch API described at https://ai.google.dev/gemini-api/docs/batch-api which
```
mkornacker: Asynchronous is confusing in this context, it always means asyncio in Python, and this is unrelated.

Best to reference 'batch' directly.

pixeltable/functions/gemini.py
mkornacker: Yes, this check belongs in the embedding index creation logic.

See EmbeddingIndex.init() and .get_index_sa_type().

pixeltable/functions/gemini.py
sergey-mkhitaryan: done -- calling it batch API now

pixeltable/functions/gemini.py
sergey-mkhitaryan: yes (manually)

pixeltable/functions/gemini.py
sergey-mkhitaryan: You're probably right, there's indirect evidence that the embedding values are float32 at most.

pixeltable/functions/gemini.py
sergey-mkhitaryan: Just to be clear, that's a call over network each time this happens. How much should the wait be? In my experience, jobs (even small with only 3 texts) can take >10min to complete.

tests/functions/test_gemini.py
sergey-mkhitaryan: 32 should be fine, see the other comment above

pixeltable/functions/gemini.py
aaron-siegel: Most embeddings are 32-bit. You should be able to tell fairly easily just by inspecting the output

pixeltable/index/embedding_index.py
```diff
@@ -127,7 +130,13 @@ def records_value_errors(self) -> bool:
 
     def get_index_sa_type(self, val_col_type: ts.ColumnType) -> sql.types.TypeEngine:
         assert isinstance(val_col_type, ts.ArrayType) and val_col_type.shape is not None
+        assert len(val_col_type.shape) == 1, (
```
aaron-siegel: These should raise pxt.Error, not fire assertions (they're user-facing errors, correct)?

pixeltable/index/embedding_index.py
```diff
@@ -127,7 +130,13 @@ def records_value_errors(self) -> bool:
...
+            f'Pixeltable does not support multi-dimensional embedding vectors: {val_col_type}'
```
aaron-siegel: I'd say "Expected a 1-dimensional array, got {n}: ..."

Technically a vector is 1-dimensional so "multi-dimensional vector" is not a meaningful term. Multi-vector embeddings are a separate concept so here it's not really a question of what's supported or not, it's a question of what inputs are meaningful.

pixeltable/index/embedding_index.py
```diff
@@ -127,7 +130,13 @@ def records_value_errors(self) -> bool:
...
+        )
         vector_size = val_col_type.shape[0]
+        assert vector_size <= MAX_EMBEDDING_VECTOR_LENGTH, (
+            f'Embedding vector size exceeds the maximum allowed size of {MAX_EMBEDDING_VECTOR_LENGTH}: {val_col_type}'
```
aaron-siegel: Add a TODO to support larger embeddings via halfvec columns

tests/functions/test_gemini.py
```diff
@@ -129,3 +131,81 @@ def test_generate_videos(self, reset_db: None) -> None:
...
+        skip_test_if_not_installed('google.genai')
+        skip_test_if_no_client('gemini')
+        from pixeltable.functions.gemini import generate_embedding
+
+        t = pxt.create_table('test', {'rowid': pxt.Int, 'text': pxt.String})
+
+        # Test embeddings as a computed column
+        config = None
+        expected_dim = 1536
+        if not default_params:
```
aaron-siegel: This isn't quite what I was suggesting. Rather than parameterize the test, just add separate columns that exercise both patterns.

`t.add_computed_column(embedding1=generate_embedding(t.text))`
`t.add_computed_column(embedding2=generate_embedding(t.text, config=..., blahblahblah=...))`

The given implementation doesn't address my concern because it still passes `config=None`, and the purpose is to test what happens if we *don't* give a `config` kwarg. (We've had issues in the past with parameters inadvertently being declared as required, and this being overlooked because they were always specified in every test. So it's important to have tests that specifically *omit* optional kwargs.)

tests/functions/test_gemini.py
```diff
@@ -129,3 +131,81 @@ def test_generate_videos(self, reset_db: None) -> None:
...
+        )
+        for row in t.collect():
+            embedding = row['embedding']
+            assert isinstance(embedding, np.ndarray)
+            assert embedding.dtype == np.float32
+            assert embedding.shape == (expected_dim,)
+
+        # Test embeddings as an index
+        if default_params:
+            t.add_embedding_index(t.text, embedding=generate_embedding)
```
aaron-siegel: Won't this fail?

[Edit: I see; we're specifying a default model, but I'd argue we shouldn't be; see above comment]

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,122 @@ def _(model: str) -> str:
...
+# Google recommends using 768, 1536, or 3072 output dimensions, but 3072 exceeds the max vector length for embedding
+# indexes (2000). It's possible to exceed 2000 as long as the output vector is not stored in an embedding index.
+_DEFAULT_EMBEDDING_DIMENSIONALITY = 1536
+
+
+@pxt.udf(resource_pool='request-rate:gemini', batch_size=32)
+async def generate_embedding(
+    input: Batch[str],
+    *,
+    model: str = 'gemini-embedding-001',
```
aaron-siegel: I'd argue we shouldn't be specifying a default here (the models are continually changing and it's best not to hardcode it - in most UDFs model is a required parameter; we should follow that convention unless there's a specific reason to deviate from it here)

tests/test_index.py
```diff
@@ -800,3 +800,19 @@ def test_date_btree(self, reset_db: None) -> None:
...
+    @staticmethod
+    def dummy_embed_max_len(text: str) -> pxt.Array[(2000,), np.float32]:
+        pass
+
+    @pxt.udf
+    @staticmethod
+    def dummy_embed_too_long(text: str) -> pxt.Array[(2001,), np.float32]:
+        pass
+
+    def test_embedding_length_limit(self, reset_db: None) -> None:
```
aaron-siegel: This doesn't quite test what I was hoping for, which is a test that vectors of length 2000 actually get written to postgres correctly. It's fine to use the dummy embeddings but I think we should have a test that creates an index of vector length 2000 *and writes something to it*.

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+
+    - `pip install google-genai`
+
+    Args:
+        input: The strings to generate embeddings for.
+        model: The Gemini model to use.
+        config: Configuration for embedding generation, corresponding to keyword arguments of
+            `genai.types.EmbedContentConfig`. For details on the parameters, see:
+            <https://googleapis.github.io/python-genai/genai.html#genai.types.EmbedContentConfig>
+        use_batch_api: If True, use Gemini's Batch API (https://ai.google.dev/gemini-api/docs/batch-api) that provides
```
aaron-siegel: Enclose URLs in angle brackets to ensure correct rendering in markdown.

Maybe even better, just make it into a link:
```
use [Gemini's Batch API](https://ai.google.dev/gemini-api/docs/batch-api)
```
[note one does *not* use angle brackets in that case; the above is standard markdown notation.]

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+
+        >>> t.add_embedding_index(
+        ...    t.text, embedding=generate_embedding.using(model='gemini-embedding-001')
+        ...)
+    """
+    env.Env.get().require_package('google.genai')
+    from google.genai import types
+
+    client = _genai_client()
+    config_ = _embedding_config(config)
```
aaron-siegel, line 291: I have one lingering concern with "hijacking" the default here, which is that if we do support 3072-dimension vectors in the future, we'll still be "locked in" to 1536 as the default for backward compatibility. That seems unfortunate. The alternative would be to simply use the defaults given by Google and live with the fact that the default dimensionality will generate an exception for the time being, if used as an embedding index.

It's also possible that different models have, or in the future will have, different default dimensionalities (there are a number of other providers with that behavior), so we might regret locking us in to a fixed override of the default dimensionality.

sergey-mkhitaryan: @mkornacker what do you think, should we stop imposing our own default embedding vector length? If we do, `add_embedding_index` won't work with this UDF's defaults, at least until we implement halfvec embedding indexes.

pixeltable/functions/gemini.py
```diff
@@ -242,6 +247,118 @@ def _(model: str) -> str:
...
+
+    Returns:
+        The generated embeddings.
+
+    Examples:
+        Add a computed column with embeddings to an existing table with a `text` column:
+
+        >>> t.add_computed_column(embedding=generate_embedding(t.text))
+
+        Add an embedding index on `text` column:
```
mkornacker, line 278: Change the example to show how to set the length of the generated vectors (such as 1536).

pixeltable/functions/gemini.py
mkornacker: I would say let's leave out a default here and simply follow this up with the halfvec changes.

For the workshop, we can set the length explicitly.

PR #987, opened 2025-12-10
Twelve Labs multimodal embeddings support

aaron-siegel:
Working implementation of Twelve Labs image, audio, video, and text_image embeddings.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -83,6 +83,7 @@ env:
   REPLICATE_API_TOKEN: ${{ secrets.REPLICATE_API_TOKEN }}
   REVE_API_KEY: ${{ secrets.REVE_API_KEY }}
   TOGETHER_API_KEY: ${{ secrets.TOGETHER_API_KEY }}
+  TWELVELABS_API_KEY: ${{ secrets.TWELVELABS_API_KEY }}
   VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
 
 jobs:


pixeltable/functions/twelvelabs.py
```diff
@@ -69,20 +64,125 @@ async def embed(
...
+
+@embed.overload
+async def _(
+    model_name: str,
+    *,
+    audio: pxt.Audio,
+    start_sec: float | None = None,
+    end_sec: float | None = None,
+    embedding_option: list[Literal['audio', 'transcription']] | None = None,
+    embedding_scope: list[Literal['clip', 'asset']] | None = None,
```
mkornacker: I don't think this parameter makes sense, since we don't expose the segmentation aspect (and I don't think we should, our Audio-/VideoSplitter are more useful for that). We should simply pass 'asset'.

pixeltable/functions/twelvelabs.py
```diff
@@ -69,20 +64,125 @@ async def embed(
...
+
+@embed.overload
+async def _(
+    model_name: str,
+    *,
+    video: pxt.Video,
+    start_sec: float | None = None,
+    end_sec: float | None = None,
+    embedding_option: list[Literal['visual', 'audio', 'transcription']] | None = None,
+    embedding_scope: list[Literal['clip', 'asset']] | None = None,
```
mkornacker: Same here.

pixeltable/functions/twelvelabs.py
aaron-siegel: Done.

pixeltable/functions/twelvelabs.py
aaron-siegel: Done.

PR #991, opened 2025-12-12
PXT-910 Add presigned_url udf

amithadke:
- Add create_presigned_url (for http GET) to all non local stores
- Refactor s3 store client to cache client per region (needed for presigned url)
- Adding test in test_destination


pixeltable/functions/presigned_url.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Returns:
+        A presigned HTTP URL for accessing the file
+
+    Raises:
+        Error: If the URI is a local file:// path
+
+    Examples:
+        >>> tbl = pxt.get_table('my_table')
+        >>> # Using with media column fileurl (e.g., Video, Image, Audio)
```
mkornacker: Examples should be formatted like so:

    Examples:
        Split a video at 1 minute intervals using fast mode:

        >>> tbl.select
        ...     segment_paths=tbl.video.segment_video(
        ...         duration=60, mode='fast'
        ...     )
        ... ).collect()

pixeltable/utils/object_stores.py
```diff
@@ -358,15 +361,33 @@ def list_objects(self, return_uri: bool, n_max: int = 10) -> list[str]:
...
+        """Create a presigned URL for downloading an object from the store.
+
+        Args:
+            soa: StorageObjectAddress containing the object location
+            expiration_seconds: Time in seconds for the URL to remain valid
+
+        Returns:
+            A presigned HTTP URL that can be used to access the object
+        """
+        raise AssertionError
```
mkornacker, line 374: shouldn't this be raise AssertionError()?

amithadke: Either works — they're functionally identical. raise AssertionError raises the class (Python instantiates it automatically), raise AssertionError() raises an instance explicitly.

(from claude)

pixeltable/functions/net.py
```diff
@@ -0,0 +1,69 @@
+"""
+Pixeltable UDF for converting media file URIs to presigned HTTP URLs.
```
aaron-siegel, line 2: Do we really need a dedicated module for `presigned_url`? If it doesn't fit neatly into an existing module, maybe we could pick a more general module name like `net.py` or `http.py`. @pierrebrunelle @apreshill Any opinions here?

pixeltable/functions/presigned_url.py
```diff
@@ -0,0 +1,69 @@
...
+Pixeltable UDF for converting media file URIs to presigned HTTP URLs.
+"""
+
+from pixeltable import exceptions as excs
+from pixeltable.func.udf import udf
+from pixeltable.utils.code import local_public_names
+from pixeltable.utils.object_stores import ObjectOps, ObjectPath, StorageTarget
+
+
+@udf(is_method=True)
```
aaron-siegel: I'm not sure this makes sense as a method of `StringType`.

pixeltable/functions/net.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Convert a blob storage URI to a presigned HTTP URL for direct access.
+
+    Generates a time-limited, publicly accessible URL from cloud storage URIs
+    (S3, GCS, Azure, etc.) that can be used to serve media files over HTTP.
+
+    Note:
+        This function uses presigned URLs from storage providers. Provider-specific
+        limitations apply:
+        - Google Cloud Storage: maximum 7-day expiration
```
aaron-siegel, line 23: Bulleted lists in markdown must be preceded by a blank line or they will render improperly in the docs

pixeltable/functions/presigned_url.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Note:
+        This function uses presigned URLs from storage providers. Provider-specific
+        limitations apply:
+        - Google Cloud Storage: maximum 7-day expiration
+        - AWS S3: requires proper region configuration
+        - Azure: subject to storage account access policies
+
+    Args:
+        uri: The media file URI (e.g., s3://bucket/path, gs://bucket/path, azure://container/path)
```
aaron-siegel: I suggest putting the URI snippets in backticks for clarity in the docs.

pixeltable/functions/presigned_url.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Raises:
+        Error: If the URI is a local file:// path
+
+    Examples:
+        Generate a presigned URL for a video column with 1-hour expiration:
+
+        >>> tbl.select(
+        ...     original_url=tbl.video.fileurl,
+        ...     presigned_url=tbl.video.fileurl.presigned_url(3600)
```
aaron-siegel: Here I'd argue the pattern should be `pxtf.net.presigned_url(tbl.video.fileurl, 3600)` or some such, rather than it being a method of `StringType`.

pixeltable/utils/object_stores.py
```diff
@@ -229,6 +230,8 @@ def parse_object_storage_addr1(cls, src_addr: str) -> StorageObjectAddress:
                 storage_target = StorageTarget.B2_STORE
             elif 'windows' in parsed.netloc:
                 storage_target = StorageTarget.AZURE_STORE
+            elif 't3.storage.dev' in parsed.netloc or parsed.netloc == 't3.storage.dev':
```
aaron-siegel: Isn't the second condition redundant?

aaron-siegel:
Approved, but I removed from merge queue because I did have some feedback that I hope can be taken into account. Thanks!

PR #992, opened 2025-12-12
Iterators Refactor Part 1

aaron-siegel:
Moves iterators from `pxt.iterators` to the appropriate `pxt.functions` module, e.g., `pxt.iterators.video.FrameIterator` becomes `pxt.functions.video.frame_iterator`. I've left the names unchanged (`_iterator`, `_splitter`) for now.

Marks the iterator `create()` methods as `@deprecated`; `pxt.iterators` can now be considered deprecated as a public package.

Fixes a few bugs in iterator initialization that were uncovered by making the initializer parameters explicit.

Updates the SDK docs for the new iterators and removes `pxt.iterators` from the SDK docs. I haven't updated the notebooks and markdown yet; I'll do that once we've finalized the names.


tests/test_component_view.py
```diff
@@ -7,8 +7,8 @@
 
 import pixeltable as pxt
 import pixeltable.type_system as ts
+from pixeltable.functions.video import frame_iterator
```
mkornacker, line 10: What about ConstantImgIterator below?

aaron-siegel: It's only used for testing, so I figured there's no need to refactor it and it might as well wait for when we update the pattern.

PR #1001, opened 2025-12-15
Add `pxt.Binary` type to type system; `bytes` support in JSON; working Gemini 3 Pro

aaron-siegel:
- Adds `pxt.Binary` / `BinaryType` to type system
- Support for storing Python `bytes` in JSON structs
- `bytes` are now externalized the same way as `ndarray`s, via CellMaterializationNode (both in independent columns and as elements of JSON structs)
- Gemini 3 Pro now works correctly
- Adds ReloadTester to all tests in `test_inlined_objects.py`
- Various test cleanup / consolidation

### DIFF
--- a/pixeltable/exec/cell_materialization_node.py
+++ b/pixeltable/exec/cell_materialization_node.py
@@ -158,7 +158,7 @@ class CellMaterializationNode(ExecNode):
 
     def _materialize_binary_cell(self, row: exprs.DataRow, col: catalog.Column, val: bytes) -> None:
         if len(val) <= self.MAX_DB_BINARY_SIZE:
-            # this binary is small enough to store in the db column (type: binary) directly
+            # this `bytes` object is small enough to store in the db column (type: binary) directly
             row.cell_vals[col.id] = val
             row.cell_md[col.id] = None
         else:


pixeltable/type_system.py
```diff
@@ -546,6 +558,10 @@ def is_media_type(self) -> bool:
         # types that refer to external media files
         return self.is_image_type() or self.is_video_type() or self.is_audio_type() or self.is_document_type()
 
+    def is_materializable(self) -> bool:
```
mkornacker: I think something along the lines of has_external_storage() would be clearer (I wasn't sure what is_materializable() meant). The critical factor is the out-of-band storage.

pixeltable/exec/cell_materialization_node.py
```diff
@@ -153,6 +156,25 @@ def _materialize_array_cell(self, row: exprs.DataRow, col: catalog.Column, val:
 
         assert row.cell_vals[col.id] is not None or row.cell_md[col.id] is not None
 
+    def _materialize_binary_cell(self, row: exprs.DataRow, col: catalog.Column, val: bytes) -> None:
+        if len(val) <= self.MAX_DB_BINARY_SIZE:
+            # this binary is small enough to store in the db column (type: binary) directly
```
mkornacker: this bytes array...

aaron-siegel: Fixed.

pixeltable/type_system.py
aaron-siegel: I changed to `supports_file_offloading()`. We can rename again if we think of a better name

PR #1005, opened 2025-12-16
Add VLM Run integration for multimodal chat completions

wbrennan899:
## Summary

This PR adds a VLM Run integration. It provides five UDFs that call VLM Run's OpenAI-compatible API (Orion 2). The default model is `vlmrun-orion-2:auto`.

Users pass Pixeltable columns directly: `pxt.Image`, `pxt.Video`, and `pxt.Document` are accepted as UDF arguments, and can also be referenced inside `chat_completions` messages. When a UDF runs, it uploads the media file to VLM Run and uses the resulting file id for that one API call. No provider file ids are stored in table state.

## UDFs

- **`chat_completions(messages, *, model, model_kwargs)`** — chat and data extraction over text, images, documents, and video. Columns are referenced in message content items (`image_url` / `video_url` / `file_url`). Supports structured JSON output through `response_format` and toolset selection through `extra_body`. Returns the API response as a dict.
- **`generate_image(prompt, *, image, ...)`** — generates an image from text, or edits the given image. Returns `PIL.Image.Image`.
- **`annotate_image(prompt, image, *, ...)`** — draws bounding boxes, keypoints, or masks on an image. Returns `PIL.Image.Image`.
- **`generate_video(prompt, *, image, video, ...)`** — generates a video from text, an image, or another video. Returns `pxt.Video`.
- **`generate_document(prompt, document, *, ...)`** — transforms a document — for example, redacting PII. Returns the result as a `pxt.Document` (PDF).

The four generation UDFs take a `timeout` parameter (default 600 seconds) that bounds how long they wait for results.

## Changes since the previous review

Each item below is numbered as in the review punch list:

1. Public UDFs now take Pixeltable column types instead of `file_path: str`. `chat_completions` resolves column references inside messages, following the same approach as `openai.py`.
2. Provider failures raise `pxt.ExternalServiceError` with `provider='vlmrun'`.
3. `Env.get().require_package('vlmrun')` runs at the start of every UDF and in the sync client helper.
4. All UDFs are declared `is_deterministic=False`.
5. Artifact polling uses exponential backoff (250ms doubling to 16s) and fails immediately on auth errors.
6. Response schemas are built once per process and cached (`functools.lru_cache`).
7. The notebook loads its sample PDF from a public `raw.githubusercontent.com` URL.
8. There is a comment explaining why a second (sync) client exists: the file upload API is only available in the sync `vlmrun` SDK.

The branch was also rebased onto current `main` and squashed to one commit. The default model moved to Orion 2 and the SDK is pinned at `vlmrun>=0.6.5`.

## Provider behaviors the code handles

The earlier review asked for a heads-up on anything that runs into VLM Run's API constraints. Two behaviors required handling in the integration:

- **Requests that run longer than about 150 seconds.** The API does not keep the connection open. Instead it responds `303 See Other` with a URL where the result will appear. That URL can answer `200` or `201` with the result, return `204`/empty responses or intermittent `500`s while the job is still running, or hold the connection open for several minutes. The `_chat_create()` helper follows the redirect and polls until the result is ready, within the UDF's `timeout`. We verified this end to end on video and document generation jobs.
- **Artifact ids sometimes come back with a file extension attached** (`vid_a1b2c3.mp4` instead of `vid_a1b2c3`). `_classify_media_ref()` strips the extension before calling the artifacts endpoint.

VLM Run also puts size limits on inputs. In our testing, editing a 2732x2147 image failed with a token-limit error, and transforming a 15-page PDF failed with an input-too-large error. Smaller inputs work: the same image edit succeeded at 1024px, and the same PDF operation succeeded on a single page. The error messages from the API say what went wrong, and the UDFs pass them through to the user.

## Testing

`tests/functions/test_vlmrun.py` has two kinds of tests:

- **Tests that run in normal CI** (no API key needed). These cover the local logic: converting column values inside messages into uploads, and parsing the ids the API returns.
- **Tests that call the real API.** These are marked `remote_api` (plus `expensive` or `very_expensive` for the generation ones) and need a `VLMRUN_API_KEY`: `pytest -m remote_api tests/functions/test_vlmrun.py`. They cover each UDF with text, image, document, and video inputs, and check that one failing row in a batch records an error without affecting the other rows.

Before opening this revision we ran everything against the live API with real files — a scanned invoice, a research paper PDF, a Word document, street video footage, a high-resolution photo, and a website screenshot. 23 operations completed end to end, including invoice extraction into JSON, object detection, face blurring, video key-moment extraction, video-to-video generation, and PDF redaction with rendered black boxes.

The docs notebook is in the nb-tests skip list because its cells call the paid API.


pixeltable/functions/vlmrun.py
```diff
@@ -0,0 +1,426 @@
...
+
+    return VLMRun(api_key=api_key, base_url='https://agent.vlm.run/v1')
+
+
+def _vlmrun_client() -> 'VLMRun':
+    return env.Env.get().get_client('vlmrun')
+
+
+@pxt.udf(resource_pool='request-rate:vlmrun')
+async def chat_image(
```
spillai: Can we make this into a `chat_completions` UDF similar to [fireworks](https://github.com/pixeltable/pixeltable/blob/main/pixeltable/functions/fireworks.py#L31) plugin?

pixeltable/functions/vlmrun.py
```diff
@@ -0,0 +1,426 @@
...
+
+        >>> tbl.add_computed_column(file_id=vlmrun.upload(tbl.document.localpath))
+        >>> tbl.add_computed_column(
+        ...     summary=vlmrun.chat_with_file(tbl.file_id, "Summarize this document")
+        ... )
+    """
+
+    def _call_api() -> str:
+        client = _vlmrun_client()
+        uploaded = client.files.upload(file=Path(media))
```
spillai: We raise errors if these are unsupported file formats, so we should be mindful of catching there here.

pixeltable/functions/vlmrun.py
```diff
@@ -0,0 +1,426 @@
...
+            temperature=0,
+        )
+        return response.choices[0].message.content
+
+    result = await asyncio.to_thread(_call_api)
+    return result
+
+
+@pxt.udf(resource_pool='request-rate:vlmrun')
+async def redact(
```
spillai: I would drop this UDF for now, let's keep the API simple.

apreshill:
Hi @wbrennan899!

Thank you again for taking this integration on, and I apologize for our delay in reviewing and responding to your PR since we last chatted. To move things along, I took a look at your PR in light of our sync in February, the TwelveLabs integration (which seems closest to VLM Run), and the Copilot review comments from April. I'm summarizing the findings below so you have one consolidated punch list for the next revision:

- The main design change is # 1 in the list below: fold the upload UDFs into the main UDFs, with no separate file-ID handling in table state. Your cache/reuse point makes sense; I think persistent provider file IDs are a broader Pixeltable design question to revisit once the pattern settles across providers.
- Items # 2, # 3, and # 4 came from comparing this PR to the patterns we used in the TwelveLabs integration, which is a useful reference point for VLM Run.
- I also put together a draft WIP integration guidelines doc for future contributors. It may be useful if you are using a coding assistant to revise this PR: [GUIDELINES_FOR_INTEGRATIONS.md](https://github.com/apreshill/pixeltable/blob/docs/integration-guidelines/docs/_guidelines/GUIDELINES_FOR_INTEGRATIONS.md). This is still a draft and needs to be finalized by our engineering team, but it captures patterns already present in our existing integrations.
- Items # 5, # 6, and # 7 also showed up in Copilot's April review.

After the next revision, we can bring this back to the team's review queue. Let me know if any of these run into VLM Run's API constraints.

## Findings

### Required for this PR

**1. Public UDFs should take Pixeltable column types rather than `file_path: str`.** All four UDFs currently accept `file_path: str`, and the notebook passes `t.col.localpath`. The public surface should take `pxt.Image` (or `PIL.Image.Image`) and `pxt.Video`/`pxt.Audio`/`pxt.Document` directly; the runtime will resolve these to PIL objects or local paths. For `chat_completions`, accept column references inside the messages list and resolve them internally before the API call, mirroring `openai.py:446` and `anthropic.py:146`. `file_path: str` still makes sense in internal helpers. Reference: `twelvelabs.py:71, 135, 156, 179`.

**2. Provider failures should be wrapped in `pxt.ExternalServiceError`.** Please use `pxt.ExternalServiceError(pxt.ErrorCode.PROVIDER_ERROR, ..., provider='vlmrun')` at two clear sites: `_download_artifact` when polling times out, and after each API call in the four UDFs when the response shape doesn't match. Reference: `twelvelabs.py:127`.

**3. `require_package('vlmrun')` should run at the call sites.** Add `Env.get().require_package('vlmrun')` as the first line of each UDF and inside `_vlmrun_sync_client`. The package is already registered in `Env.__register_packages`; this just wires up the runtime check. Reference: `twelvelabs.py:107`.

**4. Generative UDFs should be marked with `is_deterministic=False`.** Please add this to all four decorators: `@pxt.udf(is_deterministic=False, resource_pool='request-rate:vlmrun')`. Otherwise Pixeltable may reuse cached outputs on recompute when the provider would return a fresh result. References: `replicate.py:30`, `fal.py:30`, `reve.py:139, 175, 214`, `openai.py:242, 446, 572`, `anthropic.py:146`, `gemini.py:96`.

### Important cleanup

**5. Use backoff in `_download_artifact`.** Replace `time.sleep(5.0)` with exponential backoff (~250ms start, double up to ~16s). A short custom loop is fine.

**6. Cache the `model_json_schema()` results.** The generation UDFs (`generate_image`, `annotate_image`, `generate_video`) recompute their JSON schemas via `type(...).model_json_schema()` on every invocation. Please define them once at module scope or cache them, then reuse. This avoids repeated work on larger tables.

**7. Use a public URL for the notebook PDF.** The PDF cell uses `Path('../../../../tests/data/...layout-parser-paper.pdf')`, which only resolves inside the repo. Please replace it with the `raw.githubusercontent.com` URL, matching the image cell already in the notebook.

### Optional follow-ups

**8. Dual client setup.** The sync VLMRun SDK wrapped in `asyncio.to_thread` works. The refactor in #1 will reshape these call sites, so this may resolve naturally; if the dual setup persists, a one-line comment near `_vlmrun_sync_client` explaining why is fine.

**9. `chat_completions` returns `dict`.** If you want structured typing tied to `model_kwargs['response_format']`, register a `@conditional_return_type` callback. Reference: `twelvelabs.py:253-259`.

---

## Already looks right (worth keeping)

These already match the guidelines:

- `@pxt.udf(resource_pool='request-rate:vlmrun')` on all four UDFs
- `@register_client('vlmrun')` for the async client
- Lazy SDK imports (inside functions, not module top)
- `asyncio.to_thread` wrapping the sync SDK calls
- `vlmrun` in `Env.__register_packages` (`env.py:1061`)
- Test markers: `@pytest.mark.remote_api`, `@pytest.mark.expensive`, `@rerun(reruns=3, reruns_delay=8)`
- Test fixtures: `get_image_files()`, `get_documents()`
- Module footer: `__all__ = local_public_names(__name__)` and `__dir__`

pierrebrunelle:
grok review

PR #1007, opened 2025-12-17
[PXT-916] Store embedding indexes as halfvecs

sergey-mkhitaryan:
Adds a parameter to `Table.add_embedding_index` that controls the float precision used (16 or 32 bit, currently implemented using pgvector's halfvec and vector). The parameter defaults to 16 bit, which is different from the current behavior that always uses the regular float32 vector. To reconcile this difference, a catalog migration is implemented that adds precision=32bit to all embedding indexes in the schema.

### DIFF
--- a/pixeltable/index/embedding_index.py
+++ b/pixeltable/index/embedding_index.py
@@ -6,6 +6,7 @@ from typing import Any, ClassVar, Literal
 import numpy as np
 import pgvector.sqlalchemy  # type: ignore[import-untyped]
 import sqlalchemy as sql
+from sqlalchemy import cast
 
 import pixeltable.catalog as catalog
 import pixeltable.exceptions as excs
@@ -143,6 +144,11 @@ class EmbeddingIndex(IndexBase):
         assert vector_length is not None
         assert vector_length > 0
 
+        # TODO(PXT-941): Revisit embedding index precision behavior for cloud launch
+        # CockroachDB doesn't have HALFVEC. For now, always use Vector type.
+        if Env.get().is_using_cockroachdb:
+            return pgvector.sqlalchemy.Vector(vector_length)
+
         match self.precision:
             case self.Precision.FP32:
                 if vector_length > MAX_EMBEDDING_VECTOR_LENGTH:
@@ -183,10 +189,15 @@ class EmbeddingIndex(IndexBase):
         embedding = self.embeddings[item.col_type._type].exec([item.val], {})
         assert isinstance(embedding, np.ndarray)
 
+        # In arithmetic operations between floats and ints (or between vector and int), CockroachDB requires an explicit
+        # cast. Otherwise the query fails.
+        cast_ints = Env.get().is_using_cockroachdb
+        one = cast(1, sql.types.Float) if cast_ints else 1
+        neg_one = cast(-1, sql.types.Float) if cast_ints else -1
         if self.metric == self.Metric.COSINE:
-            return val_column.sa_col.cosine_distance(embedding) * -1 + 1
+            return val_column.sa_col.cosine_distance(embedding) * neg_one + one
         elif self.metric == self.Metric.IP:
-            return val_column.sa_col.max_inner_product(embedding) * -1
+            return val_column.sa_col.max_inner_product(embedding) * neg_one
         else:
             assert self.metric == self.Metric.L2
             return val_column.sa_col.l2_distance(embedding)
--- a/tests/test_index.py
+++ b/tests/test_index.py
@@ -11,6 +11,7 @@ import pytest
 
 import pixeltable as pxt
 import pixeltable.type_system as ts
+from pixeltable.env import Env
 from pixeltable.functions.huggingface import clip
 
 from .utils import (
@@ -809,18 +810,24 @@ class TestIndex:
             sim = img_t.img.similarity(string='red truck')
             _ = img_t.order_by(sim, asc=False).limit(1).collect()
 
-        with pytest.raises(
-            pxt.Error, match="Embedding index's vector dimensionality 4001 exceeds maximum of 4000 for fp16 precision"
-        ):
-            test_tbl.add_embedding_index(
-                test_tbl.c1, embedding=TestIndex.dummy_embedding.using(n=4001), precision='fp16'
-            )
-        with pytest.raises(
-            pxt.Error, match="Embedding index's vector dimensionality 2001 exceeds maximum of 2000 for fp32 precision"
-        ):
-            test_tbl.add_embedding_index(
-                test_tbl.c1, embedding=TestIndex.dummy_embedding.using(n=2001), precision='fp32'
-            )
+        if not Env.get().is_using_cockroachdb:
+            # TODO(PXT-941): Revisit embedding index precision behavior for cloud launch
+            # In CockroachDB we use VECTOR type that doesn't have the same limitation as pgvector's VECTOR and HALFVEC
+            with pytest.raises(
+                pxt.Error,
+                match="Embedding index's vector dimensionality 4001 exceeds maximum of 4000 for fp16 precision",
+            ):
+                test_tbl.add_embedding_index(
+                    test_tbl.c1, embedding=TestIndex.dummy_embedding.using(n=4001), precision='fp16'
+                )
+            with pytest.raises(
+                pxt.Error,
+                match="Embedding index's vector dimensionality 2001 exceeds maximum of 2000 for fp32 precision",
+            ):
+                test_tbl.add_embedding_index(
+                    test_tbl.c1, embedding=TestIndex.dummy_embedding.using(n=2001), precision='fp32'
+                )
+
         with pytest.raises(pxt.Error, match=r"Invalid precision.+Must be one of: \['fp16', 'fp32'\]"):
             test_tbl.add_embedding_index(
                 test_tbl.c1,


pixeltable/exec/cell_reconstruction_node.py
```diff
@@ -114,7 +115,9 @@ async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
                         assert cell_md.file_urls is not None and len(cell_md.file_urls) == 1
                         row[col_ref.slot_idx] = self._reconstruct_array(cell_md)
                     else:
-                        assert row[col_ref.slot_idx] is None or isinstance(row[col_ref.slot_idx], np.ndarray)
+                        assert row[col_ref.slot_idx] is None or isinstance(
+                            row[col_ref.slot_idx], (np.ndarray, pgvector.sqlalchemy.HalfVector)
```
mkornacker: Why is this showing up here?

pixeltable/exprs/data_row.py
```diff
@@ -274,9 +274,17 @@ def get_stored_val(self, index: int, sa_col_type: sql.types.TypeEngine | None =
             return self.file_urls[index]
 
         if self.vals[index] is not None and index in self.array_slot_idxs:
-            assert isinstance(self.vals[index], np.ndarray)
-            np_array = self.vals[index]
-            if sa_col_type is not None and isinstance(sa_col_type, pgvector.sqlalchemy.Vector):
+            if isinstance(self.vals[index], np.ndarray):
+                np_array = self.vals[index]
+            elif isinstance(self.vals[index], pgvector.sqlalchemy.HalfVector):
```
mkornacker: Why are values showing up as HalfVector? We should only be representing arrays as ndarray in datarows.

pixeltable/exprs/data_row.py
```diff
@@ -325,6 +333,10 @@ def __setitem__(self, idx: int, val: Any) -> None:
                 self.vals[idx] = self.file_paths[idx] if self.file_paths[idx] is not None else self.file_urls[idx]
         elif idx in self.array_slot_idxs and isinstance(val, bytes):
             self.vals[idx] = np.load(io.BytesIO(val))
+        elif idx in self.array_slot_idxs and isinstance(val, pgvector.sqlalchemy.HalfVector):
```
mkornacker: These need to be converted at the source. Ie, slot values should only be ndarray, not HalfVector.

pixeltable/plan.py
```diff
@@ -554,7 +554,9 @@ def _cell_md_col_refs(cls, expr_list: Iterable[exprs.Expr]) -> list[exprs.Column
         def needs_reconstruction(e: exprs.Expr) -> bool:
             assert isinstance(e, exprs.ColumnRef)
             # Vector-typed array columns are used for vector indexes, and are stored in the db
-            return e.col.col_type.is_array_type() and not isinstance(e.col.sa_col_type, pgvector.sqlalchemy.Vector)
+            return e.col.col_type.is_array_type() and not isinstance(
+                e.col.sa_col_type, (pgvector.sqlalchemy.Vector, pgvector.sqlalchemy.HALFVEC)
```
mkornacker: Add a Column.is_sa_vector_type() or something like that for this isinstance() check.

tests/test_index.py
```diff
@@ -800,3 +800,63 @@ def test_date_btree(self, reset_db: None) -> None:
...
+        return np.random.rand(n).astype(np.float32)
+
+    @dummy_embedding.conditional_return_type
+    @staticmethod
+    def _(n: int) -> ts.ArrayType:
+        return ts.ArrayType((n,), dtype=np.dtype('float32'), nullable=False)
+
+    @pytest.mark.parametrize('n,reload_cat', [(2000, True), (2000, False), (2001, True), (2001, False), (4000, False)])
+    def test_embedding_index_type_selection(self, reset_db: None, n: int, reload_cat: bool) -> None:
+        for metric in ['l2', 'cosine', 'ip']:
```
mkornacker: Might as well move that into parametrize() as well.

Apparently you can stack them:
@parametrize(n, ...)
@parametrize(reload, [True, False])
@parametrize(metric, [...])

tests/test_index.py
```diff
@@ -800,3 +800,63 @@ def test_date_btree(self, reset_db: None) -> None:
...
+                ]
+            )
+
+            reload_catalog(reload_cat)
+
+            res = t.select(t.text.embedding()).collect()
+            assert len(res.schema) == 1
+            assert ts.ArrayType((n,), np.dtype('float32')).matches(res.schema['col_0']), res.schema
+            assert len(res) == 3
+            assert isinstance(res[0, 0], np.ndarray)
```
mkornacker: assert all(isinstance(row['embedding']) for row in res)

tests/test_index.py
aaron-siegel: I'm not sure I agree it's a good idea to move them into parametrize. Overly parameterized tests leads to a combinatorial explosion of sprawling test instances... sometimes it's better to just have a loop inside a test

sergey-mkhitaryan: I like that, too, but @aaron-siegel may be of a different opinion, based on the past code reviews.

pixeltable/catalog/column.py
```diff
@@ -259,11 +259,14 @@ def has_window_fn_call(self) -> bool:
         )
         return len(window_fn_calls) > 0
 
+    def is_sa_vector_type(self) -> bool:
```
aaron-siegel: I'd suggest `has_sa_vector_type` or `is_sa_vector_col` as the name (the reference object is a column, not a type)

pixeltable/exec/sql_node.py
```diff
@@ -407,6 +409,10 @@ async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
                             output_row[slot_idx] = sql_row[i]
                     else:
                         raise RuntimeError(f'Unexpected datetime value for {e}')
+                elif isinstance(sql_row[i], pgvector.sqlalchemy.HalfVector):
+                    # Halfvecs are float16s, but because we dynamically select between Vector and Halfvec for embedding
```
aaron-siegel: It's less about dynamically or not dynamically selected, and more that we want the type of the returned object to match the declared return type of the UDF (which currently is always float32 for embedding functions).

pixeltable/exprs/data_row.py
```diff
@@ -286,7 +286,9 @@ def get_stored_val(self, index: int, sa_col_type: sql.types.TypeEngine | None =
         if self.vals[index] is not None and index in self.array_slot_idxs:
             assert isinstance(self.vals[index], np.ndarray)
             np_array = self.vals[index]
-            if sa_col_type is not None and isinstance(sa_col_type, pgvector.sqlalchemy.Vector):
+            if sa_col_type is not None and isinstance(
+                sa_col_type, (pgvector.sqlalchemy.Vector, pgvector.sqlalchemy.HALFVEC)
```
aaron-siegel, line 290: Why `HALFVEC` here but `HalfVector` elsewhere?

pixeltable/index/embedding_index.py
```diff
@@ -16,6 +16,9 @@
 
 from .base import IndexBase
 
+MAX_EMBEDDING_VECTOR_LENGTH = 2000
+MAX_EMBEDDING_HALFVEC_LENGTH = 4000
```
aaron-siegel, line 21: Aren't these definitionally redundant? (4000 = 2000 / 0.5 so the relationship is implied by the fact that they're "half" vectors?)

pixeltable/index/embedding_index.py
```diff
@@ -119,15 +127,28 @@ def records_value_errors(self) -> bool:
...
     def get_index_sa_type(self, val_col_type: ts.ColumnType) -> sql.types.TypeEngine:
         assert isinstance(val_col_type, ts.ArrayType) and val_col_type.shape is not None
-        vector_size = val_col_type.shape[0]
-        assert vector_size is not None
-        return pgvector.sqlalchemy.Vector(vector_size)
+        assert len(val_col_type.shape) == 1
+        vector_length = val_col_type.shape[0]
+        assert vector_length is not None
+
+        if vector_length <= MAX_EMBEDDING_VECTOR_LENGTH:
```
aaron-siegel: As I commented on slack - I think there's a strong argument for always defaulting to HalfVector for new embeddings - and making it a property of the embedding index that the user can control. Let's discuss at standup.

pixeltable/share/packager.py
```diff
@@ -216,6 +216,10 @@ def __to_pa_value(
             # Handle media files as described above
             assert isinstance(val, str)
             return self.__process_media_url(val)
+        if isinstance(sql_type, (sql_vector.HALFVEC)):
```
aaron-siegel: Spurious parens

tests/share/test_packager.py
```diff
@@ -298,11 +299,11 @@ def __validate_index_data(
                     for result in Env.get().conn.execute(q).fetchall():
                         v_min, v_max, val, undo = result
                         if v_min <= head_version and v_max > head_version:
-                            assert val is None or isinstance(val, np.ndarray)
+                            assert val is None or isinstance(val, (np.ndarray, pgvector.sqlalchemy.HalfVector))
```
aaron-siegel, line 302: Remove

tests/share/test_packager.py
```diff
@@ -298,11 +299,11 @@ def __validate_index_data(
...
                             assert undo is None
                         else:
                             assert val is None
-                            assert undo is None or isinstance(undo, np.ndarray)
+                            assert undo is None or isinstance(undo, (np.ndarray, pgvector.sqlalchemy.HalfVector))
```
aaron-siegel, line 306: Remove

tests/test_index.py
```diff
@@ -884,3 +884,63 @@ def test_date_btree(self, reset_db: None) -> None:
             start + datetime.timedelta(days=random.randint(0, int(delta_days))) for _ in range(self.BTREE_TEST_NUM_ROWS)
         ]
         self.run_btree_test(data, pxt.Date)
+
+    @pxt.udf
+    @staticmethod
+    def dummy_embedding(text: str, n: int) -> pxt.Array[(None,), np.float32]:
```
aaron-siegel, line 926: Do we have test coverage of the packager code too? (i.e., is that coverage implicit in the existing packager tests?)

aaron-siegel (changes requested):

pixeltable/exprs/data_row.py
sergey-mkhitaryan: Confusingly, `HALFVEC` is the column type (parallel to `Vector` -- yes, the naming style is off), whereas `HalfVector` is a class for representing a specific value (basically a thin wrapper around `numpy.ndarray` of float16s).

pixeltable/index/embedding_index.py
sergey-mkhitaryan: They are redundant if you know what you know, i.e. that this limits come from max pg page size, and that halfvec takes 2 times fewer bits per element than vector, etc. In this particular scenario my inclination is to be more explicit.

tests/test_index.py
sergey-mkhitaryan: we do now

tests/share/test_packager.py
sergey-mkhitaryan: it's necessary -- this code selects directly from postgres

pixeltable/catalog/table.py
```diff
@@ -983,6 +983,7 @@ def add_embedding_index(
         string_embed: pxt.Function | None = None,
         image_embed: pxt.Function | None = None,
         metric: Literal['cosine', 'ip', 'l2'] = 'cosine',
+        precision: Literal['16bit', '32bit'] = '16bit',
```
aaron-siegel: These are unusual names. The 'standard' names (that I've seen most often) are `float32`, `float16`, and `bfloat16`; OR `fp32`, `fp16`, and `bf16`. ["bfloat16" has a different ratio of mantissa bits to exponent bits (7:8) than the IEEE float16 (10:5).]

[EDIT: According to Gemini, pgvector uses IEEE float16.]

I guess I'd vote for fp32 and fp16 if pressed, but don't have a strong preference other than I think we should use something standard.

pixeltable/index/embedding_index.py
```diff
@@ -33,18 +36,29 @@ class Metric(enum.Enum):
         IP = 2
         L2 = 3
 
+    class Precision(enum.Enum):
+        _16BIT = '16bit'
```
aaron-siegel: The preceding underscores are not necessary here, and are inconsistent with other internal enum naming conventions

pixeltable/index/embedding_index.py
```diff
@@ -119,15 +138,37 @@ def records_value_errors(self) -> bool:
...
         assert isinstance(val_col_type, ts.ArrayType) and val_col_type.shape is not None
-        vector_size = val_col_type.shape[0]
-        assert vector_size is not None
-        return pgvector.sqlalchemy.Vector(vector_size)
+        assert len(val_col_type.shape) == 1
+        vector_length = val_col_type.shape[0]
+        assert vector_length is not None
+        assert vector_length > 0
+
+        if self.precision == self.Precision._32BIT:
```
aaron-siegel: Or use a matcher to make this pattern more symmetric:
```
match self.precision:
    case self.Precision._32BIT:
        ...
    case self.Precision._16BIT:
        ...
    case _:
        raise AssertionError(self.precision)
```

tests/share/test_packager.py
aaron-siegel: Ahhhh I see. So SQLAlchemy itself returns `np.ndarray` for 32-bit vectors, but `HalfVector` for 16-bit vectors? Weird.

aaron-siegel (changes requested):

tests/share/test_packager.py
sergey-mkhitaryan: Yes, that's what I'm seeing. I certainly did not write any code that forces HalfVector to be returned instead of `ndarray`. Claude said this is intended behavior when I checked some time ago, and I did not press for details or proof.

aaron-siegel (approved):

pixeltable/metadata/converters/convert_44.py
```diff
@@ -0,0 +1,29 @@
...
+        return None
+    if ('class_fqn', 'pixeltable.index.embedding_index.EmbeddingIndex') not in value.items():
+        return None
+    if 'init_args' not in value:
+        return None
+    assert isinstance(value['init_args'], dict), value
+    if 'precision' in value['init_args']:
+        return None
+    value['init_args']['precision'] = 'fp32'
+    return (key, value)
```
aaron-siegel: Style nit: spurious parentheses

`return key, value` is generally accepted here

tests/share/test_packager.py
```diff
@@ -819,23 +820,27 @@ def test_view_over_snapshot_round_trip(self, reset_db: None) -> None:
...
-    def test_embedding_index(self, reset_db: None, clip_embed: pxt.Function) -> None:
+    @pytest.mark.parametrize('embedding_precision', ['fp16', 'fp32'])
+    def test_embedding_index(self, reset_db: None, clip_embed: pxt.Function, embedding_precision: str) -> None:
         skip_test_if_not_installed('transformers')  # needed for CLIP
 
         t = pxt.create_table('tbl', {'image': pxt.Image})
         images = get_image_files()[:10]
         t.insert({'image': image} for image in images)
-        t.add_embedding_index('image', embedding=clip_embed)
+        t.add_embedding_index('image', embedding=clip_embed, precision=embedding_precision)  # type: ignore[arg-type]
```
aaron-siegel: Can the `type: ignore` be avoided by typing the `embedding_precision` test parameter as `Literal[...]`?

pixeltable/exec/sql_node.py
```diff
@@ -417,6 +419,10 @@ async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
                             output_row[slot_idx] = sql_row[i]
                     else:
                         raise RuntimeError(f'Unexpected datetime value for {e}')
+                elif isinstance(sql_row[i], HalfVector):
+                    # The type returned by selecting from an index should match the declared return type of the UDF
```
mkornacker: More appropriate comment:
all array data needs to be materialized as ndarrays.

PR #1013, opened 2025-12-18
Twelvelabs notebook

pierrebrunelle:
Cookbook for 12labs. I've also added the new similarity search that enables you to search with video, audio, and others, and showcase how to retrieve the embeddings as well. Please review.


pierrebrunelle:
@mkornacker @aaron-siegel
- Notebook has been re-written to be centered around the multimodality of video index.
- The change from https://github.com/pixeltable/pixeltable/pull/1023 was beneficial for image search but video and audio were not working so I'm suggesting a similar update to the column_ref for video and audio, I'm happy to discard that change if you have another fix/update in mind to provide the same seamless experience so I can simply give it an external URL HTTP as I can do now for image.

Thanks!

PR #1017, opened 2025-12-20
PXT-928: Export Json columns to parquet as pa.struct

mkornacker:


tests/utils.py
```diff
@@ -76,11 +76,11 @@ def create_table_data(t: pxt.Table, col_names: list[str] | None = None, num_rows
         col_names = []
     data: dict[str, Any] = {}
 
+    # avoid fields with empty dicts, they cannot be serialized as a struct in Parquet
```
aaron-siegel, line 79: This seems like a pretty severe limitation; what happens if I give an empty dict to the exporter? Do I get an exception? If so - should we at least have an option to discard such fields (or replace them with Nones) rather than just refusing to export?

mkornacker: There'll be an exception.

I'd prefer to leave that as a todo, it sounds like we should discuss how to handle this (in principle, this kind of data isn't compatible with Parquet).

aaron-siegel: Ok... let's add a TODO and a rough-edges ticket?

mkornacker: We should discuss first what that could look like.

It may be best/appropriate to do this as a complementary udf that cleans up json, and if you know you have broken json you add a new computed columns, which then functions as the target for export.

Traversing deeply nested json structures is pretty expensive in Python, and data that actually needs it is probably fairly rare.

PR #1018, opened 2025-12-20
Bedrock invoke_model() udf

mkornacker:


pixeltable/functions/bedrock.py
```diff
@@ -102,7 +118,166 @@ def converse(
...
+    <https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model.html>
+
+    Returns:
+        A dictionary containing the model response.
+
+    Examples:
+        Add a computed column using Amazon Titan embedding model:
+
+        >>> body = {'inputText': tbl.text, 'dimensions': 512, 'normalize': True}
+        ... tbl.add_computed_column(response=invoke_model(body, model_id='amazon.titan-embed-text-v2:0'))
```
aaron-siegel: Should use `>>>` not `...` because it's a new statement (this is necessary for doc formatting to work properly)

PR #1020, opened 2025-12-20
HF DetrForSegmentation

mkornacker:
Also included:
- udf overlay_segmentation() in pxtf.vision to visualize the segmentation map

### DIFF
--- a/tests/test_inlined_objects.py
+++ b/tests/test_inlined_objects.py
@@ -2,13 +2,14 @@ import random
 import time
 from typing import Any
 
-import PIL.Image
 import numpy as np
+import PIL.Image
 import pytest
 
 import pixeltable as pxt
 from pixeltable.env import Env
 from pixeltable.utils.local_store import LocalStore
+
 from .utils import ReloadTester, assert_columns_eq, inf_array_iterator, inf_image_iterator, validate_update_status
 
 


pixeltable/functions/huggingface.py
```diff
@@ -249,6 +249,88 @@ def detr_for_object_detection(
...
+        ...     model_id='facebook/detr-resnet-50-panoptic',
+        ...     threshold=0.5
+        ... ))
+    """
+    env.Env.get().require_package('transformers')
+    device = resolve_torch_device('auto')
+    import torch
+    from transformers import DetrForSegmentation, DetrImageProcessor
+
+    model = _lookup_model(model_id, lambda x: DetrForSegmentation.from_pretrained(x, revision=revision), device=device)
```
aaron-siegel: I think this is going to break with the `revision` parameter; it needs to be part of the key in `_model_cache` or else there will be cache collisions/corruption if two revisions of the same model are mentioned in the same process

mkornacker: I'll hardwire the revision instead.

detr_for_object_detection() has the same problem.

pixeltable/functions/huggingface.py
```diff
@@ -249,6 +249,85 @@ def detr_for_object_detection(
...
+        ...     model_id='facebook/detr-resnet-50-panoptic',
+        ...     threshold=0.5
+        ... ))
+    """
+    env.Env.get().require_package('transformers')
+    device = resolve_torch_device('auto')
+    import torch
+    from transformers import DetrForSegmentation, DetrImageProcessor
+
+    model = _lookup_model(model_id, lambda x: DetrForSegmentation.from_pretrained(x), device=device)
```
aaron-siegel: No need for the lambda.

```
_lookup_model(model_id, DetrForSegmentation.from_pretrained, device=device)
```

pixeltable/functions/huggingface.py
```diff
@@ -249,6 +249,85 @@ def detr_for_object_detection(
...
+    processor = _lookup_processor(model_id, lambda x: DetrImageProcessor.from_pretrained(x))
```
aaron-siegel: same

pixeltable/functions/huggingface.py
```diff
@@ -249,6 +249,85 @@ def detr_for_object_detection(
...
+        image: The image to segment.
+        model_id: The pretrained model to use for segmentation (e.g., 'facebook/detr-resnet-50-panoptic').
+        threshold: Confidence threshold for filtering segments.
+
+    Returns:
+        A dictionary containing the output of the segmentation model, in the following format:
+
+            ```python
+            {
+                'segmentation': np.ndarray,  # (H, W) array where each pixel value is a segment ID
```
aaron-siegel, line 272: Seems a little misleading since it's a dict value, not an annotation; something like this would be clearer:

```
'segmentation': arr,  # arr is an (H, W) np.ndarray instance where each pixel value is a segment ID
```

pixeltable/functions/vision.py
```diff
@@ -391,6 +392,106 @@ def draw_bounding_boxes(
...
+
+    Returns:
+        The image with the colored segmentation overlay.
+    """
+
+    if segmentation.shape != (img.height, img.width):
+        raise ValueError(
+            f'Segmentation shape {segmentation.shape} does not match image dimensions ({img.height}, {img.width})'
+        )
+    segment_ids = sorted([int(sid) for sid in np.unique(segmentation) if sid != background])
```
aaron-siegel: spurious list creation

pixeltable/functions/vision.py
```diff
@@ -391,6 +392,106 @@ def draw_bounding_boxes(
...
+            | padded[:-2, 2:]
+            | padded[2:, :-2]
+            | padded[2:, 2:]
+        )
+
+    return boundaries
+
+
+@pxt.udf
+def overlay_segmentation(
```
aaron-siegel, line 432: I feel like there should be a UDF for just creating the segmentation image (i.e. the image with colored regions) without necessarily overlaying.

Maybe the simplest way to do this would be to make this UDF instead something like
```
def segmentation_image(
    segmentation: pxt.Array[(None, None), np.int32],
    *,
    img: Optional[PIL.Image.Image] = None,
    alpha: float = 0.5,
    ...
)
```
If `img` is not specified, you get just the segmentation image. If `img` is specified, the segmentation map gets overlaid on top of it

mkornacker: overlay_segmentation() follows the approach taken by existing open-source libraries.

For example, Roboflow's supervision package contains a MaskAnnotator class which you can then use to create annotated images (ie, the mask overlay).

pixeltable/functions/huggingface.py

pixeltable/functions/huggingface.py

pixeltable/functions/huggingface.py
mkornacker: I'm not following. Where does it say it's an annotation?

pixeltable/functions/vision.py
mkornacker: Done.

PR #1022, opened 2025-12-22
Use head() instead of collect() in test_add_column_to_view

aaron-siegel:
We sometimes see spurious test failures because tests are comparing ResultSets without an ordering. They'll succeed ~99% of the time but occasionally fail because - well, there's no ordering, so who knows.

Let's get in the habit of fixing these anomalies when we see them rather than just re-running the tests

### DIFF
--- a/tests/test_view.py
+++ b/tests/test_view.py
@@ -217,7 +217,7 @@ class TestView:
     def test_add_column_to_view(self, reset_db: None, test_tbl: pxt.Table, reload_tester: ReloadTester) -> None:
         """Test add_column* methods for views"""
         t = test_tbl
-        t_c1_val0 = t.select(t.c1).head()[0]['c1']
+        t_c1_val0 = t.order_by(t.c1).collect()[0]['c1']
 
         # adding column with same name as a base table column at
         # the time of creating a view will raise an error now.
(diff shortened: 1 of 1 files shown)


tests/test_view.py
```diff
@@ -228,17 +228,17 @@ def test_add_column_to_view(self, reset_db: None, test_tbl: pxt.Table, reload_te
         v = pxt.create_view('test_view', t, additional_columns={'v1': pxt.Int})
         v.add_computed_column(vcol='xxx')
         assert 'vcol' in v.columns()
-        assert v.select(v.vcol).collect()[0]['vcol'] == 'xxx'
+        assert v.select(v.vcol).head()[0]['vcol'] == 'xxx'
```
mkornacker: This is the wrong way to address the problem. You need to use an order_by(). head() only guarantees that retrieval order matches insertion order, but the insertion order itself could change (eg, the plan changes).

aaron-siegel: I understand that but it shouldn't matter here, right? All that matters is that we get the rows back in a *consistent* order and head() should guarantee that. We're just doing internal comparisons - not checking against some ground truth.

mkornacker: Fair enough. We could also require order_by() universally, that way you don't have to think about it.

aaron-siegel: Sure... I switched it to order_by

PR #1023, opened 2025-12-22
Allow `similarity(image=...)` to accept a filename or URL instead of a PIL image object

aaron-siegel:


pixeltable/exprs/column_ref.py
```diff
@@ -233,8 +234,14 @@ def similarity(
...
             else:
-                if not isinstance(image, PIL.Image.Image):
-                    raise excs.Error(f'similarity(image=...): expected `PIL.Image.Image`; got `{type(image).__name__}`')
+                if not isinstance(image, (str, PIL.Image.Image)):
+                    raise excs.Error(
+                        f'similarity(image=...): expected `str` or `PIL.Image.Image`; got `{type(image).__name__}`'
+                    )
+                if isinstance(image, str):
+                    image_path = fetch_url(image, allow_local_file=True)
+                    image = PIL.Image.open(image_path)
```
mkornacker, line 243: This is changing the type of image, best to use a new variable.

It's disappointing that mypy doesn't catch this.

pixeltable/utils/__init__.py
```diff
@@ -55,3 +59,40 @@ def parse_local_file_path(file_or_url: str) -> Path | None:
         return Path(urllib.parse.unquote(urllib.request.url2pathname(parsed.path)))
     else:
         return None
+
+
+def fetch_url(url: str, allow_local_file: bool = False) -> Path:
```
mkornacker: Move into utils/http.py.

pixeltable/exprs/column_ref.py
aaron-siegel: It's not changing the type of the variable. Previously the type was `str | Image`, and this simply narrows it to `Image`. The effect on the type is exactly the same as if I'd said `assert not isinstance(image, str)`.

It's entirely stylistically correct, and there's nothing for mypy to catch

pixeltable/utils/__init__.py
aaron-siegel: Done.

PR #1024, opened 2025-12-22
PXT-875 Run perf tests on a dedicated larger runner

sergey-mkhitaryan:
The errors that we are getting at present indicate that the standard runners may have insufficient resources.


.github/workflows/stress-tests.yml
```diff
@@ -68,7 +68,7 @@ jobs:
           path: ~/.pixeltable/random-ops.log
 
   performance-tests:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-x64-t4
```
aaron-siegel: are we planning to update this to something other than x64-t4 (very expensive)?

PR #1028, opened 2025-12-24
Add BFL FLUX integration for image generation and editing

pierrebrunelle:
## Summary

Adds integration with [Black Forest Labs (BFL) FLUX API](https://docs.bfl.ai/) for state-of-the-art image generation and editing.

## Features

- **`bfl.generate()`**: Text-to-image generation with FLUX models
- **`bfl.edit()`**: Image editing with text prompts
- **`bfl.fill()`**: Inpainting with mask support
- **`bfl.expand()`**: Outpainting to extend images

## Implementation _a la Reve_ (no Python SDK)

- Async HTTP client with polling (aiohttp)
- Rate limiting via `resource_pool='request-rate:bfl'`
- Proper error handling (rate limits, moderation, timeouts)


pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,512 @@
...
+    'flux-kontext-pro': '/v1/flux-kontext-pro',
+    'flux-kontext-max': '/v1/flux-kontext-max',
+}
+
+
+@pxt.udf(resource_pool='request-rate:bfl')
+async def generate(
+    prompt: str,
+    *,
+    model: Literal[
```
aaron-siegel: Best to just take this as `model: str` rather than hardcode whatever models exist today.

Best not to specify a default for `model`.

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,512 @@
...
+    """
+    Generate an image from a text prompt using FLUX models.
+
+    This UDF wraps the BFL FLUX API endpoints. For more information, refer to the official
+    [API documentation](https://docs.bfl.ai/flux_2/flux2_text_to_image).
+
+    Args:
+        prompt: Text description of the image to generate.
+        model: FLUX model to use. Options:
+            - 'flux-2-pro': Best balance of speed and quality (default)
```
aaron-siegel: Incorrect markdown (bulleted lists must be preceded by a blank line)

But I'd argue against hard-coding a list of models in our docs. We don't do that for any other provider - unless there's a specific reason why BFL should be treated differently in this regard.

Just link to the models section of the BFL docs and let them do the work of keeping it up to date.

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,512 @@
...
+def _(api_key: str) -> _BflClient:
+    return _BflClient(api_key=api_key)
+
+
+def _client() -> _BflClient:
+    return Env.get().get_client('bfl')
+
+
+# Model endpoint mapping
+_MODEL_ENDPOINTS = {
```
aaron-siegel: Why do we need this dict? It seems like unnecessary boilerplate when

`f'/v1/{model}'`

would suffice (and be forward-compatible if new models are introduced)

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,512 @@
...
+        ... )
+
+        Generate a square image with specific seed for reproducibility:
+
+        >>> t.add_computed_column(
+        ...     image=bfl.generate(t.prompt, seed=42, width=1024, height=1024)
+        ... )
+    """
+    if model not in _MODEL_ENDPOINTS:
+        raise pxt.Error(f'Unknown model: {model}. Available: {list(_MODEL_ENDPOINTS.keys())}')
```
aaron-siegel: Quote the model any time you print it in error messages: `{model!r}`

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class TestBfl:
+    @pytest.mark.remote_api
+    def test_generate(self, reset_db: None) -> None:
+        skip_test_if_no_client('bfl')
+        from pixeltable.functions.bfl import generate
+
+        t = pxt.create_table('test_tbl', {'prompt': pxt.String})
+        t.add_computed_column(image=generate(t.prompt, model='flux-pro-1.1', width=512, height=512))
```
aaron-siegel: Always make sure every UDF has test instances that

- Test the base UDF with no optional arguments
- Test every optional parameter

This will help avoid bugs/regressions in maintaining the UDF parameter lists.

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,80 @@
...
+        validate_update_status(t.insert(prompt='A beautiful mountain landscape at sunset'), expected_rows=1)
+
+        result = t.select(t.image).collect()
+        assert result['image'][0] is not None
+        # Check that we got an image with the expected dimensions
+        img = result['image'][0]
+        assert img.width == 512
+        assert img.height == 512
+
+    @pytest.mark.remote_api
```
aaron-siegel: Instead of decorating each function separately with the pytest marker, just decorate the class once.

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,80 @@
...
+        result = t.select(t.image).collect()
+        assert result['image'][0] is not None
+        img = result['image'][0]
+        assert img.width == 1024
+        assert img.height == 1024
+
+    @pytest.mark.remote_api
+    def test_edit(self, reset_db: None) -> None:
+        skip_test_if_no_client('bfl')
+        from pathlib import Path
```
aaron-siegel: Belongs at top of file

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,512 @@
...
+    if guidance is not None:
+        payload['guidance'] = guidance
+    if seed is not None:
+        payload['seed'] = seed
+    if safety_tolerance is not None:
+        payload['safety_tolerance'] = safety_tolerance
+    if output_format is not None:
+        payload['output_format'] = output_format
+
+    return await _client().generate('/v1/flux-pro-1.0-fill', payload)
```
aaron-siegel: We should take a model name as a parameter. Even if this is the only model that works for this today, there might be newer models/versions introduced in the future. Hardcoding a single model risks causing us backward-compatibility headaches in the future.

aaron-siegel (changes requested):

pixeltable/functions/bfl.py
pierrebrunelle: Done - removed the dict entirely. Now uses `f'/v1/{model}'` directly.

pixeltable/functions/bfl.py
pierrebrunelle: Done - changed to `model: str` with no default.

pixeltable/functions/bfl.py
pierrebrunelle: Done - removed hardcoded model list, now links to BFL docs instead.

pixeltable/functions/bfl.py
pierrebrunelle: Done - removed the validation entirely since we use dynamic endpoints now.

tests/functions/test_bfl.py
pierrebrunelle: Noted - will add more comprehensive test coverage for optional params in a follow-up.

tests/functions/test_bfl.py
pierrebrunelle: Done - moved decorator to class level.

tests/functions/test_bfl.py
pierrebrunelle: Done - moved `from pathlib import Path` to top of file.

pixeltable/functions/bfl.py
pierrebrunelle: Done - added `model: str` parameter to `fill()` and `expand()` functions.

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,486 @@
...
+    """
+    Client for interacting with the BFL API. Maintains a long-lived HTTP session to the service.
+    """
+
+    api_key: str
+    session: aiohttp.ClientSession
+
+    def __init__(self, api_key: str):
+        self.api_key = api_key
+        self.session = Env.get().event_loop.run_until_complete(self._start_session())
```
aaron-siegel: @mkornacker should weigh in on whether this will play nice with our async execution engine.

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,486 @@
...
+                                raise BflUnexpectedError(f'BFL task {task_id} not found (may have expired)')
+                            case 'Pending':
+                                await asyncio.sleep(poll_interval)
+                                elapsed += poll_interval
+                            case _:
+                                # Unknown status, wait and retry
+                                await asyncio.sleep(poll_interval)
+                                elapsed += poll_interval
+                    case 429:
+                        retry_after_seconds = None
```
aaron-siegel: Isn't this (next 6 lines) identical to the preceding logic in _submit_task()?

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,486 @@
...
+                        retry_after_header = resp.headers.get('Retry-After')
+                        if retry_after_header is not None and re.fullmatch(r'\d{1,2}', retry_after_header):
+                            retry_after_seconds = int(retry_after_header)
+                        _logger.info(f'BFL task {task_id} polling rate limited, retry after: {retry_after_header}')
+                        raise BflRateLimitedError(
+                            f'BFL task {task_id} polling rate limited (429). retry-after:{retry_after_seconds}'
+                        )
+                    case _:
+                        error_text = await resp.text()
```
aaron-siegel, line 133: Same here - best to refactor this code to remove duplicate logic

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,486 @@
...
+    if height is not None:
+        payload['height'] = height
+    if seed is not None:
+        payload['seed'] = seed
+    if safety_tolerance is not None:
+        payload['safety_tolerance'] = safety_tolerance
+    if output_format is not None:
+        payload['output_format'] = output_format
+
+    # flux-2-flex specific parameters
```
aaron-siegel: I'd argue this shouldn't be our business; the logger warning will be silent to most users, so from the user perspective we're just swallowing parameters. What if BFL releases a new model tomorrow (flux-2.1-flex, or whatever) that also uses these parameters?

"Do not abstract away the AI" - just pass whatever the user gives us to BFL and let it fail if the parameters are invalid.

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,486 @@
...
+        guidance: Guidance scale 1.5-10 (flux-2-flex only). Default 4.5.
+
+    Returns:
+        A generated PIL Image.
+
+    Examples:
+        Add a computed column to generate images from prompts:
+
+        >>> t.add_computed_column(
+        ...     image=bfl.generate(t.prompt, model='flux-2-pro', width=1920, height=1080)
```
aaron-siegel: Maybe give an example without explicit width/height.

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,486 @@
...
+        payload['height'] = height
+    if seed is not None:
+        payload['seed'] = seed
+    if safety_tolerance is not None:
+        payload['safety_tolerance'] = safety_tolerance
+    if output_format is not None:
+        payload['output_format'] = output_format
+
+    # flux-2-flex specific parameters
+    if model == 'flux-2-flex':
```
aaron-siegel: Same comment as above.

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,77 @@
...
+        validate_update_status(t.insert(prompt='A beautiful mountain landscape at sunset'), expected_rows=1)
+
+        result = t.select(t.image).collect()
+        assert result['image'][0] is not None
+        # Check that we got an image with the expected dimensions
+        img = result['image'][0]
+        assert img.width == 512
+        assert img.height == 512
+
+    def test_generate_flux2(self, reset_db: None) -> None:
```
aaron-siegel: Why duplicate the test rather than use `@pytest.parametrize`?

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,77 @@
...
+        img = result['image'][0]
+        assert img.width == 1024
+        assert img.height == 1024
+
+    def test_edit(self, reset_db: None) -> None:
+        skip_test_if_no_client('bfl')
+        from pixeltable.functions.bfl import edit
+
+        # Use a test image
+        img_path = Path(__file__).parent.parent / 'data' / 'images' / 'test_image_rgb_1024x576.png'
```
aaron-siegel: I don't see this file either in the repo or in the PR.

Also, use `utils.TEST_DIR` in preference to `Path(__file__).parent.parent`.

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,77 @@
...
+        if not img_path.exists():
```
aaron-siegel: Please don't do this. "Fallback logic" makes the tests environment-dependent / nondeterministic and makes debugging harder (and tends to create headaches for people operating CI)

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test_generate_with_seed_reproducibility(self, reset_db: None) -> None:
+        """Test that using the same seed produces consistent results."""
+        skip_test_if_no_client('bfl')
+        from pixeltable.functions.bfl import generate
+
+        t = pxt.create_table('test_tbl', {'prompt': pxt.String, 'seed': pxt.Int})
+        t.add_computed_column(image=generate(t.prompt, model='flux-pro-1.1', width=512, height=512, seed=t.seed))
+
+        # Same prompt and seed should give similar results
```
aaron-siegel: Ok, but in what sense is the test actually validating that?

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,77 @@
...
+
+import pytest
+
+import pixeltable as pxt
+
+from ..utils import skip_test_if_no_client, validate_update_status
+
+
+@pytest.mark.remote_api
+class TestBfl:
```
aaron-siegel, line 15: I don't see tests for fill() and expand()?

aaron-siegel (changes requested):

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,77 @@
+from pathlib import Path
+
+import pytest
+
+import pixeltable as pxt
+
+from ..utils import skip_test_if_no_client, validate_update_status
+
+
+@pytest.mark.remote_api
```
aaron-siegel, line 12: Mark with reruns (as specified in New Integration Checklist)

Also per New Integration Checklist, please add relevant entry to `.github/workflows/pytest.yml`.

aaron-siegel (changes requested):

pixeltable/env.py
```diff
@@ -837,6 +837,7 @@ def __register_packages(self) -> None:
         """Declare optional packages that are utilized by some parts of the code."""
         self.__register_package('accelerate')
         self.__register_package('anthropic')
+        self.__register_package('bfl')
```
aaron-siegel: Wait a second. This just occurred to me.

Why are we registering a package that doesn't exist? (Or if it does, isn't in our pyproject?)

pixeltable/functions/bfl.py
pierrebrunelle: Done - extracted to `_handle_rate_limit()` helper method that's used by both `_submit_task()` and `_poll_result()`.

pixeltable/functions/bfl.py
pierrebrunelle: Done - removed the model-specific validation. Now all parameters are passed directly to BFL API.

pixeltable/functions/bfl.py
pierrebrunelle: Done - added example without width/height: `t.add_computed_column(image=bfl.generate(t.prompt, model='flux-2-pro'))`

tests/functions/test_bfl.py
pierrebrunelle: Done - now using `@pytest.mark.parametrize` for the generate tests with different models.

tests/functions/test_bfl.py
pierrebrunelle: Done - now using `TESTS_DIR / 'data' / 'images' / 'sewing-threads.heic'` which exists in the repo.

tests/functions/test_bfl.py
pierrebrunelle: Done - removed all fallback logic. Now uses a deterministic path and skips if not found.

tests/functions/test_bfl.py
pierrebrunelle: Done - added tests for `fill()` and `expand()` UDFs, each with both a base test (required args only) and a test with optional parameters.

tests/functions/test_bfl.py
pierrebrunelle: Done - added `@rerun(reruns=3, reruns_delay=8)` decorator to the test class.

pixeltable/env.py
pierrebrunelle: Good catch - removed. The BFL integration uses `aiohttp` (already a dependency), not a separate `bfl` package.

tests/functions/test_bfl.py
pierrebrunelle: Done - added tests for base UDF (no optional args) and individual tests for each optional parameter (seed, safety_tolerance, output_format).

tests/functions/test_bfl.py
pierrebrunelle: Good point - the seed test now only verifies that image generation works with a seed parameter. True reproducibility testing would require multiple runs with the same seed, which would be expensive. Renamed to `test_generate_with_seed`.

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,236 @@
...
+
+        validate_update_status(t.insert(prompt='A beautiful mountain landscape at sunset'), expected_rows=1)
+
+        result = t.select(t.image).collect()
+        assert result['image'][0] is not None
+        img = result['image'][0]
+        assert img.width == width
+        assert img.height == height
+
+    def test_generate_base(self, reset_db: None) -> None:
```
aaron-siegel: No need to have an extra test for this. Just insert 2 columns in the main test. That's how it's done in all our other integrations.

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,236 @@
...
+
+        validate_update_status(t.insert(prompt='A red apple on a wooden table'), expected_rows=1)
+
+        result = t.select(t.image).collect()
+        assert result['image'][0] is not None
+        img = result['image'][0]
+        assert img.width == 512
+        assert img.height == 512
+
+    def test_generate_with_safety_tolerance(self, reset_db: None) -> None:
```
aaron-siegel: I don't see why these all need to be separate tests. Why not a single test with 2 columns
- one that has no optional parameters
- one that uses every optional parameter

(or if there are mutually incompatible optional parameters, then 3 columns, or whatever).

Please look at how this is done in our other integrations and follow the established pattern, unless there's a specific reason to deviate here.

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,236 @@
...
+
+        result = t.select(t.image).collect()
+        assert result['image'][0] is not None
+
+    def test_edit(self, reset_db: None) -> None:
+        """Test image editing with edit UDF."""
+        skip_test_if_no_client('bfl')
+        from pixeltable.functions.bfl import edit
+
+        if not TEST_IMAGE_PATH.exists():
```
aaron-siegel: Avoid spurious pytest.skip() invocations that could mask a real error.

If this happens, it's a problem and we want the test to fail so that we're alerted to do something about it.

There are more than one instance of this in the code, I haven't left a comment on all of them, but please check to be sure we catch them all.

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,236 @@
...
+
+        result = t.select(t.edited).collect()
+        assert result['edited'][0] is not None
+
+    def test_fill(self, reset_db: None) -> None:
+        """Test inpainting with fill UDF."""
+        skip_test_if_no_client('bfl')
+        from pixeltable.functions.bfl import fill
+
+        import PIL.Image
```
aaron-siegel: move to top of file

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,236 @@
...
+                for y in range(h // 4, 3 * h // 4):
+                    mask.putpixel((x, y), 255)
+
+        t = pxt.create_table('test_tbl', {'image': pxt.Image, 'mask': pxt.Image, 'prompt': pxt.String})
+        t.add_computed_column(filled=fill(t.prompt, t.image, t.mask, model='flux-pro-1.0-fill'))
+
+        # Save temp images for insertion
+        import tempfile
+
+        with tempfile.NamedTemporaryFile(suffix='.png', delete=False) as img_f:
```
aaron-siegel: What's the point of this? Why not just insert the PIL images directly?

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,236 @@
...
+            w, h = img.size
+            for x in range(w // 4, 3 * w // 4):
+                for y in range(h // 4, 3 * h // 4):
+                    mask.putpixel((x, y), 255)
+
+        t = pxt.create_table('test_tbl', {'image': pxt.Image, 'mask': pxt.Image, 'prompt': pxt.String})
+        t.add_computed_column(filled=fill(t.prompt, t.image, t.mask, model='flux-pro-1.0-fill'))
+
+        # Save temp images for insertion
+        import tempfile
```
aaron-siegel: move to top of file

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,236 @@
...
+            mask_path = mask_f.name
+
+        validate_update_status(
+            t.insert(image=img_path, mask=mask_path, prompt='Fill with a beautiful pattern'), expected_rows=1
+        )
+
+        result = t.select(t.filled).collect()
+        assert result['filled'][0] is not None
+
+    def test_fill_base(self, reset_db: None) -> None:
```
aaron-siegel: No need for a separate test, especially one that is 95% code duplication (pure slop).

tests/functions/test_bfl.py
```diff
@@ -0,0 +1,236 @@
...
+        t.add_computed_column(expanded=expand(t.prompt, t.image, model='flux-pro-1.0-expand', left=128, right=128))
+
+        validate_update_status(
+            t.insert(image=str(TEST_IMAGE_PATH), prompt='Continue the scene to the sides'), expected_rows=1
+        )
+
+        result = t.select(t.expanded).collect()
+        assert result['expanded'][0] is not None
+
+    def test_expand_base(self, reset_db: None) -> None:
```
aaron-siegel: No need for a separate test.

aaron-siegel (changes requested):

tests/functions/test_bfl.py
pierrebrunelle: Done - merged into single test with 2 columns.

tests/functions/test_bfl.py
pierrebrunelle: Done - consolidated all generate tests into single test with 2 columns (base + all optional params).

tests/functions/test_bfl.py
pierrebrunelle: Done - removed all pytest.skip() calls.

tests/functions/test_bfl.py
pierrebrunelle: Done - moved to top of file.

tests/functions/test_bfl.py
pierrebrunelle: Done - now inserting PIL images directly.

tests/functions/test_bfl.py
pierrebrunelle: Done - moved to top of file.

tests/functions/test_bfl.py
pierrebrunelle: Done - merged into single test with 2 columns.

tests/functions/test_bfl.py
pierrebrunelle: Done - merged into single test with 2 columns.

pixeltable/functions/bfl.py
aaron-siegel: cc @mkornacker

pixeltable/functions/bfl.py
```diff
@@ -0,0 +1,474 @@
...
+    def __init__(self, api_key: str):
+        self.api_key = api_key
+        self.session = Env.get().event_loop.run_until_complete(self._start_session())
+        atexit.register(lambda: asyncio.run(self.session.close()))
+
+    async def _start_session(self) -> aiohttp.ClientSession:
+        # Don't set base_url because polling_url and image URLs are absolute
+        return aiohttp.ClientSession()
+
+    @staticmethod
```
aaron-siegel: `@staticmethod` should usually be avoided - it should only be used if there's a specific reason to use it in place of `@classmethod`. Here I don't see any, so I'd argue this should be a `@classmethod`

PR #1029, opened 2025-12-24
Add Jina AI integration for embeddings and reranking

pierrebrunelle:
- Embedding UDF (`jina.embeddings`)
- Rerank UDF (`jina.rerank`)

## Usage

```python
from pixeltable.functions import jina

# Embeddings
tbl.add_computed_column(
    embed=jina.embeddings(tbl.text, model='jina-embeddings-v3', task='retrieval.passage')
)

# Reranking
tbl.add_computed_column(
    reranked=jina.rerank(tbl.query, tbl.documents, model='jina-reranker-v2-base-multilingual', top_n=5)
)
```

### DIFF
--- a/docs/release/howto/providers/working-with-jina.ipynb
+++ b/docs/release/howto/providers/working-with-jina.ipynb
@@ -1,535 +1,564 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {},
-      "source": [
-        "# Working with Jina AI in Pixeltable\n",
-        "\n",
-        "Pixeltable's Jina AI integration enables you to access state-of-the-art embedding and reranker models via the Jina AI API.\n",
-        "\n",
-        "### Prerequisites\n",
-        "- A Jina AI account with an API key (https://jina.ai/)\n",
-        "\n",
-        "### Important notes\n",
-        "\n",
-        "- Jina AI usage may incur costs based on your Jina AI plan.\n",
-        "- Be mindful of sensitive data and consider security measures when integrating with external services.\n"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {},
-      "source": [
-        "First you'll need to install Pixeltable and set up your Jina AI API key.\n"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": null,
-      "metadata": {},
-      "outputs": [],
-      "source": [
-        "%pip install -qU pixeltable"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 17,
-      "metadata": {},
-      "outputs": [],
-      "source": [
-        "import os\n",
-        "import getpass\n",
-        "\n",
-        "if 'JINA_API_KEY' not in os.environ:\n",
-        "    os.environ['JINA_API_KEY'] = getpass.getpass('Enter your Jina AI API key: ')\n"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {},
-      "source": [
-        "Now let's create a Pixeltable directory to hold the tables for our demo.\n"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 18,
-      "metadata": {},
-      "outputs": [
-        {
-          "name": "stdout",
-          "output_type": "stream",
-          "text": [
-            "Created directory 'jina_demo'.\n"
-          ]
-        },
-        {
-          "data": {
-            "text/plain": [
-              "<pixeltable.catalog.dir.Dir at 0x1454c53d0>"
-            ]
-          },
-          "execution_count": 18,
-          "metadata": {},
-          "output_type": "execute_result"
-        }
-      ],
-      "source": [
-        "import pixeltable as pxt\n",
-        "\n",
-        "# Remove the 'jina_demo' directory and its contents, if it exists\n",
-        "pxt.drop_dir('jina_demo', force=True)\n",
-        "pxt.create_dir('jina_demo')\n"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {},
-      "source": [
-        "## Text Embeddings\n",
-        "\n",
-        "Jina AI provides frontier multilingual embedding models for semantic search and RAG applications. The `jina-embeddings-v3` model supports 89+ languages and achieves state-of-the-art performance.\n"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 19,
-      "metadata": {},
-      "outputs": [
-        {
-          "name": "stdout",
-          "output_type": "stream",
-          "text": [
-            "Created table 'documents'.\n",
-            "Added 0 column values with 0 errors.\n"
-          ]
-        },
-        {
-          "data": {
-            "text/plain": [
-              "No rows affected."
-            ]
-          },
-          "execution_count": 19,
-          "metadata": {},
-          "output_type": "execute_result"
-        }
-      ],
-      "source": [
-        "from pixeltable.functions import jina\n",
-        "\n",
-        "# Create a table for document embeddings\n",
-        "docs_t = pxt.create_table('jina_demo.documents', {'text': pxt.String})\n",
-        "\n",
-        "# Add computed column with Jina embeddings\n",
-        "# task='retrieval.passage' optimizes embeddings for documents to be searched\n",
-        "docs_t.add_computed_column(\n",
-        "    embedding=jina.embeddings(\n",
-        "        docs_t.text,\n",
-        "        model='jina-embeddings-v3',\n",
-        "        task='retrieval.passage'\n",
-        "    )\n",
-        ")\n"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 20,
-      "metadata": {},
-      "outputs": [
-        {
-          "name": "stdout",
-          "output_type": "stream",
-          "text": [
-            "Inserting rows into `documents`: 6 rows [00:00, 1394.00 rows/s]\n",
-            "Inserted 6 rows with 0 errors.\n"
-          ]
-        },
-        {
-          "data": {
-            "text/plain": [
-              "6 rows inserted, 12 values computed."
-            ]
-          },
-          "execution_count": 20,
-          "metadata": {},
-          "output_type": "execute_result"
-        }
-      ],
-      "source": [
-        "# Insert some sample documents\n",
-        "documents = [\n",
-        "    \"The Mediterranean diet emphasizes fish, olive oil, and vegetables, believed to reduce chronic diseases.\",\n",
-        "    \"Photosynthesis in plants converts light energy into glucose and produces essential oxygen.\",\n",
-        "    \"20th-century innovations, from radios to smartphones, centered on electronic advancements.\",\n",
-        "    \"Rivers provide water, irrigation, and habitat for aquatic species, vital for ecosystems.\",\n",
-        "    \"Apple's conference call to discuss fourth fiscal quarter results is scheduled for Thursday, November 2, 2023.\",\n",
-        "    \"Shakespeare's works, like 'Hamlet' and 'A Midsummer Night's Dream,' endure in literature.\"\n",
-        "]\n",
-        "\n",
-        "docs_t.insert({'text': doc} for doc in documents)\n"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 21,
-      "metadata": {},
-      "outputs": [
-        {
-          "data": {
-            "text/html": [
-              "<table border=\"1\" class=\"dataframe\">\n",
-              "  <thead>\n",
-              "    <tr style=\"text-align: right;\">\n",
-              "      <th>text</th>\n",
-              "      <th>embedding</th>\n",
-              "    </tr>\n",
-              "  </thead>\n",
-              "  <tbody>\n",
-              "    <tr>\n",
-              "      <td>The Mediterranean diet emphasizes fish, olive oil, and vegetables, believed to reduce chronic diseases.</td>\n",
-              "      <td>[-0.004  0.03   0.057 -0.048 -0.086 -0.037 ...  0.013  0.008 -0.007  0.012  0.    -0.019]</td>\n",
-              "    </tr>\n",
-              "    <tr>\n",
-              "      <td>Photosynthesis in plants converts light energy into glucose and produces essential oxygen.</td>\n",
-              "      <td>[ 0.028  0.112  0.031 -0.085 -0.082 -0.1   ...  0.045  0.012  0.003  0.015 -0.019  0.012]</td>\n",
-              "    </tr>\n",
-              "    <tr>\n",
-              "      <td>20th-century innovations, from radios to smartphones, centered on electronic advancements.</td>\n",
-              "      <td>[ 0.026  0.022 -0.036  0.026  0.016 -0.04  ...  0.021  0.006  0.012 -0.008 -0.002 -0.027]</td>\n",
-              "    </tr>\n",
-              "  </tbody>\n",
-              "</table>"
-            ],
-            "text/plain": [
-              "                                                text  \\\n",
-              "0  The Mediterranean diet emphasizes fish, olive ...   \n",
-              "1  Photosynthesis in plants converts light energy...   \n",
-              "2  20th-century innovations, from radios to smart...   \n",
-              "\n",
-              "                                           embedding  \n",
-              "0  [-0.00416068, 0.02968253, 0.05739661, -0.04815...  \n",
-              "1  [0.02841411, 0.11249779, 0.03135551, -0.084914...  \n",
-              "2  [0.02606024, 0.02163549, -0.03598222, 0.025623...  "
-            ]
-          },
-          "execution_count": 21,
-          "metadata": {},
-          "output_type": "execute_result"
-        }
-      ],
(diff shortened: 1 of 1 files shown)


aaron-siegel:
This has many of the same issues as #1028 - please audit for similar issues as #1028 and then I will code review (so that I'm not just leaving the same review twice)

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+    embedding_type: Literal['float', 'base64', 'binary', 'ubinary'] | None = None,
+) -> ts.ArrayType:
+    # If dimensions is explicitly specified, use it
+    if dimensions is not None:
+        return ts.ArrayType((dimensions,), dtype=ts.FloatType(), nullable=False)
+    # Otherwise, look up the default for this model
+    dim = _embedding_dimensions_cache.get(model)
+    if dim is None:
+        return ts.ArrayType((None,), dtype=ts.FloatType(), nullable=False)
+    return ts.ArrayType((dim,), dtype=ts.FloatType(), nullable=False)
```
sergey-mkhitaryan: I didn't have a chance to test this out yet, but do you know what happens if a non-float embedding is selected? Say if it's base64, how do we go from a base64 string to an Array? And does it work as an embedding index?

pixeltable/env.py
```diff
@@ -848,6 +848,7 @@ def __register_packages(self) -> None:
         self.__register_package('google.cloud.storage', library_name='google-cloud-storage')
         self.__register_package('google.genai', library_name='google-genai')
         self.__register_package('groq')
+        self.__register_package('jina')
```
aaron-siegel: Same comment as the BFL PR, this is a nonexistent package!

Or rather, in this case it exists, but it's something completely different:
```
Jina-serve is a framework for building and deploying AI services that communicate via gRPC, HTTP and WebSockets. Scale your services from local development to production while focusing on your core logic.
```

The Jina SDK package is `jinaai`, not `jina`. That's the one we should register, but only if we're actually using it (see next comment).

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+from pixeltable.utils.code import local_public_names
+
+_logger = logging.getLogger('pixeltable')
+
+# Default embedding dimensions for Jina models
+_embedding_dimensions_cache: dict[str, int] = {
+    'jina-embeddings-v4': 2048,
+    'jina-clip-v2': 1024,
+    'jina-embeddings-v3': 1024,
+    'jina-colbert-v2': 128,  # ColBERT models return multi-vector, dimension per token
```
aaron-siegel: Are we actually able to support this? If not, what's the reason to include it in the dict?

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+
+class JinaRateLimitedError(Exception):
+    pass
+
+
+class JinaUnexpectedError(Exception):
+    pass
+
+
+class _JinaClient:
```
aaron-siegel, line 48: Is there a reason not to use this?
https://github.com/jina-ai/jinaai-py
(I didn't investigate, just did a Google search)

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+    Request throttling:
+    Applies the rate limit set in the config (section `jina`, key `rate_limit`). If no rate
+    limit is configured, uses a default of 500 RPM.
+
+    Args:
+        input: The text to embed.
+        model: The Jina embedding model to use. See available models at
+            <https://jina.ai/embeddings/>.
+        task: Task-specific embedding optimization. Options:
+            - `retrieval.query`: For search queries
```
aaron-siegel, line 126: Invalid markdown

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+) -> Batch[pxt.Array[(None,), pxt.Float]]:
+    """
+    Creates embedding vectors for the input text using Jina AI embedding models.
+
+    Equivalent to the Jina AI embeddings API endpoint.
+    For additional details, see: <https://jina.ai/embeddings/>
+
+    Request throttling:
+    Applies the rate limit set in the config (section `jina`, key `rate_limit`). If no rate
+    limit is configured, uses a default of 500 RPM.
```
aaron-siegel: Isn't 600 rpm the default?

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+        input: The text to embed.
+        model: The Jina embedding model to use. See available models at
+            <https://jina.ai/embeddings/>.
+        task: Task-specific embedding optimization. Options:
+            - `retrieval.query`: For search queries
+            - `retrieval.passage`: For documents/passages to be searched
+            - `separation`: For clustering/separation tasks
+            - `classification`: For classification tasks
+            - `text-matching`: For semantic similarity
+        dimensions: Output embedding dimensions. Model-dependent; check docs for valid values.
```
aaron-siegel: Mention that this is optional and the model default will be used if not specified

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+        late_chunking: Enable late chunking for long documents.
+        embedding_type: Output format. Options: `float` (default), `base64`, `binary`, `ubinary`.
+            Only `float` is currently supported in Pixeltable.
```
aaron-siegel: Then why do we have this as a parameter?

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+        ...     embed=jina.embeddings(tbl.text, model='jina-embeddings-v3', task='retrieval.passage')
+        ... )
+
+        Add an embedding index:
+
+        >>> tbl.add_embedding_index('text', string_embed=jina.embeddings.using(model='jina-embeddings-v3'))
+    """
+    cl = _client()
+
+    payload: dict[str, Any] = {'model': model, 'input': list(input)}
```
aaron-siegel: list() is superfluous (`Batch[T]` *is* an annotated list)

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+async def embeddings(
+    input: Batch[str],
+    *,
+    model: str = 'jina-embeddings-v3',
+    task: Literal['retrieval.query', 'retrieval.passage', 'separation', 'classification', 'text-matching']
+    | None = None,
+    dimensions: int | None = None,
+    late_chunking: bool | None = None,
+    embedding_type: Literal['float', 'base64', 'binary', 'ubinary'] | None = None,
+) -> Batch[pxt.Array[(None,), pxt.Float]]:
```
aaron-siegel: Use explicit numpy type

pxt.Array[(None,), np.float32]

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+    if embedding_type is not None:
+        payload['embedding_type'] = embedding_type
+
+    result = await cl._post('/v1/embeddings', payload=payload)
+
+    # Extract embeddings from response - they come as a list of objects with 'embedding' field
+    embeddings_list = result.get('data', [])
+    # Sort by index to ensure correct order
+    embeddings_list.sort(key=lambda x: x.get('index', 0))
+    return [np.array(item['embedding'], dtype=np.float64) for item in embeddings_list]
```
aaron-siegel: We always use float32 for embeddings.

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+@embeddings.conditional_return_type
+def _(
+    model: str = 'jina-embeddings-v3',
```
aaron-siegel: conditional_return_type should never specify a default.

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+    task: Literal['retrieval.query', 'retrieval.passage', 'separation', 'classification', 'text-matching']
```
aaron-siegel: Omit any parameters that are not used.

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+    | None = None,
+    dimensions: int | None = None,
+    late_chunking: bool | None = None,
+    embedding_type: Literal['float', 'base64', 'binary', 'ubinary'] | None = None,
+) -> ts.ArrayType:
+    # If dimensions is explicitly specified, use it
+    if dimensions is not None:
+        return ts.ArrayType((dimensions,), dtype=ts.FloatType(), nullable=False)
```
aaron-siegel: dtype=np.float32 here and below.

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+
+def _client() -> _JinaClient:
+    return Env.get().get_client('jina')
+
+
+@pxt.udf(batch_size=128, resource_pool='request-rate:jina')
+async def embeddings(
+    input: Batch[str],
+    *,
+    model: str = 'jina-embeddings-v3',
```
aaron-siegel: I prefer this to be a required parameter.

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+        return ts.ArrayType((None,), dtype=ts.FloatType(), nullable=False)
+    return ts.ArrayType((dim,), dtype=ts.FloatType(), nullable=False)
+
+
+@pxt.udf(resource_pool='request-rate:jina')
+async def rerank(
+    query: str,
+    documents: list[str],
+    *,
+    model: str = 'jina-reranker-v2-base-multilingual',
```
aaron-siegel: I prefer this to be a required parameter.

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,270 @@
...
+    top_n: int | None = None,
+    return_documents: bool = False,
```
aaron-siegel: Why is this treated differently from all other parameters? Shouldn't it be
```
return_documents: bool | None = None
```
in order to defer to Jina to do the defaulting (and for consistency with everything else)?

tests/functions/test_jina.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+
+        # Test similarity search using the index
+        sim = t.text.similarity(string='What is machine learning?')
+        results = t.order_by(sim, asc=False).limit(2).select(t.text, similarity=sim).collect()
+
+        assert len(results) == 2
+        # The ML-related text should be ranked first
+        assert (
+            'machine learning' in results['text'][0].lower() or 'artificial intelligence' in results['text'][0].lower()
```
aaron-siegel: Why is this an `or` ? Isn't there only one string that contains either, and it contains both?

tests/functions/test_jina.py
```diff
@@ -0,0 +1,134 @@
...
+        assert 'index' in first_result
+        assert 'relevance_score' in first_result
+        assert 'document' in first_result  # We set return_documents=True
+        assert 'Apple' in first_result['document']
+
+        # Verify scores are in descending order
+        scores = [r['relevance_score'] for r in result['results']]
+        assert scores == sorted(scores, reverse=True)
+
+    def test_multilingual_embeddings(self, reset_db: None) -> None:
```
aaron-siegel: I don't see what this actually tests that the other tests don't.

tests/functions/test_jina.py
```diff
@@ -0,0 +1,134 @@
+import pytest
+
+import pixeltable as pxt
+
+from ..utils import rerun, skip_test_if_no_client, validate_update_status
+
+
+@pytest.mark.remote_api
+@rerun(reruns=3, reruns_delay=8)
+class TestJina:
```
aaron-siegel, line 10: Also add appropriate entries to `.github/workflows/pytest.yml` per New Integration Checklist.

pixeltable/functions/jina.py
pierrebrunelle: It's 2 y/o w/o update and not mentioned in their docs.

pixeltable/env.py
pierrebrunelle: Done - removed. The Jina integration uses `aiohttp` (already a dependency), not a separate `jina` package.

pixeltable/functions/jina.py
pierrebrunelle: Good point - removed ColBERT entries. They return multi-vector embeddings (one per token) which our Array type doesn't support.

pixeltable/functions/jina.py
pierrebrunelle: Done - added blank line before bullet list.

pixeltable/functions/jina.py
pierrebrunelle: Done - removed embedding_type parameter entirely since only float is supported.

pixeltable/functions/jina.py
pierrebrunelle: Done - removed superfluous `list()` call.

pixeltable/functions/jina.py
pierrebrunelle: Done - now using `dtype=np.float32` for embeddings.

pixeltable/functions/jina.py
pierrebrunelle: Done - removed defaults from conditional_return_type.

pixeltable/functions/jina.py
pierrebrunelle: Done - removed unused params (task, late_chunking, embedding_type) from conditional_return_type.

pixeltable/functions/jina.py
pierrebrunelle: Done - model is now required (no default) in embeddings().

pixeltable/functions/jina.py
pierrebrunelle: Done - model is now required (no default) in rerank().

pixeltable/functions/jina.py
pierrebrunelle: Done - return_documents is now handled consistently with other optional params (bool | None, only added to payload if not None).

tests/functions/test_jina.py
pierrebrunelle: Good catch - simplified to just check for 'machine learning'. The first document contains both terms anyway.

tests/functions/test_jina.py
pierrebrunelle: Done - removed the multilingual test. It was redundant since the other tests already verify embedding dimensions.

tests/functions/test_jina.py
pierrebrunelle: Done - added `JINA_API_KEY: ${{ secrets.JINA_API_KEY }}` to pytest.yml workflow.

pixeltable/functions/jina.py
pierrebrunelle: I'll leave that to @aaron-siegel

pixeltable/functions/jina.py
pierrebrunelle: Agreed - we're using the direct REST API with aiohttp rather than the old SDK. The integration doesn't depend on any Jina SDK package.

pixeltable/functions/jina.py
pierrebrunelle: Done - changed return type to use `np.float32`.

pixeltable/functions/jina.py
pierrebrunelle: Done - using `dtype=np.float32` in array creation.

pixeltable/functions/jina.py
pierrebrunelle: Done - updated docstring to clarify that dimensions is optional and the model's default will be used if not specified.

pixeltable/functions/jina.py
aaron-siegel: I don't see these changes in the diff?

pixeltable/functions/jina.py
aaron-siegel: I don't have context here - @sergey-mkhitaryan what do you mean by this? When would we get a base64 string as an embedding?

pixeltable/functions/jina.py
```diff
@@ -0,0 +1,251 @@
...
+        query: The query string to rank documents against.
+        documents: The list of documents to rerank.
+        model: The Jina reranker model to use. See available models at
+            <https://jina.ai/reranker/>.
+        top_n: Number of top results to return. If not specified, returns all documents.
+        return_documents: Whether to include the original document text in results.
+
+    Returns:
+        A dictionary containing:
+        - `results`: List of reranking results with `index` and `relevance_score`
```
aaron-siegel, line 207: Invalid markdown

pixeltable/functions/jina.py
sergey-mkhitaryan: this UDF had the following parameter when I reviewed it:
```
embedding_type: Literal['float', 'base64', 'binary', 'ubinary'] | None = None,
```
but cc0b9e73681fc09bc6454a651eac002996117641 got rid of it, so feel free to ignore this.

pixeltable/functions/jina.py
pierrebrunelle: Done - fixed to 600 RPM.

pixeltable/functions/jina.py
pierrebrunelle: Done - now using np.float32 in conditional_return_type.

pixeltable/functions/jina.py
pierrebrunelle: Done - added blank line before bullet list.

PR #1032, opened 2025-12-24
disable progress reporting in Jupyter if ipywidgets is not installed

mkornacker:


pixeltable/exec/exec_context.py
```diff
@@ -42,6 +43,10 @@ def __init__(
         else:
             self.show_progress = Env.get().verbosity >= 1 and Env.get().is_interactive()
 
+        # disable progress reporting in Jupyter if ipywidgets is not installed
+        if Env.get().is_notebook() and importlib.util.find_spec('ipywidgets') is None:
```
aaron-siegel, line 47: Print a warning in this scenario?

mkornacker: I think that's too spammy. We're already printing a warning when in pxt.init().

PR #1033, opened 2025-12-25
Lint markdown in notebooks

sergey-mkhitaryan:


.github/workflows/pytest.yml
```diff
@@ -231,6 +231,11 @@ jobs:
       - name: Lint
         if: ${{ matrix.test-category == 'lint' }}
         run: ruff check pixeltable tests tool
+      - name: Notebook lint
+        if: ${{ matrix.test-category == 'lint' }}
+        run: |
+          python -m pip install nbqa mdformat
```
aaron-siegel: Why is this necessary? the `'lint'` test category should be installing dev dependencies. And overriding uv with a direct `pip install` will circumvent uv's version lock mechanism, potentially causing problems.

PR #1039, opened 2025-12-30
Proper configurability for spaCy models

aaron-siegel:
Removes the hard-coded dependency on the `en_core_web_sm` spaCy model. Adds a `spacy_model` parameter to `document_splitter` and `string_splitter`. Moves spaCy model management out of `Env` and into a `pixeltable.utils.spacy` module.

Users will now need to install a spaCy model separately in order to use sentence separators, but this is the generally accepted practice. (The model will be validated in output_schema() and Pixeltable will print a helpful error message if it is not installed.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -175,6 +175,7 @@ dev = [
     "voyageai>=0.3.5",
     "scenedetect>=0.6.0.1",
     "opentelemetry-exporter-otlp>=1.38.0",
+    "ipywidgets>=8.0.0",
 ]
 test = [
     "pytest>=8.4",
--- a/tests/test_sample.py
+++ b/tests/test_sample.py
@@ -156,7 +156,7 @@ class TestSample:
         query = t.select().sample(fraction=0.123, seed=42)
         self._check_sample(query, t_rows * 0.123)
 
-        query = t.select().where(t.id < 200).sample(fraction=0.5)
+        query = t.select().where(t.id < 200).sample(fraction=0.5, seed=876)
         self._check_sample(query, 200 * 0.5)
 
     def test_sample_snapshot_reload(self, reset_db: None, reload_tester: ReloadTester) -> None:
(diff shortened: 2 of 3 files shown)


pixeltable/functions/string.py
```diff
@@ -818,14 +818,18 @@ def zfill(self: str, width: int) -> str:
...
+    text: Any, separators: str, *, spacy_model: str = 'en_core_web_sm'
+) -> tuple[type[pxt.iterators.ComponentIterator], dict[str, Any]]:
     """Iterator over chunks of a string. The string is chunked according to the specified `separators`.
 
     The iterator yields a `text` field containing the text of the chunk.
     Chunked text will be cleaned with `ftfy.fix_text` to fix up common problems with unicode sequences.
 
     Args:
         separators: separators to use to chunk the document. Currently the only supported option is `'sentence'`.
+        spacy_model: Name of the spaCy model to use for sentence segmentation. This parameter is ignored unless
```
mkornacker: string_splitter() **only** supports sentence segmentation. This is a required parameter.

pyproject.toml
```diff
@@ -146,7 +146,8 @@ dev = [
     "groq>=0.26.0",
     "boto3==1.36.23",
     "botocore==1.36.23",
-    "spacy>=3.8.7",
+    "spacy>=3.8.7,<3.9",
```
mkornacker, line 149: Why so narrow?

mkornacker (approved):

pixeltable/functions/string.py
aaron-siegel: Fixed. We should really unify string_splitter() and document_splitter() though.

pyproject.toml
aaron-siegel: Because the `en_core_web_sm` model dependency is pegged to a specific major/minor version of spacy. If we update spacy to 3.9 (or greater), we also have to change the URL for en_core_web_sm. The constraint here guarantees they stay in sync.

PR #1044, opened 2026-01-07
[PXT-629] Update plan is incomplete

sergey-mkhitaryan:
Update index columns for views when an update from the base table is propagated.

Introduce a test fixture that validates the state of the index columns in the end of each test. This fixture can be extended with more validations in the future.

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -2463,6 +2463,23 @@ class Catalog:
         sa_tbl = tv.store_tbl.sa_tbl
 
         # Validate that the Btree index value columns are in sync with the actual colums for latest version rows
+        # Example query:
+        # SELECT *,
+        #        tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_0 !=
+        #        tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_3 AS idx_mismatch_idx0,
+        #        LEFT(tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_1, 256) !=
+        #        tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_5 AS idx_mismatch_idx1,
+        #        LEFT(tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_2, 256) !=
+        #        tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_7 AS idx_mismatch_idx2
+        # FROM   tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2
+        # WHERE  tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.v_max = 9223372036854775807
+        #        AND ( tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_0 !=
+        #                    tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_3
+        #               OR LEFT(tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_1, 256) !=
+        #                  tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_5
+        #               OR LEFT(tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_2, 256) !=
+        #                  tbl_b58cafd251c84eb4ab5a13ff6c0f9bd2.col_7 )
+        # LIMIT 1;
         select_list: list[sql_elements.SQLCoreOperations[Any] | Literal['*']] = ['*']
         conditions: list[sql.ColumnExpressionArgument] = []
         for idx_info in tv.idxs.values():
@@ -2496,6 +2513,26 @@ class Catalog:
                 )
 
         # Validate that the index values are NULL for non-latest version rows
+        # Example query:
+        # SELECT *,
+        #        tbl_1d7bb633b5be4c57bd9070707ca4c552.col_3 IS NOT NULL  AS
+        #        idx_not_null_idx0,
+        #        tbl_1d7bb633b5be4c57bd9070707ca4c552.col_5 IS NOT NULL  AS
+        #        idx_not_null_idx1,
+        #        tbl_1d7bb633b5be4c57bd9070707ca4c552.col_7 IS NOT NULL  AS
+        #        idx_not_null_idx2,
+        #        tbl_1d7bb633b5be4c57bd9070707ca4c552.col_11 IS NOT NULL AS
+        #        idx_not_null_img_idx2,
+        #        tbl_1d7bb633b5be4c57bd9070707ca4c552.col_13 IS NOT NULL AS
+        #        idx_not_null_img_idx1
+        # FROM   tbl_1d7bb633b5be4c57bd9070707ca4c552
+        # WHERE  tbl_1d7bb633b5be4c57bd9070707ca4c552.v_max < 9223372036854775807
+        #        AND ( tbl_1d7bb633b5be4c57bd9070707ca4c552.col_3 IS NOT NULL
+        #               OR tbl_1d7bb633b5be4c57bd9070707ca4c552.col_5 IS NOT NULL
+        #               OR tbl_1d7bb633b5be4c57bd9070707ca4c552.col_7 IS NOT NULL
+        #               OR tbl_1d7bb633b5be4c57bd9070707ca4c552.col_11 IS NOT NULL
+        #               OR tbl_1d7bb633b5be4c57bd9070707ca4c552.col_13 IS NOT NULL )
+        # LIMIT 1;
         select_list.clear()
         select_list.append('*')
         conditions.clear()
@@ -2504,7 +2541,7 @@ class Catalog:
             # add it to where clause, and also to select clause for easier debugging
             condition = idx_info.val_col.sa_col != None
             conditions.append(condition)
-            select_label = f'idx_mismatch_{idx_info.name}'
+            select_label = f'idx_not_null_{idx_info.name}'
             select_list.append(condition.label(select_label))
         if len(conditions) > 0:
             stmt = (
@@ -2519,6 +2556,6 @@ class Catalog:
                 raise AssertionError(
                     'The table validation query should have returned nothing, but it returned row: '
                     f'{row._asdict()}.\nThis means that one of the indexes in {tbl._display_str()} is corrupted, i.e. '
-                    'the index value is not NULL for a non-latest version row. Look for idx_mismatch_*. '
+                    'the index value is not NULL for a non-latest version row. Look for idx_not_null_*. '
                     f'The query was:\n{stmt}'
                 )


tests/test_user.py
```diff
@@ -7,6 +7,12 @@
 
 class TestUser:
     def test_user_namespace(self, reset_db: None) -> None:
+        try:
+            self._test_user_namespace()
+        finally:
+            Env.get().user = None
```
aaron-siegel: This shouldn't be necessary, since it gets cleared as part of `reset_db`. Why was it needed here?

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
...
+
+@pytest.fixture(autouse=True)
+def validate_db_state_after_test(request: pytest.FixtureRequest) -> Generator[Any, Any, Any]:
+    """Lists all tables and views created during the test, and runs validation queries on them."""
+    if 'corrupts_db' in request.keywords:
+        yield
+        return
+
+    # Not all tests reset DB in the beginning. Ignore the tables that existed before the test started.
+    # TODO find a better way of handling this.
```
aaron-siegel: Why not build this into the existing `reset_db` fixture? Any tests that perform DB operations should be incorporating that fixture.

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
         return sentence_transformer.using(model_id='all-mpnet-base-v2')
     except ImportError:
         return None
+
+
+@pytest.fixture(autouse=True)
+def validate_db_state_after_test(request: pytest.FixtureRequest) -> Generator[Any, Any, Any]:
```
aaron-siegel: I believe `Iterator[None]` is the preferred way to express the return type here.

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
...
+    """Lists all tables and views created during the test, and runs validation queries on them."""
+    if 'corrupts_db' in request.keywords:
```
aaron-siegel: The logic here would be easier to follow if `yield` were only stated once; i.e., move this to after the principal `yield` and make it simply

```
if 'corrupts_db' in request.keywords:
    return
```

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
...
+    yield
+
+    # Collect tables and views that need to be validated
+    tables_to_check = set()
+    for tbl_path in set(pxt.list_tables('', recursive=True)) - pre_existing_tables:
+        tables_to_check.add(tbl_path)
+        tbl = pxt.get_table(tbl_path)
+        if tbl._tbl_version is None:
+            continue
+        tables_to_check.update(tbl.list_views(recursive=True))
```
aaron-siegel: Why is this necessary?

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
...
+
+    # Run the validation
+    with Env.get().engine.connect() as conn:
+        for tbl_path in tables_to_check:
+            _validate_table_or_view(tbl_path, conn)
+
+
+def _validate_table_or_view(tbl_path: str, conn: sql.Connection) -> None:
```
aaron-siegel: A view is a kind of table.

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
...
+
+    # Validate that the Btree index value columns are in sync with the actual colums for latest version rows
+    stmt = sql.select('*').select_from(sa_tbl).where(sa_tbl.c.v_max == Table.MAX_VERSION)
+    conditions = []
+    for idx in tv.idxs.values():
+        if isinstance(idx.idx, btree.BtreeIndex):
+            col = getattr(sa_tbl.c, f'col_{idx.col.id}')
+            index_val_col = getattr(sa_tbl.c, f'col_{idx.val_col.id}')
+            if idx.val_col.col_type.is_string_type():
+                conditions.append(func.left(col, btree.BtreeIndex.MAX_STRING_LEN) != index_val_col)
```
aaron-siegel: Usually we write `sql.func`, `sql.or_`, etc., rather than importing them separately.

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
...
+        return
+    tv = tbl._tbl_version.get()
+    sa_tbl = tv.store_tbl.sa_tbl
+
+    # Validate that the Btree index value columns are in sync with the actual colums for latest version rows
+    stmt = sql.select('*').select_from(sa_tbl).where(sa_tbl.c.v_max == Table.MAX_VERSION)
+    conditions = []
+    for idx in tv.idxs.values():
+        if isinstance(idx.idx, btree.BtreeIndex):
+            col = getattr(sa_tbl.c, f'col_{idx.col.id}')
```
aaron-siegel: I don't believe it should be necessary to reconstruct the column names manually - use `idx.col.store_name()`.

Also I believe `sa_tbl.c[col_name]` should be a friendlier access pattern than `getattr()`

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
...
+            index_val_col = getattr(sa_tbl.c, f'col_{idx.val_col.id}')
+            if idx.val_col.col_type.is_string_type():
+                conditions.append(func.left(col, btree.BtreeIndex.MAX_STRING_LEN) != index_val_col)
+            else:
+                conditions.append(col != index_val_col)
+    if len(conditions) > 0:
+        stmt = stmt.where(or_(*conditions)).limit(1)
+        _logger.info(f'Running index value column validation query on table {tbl_path}: {stmt}')
```
aaron-siegel: {tbl_path!r}

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
...
+        for row in conn.execute(stmt).all():
+            raise AssertionError(f"""The table validation query should have returned nothing, but it returned row:
+{row._asdict()}.
+This means that one of the indexes in table {tbl_path} is corrupted, i.e. the index value out of sync with the actual
```
aaron-siegel: {tbl_path!r}

tests/conftest.py
```diff
@@ -387,3 +388,76 @@ def all_mpnet_embed() -> pxt.Function:
...
+            index_val_col = getattr(sa_tbl.c, f'col_{idx.val_col.id}')
+            if idx.val_col.col_type.is_string_type():
+                conditions.append(func.left(col, btree.BtreeIndex.MAX_STRING_LEN) != index_val_col)
+            else:
+                conditions.append(col != index_val_col)
+    if len(conditions) > 0:
+        stmt = stmt.where(or_(*conditions)).limit(1)
+        _logger.info(f'Running index value column validation query on table {tbl_path}: {stmt}')
+        for row in conn.execute(stmt).all():
+            raise AssertionError(f"""The table validation query should have returned nothing, but it returned row:
```
aaron-siegel: The triple quotes seem like overkill here and compromise code readability; in most analogous places in the code, we use multiple single line strings, as in
```
raise AssertionError(
    'Line 1\n'
    'Line 2\n'
    'Line 3'
)
```
and I'd suggest keeping to that convention except for very long (such as multi paragraph) strings, or strings with many special characters.

tests/test_user.py
sergey-mkhitaryan: `reset_db` runs before the test itself, not after it

sergey-mkhitaryan: but judging by your comments, this is unexpected behavior, so let me try to change `reset_db` to clean before, then run the test, then validate and clean again

tests/conftest.py
sergey-mkhitaryan: didn't realize that list_tables returns views.

tests/conftest.py
sergey-mkhitaryan: this is now `tbl._display_str()` -- is there a better alternative?

tests/test_user.py
sergey-mkhitaryan: done. but this is still necessary for the validation to run after the test

tests/test_video.py
```diff
@@ -96,6 +96,9 @@ def test_query(self, reset_db: None) -> None:
         res = view_t.where(view_t.video == url).collect()
         assert len(res) == len(all_rows[all_rows.url == url])
 
+    # Bug: pxt.create_view('invalid_args'..., actually creates a corrupted view in the catalog (despite raising
```
sergey-mkhitaryan: @aaron-siegel you mentioned that it's a known problem and there's a jira for it. Is it https://pixeltable.atlassian.net/browse/PXT-842 ?

tests/conftest.py
```diff
@@ -165,6 +178,64 @@ def reset_db(init_env: None) -> None:
...
+
+    if 'corrupts_db' in request.keywords:
+        _logger.info('Skipping DB validation due to corrupts_db marker.')
+        return
+
+    # Run the validation
+    with Env.get().engine.connect() as conn:
+        for tbl_path in pxt.list_tables('', recursive=True):
+            tbl = pxt.get_table(tbl_path)
+            if tbl._tbl_version is None:
```
aaron-siegel: Why do we need this if it gets checked first thing in _validate_table?

tests/conftest.py
```diff
@@ -165,6 +178,64 @@ def reset_db(init_env: None) -> None:
...
+        if isinstance(idx.idx, btree.BtreeIndex):
+            col = sa_tbl.c[idx.col.store_name()]
+            index_val_col = sa_tbl.c[idx.val_col.store_name()]
+            if idx.val_col.col_type.is_string_type():
+                conditions.append(sql.func.left(col, btree.BtreeIndex.MAX_STRING_LEN) != index_val_col)
+            else:
+                conditions.append(col != index_val_col)
+    if len(conditions) > 0:
+        stmt = stmt.where(sql.or_(*conditions)).limit(1)
+        _logger.info(f'Running index value column validation query on table {tbl._display_str()}: {stmt}')
```
aaron-siegel: I believe you don't need the word "table" if you're using _display_str(): I think the existing code will give output like `table table 'my_tbl'`

tests/test_user.py
aaron-siegel: Set it to None in the fixture... that's what the fixture is for (it's still better than this one-off try/finally pattern here, which we have to remember to replicate in every future test that sets a user)

tests/test_video.py
aaron-siegel: That's correct.

tests/conftest.py
sergey-mkhitaryan: yeah, just an oversight.

tests/conftest.py
```diff
@@ -165,6 +168,64 @@ def reset_db(init_env: None) -> None:
...
+    yield
+
+    if 'corrupts_db' in request.keywords:
+        _logger.info('Skipping DB validation due to corrupts_db marker.')
+        return
+
+    Env.get().user = None
+
+    # Run the validation
+    with Env.get().engine.connect() as conn:
```
mkornacker: Let's move this into the catalog module as a "validate_store()" utility.

This is reaching into the catalog internals, no need to have it here.

pixeltable/catalog/catalog.py
```diff
@@ -2438,3 +2441,57 @@ def _handle_path_collision(
...
             self._drop_tbl(obj, force=if_exists == IfExistsParam.REPLACE_FORCE, is_replace=True)
         return None
+
+    @classmethod
+    def validate_store(cls) -> None:
+        """Validate the underlying store for testing purposes.
+        This function can and should be extended to perform more checks.
+        """
+        with Env.get().engine.connect() as conn:
+            for tbl_path in pxt.list_tables('', recursive=True):
```
mkornacker: Don't call public API functions for metadata access here, this should stay within catalog (otherwise you're creating circular dependencies).

pixeltable/catalog/catalog.py
```diff
@@ -2438,3 +2441,57 @@ def _handle_path_collision(
             assert isinstance(obj, Table)
             self._drop_tbl(obj, force=if_exists == IfExistsParam.REPLACE_FORCE, is_replace=True)
         return None
+
+    @classmethod
+    def validate_store(cls) -> None:
+        """Validate the underlying store for testing purposes.
+        This function can and should be extended to perform more checks.
+        """
+        with Env.get().engine.connect() as conn:
```
mkornacker: Use Env.begin_xact().

pixeltable/catalog/catalog.py
```diff
@@ -2438,3 +2441,57 @@ def _handle_path_collision(
...
+            for tbl_path in pxt.list_tables('', recursive=True):
+                tbl = pxt.get_table(tbl_path)
+                cls._validate_table(tbl, conn)
+
+    @classmethod
+    def _validate_table(cls, tbl: pxt.Table, conn: sql.Connection) -> None:
```
mkornacker: Use Env.conn instead of passing it in.

pixeltable/catalog/catalog.py
```diff
@@ -15,10 +15,13 @@
 import sqlalchemy as sql
 import sqlalchemy.exc as sql_exc
 
+import pixeltable as pxt
 from pixeltable import exceptions as excs
 from pixeltable.env import Env
+from pixeltable.index import btree
 from pixeltable.iterators import ComponentIterator
 from pixeltable.metadata import schema
+from pixeltable.metadata.schema import Table as SchemaTable
```
mkornacker: Simply import schema and then use schema.Table, easier to read. I had to look up SchemaTable, which isn't used anywhere else.

pixeltable/catalog/catalog.py
```diff
@@ -15,10 +15,13 @@
 import sqlalchemy as sql
 import sqlalchemy.exc as sql_exc
 
+import pixeltable as pxt
```
mkornacker: Remove and only use specific module imports, it's very easy to create circular dependencies.

pixeltable/catalog/catalog.py
```diff
@@ -2438,3 +2441,57 @@ def _handle_path_collision(
...
+    @classmethod
+    def _validate_table(cls, tbl: pxt.Table, conn: sql.Connection) -> None:
+        if tbl._tbl_version is None:
+            return
+        tv = tbl._tbl_version.get()
+        sa_tbl = tv.store_tbl.sa_tbl
+
+        # Validate that the Btree index value columns are in sync with the actual colums for latest version rows
+        stmt = sql.select('*').select_from(sa_tbl).where(sa_tbl.c.v_max == SchemaTable.MAX_VERSION)
+        conditions = []
```
mkornacker: Missing type decl.

pixeltable/catalog/catalog.py
```diff
@@ -2438,3 +2441,57 @@ def _handle_path_collision(
...
+        for idx in tv.idxs.values():
+            if isinstance(idx.idx, btree.BtreeIndex):
```
mkornacker: Import pixeltable.index as index and then use that (BtreeIndex is part of that module).

pixeltable/catalog/catalog.py
```diff
@@ -2438,3 +2441,57 @@ def _handle_path_collision(
...
+    def _validate_table(cls, tbl: pxt.Table, conn: sql.Connection) -> None:
+        if tbl._tbl_version is None:
+            return
+        tv = tbl._tbl_version.get()
+        sa_tbl = tv.store_tbl.sa_tbl
+
+        # Validate that the Btree index value columns are in sync with the actual colums for latest version rows
+        stmt = sql.select('*').select_from(sa_tbl).where(sa_tbl.c.v_max == SchemaTable.MAX_VERSION)
+        conditions = []
+        for idx in tv.idxs.values():
```
mkornacker: idx_info is a better variable name, because these are TableVersion.IndexInfo instances.

pixeltable/catalog/catalog.py
```diff
@@ -2438,3 +2441,57 @@ def _handle_path_collision(
...
+            if isinstance(idx.idx, btree.BtreeIndex):
+                col = sa_tbl.c[idx.col.store_name()]
```
mkornacker: Column.sa_col instead of this lookup

pixeltable/catalog/catalog.py
```diff
@@ -2438,3 +2441,57 @@ def _handle_path_collision(
...
+                index_val_col = sa_tbl.c[idx.val_col.store_name()]
+                if idx.val_col.col_type.is_string_type():
+                    conditions.append(sql.func.left(col, btree.BtreeIndex.MAX_STRING_LEN) != index_val_col)
+                else:
+                    conditions.append(col != index_val_col)
+        if len(conditions) > 0:
+            stmt = stmt.where(sql.or_(*conditions)).limit(1)
+            _logger.info(f'Running index value column validation query on {tbl._display_str()}: {stmt}')
+            for row in conn.execute(stmt).all():
```
mkornacker: To help with debugging, also print the ids and names of the columns with the discrepancies..

pixeltable/catalog/catalog.py
```diff
@@ -2438,3 +2441,57 @@ def _handle_path_collision(
...
+                raise AssertionError(
+                    f'The table validation query should have returned nothing, but it returned row: {row._asdict()}.'
+                    f' This means that one of the indexes in {tbl._display_str()} is corrupted, i.e. the index value'
+                    f' out of sync with the actual value for a current row. The query was: {stmt}'
+                )
+
+        # Validate that the index values are NULL for non-latest version rows
```
mkornacker, line 2515: Same comments apply here.

pixeltable/catalog/catalog.py
```diff
@@ -2442,56 +2442,83 @@ def _handle_path_collision(
...
+                if idx_info.val_col.col_type.is_string_type():
+                    condition = (
+                        sql.func.left(idx_info.col.sa_col, index.BtreeIndex.MAX_STRING_LEN) != idx_info.val_col.sa_col
+                    )
                 else:
-                    conditions.append(col != index_val_col)
+                    condition = idx_info.col.sa_col != idx_info.val_col.sa_col
+                conditions.append(condition)
+                select_label = f'idx_mismatch_{idx_info.name}'
+                select_list.append(condition.label(select_label))
```
mkornacker, line 2497: comment: what's the corresponding sql expr if i wrote this by hand? That helps with following the code. See load_tbl_version() for an example.

pixeltable/catalog/catalog.py
```diff
@@ -2442,56 +2442,83 @@ def _handle_path_collision(
...
         # Validate that the Btree index value columns are in sync with the actual colums for latest version rows
-        stmt = sql.select('*').select_from(sa_tbl).where(sa_tbl.c.v_max == SchemaTable.MAX_VERSION)
-        conditions = []
-        for idx in tv.idxs.values():
-            if isinstance(idx.idx, btree.BtreeIndex):
-                col = sa_tbl.c[idx.col.store_name()]
-                index_val_col = sa_tbl.c[idx.val_col.store_name()]
-                if idx.val_col.col_type.is_string_type():
-                    conditions.append(sql.func.left(col, btree.BtreeIndex.MAX_STRING_LEN) != index_val_col)
+        select_list: list[sql_elements.SQLCoreOperations[Any] | Literal['*']] = ['*']
```
mkornacker, line 2483: I think this can be simplified to list[sql.ColumElement] (and below).

sergey-mkhitaryan: just tried it, doesn't seem to be working : (

PR #1048, opened 2026-01-09
 PXT-866 Add validation for version in pixeltable uri

amithadke:


pixeltable/share/protocol/common.py
```diff
@@ -97,30 +104,26 @@ def _parse_and_validate_uri(cls, uri: str) -> dict:
...
-            else:
-                identifier, version = path_part, None
-        else:
-            identifier, version = path_part, None
+            parts = path_part.rsplit(':', 1)
+            if len(parts) == 2:
+                try:
+                    version_int = int(parts[1])
+                except ValueError:
+                    pass  # Not a valid number
```
aaron-siegel: shouldn't this be an error?

amithadke: we just stop looking for version there, assume rest of string as a path 
 for example if someone had path dir1.a:b 
 
 I just checked Path.parse will throw error here so perhaps i should replicate that here too.

aaron-siegel (approved):

PR #1049, opened 2026-01-09
Upgrade torch and transformers to resolve segfault on Windows

sergey-mkhitaryan:
There's some evidence (sample of 3) that this resolves the segfault in test_view_indices on Windows + python 3.10


aaron-siegel:
Torch 2.9 is going to break WhisperX I believe.

pyproject.toml
```diff
@@ -132,7 +132,7 @@ dev = [
     "mistune>=3.0.2,<4",
     "markitdown[pptx,docx,xlsx]>=0.1.3",
     "pyiceberg>=0.6.0",
-    "torch~=2.7",
+    "torch==2.8.0",
```
aaron-siegel, line 135: `torch>=2.8,<2.9`

no need to freeze it so rigidly

PR #1052, opened 2026-01-09
Ignore test_view_indices failures on Windows (but still run the test)

sergey-mkhitaryan:


aaron-siegel:
How would this solve the problem? i.e., if the test causes the pytest harness to segfault, how does ignoring the test results help?

PR #1054, opened 2026-01-13
Add local docs serving instructions to contributing guide

apreshill:
## Updated CONTRIBUTING.md with Mintlify Documentation Setup

**Changes:**
* Added "Building and Serving Documentation Locally" section with setup instructions
* Included Node.js installation steps (via nvm for macOS/Linux, direct download for Windows)
* Added Mintlify CLI installation instructions
* Documented `make` commands for building and serving docs locally (`make docs`, `make docs-serve`)
* Added reminder in PR workflow to preview doc changes locally before submitting

**Why:**
Contributors working on documentation need clear instructions for previewing their changes locally using our Mintlify setup.


CONTRIBUTING.md
```diff
@@ -64,6 +64,53 @@ already have standard developer tools such as `git` and `make` installed on your
 
 We recommend VSCode for development: <https://code.visualstudio.com/>
 
+## Building and Serving Documentation Locally
+
+Pixeltable uses [Mintlify](https://mintlify.com) for documentation. If you're contributing to the docs, you may want to build and preview them locally before submitting a PR.
+
+### Prerequisites and Installation
+
+#### 1. Install Node.js
```
aaron-siegel: This step is not necessary. If someone has followed the preceding instructions then they will have a compatible version of node.js in their conda environment via `make install`.

CONTRIBUTING.md
```diff
@@ -64,6 +64,53 @@ already have standard developer tools such as `git` and `make` installed on your
...
+
+* **Windows**: Download and install from [nodejs.org](https://nodejs.org/)
+
+* Verify your installation:
+
+    ```bash
+    node --version  # Should show v19.0.0 or higher
+    ```
+
+#### 2. Install Mintlify CLI
```
aaron-siegel: Similarly, not necessary; it's installed into the conda environment via `make install`.

aaron-siegel (changes requested):

CONTRIBUTING.md
tiennguyentony: +1

CONTRIBUTING.md
tiennguyentony: Technically, I contribute into pixeltable by using different package instead of using ```make```:

```Pip install``` and ```ruff format``` is what I used. Is Pixeltable only supported make for Anaconda now since my Conda environment is broken due to some environment variables bugs that I am quite lazy to fix.

If developer needs to firmly use Miniconda (Anaconda) then this is needed to update that Pixeltable won't accept other ways of contributing without using ```make```

Comment (aaron-siegel) [reply CONTRIBUTING.md]: conda is not required in order to *use* pixeltable (for example, if you are developing a different Python library that requires pixeltable as a dependency: you can use pip, uv, conda, or any other package manager that works).

However, conda is the only supported *development* environment for pixeltable (if you are submitting PRs or making changes to Pixeltable itself).

If you find some way around using `make install` then you're certainly free to go that route, and we'll accept PRs as long as they pass our tests :) But I'd prefer that we avoid giving the impression in the docs that it's a supported workflow.

Comment (tiennguyentony) [reply CONTRIBUTING.md]: Sounds good to me! Thanks for your confirmation!

File: CONTRIBUTING.md [hunk 3]
```diff
@@ -64,6 +64,53 @@ already have standard developer tools such as `git` and `make` installed on your
 
 We recommend VSCode for development: <https://code.visualstudio.com/>
 
+## Building and Serving Documentation Locally
```
aaron-siegel, line 79: It also occurred to me, we should include instructions for updating docs.json and public_api.opml, without which SDK additions won't appear in the docs.

CONTRIBUTING.md
apreshill: @aaron-siegel can users who need to install miniconda, but don't wish to download from Anaconda, choose miniforge instead?

https://conda-forge.org/download/

tiennguyentony: I will assume that internally we have approved for Miniforge

tiennguyentony (approved):

pierrebrunelle (approved):

aaron-siegel (approved):

PR #1055, opened 2026-01-13
Use '/' as path delimiter

amithadke:


pixeltable/catalog/globals.py
```diff
@@ -80,7 +80,16 @@ def is_valid_identifier(name: str, *, allow_system_identifiers: bool = False, al
 def is_valid_path(path: str, empty_is_valid: bool, allow_system_paths: bool = False) -> bool:
```
aaron-siegel: Move this to `Path` and consolidate the splitting logic?

Also, I notice this doesn't handle versions correctly. I'm not sure it's even actually being used anywhere.

pixeltable/catalog/path.py
```diff
@@ -105,17 +115,20 @@ def __repr__(self) -> str:
...
     def __eq__(self, other: object) -> bool:
-        return isinstance(other, Path) and str(self) == str(other)
+        if not isinstance(other, Path):
+            return False
+        return self.components == other.components and self.version == other.version
 
     def __hash__(self) -> int:
-        return hash(str(self))
+        # Hash based on components and version
+        return hash((tuple(self.components), self.version))
```
aaron-siegel: `components` should probably just be a tuple; it's immutable.

aaron-siegel (approved):

PR #1060, opened 2026-01-14
PXT-898 Allow Pixeltable API key to change in the environment mid-stream in a Python session

amithadke:


pixeltable/share/publish.py
```diff
@@ -345,6 +345,13 @@ def list_table_versions(table_uri: str) -> list[dict[str, Any]]:
 def _api_headers() -> dict[str, str]:
     headers = {'Content-Type': 'application/json'}
     api_key = Env.get().pxt_api_key
-    if api_key is not None:
-        headers['X-api-key'] = api_key
+    if api_key is None:
+        from pixeltable.config import Config
```
aaron-siegel: move to top of file

pixeltable/share/publish.py
```diff
@@ -345,6 +345,13 @@ def list_table_versions(table_uri: str) -> list[dict[str, Any]]:
...
+
+        raise excs.Error(
+            'Pixeltable API key not found. Set it via: `import os; os.environ["PIXELTABLE_API_KEY"] = "your-key"` '
+            f'Alternatively, add `api_key = "your-key"` to the `[pixeltable]` section of '
+            f'{Config.get().config_file} and restart.'
```
aaron-siegel: Add link to config doc page?

pixeltable/env.py
```diff
@@ -476,7 +476,8 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
                 except Exception as e:
                     raise excs.Error(f'Invalid {mode} media destination URI: {uri}') from e
 
-        self._pxt_api_key = config.get_string_value('api_key')
+        # Don't load API key during initialization - load it lazily when needed
+        self._pxt_api_key = None
```
aaron-siegel: I'm not sure there's value in caching it here; it makes the code a lot more convoluted, and Config.get_string_value() is pretty fast and doesn't get called *that* often. See below.

pixeltable/env.py
```diff
@@ -715,6 +716,21 @@ def _upgrade_metadata(self) -> None:
 
     @property
     def pxt_api_key(self) -> str | None:
+        """Get the Pixeltable API key, loading it from config if not already loaded."""
```
aaron-siegel: If we dispense with the `_pxt_api_key` cache, then this entire method body can be replaced by a simple `Config.get_string_value('api_key')`, which will always consult the environment as its first action.

pixeltable/share/publish.py
amithadke: done

pixeltable/share/publish.py
```diff
@@ -345,6 +346,12 @@ def list_table_versions(table_uri: str) -> list[dict[str, Any]]:
 def _api_headers() -> dict[str, str]:
     headers = {'Content-Type': 'application/json'}
     api_key = Env.get().pxt_api_key
-    if api_key is not None:
-        headers['X-api-key'] = api_key
+    if api_key is None:
+        raise excs.Error(
+            'Pixeltable API key not found. '
```
aaron-siegel: Instead of "Pixeltable API key not found"... how about the friendlier "A Pixeltable API key is required to use this feature."

You may want to add a \n somewhere to break up the error message (maybe after `section in {Config.get().config_file}.`)

PR #1061, opened 2026-01-14
[PXT-875] Disable performance test until it reliably passes

sergey-mkhitaryan:


.github/workflows/stress-tests.yml
```diff
@@ -71,6 +71,7 @@ jobs:
           path: ~/.pixeltable/random-ops.log
 
   performance-tests:
+    if: false # Disable until it can reliably pass on a GitHub runner
```
aaron-siegel, line 74: Alternatively, you can use
`continue-on-error: true`
and the job will run, but ignored if it fails.

sergey-mkhitaryan: ack but I don't think there's a point in that

PR #1062, opened 2026-01-14
[PXT-875] Fix openai perftest on github

sergey-mkhitaryan:
The key to fix is that "connect" and "pool" timeouts need to be unset in httpx. We execute many long-running HTTP requests, so they can spend an unbounded amount of time enqued. As such, we cannot enforce any reasonable timeout on waiting in the queue.

### DIFF
--- a/pixeltable/functions/openai.py
+++ b/pixeltable/functions/openai.py
@@ -45,13 +45,14 @@ def _(api_key: str, base_url: str | None = None, api_version: str | None = None)
     set_file_descriptor_limit(max_connections * 2)
     default_query = None if api_version is None else {'api-version': api_version}
 
-    # TODO: once there's enough confidence in this HTTP client configuration, we should apply it to other clients.
+    # TODO(PXT-1013): once there's enough confidence in this HTTP client configuration, we should apply it to other
+    # clients.
     # Pixeltable scheduler's retry logic takes into account the rate limit-related response headers, so in theory we can
     # benefit from disabling retries in the OpenAI client (max_retries=0). However to do that, we need to get smarter
     # about idempotency keys and possibly more.
     http_limits = httpx.Limits(max_keepalive_connections=max_keepalive_connections, max_connections=max_connections)
     # Connect and pool timeouts should be unset because requests can spend an unbounded time waiting in the queue.
-    # Note: the clock start ticking for the connect timeout when the request enters the queue, not when it is put on
+    # Note: the clock starts ticking for the connect timeout when the request enters the queue, not when it is put on
     # the wire.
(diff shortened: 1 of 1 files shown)


pixeltable/functions/openai.py
```diff
@@ -37,24 +37,33 @@
...
+    http_timeouts = httpx.Timeout(connect=None, read=read_timeout, write=write_timeout, pool=None)
+    _logger.debug(f'Initializing AsyncOpenAI client with httpx limits: {http_limits} and timeouts: {http_timeouts}')
     return openai.AsyncOpenAI(
         api_key=api_key,
         base_url=base_url,
         default_query=default_query,
-        # recommended to increase limits for async client to avoid connection errors
         http_client=httpx.AsyncClient(
-            limits=httpx.Limits(max_keepalive_connections=max_keepalive_connections, max_connections=max_connections),
+            limits=http_limits,
```
mkornacker, line 64: httpx.AsyncClient() is used in a number of our integrations. Let's fix this in a comprehensive way, so we don't need to bake it into each one of them separately.

sergey-mkhitaryan: see `# TODO: once there's enough confidence in this HTTP client configuration, we should apply it to other clients.` In other words, let this perf test run for awhile, and also investigate the row batching, and then apply this to all HTTP clients.

pixeltable/functions/openai.py
```diff
@@ -38,24 +38,33 @@
...
     set_file_descriptor_limit(max_connections * 2)
     default_query = None if api_version is None else {'api-version': api_version}
 
+    # TODO: once there's enough confidence in this HTTP client configuration, we should apply it to other clients.
     # Pixeltable scheduler's retry logic takes into account the rate limit-related response headers, so in theory we can
     # benefit from disabling retries in the OpenAI client (max_retries=0). However to do that, we need to get smarter
     # about idempotency keys and possibly more.
+    http_limits = httpx.Limits(max_keepalive_connections=max_keepalive_connections, max_connections=max_connections)
+    # Connect and pool timeouts should be unset because requests can spend an unbounded time waiting in the queue.
+    # Note: the clock start ticking for the connect timeout when the request enters the queue, not when it is put on
```
mkornacker: "starts"

pixeltable/functions/openai.py
mkornacker: Let's go with that. Do we have a follow-on jira?

sergey-mkhitaryan: yes, just filed one https://pixeltable.atlassian.net/browse/PXT-1013

PR #1065, opened 2026-01-15
PXT-139 Add default value support for columns

amithadke:
Support literal default values in table columns and views.


tests/test_snapshot.py
```diff
@@ -413,3 +413,92 @@ def test_rename_column(self, reset_db: None) -> None:
 
         # should work
         v1.rename_column('v1', 'new_v1')
+
+    def test_additional_columns_with_defaults(self, reset_db: None, reload_tester: ReloadTester) -> None:
+        """Test that snapshots with additional_columns that have default values work correctly."""
+        # Create base table with columns that have default values
+        t = pxt.create_table('base_tbl', {'c1': pxt.Int})
+        # Add columns with defaults to base table (table is empty, so this is allowed)
+        t.add_column(base_int={'type': pxt.Int, 'default': 10})
```
mkornacker: add_column(base_int=pxt.Int, default=10) would be easier to intuit.

tests/test_snapshot.py
```diff
@@ -413,3 +413,92 @@ def test_rename_column(self, reset_db: None) -> None:
...
+            assert row['base_int'] == 10, f'Expected 10, got {row["base_int"]}'
+            assert row['base_str'] == 'base_default', f"Expected 'base_default', got {row['base_str']}"
+            assert row['base_json'] == {'base': 'data'}, f"Expected {{'base': 'data'}}, got {row['base_json']}"
+
+        # Create snapshot with additional_columns that have default values
+        s1 = pxt.create_snapshot(
+            'snap1',
+            t,
+            additional_columns={
+                's1_int': {'type': pxt.Int, 'default': 100},
```
mkornacker: We should probably not allow this. It only takes up storage space and doesn't appear to have any value.

tests/test_table.py
```diff
@@ -3005,3 +3005,98 @@ def test_drop_last_column(self, reset_db: None, reload_tester: ReloadTester) ->
             pxt.Error, match="Cannot drop column 'c2' because it is the last remaining column in this table"
         ):
             t.drop_column('c2')
+
+    def test_column_defaults(self, reset_db: None, reload_tester: ReloadTester) -> None:
+        """Test adding columns with default values."""
```
mkornacker: This should also work in the initial schema spec. (Why just 'adding'?)

mkornacker (changes requested):

tests/test_snapshot.py
amithadke: I've not removed it yet , can remove it quickly.
Are you saying it doesn't provide value because applications will always be setting value of additional columns in their business logic so less likely to use default value.

Alternatively I was thinking that adding/defining column should look same across all places in pixeltable. (specially we move to class/schema based specs)

pixeltable/catalog/catalog.py
```diff
@@ -1290,6 +1290,13 @@ def add_fn() -> None:
         # force a reload in order to see the new columns/idxs
         self._clear_tv_cache(TableVersionKey(tbl.tbl_id, None, None))
         self._roll_forward()
+
+        # Populate default values for columns that were just added to existing rows
```
amithadke: im thinking of adding this to add_column_ops as last /optional operation 
current approach populates existing rows in one transaction for all columns (also avoids column creation and setting value for that column in the same transaction)

tests/test_snapshot.py
mkornacker: It's a snapshot, so you can't update it. It's simply a constant int stored in every row, unclear why anyone would want that.

pixeltable/catalog/catalog.py
mkornacker: Yeah, this needs to be a separate operation to guarantee atomicity.

tests/test_snapshot.py
amithadke: nvm i was thinking for views, snapshot will take it out

pixeltable/catalog/catalog.py
amithadke: todo : add_column_ops will add another op SetColumnValueOp that will update existing rows

pixeltable/metadata/schema.py
```diff
@@ -121,12 +121,15 @@ class ColumnMd:
...
 
-    # if set, this is a computed column
+    # if set, this is a computed column or a column with a default value
     value_expr: dict | None
 
     # if True, the column is present in the stored table
     stored: bool | None
 
+    # if True, value_expr represents a computed column; if False, value_expr represents a default value
+    is_computed_column: bool = False
```
mkornacker: It feels like having an explicit default_expr would be easier to follow.

Also, with a 'default_value_expr: dict | None = None' you don't need a migration.

pixeltable/exprs/literal.py
```diff
@@ -113,6 +119,19 @@ def _as_dict(self) -> dict:
     def as_literal(self) -> Literal | None:
         return self
 
+    @classmethod
+    def _to_stored_value(cls, val: Any, col_type: ts.ColumnType) -> Any:
```
mkornacker: This is kind of the analogue to ColumnType.create_literal() and should live there.

pixeltable/exprs/row_builder.py
```diff
@@ -497,12 +497,30 @@ def create_store_table_row(
         """
         from pixeltable.exprs.column_property_ref import ColumnPropertyRef
 
+        def _get_default_value(col: catalog.Column) -> Any:
+            """Get default value for a column, initializing value_expr if needed."""
+            col.init_value_expr(None)  # idempotent: no-op if already initialized
```
mkornacker: This really shouldn't be happening here. This is new schema data, and all of that is managed by Catalog/TableVersion. Doing partial initializations and spreading initialization around makes the state very difficult to reason about.

pixeltable/exprs/row_builder.py
```diff
@@ -497,12 +497,30 @@ def create_store_table_row(
...
+
         num_excs = 0
         table_row: list[Any] = list(pk)
         # Nulls in JSONB columns need to be stored as sql.sql.null(), otherwise it stores a json 'null'
         for col, slot_idx in self.table_columns.items():
             if col.id in data_row.cell_vals:
-                table_row.append(data_row.cell_vals[col.id])
+                val = data_row.cell_vals[col.id]
+                # Use default if value is None (or sql.sql.null() for JSON types)
+                if _is_null(val, col) and col.has_default_value:
```
mkornacker: If someone provided an explicit None, we need to keep that.

That means that the decision to use the default needs to be made upstream (the planner would know).

pixeltable/exec/in_memory_data_node.py
```diff
@@ -65,7 +65,15 @@ def _open(self) -> None:
                 col_info = user_cols_by_name.get(col_name)
                 assert col_info is not None
                 col = col_info.col
-                if col.col_type.is_image_type() and isinstance(val, bytes):
+                # If value is None and column has a default, use the default value
```
mkornacker: Wrong behavior (see rowbuilder comment).

pixeltable/exprs/literal.py
```diff
@@ -80,7 +84,9 @@ def sql_expr(self, _: SqlElementCache) -> sql.ColumnElement | None:
 
     def eval(self, data_row: DataRow, row_builder: RowBuilder) -> None:
         # this will be called, even though sql_expr() does not return None
-        data_row[self.slot_idx] = self.val
+        # Always use stored_value which is computed once during initialization
+        # This is especially important for arrays which need to be converted to stored format
+        data_row[self.slot_idx] = self.stored_value
```
mkornacker: This is wrong. Datarow.vals does not contain stored values, only the in-memory values.

mkornacker (changes requested):

pixeltable/exec/in_memory_data_node.py
```diff
@@ -74,12 +74,16 @@ def _open(self) -> None:
...
                 input_slot_idxs.add(col_info.slot_idx)
 
-            # set the remaining output slots to their default values (presently None)
+            # set the remaining output slots to their default values
             missing_slot_idxs = output_slot_idxs - input_slot_idxs
             for slot_idx in missing_slot_idxs:
                 col_info = output_cols_by_idx.get(slot_idx)
                 assert col_info is not None
-                output_row[col_info.slot_idx] = None
+                if col_info.col.has_default_value:
```
amithadke, line 82: @mkornacker I ended up keeping this change here as it does exactly what we want to do - check incoming rows and fill in missing columns if they have defaults.  If user specified none we will not overwrite it.

I didn't find it useful to replicate this back in SetDefaultValueNode. Let me know if it needs to be 
SetDefaultValueNode is used in other two cases.

Also I didn't like adding default value conversion inside CellMateriazationNode (we will expose it to non array/json column types)  - unless you think it has performance benefits

pixeltable/plan.py
```diff
@@ -405,21 +405,77 @@ def create_query_insert_plan(
...
+                input=plan,
+                maintain_input_order=False,
+            )
+
+        # register query columns in table_columns so create_store_table_row knows which slot to read for each dst column
         needs_cell_materialization = False
         for col_name, expr in zip(query.schema.keys(), query._select_list_exprs):
-            assert col_name in tbl.cols_by_name
             col = tbl.cols_by_name[col_name]
             plan.row_builder.add_table_column(col, expr.slot_idx)
```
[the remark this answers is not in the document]
amithadke: added back assert

pixeltable/catalog/column.py
```diff
@@ -219,11 +233,51 @@ def create(cls, name: str, spec: type | ColumnSpec | exprs.Expr) -> Column:
...
+                        f'(String, Int, Float, Bool, Timestamp, Date, UUID), simple JSON, '
+                        f'Array, and Binary.'
+                    )
+                # Ensure the default value's type is compatible with the column type
+                default_type = default_value_expr.col_type
+                if not col_type.is_supertype_of(default_type, ignore_nullable=True):
+                    raise excs.Error(
+                        f'Column {name!r}: Default value type {default_type} is not compatible '
+                        f'with column type {col_type}.'
+                    )
```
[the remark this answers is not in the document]
amithadke: added check during  spec validation to make sure default value is not none

pixeltable/catalog/tbl_ops.py
```diff
@@ -226,6 +226,25 @@ def undo(self, tv: TableVersion | None) -> None:
...
+    column_ids: list[int]
+
+    def exec(self, tv: TableVersion | None) -> None:
+        assert not get_runtime().in_xact
+        cols = [tv.cols_by_id[col_id] for col_id in self.column_ids]
+        with get_runtime().begin_xact():
+            tv._populate_default_values(cols)
+
+    def undo(self, tv: TableVersion | None) -> None:
+        pass
```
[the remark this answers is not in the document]
amithadke: SetColumnValueOp runs as part of add_column — if it fails or needs to roll back, the entire column gets dropped

pixeltable/catalog/table_version.py
```diff
@@ -913,6 +937,48 @@ def _add_columns(
...
+        # Set all default value columns
+        for col in default_value_cols:
+            assert col.has_default_value, f'Column {col.name!r} should have default value'
+            assert col.default_value_expr is not None, f'Column {col.name!r} default_value_expr should be set'
+            # sa_col should be initialized by create_sa_tbl() called above
+            assert col.sa_col is not None, f'Column {col.name!r} sa_col should be initialized'
+            default_val = col.col_type.to_stored_value(col.default_value_expr.val)
+            update_stmt = update_stmt.values({col.sa_col: default_val})
+
+        conn.execute(update_stmt)
```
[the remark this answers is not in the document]
amithadke: i dont think user can update these rows/columns when there is a pending op to set existing rows column value.

pixeltable/catalog/column.py
```diff
@@ -219,11 +233,51 @@ def create(cls, name: str, spec: type | ColumnSpec | exprs.Expr) -> Column:
...
+                if value_expr is not None and isinstance(value_expr, exprs.Expr):
+                    # create copy so we can modify it
+                    value_expr = value_expr.copy()
+                    value_expr.bind_rel_paths()
+            elif 'default' in spec:
+                # Column with default value
+                default_expr_obj = spec['default']
+                default_value_expr = exprs.Expr.from_object(default_expr_obj)
+                default_value_expr = default_value_expr.copy()
+                default_value_expr.bind_rel_paths()
```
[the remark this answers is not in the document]
amithadke: done using literal to construct default value expr

pixeltable/catalog/table.py
```diff
@@ -623,6 +626,7 @@ def add_columns(
     def add_column(
         self,
         *,
+        default: Any = None,
         if_exists: Literal['error', 'ignore', 'replace', 'replace_force'] = 'error',
         **kwargs: type | ColumnSpec,
```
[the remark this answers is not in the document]
amithadke: default=None meaning "not provided"  and creates a regular column is fine. Adding sentinels complicates arg serializations and typescript version. we will still keep validation for default value = Nne  in case its called from other places in code

pixeltable/exprs/column_ref.py
```diff
@@ -98,8 +101,11 @@ def __init__(
                 )
         else:
             assert perform_validation is None or not perform_validation
+
+        self.for_insert = for_insert if for_insert is not None else False
+
         if self.perform_validation:
-            non_validating_col_ref = ColumnRef(col, perform_validation=False)
+            non_validating_col_ref = ColumnRef(col, perform_validation=False, for_insert=for_insert)
             self.components = [non_validating_col_ref]
```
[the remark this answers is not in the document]
amithadke: solved this by having row builder build stored columns first then computed columns which will guarantee ordering

pixeltable/exprs/column_ref.py
```diff
@@ -441,8 +447,12 @@ def eval(self, data_row: DataRow, row_builder: RowBuilder) -> None:
...
 
         if not self.is_unstored_iter_col:
-            # supply default
-            data_row[self.slot_idx] = None
+            # Only fill default when the slot has no value and this ref is used in insert/view propagation
+            if not data_row.has_val[self.slot_idx]:
+                if self.for_insert and self.col.has_default_value and self.col.default_value_expr is not None:
+                    data_row[self.slot_idx] = self.col.default_value_expr.val
+                else:
+                    data_row[self.slot_idx] = None
```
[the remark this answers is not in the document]
amithadke: default_value_expr is a literal expression which is immutable

pixeltable/exec/in_memory_data_node.py
```diff
@@ -74,12 +74,16 @@ def _open(self) -> None:
...
+                    assert col_info.col.default_value_expr is not None
+                    output_row[col_info.slot_idx] = col_info.col.default_value_expr.val
+                else:
+                    output_row[col_info.slot_idx] = None
```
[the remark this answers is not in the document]
amithadke: same as above we dont have mutable defaults we are using literal expression

pixeltable/plan.py
```diff
@@ -405,21 +405,85 @@ def create_query_insert_plan(
...
+            analyzer=analyzer,
+            eval_ctx=query_eval_ctx,
+            with_pk=True,
+            limit=query.limit_val,
+            offset=query.offset_val,
+        )
+
+        # extra cols are in default_eval_ctx but not handled by the query plan
+        query_target_set = exprs.ExprSet(query_eval_ctx.target_exprs)
+        extra_col_exprs = [e for e in row_builder.default_eval_ctx.target_exprs if e not in query_target_set]
```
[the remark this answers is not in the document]
amithadke: extra_col_exprs = [row_builder.unique_exprs[row_builder.table_columns[col]] for col in extra_cols]

extra_cols shouldn't have duplicates and unique_exprs by definition contains distinct exprs

pixeltable/exprs/row_builder.py
```diff
@@ -149,23 +150,19 @@ def __init__(
...
                 # record a ColumnRef so that references to this column resolve to the same slot idx
                 perform_validation = (
                     None
                     if not col.col_type.is_media_type()
                     else col.media_validation == catalog.MediaValidation.ON_WRITE
                 )
-                expr = ColumnRef(col, perform_validation=perform_validation)
+                expr = ColumnRef(col, perform_validation=perform_validation, for_insert=for_insert)
                 # recursive=True: needed for validating ColumnRef
                 expr = self._record_unique_expr(expr, recursive=True)
```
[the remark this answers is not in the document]
amithadke: RowBuilder is created  output_exprs=analyzer.all_exprs and extra_cols which are not part of query but are part of the table being inserted. 
rowbuilder will initialze all non computed columns with for_insert=True and then build expressions leaving for_insert out of id_attr allows computed column value expressions to reuse slots for columns we created in the first pass with for_insert=true

we should never have an ordering issue where column_ref with for_insert=false gets in before other

pixeltable/plan.py
```diff
@@ -405,21 +405,84 @@ def create_query_insert_plan(
...
+        if extra_col_exprs:
+            plan = exec.ExprEvalNode(
+                row_builder,
+                output_exprs=extra_col_exprs,
+                input_exprs=row_builder.input_exprs,
+                input=plan,
+                maintain_input_order=False,
+            )
+
+        # register query columns in table_columns so create_store_table_row knows which slot to read for each dst column
```
amithadke, line 470: old comment not needed

PR #1067, opened 2026-01-16
rename reset_db fixture to uses_db

sergey-mkhitaryan:


aaron-siegel:
I prefer `uses_db` or `touches_db`.

sergey-mkhitaryan:
@aaron-siegel ptal

mkornacker:
What's wrong with reset_db? I think that accurately describes what it does.

sergey-mkhitaryan:
> What's wrong with reset_db? I think that accurately describes what it does.

It's not just a reset now that we've added a db validation there.

PR #1068, opened 2026-01-16
PXT-961: correct support for alpha in draw_bounding_boxes()

mkornacker:


pixeltable/functions/vision.py
```diff
@@ -319,7 +325,11 @@ def draw_bounding_boxes(
         labels: List of labels for each bounding box.
         color: Single color to be used for all bounding boxes and labels.
         box_colors: List of colors, one per bounding box.
+        alpha: Opacity (0-1) of the bounding box borders and labels. If non-`None`, overrides any alpha in
```
aaron-siegel, line 328: Is that the right behavior (overrides any alpha in ...)?

To me the most logical resolution order is:
- Use the explicit opacity in the individual color
- If the color does not specify an explicit opacity, use alpha / fill_alpha
- If neither are specified, use the defaults

pixeltable/functions/vision.py
```diff
@@ -350,27 +360,43 @@ def draw_bounding_boxes(
         label_colors = __create_label_colors(labels)
         box_colors = [label_colors[label] for label in labels]
 
-    from PIL import ImageColor, ImageDraw, ImageFont
+    rgb_box_colors = [PIL.ImageColor.getrgb(c) for c in box_colors]
+    rgba_border_colors: list[tuple[int, int, int, int]]
+    if alpha is not None:
+        # override any alpha in rgb_box_colors
```
aaron-siegel, line 366: See comment above

pixeltable/functions/vision.py
mkornacker: The reasoning is:
- you might want different alpha for the border and the fill
- if the caller explicitly sets alpha/fill_alpha, that's a clear signal of intent

PR #1072, opened 2026-01-16
misc improvements in the code

sergey-mkhitaryan:


pixeltable/store.py
```diff
@@ -315,7 +315,7 @@ def drop_column(self, col: catalog.Column) -> None:
         log_stmt(_logger, stmt)
         Env.get().conn.execute(stmt)
 
-    def load_column(self, col: catalog.Column, exec_plan: ExecNode, abort_on_exc: bool) -> int:
+    def populate_new_column(self, col: catalog.Column, exec_plan: ExecNode, abort_on_exc: bool) -> int:
```
mkornacker: Let's stick with load_column(). The logic doesn't depend on this being a new column.

mkornacker (changes requested):

pixeltable/store.py
sergey-mkhitaryan: then `populate_column()` maybe? "load" makes no sense to me in this context -- load usually means there's data somewhere in the external store, and we copy it closer to us (local disk, RAM, etc.)

aaron-siegel: Between those two I prefer `populate_column()` (`load` has many meanings and I generally prefer unambiguous verbs to ambiguous ones)

mkornacker: In that case, we already use the verb 'write' for similar activity in catalog.py, so I'd suggest write_column(), which is also unambiguous.

PR #1075, opened 2026-01-18
Transaction rollback

mkornacker:
Changes transaction behavior to roll back specific table operations when they encounter an error.

Included:
- a fix for PXT-842 (error during create_view() corrupts db)
- a multi-phase implementation of Table.add_columns()

Left for follow-up work: extending the multi-phase logic to add_computed_columns() (there's currently a lot of duplicated logic between TableVersion.add_columns_ops() and TableVersion.add_columns())

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -88,6 +88,7 @@ env:
   TOGETHER_API_KEY: ${{ secrets.TOGETHER_API_KEY }}
   TWELVELABS_API_KEY: ${{ secrets.TWELVELABS_API_KEY }}
   VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
+  RUNWAYML_API_SECRET: ${{ secrets.RUNWAYML_API_SECRET }}
 
 jobs:
   generate-matrix:
--- a/docs/public_api.opml
+++ b/docs/public_api.opml
@@ -320,6 +320,12 @@
           <outline text="udf|pixeltable.functions.reve.edit" />
           <outline text="udf|pixeltable.functions.reve.remix" />
         </outline>
+        <outline text="module|pixeltable.functions.runwayml">
+          <outline text="udf|pixeltable.functions.runwayml.text_to_image" />
+          <outline text="udf|pixeltable.functions.runwayml.text_to_video" />
+          <outline text="udf|pixeltable.functions.runwayml.image_to_video" />
+          <outline text="udf|pixeltable.functions.runwayml.video_to_video" />
+        </outline>
         <outline text="module|pixeltable.functions.together">
           <outline text="udf|pixeltable.functions.together.completions" />
           <outline text="udf|pixeltable.functions.together.chat_completions" />
--- a/docs/release/docs.json
+++ b/docs/release/docs.json
@@ -240,6 +240,7 @@
                       "sdk/latest/openrouter",
                       "sdk/latest/replicate",
                       "sdk/latest/reve",
+                      "sdk/latest/runwayml",
                       "sdk/latest/together",
                       "sdk/latest/twelvelabs",
                       "sdk/latest/vision",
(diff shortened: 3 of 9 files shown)


pixeltable/catalog/catalog.py
```diff
@@ -635,44 +633,89 @@ def _acquire_tbl_lock(
...
         """Finalize pending ops for all tables in self._roll_forward_ids."""
         for tbl_id in self._roll_forward_ids:
-            self._finalize_pending_ops(tbl_id)
-            # TODO: handle replicas
-            self._clear_tv_cache(TableVersionKey(tbl_id, None, None))
+            try:
+                # TODO: handle replicas
+                exc = self._finalize_pending_ops(tbl_id)
+                if exc is not None:
+                    raise excs.Error(f'Table operation was aborted with\n{exc!s}')
```
aaron-siegel: raise from exc (otherwise we lose the stack trace history, which is extremely annoying when debugging)

pixeltable/catalog/catalog.py
```diff
@@ -635,44 +633,89 @@ def _acquire_tbl_lock(
...
+                    q = sql.select(schema.Table.md).where(schema.Table.id == tbl_id).with_for_update()
+                    row = conn.execute(q).one_or_none()
+                    if row is None:
+                        return None
+                    tbl_md = schema.md_from_dict(schema.TableMd, row.md)
+                    if tbl_md.tbl_state == schema.TableState.LIVE:
+                        # nothing left to do
+                        return None
+                    is_rollback = tbl_md.tbl_state == schema.TableState.ROLLBACK
+                    is_snapshot = False if tbl_md.view_md is None else tbl_md.view_md.is_snapshot
```
aaron-siegel: Should TableMd have an is_snapshot property? It seems awkward to be inlining this logic

pixeltable/catalog/catalog.py
```diff
@@ -635,44 +633,89 @@ def _acquire_tbl_lock(
...
+                    assert len(rows) > 0
+                    ops = [TableOp.from_dict(dict(row.op)) for row in rows]
+
+                    # determine next op to execute/undo
+                    if is_rollback:
+                        # last aborted: in chronological order (ie, the one with the lowest op_sn)
+                        last_aborted_op = next((op for op in ops if op.status == OpStatus.ABORTED), None)
+                        if last_aborted_op is None:
+                            # we haven't aborted anything yet and need to start with the first pending op
+                            op = next(op for op in ops if op.status == OpStatus.PENDING)
```
aaron-siegel, line 706: I'm confused about why we're executing pending ops in rollback mode. More clarification would be useful.

More generally: I think a more detailed design doc that explains the data structures and overarching db ops logic would be very useful.

aaron-siegel: EDIT: After reading the logic below, I think I understand: we're cancelling the pending ops, not executing them?

pixeltable/catalog/table_version.py
```diff
@@ -420,17 +428,73 @@ def exec_op(self, op: TableOp) -> None:
...
+            Catalog.get().write_tbl_md(self.id, None, self._tbl_md, None, None, [])
+
+        elif isinstance(op, CreateStoreColumnsOp):
+            for col_id in op.column_ids:
+                with Env.get().begin_xact():
+                    self.store_tbl.drop_column(self.cols_by_id[col_id])
+
+        elif isinstance(op, (DeleteTableMdOp, DeleteTableMediaFilesOp, DropStoreTableOp)):
+            # physical deletion cannot be rolled back
+            raise AssertionError()
```
aaron-siegel, line 496: Add a note saying why we're certain these can never be part of a rollback (I assume because they always come at the end of an ops sequence?)

We could make them undoable by moving the files to a trash bin and only emptying the trash at the end of a (Pixeltable) transaction. In fact we'll need this if we want to support user transactions.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: It already does, thank you for the reminder.

pixeltable/catalog/catalog.py
mkornacker: Explained the logic in more detail.

pixeltable/catalog/table_version.py
mkornacker: Done.

PR #1078, opened 2026-01-22
to_string() method on UUIDType

aaron-siegel:


pixeltable/functions/uuid.py
```diff
@@ -106,6 +108,31 @@ def uuid7() -> uuid.UUID:
     return _uuid7()
 
 
+@pxt.udf
+def to_string(u: uuid.UUID, format: Literal['standard', 'hex'] = 'standard') -> str:
```
mkornacker: Instead of the format parameter, why not simply also have a hex() udf? That mirrors the way the Python module is organized, and I anticipate we'll also need int() and urn() udfs (which you could add now or leave to later).

aaron-siegel: This is fine with me and makes sense.

aaron-siegel: Done. I also added to_sql decorations for both UDFs.

pixeltable/functions/uuid.py
```diff
@@ -106,6 +106,55 @@ def uuid7() -> uuid.UUID:
...
+
+@to_string.to_sql
+def _(u: sql.ColumnElement) -> sql.ColumnElement:
+    return u.cast(sql.Text)
+
+
+@pxt.udf
+def hex(u: uuid.UUID) -> str:
+    """
+    Convert a UUID to its hexadecimal representation.
```
mkornacker, line 136: Link back to Python docs?

pixeltable/functions/uuid.py
```diff
@@ -106,6 +106,55 @@ def uuid7() -> uuid.UUID:
     return _uuid7()
 
 
+@pxt.udf
+def to_string(u: uuid.UUID) -> str:
```
mkornacker, line 110: Or should we call this simply str() and make it a method, and then you can write str(t.uuid_col), exactly mirroring the Python equivalent?

pixeltable/functions/uuid.py
aaron-siegel: Done.

pixeltable/functions/uuid.py
aaron-siegel: I don't love the idea of shadowing Python primitives... we can discuss at standup?

PR #1084, opened 2026-01-23
various improvements to random-ops script

sergey-mkhitaryan:


scripts/stress-tests.sh
```diff
@@ -16,6 +16,13 @@ fi
 WORKERS="$1"
 DURATION="$2"
 
+export PIXELTABLE_DB="random_ops"
+echo "Cleaning $PIXELTABLE_DB postgres DB ..."
+PIXELTABLE_HOME=~/.pixeltable
```
aaron-siegel: Why do we need to override this?

scripts/stress-tests.sh
```diff
@@ -16,6 +16,13 @@ fi
...
+POSTGRES_BIN_PATH=$(python -c 'import pixeltable_pgserver; import sys; sys.stdout.write(str(pixeltable_pgserver._commands.POSTGRES_BIN_PATH))')
+PIXELTABLE_URL="postgresql://postgres:@/postgres?host=$PIXELTABLE_HOME/pgdata"
```
aaron-siegel: Why do we need to override this? It seems dangerous? And the way it's constructed will (I believe) override psycopg3 and fall back on psycopg2 so we won't get an accurate test environment.

I think it's best to just leave this up to Pixeltable.

aaron-siegel: [Per discussion earlier today - ignore this comment - this is a harmless local variable; I had it confused with PIXELTABLE_DB_CONNECT_STR.]

scripts/stress-tests.sh
sergey-mkhitaryan: PTAL -- does this work?

aaron-siegel (approved):

PR #1087, opened 2026-01-25
PXT-973: additional_columns doesn't evaluate as expected when creating a view

mkornacker:
The bug stems from the fact that the computed column of the component view is defined with a non-deterministic function, and the planner assigns that function call to the input of the ComponentIterationNode (because it doesn't have parameters and therefore looks like it can be evaluated in the context of the base table).

To fix this, we need to recognize non-deterministic functions so that the planner can place them downstream of the ComponentIterationNode.

### DIFF
--- a/pixeltable/functions/huggingface.py
+++ b/pixeltable/functions/huggingface.py
@@ -499,7 +499,9 @@ def speech2text_for_conditional_generation(audio: pxt.Audio, *, model_id: str, l
         generated_ids = model.generate(**inputs.to(device), forced_bos_token_id=forced_bos_token_id).to('cpu')
 
     transcription = processor.batch_decode(generated_ids, skip_special_tokens=True)
-    return transcription
+    assert len(transcription) == 1
+    assert isinstance(transcription[0], str)
+    return transcription[0]
 
 
 @pxt.udf
--- a/tests/functions/test_huggingface.py
+++ b/tests/functions/test_huggingface.py
@@ -213,11 +213,6 @@ class TestHuggingface:
         assert result['label_text'] == ['meat loaf, meatloaf', 'mashed potato', 'broccoli']
 
     @pytest.mark.skipif(sys.version_info >= (3, 13), reason='Not working on Python 3.13+')
-    # Bug(PXT-944): speech2text_for_conditional_generation declares a return type of str, but actually returns a list.
-    # That list is then inserted in a string column of the table. Btree index inserts the same value to the index value
-    # column without truncating it to 256 characters because in s[: BtreeIndex.MAX_STRING_LEN], s is a list with
-    # 1 element and not a string.
-    @pytest.mark.corrupts_db
     def test_speech2text_for_conditional_generation(self, uses_db: None) -> None:
         skip_test_if_not_installed('transformers')
         from pixeltable.functions.huggingface import speech2text_for_conditional_generation


aaron-siegel:
I'm not sure I agree with this approach. It means that when users create their own UDFs, they need to know to mark them as "nondeterministic", or they'll get surprising behavior.

Why not simply treat all functions as "nondeterministic"? Are there specific user stories where we think performance will be an issue if we do? In my view, putting `additional_columns` here should guarantee *identical* behavior to an `add_computed_column()`.

mkornacker:
Marking functions/udfs in that way is standard practice for all dbms. It's needed to let optimizers move predicates around the execution tree, which is one of the most fundamental optimization techniques.

Very few functions are non-deterministic, and my impression is that making that the default is an extra burden for users.

mkornacker:
I'm not sure why you think the fix only applies to additional_columns and not those from add_computed_columns().

I expanded the test to include the latter scenario.

aaron-siegel:
> I'm not sure why you think the fix only applies to additional_columns and not those from add_computed_columns().
> 
> I expanded the test to include the latter scenario.

Because for `add_computed_column()` there's nothing to fix (the cells will be distinct whether or not it's marked with a nondeterminism flag), so this change is a no-op.

aaron-siegel:
> Marking functions/udfs in that way is standard practice for all dbms. It's needed to let optimizers move predicates around the execution tree, which is one of the most fundamental optimization techniques.
> 
> Very few functions are non-deterministic, and my impression is that making that the default is an extra burden for users.

Ok, if there are future applications of this flag beyond `additional_columns`, then it makes a lot more sense to introduce it. But I feel pretty strongly that the default should be nondeterministic if unspecified. Otherwise we are likely to surprise users with unexpected behavior. I'd argue that "presumption of determinism" imposes a substantially larger burden on users than the converse; we're requiring them to understand that nuance in order to guarantee correct behavior, versus it being an optimization that they can apply when warranted.

"Very few functions are non-deterministic" - I'm not so sure how you're defining "very few" - all generative model invocations are non-deterministic, for example. There are at least a couple dozen nondeterministic functions already in our library.

pixeltable/functions/uuid.py
```diff
@@ -79,7 +79,7 @@ def _uuid7() -> uuid.UUID:
     return uuid.UUID(int=int_uuid_7)
 
 
-@pxt.udf
+@pxt.udf(is_deterministic=False)
```
aaron-siegel, line 82: Just a thought: `deterministic=False` might be more consistent with our other APIs than `is_deterministic=False`.

For example, it's `add_computed_column(..., stored=False)`, not `add_computed_column(..., is_stored=False)`

tests/test_video.py
```diff
@@ -37,7 +37,6 @@ def create_tbls(
 
     def create_and_insert(self, stored: bool | None, paths: list[str]) -> tuple[pxt.Table, pxt.Table]:
         base_t, view_t = self.create_tbls()
-        _ = ObjectOps.count(view_t._id, default_output_dest=True)
```
aaron-siegel, line 40: Why take this out?

aaron-siegel: I guess it doesn't seem to be doing anything? :/

aaron-siegel (approved):

tests/test_video.py
mkornacker: Nope. :)

PR #1090, opened 2026-01-26
[PXT-969] Support for versioning of column metadata

sergey-mkhitaryan:
Foundation for more rich alter column operations.

1. Various column metadata (`col_type`, `is_pk`, `value_expr`) moved the non-versioned `ColumnMd` to versioned `SchemaColumn` structures.
2. In addition to user-visible (named) columns, we keep `SchemaVersionMd` for internal columns (e.g. index value and undo columns).
3. A catalog migration is added that implements 1 and 2.
4. `TableVersion.cols` deleted. `TableVersion.cols_by_id` is the replacement, with a notable difference that `cols_by_id` only has visible columns.
5. The code that determines and describes the table version history and changes is updated.
6. Fixes an off-by-one error when assigning positions to added columns in `TableVersion`.


pixeltable/metadata/schema.py
```diff
@@ -326,13 +322,32 @@ class SchemaColumn:
...
 
+    # If present, the URI for the destination for column values
+    destination: str | None = None
+
+    def __post_init__(self) -> None:
+        assert (self.pos is None) == (self.name is None), (
+            f'Invalid SchemaColumn: pos and name must both be set or both unset: {self.pos}, {self.name}'
+        )
+
+    def is_system_column(self) -> bool:
```
mkornacker, line 361: You could make that a property.

pixeltable/metadata/schema.py
```diff
@@ -116,19 +116,15 @@ class ColumnMd:
...
-
-    # if set, this is a computed column
-    value_expr: dict | None
 
     # if True, the column is present in the stored table
     stored: bool | None
 
-    # If present, the URI for the destination for column values
-    destination: str | None = None
+    def is_live_in_version(self, schema_version: int) -> bool:
```
mkornacker: exists_at_version() flows a bit better.

pixeltable/metadata/schema.py
mkornacker: Also update this.

pixeltable/catalog/catalog.py
```diff
@@ -2261,6 +2261,16 @@ def write_tbl_md(
                 assert tbl_md.current_schema_version == version_md.schema_version
             if schema_version_md is not None:
                 assert tbl_md.current_schema_version == schema_version_md.schema_version
+                # Validate that the columns in schema_version_md are consistent with tbl_md.
+                sch_col_ids = set(schema_version_md.columns.keys())
+                for tbl_col_id, tbl_col_md in tbl_md.column_md.items():
+                    if tbl_col_md.is_live_in_version(tbl_md.current_schema_version):
+                        assert tbl_col_id in sch_col_ids, (tbl_md.tbl_id, tbl_col_id)
+                        sch_col_ids.remove(tbl_col_id)
+                assert not sch_col_ids, (tbl_md.tbl_id, sch_col_ids)
```
mkornacker: Use an explicit len(...) == 0

pixeltable/catalog/table_version.py
```diff
@@ -149,6 +149,7 @@ class TableVersion:
     # record metadata stored in catalog
     _tbl_md: schema.TableMd
     _version_md: schema.VersionMd
+    # User and system columns that are live in this schema version.
```
mkornacker: We use "visible" elsewhere. Let's avoid adding jargon variants.

pixeltable/catalog/table_version.py
```diff
@@ -352,14 +352,13 @@ def create_initial_md(
                 index_md[idx_id] = md
 
             for col in index_cols:
-                col_md, _ = col.to_md()
+                col_md, sch_md = col.to_md(pos=None)
```
mkornacker: We write out 'schema' elsewhere (let's stick with existing conventions).

pixeltable/metadata/converters/convert_45.py
```diff
@@ -0,0 +1,71 @@
...
+@register_converter(version=45)
+def _(engine: sql.engine.Engine) -> None:
+    """
+    In version 46, some column metadata moved from the table level (TableMd, ColumnMd) to SchemaVersionMd to allow for
+    versioning of these fields in the future. This converter moves the affected fields accordingly by reading all table
+    and schema version metadata to memory, then updating them and writing them back.
+    """
+    with engine.connect().execution_options(isolation_level='READ COMMITTED') as conn:
+        # Read the table and table schema version metadata from the store
+        # table id -> table md
```
mkornacker: Is this -> schema.TableMd? Best to be clear.

pixeltable/metadata/converters/convert_45.py
```diff
@@ -0,0 +1,71 @@
...
+_COL_FIELDS_TO_MOVE = {'col_type': True, 'is_pk': True, 'value_expr': False, 'destination': False}
+
+
+def _convert_table_and_versions(table_md: dict, schema_versions: dict[int, dict]) -> None:
+    assert len(schema_versions) > 0, table_md.get('tbl_id')
+    for col_id, table_col_md in table_md['column_md'].items():
+        # For each column in table md, find the relevant schema versions in schema versions md and add/update columns
+        # there accordingly
+        # lowest and highest versions are inclusive:
+        lowest_schema_ver = table_col_md['schema_version_add']
```
mkornacker: We often use 'min_...'/'max_...' in the codebase for this kind of thing.

pixeltable/catalog/column.py
```diff
@@ -138,26 +138,29 @@ def __init__(
         self.sa_cellmd_col = None
         self._explicit_destination = destination
 
-    def to_md(self, pos: int | None = None) -> tuple[schema.ColumnMd, schema.SchemaColumn | None]:
-        """Returns the Column and optional SchemaColumn metadata for this Column."""
+    def to_md(self, pos: int | None) -> tuple[schema.ColumnMd, schema.SchemaColumn]:
```
aaron-siegel, line 364: I understand that this particular decision wasn't made in this PR, but: I think `ColumnVersionMd` would be a clearer / more consistent name for the struct than `SchemaColumn`

pixeltable/catalog/table_version.py
```diff
@@ -163,7 +164,7 @@ class TableVersion:
     # target for data operation propagation (only set for non-snapshots, and only records non-snapshot views)
     mutable_views: frozenset[TableVersionHandle]
 
-    # contains complete history of columns, incl dropped ones
+    # User and system columns that are live in this schema version.
```
aaron-siegel: Wait are we changing the meaning of `cols`, or was the comment incorrect before? If the meaning is changing this has major implications? I know there are places in the code where I've relied on very specific meaning of these fields.

pixeltable/catalog/table_version.py
aaron-siegel: I'd add: stick with existing conventions *or* revisit the conventions as a whole in a principled way, but not make ad hoc deviations without a specific reason to do so.

pixeltable/catalog/table_version.py
```diff
@@ -579,11 +578,14 @@ def _init_schema(self) -> None:
         # point backward.
         sorted_column_md = sorted(self.tbl_md.column_md.values(), key=lambda item: item.id)
         for col_md in sorted_column_md:
-            col_type = ts.ColumnType.from_dict(col_md.col_type)
-            schema_col_md = self.schema_version_md.columns.get(col_md.id)
+            if not col_md.is_live_in_version(self.schema_version):
```
aaron-siegel: Now I see we're changing the meaning of `cols`. This has potentially major implications. Are you confident this won't disrupt the logic of any existing references to `cols`?

pixeltable/catalog/table_version.py
```diff
@@ -579,11 +578,14 @@ def _init_schema(self) -> None:
...
+                continue
+
+            schema_col_md = self._schema_version_md.columns[col_md.id]
```
aaron-siegel, line 413: Echoing my above comment, to me, this is best thought of as "col_version_md".

pixeltable/catalog/table_version.py
```diff
@@ -888,28 +887,17 @@ def add_columns_ops(self, cols: Iterable[Column]) -> tuple[TableVersionMd, list[
         self.bump_version(bump_schema_version=True)
 
         # create column md
+        highest_pos = max((c.pos for c in self._schema_version_md.columns.values() if c.pos is not None), default=-1)
+        next_pos = itertools.count(start=highest_pos + 1, step=1)
```
aaron-siegel: Isn't `step=1` implied?

pixeltable/catalog/table_version.py
```diff
@@ -888,28 +887,17 @@ def add_columns_ops(self, cols: Iterable[Column]) -> tuple[TableVersionMd, list[
         self.bump_version(bump_schema_version=True)
 
         # create column md
+        highest_pos = max((c.pos for c in self._schema_version_md.columns.values() if c.pos is not None), default=-1)
```
aaron-siegel: Does this mean that the same `pos` may be occupied by different columns at different times?

pixeltable/catalog/table_version.py
```diff
@@ -1017,13 +1007,11 @@ def _add_columns(
...
             if col.name is not None:
                 self.cols_by_name[col.name] = col
-                col_md, sch_md = col.to_md(len(self.cols_by_name))
-                assert sch_md is not None, 'Schema column metadata must be created for user-facing columns'
-                self._tbl_md.column_md[col.id] = col_md
-                self._schema_version_md.columns[col.id] = sch_md
-            else:
-                col_md, _ = col.to_md()
-                self._tbl_md.column_md[col.id] = col_md
+            col_md, sch_md = col.to_md(next(next_pos) if col.name is not None else None)
```
aaron-siegel: Can this be deduped with the preceding logic?

pixeltable/catalog/table_version.py
```diff
@@ -1106,12 +1094,14 @@ def _drop_columns(self, cols: Iterable[Column]) -> None:
...
-            if col.name is not None:
-                del self._schema_version_md.columns[col.id]
+            del self._schema_version_md.columns[col.id]
 
         # update positions
-        for pos, schema_col in enumerate(self._schema_version_md.columns.values()):
-            schema_col.pos = pos
+        pos = 0
+        for schema_col in self._schema_version_md.columns.values():
+            if not schema_col.is_system_column():
```
aaron-siegel: In some places we're using `is_system_column` and in others, calling `name is not None` directly. Can we standardize?

pixeltable/metadata/converters/convert_45.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@register_converter(version=45)
+def _(engine: sql.engine.Engine) -> None:
+    """
+    In version 46, some column metadata moved from the table level (TableMd, ColumnMd) to SchemaVersionMd to allow for
+    versioning of these fields in the future. This converter moves the affected fields accordingly by reading all table
+    and schema version metadata to memory, then updating them and writing them back.
+    """
+    with engine.connect().execution_options(isolation_level='READ COMMITTED') as conn:
```
aaron-siegel: Any reason to make the isolation level explicit for convert_45 but not in our other migration code? And are we sure this pattern will work on cockroachdb?

tests/data/v45_schema_version_md.json
```diff
@@ -0,0 +1,63 @@
+{
```
aaron-siegel: Why do we need these .json files?

Shouldn't the db dumps be covering the breadth of the metadata changes? If not, I'd prefer to add more coverage to the db dumps, which do a ton more validation (e.g., insert-after-md-upgrade) rather than try to paper over with synthetic ("whitebox") tests.

aaron-siegel (changes requested):

pixeltable/catalog/table_version.py
mkornacker: There shouldn't be any places in the code outside of what's in the catalog module to rely on the internals of TableVersion.

We need to change that code to use public APIs (and in the future make sure that this code doesn't get checked in to start with).

pixeltable/catalog/table_version.py
mkornacker: Yes, that has always been the case.

pixeltable/catalog/column.py
mkornacker: Fine by me.

pixeltable/catalog/table_version.py
```diff
@@ -163,7 +164,7 @@ class TableVersion:
     # target for data operation propagation (only set for non-snapshots, and only records non-snapshot views)
     mutable_views: frozenset[TableVersionHandle]
 
-    # contains complete history of columns, incl dropped ones
+    # User and system columns that are live in this schema version.
     cols: list[Column]
```
mkornacker: You should remove cols altogether (or replace it with a property), it served a purpose historically (it was used to re-generate the -Md records, but we're now storing those explicitly).

cols_by_id subsumes cols.

pixeltable/catalog/column.py
sergey-mkhitaryan: left a todo, this pr is already too large

pixeltable/catalog/table_version.py
sergey-mkhitaryan: This change is the biggest reason why this pr takes so long. Yes, as far as I can tell, everything is correct. Some usages of `cols` are in fact even _more_ correct than before, e.g. `label_studio.py`

pixeltable/metadata/converters/convert_50.py
```diff
@@ -0,0 +1,150 @@
...
+    Changes in version 51:
+    - some column metadata (col_type, is_pk, value_expr) moved from the table level (ColumnMd) to
+    SchemaVersionMd in order to allow for future versioning of these metadata.
+    - Only user columns used to be represented by SchemaVersionMd, now system columns are also there.
+    - New attrs in ColumnMd: is_media_type and sa_col_type.
+
+    This converter reads all table and schema metadata in memory, updates them, and writes them back.
+    """
+    assert isinstance(dbms, (CockroachDbms, PostgresqlDbms)), dbms
+    is_cockroachdb = isinstance(dbms, CockroachDbms)
```
christopherpestano: Do we actually need the new `dbms` param? This is the only converter that uses it, and only to decide Vector vs HalfVec. The same info is already available from the `engine` that converters receive today:

```python
is_cockroachdb = engine.dialect.name == 'cockroachdb'
```

That's the same string `env.py` keys on to construct `CockroachDbms` in the first place, and `get_sa_type` in `utils/sql.py` already dispatches on `engine.dialect.name` the same way. Doing it this way would drop the signature changes to `register_converter`/`upgrade_md` and the edits to the other 41 converters.

pixeltable/metadata/converters/convert_50.py
```diff
@@ -0,0 +1,150 @@
...
+
+import sqlalchemy as sql
+
+from pixeltable.metadata import register_converter
+from pixeltable.metadata.schema import Table, TableSchemaVersion
+from pixeltable.utils.dbms import CockroachDbms, Dbms, PostgresqlDbms
+
+
+@register_converter(version=50)
+def _(engine: sql.engine.Engine, dbms: Dbms) -> None:
```
christopherpestano: Has this been validated against a real cockroachdb instance? If not you should pull the latest main and test it against the cockroachdb container, it could help validate that the conversion works as expected.

pixeltable/type_system.py
```diff
@@ -762,14 +763,33 @@ def from_np_dtype(cls, dtype: np.dtype, nullable: bool) -> ColumnType | None:
...
+    ColumnType.Type.ARRAY: sql.LargeBinary(),
+    ColumnType.Type.IMAGE: sql.String(),
+    ColumnType.Type.VIDEO: sql.String(),
+    ColumnType.Type.AUDIO: sql.String(),
+    ColumnType.Type.DOCUMENT: sql.String(),
+}
+
+
+def to_sa_type(t: ColumnType.Type) -> sql.types.TypeEngine:
+    return _SA_TYPE_MAP[t]
```
christopherpestano, line 813: Is this the right implementation, we have modified this to have non-abstract implementation without changing the behavior of the subclasses. The behavior of this SA_TYPE_MAP and the subclasses can easily diverge, try to unify somehow so there is a single source of truth.

pixeltable/metadata/schema.py
```diff
@@ -325,21 +336,41 @@ class TableVersion(Base):
...
 
-    pos: int
-    name: str
+    # pos and name must be set for user columns, and be None for system columns
+    pos: int | None
+    name: str | None
+
+    col_type: dict
+    # True if this column is part of the primary key
+    is_pk: bool
```
christopherpestano, line 343: Are we going to support ALTER primary key? I was under the presumption that primary key's were to remain fixed even post schema-evolution

christopherpestano (changes requested):

pixeltable/metadata/converters/convert_51.py
```diff
@@ -0,0 +1,146 @@
...
+            if schema_ver < table_col_md['schema_version_add']:
+                continue
+            if schema_ver >= (table_col_md.get('schema_version_drop') or Table.MAX_VERSION):
+                continue
+            # Update user-visible columns and add system columns
+            col = schema_version_md['columns'].setdefault(col_id, {'pos': None, 'name': None, 'media_validation': None})
+            for field, is_required in _COL_FIELDS_TO_MOVE.items():
+                assert field in table_col_md or not is_required, (field, table_col_md)
+                if field in table_col_md:
+                    col[field] = table_col_md[field]
```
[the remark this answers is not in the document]
sergey-mkhitaryan: `value_expr` was required in `ColumnMd`

pixeltable/metadata/converters/convert_50.py
sergey-mkhitaryan: yeah, this is good. done.

pixeltable/metadata/converters/convert_50.py
sergey-mkhitaryan: a pretty recent `main` is here

pixeltable/type_system.py
sergey-mkhitaryan: deleted leftovers

pixeltable/metadata/schema.py
sergey-mkhitaryan: technically it's just a unique constraint, so yeah we can support alter pk. whether or when we will actually do it, I don't know.

pixeltable/type_system.py
```diff
@@ -1980,3 +1984,50 @@ def as_col_type(cls, nullable: bool) -> ColumnType:
...
+    if isinstance(t, sql.types.TIMESTAMP):
+        assert t.timezone
+    if isinstance(t, sql.types.UUID):
+        assert t.as_uuid
+    if isinstance(t, sql.types.LargeBinary):
+        assert t.length is None
+    if isinstance(t, (sql_vector.HALFVEC, sql_vector.VECTOR)):
+        assert t.dim is not None
+        d['dim'] = t.dim
+    return d
```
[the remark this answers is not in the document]
sergey-mkhitaryan: test migration

pixeltable/catalog/table_version.py
```diff
@@ -428,70 +425,55 @@ def _init_schema(self) -> None:
...
                 schema_version_add=col_md.schema_version_add,
                 schema_version_drop=col_md.schema_version_drop,
                 stores_cellmd=col_md.stores_cellmd,
                 computed_with=value_expr,
-                value_expr_dict=col_md.value_expr,
+                value_expr_dict=schema_col_md.value_expr,
                 tbl_handle=self.handle,
                 destination=col_md.destination,
                 custom_metadata=schema_col_md.custom_metadata if schema_col_md is not None else None,
                 comment=schema_col_md.comment if schema_col_md is not None else '',
```
christopherpestano: Do these not None checks do anything? The type hint says not None.

pixeltable/catalog/column.py
```diff
@@ -211,23 +215,32 @@ def create(cls, name: str, spec: type | ColumnSpec | exprs.Expr) -> Column:
         custom_metadata: Any = None
         comment: str | None = None
 
+        sa_col_type: sql.types.TypeEngine = None
```
christopherpestano: Shouldn't this be `sql.types.TypeEngine | None`?

pixeltable/type_system.py
```diff
@@ -14,6 +14,8 @@
 from pathlib import Path
 from typing import Any, ClassVar, Iterable, Literal, Mapping, Sequence, Union
 
+import pgvector.sqlalchemy as sql_vector  # type: ignore[import-untyped]
```
christopherpestano: Different import style from column.py, should probably just make this import pgvector.sqlalchemy and make the refs like `pgvector.sqlalchemy.Vector`

christopherpestano (approved):

pixeltable/catalog/column.py
```diff
@@ -563,16 +578,18 @@ def source(self) -> None:
...
-        """
-        These need to be recreated for every sql.Table instance
-        """
-        assert self.is_stored
-        assert self.stores_cellmd is not None
-        # all storage columns are nullable (we deal with null errors in Pixeltable directly)
-        self.sa_col = sql.Column(self.store_name(), self.sa_col_type, nullable=True)
+    def set_sa_cols(self, sa_col: sql.Column, sa_cellmd_col: sql.Column | None) -> None:
+        # `type() is` check is weak as it ignores some SQL type parameters (e.g. vector dimensions), but anything
+        # stronger would be unnecessary complicated.
```
aaron-siegel: typo (unnecessarily)

aaron-siegel: Also, why would `assert sa_col.type == self.sa_col_type` not be correct? (Just curious - and maybe worth a code comment explaining when/why this is used)

pixeltable/catalog/column.py
```diff
@@ -563,16 +578,18 @@ def source(self) -> None:
...
 
-    def create_sa_cols(self) -> None:
-        """
-        These need to be recreated for every sql.Table instance
-        """
-        assert self.is_stored
-        assert self.stores_cellmd is not None
-        # all storage columns are nullable (we deal with null errors in Pixeltable directly)
-        self.sa_col = sql.Column(self.store_name(), self.sa_col_type, nullable=True)
+    def set_sa_cols(self, sa_col: sql.Column, sa_cellmd_col: sql.Column | None) -> None:
```
aaron-siegel, line 581: `set_sa_col()` - singular?

pixeltable/metadata/schema.py
```diff
@@ -346,13 +372,14 @@ class SchemaVersionMd:
...
     preceding_schema_version: int | None
+    # User and system columns visible in this schema version
     columns: dict[int, SchemaColumn]  # col_id -> SchemaColumn
     # TODO: Before next release, add migration to preexisting empty strings to None
     comment: str | None
 
     # default validation strategy for any media column of this table
-    # stores column.MediaValiation.name.lower()
-    media_validation: str
+    # stores column.MediaValidation.name.lower() or None if unspecified
```
aaron-siegel: What does it mean for it to be unspecified? Why do we need to distinguish between "unspecified" and "on_write"?

pixeltable/metadata/utils.py
```diff
@@ -20,28 +22,68 @@ def _diff_md(
...
         t = []
-        if len(added) > 0:
-            t.append('Added: ' + ', '.join(added.values()))
-        if len(changed) > 0:
-            t.append('Renamed: ' + ', '.join(changed.values()))
-        if len(deleted) > 0:
-            t.append('Deleted: ' + ', '.join(deleted.values()))
-        r = ', '.join(t)
-        return r
+        if added:
```
aaron-siegel: Elsewhere in the codebase we avoid "implicit booleans" where we can and write `len(added) > 0` explicitly here. I don't feel super strongly about it but there's an argument for maintaining consistency.

pixeltable/metadata/schema.py
```diff
@@ -321,21 +327,41 @@ class TableVersion(Base):
...
 
-    pos: int
-    name: str
+    # pos and name must be set for user columns, and be None for system columns
+    pos: int | None
+    name: str | None
+
+    col_type: dict
+    # True if this column is part of the primary key
+    is_pk: bool
```
mkornacker, line 343: You could argue that the primary key (and its enforcing unique constraint) is part of the physical schema and should be in ColumnMd. When we implement .alter(primary_key=[...]) we basically need to do a physical "migration" (check uniqueness, then create the index), and the enforced primary key is only ever valid for the live version anyway. Analogous to .alter_column() that would change the physical type.

Counter-argument: a snapshot could have a more restrictive primary key, which might be of interest to the user.

pixeltable/metadata/utils.py
```diff
@@ -20,28 +22,68 @@ def _diff_md(
...
-        changed = {
-            k: f'{old_md[k].name!r} to {v.name!r}'
-            for k, v in new_md.items()
-            if k in old_md and old_md[k].name != v.name
-        }
-        deleted = {k: v.name for k, v in old_md.items() if k not in new_md}
-        if len(added) == 0 and len(changed) == 0 and len(deleted) == 0:
+        # added, altered, and dropped track diffs of user-visible columns only. That's what we want to report to users
+        # in the end.
+        added, dropped = [], []
```
mkornacker: Missing type decls

(and elsewhere; please fix all of them)

pixeltable/metadata/utils.py
```diff
@@ -20,28 +22,68 @@ def _diff_md(
...
+        if old.name != new.name:
+            diff.append(f'renamed to {new.name}')
+        assert old.is_pk == new.is_pk, 'Not implemented: describe a primary key change'
+        assert old.col_type == new.col_type, 'Not implemented: describe a column type change'
+        assert old.value_expr == new.value_expr, 'Not implemented: describe a value expression change'
+        assert old.media_validation == new.media_validation, 'Not implemented: describe a media validation change'
+        assert old.comment == new.comment, 'Not implemented: describe a comment change'
+        assert old.custom_metadata == new.custom_metadata, 'Not implemented: describe a custom metadata change'
+
+        if diff:
```
mkornacker: We use explicit comparisons throughout.

pixeltable/type_system.py
```diff
@@ -684,12 +686,11 @@ def supports_file_offloading(self) -> bool:
         # types that can be offloaded to file-based storage via a CellMaterializationNode
         return self.is_array_type() or self.is_json_type() or self.is_binary_type()
 
-    @classmethod
-    @abc.abstractmethod
-    def to_sa_type(cls) -> sql.types.TypeEngine:
+    def to_sa_type(self) -> sql.types.TypeEngine:
```
mkornacker, line 689: What's the point of restructuring this? It seems like a purely syntactic change.

pixeltable/type_system.py
```diff
@@ -1980,3 +1941,50 @@ def as_col_type(cls, nullable: bool) -> ColumnType:
...
+    sql.dialects.postgresql.json.JSONB: 'JSONB',
+    pgvector.sqlalchemy.HALFVEC: 'HalfVec',
+    pgvector.sqlalchemy.Vector: 'Vector',
+}
+
+_SA_TYPE_BY_NAME: dict[str, type] = {name: t for t, name in _SA_TYPE_NAMES.items()}
+
+
+def sa_type_as_dict(t: sql.types.TypeEngine) -> dict:
+    d = {'type': _SA_TYPE_NAMES[type(t)]}
```
mkornacker, line 1991: Why not just type(t).__name__ and remove the lookup table?

pixeltable/type_system.py
```diff
@@ -1980,3 +1941,50 @@ def as_col_type(cls, nullable: bool) -> ColumnType:
...
+    if isinstance(t, sql.types.String):
+        assert t.length is None
+    if isinstance(t, sql.types.TIMESTAMP):
+        assert t.timezone
```
mkornacker, line 1995: here and elsewhere: explicit comparison

pixeltable/type_system.py
```diff
@@ -1980,3 +1941,50 @@ def as_col_type(cls, nullable: bool) -> ColumnType:
...
+    if isinstance(t, sql.types.LargeBinary):
+        assert t.length is None
+    if isinstance(t, (pgvector.sqlalchemy.HALFVEC, pgvector.sqlalchemy.Vector)):
+        assert t.dim is not None
+        d['dim'] = t.dim
+    return d
+
+
+def sa_type_from_dict(d: dict) -> sql.types.TypeEngine:
+    clazz = _SA_TYPE_BY_NAME[d['type']]
```
mkornacker, line 2007: or: getattr(sql.types, d['type'])

pixeltable/catalog/column.py
```diff
@@ -67,7 +68,7 @@ class Column:
     schema_version_drop: int | None
     stores_cellmd: bool
     sa_col: sql.schema.Column | None
-    sa_col_type: sql.types.TypeEngine
+    sa_col_type: sql.types.TypeEngine | None
```
mkornacker: comment: when is it None

pixeltable/catalog/column.py
sergey-mkhitaryan: sqlalchemy types do not implement `__eq__` so that won't work. writing a comment.

pixeltable/catalog/column.py
sergey-mkhitaryan: but it sets two columns -- data column and cellmd column

mkornacker: It's also setting sa_cellmd_col

pixeltable/metadata/schema.py
sergey-mkhitaryan: I'm actually not sure why I even changed this. Maybe I confused it with `SchemaColumn.media_validation`. Reverting.

pixeltable/metadata/schema.py
sergey-mkhitaryan: Pixeltable primary key is basically a unique constraint, which, in my mind, is not unlike b-tree or secondary indexes that the user can define. That said, @christopherpestano was also confused by why it's here, so maybe I'm the only one who sees it that way. Can move it back, fine by me either way.

pixeltable/type_system.py
sergey-mkhitaryan: I don't mind reverting this, I think it's an artifact left from the time when we wanted `ColumnMd` to contain a `ColumnType.Type` of that column. Still this structure is less boilerplate than before, and it better captures the fact that the sa type is the product of `ColumnType.Type` alone and not any other pxt type attributes.

pixeltable/type_system.py
sergey-mkhitaryan: It gets us stable persisted names. It also enforces an allow-list of types, which we do want because some types (like halfvec and vector below) need special handling here.

pixeltable/type_system.py
sergey-mkhitaryan: are you sure, it's a boolean

pixeltable/type_system.py
sergey-mkhitaryan: But not all of them are in `sql.types`. There's `sql.dialects.postgresql.json.JSONB`, `pgvector.sqlalchemy.HALFVEC`, etc.

pixeltable/type_system.py
mkornacker: I didn't realize that, I thought it was a 'is not None'.

pixeltable/type_system.py
mkornacker: The stable names argument doesn't really hold: Sqlalchemy isn't going to change the name of the String class, that would be a very major API change (for 0 gain).

pixeltable/metadata/schema.py
mkornacker: It can be useful (as schema information) in snapshots, so might as well leave it there.

But if it reflects something about the stored table (= there's only one of those) vs the different table versions that are combined in that table, it should be in ColumnMd.

mkornacker (approved):

aaron-siegel (approved):

PR #1096, opened 2026-01-27
PXT-767 Support Array Column as input to add_embedding_index

amithadke:


pixeltable/index/embedding_index.py
```diff
@@ -275,6 +282,26 @@ def _validate_embedding_fn(cls, embed_fn: func.Function) -> None:
...
             )
 
+        # If column is an array type, validate that the embedding function's return type matches
+        if column is not None and isinstance(column.col_type, ts.ArrayType):
+            col_array_type = column.col_type
+            # Validate shape compatibility
+            if shape is not None and col_array_type.shape is not None and shape != col_array_type.shape:
+                raise excs.Error(
+                    f'The function `{embed_fn.name}` returns an array with shape {shape}, '
+                    f'but column {column.name!r} requires shape {col_array_type.shape}'
```
aaron-siegel: `s/requires/has/` would be clearer.

pixeltable/index/embedding_index.py
```diff
@@ -275,6 +282,26 @@ def _validate_embedding_fn(cls, embed_fn: func.Function) -> None:
...
+                )
+            # Validate dtype compatibility
+            if (
+                return_type.dtype is not None
+                and col_array_type.dtype is not None
+                and return_type.dtype != col_array_type.dtype
+            ):
+                raise excs.Error(
+                    f'The function `{embed_fn.name}` returns an array with dtype {return_type.dtype}, '
```
aaron-siegel: I definitely don't think we should be so strict here. It should be fine if the dtypes are not identical.

pixeltable/index/embedding_index.py
```diff
@@ -65,6 +65,7 @@ def __init__(
         image_embed: func.Function | None = None,
         audio_embed: func.Function | None = None,
         video_embed: func.Function | None = None,
+        column: catalog.Column | None = None,
```
aaron-siegel: Is this passed in only for array embeddings, or for all embeddings? (Clarify)

pixeltable/index/embedding_index.py
```diff
@@ -284,15 +311,17 @@ def as_dict(self) -> dict:
 
     @classmethod
     def from_dict(cls, d: dict) -> EmbeddingIndex:
+        embed = func.Function.from_dict(d['embed']) if d.get('embed') is not None else None
```
aaron-siegel: This was intentionally left out and shouldn't be serialized. It gets resolved to the specific modalities in the EmbeddingIndex initializer.

pixeltable/plan.py
```diff
@@ -18,6 +18,16 @@ def _is_agg_fn_call(e: exprs.Expr) -> bool:
     return isinstance(e, exprs.FunctionCall) and e.is_agg_fn_call and not e.is_window_fn_call
 
 
+def _dedupe_col_refs(candidates: Iterable[exprs.Expr]) -> list[exprs.ColumnRef]:
```
aaron-siegel: Isn't this happening automatically in the planner? I guess it happens later?

What are the implications of this?

aaron-siegel (changes requested):
Two questions I have:
- What if I don't have an embedding function from a non-array modality, and I just want to be able to do top-k (indexed) lookups against a given array? Is the embedding function optional if the array is explicitly specified? (I think it should be.)
- How would I do the top-k lookup against an array (rather than a string, say)?

As I mentioned before... I think it would simplify things a bit if we simply treated "Array" as another modality, for which the "embedding function" is the identify function. Then I think we wouldn't need to special-case it; it would just fit cleanly into the existing mechanism. I don't feel super strongly about this, just pointing it out.

pixeltable/index/embedding_index.py
amithadke: just needed for an array (for validation)

pixeltable/plan.py
amithadke: no not in this path.  list_subexprs can return the same ColumnRef more than once.

We were having 2 expr one from select (array column) and then one ref coming from similarity expression and since its externally stored array we had dups in array refs during cell reconstruction.

pixeltable/index/embedding_index.py
aaron-siegel: I meant: please clarify this in the python file with a code comment.

amithadke:
There are couple of issues
- dedup its needed when u run query like t.select(t.text, t.text.similarity(string='a cat')).limit(5).collect() where t.text is selected along with similarity too
may be we can remove dedup inside CellreConstructionNode and pass row_builder.unique_exprs to _add_cell_reconstruction_node and that should fix this issue

- Redundant column ref from similarity expression
for res = t.select(t.text.similarity(string='a cat')).limit(5).collect() (we don't select column t.text)

analyzer/planner treat SimilarityExpr’s indexed column as a normal dependency and add it to the expression set that gets slots and is (or may be) evaluated, even though similarity is computed in SQL and that column’s value is never needed — so we get redundant expressions that don’t need to be evaluated

fundamentally similarity expression uses 2 components column ref for indexed column and then actual index column on which sql  is generated/evaluated. column ref for indexed column is used only for serialization and finding out correct vector index column. We use list_subexpr to traverse the tree and when we are using it to find out expressions to evaluate we are adding column ref from similarity expression which really doesnt need to be evaluated at all. there are couple of ways to fix it  
- add flag to list_subexpr so it knows to traverse expresions for evaluation and modify similarity expression to exclue index colum ref
 - or change similarity expression so that it doesnt need to store index column in components (which means we will have to change as_dict and from_dict to not rely on column ref)

amithadke:
> Two questions I have:
> 
> * What if I don't have an embedding function from a non-array modality, and I just want to be able to do top-k (indexed) lookups against a given array? Is the embedding function optional if the array is explicitly specified? (I think it should be.)
> * How would I do the top-k lookup against an array (rather than a string, say)?
> 
> As I mentioned before... I think it would simplify things a bit if we simply treated "Array" as another modality, for which the "embedding function" is the identify function. Then I think we wouldn't need to special-case it; it would just fit cleanly into the existing mechanism. I don't feel super strongly about this, just pointing it out.

cant we do both? if you just want to query indexed column with vector as input we can allow array modality and when array is passed in we lookup for function that takes vector/array as input (in most cases identity function - but doesnt need to be limited to it, it can be any function array->array) 
if string/image is passed we will match it with relevant function and then based on which embedding function was matched we can  decide to either evaluate it or return column ref.

aaron-siegel:
@amithadke I see a couple of my comments that are marked as resolved but without accompanying code changes or responses - do you have pending commits that haven't been pushed yet? (Just want to know when I should re-review)

amithadke:
> @amithadke I see a couple of my comments that are marked as resolved but without accompanying code changes or responses - do you have pending commits that haven't been pushed yet? (Just want to know when I should re-review)

I've fixed those but not pushed changes. I'm waiting for decision to add array in similarity.

pixeltable/exprs/column_ref.py
```diff
@@ -180,6 +181,7 @@ def similarity(
         image: str | PIL.Image.Image | None = None,
         audio: str | None = None,
         video: str | None = None,
+        array: np.ndarray | list[float] | None = None,
```
aaron-siegel: Do we really want to take `list[float]` here? There's an argument it's better for the user to make it explicit by wrapping it in an `np.array()`, and avoid the type munging.

pixeltable/exprs/similarity_expr.py
```diff
@@ -62,42 +71,56 @@ def sql_expr(self, _: SqlElementCache) -> sql.ColumnElement | None:
...
 
     def eval(self, data_row: DataRow, row_builder: RowBuilder) -> None:
         raise excs.Error('similarity(): cannot be used in a computed column')
 
     def _as_dict(self) -> dict:
-        return {'idx_name': self.idx_name, **super()._as_dict()}
+        return {
+            'idx_name': self.idx_name,
+            'tbl_version_key': self.tbl_version_key.as_dict(),
+            'components': [c.as_dict() for c in self.components],
```
aaron-siegel: Let's please follow the pattern used by other `Expr` classes, which is to call `**super()._as_dict()` to serialize the components.
- If that doesn't work here for some reason, please add a code comment explaining why
- If there's a principled reason to prefer this one, I'd argue it's still best to stick with the existing one for now for consistency, and then raise it as a refactoring/techdebt opportunity at the eng meeting, so that we can address it in a principled way rather than ad hoc.

pixeltable/exprs/similarity_expr.py
```diff
@@ -62,42 +71,56 @@ def sql_expr(self, _: SqlElementCache) -> sql.ColumnElement | None:
...
+        tbl_version_key = TableVersionKey.from_dict(d['tbl_version_key'])
         idx_name = d.get('idx_name')
-        assert len(components) == 2
-        assert isinstance(components[0], ColumnRef)
-        return cls(components[0], components[1], idx_name=idx_name)
+        return cls._from_index_name(item=components[0], idx_name=idx_name, tbl_version_key=tbl_version_key)
+
+    @classmethod
+    def _from_index_name(cls, *, item: Expr, idx_name: str, tbl_version_key: TableVersionKey) -> 'SimilarityExpr':
+        obj = cls.__new__(cls)
```
aaron-siegel: Why do we need to do it this way? It seems to me this makes the code harder to understand and increases the possibility of subtle bugs. I understand that the goal here is to differentiate between the two instantiation paths - new vs. from catalog. But all the other `Expr` classes have the same bifurcation, and they all route through the class initializer in both code paths. Is there a reason for preferring to do things differently here?

pixeltable/functions/array.py
```diff
@@ -0,0 +1,24 @@
+"""
+Pixeltable UDFs for array/vector operations.
+"""
+
+import pixeltable as pxt
+from pixeltable.utils.code import local_public_names
+
+
+@pxt.udf
+def identity(vec: pxt.Array[(None,), pxt.Float]) -> pxt.Array[(None,), pxt.Float]:
```
aaron-siegel: Suggestion: Put this in `pixeltable.functions` and make it a generic identity function (that can take any type).

pixeltable/index/embedding_index.py
```diff
@@ -65,9 +65,13 @@ def __init__(
         image_embed: func.Function | None = None,
         audio_embed: func.Function | None = None,
         video_embed: func.Function | None = None,
+        array_embed: func.Function | None = None,
+        column: catalog.Column | None = None,  # Used for validation when the indexed column is an array.
     ):
-        if embed is None and string_embed is None and image_embed is None:
-            raise excs.Error('At least one of `embed`, `string_embed`, or `image_embed` must be specified')
+        if embed is None and string_embed is None and image_embed is None and array_embed is None:
```
aaron-siegel: audio_embed and video_embed are missing here - that's a bug - might as well fix it now

tests/test_index.py
```diff
@@ -67,7 +68,7 @@ def test_similarity_multiple_index(
         t.drop_embedding_index(idx_name='img_idx1')
         with pytest.raises(pxt.Error) as exc_info:
             reload_tester.run_reload_test(clear=False)
-        assert "index 'img_idx1' not found" in str(exc_info.value).lower()
+        assert 'img_idx1' in str(exc_info.value) and 'not found' in str(exc_info.value).lower()
```
aaron-siegel: Note: we've been slowly migrating to using the `pytest.raises(pxt.Error, match=...)` pattern - and when you have to edit one of these for some other reason, that's a good opportunity to do so. (Not critical, but just wanted to mention this.)

tests/test_index.py
```diff
@@ -981,3 +984,90 @@ def test_embedding_index_precision(
         sim = t.text.similarity(string='one')
         res = t.select(t.rowid, t.text, sim=sim).order_by(sim, asc=False).collect()
         assert res[0]['rowid'] == 1
+
+    def test_array_column_embedding_index(
+        self, uses_db: None, e5_embed: pxt.Function, reload_tester: ReloadTester
+    ) -> None:
+        skip_test_if_not_installed('transformers')
+        texts = ['a dog playing in the park', 'a cat sitting on a mat', 'a bird flying in the sky']
+        vecs = [e5_embed.exec([s], {}) for s in texts]
```
mkornacker: Let's stick to the public interface (which this isn't part of).

tests/test_index.py
```diff
@@ -981,3 +984,90 @@ def test_embedding_index_precision(
         sim = t.text.similarity(string='one')
         res = t.select(t.rowid, t.text, sim=sim).order_by(sim, asc=False).collect()
         assert res[0]['rowid'] == 1
+
+    def test_array_column_embedding_index(
+        self, uses_db: None, e5_embed: pxt.Function, reload_tester: ReloadTester
+    ) -> None:
```
mkornacker, line 991: Use blank lines to break up code where appropriate.

tests/test_index.py
```diff
@@ -981,3 +984,90 @@ def test_embedding_index_precision(
...
+        )
+        # f32 precomputed embedding with string embedding function
+        t.add_embedding_index('vec', idx_name='emb_stored', string_embed=e5_embed, metric='cosine', precision='fp32')
+        # f64 precomputed embedding column with string embedding function
+        t.add_embedding_index(
+            'vec64', idx_name='emb_stored64', string_embed=e5_embed, metric='cosine', precision='fp32'
+        )
+        # Array column with array embedding function
+        t.add_embedding_index(
+            'vec_array_f32', idx_name='emb_array_f32_cosine', array_embed=identity, metric='cosine', precision='fp32'
```
mkornacker: I don't understand why the array_embed parameter is useful (or why it would ever be anything other than identify).

tests/test_index.py
```diff
@@ -981,3 +984,90 @@ def test_embedding_index_precision(
...
+        return np.zeros(256, dtype=np.float32)
+
+    def test_array_embedding_index_validation_errors(self, uses_db: None) -> None:
+        t = pxt.create_table(
+            'arr_val_test',
+            {'id': pxt.Int, 'vec': pxt.Array[(384,), np.float32]},  # type: ignore[misc]
+            if_exists='replace',
+        )
+        t.insert([{'id': 0, 'vec': np.zeros(384, dtype=np.float32)}])
+        with pytest.raises(pxt.Error) as exc_info:
```
mkornacker: use match=

pixeltable/catalog/table.py
```diff
@@ -1069,6 +1070,8 @@ def add_embedding_index(
             image_embed: An optional UDF to use for the image embedding component of this index.
                 Can be used in conjunction with `string_embed` to construct multimodal embeddings manually, by
                 specifying different embedding functions for different data types.
+            array_embed: An optional UDF to use for the array/vector component of this index. Use when the
```
mkornacker: Is this for the case where you're searching with an embedding directly? Why would this require an embedding function?

pixeltable/exprs/similarity_expr.py
```diff
@@ -21,22 +24,55 @@
 class SimilarityExpr(Expr):
     """
     A similarity expression against an embedding index.
+    Resolved by index name so that drop + recreate under the same name still works.
```
mkornacker: Line 32 sounds like it's contradicting that.

pixeltable/exprs/similarity_expr.py
```diff
@@ -21,22 +24,55 @@
...
+            column = tv.cols_by_id[self.col_id]
+            assert column is not None
 
         # determine index to use
-        idx_info = col_ref.tbl.get().get_idx(col_ref.col, idx_name, EmbeddingIndex)
-        self.idx_id = idx_info.id
-        self.idx_name = idx_info.name
+        if self.idx_name is None:
+            # Look up index by column
+            idx_info = tv.get_idx(column, idx_name, EmbeddingIndex)
```
mkornacker: In this case, idx_name is also None. It's clearer to pass that in directly.

pixeltable/exprs/similarity_expr.py
```diff
@@ -21,22 +24,55 @@
...
+            self.idx_name = idx_info.name
+        else:
+            # Lookup index by name
```
mkornacker: Look up

pixeltable/exprs/similarity_expr.py
```diff
@@ -50,10 +86,14 @@ def __init__(self, col_ref: ColumnRef, item: Expr, idx_name: str | None = None):
         self.id = self._create_id()
 
     def __repr__(self) -> str:
-        return f'{self.components[0]}.similarity({self.components[1]})'
+        idx_info = self._resolve_idx()
+        return f'{idx_info.name}.similarity({self.components[0]})'
```
mkornacker: This is printing the index name, not the name of the indexed column.

Is this tested anywhere?

What about idx_name in similarity()?

pixeltable/exprs/column_ref.py
```diff
@@ -180,6 +181,7 @@ def similarity(
         image: str | PIL.Image.Image | None = None,
         audio: str | None = None,
         video: str | None = None,
+        array: np.ndarray | None = None,
```
mkornacker: We could also call this embedding, just to be absolutely clear.

Since this is user-facing, start a thread in the product channel to get consensus.

pixeltable/exprs/column_ref.py
```diff
@@ -271,7 +278,27 @@ def similarity(
...
+                if not array.col_type.is_array_type():
+                    raise excs.Error(f'similarity(array=...): expected `Array`; got `{array.col_type}`')
+                expr = array
+            else:
+                if not isinstance(array, np.ndarray):
+                    raise excs.Error(
+                        f'similarity(array=...): expected `numpy.ndarray`, or array `Expr`; '
+                        f'got `{type(array).__name__}`'
+                    )
+                arr = np.asarray(array)
```
mkornacker: What is this good for?

tests/test_index.py
```diff
@@ -981,3 +984,90 @@ def test_embedding_index_precision(
...
+        texts = ['a dog playing in the park', 'a cat sitting on a mat', 'a bird flying in the sky']
+        vecs = [e5_embed.exec([s], {}) for s in texts]
+        dim = len(vecs[0])
+        vecs64 = [v.astype(np.float64) for v in vecs]
+        t = pxt.create_table(
+            'array_embedding_test',
+            {
+                'id': pxt.Int,
+                'text': pxt.String,
+                'vec': pxt.Array[(dim,), np.float32],  # type: ignore[misc]
```
mkornacker: Column naming is inconsistent.

tests/test_index.py
```diff
@@ -981,3 +984,90 @@ def test_embedding_index_precision(
...
+            res = reload_tester.run_query(query)
+            assert len(res) == 3, col_name
+            assert res[0]['text'] == best, col_name
+            sim_vals = [r['sim'] for r in res]
+            assert all(sim_vals[i] >= sim_vals[i + 1] for i in range(len(sim_vals) - 1)), (
+                f'{col_name}:{idx_name}: similarity scores must be descending; got {sim_vals}'
+            )
+        # Search by vector (array modality)
+        for idx_name in ('emb_array_f32_cosine', 'emb_array_f32_l2'):
+            sim = t.vec_array_f32.similarity(array=best_vec, idx=idx_name)
```
mkornacker: This should also work for other embedding indices.

pixeltable/exprs/column_ref.py
amithadke: ohh i had list[float] earlier will take this out as we wont support list anymore

tests/test_index.py
amithadke: had to stop using identity function, so now if we search by array/vector we by pass any embedding function valuation use value directly

pixeltable/exprs/similarity_expr.py
amithadke: fixed, added test case in test_similarity in test_expr.py 
it doesnt need to resolve index we already have index name now so added it j
it will look like - 
"img.img_idx0.similarity('red truck')"

pixeltable/index/embedding_index.py
```diff
@@ -99,17 +111,16 @@ def __init__(
                 if resolved_fn is not None:
                     self.embeddings[embed_type] = resolved_fn
 
-        if len(self.embeddings) == 0:
+        if embed and len(self.embeddings) == 0:
```
mkornacker: We generally do is not None explicitly.

pixeltable/index/embedding_index.py
```diff
@@ -178,11 +195,39 @@ def sa_create_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql
...
-        assert item.col_type._type in self.embeddings
-        embedding = self.embeddings[item.col_type._type].exec([item.val], {})
+        if item.col_type._type == ts.ColumnType.Type.ARRAY:
+            # Array value is already a vector; no embedding function needed.
+            embedding = item.val
+        else:
+            assert item.col_type._type in self.embeddings
+            embedding = self.embeddings[item.col_type._type].exec([item.val], {})
         assert isinstance(embedding, np.ndarray)
+        self._validate_similarity_input_vector(embedding, val_column, item.col_type._type)
```
mkornacker: This is only needed for the if branch, we're validating that statically if we getting an embedding udf.

pixeltable/index/embedding_index.py
```diff
@@ -178,11 +195,39 @@ def sa_create_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql
...
+    def _validate_similarity_input_vector(
+        self, input_vector: np.ndarray, val_column: catalog.Column, query_type: ts.ColumnType.Type
+    ) -> None:
+        """Validate the input vector used in the similarity clause matches the index column dimensions."""
+        if input_vector.ndim != 1:
+            raise excs.Error(f'Similarity input vector must be 1-dimensional; got shape {input_vector.shape}')
+        assert isinstance(val_column.col_type, ts.ArrayType), 'val_column must be ArrayType for embedding index'
+        col_shape = val_column.col_type.shape
+        assert col_shape is not None and len(col_shape) == 1 and col_shape[0] is not None, (
+            'val_column must have known 1D shape (enforced by get_index_sa_type)'
```
mkornacker: Is this llm output? What is the point of the "(enforced by ...)"?

pixeltable/index/embedding_index.py
```diff
@@ -178,11 +195,39 @@ def sa_create_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql
...
+        )
+        expected_len = col_shape[0]
+        if input_vector.shape[0] != expected_len:
+            raise excs.Error(
+                f'Similarity input vector length {input_vector.shape[0]} does not match '
```
mkornacker: Is this the user-facing error message? In that case, it should probably be more explicit, such as:
similarity(vector=...): vector length does not match ...

Also, "query vector" is probably easier to understand than "similarity input vector" (both in the message and in function names).

pixeltable/index/embedding_index.py
```diff
@@ -194,14 +239,18 @@ def similarity_clause(self, val_column: catalog.Column, item: exprs.Literal) ->
...
-        assert item.col_type._type in self.embeddings
-        embedding = self.embeddings[item.col_type._type].exec([item.val], {})
+        if item.col_type._type == ts.ColumnType.Type.ARRAY:
+            # Array value is already a vector; no embedding function needed.
+            embedding = item.val
+        else:
+            assert item.col_type._type in self.embeddings
+            embedding = self.embeddings[item.col_type._type].exec([item.val], {})
         assert isinstance(embedding, np.ndarray)
+        self._validate_similarity_input_vector(embedding, val_column, item.col_type._type)
```
mkornacker: Same here.

pixeltable/index/embedding_index.py
```diff
@@ -257,19 +306,38 @@ def _validate_embedding_fn(cls, embed_fn: func.Function) -> None:
...
                 f'The function `{embed_fn.name}` is not a valid embedding: '
                 f'it returns an array of invalid length {shape[0]}'
             )
 
+        # Deserialization when no column context is passed, skip column-specific validation.
+        if column is None:
+            return
+
+        # Indexing an array column with precomputed embeddings
+        if isinstance(column.col_type, ts.ArrayType):
```
mkornacker: Is it valid to pass a non-array column?

pixeltable/index/embedding_index.py
```diff
@@ -257,19 +306,38 @@ def _validate_embedding_fn(cls, embed_fn: func.Function) -> None:
...
+            col_shape = column.col_type.shape
+            if col_shape is None or len(col_shape) != 1 or col_shape[0] is None or col_shape[0] <= 0:
```
mkornacker: This check needs to happen even if you don't have embedding functions (= so it shouldn't be here).

pixeltable/index/embedding_index.py
```diff
@@ -246,7 +295,7 @@ def _resolve_embedding_fn(cls, embed_fn: func.Function, expected_type: ts.Column
         return None
 
     @classmethod
-    def _validate_embedding_fn(cls, embed_fn: func.Function) -> None:
+    def _validate_embedding_fn(cls, embed_fn: func.Function, column: catalog.Column | None = None) -> None:
```
mkornacker: The column parameter is too general, you should really only be validating the (optional) array dimensions.

tests/test_index.py
```diff
@@ -984,6 +985,99 @@ def test_embedding_index_precision(
         res = t.select(t.rowid, t.text, sim=sim).order_by(sim, asc=False).collect()
         assert res[0]['rowid'] == 1
 
+    def test_array_column_embedding_index(
+        self, uses_db: None, e5_embed: pxt.Function, reload_tester: ReloadTester
+    ) -> None:
+        skip_test_if_not_installed('transformers')
+
+        texts = ['a dog playing in the park', 'a cat sitting on a mat', 'a bird flying in the sky']
```
mkornacker, line 994: Add blank lines between larger units of logic.

tests/test_index.py
```diff
@@ -984,6 +985,99 @@ def test_embedding_index_precision(
...
+                query = t.select(t.id, t.text, sim=sim).order_by(sim, asc=False).limit(3)
+                res = reload_tester.run_query(query)
+                assert len(res) == 3, col_name
+                assert res[0]['text'] == best, col_name
+                sim_vals = [r['sim'] for r in res]
+                assert all(sim_vals[i] >= sim_vals[i + 1] for i in range(len(sim_vals) - 1)), (
+                    f'{col_name}:{idx_name}: similarity scores must be descending; got {sim_vals}'
+                )
+
+            # search by embedding vector on - should work with all embedding indices
```
mkornacker: Grammar?

pixeltable/exprs/similarity_expr.py
```diff
@@ -23,24 +26,57 @@ class SimilarityExpr(Expr):
...
         # determine index to use
-        idx_info = col_ref.tbl.get().get_idx(col_ref.col, idx_name, EmbeddingIndex)
-        self.idx_id = idx_info.id
-        self.idx_name = idx_info.name
+        if self.idx_name is None:
+            # Look up index by column
+            idx_info = tv.get_idx(column, None, EmbeddingIndex)
+            self.idx_name = idx_info.name
+        else:
+            # Look up index by name
```
mkornacker: add: ", not by id, so that the expr is still valid when the index gets dropped and re-created"

pixeltable/index/embedding_index.py
```diff
@@ -65,9 +65,21 @@ def __init__(
         image_embed: func.Function | None = None,
         audio_embed: func.Function | None = None,
         video_embed: func.Function | None = None,
+        column: catalog.Column | None = None,  # Used for validation when the indexed column is an array.
     ):
-        if embed is None and string_embed is None and image_embed is None:
-            raise excs.Error('At least one of `embed`, `string_embed`, or `image_embed` must be specified')
+        # Skip validation in two cases:
```
mkornacker: Validation of what?

tests/test_index.py
```diff
@@ -984,6 +985,99 @@ def test_embedding_index_precision(
...
+        return np.zeros(256, dtype=np.float32)
+
+    def test_array_embedding_index_validation_errors(self, uses_db: None) -> None:
+        t = pxt.create_table(
+            'arr_val_test',
+            {'id': pxt.Int, 'vec': pxt.Array[(384,), np.float32]},  # type: ignore[misc]
+            if_exists='replace',
+        )
+        t.insert([{'id': 0, 'vec': np.zeros(384, dtype=np.float32)}])
+        with pytest.raises(pxt.Error, match='shape'):
```
mkornacker, line 1094: What about a column vec2: Array[(10, 10), float32]
and add_embedding_index('vec2')

mkornacker (changes requested):

pixeltable/index/embedding_index.py
amithadke: yes weird it actually found where do we even get index sa type

pixeltable/index/embedding_index.py
amithadke: sg, using query_vector in similarity so it will look similarity(query_vector=...)

pixeltable/index/embedding_index.py
amithadke: moved check  for column, earlier now we only check for columns array shape match in here

pixeltable/index/embedding_index.py
aaron-siegel: We aligned on "vector=" in a team product meeting. We can't just be changing that again via an ad hoc decision in a PR comment thread. cc @pierrebrunelle @apreshill

aaron-siegel (changes requested):
See my latest comment re: `similarity()` parameters.

pixeltable/index/embedding_index.py
mkornacker: I didn't suggest using query_vector as the parameter name.

pixeltable/index/embedding_index.py
```diff
@@ -178,11 +210,38 @@ def sa_create_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql
         stmt = Env.get().dbms.create_vector_index_stmt(store_index_name, sa_value_col, metric=metric)
         return stmt
 
+    def drop_index(self, index_name: str, index_value_col: catalog.Column) -> None:
+        """Drop the index on the index value column"""
+        # TODO: implement
+        raise NotImplementedError()
+
+    def _validate_query_vector(self, query_vector: np.ndarray, val_column: catalog.Column) -> None:
```
mkornacker: You can tighten the signature with val_column_type: ts.ArrayType instead of val_column.

pixeltable/index/embedding_index.py
```diff
@@ -178,11 +210,38 @@ def sa_create_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql
...
+                f'similarity(vector=...): query vector length {query_vector.shape[0]} does not match '
+                f'indexed column {val_column.name!r} dimension {expected_len}'
+            )
+
     def similarity_clause(self, val_column: catalog.Column, item: exprs.Literal) -> sql.ColumnElement:
         """Create a ColumnElement that represents '<val_column> <op> <item>'"""
-        assert item.col_type._type in self.embeddings
-        embedding = self.embeddings[item.col_type._type].exec([item.val], {})
-        assert isinstance(embedding, np.ndarray)
+        if item.col_type._type == ts.ColumnType.Type.ARRAY:
```
mkornacker: if item.col_type.is_array_type()

pixeltable/index/embedding_index.py
```diff
@@ -65,9 +65,21 @@ def __init__(
...
+        column: catalog.Column | None = None,  # Used for validation when the indexed column is an array.
     ):
-        if embed is None and string_embed is None and image_embed is None:
-            raise excs.Error('At least one of `embed`, `string_embed`, or `image_embed` must be specified')
+        # Embedding functions are not required in two cases:
+        # 1. column is None: deserialization path
+        # 2. column is an array type: similarity search uses the raw vector directly
+        if (
+            column is not None
+            and column.col_type._type != ts.ColumnType.Type.ARRAY
```
mkornacker: is_array_type()

pixeltable/index/embedding_index.py
```diff
@@ -65,9 +65,21 @@ def __init__(
...
         video_embed: func.Function | None = None,
+        column: catalog.Column | None = None,  # Used for validation when the indexed column is an array.
     ):
-        if embed is None and string_embed is None and image_embed is None:
-            raise excs.Error('At least one of `embed`, `string_embed`, or `image_embed` must be specified')
+        # Embedding functions are not required in two cases:
+        # 1. column is None: deserialization path
+        # 2. column is an array type: similarity search uses the raw vector directly
+        if (
+            column is not None
```
mkornacker, line 72: I don't understand this check. The comment in line 68 says 'column' is only used for indexed array columns. Here you're explicitly checking for the opposite.

pixeltable/index/embedding_index.py
```diff
@@ -99,17 +111,31 @@ def __init__(
...
+        if column is not None and isinstance(column.col_type, ts.ArrayType):
+            array_column_shape = column.col_type.shape
+            if (
+                array_column_shape is None
+                or len(array_column_shape) != 1
+                or array_column_shape[0] is None
+                or array_column_shape[0] <= 0
+            ):
+                raise excs.Error(
+                    f'Column {column.name!r} must be a 1-dimensional array with a specific length '
```
mkornacker: Easier to follow:
"Cannot create embedding index on column ..: requires a 1-dimensional array column type with a defined length."

pixeltable/index/embedding_index.py
```diff
@@ -257,19 +326,26 @@ def _validate_embedding_fn(cls, embed_fn: func.Function) -> None:
...
         if shape[0] <= 0:
             raise excs.Error(
                 f'The function `{embed_fn.name}` is not a valid embedding: '
                 f'it returns an array of invalid length {shape[0]}'
             )
 
+        if array_column_shape is not None and shape != array_column_shape:
+            raise excs.Error(
+                f'The function `{embed_fn.name}` returns an array with shape {shape}, '
+                f'but indexed column has shape {array_column_shape}'
```
mkornacker: "but the ..."

pixeltable/exprs/similarity_expr.py
```diff
@@ -23,24 +26,58 @@ class SimilarityExpr(Expr):
...
+            idx_info = tv.idxs_by_name.get(idx_name)
+            if idx_info is None:
+                raise excs.Error(f'Index {idx_name!r} not found for column {column.name!r}')
+
         idx = idx_info.idx
         assert isinstance(idx, EmbeddingIndex)
 
-        if item.col_type._type not in idx.embeddings:
+        # Skip for array columns; similarity search uses the raw vector directly.
+        if item.col_type._type != ts.ColumnType.Type.ARRAY and item.col_type._type not in idx.embeddings:
```
mkornacker: is_array_type()

pixeltable/exprs/similarity_expr.py
```diff
@@ -50,10 +87,18 @@ def __init__(self, col_ref: ColumnRef, item: Expr, idx_name: str | None = None):
         self.id = self._create_id()
 
     def __repr__(self) -> str:
-        return f'{self.components[0]}.similarity({self.components[1]})'
+        assert self.idx_name is not None
+        assert self.col_id is not None
+
+        tbl_version = get_runtime().catalog.get_tbl_version(self.table_version_key, validate_initialized=True)
+        col = tbl_version.cols_by_id[self.col_id]
+        return f'{col.name}.{self.idx_name}.similarity({self.components[0]})'
```
mkornacker: This isn't valid syntax. The index name is passed as an argument to similarity()

amithadke: switching  to f'{col.name}.similarity({self.components[0]}, {self.idx_name})'
When similarity() has idx = none we select first embedding index so idx_name will be name of index that we found for that column - it wont exactly match with user input to similarity()

pixeltable/index/embedding_index.py
amithadke: add_embedding_index always pass in column when constructing 
 idx = EmbeddingIndex(
                metric=metric,
                precision=precision,
                embed=embedding,
                string_embed=string_embed,
                image_embed=image_embed,
                column=col,  # Pass column for shape validation
            )

that check is for non array column where at least one of the embedding function must be provided 
perhaps i should rephrase comment on line 68 or pass column only when its an array column (but that causes issue during deserialization for array column index column is none which is true for EmbeddingIndex construction with non array column)

pixeltable/exprs/column_ref.py
```diff
@@ -181,6 +182,7 @@ def similarity(
         image: str | PIL.Image.Image | None = None,
         audio: str | None = None,
         video: str | None = None,
+        vector: np.ndarray | None = None,
         idx: str | None = None,
```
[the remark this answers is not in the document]
amithadke: ignoring

pixeltable/metadata/converters/convert_48.py
```diff
@@ -0,0 +1,42 @@
+"""Convert SimilarityExpr from legacy 'indexed_col' (serialized ColumnRef) to 'table_version_key' + 'idx_name'."""
+
```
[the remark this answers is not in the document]
amithadke: ignoring

pixeltable/catalog/table_version.py
```diff
@@ -589,7 +589,7 @@ def _lookup_column(self, qid: QColumnId) -> Column | None:
         if qid.tbl_id == self.id:
             return next(col for col in self.cols if col.id == qid.col_id)
```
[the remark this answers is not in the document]
amithadke: done

pixeltable/exprs/similarity_expr.py
```diff
@@ -62,42 +121,49 @@ def sql_expr(self, _: SqlElementCache) -> sql.ColumnElement | None:
...
-        if self.idx_id not in col_ref.tbl.get().idxs:
-            raise excs.Error(f'Index {self.idx_name!r} not found')
-        idx_info = col_ref.tbl.get().idxs[self.idx_id]
+        tbl_version = get_runtime().catalog.get_tbl_version(self.table_version_key, validate_initialized=True)
+        col = tbl_version.lookup_column(self.col_qid)
+        if col is None:
+            raise excs.Error(
+                f'Embedding index {self.idx_name!r} no longer exists because column {self.col_qid!r} was dropped'
+            )
+        idx_info = tbl_version.get_idx(col, self.idx_name, EmbeddingIndex)
```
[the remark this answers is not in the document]
amithadke: done adding get_visible_column moving lookup_column back to private method

pixeltable/exprs/similarity_expr.py
```diff
@@ -50,10 +88,62 @@ def __init__(self, col_ref: ColumnRef, item: Expr, idx_name: str | None = None):
...
 
     def __repr__(self) -> str:
-        return f'{self.components[0]}.similarity({self.components[1]})'
+        assert self.idx_name is not None
+        assert self.col_qid is not None
+
+        tbl_version = get_runtime().catalog.get_tbl_version(self.table_version_key, validate_initialized=True)
+        col = tbl_version.get_visible_column(self.col_qid)
+        assert col is not None
+        return f'{col.name}.similarity({self.components[0]}, {self.idx_name!r})'
```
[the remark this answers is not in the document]
amithadke: adding
  if col is None:
            return f'<invalid>.similarity({self.components[0]}, {self.idx_name!r})'
            
very narrow case but better than an assert

pixeltable/metadata/converters/convert_48.py
```diff
@@ -0,0 +1,42 @@
...
+    if not isinstance(v, dict):
+        return None
+    if v.get('_classname') != 'SimilarityExpr':
+        return None
+    if 'table_version_key' in v:  # already migrated
+        return None
+    assert 'components' in v
+    components = v['components']
+    assert len(components) == 2
+    col_ref_dict = components[0]
```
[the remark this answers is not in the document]
amithadke: ignoring

pixeltable/catalog/table.py
```diff
@@ -959,11 +959,15 @@ def add_embedding_index(
...
-        Only `String` and `Image` columns are currently supported.
+        To add an embedding index, specify the column to be indexed and, if the column is not an `Array` column, an
+        embedding UDF. `String`, `Image`, and `Array` columns are currently supported.
+
+        For `Array` columns, which are assumed to contain precomputed embeddings, an embedding function is optional;
+        if provided, it will be used to convert query values into embeddings for similarity search.
 
         Args:
-            column: The name of, or reference to, the column to be indexed; must be a `String` or `Image` column.
+            column: The name of, or reference to, the column to be indexed; must be a `String`, `Image` or
```
mkornacker, line 969: This is outdated. We now have embedding udfs for audio and video as well.

pixeltable/exprs/similarity_expr.py
```diff
@@ -27,57 +30,55 @@ class SimilarityExpr(Expr):
     """
 
     table_version_key: TableVersionKey
-    idx_name: str | None = None  # index name could be optional
-    col_id: int | None = None  # used to find the first embedding index when index name is not provided
+    idx_name: str | None = None  # index name; None if not specified by the user
+    col_qid: QColumnId | None = None  # identifies the indexed column
```
mkornacker: The 'q' stands for qualified, so maybe qcol_id?

PR #1102, opened 2026-01-29
PXT-985 Adding custom_metadata and restricting comment field to string

christopherpestano:
PR adding user_metadata: a custom user defined field in table metadata that allows for arbitrary JSON objects to be directly placed into table metadata. Additionally, this PR adds validation that ensures non-string inputs are rejected in the comment field.

As user_metadata is a non-required field, there is no need to bump the metadata version as the converter would be a no-op, however we should incorporate this field in future dumps when the version is bumped to ensure coverage.

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -2801,7 +2801,7 @@ class TestTable:
             External Store         Type
                    project  MockProject
 
-            COMMENT: This is an intriguing table comment."""
+            Comment: This is an intriguing table comment."""
         )
         _ = v2._repr_html_()  # TODO: Is there a good way to test this output?
 
@@ -2827,7 +2827,7 @@ class TestTable:
             External Store         Type
                    project  MockProject
 
-            COMMENT: This is an intriguing table comment."""
+            Comment: This is an intriguing table comment."""
         )
 
         # test case: snapshot of base table


pixeltable/catalog/table_metadata.py
```diff
@@ -72,6 +72,8 @@ class TableMetadata(TypedDict):
     """The current schema version of the table."""
     comment: str | None
     """User-provided table comment, if one exists."""
+    user_metadata: Any | None
```
aaron-siegel: The `| None` is superfluous (`Any | None` is just `Any`)

pixeltable/metadata/schema.py
```diff
@@ -351,6 +351,7 @@ class SchemaVersionMd:
     # stores column.MediaValiation.name.lower()
     media_validation: str
     additional_md: dict[str, Any]  # deprecated
+    user_metadata: Any | None = None
```
aaron-siegel: same comment

pixeltable/globals.py
```diff
@@ -197,12 +199,15 @@ def create_table(
             'Unable to create a proper schema from supplied `source`. Please use appropriate `schema_overrides`.'
         )
 
+    assert isinstance(comment, str), 'Comment must be a string'
```
aaron-siegel: Users should never see an `AssertionError` unless there's a bug (we use `assert` purely to state-check internal logic). Best to raise an `excs.Error` here.

pixeltable/globals.py
```diff
@@ -333,6 +339,8 @@ def create_view(
                     f'{tbl_version_path.get_column(col_name).get_tbl().name}.'
                 )
 
+    assert isinstance(comment, str), 'Comment must be a string'
```
aaron-siegel: Same comment as above. Also, add this check to create_snapshot

tests/test_table.py
```diff
@@ -3007,3 +3011,20 @@ def test_drop_last_column(self, uses_db: None, reload_tester: ReloadTester) -> N
...
+        with pytest.raises(AssertionError, match='Comment must be a string'):
+            t = pxt.create_table('tbl_invalid', {'c': pxt.Int}, comment={'comment': 'This is a test table.'})  # type: ignore[arg-type]
+
+    def test_table_user_metadata(self, uses_db: None, reload_tester: ReloadTester) -> None:
+        user_metadata = {'key1': 'value1', 'key2': 2, 'key3': [1, 2, 3]}
+        t = pxt.create_table('tbl', {'c': pxt.Int}, user_metadata=user_metadata)
+        assert t.get_metadata()['user_metadata'] == user_metadata
+
+        # check that invalid JSON user metadata are rejected
+        with pytest.raises(TypeError):
```
aaron-siegel: My preference would be to trap that error and re-raise it as an `excs.Error`.

tests/test_table.py
```diff
@@ -3007,3 +3011,20 @@ def test_drop_last_column(self, uses_db: None, reload_tester: ReloadTester) -> N
             pxt.Error, match="Cannot drop column 'c2' because it is the last remaining column in this table"
         ):
             t.drop_column('c2')
+
+    def test_table_comment(self, uses_db: None, reload_tester: ReloadTester) -> None:
+        t = pxt.create_table('tbl', {'c': pxt.Int}, comment='This is a test table.')
```
aaron-siegel, line 3016: Also run all these tests on `create_view()` and `create_snapshot()`.

aaron-siegel (commented):
I'm aware it's a draft but I took a quick look and had a few comments. If any of them are things you're already planning to do (or are obviated by something else you're planning to do), just ignore them.

I think `Any` is probably the correct type for `user_metadata`, unfortunately. We just need to clearly document that it's expected to be valid JSON.

tests/test_table.py
christopherpestano: Added similar tests to test_snapshot.py and test_view.py.

tests/test_table.py
christopherpestano: Modified to raise excs.Error, as well as added tests to table/view/snapshot to ensure that all three are raising correct error.

pixeltable/globals.py
christopherpestano: Modified to be excs.Error

pixeltable/globals.py
christopherpestano: Modified to be excs.Error, question though: this code path is also used by create_snapshot (validated with test_snapshot.py), should there be a double validation since create_snapshot calls into create_view?

pixeltable/metadata/schema.py
christopherpestano: Changed type hint to be just Any

pixeltable/catalog/table_metadata.py
christopherpestano: Changed type hints throughout PR to just be Any.

pixeltable/metadata/schema.py
```diff
@@ -351,6 +351,7 @@ class SchemaVersionMd:
     # stores column.MediaValiation.name.lower()
     media_validation: str
     additional_md: dict[str, Any]  # deprecated
+    user_metadata: Any = None
```
mkornacker: This could subsume 'comment', there is no good reason to keep it.

Existing non-empty comments can be converted to {'comment': ...}.

aaron-siegel: We discussed this exact question at length last week, and reached a consensus to keep them separate (@pierrebrunelle noted that users will expect a separate `comment` field that's a simple string).

pixeltable/globals.py
christopherpestano: Resolved during daily standup.

pixeltable/catalog/table_version_path.py
```diff
@@ -152,6 +153,10 @@ def comment(self) -> str:
         self.refresh_cached_md()
         return self._cached_tbl_version.comment
 
+    def custom_metadata(self) -> Any:
+        self.refresh_cached_md()
+        return self._cached_tbl_version.custom_metadata
```
aaron-siegel: I wonder if we should be returning a deepcopy() of the metadata here, in case clients inadvertently try to modify it.

pixeltable/globals.py
```diff
@@ -197,12 +200,21 @@ def create_table(
             'Unable to create a proper schema from supplied `source`. Please use appropriate `schema_overrides`.'
         )
 
+    if not isinstance(comment, str):
+        raise excs.Error('Comment must be a string')
```
aaron-siegel:
```
`comment` must be a string
```
(for consistency with other error message formatting)

pixeltable/globals.py
```diff
@@ -93,6 +95,7 @@ def create_table(
             table.
         num_retained_versions: Number of versions of the table to retain.
         comment: An optional comment; its meaning is user-defined.
+        custom_metadata: Optional user-defined JSON metadata to associate with the table.
```
aaron-siegel: "JSON metadata" maybe isn't the clearest way to describe it? It's a Python object. Maybe something like "Optional user-defined metadata to associate with the table. Must be a valid JSON-serializable object (str, int, float, bool, dict, or list)?"

tests/test_table.py
```diff
@@ -3007,3 +3011,73 @@ def test_drop_last_column(self, uses_db: None, reload_tester: ReloadTester) -> N
             pxt.Error, match="Cannot drop column 'c2' because it is the last remaining column in this table"
         ):
             t.drop_column('c2')
+
+    def test_table_comment(self, uses_db: None, reload_tester: ReloadTester) -> None:
```
aaron-siegel: In all these tests you're pulling in a `ReloadTester` artifact, but not actually using it. I'm not sure it serves any purpose here since it only checks table data, not metadata; but I do think reload tests would be useful. Maybe get rid of the `ReloadTester` and manually add checks that comments and metadata survive catalog reloads.

I'd suggest we also include comments and custom metadata in `create_test_db_dump.py` and add checks for them in `test_migration.py`.

tests/test_table.py
```diff
@@ -3007,3 +3011,73 @@ def test_drop_last_column(self, uses_db: None, reload_tester: ReloadTester) -> N
...
+        with pytest.raises(pxt.Error, match='Comment must be a string'):
+            pxt.create_table('tbl_invalid', {'c': pxt.Int}, comment={'comment': 'This is a test table.'})  # type: ignore[arg-type]
+
+    def test_table_custom_metadata(self, uses_db: None, reload_tester: ReloadTester) -> None:
+        custom_metadata = {'key1': 'value1', 'key2': 2, 'key3': [1, 2, 3]}
+        t = pxt.create_table('tbl', {'c': pxt.Int}, custom_metadata=custom_metadata)
+        assert t.get_metadata()['custom_metadata'] == custom_metadata
+
+        # check that invalid JSON user metadata are rejected
+        with pytest.raises(pxt.Error):
```
aaron-siegel: Use `match=` to capture the error message text (always).

pixeltable/catalog/table.py
```diff
@@ -1898,3 +1904,44 @@ def __check_mutable(self, op_descr: str) -> None:
...
+    def _format_custom_metadata(
+        self, metadata: Any, max_elements: int = 5, max_character_limit: int = 100, indent: int = 4, interpose: int = 10
+    ) -> str:
+        indent = ' ' * indent
+        if isinstance(metadata, dict):
+            items = list(metadata.items())[:max_elements]
+            parts = []
+            for key, value in items:
+                if isinstance(value, str):
+                    value_str = f'"{value}"'
```
aaron-siegel: You can fold this into the next case (`json.dumps(value)` will give the expected output if `value` is a string)

pixeltable/catalog/table.py
```diff
@@ -1898,3 +1904,44 @@ def __check_mutable(self, op_descr: str) -> None:
...
+                elif isinstance(value, (dict, list)):
+                    value_str = json.dumps(value)
+                else:
+                    value_str = str(value)
```
aaron-siegel: Why not just use `json.dumps(value)` for everything?

pixeltable/catalog/table.py
```diff
@@ -1898,3 +1904,44 @@ def __check_mutable(self, op_descr: str) -> None:
...
+
+                if len(value_str) > max_character_limit:
+                    value_str = value_str[: max_character_limit - interpose] + ' ... ' + value_str[-interpose:]
```
aaron-siegel: I suggest
`   ......   `
for the bridge rather than
`  ...  `
the reason being that the latter might reasonably occur in user data, but the former is much less likely to, so it will stand out better.

pixeltable/catalog/table.py
```diff
@@ -1898,3 +1904,44 @@ def __check_mutable(self, op_descr: str) -> None:
...
+
+                parts.append(f'{indent}"{key}": {value_str}')
+            result = '{\n' + ',\n'.join(parts)
+            if len(metadata) > max_elements:
+                result += f',\n{indent}... ({len(metadata) - max_elements} more)'
+            result += '\n}'
+
+            return result
+        elif isinstance(metadata, list):
```
aaron-siegel: The remaining cases should fold into one.

pixeltable/metadata/schema.py
```diff
@@ -351,6 +351,7 @@ class SchemaVersionMd:
     # stores column.MediaValiation.name.lower()
     media_validation: str
     additional_md: dict[str, Any]  # deprecated
+    custom_metadata: Any = None
```
mkornacker, line 355: Mention additional constraints (eg, 'json-serializable') in a comment.

pixeltable/catalog/table.py
```diff
@@ -1898,3 +1904,44 @@ def __check_mutable(self, op_descr: str) -> None:
             raise excs.Error(f'{self._display_str()}: Cannot {op_descr} a replica.')
         if self._tbl_version_path.is_snapshot():
             raise excs.Error(f'{self._display_str()}: Cannot {op_descr} a snapshot.')
+
+    def _format_custom_metadata(
```
mkornacker: We already have Formatter::format_json(). Fold it into that.

pixeltable/catalog/table.py
```diff
@@ -163,6 +163,7 @@ def _get_metadata(self) -> TableMetadata:
             version_created=datetime.datetime.fromtimestamp(tv.created_at, tz=datetime.timezone.utc),
             schema_version=tvp.schema_version(),
             comment=self._get_comment(),
+            custom_metadata=self._get_custom_metadata(),
```
mkornacker, line 167: Let's return a copy here.

pixeltable/catalog/table.py
aaron-siegel: @mkornacker We talked about this at length yesterday. The differences in formatting logic are substantial enough that I believe this merits a separate implementation.

pixeltable/catalog/table.py
aaron-siegel: deepcopy :)

pixeltable/catalog/table.py
christopherpestano: Good point, when I was writing this section I overcomplicated it because I wasn't certain if certain types would need special formatting.

pixeltable/catalog/table.py
christopherpestano: Yeah, changed the entire thing to just be json.dumps(value)

pixeltable/catalog/table.py
christopherpestano: Yep just like the other part, modified to just be json.dumps(metadata)

pixeltable/catalog/table.py
mkornacker: Let's still move it into Formatter and call it something else.

pixeltable/catalog/table_version_path.py
christopherpestano: Yeah I agree the client shouldn't be able to mutate through a getter.

pixeltable/catalog/table.py
christopherpestano: Modified so that custom_metadata returns deepcopy

pixeltable/catalog/table.py
christopherpestano: Visually something about that looks off to me, I think personally that maybe [...] would be a better bridge as that's what python standard library uses in things like textwrap.

pixeltable/metadata/schema.py
christopherpestano: Discussed and resolved.

tests/test_table.py
christopherpestano: Makes sense, added match with appropriate error message.

pixeltable/globals.py
christopherpestano: Sure, modified to be appropriate error.

pixeltable/globals.py
christopherpestano: Sure I think that's clearer, I'll add something similar as well to schema.py

pixeltable/metadata/schema.py
christopherpestano: Added following comment to clarify that object must be valid json
`user-defined metadata - must be a valid JSON-serializable object`

pixeltable/catalog/table.py
christopherpestano: I think the idea of moving it into the formatter is good, as it makes the function reusable should the case arise that we want similar formatting logic elsewhere for other json fields. Added to Formatter as summarize_json, let me know if you would prefer a different name, not too set on the current one.

aaron-siegel: Agreed.

pixeltable/catalog/table.py
aaron-siegel: ok

tests/test_view.py
```diff
@@ -1264,3 +1267,24 @@ def test_circular_view_def(self, uses_db: None) -> None:
         pxt.drop_table('my_view_2')
         pxt.drop_table('my_view_1')
         pxt.drop_table('my_tbl')
+
+    def test_view_comment(self, uses_db: None, reload_tester: ReloadTester) -> None:
```
aaron-siegel: I don't think these reload_testers are actually doing anything?

pixeltable/catalog/table.py
```diff
@@ -379,6 +384,8 @@ def _descriptors(self) -> DescriptionHelper:
                 helper.append(stores)
             if self._get_comment():
                 helper.append(f'COMMENT: {self._get_comment()}')
+            if self._get_custom_metadata():
+                helper.append(f'CUSTOM METADATA: {Formatter.summarize_json(self._get_custom_metadata())}')
```
aaron-siegel: Last comment - I find the ALL CAPS distracting here - I realize that's how "comment" was before, but maybe that was a mistake, and with multiple properties now being shown in all caps, I find it visually distracting. Maybe we should just write "Comment:" and "Custom Metadata:"?

aaron-siegel (approved):

tests/test_table.py
christopherpestano: Added some reload coverage using @pytest.mark.parametrize for reload_catalog and removed the reload_tester for the 3 primary ut files. Also moved the summarize_json ut to the appropriate test file for all the formatter utils. Also added those reload tests using comment and user_metadata. Let me know if there's still any coverage missing.

tests/test_view.py
christopherpestano: Replaced with reload_catalog()

pixeltable/catalog/table.py
christopherpestano: Agreed, implemented suggestion.

PR #1104, opened 2026-01-29
PXT-965 Support http url and blob store uri for creating  json/parquet/csv tables

amithadke:
Also adds support for jsonl files (uses jsonlines)


pyproject.toml
```diff
@@ -47,6 +47,7 @@ dependencies = [
     "toml>=0.10",
     "jsonschema>=4.1.0",
     "jmespath>=1.0.1",
+    "jsonlines>=4.0.0",
```
aaron-siegel: I'm not convinced it's worth introducing a new required dependency for this. `jsonl` should be easy enough to parse without the extra dependency, right?

tests/io/test_import.py
```diff
@@ -96,6 +98,57 @@ def test_import_json(self, uses_db: None) -> None:
...
+        assert tab._get_schema() == {
+            'name': ts.StringType(nullable=True),
+            'human': ts.BoolType(nullable=True),
+            'parents': ts.JsonType(nullable=True),
+            'age': ts.FloatType(nullable=True),
+            'metadata': ts.JsonType(nullable=True),
+            'children': ts.IntType(nullable=True),
+        }
+
+    def test_import_json_from_s3_uri(self, uses_db: None) -> None:
```
aaron-siegel: These tests are almost exact duplicates. Why not just have one test that inserts 2 rows? (Same for the others with separate http/s3 tests)

pixeltable/utils/__init__.py
```diff
@@ -55,3 +57,15 @@ def parse_local_file_path(file_or_url: str) -> Path | None:
         return Path(urllib.parse.unquote(urllib.request.url2pathname(parsed.path)))
     else:
         return None
+
+
+def resolve_table_source_to_path(source: str) -> Path:
```
aaron-siegel: This shouldn't be necessary - just do `fetch_url(source, allow_local_file=True)`

pixeltable/io/table_data_conduit.py
```diff
@@ -686,7 +702,9 @@ def specialize(self) -> TableDataConduit:
             return CSVTableDataConduit.from_tds(self)
         if self.source_format == 'excel' or (isinstance(self.source, str) and '.xls' in self.source.lower()):
             return ExcelTableDataConduit.from_tds(self)
-        if self.source_format == 'json' or (isinstance(self.source, str) and '.json' in self.source.lower()):
+        if self.source_format == 'json' or (
+            isinstance(self.source, str) and ('.json' in self.source.lower() or '.jsonl' in self.source.lower())
```
aaron-siegel: Doesn't `'.jsonl' in self.source.lower()` necessarily imply `'.json' in self.source.lower()`?

tests/io/test_import.py
amithadke: s3 tests can be skipped http are referring to file in main on our github.
i will add params

pixeltable/io/table_data_conduit.py
amithadke: yes I'm thinking if we should be doing checks with endswith over in string  .json/.jsonl and other places.

tests/utils.py
```diff
@@ -597,6 +597,19 @@ def skip_test_if_no_aws_credentials() -> None:
         pytest.skip(str(exc))
 
 
+_S3_PYTEST_RESOURCES = 's3://pxt-test/pytest-resources'
+
+
+def ensure_s3_pytest_resources_access() -> None:
+    """Skip if s3://pxt-test/pytest-resources is not reachable (no creds or no access)."""
+    if not Env.get().is_installed_package('boto3'):
```
aaron-siegel: Use the standard utility please (utils.skip_test_if_not_installed)

PR #1105, opened 2026-01-30
PXT-887 get_table with if_not_exists='ignore' should not raise an error for missing parent dir

amithadke:


amithadke:
you are right I thought we had missing case there. 

Jira earlier had if_exists=ignore like this
 pxt.get_table('my_dir.my_tbl') throws an exception if `my_dir` does not exist, even if `if_exists='ignore'`
 
 but get_table doesnt have 'if_exists' param it has 
 def get_table(path: str, if_not_exists: Literal['error', 'ignore'] = 'error') -> catalog.Table | None:

I thought it was a  typo in bug description.

PR #1106, opened 2026-01-30
fix: Make notebook outputs visible in dark mode

pierrebrunelle:
update


pierrebrunelle:
Closing in favor of a clean PR with only the CSS fix

PR #1111, opened 2026-02-01
@pxt.iterator decorator

aaron-siegel:
Refactors the iterator abstraction to use a new `@pxt.iterator` decorator for declaring iterators.
- Iterators can be defined via `def my_iterator(...) -> Iterator[MyRow]` or via `class my_iterator(PxtIterator[MyRow])`, where `MyRow` is a `TypedDict` class
- Iterators can be configured with validators and/or conditional_output_schemas (if a conditional_output_schema is defined, then a simple `dict` can be used as the return type, in place of `MyRow`)
- Iterator schemas are persisted to the catalog now, and allow for column renames and subselection; currently this is used purely for migrations/backward-compatibility and flexible evolution of iterator schemas; in the future we can allow for explicit (user-specified) renames and subselection
- `audio_splitter` inputs and outputs updated to align with `video_splitter`, with appropriate migration instance for backward-compatibility
- all builtin iterators have been migrated to the new pattern, with appropriate migration instances
- "graceful evolution" of iterators is supported the same way as for UDFs (backward-incompatible changes to an iterator will cause views to be marked as invalid, but they'll still be queryable except for unstored columns) 
- custom iterators that use the old pattern are retrofitted into the new framework
- mypy plugin updated to recognize type semantics of `PxtIterator`
- doctools updated for new structure, along with docstrings for all builtin iterators
- tutorial notebook for user-defined iterators

### DIFF
--- a/pixeltable/iterators/base.py
+++ b/pixeltable/iterators/base.py
@@ -56,7 +56,7 @@ class ComponentIterator(ABC):
     @deprecated(
         'The `ComponentIterator` class is deprecated; please use the `@pxt.iterator` decorator instead.\n'
         f'For details see: {ITERATOR_GUIDE_URL}',
-        version='0.5.18',
+        version='0.5.19',
         category=excs.PixeltableDeprecationWarning,
     )
     def create(cls, **kwargs: Any) -> GeneratingFunctionCall:


pixeltable/func/iterator.py
```diff
@@ -0,0 +1,493 @@
...
+    def from_dict(cls, d: dict[str, Any]) -> 'IteratorOutput':
+        return cls(orig_name=d['orig_name'], is_stored=d['is_stored'], col_type=ts.ColumnType.from_dict(d['col_type']))
+
+
+class PxtIterator(abc.ABC, Iterator[T], Generic[T]):
+    def __iter__(self) -> Self:
+        return self
+
+    @abc.abstractmethod
+    def __next__(self) -> T: ...
```
mkornacker, line 51: Can we drop this and simply have __iter__() -> Iterator[T]?

pixeltable/functions/audio.py
```diff
@@ -127,39 +133,206 @@ def encode_audio(
...
+            current_pos = segment_end - overlap
+        # If the last segment is smaller than min_segment_duration then drop the last segment from the list
+        if (
+            len(segments_to_extract_in) > 0
+            and (segments_to_extract_in[-1][1] - segments_to_extract_in[-1][0]) < min_segment_duration
+        ):
+            return segments_to_extract_in[:-1]  # return all but the last segment
+        return segments_to_extract_in
+
+    def __next__(self) -> AudioSegment:
```
mkornacker, line 259: Would be nice simply to have __iter__() -> Iterator[AudioSegment], while you're at it.

pixeltable/functions/video.py
```diff
@@ -1438,43 +1443,55 @@ def process_cuts(cuts: list[FrameTimecode]) -> None:
...
+    time: float
+    is_corrupt: bool
+    key_frame: bool
+    pict_type: int
+    interlaced_frame: bool
+
+
+class Frame(TypedDict):
+    frame: pxt.Image
+    frame_attrs: FrameAttrs
```
mkornacker, line 1560: That's problematic for anyone running queries with predicates on the attrs.

The way those are handled today would imply a very steep performance drop if these aren't actual columns.

pixeltable/func/iterator.py
aaron-siegel: That doesn’t work with seek().

pixeltable/functions/audio.py
aaron-siegel: In general that doesn't work for seek(). The contract needs to be that PxtIterator implements the iterator protocol, which means it needs a __next__().

Since AudioSegment doesn't need a seek(), it can (and perhaps should) be reimplemented as a generator function with a for/yield pattern. But my goal with this PR was to port the existing iterators as-is to the new interfaces, and avoid rewriting their design patterns as much as possible; it's already a fairly heavy PR. We could add a follow-up item to restructure AudioSplitter.

pixeltable/functions/video.py
aaron-siegel: Let's discuss this. Using `FrameAttrs` has one big advantage, which is that if we add new attributes to the dict in the future, the column structure won't change, so existing implementations will reap the benefits of that change without needing a heavyweight schema change.

It seems like the real problem is that JSON queries in SQL aren't performant; let's discuss if there's a way to fix that?

tests/test_component_view.py
```diff
@@ -96,18 +69,13 @@ def test_basic(self, uses_db: None) -> None:
         video_t = pxt.create_table('video_tbl', schema)
         video_filepaths = get_test_video_files()
 
-        # cannot add 'pos' column
```
mkornacker, line 99: What happened?

pixeltable/func/iterator.py
```diff
@@ -0,0 +1,493 @@
...
+    @classmethod
+    def validate(cls, bound_args: dict[str, Any]) -> None:
+        pass
+
+    @classmethod
+    def conditional_output_schema(cls, bound_args: dict[str, Any]) -> dict[str, type] | None:
+        return None
+
+
+class GeneratingFunction:
```
mkornacker, line 68: Why not simply a subclass of Function?

Why not just IteratorFunction?

pixeltable/func/iterator.py
```diff
@@ -0,0 +1,493 @@
...
+    """
+    A function that evaluates to iterators over its inputs.
+
+    It is the Pixeltable equivalent of a table generating function in SQL.
+
+    It is the "lift" of a PxtIterator: a PxtIterator represents a one-to-many expansion of its inputs; the
+    corresponding GeneratingFunction represents a one-to-many expansion of *columns* of inputs.
```
mkornacker, line 75: I don't understand that comment.

pixeltable/func/iterator.py
```diff
@@ -0,0 +1,493 @@
...
+
+            iter_fn = self.decorated_callable.__next__
+            return_type = typing.get_type_hints(iter_fn).get('return')
+
+            # remove type args from return_type (e.g., convert `dict[str, Any]` to `dict`)
+            element_type = typing.get_origin(return_type) or return_type
+            if not isinstance(element_type, type) or not issubclass(element_type, dict):
+                raise excs.Error(
+                    f'`__next__()` method of @pxt.iterator-decorated class `{self.fqn}` '
+                    'must have return type `dict` or `MyTypedDict`.'
```
mkornacker: MyTypedDict?

pixeltable/func/iterator.py
```diff
@@ -0,0 +1,493 @@
...
+        self.signature = Signature.create(decorated_callable, return_type=ts.JsonType())
+
+        self._infer_properties()
+
+        if len(self.unstored_cols) > 0 and not self.has_seek:
+            raise excs.Error(f'Iterator `{self.fqn}` with `unstored_cols` must implement a `seek()` method.')
+
+        self.is_legacy_retrofit = False
+
+    def _infer_properties(self) -> None:
```
mkornacker, line 109: This is the core part of initialization, no?

pixeltable/func/iterator.py
```diff
@@ -0,0 +1,493 @@
...
+
+    def _validate(self, bound_args: dict[str, Any]) -> None:
+        raise excs.Error(f'The iterator `{self.fqn}` cannot be used, because\n{self.error_msg}')
+
+    def as_dict(self) -> dict[str, Any]:
+        return self.fn_dict
+
+
+@dataclass(frozen=True)
+class GeneratingFunctionCall:
```
mkornacker, line 347: Why wouldn't that be an expr?

This probably warrants more discussion (eg, return type, etc.).

mkornacker (changes requested):

tests/test_component_view.py
aaron-siegel: `pos` is no longer a reserved column name (if there's a column `pos`, Pixeltable will call the new one `pos_1`, etc., following the usual convention).

pixeltable/func/iterator.py
aaron-siegel: It doesn't follow the Function interface (specifically, the return type of __call__ does not conform to the Function interface).

I'm not super opinionated about the name; I chose GeneratingFunction because that's what SQL uses.

pixeltable/func/iterator.py
aaron-siegel: This isn't a useful comment, but feel free to reopen with something more actionable.

pixeltable/func/iterator.py
aaron-siegel: Rephrased it.

pixeltable/func/iterator.py
aaron-siegel: Yes

pixeltable/func/iterator.py
aaron-siegel: It doesn't conform to the Expr interface (it returns iterators, not values, and in particular isn't composable the way other expressions are).

Happy to discuss in more detail but in general I would advocate strongly for clean class hierarchies with clear semantics.

pixeltable/func/iterator.py
mkornacker: The __call__() should return an Expr subclass.

pixeltable/func/iterator.py
mkornacker: It needs to be an Expr subclass so that we work with it in the planner/row builder.

That doesn't mean that it implements an eval(). The return type should be list of dict.

aaron-siegel: There is nothing in the planner/row builder today that needs it to be an Expr class - since everything works today (and the old iterator pattern didn't involve Exprs). Are you thinking of a future in which dynamic iterator queries are possible?

mkornacker: I don't understand why we want to check it in in this state. I think this will require non-mechanical changes to conform to the Expr/Function system. I don't want to spend the time to review an intermediate state.

aaron-siegel: It's not in an "intermediate state". It's fully functional, solves the problems outlined in the design doc and in a way that's consistent with the stated design goals, and introduces no regressions (ComponentIterator didn't conform to the Function/Expr hierarchy either; this PR simply swaps it out for the new decorator-based PxtIterator, without any real restructuring of the execution logic).

Like I said I'm happy to discuss whether/how it fits into Function/Expr, but I think that's a nontrivial conversation (we'd be bending the existing abstractions somewhat), and best to do as a follow-up.

PR #1119, opened 2026-02-03
[PXT-1007 + PXT-1010] Modifying add_columns to support column metadata and introducing standard ColumnSpec

christopherpestano:
Currently there is an issue where users cannot add new columns with column metadata, only during table creation is column metadata accepted. To address this issue, I have modified add_columns to accept column schemas aligning the add_columns codepath with the logic currently present in create_table for columns.


pixeltable/catalog/table.py
```diff
@@ -585,15 +584,19 @@ def add_column(
...
+            Error: If the column name is invalid, or already exists and `if_exists='error'`,
                 or `if_exists='replace*'` but the column has dependents or is a basetable column.
 
         Examples:
             Add an int column:
 
             >>> tbl.add_column(new_col=pxt.Int)
 
-            Alternatively, this can also be expressed as:
+            Add a column with column metadata using a dict:
```
sergey-mkhitaryan, line 614: nice. please also add or update the existing example on `add_columns()`

tests/test_table.py
```diff
@@ -3006,3 +3006,59 @@ def test_drop_last_column(self, uses_db: None, reload_tester: ReloadTester) -> N
...
+
+        # valid metadata parameters are accepted
+        t.add_columns({'c3': {'type': pxt.Image, 'stored': True, 'media_validation': 'on_write'}})
+
+        # make sure this metadata is persisted
+        reload_catalog()
+
+        t = pxt.get_table('tbl')
+        assert 'c3' in t.columns()
+        assert t.c3.col.stored == True
```
sergey-mkhitaryan: just `assert t.c3.col.stored`

tests/test_table.py
```diff
@@ -3006,3 +3006,59 @@ def test_drop_last_column(self, uses_db: None, reload_tester: ReloadTester) -> N
...
+        assert str(t.c3.col._media_validation) == 'MediaValidation.ON_WRITE'
```
sergey-mkhitaryan: there should be `col.media_validation` that returns a `MediaValidation`

pixeltable/catalog/table.py
```diff
@@ -573,7 +572,7 @@ def add_column(
         Adds an ordinary (non-computed) column to the table.
 
         Args:
-            kwargs: Exactly one keyword argument of the form `col_name=col_type`.
+            kwargs: Exactly one keyword argument of the form `col_name=col_type` or `col_name=col_spec_dict`.
```
mkornacker: Expand with a description of what can be in that dict.

pixeltable/catalog/table.py
mkornacker: This is not correct (not this PR's fault, but you might as well fix it).

tests/test_table.py
```diff
@@ -3006,3 +3006,59 @@ def test_drop_last_column(self, uses_db: None, reload_tester: ReloadTester) -> N
...
+        with pytest.raises(pxt.Error, match=r"media_validation must be one of: \['on_read', 'on_write']"):
+            t.add_columns({'non_existing_col1': {'type': pxt.Image, 'media_validation': 'on_error'}})
+        with pytest.raises(pxt.Error, match="'stored' must be a bool; got <class 'float'>"):
+            t.add_columns({'non_existing_col1': {'type': pxt.Image, 'stored': float}})
+
+        # valid metadata parameters are accepted
+        t.add_columns({'c3': {'type': pxt.Image, 'stored': True, 'media_validation': 'on_write'}})
+
+        # make sure this metadata is persisted
+        reload_catalog()
```
mkornacker: use pytest.parametrize() with reload_catalog(bool)

tests/test_table.py
mkornacker: Don't use ColumnRef internals, use the public API instead.

Table.get_metadata().

mkornacker (changes requested):

pixeltable/catalog/table.py
christopherpestano: Added additional example in add_columns()

tests/test_table.py
christopherpestano: Followed coding style from test_array_type.py and added to both test_add_columns_with_metadata and test_add_column_with_metadata. Currently working as expected.

pixeltable/catalog/table.py
aaron-siegel: Expand type hint to include the new pattern.

pixeltable/catalog/table.py
```diff
@@ -573,7 +581,12 @@ def add_column(
         Adds an ordinary (non-computed) column to the table.
 
         Args:
-            kwargs: Exactly one keyword argument of the form `col_name=col_type`.
+            kwargs: Exactly one keyword argument of the form `col_name=col_type` or `col_name=col_spec_dict`,
+                where `col_spec_dict` is a dict with keys:
```
aaron-siegel: Really it should be a `TypedDict`.

pixeltable/catalog/table.py
aaron-siegel: Updated type hint

Also, I think the `exprs.Expr` here is wrong? (legacy issue that we might as well fix)

tests/test_table.py
```diff
@@ -3006,3 +3007,61 @@ def test_drop_last_column(self, uses_db: None, reload_tester: ReloadTester) -> N
             pxt.Error, match="Cannot drop column 'c2' because it is the last remaining column in this table"
         ):
             t.drop_column('c2')
+
+    @pytest.mark.parametrize('do_reload_catalog', [False, True], ids=['no_reload_catalog', 'reload_catalog'])
```
aaron-siegel, line 3021: I don't see a need for the `@parametrize`: just do the checks once, then reload, then do the checks again; there's no need to run through the table creation logic a second time (unless I'm missing something)

tests/test_table.py
christopherpestano: Modified to use get_metadata

christopherpestano: Implemented suggestion but with table.get_metadata() instead of t.c3.col.stored.

tests/test_table.py
christopherpestano: Ended up ultimately using table.get_metadata() which gives a string definition of the media_validation.

tests/test_table.py
christopherpestano: I think that in this case there is benefit to doing the `@parameterize` here, as it provides additional coverage both in the case of just directly retrieving the table from catalog as well as retrieving it when it's been persisted. Maybe the additional just getting directly from catalog isn't likely to break, but there's no downside to having the additional coverage.

aaron-siegel: i was saying you could get the same coverage without the parametrize, but i'm ok either way

pixeltable/catalog/table.py
```diff
@@ -491,7 +492,7 @@ def _ignore_or_drop_existing_columns(self, new_col_names: list[str], if_exists:
 
     def add_columns(
         self,
-        schema: dict[str, ts.ColumnType | builtins.type | _GenericAlias],
+        schema: dict[str, ColumnSpec],
```
aaron-siegel: This isn't correct in this case; `ColumnSpec` includes `expr.Expr` which isn't an allowed type here (it's only allowed for computed columns).

I'm also thinking we should remove `ts.ColumnType`; it's no longer part of the public API, and was there for backward compatibility, but I doubt  anyone's using it anymore.

So my preference would be to call the TypedDict "ColumnSpec" and then put the type here explicitly as
```
schema: dict[str, type | ColumnSpec]
```
Giving a name to the compound type makes things a little less transparent for the user, and it's not clear it actually solves a problem in this case.

pixeltable/catalog/table.py
```diff
@@ -501,7 +502,12 @@ def add_columns(
         The format of the `schema` argument is a dict mapping column names to their types.
 
         Args:
-            schema: A dictionary mapping column names to types.
+            schema: A dictionary mapping column names to a `col_type` or a `col_spec_dict`
```
aaron-siegel: The docstring needs to be updated again.

pixeltable/catalog/table.py
```diff
@@ -567,13 +581,18 @@ def add_column(
         self,
         *,
         if_exists: Literal['error', 'ignore', 'replace', 'replace_force'] = 'error',
-        **kwargs: ts.ColumnType | builtins.type | _GenericAlias | exprs.Expr,
+        **kwargs: ColumnSpec,
```
aaron-siegel: Again, `exprs.Expr` and `ts.ColumnType` don't belong here.

pixeltable/catalog/table_metadata.py
```diff
@@ -99,3 +102,37 @@ class VersionMetadata(TypedDict):
...
+        'type': ts.ColumnType | builtins.type,
+        'value': exprs.Expr,
+        'primary_key': bool,
+        'stored': bool,
+        'media_validation': Literal['on_read', 'on_write'],
+        'destination': str,
+    },
+    total=False,
+)
+ColumnOptions.__doc__ = """Optional metadata for a column specification.
```
aaron-siegel: Instead of doing it this way, a cleaner approach is
```
class ColumnOptions(TypedDict):
    """Docstring"""
    type: type
    value: exprs.Expr
    # etc
```

pixeltable/catalog/table_metadata.py
```diff
@@ -99,3 +102,37 @@ class VersionMetadata(TypedDict):
...
+    type: The column type (e.g., `pxt.Image`, `str`). Required unless `value` is specified.
+    value: A Pixeltable expression for computed columns. Mutually exclusive with `type`.
+    primary_key: Whether this column is part of the primary key. Defaults to `False`.
+    stored: Whether to store the column data. Defaults vary by column type.
+    media_validation: When to validate media; `'on_read'` or `'on_write'`.
+    destination: An object store reference for persisting computed files. Only applicable for computed columns.
+"""
+
+ColumnSpec = ts.ColumnType | builtins.type | ColumnOptions | exprs.Expr
+"""A column specification for use in a schema dict.
```
aaron-siegel: This won't actually appear in our docs. But see my above comment; I'm not convinced the named composite type serves a useful purpose here.

pixeltable/catalog/table_metadata.py
```diff
@@ -99,3 +102,37 @@ class VersionMetadata(TypedDict):
     """The number of errors encountered during this version."""
     schema_change: str | None
     """A description of the schema change that occurred in this version, if any."""
+
+
+ColumnOptions = TypedDict(
```
aaron-siegel: I'm not sure if `table_metadata.py` is the right place for this to go. Other packages have a `types.py` module for things like this; maybe it's time for us to create one?

pixeltable/catalog/table.py
```diff
@@ -723,7 +749,7 @@ def _validate_column_spec(cls, name: str, spec: dict[str, Any]) -> None:
         if 'type' not in spec and 'value' not in spec:
             raise excs.Error(f"Column {name!r}: 'type' or 'value' must be specified")
 
-        if 'type' in spec and not isinstance(spec['type'], (ts.ColumnType, type, _GenericAlias)):
+        if 'type' in spec and not isinstance(spec['type'], (ts.ColumnType, type)):
             raise excs.Error(f"Column {name!r}: 'type' must be a type or ColumnType; got {spec['type']}")
```
aaron-siegel: We can change this error to just say it must be a type.

It's fine if we accept ColumnType for backward compatibility (as a deprecated pattern) but we shouldn't expect users to know what they are.

pixeltable/catalog/table.py
```diff
@@ -755,7 +781,8 @@ def _create_columns(cls, schema: dict[str, Any]) -> list[Column]:
             stored = True
             destination: str | None = None
 
-            if isinstance(spec, (ts.ColumnType, type, _GenericAlias)):
+            # TODO: Should we fully deprecate passing ts.ColumnType here?
```
aaron-siegel, line 799: Another option is to raise a DeprecationWarning, but I agree let's keep this as a TODO and it doesn't have to be decided now.

tests/test_table.py
```diff
@@ -870,15 +870,15 @@ def test_schema_types(self, uses_db: None) -> None:
             'json_col': pxt.Json,
             'req_json_col': pxt.Required[pxt.Json],
             'array_col': pxt.Array[(5, None, 3), pxt.Int],  # type: ignore[misc]
-            'req_array_col': pxt.Required[pxt.Array[(5, None, 3), pxt.Int]],  # type: ignore[misc]
+            'req_array_col': pxt.Required[pxt.Array[(5, None, 3), pxt.Int]],  # type: ignore[dict-item, misc]
```
aaron-siegel, line 884: Hmmmm, is this necessary because we removed the references to `_GenericAlias`?

pixeltable/catalog/table.py
christopherpestano: Makes sense I changed it to be such that type | ColumnSpec is the type here. As to the point of expr.Expr, I think it makes more sense to have a unified ColumnSpec as we can then enforce improve code sharing between the computed and non-computed column scenarios.

pixeltable/catalog/table.py
christopherpestano: Changed to type | ColumnSpec

pixeltable/catalog/table_metadata.py
christopherpestano: When I tried this it didn't worked since type is already a reserved keyword in python. The functional declaration I followed is the recommended path when you can't declare the keys in that format you pasted.

pixeltable/catalog/table_metadata.py
christopherpestano: Removed composite

pixeltable/catalog/table_metadata.py
christopherpestano: Never mind, works, changed it to be the class definition though type: type is a bit odd.

tests/test_table.py
christopherpestano: Seems like so, I think _GenericAlias was covering some of these weirder Required cases

pixeltable/catalog/table.py
christopherpestano: Modified to reflect new schema type of type | ColumnSpec

christopherpestano: Expanded to include ColumnSpec

pixeltable/catalog/table.py
christopherpestano: Yeah I think what we should do going forward about _GenericAlias should probably be discussed in a wider forum.

pixeltable/catalog/table.py
christopherpestano: Makes sense to me, changed it to just be 'type' must be a type; got spec

pixeltable/catalog/table.py
christopherpestano: Expanded

pixeltable/catalog/table.py
christopherpestano: Changed terminology to TypedDict with an additional section indicating what fields there are within that TypedDict.

pixeltable/catalog/table.py
christopherpestano: Updated

pixeltable/catalog/table.py
christopherpestano: Moreover, there isn't necessarily anything code-wise that would reasonably stop someone from being able to add a computed column with add_columns, just that it makes more sense for them to use the actually named method for computed columns.

pixeltable/catalog/table.py
christopherpestano: Removed in several places, let me know if any still remains.

pixeltable/catalog/table_metadata.py
christopherpestano: This seems like it could be related to the _GenericAlias conversation, can we maybe hold off on this until we decide how we're going to standardize the typing for our fn inputs? If not totally fine I can make that change but it'd make more sense implement that with a plan on how we're going to standardize typing for this repo.

pixeltable/catalog/table_metadata.py
```diff
@@ -101,3 +104,20 @@ class VersionMetadata(TypedDict):
     """The number of errors encountered during this version."""
     schema_change: str | None
     """A description of the schema change that occurred in this version, if any."""
+
+
+class ColumnSpec(TypedDict, total=False):
+    """Column specification, a dictionary representation of a column's schema."""
```
aaron-siegel: It'd be clearer to say here: "Exactly one of `type` or `value` must be included in the dictionary," rather than in (or in addition to) the parameter docstrings.

pixeltable/catalog/table_metadata.py
```diff
@@ -101,3 +104,20 @@ class VersionMetadata(TypedDict):
...
+    value: exprs.Expr
+    """A Pixeltable expression for computed columns. Mutually exclusive with `type`."""
+    primary_key: bool
+    """Whether this column is part of the primary key. Defaults to `False`."""
+    stored: bool
+    """Whether to store the column data. Defaults vary by column type."""
+    media_validation: Literal['on_read', 'on_write']
+    """When to validate media; `'on_read'` or `'on_write'`."""
+    destination: str | Path
+    """An object store reference for persisting computed files. Only applicable for computed columns."""
```
aaron-siegel: Clearest would be something like:
```
Destination for storing computed output files. Can be either:

- A local pathname (such as `path/to/outputs/`), or
- The URI of an object store (such as `s3://my-bucket/outputs/`).
```
I'm concerned that "object store reference" might leave some users guessing.

tests/test_index.py
```diff
@@ -416,9 +416,9 @@ def test_add_embedding_index_if_exists(
         # cannot use if_exists to ignore or replace an existing index
         # that is not an embedding (like, default btree indexes).
         assert indexes[0]['_name'] == 'idx0'
-        for ie in ('ignore', 'replace', 'replace_force'):
+        for ie in ['ignore', 'replace', 'replace_force']:
```
aaron-siegel: Why did this change? It's correct to use a tuple here. Changing to the list is what necessitated adding the `type: ignore` directive below.

tests/test_table.py
aaron-siegel: Ok.

tests/test_types.py
```diff
@@ -171,10 +169,10 @@ def test_from_python_type(self, init_env: None) -> None:
                 "Image[(100, 200), 'RGB']",
             ),
             Image['RGB']: (ImageType(height=None, width=None, mode='RGB', nullable=False), "Image['RGB']"),  # type: ignore[misc]
-            Literal['a', 'b', 'c']: (StringType(nullable=False), 'String'),
-            Literal[1, 2, 3]: (IntType(nullable=False), 'Int'),
-            Literal[1, 2.0, 3]: (FloatType(nullable=False), 'Float'),
-            Literal['a', 'b', None]: (StringType(nullable=True), 'String'),  # noqa: PYI061
+            Literal['a', 'b', 'c']: (StringType(nullable=False), 'String'),  # type: ignore[dict-item]
```
aaron-siegel: I don't love littering the code with these type: ignores; since this is testing code, maybe it's better to just make the dict into a `dict[Any, tuple[ColumnType, str]]`?

tests/test_view.py
```diff
@@ -514,7 +514,7 @@ def test_unstored_columns_non_image(self, uses_db: None) -> None:
             'ud2': {'value': t.d2, 'stored': False},
         }
 
-        v1 = pxt.create_view('v1', t, additional_columns=add_schema1)
+        v1 = pxt.create_view('v1', t, additional_columns=add_schema1)  # type: ignore[arg-type]
```
aaron-siegel: Why do we need all these type: ignores? Is that really the best/only solution to this problem?

aaron-siegel (changes requested):

tests/test_index.py
christopherpestano: Reverted back to tuple, I think I made this adjustment based on a bad mypy hint when writing the code.

tests/test_types.py
christopherpestano: Sure that works, though this section already does have quite a few type ignores, changed key to any type and removed added ignores.

tests/test_view.py
christopherpestano: Depends, in this case I am able to remove the type ignore by explicitly adding the type dict[str, ColumnSpec] to add_schema1. However, for the case where we are intentionally raising an error and using # type: ignore[dict-item] we have to include the ignore since mypy will warn you for passing an invalid value into ColumnSpec (which we expect).

pixeltable/catalog/table_metadata.py
christopherpestano: Added specification for valid input types

pixeltable/catalog/table_metadata.py
christopherpestano: Ok based off discussion, followed our decision to create types.py. Currently just contains ColumnSpec in the future we should move the rest of the typedicts into there.

pixeltable/catalog/table_metadata.py
christopherpestano: Sure added it in addition to the parameter docstrings.

pixeltable/catalog/column.py
```diff
@@ -136,7 +137,11 @@ def __init__(
 
         # computed cols also have storage columns for the exception string and type
         self.sa_cellmd_col = None
-        self._explicit_destination = destination
+
+        if isinstance(destination, Path):
+            destination = str(destination)
+        else:
```
aaron-siegel: Don't we still want to set `self._explicit_destination` in cases where it's a Path?

tests/test_index.py
```diff
@@ -459,7 +459,7 @@ def test_update_img(
             new_rows.append(row)
 
         # create table with fewer rows to speed up testing
-        schema = {'pkey': ts.IntType(nullable=False), 'img': pxt.Image, 'category': pxt.String, 'split': pxt.String}
+        schema = {'pkey': Required[pxt.Int], 'img': pxt.Image, 'category': pxt.String, 'split': pxt.String}
```
aaron-siegel: pxt.Required

pixeltable/types.py
```diff
@@ -0,0 +1,31 @@
...
+    """Whether this column is part of the primary key. Defaults to `False`."""
+    stored: bool
+    """Whether to store the column data. Defaults vary by column type."""
+    media_validation: Literal['on_read', 'on_write']
+    """When to validate media; `'on_read'` or `'on_write'`."""
+    destination: str | Path
+    """
+    Destination for storing computed output files. Only applicable for computed columns.
+    Can be:
+        - A local pathname (such as `path/to/outputs/`), or
```
aaron-siegel, line 31: Blank line required before markdown list (or else rendering will fail in mintlify)

pixeltable/types.py
```diff
@@ -0,0 +1,31 @@
+from pathlib import Path
+from typing import Literal, TypedDict
+
+from pixeltable import exprs
```
aaron-siegel: Guard this with `if TYPE_CHECKING:` (as discussed at standup)? With the goal of making this module "universally importable"

pixeltable/catalog/column.py
sergey-mkhitaryan: should this block be simply `self._explicit_destination = str(destination)`?

pixeltable/types.py
```diff
@@ -0,0 +1,31 @@
...
+
+"""Types used for type annotations across the Pixeltable codebase."""
+
+
+class ColumnSpec(TypedDict, total=False):
```
sergey-mkhitaryan, line 10: Do we want this to show up in the public docs? if so it needs to be added to `public_api.opml`. Run `make docs-serve` to preview.

aaron-siegel: Yes I think we do.

tests/test_index.py
christopherpestano: Fixed

pixeltable/types.py
christopherpestano: Fixed

pixeltable/types.py
christopherpestano: Added, was having issues but seems I needed to wrap exprs.Expr in a 'exprs.Expr'

pixeltable/types.py
christopherpestano: Added to public docs, put it under core api since most users will be searching for it when they see it referenced in the other public facing api fns

pixeltable/catalog/column.py
christopherpestano: > should this block be simply `self._explicit_destination = str(destination)`?

In this case we don't want to convert automatically to string, since we haven't actually validated that the destination is of type str | Path, which is done in the next step _verify_column.

christopherpestano: > Don't we still want to set `self._explicit_destination` in cases where it's a Path?

Removed the else so that _explict_destination is actually being set.

pixeltable/catalog/table.py
```diff
@@ -508,7 +509,15 @@ def add_columns(
         The format of the `schema` argument is a dict mapping column names to their types.
 
         Args:
-            schema: A dictionary mapping column names to types.
+            schema: A dictionary mapping column names to a `type` or a `ColumnSpec` dict,
```
aaron-siegel: We can probably just link to the `ColumnSpec` class now, rather than spelling out the details of the dict here.

pixeltable/catalog/table.py
```diff
@@ -571,13 +592,21 @@ def add_column(
...
         if_exists: Literal['error', 'ignore', 'replace', 'replace_force'] = 'error',
-        **kwargs: ts.ColumnType | builtins.type | _GenericAlias | exprs.Expr,
+        **kwargs: type | ColumnSpec,
     ) -> UpdateStatus:
         """
         Adds an ordinary (non-computed) column to the table.
 
         Args:
-            kwargs: Exactly one keyword argument of the form `col_name=col_type`.
+            kwargs: Exactly one keyword argument of the form `col_name=type` or `col_name=col_spec_dict`,
```
aaron-siegel, line 593: Likewise - linking to the ColumnSpec docs will tighten up and deduplicate the documentation.

pixeltable/__init__.py
```diff
@@ -60,6 +60,7 @@
     Timestamp,
     Video,
 )
+from .types import ColumnSpec
```
aaron-siegel: It's not clear we want this, versus just treating `types` as a public module.

tests/test_view.py
```diff
@@ -7,7 +7,9 @@
 
 import pixeltable as pxt
 from pixeltable.catalog import Catalog
+from pixeltable.exprs import expr
```
aaron-siegel: Typically we just do `from pixeltable import exprs` and reference `Expr` as `exprs.Expr` (rather than `exprs.expr.Expr`). (Either will work - just pointing out the convention we use)

pixeltable/catalog/table.py
christopherpestano: Replaced with link

pixeltable/catalog/table.py
christopherpestano: Replaced with link

pixeltable/__init__.py
christopherpestano: Yeah, I'm realizing that even this wouldn't work since types isn't in the blacklist, removed this line and allowing types to just be public.

tests/test_view.py
christopherpestano: Changed to be appropriate import, think this was auto import.

pixeltable/types.py
```diff
@@ -0,0 +1,33 @@
+"""Types used for type annotations across the Pixeltable codebase."""
```
mkornacker: "User-facing types used ..."

pixeltable/types.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+
+    type: type
+    """The column type (e.g., `pxt.Image`, `str`). Required unless `value` is specified."""
+    value: 'exprs.Expr'
+    """A Pixeltable expression for computed columns. Mutually exclusive with `type`."""
+    primary_key: bool
+    """Whether this column is part of the primary key. Defaults to `False`."""
+    stored: bool
+    """Whether to store the column data. Defaults vary by column type."""
```
mkornacker: They actually don't, the default is True.

tests/test_table.py
```diff
@@ -448,14 +449,20 @@ def test_table_metadata(self, uses_db: None, clip_embed: pxt.Function) -> None:
             )
 
     def test_media_validation(self, uses_db: None) -> None:
-        tbl_schema = {'img': {'type': pxt.Image, 'media_validation': 'on_write'}, 'video': pxt.Video}
+        tbl_schema: dict[str, ColumnSpec | type] = {
```
mkornacker, line 452: Is this what users need to write in order to avoid mypy errors?

pixeltable/catalog/table.py
```diff
@@ -508,7 +509,8 @@ def add_columns(
         The format of the `schema` argument is a dict mapping column names to their types.
 
         Args:
-            schema: A dictionary mapping column names to types.
+            schema: A dictionary mapping column names to a `type` or a
+                [`ColumnSpec`][pixeltable.ColumnSpec] TypedDict.
```
mkornacker: TypedDict -> dict

pixeltable/catalog/table.py
```diff
@@ -571,13 +585,14 @@ def add_column(
...
-        **kwargs: ts.ColumnType | builtins.type | _GenericAlias | exprs.Expr,
+        **kwargs: type | ColumnSpec,
     ) -> UpdateStatus:
         """
         Adds an ordinary (non-computed) column to the table.
 
         Args:
-            kwargs: Exactly one keyword argument of the form `col_name=col_type`.
+            kwargs: Exactly one keyword argument of the form `col_name=type` or `col_name=col_spec_dict`,
+                where `col_spec_dict` is a [`ColumnSpec`][pixeltable.ColumnSpec] TypedDict.
```
mkornacker: "is a ... dict" reads better (and is consistent with your example).

pixeltable/catalog/table.py
```diff
@@ -597,9 +612,31 @@ def add_column(
...
+
+            >>> tbl.add_column(
+            ...     img_col={
+            ...         'type': pxt.Image,
+            ...         'stored': True,
+            ...         'media_validation': 'on_write',
+            ...     }
+            ... )
+
+            Alternatively, simple column insertions can be expressed using `add_columns`:
```
mkornacker: What are simple column insertions?

Also, "column insertion" isn't a thing in the database world - you insert rows and you're adding columns (in Sql: Alter Table ... Add Column ...).

tests/test_table.py
christopherpestano: Yes, mypy if not given an explict type for dicts with mixed types for values, will assume the generic object as being the value of the dict. As it does not create a union type from the given values, rather it gives the values the type object which is technically correct. This is an open issue on mypy's end https://github.com/python/mypy/issues/3816

christopherpestano: aka. dict = {'a': 1, 'b': 'c'} will resolve to dict[str, object] and list = ['a', 2] will resolve to list[object]

pixeltable/catalog/table.py
christopherpestano: Sure, changed

pixeltable/catalog/table.py
christopherpestano: Changed to dict, makes sense it's a bit easier for users to comprehend what is valid input

pixeltable/catalog/table.py
christopherpestano: Yeah insertion isn't the correct terminology here, changed that to just be adding column. As for simple I think that was a miswrite, I think when I was doing this I intended for that to be `single` not `simple` anycase remove the word entirely.

pixeltable/types.py
christopherpestano: Added

pixeltable/types.py
christopherpestano: My bad, fixed documentation to reflect this

tests/test_table.py
aaron-siegel: @mkornacker This only matters in cases where the user mixes `type` instances and `ColumnSpec` instances _in the same schema_, which is probably fairly unusual. (And of course they also have to be using mypy for it to matter.)

mkornacker (approved):

PR #1123, opened 2026-02-03
[PXT-995] Literal should serialize its entire type info

sergey-mkhitaryan:
Such as array dtype.

Includes a catalog conversion so that we don't have to support two formats.

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -1322,7 +1322,7 @@ class Table(SchemaObject):
         if (column is None) == (idx_name is None):
             raise excs.Error("Exactly one of 'column' or 'idx_name' must be provided")
 
-        with Catalog.get().begin_xact(tbl=self._tbl_version_path, for_write=True, lock_mutable_tree=False):
+        with Catalog.get().begin_xact(tbl=self._tbl_version_path, for_write=True, lock_mutable_tree=True):
             col: Column = None
             if idx_name is None:
                 col = self._resolve_column_parameter(column)
--- a/pixeltable/catalog/table_version.py
+++ b/pixeltable/catalog/table_version.py
@@ -845,6 +845,10 @@ class TableVersion:
         idx_md.schema_version_drop = self.schema_version
         assert idx_md.name in self.idxs_by_name
         idx_info = self.idxs_by_name[idx_md.name]
+
+        # Drop the physical index from the store
+        self.store_tbl.drop_index(idx_id)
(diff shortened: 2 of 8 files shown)


pixeltable/exprs/literal.py
mkornacker: Let's use col_type.as_dict() everywhere here, that should simplify the code.

pixeltable/exprs/literal.py
```diff
@@ -85,54 +85,66 @@ def eval(self, data_row: DataRow, row_builder: RowBuilder) -> None:
...
-            return {'val': self.val.tolist(), 'val_t': self.col_type._type.name, **super()._as_dict()}
+            encoded_val = self.val.tolist()
         else:
-            return {'val': self.val, **super()._as_dict()}
+            encode_col_type = False
+            encoded_val = self.val
+
+        res = {'val': encoded_val, **super()._as_dict()}
+        if encode_col_type:
+            res.update({'val_t': self.col_type.as_dict()})
```
aaron-siegel: Why not just *always* include the col_type? And call it the standard `col_type` rather than the ad hoc `val_t` (I take the blame for that one)

pyproject.toml
```diff
@@ -257,6 +257,7 @@ markers = [
 filterwarnings = [
     # Treat *our* warnings as errors (we shouldn't see them fall through in tests)
     'error::pixeltable.exceptions.PixeltableDeprecationWarning',
+    'error::pixeltable.exceptions.PixeltableWarning',
```
aaron-siegel, line 259: We can remove `PixeltableDeprecationWarning` now, since it's a subclass of `PixeltableWarning`

tool/create_test_db_dump.py
```diff
@@ -234,7 +234,9 @@ def create_tables(self) -> None:
         ]
 
         self.__add_expr_columns(t, 'base_table')
-        t.insert(rows)
+        status = t.insert(rows)
```
aaron-siegel, line 237: Replace this and the 2 subsequent lines with

`validate_update_status(t.insert(rows), expected_rows=num_rows)`

sergey-mkhitaryan: `tests` is not a proper package (and we are not in it), so my understanding that there's no way to easily import `validate_update_status` here. Let me know if there is a way.

aaron-siegel: You are correct.

PR #1124, opened 2026-02-05
PXT-1023: Support for ragged arrays in export_parquet()

mkornacker:


pixeltable/io/parquet.py
```diff
@@ -57,6 +59,7 @@ def export_parquet(
                         If False, will raise an error if the Query has any image column.
                         Default False.
     """
+    Env.get().require_package('pyarrow')
```
aaron-siegel: `pyarrow` is a required dependency of Pixeltable, so this shouldn't be necessary.

tests/io/test_parquet.py
```diff
@@ -11,11 +12,40 @@
...
     skip_test_if_not_installed,
+    validate_update_status,
 )
 
 
 class TestParquet:
-    def test_import_parquet_examples(self, uses_db: None, tmp_path: pathlib.Path) -> None:
+    def validate_parquet_files(self, path: pathlib.Path, rows: list[dict]) -> None:
+        """Validate that the parquet files in the directory match the rows."""
+        from pyarrow import parquet
```
aaron-siegel: Move to top of file

tests/io/test_parquet.py
```diff
@@ -11,11 +12,40 @@
...
+
+        from pixeltable.utils.arrow import to_pydict
```
aaron-siegel: Move to top of file

pixeltable/utils/arrow.py
```diff
@@ -211,20 +228,27 @@ def create_arrow_schema() -> None:
...
+    FixedShapeTensor columns are converted to numpy arrays with the correct shape.
     """
     out: dict[str, list | np.ndarray] = {}
     for k, name in enumerate(batch.schema.names):
         col = batch.column(k)
         if isinstance(col.type, pa.FixedShapeTensorType):
-            # treat array columns as numpy arrays to easily preserve numpy type
-            out[name] = col.to_numpy(zero_copy_only=False)
+            # we want a list of ndarrays of the correct shape
+            col_array: pa.Array
```
aaron-siegel: `col` is already known to have type `pa.ChunkedArray | pa.Array`, so you can do this more simply by
```
if isinstance(col, pa.ChunkedArray):
    col = col.combine_chunks()
```
it's 2 lines of code instead of 6, and one fewer local variable. (mypy understands what's going on: you're refining the type of `col` via the conditional)

tests/io/test_parquet.py
```diff
@@ -98,7 +128,7 @@ def test_import_parquet_from_remote(self, uses_db: None, source: str) -> None:
         assert r1['bigint_col'] == 10 and r1['double_col'] == 10.1
         assert abs(r1['float_col'] - 1.1) < 1e-5  # float32 precision
 
-    def test_import_parquet(self, uses_db: None, tmp_path: pathlib.Path) -> None:
+    def test_import(self, uses_db: None, tmp_path: pathlib.Path) -> None:
```
aaron-siegel: I have a fairly strong preference for the more specific test names. The reason is that I often run individual tests using `pytest -k`, as in
```
pytest -vk test_import_parquet
```
and if the test has some generic name like `test_import` then it's impossible to do it this way (it will pick up a million unrelated tests that start with `test_import`).

I think we should stick with the more specific test names.

pixeltable/io/parquet.py
mkornacker: Done.

pixeltable/utils/arrow.py
mkornacker: Done.

tests/io/test_parquet.py
mkornacker: Done.

tests/io/test_parquet.py
mkornacker: Done.

tests/io/test_parquet.py
mkornacker: That's just repeating the class name in the method name.

Why not just run it as
pytest -v tests/io/test_parquet.py::TestParquet::test_import
which has the added advantage of not taking 20 seconds to figure out what exactly to run.

aaron-siegel: It's a lot more work to type that. I'd lose 60 seconds finding the place to copy/paste that string from, in order to save those 20 seconds (and `pytest -vk` definitely doesn't take 20 seconds on my machine).

Let me ask this: Shortening the names has a cost: increased developer friction in interacting with the test system. What do we gain in return by doing so? (i.e., what problem are you trying to solve?)

PR #1127, opened 2026-02-06
[PXT-969] Materialize stores_cellmd as boolean property in ColumnMd

sergey-mkhitaryan:
This is required for moving `col_type` and other potentially alterable properties to `SchemaColumn`.

The logic behind `stores_cellmd` is:
1. For index value columns, the index decides if cellmd column is needed
2. For index undo columns, it's always False
3. For non-stored columns, it's always False
4. cellmd is enabled for computed and media columns, and column types that support offloading to a file (json, binary, etc.)

This logic is spread across `TableVersion` and `Column`. This change includes a refactoring that consolidates it in `Column` by introducing `Column` factory methods. It also moves some other `Column`-related code to `Column`.

This change also includes the necessary catalog conversion that applies the above `stores_cellmd` logic to pre-existing columns.


pixeltable/metadata/converters/convert_45.py
```diff
@@ -0,0 +1,47 @@
...
+                case 'pixeltable.index.btree.BtreeIndex':
+                    stores_cellmd = False
+                case 'pixeltable.index.embedding_index.EmbeddingIndex':
+                    stores_cellmd = True
+                case _:
+                    raise AssertionError(f'Unexpected index class {idx["class_fqn"]}')
+        elif column_md['id'] in idx_undo_cols:
+            stores_cellmd = False
+        else:
+            stored = column_md['stored']
```
sergey-mkhitaryan: @aaron-siegel @mkornacker any idea why `stored` is optional in `ColumnMd`? Is it safer to treat it as required (the conversion can fail) or is it better to assume stored/not stored if this is unset?

aaron-siegel: No idea - @mkornacker might know?

pixeltable/catalog/column.py
```diff
@@ -138,6 +140,249 @@ def __init__(
...
+            cls_name = md.class_fqn.rsplit('.', 1)[-1]
+            idx_cls = getattr(index, cls_name)
+            idx = idx_cls.from_dict(md.init_args)
+            val_col_idxs[md.index_val_col_id] = idx
+            undo_col_idxs[md.index_val_undo_col_id] = idx
+
+        # Sort columns in column_md by the position specified in col_md.id to guarantee that all references
+        # point backward.
+        sorted_column_md = sorted(tbl_version.tbl_md.column_md.values(), key=lambda item: item.id)
+        cols = []
```
aaron-siegel: needs type hint

pixeltable/catalog/column.py
```diff
@@ -161,6 +407,23 @@ def to_md(self, pos: int | None = None) -> tuple[schema.ColumnMd, schema.SchemaC
...
+    def verify(self) -> None:
+        """Check integrity of user-supplied Column"""
+        if self.name is None:
+            return
+        Column.validate_name(self.name)
+        if self.stored is False and not self.is_computed:
+            raise excs.Error(f'Column {self.name!r}: `stored={self.stored}` only applies to computed columns')
+        if self.stored is False and self.has_window_fn_call():
+            raise excs.Error(
+                (
```
aaron-siegel: legacy issue but might as well fix: spurious parens

pixeltable/catalog/column.py
```diff
@@ -138,6 +140,249 @@ def __init__(
         self.sa_cellmd_col = None
         self._explicit_destination = destination
 
+    @classmethod
+    def instantiate_cols(cls, tbl_version: TableVersion) -> list[Column]:
+        """Instantiates and returns Column objects based on the metadata of the given TableVersion.
+
+        Args:
```
mkornacker: Style note: docstrings for functions that aren't user-facing should limit themselves to non-obvious commentary (rather than aiming for completeness).

pixeltable/catalog/column.py
```diff
@@ -138,6 +140,249 @@ def __init__(
...
+            tbl_version: the TableVersion being initialized
+        """
+        val_col_idxs: dict[int, index.IndexBase] = {}  # value column id -> index
+        undo_col_idxs: dict[int, index.IndexBase] = {}  # undo column id -> index
+        for md in tbl_version.tbl_md.index_md.values():
```
mkornacker: This code really belongs in TableVersion (_init_cols()?), because it need to access its internal state.

pixeltable/catalog/column.py
```diff
@@ -138,6 +140,249 @@ def __init__(
...
+            cls_name = md.class_fqn.rsplit('.', 1)[-1]
+            idx_cls = getattr(index, cls_name)
+            idx = idx_cls.from_dict(md.init_args)
+            val_col_idxs[md.index_val_col_id] = idx
+            undo_col_idxs[md.index_val_undo_col_id] = idx
+
+        # Sort columns in column_md by the position specified in col_md.id to guarantee that all references
```
mkornacker: Fix comment, it doesn't make sense (id != position).

pixeltable/catalog/column.py
```diff
@@ -138,6 +140,249 @@ def __init__(
...
+            schema_version_add=schema_version,
+            schema_version_drop=None,
+        )
+        undo_col.col_type = undo_col.col_type.copy(nullable=True)
+        undo_col.tbl_handle = tbl_handle
+        return val_col, undo_col
+
+    @classmethod
+    def create_column(cls, name: str, spec: Any) -> Column:
+        """Creates a brand new Column based on the provided spec"""
```
mkornacker: Remove "brand new" (it doesn't add any information).

"and validates all column parameters (eg, destination)"

pixeltable/catalog/column.py
```diff
@@ -138,6 +140,249 @@ def __init__(
...
+    def create_index_columns(
+        cls,
+        tbl_handle: TableVersionHandle,
+        col: Column,
+        idx: index.IndexBase,
+        val_col_id: int,
+        undo_col_id: int,
+        schema_version: int,
+    ) -> tuple[Column, Column]:
+        """Create value and undo columns for a brand new index.
```
mkornacker: Remove "brand new" (it doesn't add any information).

pixeltable/catalog/column.py
```diff
@@ -138,6 +140,249 @@ def __init__(
...
+            stores_cellmd=False,
+            schema_version_add=schema_version,
+            schema_version_drop=None,
+        )
+        undo_col.col_type = undo_col.col_type.copy(nullable=True)
+        undo_col.tbl_handle = tbl_handle
+        return val_col, undo_col
+
+    @classmethod
+    def create_column(cls, name: str, spec: Any) -> Column:
```
mkornacker: This is a method of the Column class (create() is sufficient).

pixeltable/catalog/column.py
```diff
@@ -138,6 +140,249 @@ def __init__(
...
+            name,
+            col_type=col_type,
+            computed_with=value_expr,
+            stored=stored,
+            is_pk=primary_key,
+            media_validation=media_validation,
+            destination=destination,
+            stores_cellmd=stores_cellmd,
+        )
+        # Validate the column's resolved_destination. This will ensure that if the column uses a default (global)
```
mkornacker: Remove the comment, it doesn't say anything beyond "it validates destination", which we already pointed out in the method comment.

pixeltable/catalog/column.py
```diff
@@ -161,6 +407,21 @@ def to_md(self, pos: int | None = None) -> tuple[schema.ColumnMd, schema.SchemaC
         )
         return col_md, sch_md
 
+    def verify(self) -> None:
+        """Check integrity of user-supplied Column"""
```
mkornacker: We don't have any more user-supplied Column instances.

pixeltable/catalog/column.py
```diff
@@ -138,6 +140,249 @@ def __init__(
...
+        for col_name, col_type in output_dict.items():
+            stored = col_name not in unstored_cols
+            stores_cellmd = stored and (col_type.is_media_type() or col_type.supports_file_offloading())
+            col = cls(col_name, col_type, is_iterator_col=True, stored=stored, stores_cellmd=stores_cellmd)
+            iterator_cols.append(col)
+        return iterator_cols
+
+    @classmethod
+    def create_stored_proxy_column(cls, col: Column) -> Column:
+        """Creates a proxy column for the specified column."""
```
mkornacker, line 304: Remove "proxy" and replace with something more meaningful ("copy"?).

pixeltable/catalog/table.py
```diff
@@ -704,127 +695,17 @@ def add_computed_column(
...
-    @classmethod
-    def validate_column_name(cls, name: str) -> None:
-        """Check that a name is usable as a pixeltable column name"""
-        if is_system_column_name(name) or is_python_keyword(name):
-            raise excs.Error(f'{name!r} is a reserved name in Pixeltable; please choose a different column name.')
-        if not is_valid_identifier(name):
-            raise excs.Error(f'Invalid column name: {name}')
-
     @classmethod
     def _verify_column(cls, col: Column) -> None:
```
mkornacker, line 789: This doesn't look useful anymore.

pixeltable/catalog/view.py
```diff
@@ -93,9 +93,11 @@ def _create(
         select_list_columns: List[Column] = []
         if not include_base_columns:
             r = cls.select_list_to_additional_columns(select_list)
-            select_list_columns = cls._create_columns(r)
+            select_list_columns = [Column.create_column(name, spec) for name, spec in r.items()]
 
-        columns_from_additional_columns = cls._create_columns(additional_columns)
+        columns_from_additional_columns = [
```
mkornacker: Very awkward name. additional_columns is a misnomer. Let's clean this up.

pixeltable/catalog/column.py
aaron-siegel: This isn't something that was introduced in this PR - the code and comment was copied from elsewhere. Since this PR has been open for awhile, I'd suggest we focus on things that actually changed in this PR and if we notice other things that could stand to be fixed, file rough edges tickets instead.

pixeltable/catalog/column.py
aaron-siegel: Likewise, a legacy issue (not something that changed in this PR) and the terminology "proxy column" is used in multiple other places, so I think trying to "fix" it in just one of them would only create confusion/ambiguity.

So (cf. my previous comment) this terminology change is unrelated to the PR and if we think it's important to address, let's do it as a rough-edges ticket in a principled/structured way.

pixeltable/catalog/column.py
sergey-mkhitaryan: just removed this. it's not unusual when a constructor of a factory does some validation, we don't need to specifically talk about it.

pixeltable/catalog/column.py
sergey-mkhitaryan: Yes, let's not do this here. It would add a non-trivial number of LOC changed that have no direct link to the purpose of this pr.

pixeltable/catalog/table.py
sergey-mkhitaryan: unfortunately this can't be easily inlined because `View` overrides it

pixeltable/catalog/column.py
```diff
@@ -143,6 +145,224 @@ def __init__(
 
         self._explicit_destination = destination
 
+    @classmethod
+    def instantiate_cols(
```
mkornacker: From the last review round:

This code really belongs in TableVersion (_init_cols()?), because it need to access its internal state.

pixeltable/catalog/column.py
```diff
@@ -143,6 +145,224 @@ def __init__(
...
+            name=None,
+            computed_with=value_expr,
+            sa_col_type=idx.get_index_sa_type(value_expr.col_type),
+            stored=True,
+            stores_cellmd=idx.records_value_errors(),
+            schema_version_add=schema_version,
+            schema_version_drop=None,
+        )
+        val_col.col_type = val_col.col_type.copy(nullable=True)
+        val_col.tbl_handle = tbl_handle
```
mkornacker: You can do that via the c'tor.

pixeltable/catalog/column.py
```diff
@@ -143,6 +145,224 @@ def __init__(
...
+            stores_cellmd=stores_cellmd,
+        )
+        ObjectOps.validate_destination(column.destination, column.name)
+        return column
+
+    @classmethod
+    def _validate_column_spec(cls, name: str, spec: ColumnSpec) -> None:
+        """Check integrity of user-supplied Column spec
+
+        We unfortunately can't use something like jsonschema for validation, because this isn't strictly a JSON schema
```
mkornacker: Remove comment, it's not useful.

pixeltable/catalog/column.py
```diff
@@ -166,6 +387,21 @@ def to_md(self, pos: int | None = None) -> tuple[schema.ColumnMd, schema.SchemaC
         )
         return col_md, sch_md
 
+    def verify(self) -> None:
```
mkornacker, line 354: Explain in a docstring what gets validated.

pixeltable/catalog/column.py
```diff
@@ -166,6 +387,21 @@ def to_md(self, pos: int | None = None) -> tuple[schema.ColumnMd, schema.SchemaC
...
+        # TODO this verification should be done before or during the Column construction, not after it
+        if self.name is None:
+            return
```
mkornacker, line 363: That should explain why name being None means you don't need validation.

pixeltable/catalog/column.py
```diff
@@ -166,6 +387,21 @@ def to_md(self, pos: int | None = None) -> tuple[schema.ColumnMd, schema.SchemaC
...
+        Column.validate_name(self.name)
+        if self.stored is False and not self.is_computed:
+            raise excs.Error(f'Column {self.name!r}: `stored={self.stored}` only applies to computed columns')
```
mkornacker, line 366: Error is used for user-facing errors. Are these expected to bubble up to the user?

We're already validating ColumnSpec, is there any reason we need to validate the resulting Column as well? The division of labor is confusing.

For the view-specific validation: Table._verify_column() should be a no-op and View can add its specific logic.

pixeltable/catalog/column.py
sergey-mkhitaryan: sorry, it fell through the cracks. done now.

pixeltable/catalog/column.py
sergey-mkhitaryan: This particular diff is a part of a "refactoring" aspect of this pr, so let's leave it be, otherwise I'll never be able to merge it. I agree that this column validation is wonky, and I left a TODO to fix it later. Once columns self-validate automatically, this function will go away, and so will `Table._verify_column()`.

sergey-mkhitaryan: Actually I don't know if `Table._verify_column()` can go away completely even if it's a no-op because `Table` needs a way to trigger the validation even if any validation is for `View` columns only.

pixeltable/catalog/column.py
```diff
@@ -151,6 +150,180 @@ def __init__(
...
+            destination=destination,
+            custom_metadata=custom_metadata,
+            comment=comment,
+            stores_cellmd=stores_cellmd,
+        )
+        ObjectOps.validate_destination(column.destination, column.name)
+        return column
+
+    @classmethod
+    def create_iterator_column(cls, name: str, col_type: ts.ColumnType, is_stored: bool) -> Column:
```
mkornacker: This doesn't have much to do with iterators.

A simple 
def stores_cellmd(cls, is_stored, col_type) -> bool
and invoking the c'tor directly would be more transparent.

pixeltable/catalog/table_version.py
mkornacker, line 469: I think it would still be worth breaking out the column initialization part, like you had before. But leave it in TableVersion.

PR #1128, opened 2026-02-07
[PXT-995] improve test migration coverage of literals of various types

sergey-mkhitaryan:
Adds some types of columns and literals to the dump that weren't there before.

### DIFF
--- a/pixeltable/catalog/table.py
+++ b/pixeltable/catalog/table.py
@@ -589,7 +589,7 @@ class Table(SchemaObject):
             Information about the execution status of the operation.
 
         Raises:
-            Error: If the column name is invalid, or already exists and `if_exists='erorr'`,
+            Error: If the column name is invalid, or already exists and `if_exists='error'`,
                 or `if_exists='replace*'` but the column has dependents or is a basetable column.
 
         Examples:
--- a/pixeltable/catalog/table_version.py
+++ b/pixeltable/catalog/table_version.py
@@ -1011,7 +1011,6 @@ class TableVersion:
         cols_with_excs: list[Column] = []
         for col in cols_to_add:
             assert col.id is not None
-            excs_per_col = 0
             col.schema_version_add = self.schema_version
             # add the column to the lookup structures now, rather than after the store changes executed successfully,
             # because it might be referenced by the next column's value_expr
@@ -1035,10 +1034,11 @@ class TableVersion:
 
             # populate the column
             plan = Planner.create_add_column_plan(self.path, col)
+            excs_per_col = 0
             with Env.get().report_progress():
                 try:
                     plan.ctx.title = self.display_str()
-                    excs_per_col = self.store_tbl.load_column(col, plan, on_error == 'abort')
+                    excs_per_col = self.store_tbl.write_column(col, plan, on_error == 'abort')
                 except sql_exc.DBAPIError as exc:
                     Catalog.get().convert_sql_exc(exc, self.id, self.handle, convert_db_excs=True)
                     # If it wasn't converted, re-raise as a generic Pixeltable error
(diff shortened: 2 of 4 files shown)


tool/create_test_db_dump.py
```diff
@@ -159,6 +161,11 @@ def create_tables(self) -> None:
             'c9': pxt.Audio,
             'c10': pxt.Video,
             'c11': pxt.Document,
+            'c12': pxt.Array[np.float64, (10,)],  # type: ignore[misc]
```
mkornacker, line 165: We should probably include other ArrayType variants, such as one without a shape and maybe also one with Nones in the shape.

tool/create_test_db_dump.py
```diff
@@ -159,6 +161,11 @@ def create_tables(self) -> None:
...
+            'c13': pxt.UUID,
+            'c14': pxt.Date,
+            'c15': pxt.Timestamp,
```
aaron-siegel: `Timestamp` is already included (unless there's a specific reason to include it a second time - this just increases the size of the dumps)

tool/create_test_db_dump.py
```diff
@@ -436,6 +459,24 @@ def test_udf_stored_batched(strings: Batch[str], *, upper: bool = True) -> Batch
     return [string.upper() if upper else string.lower() for string in strings]
 
 
+_DEFAULT_B = np.ones(10, dtype=np.float64)
+
+
+@pxt.udf(_force_stored=True)
```
aaron-siegel: probably best to avoid using `_force_stored` unless specifically trying to test pickled UDFs (otherwise will create complications when we move away from pickled UDFs)

tool/create_test_db_dump.py
```diff
@@ -436,6 +459,24 @@ def test_udf_stored_batched(strings: Batch[str], *, upper: bool = True) -> Batch
...
+def udf_with_array_literal(
```
aaron-siegel: Can we be consistent in the naming? (`udf_` vs. `test_udf_`)

tool/create_test_db_dump.py
```diff
@@ -425,6 +437,17 @@ def q2(s: str) -> pxt.Query:
 
         add_computed_column('sim_output', q2(t.c1))
 
+        add_computed_column('c12_udf_with_array_literal', udf_with_array_literal(t.c12), stored=True)
```
aaron-siegel: Let's be consistent with column names? `_udf_with` vs. `_expr_with`

tool/create_test_db_dump.py
```diff
@@ -425,6 +437,17 @@ def q2(s: str) -> pxt.Query:
...
+        add_computed_column('c13_expr_with_uuid_literal', test_udf_with_default_uuid(), stored=True)
+        add_computed_column('c14_expr_with_date_literal', pxtf.date.isocalendar(datetime.date(2026, 2, 6)), stored=True)
+        add_computed_column(
+            'c15_expr_with_timestamp_literal',
+            pxtf.timestamp.posix_timestamp(t.c15)
+            - pxtf.timestamp.posix_timestamp(datetime.datetime(2026, 1, 1, tzinfo=ZoneInfo('UTC'))),
+            stored=True,
```
aaron-siegel: `stored=True` shouldn't be necessary

sergey-mkhitaryan:
@aaron-siegel 

> if you're trying to test storing literals in the database, you need to include some literals as explicit arguments to the UDF call. If they only show up as parameter defaults in the UDF definitions, they won't be serialized to the DB.

I verified that the literals do show up in the database -- some of them in the `functions` table, others in `tables`, depending on where the literal appeared originally. I also double checked that with this dump, `test_db_migration` starts exercising the Literal deserialization code for all these literal types (the binary literal needed a fix).

> It might be a good idea to also update the insertion code in test_migration.py to test that new values can be properly inserted into the db once it's restored.

Yes, added.

PR #1132, opened 2026-02-09
PXT-985 + PXT-1041 - Adding custom_metadata and comment for columns

christopherpestano:
Pull request adding custom metadata for columns, currently most column metadata fields are non-versioned but since we are in the process of migrating to having most be versioned I have placed the custom_metadata field as a versioned field. One thing we should note is that we still do not have a method for overwriting custom_metadata/comment for either tables or columns, which will be remedied in the next PR of this series which enables updating these fields with versioning enabled.

Note: We are also changing comments in this PR to by default be None instead of an empty string, this is to keep parity with other metadata fields where the default is usually None.

### DIFF
--- a/pixeltable/func/udf.py
+++ b/pixeltable/func/udf.py
@@ -317,7 +317,7 @@ def from_table(tbl: catalog.Table, return_value: 'exprs.Expr' | None, descriptio
     if description is None:
         # Default description is the table comment
         description = tbl._get_comment()
-        if description is None:
+        if description is None or len(description) == 0:
             description = f"UDF for table '{tbl._name}'"
 
     # TODO: Use column comments as parameter descriptions, when we have them


pixeltable/catalog/table.py
```diff
@@ -722,7 +723,7 @@ def _validate_column_spec(cls, name: str, spec: dict[str, Any]) -> None:
         (on account of containing Python Callables or Exprs).
         """
         assert isinstance(spec, dict)
-        valid_keys = {'type', 'value', 'stored', 'media_validation', 'destination'}
+        valid_keys = {'type', 'value', 'stored', 'media_validation', 'destination', 'custom_metadata'}
```
aaron-siegel: Can we get these from the `TypedDict` class now that we have it (once the "other" PR is merged)?

pixeltable/catalog/table.py
```diff
@@ -750,6 +751,15 @@ def _validate_column_spec(cls, name: str, spec: dict[str, Any]) -> None:
         if d is not None and not isinstance(d, (str, Path)):
             raise excs.Error(f'Column {name!r}: `destination` must be a string or path; got {d}')
 
+        if 'custom_metadata' in spec:
+            # we require custom_metadata to be JSON-serializable
+            try:
+                json.dumps(spec['custom_metadata'])
+            except (TypeError, ValueError) as err:
+                raise excs.Error(
+                    f'Column {name!r}: `custom_metadata` must be JSON-serializable; got {spec["custom_metadata"]}'
```
aaron-siegel: Not sure it's wise to stringify `spec['custom_metadata']` into the error message; it could be a large structure?

pixeltable/catalog/column.py
```diff
@@ -86,6 +88,8 @@ def __init__(
         value_expr_dict: dict[str, Any] | None = None,
         tbl_handle: 'TableVersionHandle' | None = None,
         destination: str | None = None,
+        comment: str = '',
```
aaron-siegel: As I mentioned today - I have a preference for `comment: str | None = None` as the API, but I don't feel super strongly about it.

christopherpestano: I think this could be problematic though since then we could have two ways of representing no comment, an empty string and None. We could then disallow empty strings (or implictly convert them to None) but that ultimately diverges from the existing pattern present for Table comment's. We could discuss this more if it's a major point of contention, it is a bit weird only comment has the default null value as something other than None.

christopherpestano: I think this could be problematic though since then we could have two ways of representing no comment, an empty string and None. We could then disallow empty strings (or implictly convert them to None) but that ultimately diverges from the existing pattern present for Table comment's. We could discuss this more if it's a major point of contention, it is a bit weird only comment has the default null value as something other than None.

pixeltable/catalog/table.py
christopherpestano: Changed it to instead of stringifying custom_metadata it instead converts the underlying Type/Value error to a string within the excs.Error as that will tell users specifically what about their JSON is not serializable. Sample error 


`memories = pxt.create_table(
    'agent_demo/memories',
    {
        'content': {'type': pxt.String, 'custom_metadata': {'embedding': set()}},
        'category': pxt.String,  # Optional category (preference, fact, etc.)
        'created_at': pxt.Timestamp,  # When the memory was stored
    },
)`

`Column 'content': `custom_metadata` must be JSON-serializable; got Error: Object of type set is not JSON serializable`

pixeltable/catalog/table.py
christopherpestano: Added

PR #1133, opened 2026-02-11
Fix: Implement drop_index() for BtreeIndex and EmbeddingIndex

KeeProMise:
Fixes a critical bug where PostgreSQL indices were never physically deleted, causing space leaks and metadata inconsistency.

## Problem
- BtreeIndex.drop_index() raised NotImplementedError
- EmbeddingIndex.drop_index() raised NotImplementedError
- store.py:drop_index() incorrectly called sa_create_stmt() instead of dropping the index
- Result: Every "deleted" index remained in PostgreSQL, leaking space

## Impact
Before this fix:
- Calling Table.drop_index() only removed metadata
- PostgreSQL indices were never deleted from disk
- Each deletion leaked disk space
- System state became inconsistent (metadata vs storage)

After this fix:
- Indices are properly removed from PostgreSQL
- Disk space is freed
- Metadata and storage remain consistent
- Operations are idempotent (IF EXISTS)

## Implementation
1. Implemented BtreeIndex.drop_index() using SQLAlchemy DropIndex
2. Implemented EmbeddingIndex.drop_index() using SQLAlchemy DropIndex
3. Fixed store.py to call actual drop_index() method
4. Used IF EXISTS for idempotent operations
5. Added comprehensive tests verifying physical deletion

## Testing
- test_btree_index_drop: Verifies B-tree index physical deletion
- test_embedding_index_drop_physical: Verifies embedding index deletion
- test_drop_index_idempotent: Verifies IF EXISTS behavior

All tests verify both metadata cleanup AND physical index deletion from PostgreSQL.

### DIFF
--- a/pixeltable/index/btree.py
+++ b/pixeltable/index/btree.py
@@ -60,15 +60,6 @@ class BtreeIndex(IndexBase):
         sa_idx = sql.Index(store_index_name, sa_value_col, postgresql_using='btree')
         return sql.schema.CreateIndex(sa_idx, if_not_exists=True).compile(dialect=postgresql.dialect())
 
-    def sa_drop_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql.Compiled:
-        from sqlalchemy.dialects import postgresql
-
-        # Generate DROP INDEX statement
-        # Using IF EXISTS to make the operation idempotent
-        return sql.schema.DropIndex(sql.Index(store_index_name, sa_value_col), if_exists=True).compile(
-            dialect=postgresql.dialect()
-        )
-
     @classmethod
     def display_name(cls) -> str:
         return 'btree'
--- a/pixeltable/index/base.py
+++ b/pixeltable/index/base.py
@@ -36,9 +36,13 @@ class IndexBase(abc.ABC):
     def sa_create_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql.Compiled:
         """Return a sqlalchemy statement for creating the index"""
 
-    @abc.abstractmethod
     def sa_drop_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql.Compiled:
         """Return a sqlalchemy statement for dropping the index"""
+        from sqlalchemy.dialects import postgresql
+
+        return sql.schema.DropIndex(sql.Index(store_index_name, sa_value_col), if_exists=True).compile(
+            dialect=postgresql.dialect()
+        )
 
     @classmethod
     @abc.abstractmethod
--- a/pixeltable/index/embedding_index.py
+++ b/pixeltable/index/embedding_index.py
@@ -178,15 +178,6 @@ class EmbeddingIndex(IndexBase):
         stmt = Env.get().dbms.create_vector_index_stmt(store_index_name, sa_value_col, metric=metric)
         return stmt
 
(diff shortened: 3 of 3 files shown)


KeeProMise:
Hi @aaron-siegel 
Hi, if you have time, please help to take a look at this PR, thanks!

aaron-siegel:
@KeeProMise Thank you for the contribution! The Pixeltable team is looking into this and we'll get back to you shortly with next steps.

sergey-mkhitaryan:
@KeeProMise thanks for your contribution! We reviewed your PR. There were a few improvements and nits that it needed, but we decided that it'll be easier to just do that ourselves as opposed to requesting changes, so I'm going to push those updates now.

pixeltable/store.py
```diff
@@ -257,7 +257,8 @@ def create_index(self, idx_id: int) -> None:
     def drop_index(self, idx_id: int) -> None:
         """Drop If Exists for this index"""
         idx_info = self.tbl_version.get().idxs[idx_id]
-        stmt = idx_info.idx.sa_create_stmt(self.tbl_version.get()._store_idx_name(idx_id), idx_info.val_col.sa_col)
+        store_index_name = self.tbl_version.get()._store_idx_name(idx_id)
+        stmt = idx_info.idx.sa_drop_stmt(store_index_name, idx_info.val_col.sa_col)
```
mkornacker, line 261: I'm just noticing that this runs _exec_if_not_exists(), which is wrong here (it would be "if exists").

pixeltable/index/btree.py
```diff
@@ -60,10 +60,14 @@ def sa_create_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql
         sa_idx = sql.Index(store_index_name, sa_value_col, postgresql_using='btree')
         return sql.schema.CreateIndex(sa_idx, if_not_exists=True).compile(dialect=postgresql.dialect())
 
-    def drop_index(self, index_name: str, index_value_col: 'catalog.Column') -> None:
-        """Drop the index on the index value column"""
-        # TODO: implement
-        raise NotImplementedError()
+    def sa_drop_stmt(self, store_index_name: str, sa_value_col: sql.Column) -> sql.Compiled:
+        from sqlalchemy.dialects import postgresql
```
mkornacker: This should be the generic implementation in the base class.

tests/test_index.py
```diff
@@ -981,3 +982,41 @@ def test_embedding_index_precision(
...
+
+        # Drop it
+        if index_type == 'btree':
+            t.drop_index(column='text')
+        else:
+            t.drop_embedding_index(idx_name='text_idx')
+
+        # Verify index no longer exists in the store
+        assert store_idx_name not in list_store_indexes(t), f'Index {store_idx_name} should not exist after drop'
+        # Or the metadata
```
mkornacker, line 1018: Use the public API for that purpose (running a similarity() query should return an error).

pixeltable/store.py
sergey-mkhitaryan: rewrote. ptal

tests/test_index.py
sergey-mkhitaryan: Yeah, but `similarity()` would return an error even without this change. The point here is to verify that the store no longer has an index, not just that the index metadata doesn't exist. I don't know if that's possible to do in a blackbox manner.

PR #1138, opened 2026-02-13
[PXT-1002] Fix table md caching when an insert finalizes view creation

sergey-mkhitaryan:
The scenario that this fixes is:
1. There exists a base table
2. A view is added, but its pending ops are not finalized and its state is ROLLFORWARD.
3. An insert into the base table is executed. The insert finalizes the view's pending ops, and executes the insert.

Without this fix, the TV cache is not cleared after pending ops are finalized, and the insert blindly overwrites the view's state from LIVE back to ROLLFORWARD, which, in the absense of pending ops, is an invalid state.

### DIFF
--- a/tests/test_iterator.py
+++ b/tests/test_iterator.py
@@ -10,7 +10,7 @@ import pixeltable as pxt
 import pixeltable.functions as pxtf
 from pixeltable import func, type_system as ts
 from pixeltable.iterators.base import ComponentIterator
-from tests.utils import ReloadTester, reload_catalog
+from tests.utils import reload_catalog
 
 
 class MyRow(TypedDict):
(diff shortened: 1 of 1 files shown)


pixeltable/catalog/catalog.py
mkornacker, line 637: Also update this comment, invalidation of cached md is part of the function's "contract".

pixeltable/catalog/catalog.py
```diff
@@ -754,6 +751,7 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
                         self.mark_modified_tvs(tv.handle)
 
                         if is_final_op:
+                            self._clear_tv_cache(TableVersionKey(tbl_id, None, None))
```
mkornacker: Should this happen in an enclosing try finally block? We want to clear the cache on every exit path, no?

pixeltable/catalog/catalog.py
```diff
@@ -1514,6 +1514,7 @@ def update_additional_md(self, tbl_id: UUID, additional_md: dict[str, Any]) -> N
             .where(schema.Table.id == str(tbl_id))
             .values({schema.Table.additional_md: schema.Table.additional_md.op('||')(additional_md)})
         )
+        self._clear_tv_cache(TableVersionKey(tbl_id, None, None))
```
mkornacker, line 1505: Also add this to the class docstring (everything that writes table metadata needs to clear the cache).

pixeltable/catalog/catalog.py
sergey-mkhitaryan: I guess that's true, we are unlikely to enter this method and not end up needing to invalidate the cache.

PR #1139, opened 2026-02-13
[PXT-1002] Make non-transactional TableOps idempotent

sergey-mkhitaryan:
Otherwise they can fail when run concurrently


pixeltable/store.py
```diff
@@ -290,16 +284,20 @@ def _add_column_stmt(self, sa_col: sql.Column) -> str:
...
         )
 
-    def add_column(self, col: catalog.Column) -> None:
+    def add_column(self, col: catalog.Column, if_not_exists: bool) -> None:
         """Add column(s) to the store-resident table based on a catalog column"""
         assert col.is_stored
         conn = Env.get().conn
         col_type_str = col.sa_col_type.compile(dialect=conn.dialect)
-        s_txt = f'ALTER TABLE {self._storage_name()} ADD COLUMN {col.store_name()} {col_type_str} NULL'
+        if_not_exists_operator = 'IF NOT EXISTS' if if_not_exists else ''
```
mkornacker: Should be _clause, not _operator.

pixeltable/catalog/table_version.py
```diff
@@ -468,9 +468,8 @@ def undo_op(self, op: TableOp) -> None:
                 self.store_tbl.drop()
 
         elif isinstance(op, CreateStoreIdxsOp):
-            for idx_id in op.idx_ids:
-                with Env.get().begin_xact():
-                    self.store_tbl.drop_index(idx_id)
+            # Indexes are dropped when the table is dropped
```
aaron-siegel: What if the operation we're undoing is something other than the table create? Won't that cause orphaned indices to remain in the store?

sergey-mkhitaryan: Good point. I updated my comment, please take a look. Note that `self.store_tbl.drop_index(idx_id)` that I'm dropping in this pr _creates_ indexes instead of dropping them, which makes even less sense to me than doing nothing.

pixeltable/catalog/tbl_ops.py
```diff
@@ -22,6 +22,19 @@ class OpStatus(Enum):
 
 @dataclasses.dataclass
 class TableOp:
+    """TableOp describes an isolated operation that needs to be performed on the table, typically as part of a schema
+    change, in order to unblock further user operations on the table.
+
+    TableOps are executed sequentially in order when the table state is ROLLFORWARD. If something goes wrong or the user
```
mkornacker: This is unrelated to the TableOp semantics itself and belongs in Catalog (_finalize_pending_ops() already describes that).

pixeltable/catalog/tbl_ops.py
```diff
@@ -22,6 +22,19 @@ class OpStatus(Enum):
 
 @dataclasses.dataclass
 class TableOp:
+    """TableOp describes an isolated operation that needs to be performed on the table, typically as part of a schema
```
mkornacker: Stick to the TableOp semantics here ("... to unblock .." has nothing to do with those):

Also, what does "isolated" mean? From what?

pixeltable/catalog/tbl_ops.py
```diff
@@ -22,6 +22,19 @@ class OpStatus(Enum):
...
+    change, in order to unblock further user operations on the table.
+
+    TableOps are executed sequentially in order when the table state is ROLLFORWARD. If something goes wrong or the user
+    aborts, the table's state is changed to ROLLBACK, and the already completed TableOps are undone sequentially in
+    the reverse order.
+
+    If needs_xact is True, the op is executed, and its state is updated as part of a single store transaction.
```
mkornacker: These are the TableOp semantics.

mkornacker (changes requested):

aaron-siegel (approved):

pixeltable/catalog/tbl_ops.py
```diff
@@ -22,6 +22,15 @@ class OpStatus(Enum):
 
 @dataclasses.dataclass
 class TableOp:
+    """TableOp describes an individual operation that needs to be performed on the table, typically as part of a schema
```
mkornacker: Remove the schema change part, that's not really true (it just happens to describe current reality).

mkornacker (approved):

PR #1142, opened 2026-02-15
[PXT-1040] disable all twelvelabs tests

sergey-mkhitaryan:
They are all affected by the rate limiting issue


christopherpestano:
I think this is not blocking the merge queue no? I was able to merge with only the notebook tests disabled

aaron-siegel (approved):

PR #1143, opened 2026-02-15
Add vLLM integration for local LLM inference

pierrebrunelle:
Adds a new vLLM integration for Pixeltable https://github.com/vllm-project/vllm. This follows the same pattern as the existing `llama_cpp` integration.


pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,278 @@
...
+            'model': '...',
+            'usage': {
+                'prompt_tokens': ...,
+                'completion_tokens': ...,
+                'total_tokens': ...,
+            },
+        }
+        ```
+
+    Example:
```
aaron-siegel: Please use standard mkdocs formatting for Python code examples.

```
>>> x = 1
... x += 1
```
etc. Please follow the existing patterns used elsewhere in code example docstrings (focused examples with no import statements, no table setup code)

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,278 @@
...
+            )
+        )
+        ```
+    """
+    Env.get().require_package('vllm')
+
+    if model_kwargs is None:
+        model_kwargs = {}
+
+    import vllm as vllm_lib
```
aaron-siegel: Move to top of function. Do a straight import please (`import vllm`, not `import vllm as vllm_lib`) unless I'm missing some compelling reason not to.

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,278 @@
...
+
+    # Separate LLM constructor kwargs from SamplingParams kwargs
```
aaron-siegel: Why? Why not just take two separate kwargs as input? That seems a lot cleaner to me.

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,278 @@
...
+    kwargs_key = tuple(sorted(engine_kwargs.items())) if engine_kwargs else ()
+    key = (model, kwargs_key)
+    if key not in _model_cache:
+        llm = vllm_lib.LLM(model=model, **engine_kwargs)
+        _model_cache[key] = llm
+    return _model_cache[key]
+
+
+def _format_chat_response(output: Any, model: str) -> dict:
+    """Format a vLLM RequestOutput into an OpenAI-compatible chat completion response."""
```
aaron-siegel: Why? Why wouldn't we return the results in vLLM's native format? That's what we do in all our other interfaces (not all of them follow OpenAI and we ordinarily don't mess with the output - we use the native response format - what's the reason to do it differently here?)

tests/functions/test_vllm.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+
+        validate_update_status(t.insert(input='What are some edible species of fish?'), expected_rows=1)
+        result = t.collect()['output'][0]
+        result2 = t.collect()['output2'][0]
+        print(result)
+        assert len(result['choices'][0]['message']['content']) > 0
+        assert len(result2['choices'][0]['message']['content']) > 0
+
+        vllm.cleanup()  # Clean up the model cache after the test
```
aaron-siegel: This isn't the right place to do this - if the test raises an exception then it will be skipped. Ok, I see that llama_cpp is also doing this; it probably shouldn't be, but let's worry about that separately.

Anyway he llama_cpp thing is really a special case; vLLM is more analogous to Hugging Face; this isn't really a cleanup operation but a cache clear. I'd say get rid of cleanup( ) and follow the pattern used for Hugging Face in conftest.py::pytest_runtest_setup().

pyproject.toml
```diff
@@ -194,7 +194,8 @@ test = [
 ]
 extra-dev = [
     # This is for dependencies that will be installed locally for devs (via `make install`), but not in CI.
-    "llama-cpp-python>=0.3.16 ; sys_platform != 'win32'"
+    "llama-cpp-python>=0.3.16 ; sys_platform != 'win32'",
+    "vllm>=0.6.0 ; sys_platform == 'linux'"
```
aaron-siegel: Why not run in CI?

tests/functions/test_vllm.py
```diff
@@ -0,0 +1,61 @@
+import pytest
+
+import pixeltable as pxt
+
+from ..utils import skip_test_if_not_installed, validate_update_status
+
+
+@pytest.mark.expensive
```
aaron-siegel, line 15: Why?

aaron-siegel (changes requested):

pixeltable/functions/vllm.py
pierrebrunelle: Done — switched to standard `>>>` mkdocs format with focused examples (no imports or table setup). Matches the pattern used in `huggingface.py`, `whisper.py`, etc.

pixeltable/functions/vllm.py
pierrebrunelle: Done — changed to straight `import vllm` at the top of the function body (right after `require_package`). The `as vllm_lib` alias was to avoid shadowing the module-level TYPE_CHECKING import, but it's unnecessary since the local import naturally takes precedence.

pixeltable/functions/vllm.py
pierrebrunelle: split into two explicit parameters: `engine_kwargs` (for `LLM` constructor args like `max_model_len`, `dtype`) and `sampling_params` (for `SamplingParams` args like `max_tokens`, `temperature`). Removed the introspection-based auto-routing logic entirely.

pixeltable/functions/vllm.py
pierrebrunelle: Good point — now returning vLLM's native `RequestOutput` as a dict (preserving fields like `request_id`, `prompt`, `prompt_token_ids`, `outputs`, `finished`). Removed the OpenAI-compatible reformatting helpers. Users access generated text via `result.outputs[0].text`.

tests/functions/test_vllm.py
pierrebrunelle: Done — removed `cleanup()` from vllm module. Cache clearing now happens in `conftest.py::pytest_runtest_setup()` via `pxtf.vllm._model_cache.clear()`, matching the HuggingFace pattern.

pyproject.toml
pierrebrunelle: moved `vllm` from `extra-dev` to `test` dependencies so it gets installed in CI (Linux only via `sys_platform == 'linux'` marker).

tests/functions/test_vllm.py
pierrebrunelle: no longer marked `@pytest.mark.expensive`. Since vllm is now in `test` deps, these tests will run in CI (on Linux runners with GPU support).

pierrebrunelle:
Blocked on #1189

tests/functions/test_vllm.py
```diff
@@ -0,0 +1,61 @@
...
+class TestVllm:
+    def test_chat_completions(self, uses_db: None) -> None:
+        skip_test_if_not_installed('vllm')
+        from pixeltable.functions import vllm
```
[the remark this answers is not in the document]
pierrebrunelle: Good catch — added @rerun(reruns=3, reruns_delay=15) to TestVllm, matching the pattern in test_llama_cpp.py. Fixed in 74b022b.

docs/release/howto/providers/working-with-vllm.ipynb
```diff
@@ -0,0 +1,568 @@
...
+      "metadata": {
+        "vscode": {
+          "languageId": "raw"
+        }
+      },
+      "source": [
+        "---\n",
+        "title: \"Working with vLLM\"\n",
+        "icon: \"notebook\"\n",
+        "description: \"[Open in Kaggle](https://kaggle.com/kernels/welcome?src=https://github.com/pixeltable/pixeltable/blob/release/docs/notebooks/howto/providers/working-with-vllm.ipynb) | [Open in Colab](https://colab.research.google.com/github/pixeltable/pixeltable/blob/release/docs/notebooks/howto/providers/working-with-vllm.ipynb) | [View on GitHub](https://github.com/pixeltable/pixeltable/blob/release/docs/notebooks/howto/providers/working-with-vllm.ipynb)\"\n",
```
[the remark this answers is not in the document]
pierrebrunelle: Good catch — updated the frontmatter links to use docs/release/howto/providers/ to match the actual file location (consistent with working-with-fabric.ipynb). Fixed in 74b022b.

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,180 @@
...
+        ...         model='Qwen/Qwen2.5-0.5B-Instruct',
+        ...         sampling_params={'max_tokens': 256, 'temperature': 0.7},
+        ...     )
+        ... )
+    """
+    Env.get().require_package('vllm')
+    import vllm
+
+    llm = _lookup_model(model, engine_kwargs or {})
+    sp = vllm.SamplingParams(**(sampling_params or {})) if sampling_params else None
```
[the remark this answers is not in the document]
pierrebrunelle: Good point — added min_version=[0, 6, 0] to both require_package('vllm') calls, matching the pattern in llama_cpp.py. Fixed in 74b022b.

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def _lookup_model(model: str, engine_kwargs: dict[str, Any]) -> 'vllm.LLM':
+    import vllm
+
+    kwargs_key = tuple(sorted(engine_kwargs.items())) if engine_kwargs else ()
+    key = (model, kwargs_key)
+    if key not in _model_cache:
+        _model_cache[key] = vllm.LLM(model=model, **engine_kwargs)
+    return _model_cache[key]
```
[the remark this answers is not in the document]
pierrebrunelle: Fair point — switched the cache key from tuple(sorted(items())) to json.dumps(engine_kwargs, sort_keys=True, default=str), which handles unhashable values (lists, nested dicts) gracefully. Fixed in 74b022b.

tests/functions/test_vllm.py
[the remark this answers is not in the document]
pierrebrunelle: Already covered by the combination of existing guards:

1. `@pytest.mark.expensive` — skipped in standard CI runs
2. `skip_test_if_not_installed('vllm')` — skips if vllm isn't present
3. `vllm` is declared with `sys_platform == 'linux'` in `pyproject.toml`, so it won't be installed on non-Linux systems

If someone manually installs vllm on a machine without a GPU, vLLM itself raises a clear `RuntimeError` at engine init — adding a separate GPU probe would add complexity for a scenario that's already handled well by the library's own error messaging.

pierrebrunelle:
@aaron-siegel no rush but when you have a chance since blocker has been removed.

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,181 @@
...
+        ...         model='Qwen/Qwen2.5-0.5B-Instruct',
+        ...         sampling_params={'max_tokens': 256, 'temperature': 0.7},
+        ...     )
+        ... )
+    """
+    Env.get().require_package('vllm', min_version=[0, 6, 0])
+    import vllm
+
+    llm = _lookup_model(model, engine_kwargs or {})
+    sp = vllm.SamplingParams(**(sampling_params or {})) if sampling_params else None
```
aaron-siegel: Usually we spell out `if sampling_params is not None`.

Also, what's the point in saying `sampling_params or {}`, if that clause will only be invoked when `sampling_params` is truthy?

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,181 @@
...
+if TYPE_CHECKING:
+    import vllm
+
+
+@pxt.udf(is_deterministic=False)
+def chat_completions(
+    messages: list[dict],
+    *,
+    model: str,
+    engine_kwargs: dict[str, Any] | None = None,
```
aaron-siegel: Wouldn't this be engine_args, if we're going with vLLM's nomenclature?

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,181 @@
...
+
+    - `pip install vllm`
+
+    Args:
+        messages: A list of messages to generate a response for. Each message should be a dict
+            with `role` and `content` keys, following the OpenAI chat format.
+        model: The HuggingFace model identifier (e.g., `'Qwen/Qwen2.5-0.5B-Instruct'`).
+        engine_kwargs: Additional keyword args for the vLLM `LLM` constructor, such as `dtype`,
+            `max_model_len`, `gpu_memory_utilization`, `tensor_parallel_size`. For details, see the
+            [vLLM engine args documentation](https://docs.vllm.ai/en/stable/serving/engine_args.html).
```
aaron-siegel: URL gives a 404 error

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,181 @@
...
+        sampling_params: Keyword args for vLLM `SamplingParams`, such as `max_tokens`,
+            `temperature`, `top_p`, `top_k`. For details, see the
+            [vLLM sampling params documentation](https://docs.vllm.ai/en/stable/dev/sampling_params.html).
```
aaron-siegel: URL redirects to a non-relevant page

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,181 @@
...
+    import vllm
+
+    llm = _lookup_model(model, engine_kwargs or {})
+    sp = vllm.SamplingParams(**(sampling_params or {})) if sampling_params else None
+
+    chat_kwargs: dict[str, Any] = {'use_tqdm': False}
+    if sp is not None:
+        chat_kwargs['sampling_params'] = sp
+
+    chat_messages: Any = [messages]
```
aaron-siegel: why is the type hint `Any`?

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,181 @@
...
+    import vllm
+
+    kwargs_key = json.dumps(engine_kwargs, sort_keys=True, default=str) if engine_kwargs else ''
+    key = (model, kwargs_key)
+    if key not in _model_cache:
+        _model_cache[key] = vllm.LLM(model=model, **engine_kwargs)
+    return _model_cache[key]
+
+
+def _request_output_to_dict(output: Any) -> dict:
```
aaron-siegel: Can we please try to be more precise with the type hints.

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,181 @@
...
+                'finish_reason': o.finish_reason,
+                'stop_reason': o.stop_reason,
+            }
+            for o in output.outputs
+        ],
+        'finished': output.finished,
+    }
+
+
+_model_cache: dict[tuple, 'vllm.LLM'] = {}
```
aaron-siegel: dict[tuple[str, str], 'vllm.LLM']

scripts/prepare-nb-tests.sh
```diff
@@ -50,6 +50,7 @@ rm "$TARGET_DIR/video-generate-ai.ipynb"  # Expensive
 rm "$TARGET_DIR/img-image-to-image.ipynb"  # Expensive (downloads ~5GB model)
 rm "$TARGET_DIR/llm-tool-calling.ipynb"  # Flaky
 rm -f "$TARGET_DIR/working-with-fabric.ipynb"  # Requires Microsoft Fabric environment
+rm -f "$TARGET_DIR/working-with-vllm.ipynb"  # Requires GPU (vLLM)
```
aaron-siegel: Are we sure about this?

Why do we need to exclude the notebook here, if the unit tests run on CPU configurations in CI?

tests/functions/test_vllm.py
```diff
@@ -0,0 +1,62 @@
+import pytest
+
+import pixeltable as pxt
+
+from ..utils import rerun, skip_test_if_not_installed, validate_update_status
+
+
+@pytest.mark.expensive
```
aaron-siegel, line 15: It's still marked pytest.mark.expensive.

pyproject.toml
```diff
@@ -194,7 +194,8 @@ test = [
 ]
 extra-dev = [
     # This is for dependencies that will be installed locally for devs (via `make install`), but not in CI.
-    "llama-cpp-python>=0.3.16 ; sys_platform != 'win32'"
+    "llama-cpp-python>=0.3.16 ; sys_platform != 'win32'",
+    "vllm>=0.6.0 ; sys_platform == 'linux'",
```
aaron-siegel: Why are we excluding it from CI?

pyproject.toml
```diff
@@ -247,6 +248,10 @@ strict_optional = false
 warn_return_any = true
 warn_unused_ignores = true
 
+[[tool.mypy.overrides]]
```
aaron-siegel, line 268: Please don't add this here - it will mask future changes to vLLM (e.g. if they provide type hints in the future). Use appropriate `# type: ignore` directives instead (that's what we do everywhere else in the codebase, which is why pyproject doesn't have a `tool.mypy.overrides` section).

aaron-siegel (changes requested):

pixeltable/functions/vllm.py
pierrebrunelle: Fixed — changed to `if sampling_params is not None` and removed the redundant `or {}`.

pixeltable/functions/vllm.py
pierrebrunelle: Good catch — renamed to `engine_args` throughout (UDFs, tests, notebook).

pixeltable/functions/vllm.py
pierrebrunelle: Fixed — updated to https://docs.vllm.ai/en/stable/configuration/engine_args/

pixeltable/functions/vllm.py
pierrebrunelle: Fixed — updated to https://docs.vllm.ai/en/stable/api/vllm/sampling_params.html

pixeltable/functions/vllm.py
pierrebrunelle: Removed the `Any` workaround. This was there to avoid a "unused type: ignore" warning when vllm wasn't installed (types resolved to Any). Now that vllm is in `test` deps (installed in CI), we can use a straightforward `# type: ignore[arg-type]` on the `llm.chat()` call — the underlying issue is list invariance in vLLM's type stubs.

pixeltable/functions/vllm.py
pierrebrunelle: Fixed — `_request_output_to_dict` now takes `'vllm.RequestOutput'` (forward ref) instead of `Any`.

pixeltable/functions/vllm.py
pierrebrunelle: Done — `dict[tuple[str, str], 'vllm.LLM']`.

scripts/prepare-nb-tests.sh
pierrebrunelle: Keeping this — the notebook does actual vLLM inference (loads model, runs generation), which requires a CUDA GPU. Unlike API-key-gated notebooks that the script handles automatically, there's no env var to check. The unit tests now skip gracefully via `_skip_if_no_gpu()`, but the notebook doesn't have that mechanism.

tests/functions/test_vllm.py
pierrebrunelle: Removed `@pytest.mark.expensive`. Tests now use `_skip_if_no_gpu()` which calls `pytest.importorskip('torch')` + checks `torch.cuda.is_available()`, so they'll be discovered in all CI runs but skip gracefully on CPU-only runners.

pyproject.toml
pierrebrunelle: Moved vllm to `test` deps so it's installed in CI. Tests skip gracefully on CPU runners via `_skip_if_no_gpu()` (`torch.cuda.is_available()` check). This way mypy also sees the vllm stubs, so we don't need the `[[tool.mypy.overrides]]` workaround.

pyproject.toml
pierrebrunelle: Removed — now using `# type: ignore[arg-type]` on the `llm.chat()` call (the only place that needs it). Since vllm is now in `test` deps, mypy sees the real stubs and the ignore is properly targeted.

pyproject.toml
pierrebrunelle: Update: moved vllm out of pyproject.toml entirely. Adding it to any dependency group (test or extra-dev) caused uv's universal lockfile resolution to expand platform markers for shared packages like torch/timm, which broke `test_detr_for_segmentation` in CI (timm no longer found by `importlib.util.find_spec`). vllm is now a manual `pip install vllm` for developers with GPU access. Tests skip gracefully via `skip_test_if_not_installed('vllm')` + `_skip_if_no_gpu()`.

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,176 @@
...
+
+import json
+from typing import TYPE_CHECKING, Any
+
+import pixeltable as pxt
+from pixeltable.env import Env
+from pixeltable.utils.code import local_public_names
+
+if TYPE_CHECKING:
+    import vllm  # type: ignore[import-not-found,unused-ignore]
```
aaron-siegel, line 16: `# type: ignore[unused-ignore]`? why?

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,176 @@
...
+    import vllm
+
+    llm = _lookup_model(model, engine_args or {})
+    sp = vllm.SamplingParams(**sampling_params) if sampling_params is not None else None
+
+    chat_kwargs: dict[str, Any] = {'use_tqdm': False}
+    if sp is not None:
+        chat_kwargs['sampling_params'] = sp
+
+    outputs = llm.chat([messages], **chat_kwargs)  # type: ignore[arg-type,unused-ignore]
```
aaron-siegel, line 83: Again. They shouldn't be necessary (never seen this one before).

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,176 @@
...
+        gen_kwargs['sampling_params'] = sp
+
+    outputs = llm.generate([prompt], **gen_kwargs)
+    return _request_output_to_dict(outputs[0])
+
+
+def _lookup_model(model: str, engine_args: dict[str, Any]) -> 'vllm.LLM':
+    import vllm
+
+    args_key = json.dumps(engine_args, sort_keys=True, default=str) if engine_args else ''
```
aaron-siegel, line 141: Why is `if engine_args else ''` necessary?

pixeltable/functions/vllm.py
pierrebrunelle: Fixed — removed all inline `type: ignore` directives. Added `[[tool.mypy.overrides]]` for `vllm.*` in pyproject.toml instead (standard mypy mechanism for optional deps that cannot be in the lockfile).

pixeltable/functions/vllm.py
pierrebrunelle: Fixed — removed entirely. With the mypy override, vllm types resolve cleanly whether or not the package is installed, so no `type: ignore` is needed on this line at all.

pixeltable/functions/vllm.py
pierrebrunelle: Good catch — removed the conditional. `json.dumps({})` produces `'{}'` which works fine as a cache key.

PR #1146, opened 2026-02-16
Add Bedrock API Key auth support and notebook outputs

pierrebrunelle:
### Bedrock API Key (bearer token) authentication
- Updated `register_client('bedrock')` to accept optional `api_key` and `region_name` parameters
- When `BEDROCK_API_KEY` env var is set, uses bearer token auth (injected via botocore `before-send` event handler) instead of SigV4 signing
- Falls back to standard boto3 credential chain when no API key is provided
- Added `region_name` to `KNOWN_CONFIG_OPTIONS` for Bedrock (`BEDROCK_REGION_NAME`, defaults to `us-east-1`)

### Notebook outputs
- Re-ran notebooks to generate outputs for docs: bedrock, fal, openrouter, fiftyone


pierrebrunelle:
@claude review this

PR #1149, opened 2026-02-17
dumps() udf for json

mkornacker:


pixeltable/functions/json.py
```diff
@@ -11,12 +11,29 @@
...
 """
 
+import json
 from typing import Any
 
 import pixeltable as pxt
 from pixeltable.utils.code import local_public_names
 
 
+@pxt.udf
```
aaron-siegel, line 23: Is there a `to_sql()` equivalent?

pixeltable/functions/json.py
```diff
@@ -11,12 +11,29 @@
...
+
+    Equivalent to [`json.dumps()`](https://docs.python.org/3/library/json.html#json.dumps).
+
+    Args:
+        obj: A JSON-serializable object (dict, list, or scalar).
+
+    Returns:
+        A JSON-formatted string.
+    """
+    return json.dumps(obj)
```
aaron-siegel, line 36: What if `obj` contains images or arrays? We just surface the `TypeError`? Should we be wrapping it in a `pxt.Error`?

mkornacker: We generally record udf exceptions as-is, since they get surfaced via the errortype property.

Should we do it differently here?

pixeltable/functions/json.py
mkornacker: Done.

PR #1150, opened 2026-02-18
PXT-877 Fixing if_exists='replace' cannot be used to replace a Table with a View/Snapshot or vice-versa

christopherpestano:
**`if_exists='replace'` / `'replace_force'`**: Previously, replacing a table with a view (or vice-versa) raised an error. This PR removes that restriction, all table subtypes (table, view, snapshot) can now replace each other freely. Dirs still cannot replace table subtypes and vice-versa.

**`if_exists='ignore'`**: The ignore path now also enforces type compatibility. If the existing object is a different subtype (e.g., a table exists but you're calling `create_view`), it raises an error rather than silently returning the wrong object type. Previously this check was missing.

## Repro for the `replace` fix

```python
t = pxt.create_table('a1', {'icol': pxt.Int}, if_exists='replace_force')
v = pxt.create_view('b1', t, if_exists='replace')
t2 = pxt.create_table('b1', {'scol': pxt.String}, if_exists='replace')
```

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -2601,7 +2601,7 @@ class Catalog:
                     raise excs.Error(f'Path {path!r} already exists as a {obj_type_str} with a different base table')
             return obj
 
-        # IfExistsParam.REPLACE or IfExistsParam.REPLACE_FORCE: Drop existing object if it matches expected type (and base table for views/snapshots)
+        # IfExistsParam.REPLACE or IfExistsParam.REPLACE_FORCE
         assert if_exists in (IfExistsParam.REPLACE, IfExistsParam.REPLACE_FORCE)
 
         # check to ensure that dirs can only be replaced with dirs, and all table subtypes can replace each other
--- a/tests/test_dirs.py
+++ b/tests/test_dirs.py
@@ -75,7 +75,7 @@ class TestDirs:
 
         # invalid if_exists value is rejected
         with pytest.raises(
-            pxt.Error, match="if_exists must be one of: \['error', 'ignore', 'replace', 'replace_force'\]"
+            pxt.Error, match=r"if_exists must be one of: \['error', 'ignore', 'replace', 'replace_force'\]"
         ):
             pxt.create_dir('dir1', if_exists='invalid')  # type: ignore[arg-type]
 
--- a/tests/test_view.py
+++ b/tests/test_view.py
@@ -161,7 +161,7 @@ class TestView:
 
         # invalid if_exists value is rejected
         with pytest.raises(
-            pxt.Error, match="if_exists must be one of: \['error', 'ignore', 'replace', 'replace_force'\]"
+            pxt.Error, match=r"if_exists must be one of: \['error', 'ignore', 'replace', 'replace_force'\]"
         ):
             _ = pxt.create_view('test_view', t, if_exists='invalid')  # type: ignore[arg-type]
 
--- a/tests/test_snapshot.py
+++ b/tests/test_snapshot.py
@@ -118,7 +118,7 @@ class TestSnapshot:
         id_before = s._id
         # invalid if_exists value is rejected
         with pytest.raises(
-            pxt.Error, match="if_exists must be one of: \['error', 'ignore', 'replace', 'replace_force'\]"
+            pxt.Error, match=r"if_exists must be one of: \['error', 'ignore', 'replace', 'replace_force'\]"
         ):
             pxt.create_snapshot(sname, t, if_exists='invalid')  # type: ignore[arg-type]
         # scenario 1: a snapshot exists at the path already
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -167,7 +167,7 @@ class TestTable:
 
         # invalid if_exists value is rejected
         with pytest.raises(
-            pxt.Error, match="if_exists must be one of: \['error', 'ignore', 'replace', 'replace_force'\]"
+            pxt.Error, match=r"if_exists must be one of: \['error', 'ignore', 'replace', 'replace_force'\]"
         ):
             pxt.create_table('test', schema, if_exists='invalid')  # type: ignore[arg-type]
 
@@ -230,7 +230,7 @@ class TestTable:
             _ = pxt.create_table('dir1', schema)
         assert len(tbl.select().collect()) == 1
         for ie in ('ignore', 'replace', 'replace_force'):
-            with pytest.raises(pxt.Error, match='already exists') as exc_info:
+            with pytest.raises(pxt.Error, match='already exists'):
                 pxt.create_table('dir1', schema, if_exists=ie)
             assert len(tbl.select().collect()) == 1, f'with if_exists={ie}'
             assert 'dir1' in pxt.list_dirs(), f'with if_exists={ie}'


tests/test_catalog.py
```diff
@@ -176,3 +176,66 @@ def test_ls(self, uses_db: None) -> None:
...
+        assert 'target' in pxt.list_tables()
+        view2 = pxt.create_view('target', base_table, if_exists='replace')
+        assert 'target' in pxt.list_tables()
+        assert view2.column_names() == ['c1']  # inherits from base_table
+
+        # View can replace snapshot
+        pxt.create_snapshot('target', base_table, if_exists='replace')
+        assert 'target' in pxt.list_tables()
+        pxt.create_view('target', base_table, if_exists='replace')
+        assert 'target' in pxt.list_tables()
```
sergey-mkhitaryan: here and below, why no longer check column names?

tests/test_catalog.py
```diff
@@ -176,3 +176,66 @@ def test_ls(self, uses_db: None) -> None:
...
+        pxt.drop_table('target')
+        pxt.create_dir('target')
+        with pytest.raises(excs.Error, match='expected a table, view or snapshot'):
+            pxt.create_table('target', {'c1': pxt.Int}, if_exists='replace')
+        with pytest.raises(excs.Error, match='expected a table, view or snapshot'):
+            pxt.create_view('target', base_table, if_exists='replace')
+        with pytest.raises(excs.Error, match='expected a table, view or snapshot'):
+            pxt.create_snapshot('target', base_table, if_exists='replace')
+
+        # And table subtypes cannot replace dirs
```
sergey-mkhitaryan: a dir cannot replace a table

christopherpestano: Fixed comment

tests/test_catalog.py
christopherpestano: Added additional columns so there are distinct column names to check

pixeltable/catalog/catalog.py
```diff
@@ -2570,24 +2570,29 @@ def _handle_path_collision(
 
         if if_exists == IfExistsParam.ERROR and obj is not None:
             raise excs.Error(f'Path {path!r} is an existing {obj._display_name()}')
-        else:
-            is_snapshot = isinstance(obj, View) and obj._tbl_version_path.is_snapshot()
-            if obj is not None and (not isinstance(obj, expected_obj_type) or (expected_snapshot and not is_snapshot)):
+
+        # dirs can only collide with dirs, all table subtypes can collide with eachother
```
mkornacker: Unclear comment (what's the meaning of "can collide"?). Phrase in terms of conflict?

tests/test_catalog.py
```diff
@@ -176,3 +176,69 @@ def test_ls(self, uses_db: None) -> None:
...
+
+    def test_cross_type_replacement(self, uses_db: None) -> None:
+        """Test that tables, views, and snapshots can replace each other with if_exists='replace'.
+
+        This tests the path collision handling logic: dirs can only collide with dirs,
+        but all table subtypes (table, view, snapshot) can collide with each other.
+        """
+        base_table = pxt.create_table('base', {'c1': pxt.Int})
+
+        # Table can replace view
```
mkornacker: Can you express these cases more succinctly with 3 lambdas (per create_x(...)) and two nested loops over those lambdas? It looks very repetitive.

tests/test_view.py
```diff
@@ -201,15 +201,15 @@ def test_create_if_exists(self, uses_db: None, reload_tester: ReloadTester) -> N
         assert 'test_view_on_view' not in pxt.list_tables()
 
         # scenario 3: path exists but is not a view
-        _ = pxt.create_table('not_view', {'c1': pxt.String})
-        with pytest.raises(pxt.Error, match='is an existing table'):
+        _ = pxt.create_dir('not_view')
```
mkornacker: Why change this? This should still return an error w/ create_table()?

pixeltable/catalog/catalog.py
christopherpestano: Changed comment to instead be "can be replaced" for clarity on how the dirs/tables interact with eachother in this situation.

tests/test_view.py
christopherpestano: In this case I had wanted to preserve the original logic, of having all three of replace/replace_force/ignore failing, but thinking further it makes more sense to reflect the logic change in this pr by having replace/replace_force now succeed.

tests/test_catalog.py
christopherpestano: Makes sense, replaced it with the three lambdas and made it such that we have two nested loops for this, with the actual column checking being an additional value put with the lambda. I used a dict here instead of a list for readability, though since we never use the actual keys a list could also work. Also replaced the replace dir failing check with the lambdas as well for consistency.

pixeltable/catalog/catalog.py
```diff
@@ -2573,24 +2573,36 @@ def _handle_path_collision(
...
         if if_exists == IfExistsParam.ERROR and obj is not None:
             raise excs.Error(f'Path {path!r} is an existing {obj._display_name()}')
-        else:
-            is_snapshot = isinstance(obj, View) and obj._tbl_version_path.is_snapshot()
-            if obj is not None and (not isinstance(obj, expected_obj_type) or (expected_snapshot and not is_snapshot)):
+
+        # dirs can only be replaced with dirs, all table subtypes can replace eachother
+        if obj is not None:
+            if expected_obj_type == Dir and not isinstance(obj, Dir):
+                raise excs.Error(f'Path {path!r} is an existing {obj._display_name()}, expected a directory')
```
aaron-siegel: Nit: comma splice (comma should be a semicolon); same issue in next error

pixeltable/catalog/catalog.py
```diff
@@ -2573,24 +2573,36 @@ def _handle_path_collision(
...
+            if expected_obj_type != Dir and isinstance(obj, Dir):
+                raise excs.Error(f'Path {path!r} is an existing directory, expected a table, view or snapshot')
+
+        if obj is None:
```
aaron-siegel, line 2574: Move this up; then the preceding `if obj is not None` can be simplified away

tests/test_dirs.py
```diff
@@ -125,13 +125,13 @@ def test_create_if_exists(self, uses_db: None) -> None:
...
 
-        # scenrio 2: path already exists but is not a Dir
+        # scenario 2: path already exists but is not a Dir
         make_tbl('dir1/t1')
         for if_exists in ['ignore', 'replace', 'replace_force']:
             with pytest.raises(pxt.Error) as exc_info:
                 pxt.create_dir('dir1/t1', if_exists=if_exists)  # type: ignore[arg-type]
             err_msg = str(exc_info.value).lower()
-            assert 'already exists' in err_msg and 'not a dir' in err_msg, f' for if_exists={if_exists!r}'
+            assert 'is an existing' in err_msg and 'expected a directory' in err_msg, f' for if_exists={if_exists!r}'
```
aaron-siegel: Modernize to using `pytest.raises(pxt.Error, match=...)` syntax, while you're at it?

tests/test_view.py
```diff
@@ -204,12 +204,28 @@ def test_create_if_exists(self, uses_db: None, reload_tester: ReloadTester) -> N
...
+            _ = pxt.create_table('not_view', {'c1': pxt.String})
+
+        # scenario 4: view exists but with a different base table
+        pxt.drop_table('not_view')
+        other_base = pxt.create_table('other_base', {'c1': pxt.String})
+        _ = pxt.create_view('view_with_base', t)
+        with pytest.raises(pxt.Error) as exc_info:
+            _ = pxt.create_view('view_with_base', other_base, if_exists='ignore')
+        err_msg = str(exc_info.value).lower()
+        assert 'already exists' in err_msg and 'different base table' in err_msg
```
aaron-siegel: Definitely use match=... syntax for any new code.

aaron-siegel (changes requested):

aaron-siegel:
Please update the PR description to be more specific about what's changing (is the `if_exists='ignore'` logic changing too? A naive reading of the code suggests Yes)

tests/test_catalog.py
```diff
@@ -177,6 +177,48 @@ def test_ls(self, uses_db: None) -> None:
...
+        pxt.drop_table('target')
+        pxt.create_dir('target')
+        for creator, _ in creators.values():
+            with pytest.raises(excs.Error, match='expected a table, view or snapshot'):
+                creator()
+
+        # And table subtypes cannot be replaced by dirs
+        pxt.drop_dir('target')
+        pxt.create_table('target', {'c1': pxt.Int})
+        with pytest.raises(excs.Error, match='expected a directory'):
```
sergey-mkhitaryan: nit: you can do this in the first loop instead -- that will give you more test coverage in fewer lines of code

pixeltable/catalog/catalog.py
christopherpestano: Fixed

pixeltable/catalog/catalog.py
christopherpestano: Makes sense, this reduces alot of the obj is not None checks in the code overall.

tests/test_dirs.py
christopherpestano: Sure can do, changed function to use modern match. In the future though we should probably have claude scan the test files and do a mass replacement for these things.

tests/test_view.py
christopherpestano: Sure, changed the other asserts within this ut as well to that syntax as well for consistency.

tests/test_catalog.py
christopherpestano: I suppose but ideally I want to keep the error case check separate from the non-error logic check. But I think the loop is a good idea, changed it to instead loop over the existing creators with checks for table type replacing dir and dir replacing table type,

PR #1152, opened 2026-02-18
PXT-933 Support videos in gemini generate_content

amithadke:


pixeltable/functions/gemini.py
```diff
@@ -84,6 +96,7 @@ async def generate_content(
     Args:
         contents: The input content to generate from. Can be a prompt, or a list containing images and text
             prompts, as described in: <https://ai.google.dev/gemini-api/docs/text-generation>
+            For video columns, use the ``.localpath`` property (e.g. ``tbl.video_col.localpath``)
```
aaron-siegel: Is there a reason to use double backticks here instead of the usual convention? Are we sure it will render properly in mintlify?

pixeltable/functions/gemini.py
```diff
@@ -102,7 +115,14 @@ async def generate_content(
         ...     response=generate_content(tbl.prompt, model='gemini-2.5-flash')
         ... )
 
-        Add a computed column that applies the model `gemini-2.5-flash` for image understanding
+        Add a computed column that applies the model `gemini-2.5-flash` for image understanding.
+
+        With video, use the column's ``.localpath`` property
```
aaron-siegel: Hmmmm. This seems like an antipattern. What goes wrong when you use `tbl.video_col` instead of `tbl.video_col.localpath`? If something breaks when you try that, it might be better to fix it, rather than force the user to explicitly rely on `.localpath`.

amithadke: tbl.video_col works too removed localpath

PR #1153, opened 2026-02-20
TableOp refactoring so that TableVersion is not required for some ops

sergey-mkhitaryan:
Command pattern: each `TableOp` is capable of executing and undoing itself (provided the target `TableVersion`).

The motivation for this refactoring is that not all `TableOp`s need a `TableVersion` to execute or undo. Not requiring `Catalog` to load a `TableVersion` for some pending ops will resolve the bug found in https://github.com/pixeltable/pixeltable/pull/1138 where a table cannot be deleted, because its `TableVersion` can no longer be instantiated at the last pending op.

Also in this refactoring: `needs_xact` is no longer a member variable of `TableOp` because whether or not a `TableOp` requires a transaction, can be determined statically based on the op type.


pixeltable/catalog/tbl_ops.py
```diff
@@ -36,61 +49,202 @@ def to_dict(self) -> dict:
...
+        # needs_xact used to be a member variable. Remove it from the dict for backward compatibility.
+        needs_xact_legacy = data.pop('needs_xact', None)
         op_class = getattr(sys.modules[__name__], classname)
-        return schema.md_from_dict(op_class, data)
+        op = schema.md_from_dict(op_class, data)
+        if needs_xact_legacy is not None:
+            assert op.needs_xact == needs_xact_legacy
+        return op
+
+    def roll(self, tv: TableVersion | None, *, is_rollback: bool) -> None:
```
mkornacker: roll()?

Also, why the indirection (since the caller needs to know whether it's redo/undo)?

pixeltable/catalog/tbl_ops.py
```diff
@@ -36,61 +49,202 @@ def to_dict(self) -> dict:
     @classmethod
     def from_dict(cls, data: dict) -> TableOp:
         classname = data.pop('_classname')
+        # needs_xact used to be a member variable. Remove it from the dict for backward compatibility.
```
mkornacker, line 52: Migration, please. Otherwise we'll have to deal with this forever.

pixeltable/catalog/tbl_ops.py
```diff
@@ -22,10 +33,12 @@ class OpStatus(Enum):
 
 @dataclasses.dataclass
 class TableOp:
+    needs_tv: ClassVar[bool] = True  # if False, exec/undo can be called with tv=None
```
mkornacker: Why a default here (but not for needs_xact)?

Maybe best not to have defaults, so that adding another subclass means you have to make an explicit decision.

pixeltable/catalog/tbl_ops.py
sergey-mkhitaryan: this piece of code is duplicated in catalog so I extracted it. can revert, not a problem.

pixeltable/catalog/tbl_ops.py
sergey-mkhitaryan: this seemed like a reasonable and safe default, but sure

pixeltable/catalog/tbl_ops.py
aaron-siegel: My own preference would be to avoid the migration here, and perhaps roll it in to the next unavoidable migration. In general there's some value in finding a way to reduce the frequency of migrations. In this case, the "band-aid" code is minimal and the extraneous dict entry won't persist long in any catalog, since TableOps will all be flushed by the end of the next successful catalog operation (if I understand things right).

Looks like it's already been implemented and I'm not proposing we undo that - just making a point that I think we'd benefit from avoiding inessential migrations.

sergey-mkhitaryan: Let's discuss tomorrow. Even though I already implemented a migration, I think it's an overkill and I would rather revert it. I can reduce this code to 1 LOC today, and we can delete it in a couple of months, and the chance of somebody hitting this are so small (and if they do, the one-time fix for them is not difficult).

sergey-mkhitaryan: I removed the migration code and added a TODO to delete this workaround later.

PR #1154, opened 2026-02-23
PXT-920 + PXT-1046 Refactor execution dispatch/init_rows to improve performance and adding perf testing

christopherpestano:
Perf diff after changes, halves runtime for large batches. Example: processing 100k rows now takes roughly the same amount of time it previously took to process 50k rows.

Both Changes
image
Difference between init_rows() + dispatch() vs just dispatch(): Slightly faster
image

### DIFF
--- a/pixeltable/exec/expr_eval/schedulers.py
+++ b/pixeltable/exec/expr_eval/schedulers.py
@@ -148,6 +148,9 @@ class RateLimitsScheduler(Scheduler):
             return list(self.pool_info.resource_limits.keys())
 
     def _get_request_resources(self, request: FnCallArgs) -> dict[str, int]:
+        # Fall back to default estimate if the request's function signature doesn't match the pool's estimator
+        if not all(name in request.fn_call.fn.signature.parameters for name in self.get_request_resources_param_names):
+            return dict.fromkeys(self._resources, 1)
         kwargs_batch = request.fn_call.get_param_values(self.get_request_resources_param_names, request.rows)
         if not request.is_batched:
             return self.pool_info.get_request_resources(**kwargs_batch[0])
--- a/tests/functions/test_openai.py
+++ b/tests/functions/test_openai.py
@@ -486,3 +486,28 @@ class TestOpenai:
         validate_update_status(t.insert(input='Where did the game of Backgammon originate?'), 1)
         result = t.collect()
         assert 'Mesopotamia' in result['chat_output'][0]['choices'][0]['message']['content']
+
+    def test_shared_rate_limits_pool_different_signatures(self, uses_db: None) -> None:
+        """Verify that functions sharing a rate-limits pool with different get_request_resources signatures
+        don't crash the scheduler.
+
+        The pool caches param names from chat_completions (messages, model, model_kwargs), but vision has
+        a different signature (prompt, image, model, model_kwargs). Inserting 2+ rows ensures the second
+        call hits the rate-limited path where the mismatch would previously trigger an AssertionError.
+        """
+        skip_test_if_not_installed('openai')
+        skip_test_if_no_client('openai')
+        from pixeltable.functions.openai import vision
+
+        t = pxt.create_table('test_tbl', {'img': pxt.Image})
+        t.add_computed_column(response=vision(prompt="What's in this image?", image=t.img, model='gpt-4o-mini'))
+        # Insert 2 rows: first initializes the pool synchronously, second goes through _get_request_resources
+        with pytest.warns(
+            pxt.exceptions.PixeltableDeprecationWarning,
+            match=r'vision\(\) is deprecated as a separate API; use chat_completions\(\) instead',
+        ):
+            validate_update_status(t.insert([{'img': SAMPLE_IMAGE_URL}, {'img': SAMPLE_IMAGE_URL}]), expected_rows=2)
+        result = t.collect()
+        assert len(result) == 2
+        for row in result:
+            assert len(row['response']) > 0


aaron-siegel:
Huge! Great work, @christopherpestano !

Which platforms did you test this on? In the past, we've found that performance improvements can vary unevenly across platforms (e.g., a previous round of optimizations resulted in a 50% improvement on MacOS, but only 20-25% on linux).

I love the benchmark tests in CI. How long do they take to run? If they're relatively quick (under 30 minutes, say), we should probably be running them on PRs and in merge queue, not just in stress-tests (which will only catch them after the fact).

.github/workflows/stress-tests.yml
```diff
@@ -117,3 +117,63 @@ jobs:
       - name: Report failure
         if: failure()
         run: echo "Performance test failed. Check the logs for details."
+
+  benchmark-tests:
+    runs-on: ubuntu-large
```
aaron-siegel: Are you sure we need this? Larger instances are more expensive and not necessarily any more performant for the things we're trying to measure (they just have more cores, disk, and RAM).

.github/workflows/stress-tests.yml
```diff
@@ -117,3 +117,63 @@ jobs:
...
+      - name: Run benchmark tests
+        run: |
+          pytest -m benchmark --benchmark-json=benchmark-results.json tests/
+
+      - name: Store benchmark results
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          tool: 'pytest'
+          output-file-path: benchmark-results.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
aaron-siegel: Is this something we need to configure in actions secrets?

tests/benchmarks/test_expr_eval_perf.py
```diff
@@ -0,0 +1,132 @@
...
+    These benchmarks include both SQL-pushdown functions (string operations)
+    and UDFs that must go through ExprEvalNode (image, video, audio functions).
+    """
+
+    @pytest.mark.benchmark(group='batch_scaling')
+    @pytest.mark.parametrize('row_count', [1000, 10000, 50000, 100000])
+    def test_select_batch_scaling(self, uses_db: None, benchmark: Any, row_count: int) -> None:
+        """Test how performance scales with row count (vectorization benefit)."""
+        t = pxt.create_table(f'scale_tbl_{row_count}', {'c1': pxt.Int, 'c2': pxt.String})
+        t.insert([{'c1': i, 'c2': f'str_{i}'} for i in range(row_count)])
```
aaron-siegel: spurious list creation (here and in other similar instances below)

tests/benchmarks/test_expr_eval_perf.py
```diff
@@ -0,0 +1,132 @@
+from typing import Any
+
+import pyarrow as pa
+import pytest
+import sqlalchemy as sa
+
+import pixeltable as pxt
+import pixeltable.functions.audio as pxt_audio
```
aaron-siegel: More conventional would be `import pixeltable.functions as pxtf` and then refer to `pxtf.audio`, etc.

tests/benchmarks/test_expr_eval_perf.py
```diff
@@ -0,0 +1,132 @@
...
+class TestExprEvalPerformance:
+    """Benchmarks for expression evaluation dispatch optimization.
+
+    These benchmarks include both SQL-pushdown functions (string operations)
+    and UDFs that must go through ExprEvalNode (image, video, audio functions).
+    """
+
+    @pytest.mark.benchmark(group='batch_scaling')
+    @pytest.mark.parametrize('row_count', [1000, 10000, 50000, 100000])
+    def test_select_batch_scaling(self, uses_db: None, benchmark: Any, row_count: int) -> None:
```
aaron-siegel: To get the correct type hint for `benchmark`:
```
from pytest_benchmark.fixture import BenchmarkFixture

benchmark: BenchmarkFixture
```
(Note: I figured this out by putting `assert False, type(benchmark)` in the body of the test and running it)

tests/benchmarks/test_expr_eval_perf.py
```diff
@@ -0,0 +1,132 @@
...
+            t.insert([{'img': SAMPLE_IMAGE_URL} for _ in range(row_count)])
+
+        benchmark(insert_resized)
+
+    @pytest.mark.benchmark(group='video_frames')
+    def test_insert_video_frame_extraction(self, uses_db: None, benchmark: Any) -> None:
+        """Benchmark video frame extraction."""
+        video_files = get_video_files()
+        if not video_files:
+            pytest.skip('No test videos available')
```
aaron-siegel: Best to avoid these skip-on-error directives - they could mask a real error and cause it to go unnoticed.

If we want to case on video_files being empty, better to do `assert len(video_files) > 0`

tests/benchmarks/test_expr_eval_perf.py
```diff
@@ -0,0 +1,132 @@
...
+        video_path = video_files[0]
+
+        t = pxt.create_table('video_frame_tbl', {'video': pxt.Video, 'timestamp': pxt.Float})
+        t.add_computed_column(frame=pxt_video.extract_frame(t.video, timestamp=t.timestamp))
```
aaron-siegel: Would it be better to use an iterator here?

aaron-siegel (changes requested):

christopherpestano:
> Huge! Great work, @christopherpestano !
> 
> Which platforms did you test this on? In the past, we've found that performance improvements can vary unevenly across platforms (e.g., a previous round of optimizations resulted in a 50% improvement on MacOS, but only 20-25% on linux).
> 
> I love the benchmark tests in CI. How long do they take to run? If they're relatively quick (under 30 minutes, say), we should probably be running them on PRs and in merge queue, not just in stress-tests (which will only catch them after the fact).

Thanks! This was done on MacOS, but I can check and see if there are any improvements in an ubuntu image locally. As to the CI tests, they are fairly short currently around 10 minutes (though that may change as we add coverage). But the worry about adding it as a requirement for PRs was that since the failure threshold is only 15%, we may see flaky perf tests blocking PRs. The idea for now is to trial it in the daily stress test for a bit to see how consistent it is, and given that it's not flaky we can just move the job to the pr workflow.

.github/workflows/stress-tests.yml
christopherpestano: Hmm, I suppose I don't really need this large of an instance as currently the tests are fairly small and low memory. Downsized for now, if this ends up being an issue as we add tests we can always scale up.

.github/workflows/stress-tests.yml
christopherpestano: I don't think so, it's automatically provisioned by github https://docs.github.com/en/actions/concepts/security/github_token

tests/benchmarks/test_expr_eval_perf.py
christopherpestano: Whoops didn't notice that, changed style in general/moved out of benchmark so it doesn't impact benchmark times.

tests/benchmarks/test_expr_eval_perf.py
christopherpestano: Yeah I think so, modified so instead we are using iterator.

tests/benchmarks/test_expr_eval_perf.py
christopherpestano: Interesting, I was having issues with that since I didn't have pytest_benchmark in the dev deps only the test deps. Moved it into the dev deps and added that so pylance isn't giving me trouble.

tests/benchmarks/test_expr_eval_perf.py
christopherpestano: Makes sense, modified to be an assertion

tests/benchmarks/test_expr_eval_perf.py
christopherpestano: Hmm I suppose that is a bit cleaner, modified the two imports to instead be that one you put

christopherpestano:
Note for folks looking at the tests: I modified the tests to instead use a no-op udf since we have another PR that will break this perf test by pushing the udfs currently used into sql. Having a custom defined udf ensures that we continue to benchmark what we intended to benchmark (that is the expr_eval_node).

tests/benchmarks/test_expr_eval_perf.py
aaron-siegel: Oh I wasn't concerned about performance (I doubt it's material relative to the benchmarks). What I meant is there's no reason to enclose the generator in list creation brackets. Instead of
```python
t.insert([{'i': i} for i in range(n)])
```
it's cleaner and more performant to just write
```python
t.insert({'i': i} for i in range(n))
```

tests/benchmarks/test_expr_eval_perf.py
aaron-siegel: test deps is the right place for it; please keep it there or it may cause problems in certain CI configs.

pylance should be ok with it as long as it's pointed at the right conda environment and you've done a `make install` in that environment.

christopherpestano: Hmm ok changed it back then, I guess when I was running this code I forgot to make install/something was bugging out.

tests/test_query.py
```diff
@@ -959,3 +960,17 @@ def extract_fields(exc_info: pytest.ExceptionInfo) -> set[str]:
         with pytest.raises(pxt.Error, match=r'Extra fields .* are not allowed in model') as exc_info:
             _ = list(t.select(t.i, t.s, t.f, t.b, t.ts, t.d, extra=t.i + t.f).collect().to_pydantic(StrictTestModel))
         assert extract_fields(exc_info) == {'extra'}
+
+    @pytest.mark.benchmark(group='select_inexpensive')
+    def test_select_inexpensive(self, uses_db: None, benchmark: Any) -> None:
```
sergey-mkhitaryan, line 965: do we intend for these tests to run on each commit? how much do they contribute to the total test duration? is there an actual performance validation that happens then, or if it doesn't fail, it passes?

christopherpestano: No, these tests are supposed to be run as part of the daily stress tests, yes currently there is a performance validation present in the yml where if there is a 15% regression the job will fail. Depending on how many rounds we do the tests will be long or short, currently they are around 10 minutes but as we add tests that can quickly go up (though we can always scale down the max row count we use for the tests).

christopherpestano: Nevermind we are now changing this to point at Grafana instead, closing

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,107 @@ def dispatch_exc(
...
+        num_rows = len(rows)
+        num_slots = exec_ctx.row_builder.num_materialized
+
+        # Static arrays from row_builder
+        dependencies = exec_ctx.row_builder.dependencies  # (num_slots, num_slots)
+        num_dependencies = exec_ctx.row_builder.num_dependencies  # (num_slots,)
+
+        # Stack row arrays into 2D arrays for vectorized operations
+        # Shape: (num_rows, num_slots)
+        has_val = np.stack([r.has_val for r in rows], axis=0)
```
sergey-mkhitaryan, line 363: is there any particular reason `axis=0` is explicit?

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,107 @@ def dispatch_exc(
...
+        # ---------- Compute progress (output slot materialization) ----------
         num_computed_outputs = 0
+        if self.eval_ctx is exec_ctx:
+            # Count currently non-materialized output slots (before updating missing_slots)
+            missing_outputs_before = (missing_slots & self.outputs).sum()
+
+            # Update missing_slots: clear slots that now have values
+            missing_slots &= ~has_val
+
+            missing_outputs_after = (missing_slots & self.outputs).sum()
```
sergey-mkhitaryan: I would rewrite this block to avoid code duplication (even if it's one line) and to make it clear which parts are for progress reporting and which are not. You can also cut on unnecessary computation. Something like this:

```
report_progress = self.eval_ctx is exec_ctx and self.progress_reporter is not None
if report_progress:
    missing_outputs_before = (missing_slots & self.outputs).sum()

missing_slots &= ~has_val

if report_progress:
    missing_outputs_after = (missing_slots & self.outputs).sum()
    num_computed_outputs = missing_outputs_before - missing_outputs_after
```

@mkornacker we need to find a way to report progress without incurring a O(N*N) computation, even in native code. Can schedulers update the reporter instead? It would be smoother, too, because `dispatch()` can only report chunks of work.

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,107 @@ def dispatch_exc(
...
+            num_computed_outputs = missing_outputs_before - missing_outputs_after
+        else:
+            # Update missing_slots: clear slots that now have values
+            missing_slots &= ~has_val
+
+        # ---------- Identify completed rows ----------
+        missing_slot_counts = missing_slots.sum(axis=1)  # (num_rows,)
+        completed_mask = missing_slot_counts == 0
```
sergey-mkhitaryan: Some unnecessary computation here. `completed_mask` and `missing_slot_counts` can be completely eliminated:
```
non_completed_mask = missing_slots.any(axis=1)
```

Also let's rename non_completed_mask to maybe non_completed_rows or something that makes it clear that it's rows.

Down in the "Handle completed rows" block:
```
        if not np.all(non_completed_mask):
            completed_idxs = list((~non_completed_mask).nonzero()[0])
```

(Yes, here we basically calculate `completed_mask` again, but lazily. I have a feeling that the last line can be further improved, but not yet sure how.)

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,107 @@ def dispatch_exc(
...
+
+        # ---------- Compute ready slots for non-completed rows ----------
+        # ready_slots: (num_rows, num_slots)
+        ready_slots = np.zeros((num_rows, num_slots), dtype=bool)
+
+        non_completed_mask = ~completed_mask
+        if np.any(non_completed_mask):
+            # Only compute for non-completed rows
+            nc_missing_slots = missing_slots[non_completed_mask]
```
sergey-mkhitaryan: how much of a performance improvement we are getting by creating these submatrixes and computing only on them? because there's a pretty significant code complexity cost to this

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,107 @@ def dispatch_exc(
...
+        # ---------- GC computation ----------
+        # Compute new missing_dependents for all rows
+        # missing_dependents[i, slot] = count of slots that depend on 'slot' and don't have a value yet
+        # dependencies[i, j] means expr i depends on expr j
+        # For each slot j, we count how many slots i (that don't have values) depend on j
+        # This is: (~has_val) @ dependencies, summing over the first axis
+        new_missing_dependents = (~has_val).astype(np.int16) @ dependencies.astype(np.int16)
+
+        # Identify slots to garbage collect per row
+        gc_targets_2d = (new_missing_dependents == 0) & (missing_dependents > 0) & exec_ctx.gc_targets[np.newaxis, :]
```
sergey-mkhitaryan: `gc_targets_2d` is not a great name, please rename it. Certainly drop the _2d part, instead you can write a comment explaining what `[i][j]` of that matrix means.

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,107 @@ def dispatch_exc(
...
-            missing_dependents = np.sum(exec_ctx.row_builder.dependencies[row.has_val == False], axis=0)
-            gc_targets = (missing_dependents == 0) & (row.missing_dependents > 0) & exec_ctx.gc_targets
-            row.clear(gc_targets)
-            row.missing_dependents = missing_dependents
+            # Perform GC clear for slots that are no longer needed
+            if gc_targets_2d[i].any():
+                row.clear(gc_targets_2d[i])
+            row.missing_dependents = new_missing_dependents[i]
 
+        # ---------- Progress reporting ----------
```
sergey-mkhitaryan: we need to move this section up next to where `num_computed_outputs` is computed. It's out of place here.

sergey-mkhitaryan (changes requested):
I can't finish this code review in one go, but here are some comments so far. Nicely done, the performance improvements look great!

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: It's just for readability, so people doing a first glance at the code know we're stacking the rows as rows here and not stacking them as columns.

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Renamed and modified comment to claify that this is a boolean array where arr[i][j] just means for row i slot j will be set to True in the case it can be garbage collected.

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: I would say it's a pretty large gain considering in the next few lines of code we use the mask 7 times, meaning we are saving 6 2D operations by doing it in this pattern instead of doing ~completed_mask.

christopherpestano: We lose about 10~15% performance by removing this logic, I think we shouldn't be sacrificing too much performance here for easy of understanding as the code here is fairly self contained. alt text

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Makes sense moved up

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Ran the benchmark test with this change, there isn't really a significant difference in performance with or without this change. Personally, I think it adds significant code complexity of having completed_mask exist implicitly than as a named variable, so my preference would be not to do this change.
image

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Hmm this seems like it would add to the code complexity. By doing this we dedupe the missing_slots &= ~has_val, however we end up with the missing_outputs_before being used in a completely different if statement than the one that produced it. That seems like a bad pattern to be introducing into the code, variables in an if statement should be scoped locally to just that if statement generally speaking.

christopherpestano: Resolved off line, added changes without predeclaring any variables.

pixeltable/exec/expr_eval/globals.py
```diff
@@ -191,10 +191,33 @@ def _init_slot_evaluators(self, dispatcher: Dispatcher, target_slot_idxs: list[i
...
-            row.missing_dependents = np.sum(self.row_builder.dependencies[row.has_val == False], axis=0)
-            row.missing_slots = self.eval_ctx & (row.has_val == False)
+
+        # Stack has_val for vectorized computation
+        # Shape: (num_rows, num_slots)
+        has_val = np.stack([r.has_val for r in rows], axis=0)
+
+        # Vectorized missing_dependents computation
+        # dependencies[i, j] means expr i depends on expr j
+        # missing_dependents[slot] = count of slots that depend on 'slot' and don't have a value
```
mkornacker: explain it in terms of missing_dependents[i]

pixeltable/exec/expr_eval/globals.py
```diff
@@ -191,10 +191,33 @@ def _init_slot_evaluators(self, dispatcher: Dispatcher, target_slot_idxs: list[i
...
+        dependencies = self.row_builder.dependencies
+        new_missing_dependents = (~has_val).astype(np.int16) @ dependencies.astype(np.int16)
```
mkornacker: explain int16 conversion

For every matrix/numpy operation, comment on output shape.

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,104 @@ def dispatch_exc(
...
+        num_dependencies = exec_ctx.row_builder.num_dependencies  # (num_slots,)
+
+        # Stack row arrays into 2D arrays for vectorized operations
+        # Shape: (num_rows, num_slots)
+        has_val = np.stack([r.has_val for r in rows], axis=0)
+        missing_slots = np.stack([r.missing_slots for r in rows], axis=0)
+        is_scheduled = np.stack([r.is_scheduled for r in rows], axis=0)
+        missing_dependents = np.stack([r.missing_dependents for r in rows], axis=0)
+
+        # ---------- Compute progress (output slot materialization) ----------
```
mkornacker: Get rid of -----------, we don't use that elsewhere.

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,104 @@ def dispatch_exc(
...
+        report_progress = self.progress_reporter is not None and self.eval_ctx is exec_ctx
+        if report_progress:
+            # Count currently non-materialized output slots (before updating missing_slots)
+            missing_outputs_before = (missing_slots & self.outputs).sum()
```
mkornacker: Here, too: comment on the result shape for all numpy operations.

tests/test_query.py
sergey-mkhitaryan: These tests still run on commit, which I don't think they should due to their rather long duration. Please take a look.

.github/workflows/stress-tests.yml
```diff
@@ -118,3 +118,51 @@ jobs:
...
+          python -m pip install "uv==$UV_VERSION"
+
+      - name: Install the project dependencies
+        run: |
+          export VIRTUAL_ENV="$CONDA_PREFIX"
+          ./scripts/retry.sh 3 60 uv sync --active --group test
+
+      - name: Run benchmark tests
+        run: |
+          pytest -m benchmark --benchmark-json=benchmark-results.json tests/
```
sergey-mkhitaryan, line 162: does this print the benchmark results in the end to stdout? if not please insert a step that does `cat <json file>` or something

pixeltable/exec/expr_eval/expr_eval_node.py
sergey-mkhitaryan: From the previous round:

> Also let's rename non_completed_mask to maybe non_completed_rows or something that makes it clear that it's rows.

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,104 @@ def dispatch_exc(
...
+            # missing_dependencies: dependencies needed for missing slots
+            # Shape: (num_nc_rows, num_slots)
+            nc_missing_dependencies = num_dependencies * nc_missing_slots
+
+            # num_mat_dependencies: array of how many dependencies are materialized per slot, per row
+            # has_val @ dependencies.T gives count of materialized dependencies per slot
+            # dependencies[i, j] == True means expr i depends on expr j
+            # dependencies.T[j, i] == True means expr i depends on expr j
+            # We want: for each slot i, count how many of its dependencies have values
+            nc_num_mat_dependencies = nc_has_val.astype(np.int32) @ dependencies.T.astype(np.int32)
```
sergey-mkhitaryan: any particular reason why it's int32 here and int16 elsewhere?

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,104 @@ def dispatch_exc(
...
+            nc_ready = (nc_num_missing == 0) & (~nc_is_scheduled) & nc_missing_slots
+
+            ready_slots[non_completed_mask] = nc_ready
+
+            # Update is_scheduled for non-completed rows
+            is_scheduled[non_completed_mask] |= nc_ready
+
+        # GC computation 
+        # Compute new missing_dependents for all rows
+        # missing_dependents[i, slot] = count of slots that depend on 'slot' and don't have a value yet
```
sergey-mkhitaryan: s/missing_dependents/new_missing_dependents

Please make sure the name in the comments (everywhere, not just here) matches the actual variable name, otherwise it's slightly confusing

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,104 @@ def dispatch_exc(
...
-
-        if np.any(completed_rows):
-            completed_idxs = list(completed_rows.nonzero()[0])
-            if rows[i].parent_row is not None:
+            row.missing_slots = missing_slots[i]
+            row.is_scheduled = is_scheduled[i]
+
+            # Perform GC clear for slots that are no longer needed
+            if gc_targets[i].any():
+                row.clear(gc_targets[i])
```
sergey-mkhitaryan, line 440: I think there's a bug here in GC (that existed before your change, too). The root cause is that we do not distinguish between "has no value because we haven't computed it yet" and "has no value because we already GC'd it".  There's just `has_val` that means either one of the two. This lack of a distinction results in never GC'ing a (logically no longer needed) slot whose dependent was already GC'd (because after GC the dependent will have `has_val == False` as if it's yet to be computed). Please see the attached writeup for a specific scenario, and let me know if that's not clear enough. Apologies for the cursive.
IMG_3366 copy

tests/benchmarks/test_expr_eval_perf.py
christopherpestano: Cool changed all functions that can accept generators to accept generators. Let's note here though that for sqlalchemy and pyarrow this isn't supported though, as this test contains those methods calls as well with this generator to list logic as well.
image
image

tests/test_query.py
christopherpestano: Ah didn't see that, when I had made this change I wasn't aware of ci_tools.py. Modified it there to disable this on normal pr test runs.

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: I think this naming change is a bit misleading for a reader. Non-completed rows would be along the lines rows[non_completed_mask] since the mask itself is just an array of booleans that can be applied to get the non-completed rows. It by itself is not the non-completed rows.

.github/workflows/stress-tests.yml
christopherpestano: Already prints the results to the stdout of the gh action. Image attached 
image

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Good catch, there's actually no particular reason, let me switch to int16 for memory efficiency.

tests/test_query.py
```diff
@@ -959,3 +960,17 @@ def extract_fields(exc_info: pytest.ExceptionInfo) -> set[str]:
...
+        t = pxt.create_table('test_inexpensive', {'c1': pxt.Int, 'c2': pxt.String})
+
+        row_count = 100000
+
+        t.insert([{'c1': i, 'c2': f'str_{i}'} for i in range(row_count)])
```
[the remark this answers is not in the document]
christopherpestano: Implemented

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Discussed in slack chat, confirmed the existence of this bug, however it is beyond of the scope of this PR. Will raise different pr to resolve this issue.

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Done changed to new_missing_dependents

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Hmm the naming has grown on me after I changed it over, modified to match what you expected
non_completed_mask -> non_completed_rows
completed_mask -> completed_rows

pixeltable/exec/expr_eval/globals.py
christopherpestano: explained better

pixeltable/exec/expr_eval/globals.py
christopherpestano: Explained that it's to avoid doing boolean addition instead of arithmetic

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Removed

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Added shape where not previously present

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,104 @@ def dispatch_exc(
...
+            # Count currently non-materialized output slots (before updating missing_slots)
+            missing_outputs_before = (missing_slots & self.outputs).sum()
+
+        # Update missing_slots: clear slots that now have values
+        missing_slots &= ~has_val
+
+        # Progress reporting
+        if report_progress:
+            missing_outputs_after = (missing_slots & self.outputs).sum()
+            num_computed_outputs = missing_outputs_before - missing_outputs_after
```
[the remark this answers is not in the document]
christopherpestano: added

pixeltable/exec/expr_eval/globals.py
```diff
@@ -191,10 +191,33 @@ def _init_slot_evaluators(self, dispatcher: Dispatcher, target_slot_idxs: list[i
...
+
```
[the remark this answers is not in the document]
christopherpestano: adds too much code complexity, bad suggestion

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,104 @@ def dispatch_exc(
...
-        ready_slots = np.zeros((len(rows), exec_ctx.row_builder.num_materialized), dtype=bool)
-        completed_rows = np.zeros(len(rows), dtype=bool)
-        num_computed_outputs = 0
+        num_rows = len(rows)
+        num_slots = exec_ctx.row_builder.num_materialized
+
+        # Static arrays from row_builder
+        dependencies = exec_ctx.row_builder.dependencies  # (num_slots, num_slots)
+        num_dependencies = exec_ctx.row_builder.num_dependencies  # (num_slots,)
+
```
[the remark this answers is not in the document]
christopherpestano: adds too much complexity bad suggestion, lets avoid over optimizing native operations when most of runtime is spent in actual python

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -344,52 +344,106 @@ def dispatch_exc(
...
+        has_val = np.stack([r.has_val for r in rows], axis=0)
+        missing_slots = np.stack([r.missing_slots for r in rows], axis=0)
+        is_scheduled = np.stack([r.is_scheduled for r in rows], axis=0)
+        missing_dependents = np.stack([r.missing_dependents for r in rows], axis=0)
+
+        # Compute progress (output slot materialization)
+        report_progress = self.progress_reporter is not None and self.eval_ctx is exec_ctx
+        if report_progress:
+            # Count currently non-materialized output slots (before updating missing_slots)
+            missing_outputs_before = (missing_slots & self.outputs).sum()  # scalar
```
sergey-mkhitaryan: instead of `# scalar` you can just put a type hint `np.int64` or whatever it is. same below.

aaron-siegel (approved):

pixeltable/exec/expr_eval/expr_eval_node.py
christopherpestano: Ok changed to np.int64

PR #1155, opened 2026-02-24
PXT-1020: support for multi-threaded API calls

mkornacker:
Introduces a new Runtime class that holds thread-local global state (DB connection/session state, event loop, progress tracking, Catalog) that previously existed as part of the Catalog and Env singletons

Also:
- Env is now purely process-level shared state (engine, paths, logging, config); all per-thread mutable state moved out
- Added locks to Config.init(), Env.get(), Env.get_resource_pool_info(), and RateLimitsInfo methods to prevent races during initialization and rate limit tracking
- New tests/test_concurrent.py with a rudimentary multi-threaded insert+select

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -2555,11 +2555,18 @@ class Catalog:
     ) -> SchemaObject | None:
         obj, _, _ = self._prepare_dir_op(add_dir_path=path.parent, add_name=path.name)
 
-        if if_exists == IfExistsParam.ERROR and obj is not None:
+        if obj is None:
+            return None
+
+        # IfExistsParam.ERROR: Error given there is existing schema object
+        if if_exists == IfExistsParam.ERROR:
             raise excs.Error(f'Path {path!r} is an existing {obj._display_name()}')
-        else:
-            is_snapshot = isinstance(obj, View) and obj._tbl_version_path.is_snapshot()
-            if obj is not None and (not isinstance(obj, expected_obj_type) or (expected_snapshot and not is_snapshot)):
+
+        # IfExistsParam.IGNORE: Return existing object if it matches expected type (and base table for views/snapshots)
+        if if_exists == IfExistsParam.IGNORE:
+            # for ignore, we can only return the existing object if it matches the expected type
+            is_existing_snapshot = isinstance(obj, View) and obj._tbl_version_path.is_snapshot()
+            if not isinstance(obj, expected_obj_type) or (expected_snapshot and not is_existing_snapshot):
                 if expected_obj_type is Dir:
                     obj_type_str = 'directory'
                 elif expected_obj_type is InsertableTable:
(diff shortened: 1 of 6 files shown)


pixeltable/runtime.py
```diff
@@ -0,0 +1,180 @@
...
+            _logger.debug('Patched running loop')
+        except RuntimeError:
+            self._event_loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(self._event_loop)
+            # we set a deliberately long duration to avoid warnings getting printed to the console in debug mode
+            self._event_loop.slow_callback_duration = 3600
+
+        # always allow nested event loops, we need that to run async udfs synchronously (eg, for SimilarityExpr);
+        # see run_coroutine_synchronously()
+        nest_asyncio.apply()
```
amithadke: so we will use nest_asyncio.apply() only when jupyter is detected
if we see uvloop should we fail with an error message?

pixeltable/functions/jina.py
```diff
@@ -45,21 +43,15 @@ class JinaUnexpectedError(Exception):
     pass
 
 
-class _JinaClient:
+class _JinaClient:  # noqa: B903
```
aaron-siegel: I think this is ruff trying to say this should be a `@dataclass`. We could consider disabling that linting rule globally, it seems kind of silly.

pixeltable/functions/jina.py
```diff
@@ -94,6 +89,8 @@ def _(api_key: str) -> _JinaClient:
 
 
 def _client() -> _JinaClient:
+    from pixeltable.env import Env
```
aaron-siegel: Does importing `Env` into global scope no longer work as expected?

pixeltable/env.py
aaron-siegel, line 45: This PR doesn't contain any changes to client lifecycle (unless I'm missing something). Is it safe to assume all client objects (e.g. OpenAI client) are threadsafe?

mkornacker: Good point, and unlikely to be the case. Let me investigate and possibly follow up with moving more state into Runtime.

pixeltable/catalog/table.py
```diff
@@ -1373,7 +1359,9 @@ def _drop_index(
         # Find out if anything depends on this index
         val_col = idx_info.val_col
         dependent_user_cols = [
-            c for c in Catalog.get().get_column_dependents(val_col.get_tbl().id, val_col.id) if c.name is not None
+            c
+            for c in get_runtime().catalog.get_column_dependents(val_col.get_tbl().id, val_col.id)
```
sergey-mkhitaryan: legacy issue, but this line is overloaded and difficult to parse. I would extract `get_column_dependents()` result as a variable.

pixeltable/_query.py
```diff
@@ -336,7 +337,7 @@ def exec_plan() -> Iterator[exprs.DataRow]:
             with plan:
                 for row_batch in plan:
                     # stop progress output before we display anything, otherwise it'll mess up the output
-                    Env.get().stop_progress()
+                    get_runtime().stop_progress()
```
sergey-mkhitaryan, line 340: "runtime" can mean many different things, would `QueryRuntime`/`query_runtime()` be a better naming?

pixeltable/runtime.py
```diff
@@ -0,0 +1,190 @@
...
+        """Context manager for the Progress instance."""
+        try:
+            yield
+        finally:
+            self.stop_progress()
+
+
+def get_runtime() -> Runtime:
+    """Get the current thread's Runtime instance, creating one if needed."""
+    runtime = getattr(_thread_local, 'runtime', None)
```
sergey-mkhitaryan, line 190: I'm wondering if thread local is the right approach. FastApi supports both sync (one request per thread) and async (asyncio) request handlers, is it a good idea to restrict ourselves (and the users) by only supporting sync handlers? There's something called `contextvars` (https://docs.python.org/3/library/contextvars.html) that are kind of like thread local but are natively supported by asyncio, can that be a better choice?

pixeltable/config.py
```diff
@@ -23,6 +24,7 @@ class Config:
     """
 
     __instance: ClassVar[Config | None] = None
+    __init_lock: ClassVar[threading.Lock] = threading.Lock()
```
sergey-mkhitaryan, line 27: Is the idea that `__init_lock` protects `__instance` from concurrent read-writes? Could you write a comment that explicitly states that?

pixeltable/config.py
```diff
@@ -79,21 +81,24 @@ def config_file(self) -> Path:
 
     @classmethod
     def get(cls) -> Config:
+        if cls.__instance is not None:
```
sergey-mkhitaryan, line 84: should the entire body of `get()` be behind `cls.__init_lock`? (you may have to switch to `threading.RLock` though)

pixeltable/_query.py
mkornacker: It applies to all pxt API calls, though, no just queries.

pixeltable/functions/jina.py
mkornacker: Done.

pixeltable/catalog/table.py

pixeltable/_query.py
mkornacker: But happy to rename it to something more meaningful.

pixeltable/runtime.py
mkornacker: It's unclear how async handlers could possibly work, given that Table won't be thread-safe/async-safe. That's no different than dbms connections (such as those handed out by Sqlalchemy's engine), which are never thread-safe.

pixeltable/config.py
sergey-mkhitaryan: and, strictly speaking, line 35 (`assert self.__instance is None`) should also be behind the lock, I think.

pixeltable/functions/reve.py
```diff
@@ -59,7 +50,10 @@ async def _post(self, endpoint: str, *, payload: dict) -> PIL.Image.Image:
             'Accept': requested_content_type,
         }
 
-        async with self.session.post(endpoint, json=payload, headers=request_headers) as resp:
+        async with (
+            aiohttp.ClientSession(base_url='https://api.reve.com') as session,
```
sergey-mkhitaryan: why is this change necessary? it eliminates connection pooling to Reve, which is going to worsen the performance. is there any way to avoid it?

mkornacker: The clients need another round of changes.

PR #1135, opened 2026-02-11
[PXT-1018] Add a "source" field to list of columns in t.describe()

sergey-mkhitaryan:

PR #1159, opened 2026-02-25
Fix TableVersion.is_iterator_column

sergey-mkhitaryan:


tests/test_iterator.py
```diff
@@ -650,5 +650,73 @@ def test_retrofit(self, uses_db: None) -> None:
             {'input': 'balloon', 'pos': 3, 'output_text': 'stored balloon 3', 'unstored_text': 'unstored balloon 3'},
         ]
 
+    def test_nested_iterator(self, uses_db: None) -> None:
```
aaron-siegel, line 653: Can we capture the user-visible impact of this change in the test? i.e., construct a user action that fails or generates incorrect output without it, but succeeds with it? Otherwise we're not actually capturing the functional impact of the change. (`is_iterator_column()` and `is_iterator_col` are non-public implementation details)

sergey-mkhitaryan: The user-visible impact is coming in https://github.com/pixeltable/pixeltable/pull/1135 which is how I found this issue. I can try to come up with a test that detects this via the row_builder usage, but that may be tricky.

aaron-siegel: That's what I meant. Does this actually impact RowBuilder?

aaron-siegel (approved):

tests/test_iterator.py
sergey-mkhitaryan: Row builder uses it:

```
        def refs_unstored_iter_col(col_ref: ColumnRef) -> bool:
            tbl = col_ref.col.get_tbl()
            return tbl.is_component_view and tbl.is_iterator_column(col_ref.col) and not col_ref.col.is_stored
```

I can't quite understand the end result of this, but I guess under some pretty specific conditions you might hit it, but even then not sure what, if any, the user-visible impact of this bug would be.

aaron-siegel: Ok, let's go ahead and merge it and #1135 will provide a functional check.

pixeltable/catalog/table_version.py
```diff
@@ -242,7 +242,9 @@ def __init__(
         self.num_iterator_cols = 0
         if self.view_md is not None and self.view_md.iterator_call is not None:
             self.iterator_call = GeneratingFunctionCall.from_dict(self.view_md.iterator_call)
-            self.num_iterator_cols = len(self.iterator_call.outputs)
+            # iterator_call.outputs includes the automatically added pos column, which we do not consider an iterator
```
mkornacker, line 245: As an aside, I don't think pos should be considered part of the iterator output (and therefore be included in 'outputs'). That's a system column added by the iterator view itself, in order to reconstruct the sequence.

mkornacker (approved):

pixeltable/catalog/table_version.py
sergey-mkhitaryan: cc @aaron-siegel

PR #1160, opened 2026-02-25
[PXT-1043] Support video embeddings in VoyageAI

aaron-siegel:


pixeltable/functions/voyageai.py
```diff
@@ -207,16 +208,12 @@ async def rerank(
     }
 
 
-@pxt.udf(batch_size=32, resource_pool='request-rate:voyageai')
+@pxt.udf(batch_size=4, resource_pool='request-rate:voyageai')
```
mkornacker, line 212: Does the ideal batch size depend on the parameter type (ie, 4 for video, 32 for text)?

pixeltable/functions/voyageai.py
```diff
@@ -250,52 +249,52 @@ async def multimodal_embed(
...
-    if input_type is not None:
-        kwargs['input_type'] = input_type
-    if truncation is not None:
-        kwargs['truncation'] = truncation
 
-    result = await cl.multimodal_embed(inputs=inputs, model=model, **kwargs)
-    return [np.array(emb, dtype=np.float64) for emb in result.embeddings]
+@multimodal_embed.overload
+async def _(
+    input: Batch[pxt.Video],
```
mkornacker: Call it video, as per the docstring above?

pixeltable/functions/voyageai.py
aaron-siegel: Probably, but we don't currently have a way to tune batch size by signature/parameters. Some other thoughts,
- I doubt anyone will be using this endpoint for pure text (there's `embeddings` for that)
- I'm not sure this multi-sig approach is the right one; Voyage's examples involve mixed modalities in the same query, which isn't possible with this API. I went with the existing structure for now but we should consider refactoring and/or adding an additional signature that accepts mixed modalities.

pixeltable/functions/voyageai.py
aaron-siegel: Good catch. Fixed.

PR #1164, opened 2026-02-27
uvloop compatibility

mkornacker:
Removes nest_asyncio.apply() and instead runs ExecNode.iter() in a background thread with a separate event loop. Also removes the nest_asyncio package dependency.

Testing: it's not clear how to test this other than creating a FastAPI test server (and driver).

### DIFF
--- a/pixeltable/functions/gemini.py
+++ b/pixeltable/functions/gemini.py
@@ -6,13 +6,16 @@ the [Working with Gemini](https://docs.pixeltable.com/notebooks/integrations/wor
 """
 
 import asyncio
+import base64
 import io
 import logging
+import mimetypes
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
 
 import numpy as np
 import PIL.Image
+from tenacity import RetryCallState, retry, retry_if_result, stop_after_delay, wait_exponential
 
 import pixeltable as pxt
 from pixeltable import env, exceptions as excs, exprs, type_system as ts
@@ -23,9 +26,18 @@ from pixeltable.utils.http import exponential_backoff, parse_duration_str
 from pixeltable.utils.local_store import TempStore
 
 if TYPE_CHECKING:
+    from typing import Coroutine
+
     from google import genai
 
 _logger = logging.getLogger('pixeltable')
+
+# Max raw file size (bytes) for inline_data; larger files use the Files API.
+# The API limit is 100MB for the base64-encoded payload, so we use ~75MB raw (75 * 4/3 ≈ 100MB encoded).
+GEMINI_INLINE_VIDEO_LIMIT_BYTES = 75 * 1024**2
+
+# Placeholder key used in first pass for large file uploads.
+_UPLOAD_PLACEHOLDER_KEY = '__google_genai_upload_ref__'
 
 
 @env.register_client('gemini')
@@ -102,8 +114,6 @@ async def generate_content(
         >>> tbl.add_computed_column(
         ...     response=generate_content(tbl.prompt, model='gemini-2.5-flash')
         ... )
-
-        Add a computed column that applies the model `gemini-2.5-flash` for image understanding
     """
     env.Env.get().require_package('google.genai')
     from google.genai import types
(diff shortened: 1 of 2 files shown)


pixeltable/exec/exec_node.py
```diff
@@ -56,19 +59,65 @@ def __aiter__(self) -> AsyncIterator[DataRowBatch]:
...
+                while True:
+                    batch: DataRowBatch = loop.run_until_complete(aiter.__anext__())
+                    yield batch
+            except StopAsyncIteration:
+                pass
+
+    _SENTINEL = object()
+
+    def _thread_iter(self) -> Iterator[DataRowBatch]:
+        # maxsize=2: we want a minimal amount of buffering
```
aaron-siegel: Why?

pixeltable/exec/exec_node.py
```diff
@@ -56,19 +59,65 @@ def __aiter__(self) -> AsyncIterator[DataRowBatch]:
...
+            try:
+                while True:
+                    batch: DataRowBatch = loop.run_until_complete(aiter.__anext__())
+                    yield batch
+            except StopAsyncIteration:
+                pass
+
+    _SENTINEL = object()
+
+    def _thread_iter(self) -> Iterator[DataRowBatch]:
```
aaron-siegel, line 83: This would benefit from a comment explaining its function.

pixeltable/exec/exec_node.py
```diff
@@ -56,19 +59,65 @@ def __aiter__(self) -> AsyncIterator[DataRowBatch]:
...
+                    async for batch in aiter(self):
+                        result_queue.put(batch)
+
+                loop.run_until_complete(produce())
+                result_queue.put(ExecNode._SENTINEL)
+            except BaseException as e:
+                result_queue.put(e)
+            finally:
+                loop.close()
+                thread_runtime.conn = None
```
aaron-siegel: Factor out into a `thread_runtime.clear_db_context()`?

pixeltable/exec/exec_node.py
```diff
@@ -56,19 +59,65 @@ def __aiter__(self) -> AsyncIterator[DataRowBatch]:
...
+
+    def _thread_iter(self) -> Iterator[DataRowBatch]:
+        # maxsize=2: we want a minimal amount of buffering
+        result_queue: queue.Queue = queue.Queue(maxsize=2)
+        caller_runtime = get_runtime()
+
+        def run() -> None:
+            thread_runtime = get_runtime()
+            # the execution needs to happen in the same db context as the caller, but on a new event loop
+            thread_runtime.copy_db_context(caller_runtime)
```
aaron-siegel, line 94: This seems a little fraught to me. What we really want to do is share the entire runtime, right? Because (if I understand things right) the new thread is nonconcurrent; it exists only to isolate the event loop.

Another option would be to use a more flexible mechanism for mapping threads onto runtimes, rather than the current hardcoded 1:1 mapping. Then "child threads" spawned for event loop purposes could simply inherit the runtime of their parent.

A downside of the present architecture, for example, is that it will cause a new client to be initialized every time one of these threads is spawned, which could end up being inefficient. We could go through and try to "share" all of the relevant state between runtimes, the same way we're doing with db context now, but it seems more robust (and future-proof) to just share the runtime.

pixeltable/exec/exec_node.py
```diff
@@ -56,19 +59,65 @@ def __aiter__(self) -> AsyncIterator[DataRowBatch]:
...
+            loop = get_runtime().event_loop
+            aiter = self.__aiter__()
+            try:
+                while True:
+                    batch: DataRowBatch = loop.run_until_complete(aiter.__anext__())
+                    yield batch
+            except StopAsyncIteration:
+                pass
+
+    _SENTINEL = object()
```
aaron-siegel: Might benefit from a more descriptive name, such as `_DONE_MARKER` or some such.

pierrebrunelle:
Tested locally ✅ and added updates to docs in that PR @mkornacker if you don't mind so I don't have to do that separately. I added a very clear section about why **sync endpoint and thread pool / uvloop and workers**: http://localhost:3333/howto/deployment/operations#web-framework-concurrency

pixeltable/exec/exec_node.py

pixeltable/exec/exec_node.py
mkornacker: Done.

pixeltable/exec/exec_node.py
mkornacker: Removed altogether, not really useful, since the thread simply goes away.

pixeltable/exec/exec_node.py
mkornacker: We can't share the clients, since these are can be tied to the event loop that already exists in the caller. We also can't share the event loop, because running on a separate one is exactly the point of this thread.

I added more comments. This is the inefficient path which is only needed in exceptional circumstances (jupyter, lancedb's create_table()) that aren't performance-sensitive. For Fastapi/uvloop, we run in the same thread.

pixeltable/exec/exec_node.py

PR #1173, opened 2026-03-03
PXT-1061 Handle non-path text strings in process_media_contents

amithadke:


pixeltable/functions/gemini.py
mkornacker: What does it return?

pixeltable/functions/gemini.py
```diff
@@ -487,12 +487,15 @@ def _process_media_contents(
     if isinstance(data, list):
         return [_process_media_contents(v, client, upload_tasks, large_video_paths) for v in data]
     if isinstance(data, str):
-        local_path = Path(data).expanduser()
-        if not local_path.exists():
-            return data
-        mime_type, _ = mimetypes.guess_type(str(local_path), strict=False)
+        mime_type, _ = mimetypes.guess_type(data, strict=False)
```
mkornacker, line 498: comment: what are you trying to do here?

tests/functions/test_gemini.py
```diff
@@ -6,6 +6,9 @@
 
 import pixeltable as pxt
 import pixeltable.type_system as ts
+from unittest.mock import MagicMock
+from pixeltable.functions.gemini import _process_media_contents
```
mkornacker: You can't import non-public functions for tests. Find an end-to-end test for this scenario.

amithadke: removed this test added long text case in test_generate_content

pixeltable/functions/gemini.py
amithadke: added doc string about return type and content

PR #1176, opened 2026-03-03
[PXT-1088] Fix RateLimitsScheduler over-scheduling causing 429 cascades

christopherpestano:
Note: We are also fixing perftest_providers in this PR as it had previously been broken by a change to how we validate udfs, however this is a minor bugfix of just moving the udfs to a separate file.

Validation of change was done through the usage of log debugging to identify reduction of 429s. Unit tests were also performed before and after change and results were measured both in terms of performance as well as number of retries. As can be seen in the images below after the change we experience a great reduction in total number of retried requests as well as total number of requests required to yield 50 rows. However, there was no difference in rows/second throughput, both roughly performed at the theoretical limit of 10 rows/minute aka the rate limit we imposed on the api key. 
Pre Change:
image
Post Change:
image

### DIFF
--- a/tests/functions/test_gemini.py
+++ b/tests/functions/test_gemini.py
@@ -291,6 +291,10 @@ class TestGemini:
         """
         Scheduler stress test: 30 rows on gemini-2.5-flash free tier (~10 RPM) triggers 429s
         and verifies that retry logic recovers all rows.
+
+        Meant to be used in conjunction with a free tier gemini api key, to manually verify that
+        the retry logic in RateLimitsScheduler is working properly. If the test fails with a non-zero
+        number of exceptions, check the logs to see if they were 429 errors and if retries were attempted.
         """
         skip_test_if_not_installed('google.genai')
         skip_test_if_no_client('gemini')
--- a/tests/functions/test_openai.py
+++ b/tests/functions/test_openai.py
@@ -17,14 +17,6 @@ from .tool_utils import run_tool_invocations_test, server_state, stock_price, we
 _logger = logging.getLogger('pixeltable')
 
 
-@pxt.udf
-def _throughput_test_prompt(word1: str, word2: str) -> list[dict[str, str]]:
-    return [
-        {'role': 'system', 'content': 'You are a helpful assistant. Be concise.'},
-        {'role': 'user', 'content': f'Use "{word1}" and "{word2}" in one short sentence.'},
-    ]
-
-
 @pytest.mark.remote_api
 @rerun(reruns=3, reruns_delay=8)
 class TestOpenai:
@@ -625,8 +617,14 @@ class TestOpenai:
 
     @pytest.mark.expensive
     def test_chat_completions_scheduler(self, uses_db: None) -> None:
-        """Scheduler throughput test: 20 rows through gpt-4o-mini to verify the rate-limit
-        scheduler dispatches and completes requests without errors."""
+        """
+        Scheduler throughput test: 20 rows through gpt-4o-mini to verify the rate-limit
+        scheduler dispatches and completes requests without errors.
+
+        Meant to be used in conjunction with a rate limit enabled openai api key, to manually verify that
+        the retry logic in RateLimitsScheduler is working properly. If the test fails with a non-zero
+        number of exceptions, check the logs to see if they were 429 errors and if retries were attempted.
+        """
         skip_test_if_not_installed('openai')
         skip_test_if_no_client('openai')
         from pixeltable.functions.openai import chat_completions
@@ -638,7 +636,12 @@ class TestOpenai:
         model = 'gpt-4o-mini'
 
         t = pxt.create_table('scheduler_tbl', {'word1': pxt.String, 'word2': pxt.String})
-        t.add_computed_column(prompt=_throughput_test_prompt(t.word1, t.word2))
+        t.add_computed_column(
+            prompt=[
+                {'role': 'system', 'content': 'You are a helpful assistant. Be concise.'},
+                {'role': 'user', 'content': 'Use ' + t.word1 + ' and ' + t.word2 + ' in one short sentence.'},
+            ]
+        )
         t.add_computed_column(response=chat_completions(t.prompt, model=model))
 
         rows = [{'word1': w1, 'word2': w2} for w1, w2 in (random.sample(wordlist, k=2) for _ in range(num_rows))]


pierrebrunelle:
@claude review

tests/functions/test_openai.py
```diff
@@ -486,3 +495,115 @@ def test_azure_openai(self, uses_db: None) -> None:
...
+        """
+        skip_test_if_not_installed('openai')
+        skip_test_if_no_client('openai')
+        from pixeltable.functions.openai import chat_completions
+
+        with open('tests/data/random_words', encoding='utf-8') as f:
+            wordlist = [w.strip() for w in f if w.strip() and not w.startswith('#')]
+
+        n = 20
+        model = 'gpt-5-nano'  # cheapest model; each row costs ~$0.0001
```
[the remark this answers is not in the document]
christopherpestano: Wrong

tests/functions/test_openai.py
```diff
@@ -486,3 +495,115 @@ def test_azure_openai(self, uses_db: None) -> None:
...
+            print(
+                f'\n  [{label}]'
+                f'\n    rows={n}, errors={status.num_excs}'
+                f'\n    elapsed={elapsed:.2f}s  ({n / elapsed:.2f} req/s)'
+            )
+            return elapsed, status.num_excs
+
+        elapsed_unconstrained, errs_unconstrained = _run_trial('unconstrained no max_tokens', max_tokens=None)
+
+        # Both trials should complete without total failure
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The docstring has been rewritten to describe the single trial behavior that the test actually implements.

tests/functions/test_openai.py
```diff
@@ -486,3 +495,115 @@ def test_azure_openai(self, uses_db: None) -> None:
...
+        (RateLimitsScheduler._exec) kicks in.
+
+        What to look for in output:
+          - errors=0  (all rows eventually succeed after retries)
+          - retries>0 (429s were encountered and retried)
+          - elapsed  (inflated by Bug 1: estimated_resource_refill_delay returns a delay
+                      much shorter than the actual window reset time, causing repeated
+                      429->retry->429 cascades before the window resets)
+
+        Approximate cost: ~$0.05 for 200 rows on gpt-4o-mini.
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The cost estimate now correctly says ~$0.50 for 2000 rows.

tests/functions/test_openai.py
```diff
@@ -486,3 +495,115 @@ def test_azure_openai(self, uses_db: None) -> None:
...
+
+    @pytest.mark.expensive
+    def test_chat_completions_throughput(self, uses_db: None) -> None:
+        """
+        Performance test: sends N chat_completions requests and reports throughput.
+
+        Runs two back-to-back trials to surface the token-estimate performance bug:
+          Trial 1 (constrained):   max_tokens=20  — scheduler estimate ≈ actual usage
+          Trial 2 (unconstrained): no max_tokens  — scheduler uses _default_max_tokens()
+                                                    which is 16384 for gpt-4o-mini, causing
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The docstring has been rewritten and no longer references specific default_max_tokens values or gpt-4o-mini.

tests/functions/test_openai.py
```diff
@@ -486,3 +495,115 @@ def test_azure_openai(self, uses_db: None) -> None:
...
+        concurrent requests. Each request consumes ~500 tokens, so a 200k TPM account
+        exhausts within the first ~400 requests. Once 429s start arriving, the retry logic
+        (RateLimitsScheduler._exec) kicks in.
+
+        What to look for in output:
+          - errors=0  (all rows eventually succeed after retries)
+          - retries>0 (429s were encountered and retried)
+          - elapsed  (inflated by Bug 1: estimated_resource_refill_delay returns a delay
+                      much shorter than the actual window reset time, causing repeated
+                      429->retry->429 cascades before the window resets)
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The "What to look for" section has been simplified to just list errors=0 as the success criterion, with no stale bug references.

tests/functions/test_gemini.py
```diff
@@ -267,3 +273,121 @@ def test_generate_embeddings_batch_api(self, uses_db: None) -> None:
...
+        All N requests fire concurrently. On the free tier (~10 RPM for gemini-2.5-flash),
+        the first ~10 succeed and the rest receive 429s.
+
+        Recovery path (GeminiRateLimitsInfo.get_retry_delay):
+          1. Parses `retryDelay` from exc.details['error']['details'] (Gemini-specific)
+          2. Falls back to exponential_backoff(attempt) if that field is absent
+
+        What to look for in output:
+          - errors=0  (all rows eventually succeed after retries)
+          - elapsed   (longer than n/RPM seconds due to retry waits and est_usage reset bug)
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The elapsed description now says "longer than RPM seconds due to retry waits" with no bug reference.

tool/perftest_providers.py
```diff
@@ -8,6 +8,12 @@
 from datetime import datetime, timedelta
 from typing import Any
 
+from perftest_udfs import (  # type: ignore[import-not-found]
+    create_chatgpt_prompt,
+    create_simple_messages_prompt,
+    create_simple_prompt,
+)
```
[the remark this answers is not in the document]
christopherpestano: Fixed, switched to relative import.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -103,37 +106,17 @@ async def _main_loop(self) -> None:
...
-            if len(aws) > 0:
-                # we have something to wait for
-                done, pending = await asyncio.wait(aws, return_when=asyncio.FIRST_COMPLETED)
-                for task in pending:
-                    task.cancel()
-                if completed_aw in done:
-                    _logger.debug(f'wait(): completed request for {self.resource_pool}')
+                await asyncio.sleep(resource_delay)
                 # re-evaluate current capacity for current item
                 continue
```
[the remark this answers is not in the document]
christopherpestano: This is intentional. The old wake-up mechanism was racy; it could wake up and re-evaluate before est_usage was actually decremented, since the Event was set from a different task. With the new per-request cost tracking in _inflight_costs, est_usage is accurately maintained, so the sleep duration from _resource_delay is based on correct in-flight estimates. The continue after sleep re-evaluates immediately. In practice resource_delay is short (sub-second) so the difference vs waking early is negligible.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -39,10 +39,13 @@ class RateLimitsScheduler(Scheduler):
 
     # scheduling-related state
     pool_info: env.RateLimitsInfo | None
-    est_usage: dict[str, int]  # value per resource; accumulated estimates since the last util. report
+    est_usage: dict[str, int]  # value per resource; running sum of estimated costs for in-flight requests
+
+    # Per-request estimated costs stored when a task is fired, keyed by id(request).
+    # Looked up and removed in _exec's finally block so we subtract only that request's contribution from est_usage.
+    _inflight_costs: dict[int, dict[str, int]]
```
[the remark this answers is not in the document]
christopherpestano: The est_usage accounting is straightforward arithmetic (increment on fire, decrement in finally). The mock scaffolding needed to unit test it would be significantly more complex than the logic itself. The meaningful validation for this change is integration behavior under actual rate limits, which the existing remote_api tests cover.

pixeltable/exec/expr_eval/schedulers.py
mkornacker: This needs to be updated with whatever new insights you gained (just so we don't make the same mistakes twice).

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -103,37 +106,17 @@ async def _main_loop(self) -> None:
             # check rate limits
             request_resources = self._get_request_resources(item.request)
             resource_delay = self._resource_delay(request_resources)
-            aws: list[Awaitable[None]] = []
-            completed_aw: asyncio.Task | None = None
-            wait_for_reset: asyncio.Task | None = None
             if resource_delay > 0:
-                # Some resource or resources are nearing depletion
-
-                if self.num_in_flight > 0:
```
mkornacker, line 111: So this isn't a good idea? Probably also good to explain the class docstring.

tests/functions/test_gemini.py
```diff
@@ -280,3 +285,89 @@ def test_generate_embeddings_batch_api(self, uses_db: None) -> None:
...
+        rows = [{'word1': w1, 'word2': w2} for w1, w2 in (random.sample(wordlist, k=2) for _ in range(n))]
+        t0 = time.monotonic()
+        t.insert(rows)
+        elapsed = time.monotonic() - t0
+        pxt.drop_table('perf_tbl')
+
+        _logger.debug(f'rows={n}, elapsed={elapsed:.2f}s  ({n / max(elapsed, 0.001):.2f} req/s)')
+
+    @pytest.mark.expensive
+    def test_generate_content_429_recovery(self, uses_db: None) -> None:
```
mkornacker: This is almost a verbatim copy of the preceding test. Factor out the common part.

tests/functions/test_openai.py
```diff
@@ -487,6 +500,94 @@ def test_azure_openai(self, uses_db: None) -> None:
...
+        rows = [{'word1': w1, 'word2': w2} for w1, w2 in (random.sample(wordlist, k=2) for _ in range(n))]
+        t0 = time.monotonic()
+        t.insert(rows)
+        elapsed = time.monotonic() - t0
+        pxt.drop_table('perf_tbl')
+
+        _logger.debug(f'rows={n}, elapsed={elapsed:.2f}s  ({n / max(elapsed, 0.001):.2f} req/s)')
+
+    @pytest.mark.expensive
+    def test_chat_completions_429_recovery(self, uses_db: None) -> None:
```
mkornacker: Too much code duplication.

christopherpestano: Fair enough there is a meaningful difference between the two tests here though, so I changed it to be parameterized as a single test to dedupe.

tests/functions/test_gemini.py
christopherpestano: Took a look, I don't think there is a meaningful enough difference here between the two tests to justify having two instead of the one (since unlike openai, we don't do gemini token rate limiting). Modified to be just one test

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: Put in the class docstring, this wasn't really a good idea. Before the in flight cost fix this would cause cascading failures since we would start incorrectly firing out requests since we zero'ed out the est cost. After the in flight cost fix we were just wasting cpu cycles here, since what we would see is that we'd send out 8 requests then for each one that came back it would wake this up and then immediately go back to sleep since the response arrived before the replenishment. And even if it did arrive after replenishment, that would mean that the wait_for_reset would usually trigger first since it wakes up after enough rate limit is freed up for a single request.

That's why I made the decision to remove the section entirely, we reduce code complexity by not having to wait on multiple events while not losing any performance since the resource_delay is sufficient to saturate the rate limit.

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: Added insights, basically three things
1) We always need to account for inflight requests, returning packets rate limits are stale unless they are the last request in a sequence
2) Early wakeup doesn't actually improve performance, the resource_delay is already waking up as soon as we are able to schedule a packet without a 429.
3) GIL locking means we are CPU bound here for large rate limits, in practice we aren't able to do more than 2k req per minute.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -30,19 +30,29 @@ class RateLimitsScheduler(Scheduler):
       (obtained via RateLimitsInfo.get_request_resources())
     - issue synchronous requests when we don't have a RateLimitsInfo yet or when we depleted a resource and need to
       wait for a reset
+    - does not wake early on in-flight request completion: API quota is refilled linearly by the provider, not
+      freed by returning responses, so waking on every return just wastes CPU to go back to sleep immediately
+    - est_usage must not be zeroed out when responses return: the provider's remaining quota (from response headers)
+      only reflects completed requests, not in-flight ones. Without est_usage we'd over-schedule based on stale
+      remaining values that don't account for ongoing requests
```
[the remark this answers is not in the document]
christopherpestano: Hmm actually a good suggestion, removed the depleted part

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -28,21 +28,30 @@ class RateLimitsScheduler(Scheduler):
     - try to stay below resource limits by utilizing reported RateLimitInfo.remaining
     - also take into account the estimated resource usage for in-flight requests
       (obtained via RateLimitsInfo.get_request_resources())
-    - issue synchronous requests when we don't have a RateLimitsInfo yet or when we depleted a resource and need to
-      wait for a reset
+    - issue a synchronous request when we don't have a RateLimitsInfo yet (first request bootstraps rate limit info)
+    - does not wake early on in-flight request completion: API quota is refilled linearly by the provider, not
```
sergey-mkhitaryan, line 32: It's not very interesting to read that this class does not do XYZ. I understand that you are changing this behavior, but think about when this pr will get merged. Please focus on what it does do, e.g. when it does wake up to send more requests. That is unless you want to highlight a difference with `RequestRateScheduler`, but I don't think that's the case.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -28,21 +28,30 @@ class RateLimitsScheduler(Scheduler):
...
+      freed by returning responses, so waking on every return just wastes CPU to go back to sleep immediately
+    - est_usage must not be zeroed out when responses return: the provider's remaining quota (from response headers)
+      only reflects completed requests, not in-flight ones. Without est_usage we'd over-schedule based on stale
+      remaining values that don't account for ongoing requests
 
     TODO:
     - limit the number of in-flight requests based on the open file limit
+    - single-threaded asyncio caps throughput at ~2k RPM even when providers allow 10k RPM (scheduling overhead
```
sergey-mkhitaryan, line 41: So this scheduler can only do ~33rps? That seems really bad, please file a jira instead of a TODO. Also not sure if single-threadedness is relevant or matters, Pixeltable is no longer single-threaded.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -28,21 +28,30 @@ class RateLimitsScheduler(Scheduler):
...
+      saturates one core). Address when we have free-threading (3.14t) or Rust kernels.
     """
 
     get_request_resources_param_names: list[str]  # names of parameters of RateLimitsInfo.get_request_resources()
 
     # scheduling-related state
     pool_info: env.RateLimitsInfo | None
-    est_usage: dict[str, int]  # value per resource; accumulated estimates since the last util. report
+    est_usage: dict[str, int]  # value per resource; running sum of estimated costs for in-flight requests
```
sergey-mkhitaryan: This always confused me, please rename to something like `_est_in_flight_usage`. "In flight" is the key.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -28,21 +28,30 @@ class RateLimitsScheduler(Scheduler):
...
+
+    # Per-request estimated costs stored when a task is fired, keyed by id(request).
+    # Looked up and removed in _exec's finally block so we subtract only that request's contribution from est_usage.
+    _inflight_costs: dict[int, dict[str, int]]
 
     num_in_flight: int  # unfinished tasks
```
sergey-mkhitaryan: do we still need `num_in_flight`? It might be the same as `len(_inflight_costs)`

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -28,21 +28,30 @@ class RateLimitsScheduler(Scheduler):
...
 
     get_request_resources_param_names: list[str]  # names of parameters of RateLimitsInfo.get_request_resources()
 
     # scheduling-related state
     pool_info: env.RateLimitsInfo | None
-    est_usage: dict[str, int]  # value per resource; accumulated estimates since the last util. report
+    est_usage: dict[str, int]  # value per resource; running sum of estimated costs for in-flight requests
+
+    # Per-request estimated costs stored when a task is fired, keyed by id(request).
+    # Looked up and removed in _exec's finally block so we subtract only that request's contribution from est_usage.
```
sergey-mkhitaryan: Too much detail, just say that this is used to maintain `est_usage` as accurately as possible.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -102,37 +111,17 @@ async def _main_loop(self) -> None:
...
                 _logger.debug(f'waiting {resource_delay:.1f}s for resource availability')
-
-            if len(aws) > 0:
-                # we have something to wait for
-                done, pending = await asyncio.wait(aws, return_when=asyncio.FIRST_COMPLETED)
-                for task in pending:
-                    task.cancel()
-                if completed_aw in done:
-                    _logger.debug(f'wait(): completed request for {self.resource_pool}')
+                await asyncio.sleep(resource_delay)
```
sergey-mkhitaryan, line 113: how does this pr address PXT-997? `RateLimitsScheduler` should get `rate_limit`/`rate_limits` from `Config` the way `RequestRateScheduler` does, and use that to limit self.

tests/functions/test_openai.py
```diff
@@ -487,6 +500,56 @@ def test_azure_openai(self, uses_db: None) -> None:
...
+    )
+    def test_chat_completions_scheduler(self, uses_db: None, num_rows: int, max_tokens: int | None) -> None:
+        """
+        Scheduler stress test: verifies throughput and 429 retry recovery.
+
+        Two scenarios:
+          - throughput: 20 rows, no max_tokens (tests conservative default estimator)
+          - 429_recovery: 2000 rows, max_tokens=500 (exhausts TPM, tests retry logic)
+
+        Run with:
```
aaron-siegel: No need to include this here; it will get highly redundant if we include instructions like this in every test. If we feel the pytest instructions are inadequate for Pixeltable let's try to address that in a more holistic way.

Also note that a simple `-m ''` will suffice.

tool/perftest_udfs.py
```diff
@@ -0,0 +1,31 @@
+"""UDFs for perftest_providers.py. Separated into a module to satisfy Pixeltable's UDF requirements."""
```
aaron-siegel, line 1: Did the perftests break inadvertently due to some intervening PR? We should consider adding a test in `tests/tool` that validates the perftest tool is still functional.

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: Fair but this was to address Marcel's comment about explaining in the class doc string as to why we no longer do the early wakeup. The importance of keeping this here even after the pr is merged is to block off the "bad" codepath to anyone who changes the RateLimitScheduler in the future. I think there is value in having this here

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: Yeah I think that's pretty bad too, made a Jira for addressing this. I'll keep the todo to follow the format of the other todo within this file. One thing to note too is that this portion of the code is technically single threaded, as this is all running within one event loop. While Pixeltable may not be single threaded, this specific aspect is single threaded.

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: Yeah that confused me too, changed to _est_in_flight_usage

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: Hmm yeah that makes sense, originally I kept it to align with how the exec node and other scheduler works, but it actually makes no sense in the context of the rate limit scheduler now. Removed

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: You're right, this PR doesn't address PXT-997 as written. The fix here is for the est_usage reset bug that caused over-scheduling and 429 cascades with is related but a different issue. I've relabeled this PR to a new ticket (PXT-1088) and moved PXT-997 back to the backlog for the Config-based rate limit cap work.

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: Done.

tests/functions/test_openai.py
christopherpestano: Removed

tool/perftest_udfs.py
christopherpestano: Added `tests/tool/test_perftest_tool.py` in this PR to validate the perftest tool.

tests/functions/test_gemini.py
```diff
@@ -281,6 +286,40 @@ def test_embed_content_batch_api(self, uses_db: None) -> None:
         res = t.select(t.rowid, t.text, sim=sim).order_by(sim, asc=False).collect()
         assert res[0]['rowid'] == 3
 
+    @pytest.mark.expensive
+    def test_generate_content_scheduler(self, uses_db: None) -> None:
+        """
+        Scheduler stress test: 30 rows on gemini-2.5-flash free tier (~10 RPM) triggers 429s
+        and verifies that retry logic recovers all rows.
```
sergey-mkhitaryan, line 293: Sorry, not loving the tests in this pr. They make some assumptions (free tier api key, some responses are 429s) but don't verify them. If these assumptions are no longer true in the future (e.g. free tier allows for a higher rate), we will silently lose coverage. They are also very expensive, as you pointed out.

Also, this particular test consistently fails on my laptop (`AssertionError: 30 rows failed permanently — retries did not recover them`) with my free tier api key.

You should be able to write a test with your own test UDF. I'm thinking a UDF that acquires a lock or a semaphore (so that the test can control when it blocks and when it returns). The test, or the UDF itself would also use something like an atomic counter to track how many UDF calls are currently in flight, and verify that those do not exceed the expected maximum based on the estimated usage and known limits.

I know this is not a trivial request, but I can't think of a more on point, fast, and inexpensive blackbox test one can write.

christopherpestano: Ok I see what you mean and removed the most expensive test (openai 429) and kept the rest. I've implemented a test that simulates the 429 logic as you said, however we don't need any locks as the actual execution logic you are referencing is all running within a single event loop so the operations we are doing in the test are atomic (since we don't yield midway & only one coroutine alters state at a time).

As to the 30 row failure I think that's probably an issue with your API key, the test works both in CI and locally for me, it would be odd even at the free tier for 30 rows to fail completely (considering the allowance is 10 rows per minute essentially).

As for the logic for keeping the other tests, even if they become meaningless no-ops because of some provider change, it is useful to have them around as a repro if we ever see users complaining about rate limiting issues for gemini or openai (two of the most common providers). Therefore these non-expensive tests can stay as a fallback checking us if we introduce a majorly breaking change.

sergey-mkhitaryan: Yes, the problem was with my api key. I get 100% 429s with my free tier key and no retries succeed. It may be that there's a per-account lifetime max requests that you can make without paying. Be aware of that if this test will at some point run with a free tier key (which at this moment is not the case).

Thanks for writing the scheduler test, I'll take a look. W.r.t. the utility of this one (and the openai one), I can only agree to disagree I'm afraid.

tests/tool/test_perftest_tool.py
```diff
@@ -0,0 +1,15 @@
+import subprocess
+
+import pytest
+
+from ..utils import skip_test_if_not_installed
+
+
+@pytest.mark.remote_api
+class TestPerftestTool:
```
sergey-mkhitaryan, line 11: please indicate that it's a smoke test, either by writing a comment, or simply by naming it ~ test_perftest_providers_smoke_test

tests/tool/test_perftest_tool.py
```diff
@@ -0,0 +1,15 @@
...
+    def test_perftest_tool(self) -> None:
+        skip_test_if_not_installed('openai')
+        skip_test_if_not_installed('google.genai')
```
sergey-mkhitaryan, line 14: is google.genai required if openai is selected?

tests/test_schedulers.py
```diff
@@ -1,17 +1,131 @@
...
+    """
+    pool_info = env.Env.get().get_resource_pool_info(_POOL_NAME, lambda: _TestRateLimitsInfo(_REQUEST_LIMIT))
+
+    current = _provider.enter()
+    try:
+        if current > _provider.capacity:
+            _provider.reject()
+            raise DummyError(f'simulated 429: {current} in-flight exceeds capacity {_provider.capacity}')
+
+        await asyncio.sleep(0.05)
```
sergey-mkhitaryan: I think the idea is correct here, but there are a few issues with the implementation.

First, sleeping in a test is _almost_ always not the best thing you can do. It makes no sense that if I change this to `sleep(2)`, the test starts to fail. In the best case it means that the test depends on a particular performance/timing of things, and can become flaky over time. In the worst case, it means that the test is just not validating what it needs to be.

I think it's the latter, because the other issue is that this test passes on `main`. It means that it doesn't really verify your fix.

I suspect that both issues have to do with this arbitrary sleep. Based on my understanding of the problem this pr solves, the test should instead:

1. Submit a large batch of UDF calls (way more than the scheduler should fire at a time, based on the resource numbers). Each UDF call blocks until the test tells it to proceed at step 4 (I previously incorrectly mentioned that you need a Semaphore for that, but actually `threading.Condition` is the right choice.)
2. Each UDF call increments a counter that tells the test how many were fired simultaneously.
3. The test verifies that the number of concurrent UDF calls is within the expectation. You need to come up with a way to verify that at the right time. If you do it too soon, it may be that too many tasks _are_ fired, but the test just doesn't wait long enough to observe that. If you wait for too long, the test will be unnecessarily slow (which is still better than not validating correctly). 
4. The test does `.notify_all()` on the shared `Condition`, finally unblocking the UDFs. 

There will likely be some multithreading involved, I don't know how else to do this.

Happy to jump on a call to discuss more.

tests/tool/test_perftest_tool.py
christopherpestano: Yeah the tool doesn't have any split import, so regardless of selection we need both.

tests/tool/test_perftest_tool.py
christopherpestano: Added comment indicating that it's smoke test.

tests/test_schedulers.py
christopherpestano: Went with a somewhat altered approach for getting the test to fail in main vs not fail in this branch. Rather than having the weak assertion of "less" failures, modified it such that we now assert no 429s at all. The provider mock simply gives the client the real rate limit it has internally and it is up to the client to respect that rate limit irrespective of timing. Take a look I think it models pretty well the actual behavior we have with openai/anthropic.

sergey-mkhitaryan: But it still depends on wall clock, which means it can behave differently depending on the environment (as the failure on windows confirms). I think you should steer clear from using time when it can result in a false positive.

sergey-mkhitaryan: In order to avoid the wall clock dependency, you can experiment with simulating it. I.e. `_ProviderSimulator` uses a variable instead of `time.monotonic()`, and it advances it every time the UDF is called. But again that would not be my top choice for testing this.

tests/tool/test_perftest_tool.py
```diff
@@ -0,0 +1,17 @@
+import subprocess
+
+import pytest
+
+from ..utils import skip_test_if_not_installed
+
+"""
+Smoke test for the perftest tool. This just runs the tool with a single request and checks that it completes
+"""
```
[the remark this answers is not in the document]
christopherpestano: ruff doesn't flag

tests/test_schedulers.py
```diff
@@ -62,3 +182,46 @@ async def run_test():
             loop.run_until_complete(run_test())
         finally:
             loop.close()
```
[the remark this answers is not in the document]
christopherpestano: ignore

tests/tool/test_perftest_tool.py
```diff
@@ -0,0 +1,17 @@
...
+@pytest.mark.remote_api
+class TestPerftestTool:
+    def test_perftest_tool(self) -> None:
+        skip_test_if_not_installed('openai')
+        skip_test_if_not_installed('google.genai')
+        subprocess.run(
+            ('python', '-m', 'tool.perftest_providers', '--provider', 'openai', '--n', '1', '--t', '10'), check=True
```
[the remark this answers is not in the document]
christopherpestano: added skip if no client

pixeltable/env.py
```diff
@@ -1101,8 +1101,8 @@ def estimated_resource_refill_delay(self, target_remaining: int) -> float | None
             return 0
         if self.request_start_ts >= self.reset_at:
             return 0
-        if self.limit < target_remaining:
-            return None
+        if self.remaining >= self.limit:
+            return None  # nothing consumed yet; can't estimate refill rate
 
         # Estimate resource refill rate based on the recorded state and timestamps. Assumes linear refill.
         refill_rate = (self.limit - self.remaining) / (self.reset_at - self.request_start_ts).total_seconds()
```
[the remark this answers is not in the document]
christopherpestano: bad

christopherpestano:
Removed `test_rate_limits_scheduler` because it tests nondeterministic behavior (wall-clock-dependent scheduling) with arbitrary tolerance (5% rejection threshold). The `_ProviderSimulator` uses `time.monotonic()` for its linear refill model, making the test timing-sensitive, already failing on Windows CI where `asyncio.sleep` has ~15.6ms timer resolution. Injecting a simulated clock into the simulator alone doesn't help because the scheduler itself also paces via `asyncio.sleep` and `time.monotonic`, so the two clocks would diverge. In this case we can't easily build a deterministic test that ends up simulating real behavior, best we could do in the future is create a separate CI pipeline that tests the provider integrations using lower tier provider keys to ensure no failures for regular users.

tests/functions/test_gemini.py
christopherpestano: Ultimately removed the scheduler test, explanation left at bottom.

tests/test_schedulers.py
christopherpestano: Ultimately removed the scheduler test, explanation left at bottom.

PR #1180, opened 2026-03-04
[PXT-1024] Catch up Reve UDFs with the API updates

sergey-mkhitaryan:


pixeltable/functions/reve.py
```diff
@@ -151,18 +159,22 @@ async def create(prompt: str, *, aspect_ratio: str | None = None, version: str |
 
         >>> t.add_computed_column(img=reve.create(t.prompt, aspect_ratio='1:1'))
     """
-    payload = {'prompt': prompt}
+    payload: dict = {'prompt': prompt}
```
aaron-siegel: payload: dict[str, Any]

tests/functions/test_reve.py
```diff
@@ -19,32 +19,53 @@ def test_create_edit_remix(self, default_params: bool, uses_db: None) -> None:
...
+        )
         t.add_computed_column(
-            just_logo=(
-                reve.edit(
-                    t.pixeltable_logo,
-                    'extract the company logo and drop the name',
-                    **({} if default_params else {'version': 'latest'}),
-                )
-            )
+            just_logo=(reve.edit(t.pixeltable_logo, 'extract the company logo and drop the name', **kwargs))
```
aaron-siegel, line 30: All UDFs should ideally be tested two ways:
- with only required parameters (no optional parameters)
- with all optional parameters

We've had bugs in both directions in the past, e.g., parameters that were intended to be optional but were inadvertently made required, that made it through CI because every test case supplied those optional arguments.

aaron-siegel (changes requested):

tests/functions/test_reve.py
sergey-mkhitaryan: But that's exactly the point of `kwargs`. What are you suggesting exactly?

aaron-siegel: Ah now I see, comment withdrawn.

PR #1184, opened 2026-03-05
Log thread name now that Pixeltable can be used in a multi-threaded f…

sergey-mkhitaryan:
…ashion

### DIFF
--- a/tool/print_random_ops_stats.py
+++ b/tool/print_random_ops_stats.py
@@ -6,7 +6,7 @@ from pixeltable.config import Config
 from tool.random_ops import TABLE_OPS
 
 LINE_PARSER = re.compile(
-    r'^(?P<timestamp>.+) (?P<pid>\d+) \[Worker (?P<worker_id>\d+)\] (?P<thread_name>\w+) '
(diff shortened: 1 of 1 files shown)


pixeltable/env.py
```diff
@@ -52,7 +52,7 @@ class Env:
     _instance: Env | None = None
     __initializing: bool = False
     _init_lock: threading.RLock = threading.RLock()
-    _log_fmt_str = '%(asctime)s %(levelname)s %(name)s %(filename)s:%(lineno)d: %(message)s'
+    _log_fmt_str = '%(asctime)s %(levelname)s %(threadName)s %(name)s %(filename)s:%(lineno)d: %(message)s'
```
mkornacker, line 55: Do threads have meaningful names? What does this look like?

sergey-mkhitaryan: That fully depends on what creates those threads. With FastApi, thread names are probably like `Thread-1`, `Thread-2`, etc. Which is still more meaningful than thread IDs.

But when Pixeltable is called from `main()`, it looks like this:

```
2026-03-09 15:48:11,813 DEBUG MainThread pixeltable catalog.py:428: invalidating table version...
```

PR #1190, opened 2026-03-08
PXT-968: bounding box udfs

mkornacker:
- new udfs bboxes_convert(), bboxes_resize(), including tests
- renamed draw_bounding_boxes() to bboxes_draw()
- provided stubs for the remaining bounding box udfs, so we can agree on the signatures

### DIFF
--- a/docs/public_api.opml
+++ b/docs/public_api.opml
@@ -253,6 +253,7 @@
         <outline text="module|pixeltable.functions.video">
           <outline text="iter|pixeltable.functions.video.frame_iterator" />
           <outline text="iter|pixeltable.functions.video.video_splitter" />
+          <outline text="uda|pixeltable.functions.video.concat_videos_agg" />
           <outline text="uda|pixeltable.functions.video.make_video" />
           <outline text="udf|pixeltable.functions.video.clip" />
           <outline text="udf|pixeltable.functions.video.crop" />
--- a/pixeltable/functions/video.py
+++ b/pixeltable/functions/video.py
@@ -500,24 +500,11 @@ def segment_video(
             _handle_ffmpeg_error(e)
 
 
-@pxt.udf(is_method=True)
-def concat_videos(videos: list[pxt.Video]) -> pxt.Video:
-    """
-    Merge multiple videos into a single video.
-
-    __Requirements:__
-
-    - `ffmpeg` needs to be installed and in PATH
-
-    Args:
-        videos: List of videos to merge.
-
-    Returns:
-        A new video containing the merged videos.
-    """
+def _concat_videos(videos: list[str], error_prefix: str) -> str | None:
+    """Concatenate videos and return the path to the output video, or None for an empty list"""
     Env.get().require_binary('ffmpeg')
     if len(videos) == 0:
-        raise pxt.Error('concat_videos(): empty argument list')
+        return None
 
     # Check that all videos have the same resolution
     resolutions: list[tuple[int, int]] = []
@@ -525,7 +512,7 @@ def concat_videos(videos: list[pxt.Video]) -> pxt.Video:
         metadata = av_utils.get_metadata(str(video))
         video_stream = next((stream for stream in metadata['streams'] if stream['type'] == 'video'), None)
         if video_stream is None:
-            raise pxt.Error(f'concat_videos(): file {video!r} has no video stream')
+            raise pxt.Error(f'{error_prefix}: file {video!r} has no video stream')
         resolutions.append((video_stream['width'], video_stream['height']))
 
     # check for divergence
@@ -533,7 +520,7 @@ def concat_videos(videos: list[pxt.Video]) -> pxt.Video:
     for i, (x, y) in enumerate(resolutions[1:], start=1):
         if (x0, y0) != (x, y):
             raise pxt.Error(
-                f'concat_videos(): requires that all videos have the same resolution, but:'
+                f'{error_prefix}: requires that all videos have the same resolution, but:'
                 f'\n  video 0 ({videos[0]!r}): {x0}x{y0}'
                 f'\n  video {i} ({videos[i]!r}): {x}x{y}.'
             )
@@ -549,7 +536,7 @@ def concat_videos(videos: list[pxt.Video]) -> pxt.Video:
     try:
         # First attempt: fast copy without re-encoding (works for compatible formats)
         cmd = ['ffmpeg', '-f', 'concat', '-safe', '0', '-i', str(filelist_path), '-c', 'copy', '-y', str(output_path)]
-        _logger.debug(f'concat_videos(): {" ".join(cmd)}')
+        _logger.debug(f'_concat_videos(): {" ".join(cmd)}')
         try:
             _ = subprocess.run(cmd, capture_output=True, text=True, check=True)
             return str(output_path)
@@ -604,6 +591,59 @@ def concat_videos(videos: list[pxt.Video]) -> pxt.Video:
         _handle_ffmpeg_error(e)
     finally:
         filelist_path.unlink()
+
+
+@pxt.udf(is_method=True)
+def concat_videos(videos: list[pxt.Video]) -> pxt.Video | None:
+    """
+    Merge multiple videos into a single video.
+
+    __Requirements:__
+
+    - `ffmpeg` needs to be installed and in PATH
+
+    Args:
+        videos: List of videos to merge.
+
+    Returns:
+        A new video containing the merged videos, or None if the input list is empty.
+    """
+    Env.get().require_binary('ffmpeg')
+    videos = [v for v in videos if v is not None]
+    return _concat_videos(videos, error_prefix='concat_videos()')
+
+
+@pxt.uda(requires_order_by=True)
+class concat_videos_agg(pxt.Aggregator):
+    """
+    Aggregate function that merges videos into a single video.
+
+    __Requirements:__
+
+    - `ffmpeg` needs to be installed and in PATH
+    - All videos must have the same resolution
+
+    Returns:
+        A new video containing all input videos concatenated in order, or None if all inputs are None.
+
+    Examples:
+        Concatenate all videos in a table, ordered by timestamp:
+
+        >>> tbl.select(concat_videos_agg(tbl.timestamp, tbl.video)).collect()
+    """
+
+    videos: list[str]
+
+    def __init__(self) -> None:
+        Env.get().require_binary('ffmpeg')
+        self.videos = []
+
+    def update(self, video: pxt.Video) -> None:
+        if video is not None:
+            self.videos.append(str(video))
+
+    def value(self) -> pxt.Video | None:
+        return _concat_videos(self.videos, error_prefix='concat_videos_agg()')
 
 
 @pxt.udf
--- a/tests/test_video.py
+++ b/tests/test_video.py
@@ -10,7 +10,7 @@ import pytest
 import pixeltable as pxt
 import pixeltable.functions as pxtf
 from pixeltable.env import Env
-from pixeltable.functions.video import frame_iterator, legacy_frame_iterator, video_splitter
+from pixeltable.functions.video import concat_videos_agg, frame_iterator, legacy_frame_iterator, video_splitter
 from pixeltable.utils.object_stores import ObjectOps
 
 from .utils import (
@@ -576,15 +576,16 @@ class TestVideo:
         )
         assert status.num_excs == 0
         res = u.select(
-            u.v1.get_metadata().streams[0].duration_seconds,
-            u.v2.get_metadata().streams[0].duration_seconds,
-            u.v3.get_metadata().streams[0].duration_seconds,
-            u.concat.get_metadata().streams[0].duration_seconds,
-        ).collect()
-        # Verify all videos were concatenated
-        durations = res.to_pandas().iloc[0]
-        concat_duration = durations.iloc[3]
-        assert concat_duration is not None
+            d1=u.v1.get_duration(), d2=u.v2.get_duration(), d3=u.v3.get_duration(), d_concat=u.concat.get_duration()
+        ).collect()[0]
+        input_duration = res['d1'] + res['d2'] + res['d3']
+        concat_duration = res['d_concat']
+        assert pytest.approx(input_duration, abs=0.1) == concat_duration
+
+        # empty inputs
+        validate_update_status(u.insert([{'v1': None, 'v2': None, 'v3': None}]), expected_rows=1)
+        res = u.where(u.v1 == None).select(u.concat).collect()
+        assert res[0]['concat'] is None
 
     def test_concat_videos_mixed_formats(self, uses_db: None, tmp_path: Path) -> None:
         from pixeltable.functions.video import concat_videos
@@ -629,6 +630,66 @@ class TestVideo:
             _ = t.add_computed_column(
                 concat=concat_videos([t.v1.astype(pxt.String), t.v2.astype(pxt.String), t.v3.astype(pxt.String)])
             )
+
+    def test_concat_videos_agg(self, uses_db: None) -> None:
+        video_filepaths = get_video_files()[:3]
+        t = pxt.create_table('concat_agg_test', {'id': pxt.Int, 'video': pxt.Video})
+        t.insert({'id': i, 'video': p} for i, p in enumerate(video_filepaths))
+
+        # split each video into segments, then reassemble with group_by
+        segments = pxt.create_view('segments', t, iterator=video_splitter(t.video, duration=5.0))
+        result = (
+            segments.select(segments.id, video=concat_videos_agg(segments.pos, segments.video_segment))
+            .group_by(segments.id)
+            .collect()
+        )
+        assert len(result) == len(video_filepaths)
+
+        # verify reassembled durations match the originals
+        u = pxt.create_table('concat_results', {'id': pxt.Int, 'video': pxt.Video})
+        u.insert(list(result))
+        u.add_computed_column(duration=u.video.get_duration())
+        orig_durations = t.select(duration=t.video.get_duration()).order_by(t.id).collect()
+        concat_durations = u.select(u.duration).order_by(u.id).collect()
+        assert all(
+            abs(orig - concat) < 0.1 for orig, concat in zip(orig_durations['duration'], concat_durations['duration'])
+        )
+
+        # empty group: all-None videos should produce None
+        e = pxt.create_table('concat_empty', {'id': pxt.Int, 'video': pxt.Video, 'pos': pxt.Int})
+        e.insert(
+            [
+                {'id': 0, 'video': video_filepaths[0], 'pos': 0},
+                {'id': 1, 'video': None, 'pos': 0},
+                {'id': 1, 'video': None, 'pos': 1},
+            ]
+        )
+        result = e.select(e.id, video=concat_videos_agg(e.pos, e.video)).group_by(e.id).order_by(e.id).collect()
+        assert len(result) == 2
+        assert result[0]['video'] is not None  # id 0: has a video
+        assert result[1]['video'] is None  # id 1: empty group returns None
+
+    def test_concat_videos_agg_mixed_formats(self, uses_db: None, tmp_path: Path) -> None:
+        # mixed audio
+        no_audio = generate_test_video(tmp_path, duration=1.0, has_audio=False)
+        with_audio = generate_test_video(tmp_path, duration=1.5, has_audio=True)
+
+        t = pxt.create_table('test_agg_mixed_audio', {'id': pxt.Int, 'video': pxt.Video})
+        t.insert([{'id': 0, 'video': no_audio}, {'id': 1, 'video': with_audio}, {'id': 2, 'video': no_audio}])
+        concat_output = concat_videos_agg(t.id, t.video)
+        result = t.select(output=concat_output, duration=concat_output.get_duration()).collect()
+        assert len(result) == 1
+        concat_duration = result[0]['duration']
+        assert abs(concat_duration - (1.0 + 1.5 + 1.0)) < 0.2
+
+        # error case: mixed resolutions
+        low_res = generate_test_video(tmp_path, duration=0.5, size='176x144', has_audio=False)
+        high_res = generate_test_video(tmp_path, duration=0.5, size='1920x1080', has_audio=False)
+
+        t2 = pxt.create_table('test_agg_resolution', {'id': pxt.Int, 'video': pxt.Video})
+        t2.insert([{'id': 0, 'video': low_res}, {'id': 1, 'video': high_res}])
+        with pytest.raises(pxt.Error, match='requires that all videos have the same resolution'):
+            t2.select(concat_videos_agg(t2.id, t2.video)).collect()
 
     def _validate_splitter_segments(
         self,


pixeltable/functions/vision.py
```diff
@@ -423,6 +425,379 @@ def draw_bounding_boxes(
...
+    if src_format == dst_format:
+        return bboxes
+    if src_format not in ['xyxy', 'xywh', 'cxcywh']:
+        raise pxt.Error(f'Invalid src_format: {src_format!r}')
+    if dst_format not in ['xyxy', 'xywh', 'cxcywh']:
+        raise pxt.Error(f'Invalid dst_format: {dst_format!r}')
+
+    _ = _validate_bboxes(bboxes, 'bboxes_convert()')
+    arr = np.array(bboxes)
+    assert arr.ndim == 2 and arr.shape[1] == 4
```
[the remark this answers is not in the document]
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -423,6 +425,379 @@ def draw_bounding_boxes(
...
+        bboxes: List of bounding boxes, each either specified with absolute pixel coordinates or relative
+            coordinates in [0, 1].
+        format: Format of the bounding box coordinates, one of 'xyxy', 'xywh', 'cxcywh'.
+        width: Target width in absolute pixels. Requires bboxes to be specified with absolute pixel coordinates.
+        width_f: Target width as a float. Requires bboxes to be specified with relative coordinates in [0, 1].
+        height: Target height in absolute pixels. Requires bboxes to be specified with absolute pixel coordinates.
+        height_f: Target height as a float. Requires bboxes to be specified with relative coordinates in [0, 1].
+        aspect: Target aspect ratio as a string 'W:H' (e.g., '16:9'). Resizes either the width
+            or height to match the specified aspect ratio, maintaining the other dimension. Requires exactly one of
+            `crop` or `pad` to be True.
```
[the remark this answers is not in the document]
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -423,6 +425,379 @@ def draw_bounding_boxes(
...
+        width: Canvas width in absolute pixels.
+        height: Canvas height in absolute pixels.
+        min_visibility: Minimum fraction of the bounding box that must be visible after clipping. If the visibility
+            is less than this value, returns None.
+        min_area: Minimum area of the bounding box after clipping. If the area is less than this value, returns None.
+
+    Returns:
+        List of clipped bounding boxes in the same format as the input.
+    """
+    return []
```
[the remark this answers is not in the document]
mkornacker: Done.

tests/functions/test_vision.py
```diff
@@ -124,21 +127,255 @@ def test_draw_bounding_boxes(self, uses_db: None) -> None:
...
             # labels don't match boxes
-            _ = v.select(draw_bounding_boxes(v.frame_s, boxes=v.detections_a.bboxes, labels=[2])).collect()
+            _ = v.select(bboxes_draw(v.frame_s, boxes=v.detections_a.bboxes, labels=[2])).collect()
         assert 'number of boxes and labels must match' in str(exc_info.value).lower()
 
         with pytest.raises(pxt.Error) as exc_info:
             # box_colors don't match boxes
-            _ = v.select(draw_bounding_boxes(v.frame_s, boxes=v.detections_a.bboxes, box_colors=['red'])).collect()
+            _ = v.select(bboxes_draw(v.frame_s, boxes=v.detections_a.bboxes, box_colors=['red'])).collect()
         assert 'number of boxes and box colors must match' in str(exc_info.value).lower()
```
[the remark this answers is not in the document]
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -423,6 +425,379 @@ def draw_bounding_boxes(
...
+
+    Returns:
+        List of bounding boxes in dst_format.
+    """
+    if src_format == dst_format:
+        return bboxes
+    if src_format not in ['xyxy', 'xywh', 'cxcywh']:
+        raise pxt.Error(f'Invalid src_format: {src_format!r}')
+    if dst_format not in ['xyxy', 'xywh', 'cxcywh']:
+        raise pxt.Error(f'Invalid dst_format: {dst_format!r}')
```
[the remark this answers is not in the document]
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -423,6 +425,379 @@ def draw_bounding_boxes(
...
+        canvas_width: Canvas width.
+        canvas_height: Canvas height.
+        canvas_region: Canvas region that was cropped, either specified with absolute pixel coordinates or relative
+            coordinates, in the format specified by `canvas_region_format`.
+        canvas_region_format: Format of the `canvas_region` coordinates, one of 'xyxy', 'xywh', 'cxcywh'.
+
+    Returns:
+        List of adjusted bounding boxes in the same format as the input. They can extend beyond the canvas boundaries.
+    """
+    return []
```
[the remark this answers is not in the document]
mkornacker: Done.

tests/functions/test_vision.py
```diff
@@ -112,7 +115,7 @@ def test_draw_bounding_boxes(self, uses_db: None) -> None:
         with pytest.raises(pxt.Error) as exc_info:
             # multiple color specifications
             _ = v.select(
-                draw_bounding_boxes(
+                bboxes_draw(
```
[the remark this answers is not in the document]
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -423,6 +425,379 @@ def draw_bounding_boxes(
...
+            coordinates in [0, 1].
+        format: Format of the bounding box coordinates, one of 'xyxy', 'xywh', 'cxcywh'.
+        factor: Scale factor to apply to both box dimensions.
+        x_factor: Scale factor to apply to the box width.
+        y_factor: Scale factor to apply to the box height.
+
+    Returns:
+        List of scaled bounding boxes in the same format as the input.
+    """
+    return []
```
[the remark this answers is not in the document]
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,385 @@ def draw_bounding_boxes(
...
+        aspect_f = float(match.group(1)) / float(match.group(2))
+    has_width = width is not None or width_f is not None
+    has_height = height is not None or height_f is not None
+    has_aspect = aspect is not None or aspect_f is not None
+    if has_width + has_height + has_aspect != 1:
+        raise pxt.Error('Exactly one of width/width_f, height/height_f, or aspect/aspect_f must be specified')
+    if has_aspect and aspect_mode is None:
+        raise pxt.Error("aspect_mode ('crop' or 'pad') is required when aspect is specified")
+    if not has_aspect and aspect_mode is not None:
+        raise pxt.Error('aspect_mode is only valid when aspect is specified')
```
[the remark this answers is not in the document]
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -8,15 +8,17 @@
 
 t = pxt.get_table(...)
 t.select(
-    pxtv.draw_bounding_boxes(t.img, boxes=t.boxes, label=t.labels)
+    pxtv.bboxes_draw(t.img, boxes=t.boxes, label=t.labels)
```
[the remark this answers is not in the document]
mkornacker: Done

docs/public_api.opml
```diff
@@ -375,7 +375,7 @@
         </outline>
         <outline text="module|pixeltable.functions.vision">
           <outline text="func|pixeltable.functions.vision.eval_detections" />
-          <outline text="func|pixeltable.functions.vision.draw_bounding_boxes" />
+          <outline text="func|pixeltable.functions.vision.bboxes_draw" />
```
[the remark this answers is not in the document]
mkornacker: Done.

tests/functions/test_vision.py
```diff
@@ -48,12 +57,11 @@ def test_eval(self, uses_db: None) -> None:
         _ = v.select(mean_ap(v.eval_b)).show()[0, 0]
 
         _ = v.select(
-            draw_bounding_boxes(v.frame_s, boxes=v.detections_a.bboxes, labels=v.detections_a.labels, fill=True)
+            bboxes_draw(v.frame_s, boxes=v.detections_a.bboxes, labels=v.detections_a.labels, fill=True)
         ).collect()
 
     def test_draw_bounding_boxes(self, uses_db: None) -> None:
```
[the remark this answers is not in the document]
mkornacker: Done

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,399 @@ def draw_bounding_boxes(
...
+        raise pxt.Error(f'Invalid src_format: {src_format!r}')
+    if dst_format not in ['xyxy', 'xywh', 'cxcywh']:
+        raise pxt.Error(f'Invalid dst_format: {dst_format!r}')
+    if src_format == dst_format:
+        return bboxes
+
+    _ = _validate_bboxes(bboxes, 'bboxes_convert()')
+    arr = np.array(bboxes)
+    assert arr.ndim == 2 and arr.shape[1] == 4
+    c0, c1, c2, c3 = arr[:, 0], arr[:, 1], arr[:, 2], arr[:, 3]
```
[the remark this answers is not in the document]
mkornacker: Done

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,383 @@ def draw_bounding_boxes(
...
+
+@pxt.udf
+def bboxes_crop_canvas(
+    bboxes: list,
+    format: Literal['xyxy', 'xywh', 'cxcywh'],
+    *,
+    canvas_width: int | None = None,
+    canvas_height: int | None = None,
+    canvas_region: list,
+    canvas_region_format: Literal['xyxy', 'xywh', 'cxcywh'],
```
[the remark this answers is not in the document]
mkornacker: Done

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,383 @@ def draw_bounding_boxes(
...
+        raise pxt.Error('Only one of width or width_f can be specified')
+    if height is not None and height_f is not None:
+        raise pxt.Error('Only one of height or height_f can be specified')
+    if aspect is not None and aspect_f is not None:
+        raise pxt.Error('Only one of aspect or aspect_f can be specified')
+    if aspect is not None:
+        match = ASPECT_RATIO_RE.match(aspect)
+        if match is None:
+            raise pxt.Error(f'Invalid aspect ratio: {aspect!r}; expected "W:H"')
+        aspect_f = float(match.group(1)) / float(match.group(2))
```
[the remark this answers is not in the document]
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,383 @@ def draw_bounding_boxes(
...
+    has_height = height is not None or height_f is not None
+    has_aspect = aspect is not None or aspect_f is not None
+    if has_width + has_height + has_aspect != 1:
+        raise pxt.Error('Exactly one of width/width_f, height/height_f, or aspect/aspect_f must be specified')
+    if has_aspect and aspect_mode is None:
+        raise pxt.Error("aspect_mode ('crop' or 'pad') is required when aspect is specified")
+    if not has_aspect and aspect_mode is not None:
+        raise pxt.Error('aspect_mode is only valid when aspect is specified')
+
+    is_absolute = _validate_bboxes(bboxes, 'bboxes_resize()')
```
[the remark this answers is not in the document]
mkornacker: Done

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,391 @@ def draw_bounding_boxes(
...
+        result = np.floor(result + 0.5).astype(int)
+    return result.tolist()
+
+
+ASPECT_RATIO_RE = re.compile(r'(\d+):(\d+)')
+
+
+@pxt.udf
+def bboxes_resize(
+    bboxes: list,  # should be: list[list[int]] | list[list[float]]
```
sergey-mkhitaryan, line 504: what is the purpose of these "should be" comments? can this just become the type hint?

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,391 @@ def draw_bounding_boxes(
...
+        result = np.column_stack([c0, c1, c0 + c2, c1 + c3])
+    elif src_format == 'xywh' and dst_format == 'cxcywh':
+        result = np.column_stack([c0 + c2 / 2, c1 + c3 / 2, c2, c3])
+    elif src_format == 'cxcywh' and dst_format == 'xyxy':
+        result = np.column_stack([c0 - c2 / 2, c1 - c3 / 2, c0 + c2 / 2, c1 + c3 / 2])
+    else:  # cxcywh -> xywh
+        result = np.column_stack([c0 - c2 / 2, c1 - c3 / 2, c2, c3])
+
+    if is_absolute:
+        result = np.floor(result + 0.5).astype(int)
```
sergey-mkhitaryan, line 495: `np.round()`?

pixeltable/functions/vision.py
mkornacker: No, our type system can't represent this.

I left this in there to remind ourselves to deal with it later.

pixeltable/functions/vision.py
mkornacker: Added comment.

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,392 @@ def draw_bounding_boxes(
...
+    """
+    if len(bboxes) == 0:
+        return []
+
+    if src_format not in ['xyxy', 'xywh', 'cxcywh']:
+        raise pxt.Error(f'Invalid src_format: {src_format!r}')
+    if dst_format not in ['xyxy', 'xywh', 'cxcywh']:
+        raise pxt.Error(f'Invalid dst_format: {dst_format!r}')
+    if src_format == dst_format:
+        return bboxes
```
sergey-mkhitaryan, line 472: should we still do the validation even if formats are the same?

pixeltable/functions/vision.py
```diff
@@ -298,7 +300,7 @@ def __create_label_colors(labels: list[Any]) -> dict[Any, str]:
 
 
 @pxt.udf
-def draw_bounding_boxes(
+def bboxes_draw(
     img: PIL.Image.Image,
     boxes: list[list[int]],
```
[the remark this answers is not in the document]
christopherpestano: Yeah copilot is correct here, this is an breaking api change, it is important that we at least attempt to have some backwards compatability otherwise it'll result in bad user experience.

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,392 @@ def draw_bounding_boxes(
...
+            dimension to match the aspect ratio. If `pad`, extends the undersized dimension to match the aspect ratio.
+
+    Returns:
+        List of resized bounding boxes in the same format as the input.
+    """
+    if len(bboxes) == 0:
+        return []
+
+    # TODO: this is a lot of repeated per-call validation; find a way to do this at plan generation time, where possible
+    if width is not None and width_f is not None:
```
sergey-mkhitaryan: needs a validation that w and h and their _f equivalents are positive

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,392 @@ def draw_bounding_boxes(
...
+    return result.tolist()
+
+
+ASPECT_RATIO_RE = re.compile(r'(\d+):(\d+)')
+
+
+@pxt.udf
+def bboxes_resize(
+    bboxes: list,  # should be: list[list[int]] | list[list[float]]
+    format: Literal['xyxy', 'xywh', 'cxcywh'],
```
sergey-mkhitaryan, line 505: `format` variable shadows the python's built-in with the same name

tests/functions/test_vision.py
```diff
@@ -109,36 +115,319 @@ def test_draw_bounding_boxes(self, uses_db: None) -> None:
...
+            pxt.drop_table(t)
+
+            input_bboxes = [convert_fmt(*b, fmt) for b in abs_boxes]
+            t = pxt.create_table('bbox_abs', {'bboxes': pxt.Json})
+            validate_update_status(t.insert([{'bboxes': input_bboxes}]), expected_rows=1)
+
+            # width
+            res = t.select(out=bboxes_resize(t.bboxes, fmt, width=50)).collect()
+            assert all(get_w(b, fmt) == 50 for b in res['out'][0])
+            assert all(get_aspect(b1, fmt) == get_aspect(b2, fmt) for b1, b2 in zip(input_bboxes, res['out'][0]))
```
sergey-mkhitaryan: use `pytest.approx` like you did for the height

tests/functions/test_vision.py
```diff
@@ -109,36 +115,319 @@ def test_draw_bounding_boxes(self, uses_db: None) -> None:
...
+            res = t.select(out=bboxes_resize(t.bboxes, fmt, height=100)).collect()
+            assert all(get_h(b, fmt) == 100 for b in res['out'][0])
+            assert all(
+                get_aspect(b1, fmt) == pytest.approx(get_aspect(b2, fmt), rel=0.01)
+                for b1, b2 in zip(input_bboxes, res['out'][0])
+            )
+
+            # aspect 16:9 crop
+            res = t.select(out=bboxes_resize(t.bboxes, fmt, aspect='16:9', aspect_mode='crop')).collect()
+            assert all(get_aspect(b, fmt) == pytest.approx(16 / 9, abs=1) for b in res['out'][0])
```
sergey-mkhitaryan: `abs=1` seems too high, no?

pixeltable/functions/vision.py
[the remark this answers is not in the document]
christopherpestano: Also, what was the reason for the rename? Looking through the diff for things such as the notebooks it seems like there's a good bit of noise that arises from this change. Perhaps it would be better to keep the previous draw_bounding_boxes name to reduce the amount of noise in this pr.

tests/functions/test_vision.py
```diff
@@ -109,36 +115,319 @@ def test_draw_bounding_boxes(self, uses_db: None) -> None:
...
+                        get_h(b_out, dst_fmt) == pytest.approx(get_h(b_in, src_fmt), abs=1e-9)
+                        for b_in, b_out in zip(input_bboxes, converted)
+                    )
+
+                    # round-trip: convert back dst -> src, should match original;
+                    # cxcywh with odd integer dimensions can shift position by 1
+                    # (half-pixel center rounded to int), so relax tolerance
+                    uses_cxcywh = 'cxcywh' in (src_fmt, dst_fmt)
+                    is_abs = boxes is abs_boxes
+                    delta = 1 if (uses_cxcywh and is_abs) else 1e-9
```
sergey-mkhitaryan, line 407: should this relaxed tolerance apply to the assertions right above?

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,392 @@ def draw_bounding_boxes(
...
+        # don't use round() here, it rounds to the nearest even number
+        result = np.floor(result + 0.5).astype(int)
+    return result.tolist()
+
+
+ASPECT_RATIO_RE = re.compile(r'(\d+):(\d+)')
+
+
+@pxt.udf
+def bboxes_resize(
```
christopherpestano, line 503: We are adding quite a few of these new udfs without adding a corresponding example in the notebooks, it would be nice to be able to quickly look at a notebook and see the visualized version of these to ensure that when applied to an image they look like they make sense. Might be fine for now though

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,392 @@ def draw_bounding_boxes(
...
 
+def _validate_bboxes(bboxes: list, error_prefix: str) -> bool:
+    """Check that bboxes are either all int or all float. Return True for absolute, False for relative."""
+    is_absolute = all(isinstance(x, int) and x >= 0 for x in itertools.chain.from_iterable(bboxes))
+    is_relative = all(isinstance(x, float) and 0.0 <= x <= 1.0 for x in itertools.chain.from_iterable(bboxes))
+    if not (is_absolute or is_relative):
+        raise pxt.Error(
+            f'{error_prefix}: bounding box coordinates must be either all int (>= 0) or all float (in [0, 1])'
+        )
+    if not all(len(b) == 4 for b in bboxes):
```
sergey-mkhitaryan: it would make more sense to me if the shape validation came first

pixeltable/functions/vision.py
mkornacker: Out of scope.

pixeltable/functions/vision.py
mkornacker: Done.

pixeltable/functions/vision.py
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,392 @@ def draw_bounding_boxes(
...
+        if is_absolute:
+            result = np.floor(result + 0.5).astype(int)
+
+    if not valid.all():
+        # leave invalid boxes as-is
+        result[~valid] = orig[~valid]
+    return result.tolist()
+
+
+@pxt.udf
```
christopherpestano, line 699: Question about strategy here, there are quite a few stubs for bbox transforms (resize, scale, pad, clip, crop canvas, resize canvas) that will each need similar coordinate math with edge cases around rounding, zero-width boxes, relative vs absolute coords, etc. Have we considered wrapping something like        albumentations instead of hand-rolling these? It already handles keeping bboxes in sync with image transforms and has been battle-tested pretty extensively.

pixeltable/functions/vision.py
mkornacker: Hm, it's called the same in torchvision's BoundingBox.

pixeltable/functions/vision.py
mkornacker: Done

pixeltable/functions/vision.py
mkornacker: re strategy: we want to minimize external dependencies. It doesn't seem like it should be too difficult to create correct implementations.

pixeltable/functions/vision.py
christopherpestano, line 503: Nit on the if/elif chains for format handling. Have we considered using a canonical intermediate form instead? Something like always converting to cxcywh internally, doing the operation, then converting back out. That way each function only needs one code path and the format conversion lives in one place. Would also make the stubs a lot simpler to implement since they wouldn't each need to handle all three formats.

mkornacker: It was a conscious choice not to do that, because it's less efficient - you need to compute a bunch of output you don't use.

That's how the code started out.

tests/functions/test_vision.py
mkornacker: Done.

tests/functions/test_vision.py
mkornacker: Done

tests/functions/test_vision.py
mkornacker: The test passes, so why not use the lower tolerance. The delta here specifically applies to the roundtrip effect.

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,392 @@ def draw_bounding_boxes(
...
+        src_format: Source format, one of 'xyxy', 'xywh', 'cxcywh'.
+        dst_format: Destination format, one of 'xyxy', 'xywh', 'cxcywh'.
+
+    Returns:
+        List of bounding boxes in dst_format.
+    """
+    if len(bboxes) == 0:
+        return []
+
+    if src_format not in ['xyxy', 'xywh', 'cxcywh']:
```
aaron-siegel: Nit: use a tuple here
```
if src_format not in ('xyxy', 'xywh', 'cxcywh'):
```
It's how we do it everywhere else, and it's (slightly) more efficient (tuple is a more efficient data structure, since it's immutable: for example, it will always allocate exactly as much memory as needed)

docs/public_api.opml
```diff
@@ -375,7 +375,9 @@
...
```
aaron-siegel: Nit: Keep UDFs in alphabetical order in public_api.opml (makes it easier to maintain)

pixeltable/functions/vision.py
aaron-siegel, line 7: The usual convention here (as used by other libaries) would be `pxtfv`. My preference is to always `from pixeltable import functions as pxtf` and then explicitly `pxtf.vision.bboxes_draw`; we use that pattern a lot, and direct renames of submodules rarely (I'm struggling to recall another place in the codebase where we use that pattern). We should think about establishing some clear naming standards to be consistent.

pixeltable/functions/vision.py
aaron-siegel: @mkornacker  Did you create a `documentation` ticket for it?

pixeltable/functions/vision.py
```diff
@@ -423,6 +428,392 @@ def draw_bounding_boxes(
...
+ASPECT_RATIO_RE = re.compile(r'(\d+):(\d+)')
+
+
+@pxt.udf
+def bboxes_resize(
+    bboxes: list,  # should be: list[list[int]] | list[list[float]]
+    format: Literal['xyxy', 'xywh', 'cxcywh'],
+    *,
+    width: int | None = None,
+    width_f: float | None = None,
```
aaron-siegel: I don't love this API. To me these are somewhat confusing/overlapping parameters, and it'd be cleaner (and more consistent with similar occurrences in the SDK) to use a polymorphic UDF: one signature for "absolute"/integer values, and one for floating-point values, as in
```
@pxt.udf
def bboxes_resize(
   ...,
   width: int | None,
   height: int | None,
   aspect: str | None,
   ...
)

@bboxes_resize.overload
def _(
    ...,
    width: float | None,
    height: float | None,
    aspect: float | None,
    ...
)
```

pixeltable/functions/vision.py
mkornacker: Done

docs/public_api.opml
mkornacker: Done

pixeltable/functions/vision.py
mkornacker: Done

pixeltable/functions/vision.py
mkornacker: Done

PR #1191, opened 2026-03-09
[PXT-836] Decouple resource estimation from rate-limits pool

christopherpestano:
Moves resource estimation from `RateLimitsInfo` (the shared pool class) to per-UDF `@fn.resource_estimator` decorators on `Function`. Previously, `get_request_resources` lived on the pool and its parameter signature was cached once from whichever function initialized the pool first. This caused an `AssertionError` when a second function with a different signature (e.g., `vision` vs `chat_completions`) shared the same pool. Now each function owns its own estimator, and the scheduler dispatches to the correct one per-request. Decoration-time validation ensures estimator parameters are a subset of the UDF's signature.

### DIFF
--- a/docs/public_api.opml
+++ b/docs/public_api.opml
@@ -158,6 +158,7 @@
           <outline text="udf|pixeltable.functions.image.mode" />
         </outline>
         <outline text="module|pixeltable.functions.json">
+          <outline text="iter|pixeltable.functions.json.list_iterator" />
           <outline text="udf|pixeltable.functions.json.dumps" />
           <outline text="udf|pixeltable.functions.json.make_list" />
         </outline>
--- a/pixeltable/catalog/view.py
+++ b/pixeltable/catalog/view.py
@@ -285,9 +285,13 @@ class View(Table):
             for col in columns:
                 if col.name in md['columns'] and tv.is_iterator_column(col):
                     md['columns'][col.name]['is_iterator_col'] = True
-            # Build the iterator expression string: "IteratorName(arg1=expr1, arg2=expr2)"
-            args_str = ', '.join(f'{k}={v.display_str(inline=False)}' for k, v in tv.iterator_call.bound_args.items())
-            md['iterator_call'] = f'{tv.iterator_call.it.name}({args_str})'
+            # Build the iterator expression string: "iterator_name(arg1, arg2=expr2, ...)"
+            arg_strs: list[str] = []
+            for arg_expr in tv.iterator_call.args:
+                arg_strs.append(arg_expr.display_str(inline=True))
+            for arg_name, arg_expr in tv.iterator_call.kwargs.items():
+                arg_strs.append(f'{arg_name}={arg_expr.display_str(inline=True)}')
+            md['iterator_call'] = f'{tv.iterator_call.it.name}({", ".join(arg_strs)})'
 
         return md
 
--- a/pixeltable/func/iterator.py
+++ b/pixeltable/func/iterator.py
@@ -57,11 +57,11 @@ class PxtIterator(abc.ABC, Iterator[T], Generic[T]):
         raise NotImplementedError()
 
     @classmethod
-    def validate(cls, bound_args: dict[str, Any]) -> None:
+    def validate(cls, literal_args: dict[str, Any]) -> None:
         pass
 
     @classmethod
-    def conditional_output_schema(cls, bound_args: dict[str, Any]) -> dict[str, type] | None:
+    def conditional_output_schema(cls, bound_args: dict[str, 'exprs.Expr']) -> dict[str, type] | None:
         return None
 
 
@@ -84,7 +84,7 @@ class GeneratingFunction:
     is_legacy_retrofit: bool
 
     _default_output_schema: dict[str, ts.ColumnType] | None
-    _conditional_output_schema: Callable[[dict[str, Any]], dict[str, type]] | None
+    _conditional_output_schema: Callable[[dict[str, 'exprs.Expr']], dict[str, type]] | None
     _validate: Callable[[dict[str, Any]], bool] | None
 
     def __init__(self, decorated_callable: Callable, unstored_cols: list[str], fqn: str | None = None) -> None:
@@ -198,7 +198,7 @@ class GeneratingFunction:
                 )
             self._default_output_schema[name] = col_type
 
-    def call_output_schema(self, bound_kwargs: dict[str, Any]) -> dict[str, ts.ColumnType]:
+    def call_output_schema(self, bound_args: dict[str, 'exprs.Expr']) -> dict[str, ts.ColumnType]:
         if self._conditional_output_schema is None:
             if self._default_output_schema is None:
                 raise excs.Error(
@@ -207,13 +207,19 @@ class GeneratingFunction:
             return self._default_output_schema
 
         else:
-            output_schema = self._conditional_output_schema(bound_kwargs)
+            output_schema = self._conditional_output_schema(bound_args)
             if output_schema is None:
                 raise excs.Error(
                     f'The `conditional_output_schema` for iterator `{self.fqn}` returned None; '
                     'it must return a valid output schema dictionary.'
                 )
-            return {name: ts.ColumnType.from_python_type(type_) for name, type_ in output_schema.items()}
+            result: dict[str, ts.ColumnType] = {}
+            for name, type_ in output_schema.items():
+                if isinstance(type_, ts.ColumnType):
+                    result[name] = type_
+                else:
+                    result[name] = ts.ColumnType.from_python_type(type_)
+            return result
 
     def bind_to_signature(self, args: list['exprs.Expr'], kwargs: dict[str, 'exprs.Expr']) -> dict[str, 'exprs.Expr']:
         try:
@@ -242,7 +248,7 @@ class GeneratingFunction:
         if self._validate is not None:
             self._validate(literal_args)
 
-        output_schema = self.call_output_schema(literal_args)
+        output_schema = self.call_output_schema(bound_args)
 
         # a component view exposes the pos column of its rowid;
         # we create that column here, so it gets assigned a column id;
@@ -266,8 +272,9 @@ class GeneratingFunction:
         it.py_sig = inspect.signature(iterator_cls)
         it.unstored_cols = []
 
-        def call_output_schema(bound_kwargs: dict[str, Any]) -> dict[str, ts.ColumnType]:
-            schema, _ = iterator_cls.output_schema(**bound_kwargs)
+        def call_output_schema(bound_args: dict[str, exprs.Expr]) -> dict[str, ts.ColumnType]:
+            literal_args = {k: v.val for k, v in bound_args.items() if isinstance(v, exprs.Literal)}
+            schema, _ = iterator_cls.output_schema(**literal_args)
             return schema
 
         it.call_output_schema = call_output_schema  # type: ignore[method-assign]
@@ -284,8 +291,8 @@ class GeneratingFunction:
 
     # conditional_output_schema decorator
     def conditional_output_schema(
-        self, fn: Callable[[dict[str, Any]], dict[str, type]]
-    ) -> Callable[[dict[str, Any]], dict[str, type]]:
+        self, fn: Callable[[dict[str, 'exprs.Expr']], dict[str, type]]
+    ) -> Callable[[dict[str, 'exprs.Expr']], dict[str, type]]:
         if self._conditional_output_schema is not None:
             raise excs.Error(f'@pxt.iterator `{self.fqn}` already defines a `conditional_output_schema()` method.')
         self._conditional_output_schema = fn
@@ -333,7 +340,7 @@ class InvalidGeneratingFunction(GeneratingFunction):
     def eval(self, bound_args: dict[str, Any]) -> Iterator[dict]:
         raise excs.Error(f'The iterator `{self.fqn}` cannot be used, because\n{self.error_msg}')
 
-    def call_output_schema(self, bound_kwargs: dict[str, Any]) -> dict[str, ts.ColumnType]:
+    def call_output_schema(self, bound_args: dict[str, 'exprs.Expr']) -> dict[str, ts.ColumnType]:
         raise excs.Error(f'The iterator `{self.fqn}` cannot be used, because\n{self.error_msg}')
 
     def _validate(self, bound_args: dict[str, Any]) -> None:
@@ -415,17 +422,17 @@ class GeneratingFunctionCall:
             ).strip()
             return cls(it, args, kwargs, {}, outputs, validation_error)
 
-        literal_args = {k: v.val for k, v in bound_args.items() if isinstance(v, exprs.Literal)}
-        for param_name, param in it.py_sig.parameters.items():
-            if param_name not in bound_args and param.default is not inspect.Parameter.empty:
-                literal_args[param_name] = param.default
-        output_schema = it.call_output_schema(literal_args)
+        output_schema = it.call_output_schema(bound_args)
 
         if outputs is None:
             # For legacy iterators, `outputs` was not persisted, and there is no practical way to reconstruct it as
             # part of the schema migration. In that case we just query the iterator for it, but we lose the ability to
             # sanity check against any schema evolution of the iterator.
             if it.is_legacy_retrofit:
+                literal_args = {k: v.val for k, v in bound_args.items() if isinstance(v, exprs.Literal)}
+                for param_name, param in it.py_sig.parameters.items():
+                    if param_name not in bound_args and param.default is not inspect.Parameter.empty:
+                        literal_args[param_name] = param.default
                 _, unstored_cols = it.decorated_callable.output_schema(literal_args)  # type: ignore[attr-defined]
             else:
                 unstored_cols = it.unstored_cols
@@ -434,7 +441,7 @@ class GeneratingFunctionCall:
             }
             outputs.update(
                 {
-                    name: IteratorOutput(orig_name=name, is_stored=(name in unstored_cols), col_type=col_type)
+                    name: IteratorOutput(orig_name=name, is_stored=(name not in unstored_cols), col_type=col_type)
                     for name, col_type in output_schema.items()
                 }
             )
--- a/pixeltable/functions/document.py
+++ b/pixeltable/functions/document.py
@@ -13,7 +13,7 @@ from bs4.element import NavigableString, Tag
 from pypdfium2 import PdfDocument  # type: ignore[import-untyped]
 
 import pixeltable as pxt
-from pixeltable import exceptions as excs, type_system as ts
+from pixeltable import exceptions as excs, exprs, type_system as ts
 from pixeltable.env import Env
 from pixeltable.utils.documents import get_document_handle
 from pixeltable.utils.spacy import get_spacy_model
@@ -116,7 +116,7 @@ def _parse_elements(elements: list[Literal['text', 'image']] | None) -> list[Ele
 
 _HTML_HEADINGS = {'h1', 'h2', 'h3', 'h4', 'h5', 'h6'}
 
-_METADATA_COLUMN_TYPES: dict = {
+_METADATA_COLUMN_TYPES = {
     ChunkMetadata.TITLE: ts.String | None,
     ChunkMetadata.HEADING: ts.Json | None,
     ChunkMetadata.SOURCELINE: ts.Int | None,
@@ -559,17 +559,25 @@ class document_splitter(pxt.PxtIterator):
             Env.get().require_package('tiktoken')
 
     @classmethod
-    def conditional_output_schema(cls, bound_args: dict[str, Any]) -> dict[str, type]:
+    def conditional_output_schema(cls, bound_args: dict[str, exprs.Expr]) -> dict[str, type]:
         schema: dict[str, type] = {}
-        elements = _parse_elements(bound_args.get('elements', ['text']))
+
+        elements = [Element.TEXT]
+        if bound_args.get('elements') is not None:
+            if not isinstance(bound_args['elements'], exprs.Literal):
+                raise excs.Error('document_splitter(): `elements` must be a constant expression.')
+            elements = _parse_elements(bound_args['elements'].val)
         for element in elements:
             if element == Element.TEXT:
                 schema['text'] = ts.String
             elif element == Element.IMAGE:
                 schema['image'] = ts.Image
 
-        md_fields = _parse_metadata(bound_args.get('metadata', ''))
-        for md_field in md_fields:
-            schema[md_field.name.lower()] = _METADATA_COLUMN_TYPES[md_field]
+        if bound_args.get('metadata') is not None:
+            if not isinstance(bound_args['metadata'], exprs.Literal):
+                raise excs.Error('document_splitter(): `metadata` must be a constant expression.')
+            md_fields = _parse_metadata(bound_args['metadata'].val)
+            for md_field in md_fields:
+                schema[md_field.name.lower()] = _METADATA_COLUMN_TYPES[md_field]  # type: ignore[assignment]
 
         return schema
--- a/pixeltable/functions/huggingface.py
+++ b/pixeltable/functions/huggingface.py
@@ -7,7 +7,7 @@ first `pip install transformers` (or in some cases, `sentence-transformers`, as 
 UDFs).
 """
 
-from typing import Any, Callable, Literal, TypedDict, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Literal, TypedDict, TypeVar
 
 import av
 import numpy as np
@@ -23,6 +23,9 @@ from pixeltable.utils.code import local_public_names
 from pixeltable.utils.local_store import TempStore
 
 T = TypeVar('T')
+
+if TYPE_CHECKING:
+    from transformers import DetrConfig
 
 
 @pxt.udf(batch_size=32)
@@ -211,6 +214,21 @@ class DetrForObjectDetectionResponse(TypedDict):
     boxes: list[list[float]]
 
 
+def _detr_config(model_id: str, revision: str) -> 'DetrConfig':
+    """Load DetrConfig with workaround for dilation=None validation error.
+
+    The no_timm revision of facebook/detr-resnet-50 stores dilation: null
+    in its config, but newer huggingface_hub versions reject None for the
+    bool-typed field.
+    """
+    from transformers import DetrConfig
+
+    config = DetrConfig.from_pretrained(model_id, revision=revision)
+    if config.dilation is None:
+        config.dilation = False
+    return config
+
+
 @pxt.udf(batch_size=4)
 def detr_for_object_detection(
     image: Batch[PIL.Image.Image], *, model_id: str, threshold: float = 0.5, revision: str = 'no_timm'
@@ -262,9 +280,12 @@ def detr_for_object_detection(
     from transformers import DetrForObjectDetection, DetrImageProcessor
 
     model = _lookup_model(
-        model_id, lambda x: DetrForObjectDetection.from_pretrained(x, revision=revision), device=device
+        model_id,
+        lambda x: DetrForObjectDetection.from_pretrained(x, revision=revision, config=_detr_config(x, revision)),
+        device=device,
+        cache_key=(model_id, DetrForObjectDetection.from_pretrained, device, ('revision', revision)),
     )
-    processor = _lookup_processor(model_id, lambda x: DetrImageProcessor.from_pretrained(x, revision=revision))
+    processor = _lookup_processor(model_id, DetrImageProcessor.from_pretrained, revision=revision)
     normalized_images = [normalize_image_mode(img) for img in image]
 
     with torch.no_grad():
@@ -1122,14 +1143,12 @@ def text_to_image(
 
     pipeline = _lookup_model(
         model_id,
-        lambda x: AutoPipelineForText2Image.from_pretrained(
-            x,
-            dtype=torch.float16 if device == 'cuda' else torch.float32,
-            device_map='auto' if device == 'cuda' else None,
-            safety_checker=None,  # Disable safety checker for performance
-            requires_safety_checker=False,
-        ),
+        AutoPipelineForText2Image.from_pretrained,
         device=device,
+        dtype=torch.float16 if device == 'cuda' else torch.float32,
+        device_map='auto' if device == 'cuda' else None,
+        safety_checker=None,
+        requires_safety_checker=False,
     )
 
     try:
@@ -1320,14 +1339,12 @@ def image_to_image(
 
     pipeline = _lookup_model(
         model_id,
-        lambda x: AutoPipelineForImage2Image.from_pretrained(
-            x,
-            dtype=torch.float16 if device == 'cuda' else torch.float32,
-            device_map='auto' if device == 'cuda' else None,
-            safety_checker=None,  # Disable safety checker for performance
-            requires_safety_checker=False,
-        ),
+        AutoPipelineForImage2Image.from_pretrained,
         device=device,
+        dtype=torch.float16 if device == 'cuda' else torch.float32,
+        device_map='auto' if device == 'cuda' else None,
+        safety_checker=None,
+        requires_safety_checker=False,
     )
 
     try:
@@ -1568,12 +1585,10 @@ def image_to_video(
 
     pipe = _lookup_model(
         model_id,
-        lambda x: StableVideoDiffusionPipeline.from_pretrained(
-            x,
-            torch_dtype=torch.float16 if device == 'cuda' else torch.float32,
-            variant='fp16' if device == 'cuda' else None,
-        ),
+        StableVideoDiffusionPipeline.from_pretrained,
         device=device,
+        torch_dtype=torch.float16 if device == 'cuda' else torch.float32,
+        variant='fp16' if device == 'cuda' else None,
     )
 
     try:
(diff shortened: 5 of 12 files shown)


pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -148,16 +142,17 @@ def _resources(self) -> Collection[str]:
...
 
     def _get_request_resources(self, request: FnCallArgs) -> dict[str, int]:
-        # Fall back to default estimate if the request's function signature doesn't match the pool's estimator
-        if not all(name in request.fn_call.fn.signature.parameters for name in self.get_request_resources_param_names):
-            return dict.fromkeys(self._resources, 1)
-        kwargs_batch = request.fn_call.get_param_values(self.get_request_resources_param_names, request.rows)
+        estimator = request.fn_call.fn._resource_estimator
+        param_names = [p.name for p in inspect.signature(estimator).parameters.values()]
+        if len(param_names) == 0:
+            return estimator()
```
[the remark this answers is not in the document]
christopherpestano: Good catch. Fixed by filtering the estimator output to only include keys present in self._resources before passing to _resource_delay().

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -148,16 +142,17 @@ def _resources(self) -> Collection[str]:
...
+        kwargs_batch = request.fn_call.get_param_values(param_names, request.rows)
```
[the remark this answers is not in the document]
christopherpestano: inspect.signature() on a plain Python function is microsecond scale, and this runs once per scheduling decision which itself involves async I/O waits that are orders of magnitude slower. Not worth optimizing here.

pixeltable/func/function.py
```diff
@@ -429,6 +435,25 @@ def resource_pool(self, fn: Callable[..., str]) -> Callable[..., str]:
     def __default_resource_pool(self) -> str | None:
         return None
 
+    def resource_estimator(self, fn: Callable[..., dict[str, int]]) -> Callable[..., dict[str, int]]:
+        """Instance decorator for specifying the resource estimator of this function.
+
+        The decorated function accepts a subset of this function's parameters and returns a dict mapping
+        resource names to estimated costs for a single request.
```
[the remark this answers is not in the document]
christopherpestano: Added test_resource_estimator_validation in tests/test_function.py covering: valid estimator with param subset, valid zero arg estimator, rejection of params not in the function signature, and rejection of polymorphic functions.

pixeltable/func/function.py
```diff
@@ -429,6 +435,25 @@ def resource_pool(self, fn: Callable[..., str]) -> Callable[..., str]:
...
+        """
+        estimator_params = set(inspect.signature(fn).parameters.keys())
+        fn_params = set(self.signature.parameters.keys())
+        if not estimator_params.issubset(fn_params):
+            raise excs.Error(
```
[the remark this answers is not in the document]
christopherpestano: Good catch. Added an explicit is_polymorphic check that raises excs.Error with a clear message instead of hitting a bare AssertionError.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -148,16 +142,21 @@ def _resources(self) -> Collection[str]:
...
+            kwargs_batch = request.fn_call.get_param_values(param_names, request.rows)
+            if not request.is_batched:
+                result = estimator(**kwargs_batch[0])
+            else:
+                batch_kwargs = {k: [d[k] for d in kwargs_batch] for k in kwargs_batch[0]}
+                constant_kwargs, batch_kwargs = request.pxt_fn.create_batch_kwargs(batch_kwargs)
+                result = estimator(**constant_kwargs, **batch_kwargs)
+        # Filter to resources known to the pool to avoid KeyError in _resource_delay()
+        known = self._resources
+        return {k: v for k, v in result.items() if k in known}
```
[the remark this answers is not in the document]
christopherpestano: No, this would be a behavior change, when a udf opts out of this path the current semantics is that it shouldn't do the rate limiting logic (for example how gemini works now). Won't fix

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -148,16 +142,21 @@ def _resources(self) -> Collection[str]:
...
-            batch_kwargs = {k: [d[k] for d in kwargs_batch] for k in kwargs_batch[0]}
-            constant_kwargs, batch_kwargs = request.pxt_fn.create_batch_kwargs(batch_kwargs)
-            return self.pool_info.get_request_resources(**constant_kwargs, **batch_kwargs)
+            kwargs_batch = request.fn_call.get_param_values(param_names, request.rows)
+            if not request.is_batched:
+                result = estimator(**kwargs_batch[0])
+            else:
+                batch_kwargs = {k: [d[k] for d in kwargs_batch] for k in kwargs_batch[0]}
+                constant_kwargs, batch_kwargs = request.pxt_fn.create_batch_kwargs(batch_kwargs)
+                result = estimator(**constant_kwargs, **batch_kwargs)
```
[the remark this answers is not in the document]
christopherpestano: The decorator-time validation (param subset, polymorphic rejection) is covered in test_resource_estimator_validation. The scheduler dispatch path is exercised by the existing remote_api tests against OpenAI. Adding a local unit test here would require mocking FnCallArgs, FunctionCall, DataRow, and pool state for minimal additional coverage over what already exists. Not worth the overhead.

pixeltable/exprs/function_call.py
```diff
@@ -125,7 +125,7 @@ def __init__(
         bindings = fn.signature.py_signature.bind(*self.arg_idxs, **self.kwarg_idxs)
         self.bound_idxs = bindings.arguments
 
-        # Separately generate bound_args for purposes of determining the resource pool.
+        # Separately generate bound_args for purposes of determining the resource pool and estimator.
```
sergey-mkhitaryan: what role does `bound_args` play in determining the estimator?

pixeltable/functions/gemini.py
```diff
@@ -52,13 +52,6 @@ def _genai_client() -> 'genai.client.Client':
 
 
 class GeminiRateLimitsInfo(env.RateLimitsInfo):
-    def __init__(self) -> None:
-        super().__init__(self._get_request_resources)
-
-    def _get_request_resources(self) -> dict[str, int]:
-        # TODO(PXT-996): Improve resource tracking for Gemini UDFs
```
sergey-mkhitaryan, line 79: can you keep the TODO somewhere?

pixeltable/env.py
```diff
@@ -968,18 +968,14 @@ class RateLimitsInfo:
 
     Subclasses provide operational customization via:
     - get_retry_delay()
-    - get_request_resources(self, ...) -> dict[str, int]
-    with parameters that are a subset of those of the udf that creates the subclass's instance
+    - record_exc()
+
+    Resource estimation is handled per-UDF via the @fn.resource_estimator decorator on Function,
```
sergey-mkhitaryan: since the resource estimation is not here, no need to talk about it

tests/functions/test_openai.py
```diff
@@ -488,12 +488,10 @@ def test_azure_openai(self, uses_db: None) -> None:
         assert 'Mesopotamia' in result['chat_output'][0]['choices'][0]['message']['content']
 
     def test_shared_rate_limits_pool_different_signatures(self, uses_db: None) -> None:
-        """Verify that functions sharing a rate-limits pool with different get_request_resources signatures
-        don't crash the scheduler.
+        """Verify that functions sharing a rate-limits pool with different resource estimators work correctly.
```
sergey-mkhitaryan, line 663: This test made sense before this change, but I don't think that it's needed anymore.

pixeltable/func/function.py
```diff
@@ -42,15 +42,20 @@ class Function(ABC):
...
     # Returns the resource pool to use for calling this function with the given arguments.
-    # Overriden for specific Function instances via the resource_pool() decorator. The override must accept a subset
+    # Overridden for specific Function instances via the resource_pool() decorator. The override must accept a subset
     # of the parameters of the original function, with the same type.
     _resource_pool: Callable[..., str | None]
 
+    # Returns estimated resources needed for a specific request as a dict (key: resource name, value: estimated cost).
+    # Overridden for specific Function instances via the resource_estimator() decorator. The override must accept a
+    # subset of the parameters of the original function.
+    _resource_estimator: Callable[..., dict[str, int]]
```
sergey-mkhitaryan, line 57: since `scheduler` accesses it, it's not private. let's remove `_` and move it up.

tests/test_function.py
```diff
@@ -1319,6 +1319,65 @@ def test_udf_in_global_namespace(self, uses_db: None) -> None:
...
+            @_estimator_poly_fn.resource_estimator
+            def _() -> dict[str, int]:
+                return {'requests': 1}
+
+
+# Module-level UDFs for test_resource_estimator_validation
+
+
+@pxt.udf
+def _estimator_fn(prompt: str, model: str = 'gpt-4') -> str:
```
sergey-mkhitaryan: please try to use the existing udfs in this file when possible. for example, there's `overloaded_udf` that can replace `_estimator_poly_fn`.

pixeltable/func/function.py
```diff
@@ -429,6 +435,27 @@ def resource_pool(self, fn: Callable[..., str]) -> Callable[..., str]:
...
+        if self.is_polymorphic:
+            raise excs.Error(f'resource_estimator cannot be used with polymorphic function {self.self_path or self}')
+        estimator_params = set(inspect.signature(fn).parameters.keys())
+        fn_params = set(self.signature.parameters.keys())
+        if not estimator_params.issubset(fn_params):
+            raise excs.Error(
+                f'resource_estimator for {self.self_path or self} has parameters '
+                f'{estimator_params - fn_params} that are not in the function signature'
+            )
+        self._resource_estimator = fn
```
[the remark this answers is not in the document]
christopherpestano: Ignore

pixeltable/functions/gemini.py
christopherpestano: Added back

pixeltable/env.py
christopherpestano: Makes sense, removed comment

tests/functions/test_openai.py
christopherpestano: Pushing back on this one. The change fixes an assertion error when functions with different signatures share a rate limits pool, so this test is needed to confirm that the fix works correctly.

pixeltable/func/function.py
christopherpestano: Renamed to resource_estimator_fn & moved, I originally went with _resource_estimator since the resource_estimator name is the decorator

tests/test_function.py
christopherpestano: My worry was that we shouldn't randomly mutate existing test udfs with resource_estimators, but I realized I could just define the resource_estimator locally to the test, so did that instead of defining new udfs.

pixeltable/exprs/function_call.py
christopherpestano: I see, removed the reference to estimator. I realize now that the estimator gets its values at execution time through get_param_values() in the scheduler, so bound_args is only used for the resource pool.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -148,16 +142,21 @@ def _resources(self) -> Collection[str]:
...
+        # Filter to resources known to the pool to avoid KeyError in _resource_delay()
```
sergey-mkhitaryan: curious what is the scenario that leads to unrecognized resources appearing here?

pixeltable/func/function.py
```diff
@@ -429,6 +434,27 @@ def resource_pool(self, fn: Callable[..., str]) -> Callable[..., str]:
...
         return None
 
+    def resource_estimator(self, fn: Callable[..., dict[str, int]]) -> Callable[..., dict[str, int]]:
+        """Instance decorator for specifying the resource estimator of this function.
+
+        The decorated function accepts a subset of this function's parameters and returns a dict mapping
+        resource names to estimated costs for a single request.
+        """
+        if self.is_polymorphic:
+            raise excs.Error(f'resource_estimator cannot be used with polymorphic function {self.self_path or self}')
```
sergey-mkhitaryan: this is a new limitation, right? not saying it's a bad idea necessarily, but we should definitely check with somebody from the leadership if this is ok. I don't have an intuition for this.

pixeltable/func/function.py
```diff
@@ -429,6 +434,27 @@ def resource_pool(self, fn: Callable[..., str]) -> Callable[..., str]:
     def __default_resource_pool(self) -> str | None:
         return None
 
+    def resource_estimator(self, fn: Callable[..., dict[str, int]]) -> Callable[..., dict[str, int]]:
```
sergey-mkhitaryan, line 438: something I just realized, do you know if this works correctly with non-module-level UDFs?

The metadata refers to module-level UDFs by just module name + function name, or something like that. The UDF's code is expected to always be there.

UDFs that are not defined on the module level (e.g. a udf defined in a notebook) get a special treatment -- we serialize and deserialize them with pickle (CallableFunction.to_store/from_store). I'm now wondering if such a udf's resource estimator will not survive a kernel restart.

Note: the long term plan is to get rid of pickle, but I don't believe that we have any specific design or ideas on how to do that.

aaron-siegel: It should work fine.

pixeltable/func/function.py
```diff
@@ -71,6 +75,7 @@ def __init__(
         self.__resolved_fns = []
         self._to_sql = self.__default_to_sql
         self._resource_pool = self.__default_resource_pool
+        self.resource_estimator_fn = self.__default_resource_estimator
```
aaron-siegel: Please try to use consistent nomenclature. The output of `@resource_pool` is called `_resource_pool`, not `resource_pool_fn` (likewise `_to_sql`, not `to_sql_fn`), so this (exactly analogous) member variable should be called `_resource_estimator`.

If you believe `to_sql_fn`, `resource_pool_fn` and `resource_estimator_fn` are better names, the preferred way to address it is to open a discussion in `#pixeltable_dev` about changing the pattern globally, rather than introducing an inconsistency in the code base.

pixeltable/func/function.py
aaron-siegel: @christopherpestano Why is it necessary?

christopherpestano:
It is not necessary to disinclude polymorphic functions, however it does make it a bit more difficult to account for  mismatched parameters. Although the way pixeltable works means that we get only a single resolved Function expression to actually execute, if the estimator wasn't designed with the fn params in mind we could get some undefined behavior. Doing a check at decoration time is problematic since it is a bit inelegant to have to check for matching params both in the overload as well as the resource estimator definition (since we can't presume ordering). The solution I'm going to go with here is to have a runtime check that estimator params are a subset of the resolved functions params, without type checking. It will be up to the "user" (likely us since normal users likely aren't defining their own resource estimators) to ensure that the params are named appropriately as well as the estimator has branching logic for all the possible subtypes of its params.

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: For example Gemini doesn't give you any rate limit info but you may want to have the resource estimator have some default value e.g. {'request': 1} that only makes sense when some config is given. So for this we ignore these useful values in the case we don't have this in our config and don't ignore in the case we do have this in our config.

pixeltable/func/function.py
christopherpestano: My bad, fixed nomenclature.

pixeltable/func/function.py
christopherpestano: It wasn't but it was convenient, changed it such that we are able to use polymorphic functions.

tests/test_function.py
```diff
@@ -1319,6 +1319,35 @@ def test_udf_in_global_namespace(self, uses_db: None) -> None:
...
+            return {'requests': 1, 'tokens': a}
+
+        assert self.f1.resource_estimator_fn(5) == {'requests': 1, 'tokens': 5}
+
+        # Valid: zero-arg estimator
+        @self.f2.resource_estimator
+        def _() -> dict[str, int]:
+            return {'requests': 1}
+
+        assert self.f2.resource_estimator_fn() == {'requests': 1}
```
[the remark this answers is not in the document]
christopherpestano: ignore

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -147,16 +141,37 @@ def _resources(self) -> Collection[str]:
...
         else:
-            batch_kwargs = {k: [d[k] for d in kwargs_batch] for k in kwargs_batch[0]}
-            constant_kwargs, batch_kwargs = request.pxt_fn.create_batch_kwargs(batch_kwargs)
-            return self.pool_info.get_request_resources(**constant_kwargs, **batch_kwargs)
+            # If the estimator declares '_param_types', inject a dict mapping param names to Pixeltable
+            # ColumnTypes for the resolved overload. This lets a shared estimator on a polymorphic function
+            # distinguish overloads that share the same Python types (e.g., str for Document vs Video).
+            if '_param_types' in param_names:
+                param_names = [n for n in param_names if n != '_param_types']
+            extra = set(param_names) - set(request.fn_call.fn.signature.parameters.keys())
```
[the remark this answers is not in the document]
christopherpestano: Good comment, removed artifact

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -147,16 +141,35 @@ def _resources(self) -> Collection[str]:
...
+            if '_param_types' in inspect.signature(estimator).parameters:
+                param_types = {name: p.col_type for name, p in request.fn_call.fn.signature.parameters.items()}
+                for d in kwargs_batch:
+                    d['_param_types'] = param_types
+            if not request.is_batched:
+                result = estimator(**kwargs_batch[0])
+            else:
+                batch_kwargs = {k: [d[k] for d in kwargs_batch] for k in kwargs_batch[0]}
+                constant_kwargs, batch_kwargs = request.pxt_fn.create_batch_kwargs(batch_kwargs)
+                result = estimator(**constant_kwargs, **batch_kwargs)
```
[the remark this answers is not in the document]
christopherpestano: That muddies it up too much.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -147,16 +141,35 @@ def _resources(self) -> Collection[str]:
...
+                )
+            kwargs_batch = request.fn_call.get_param_values(param_names, request.rows)
+            if '_param_types' in inspect.signature(estimator).parameters:
+                param_types = {name: p.col_type for name, p in request.fn_call.fn.signature.parameters.items()}
+                for d in kwargs_batch:
+                    d['_param_types'] = param_types
+            if not request.is_batched:
+                result = estimator(**kwargs_batch[0])
+            else:
+                batch_kwargs = {k: [d[k] for d in kwargs_batch] for k in kwargs_batch[0]}
```
sergey-mkhitaryan: Would it be possible to find better names for these vars? `kwargs_batch` vs `batch_kwargs` is pretty confusing.

Also, on the next line, the meaning of `batch_kwargs` changes, so it would be good to have a separate variables for what is was and for what it becomes.

pixeltable/exec/expr_eval/schedulers.py
```diff
@@ -147,16 +141,35 @@ def _resources(self) -> Collection[str]:
...
-            return self.pool_info.get_request_resources(**kwargs_batch[0])
+        estimator = request.fn_call.fn._resource_estimator
+        param_names = [p.name for p in inspect.signature(estimator).parameters.values() if p.name != '_param_types']
+        if len(param_names) == 0:
+            result = estimator()
         else:
-            batch_kwargs = {k: [d[k] for d in kwargs_batch] for k in kwargs_batch[0]}
-            constant_kwargs, batch_kwargs = request.pxt_fn.create_batch_kwargs(batch_kwargs)
-            return self.pool_info.get_request_resources(**constant_kwargs, **batch_kwargs)
+            # If the estimator declares '_param_types', inject a dict mapping param names to Pixeltable
```
sergey-mkhitaryan: this comment should be at the `if '_param_types' in` line, not here

pixeltable/func/function.py
```diff
@@ -34,6 +34,10 @@ class Function(ABC):
     is_method: bool
     is_property: bool
     is_deterministic: bool
+    # Returns estimated resources needed for a specific request as a dict (key: resource name, value: estimated cost).
+    # Overridden for specific Function instances via the resource_estimator() decorator. The override must accept a
```
sergey-mkhitaryan: the override or the estimator?

pixeltable/func/function.py
```diff
@@ -429,6 +434,25 @@ def resource_pool(self, fn: Callable[..., str]) -> Callable[..., str]:
     def __default_resource_pool(self) -> str | None:
         return None
 
+    def resource_estimator(self, fn: Callable[..., dict[str, int]]) -> Callable[..., dict[str, int]]:
+        """Instance decorator for specifying the resource estimator of this function.
```
sergey-mkhitaryan: I'd just call it a "decorator". I'm not sure what makes this an "instance decorator"

pixeltable/func/function.py
```diff
@@ -429,6 +434,25 @@ def resource_pool(self, fn: Callable[..., str]) -> Callable[..., str]:
...
+
+        The decorated function accepts a subset of this function's parameters and returns a dict mapping
+        resource names to estimated costs for a single request.
```
sergey-mkhitaryan: not necessarily a single request, you support batch requests, too

pixeltable/func/function.py
```diff
@@ -429,6 +434,25 @@ def resource_pool(self, fn: Callable[..., str]) -> Callable[..., str]:
...
+
+        Parameters are validated at runtime, requiring the resource estimator contain a subset of the
```
sergey-mkhitaryan, line 444: I think it would be better if it said "accept" instead of "contain"

pixeltable/functions/openai.py
```diff
@@ -561,6 +527,37 @@ async def chat_completions(
...
+
+    max_completion_tokens = model_kwargs.get('max_completion_tokens')
+    max_tokens = model_kwargs.get('max_tokens')
+    n = model_kwargs.get('n')
+
+    completion_tokens = (n or 1) * (max_completion_tokens or max_tokens or _default_max_tokens(model))
+
+    num_tokens = 0.0
+    for message in messages:
+        num_tokens += 4  # every message follows <im_start>{role/name}\n{content}<im_end>\n
```
sergey-mkhitaryan, line 543: nit: num_tokens = len(messages) * 4

pixeltable/functions/openai.py
```diff
@@ -561,6 +527,37 @@ async def chat_completions(
...
+                num_tokens += len(value) / 4
+            elif isinstance(value, list):
+                for part in value:
+                    if isinstance(part, dict):
+                        if part.get('type') == 'text' and isinstance(part.get('text'), str):
+                            num_tokens += len(part['text']) / 4
+                        elif part.get('type') == 'image_url' and isinstance(part.get('image_url'), PIL.Image.Image):
+                            num_tokens += _token_count_for_image(part['image_url'])
+            if key == 'name':  # if there's a name, the role is omitted
+                num_tokens -= 1  # role is always required and always 1 token
```
sergey-mkhitaryan, line 555: This confused me. It's odd to be subtracting anything from num tokens. I think you should do +3 per message instead, and also +1 `if 'name' not in message.items()`.

Also, "role is always required" contradicts "the role is omitted" on the line above.

pixeltable/functions/openai.py
```diff
@@ -561,6 +527,37 @@ async def chat_completions(
...
+    if model_kwargs is None:
+        model_kwargs = {}
+
+    max_completion_tokens = model_kwargs.get('max_completion_tokens')
+    max_tokens = model_kwargs.get('max_tokens')
+    n = model_kwargs.get('n')
+
+    completion_tokens = (n or 1) * (max_completion_tokens or max_tokens or _default_max_tokens(model))
+
+    num_tokens = 0.0
```
sergey-mkhitaryan, line 541: I have a few comments about this code, and I know that you're not the original author, but still. Can this be an `int` instead?

tests/test_function.py
```diff
@@ -1319,6 +1319,68 @@ def test_udf_in_global_namespace(self, uses_db: None) -> None:
         assert process.returncode != 0  # The script should fail with an appropriate error message
         assert "Defining the UDF 'inline_udf' directly in the global namespace of a Python script" in process.stderr
 
+    def test_resource_estimator_polymorphic(self, uses_db: None) -> None:
```
sergey-mkhitaryan, line 1330: can you write a test with a non-polymorphic udf, and a test with a batch udf?

tests/test_function.py
```diff
@@ -1319,6 +1319,68 @@ def test_udf_in_global_namespace(self, uses_db: None) -> None:
...
+async def _(video: pxt.Video, model: str = 'default') -> list[float]:  # noqa: RUF029
+    init_test_pool('rate-limits:test-embed')
+    return [0.7, 0.8, 0.9]
+
+
+@mock_embed.resource_estimator
+def _(content: str, _param_types: dict) -> dict[str, int]:
+    if isinstance(_param_types.get('content'), ts.ImageType):
+        return {'requests': 1, 'tokens': 1000}
+    return {'requests': 1, 'tokens': len(content) // 4}
```
sergey-mkhitaryan, line 1418: assert that _param_types.get('content') is a StringType

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: moved

pixeltable/func/function.py
christopherpestano: The estimator is overridden, every function gets a default estimator and when you define a non-default one then you override the default with your own.

pixeltable/func/function.py
christopherpestano: This is the common naming convention for to_sql and resource_pool, an instance decorator is a class based decorator that wraps a function. Function is a class, this decorator modifies that object and is used to wrap python functions thus an instance decorator.

pixeltable/func/function.py
christopherpestano: Makes sense changed

pixeltable/functions/openai.py
christopherpestano: Putting this into a separate pr, on Jira https://pixeltable.atlassian.net/browse/PXT-1100

pixeltable/functions/openai.py
christopherpestano: Putting this into a separate pr, on Jira https://pixeltable.atlassian.net/browse/PXT-1100

pixeltable/functions/openai.py
christopherpestano: Putting this into a separate pr, on Jira https://pixeltable.atlassian.net/browse/PXT-1100

tests/test_function.py
christopherpestano: added

tests/test_function.py
christopherpestano: Added.

pixeltable/exec/expr_eval/schedulers.py
christopherpestano: Sure changed kwargs_batch to be per row kwargs, and batch_kwargs to be columnar kwargs which is used to create the real batch kwargs

pixeltable/func/function.py
christopherpestano: Fixed

pixeltable/func/function.py
sergey-mkhitaryan: Yeah I saw that the precedent above, but I would rather call it a nested decorator, or, per Claude's suggestion, although not the best IMO, an "attribute decorator". I'm going to leave it up to you. Unlike "instance decorator", "class-based decorator" or "class decorator" has a specific meaning in Python, but `resource_estimator` is not it.

tests/test_function.py
```diff
@@ -1319,6 +1319,111 @@ def test_udf_in_global_namespace(self, uses_db: None) -> None:
...
+
+    pool_info = env.Env.get().get_resource_pool_info(pool_name, env.RateLimitsInfo)
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
+    pool_info.record(now, requests=(100, 99, now))
+
+
+# Ruff ignore: async is required for the rate-limits scheduler path we could put some trival
+# await, but regardless it would be a hack to trigger the rate limited path.
+@pxt.udf(is_deterministic=False, resource_pool='rate-limits:test-embed')
+async def mock_embed(content: str, model: str = 'default') -> list[float]:  # noqa: RUF029
```
sergey-mkhitaryan: It's surprising that only async UDFs go through resource management. How hard it is to implement it for sync UDFs as well?

At very least we should raise an error if `@resource_estimator` is used on a sync UDF, otherwise, IIUC, it is silently ignored.

christopherpestano: There would only be two places where we would have to make major changes, one of those is the evaluator branching logic. Currently we branch off of is_async to even try to go down the rate limit path but it would be pretty easy to change the branching logic. The other is the Scheduler would be a bit more difficult to rewrite because the whole logic there is based off of assuming that the functions passed to it are all asynchronous, but it could be easy enough to just have it branch on function execution that if the function is sync put it in a thread or smth (or just accept the event loop will be blocked for function execution).

christopherpestano: Added change so that validate at runtime that if the function has an estimator it is asynchronous.

pixeltable/func/function.py
christopherpestano: Ok switching to just decorator then.

pixeltable/func/function.py
```diff
@@ -429,6 +435,25 @@ def resource_pool(self, fn: Callable[..., str]) -> Callable[..., str]:
...
+    def resource_estimator(self, fn: Callable[..., dict[str, int]]) -> Callable[..., dict[str, int]]:
+        """Decorator for specifying the resource estimator of this function.
+
+        The decorated function accepts a subset of this function's parameters and returns a dict mapping
+        resource names to estimated costs for a single or set of requests.
+
+        Parameters are validated at runtime, requiring the resource estimator contain a subset of the
+        parameters of the resolved function type.
+
+        If the estimator declares a parameter named '_param_types', it will receive a dict mapping
```
aaron-siegel, line 447: `ColumnType` isn't part of the public API, so this isn't suitable as a user-facing interface. Maybe we consider this feature as being only available to internal implementations for the time being.

pixeltable/functions/openai.py
aaron-siegel: Just to add some more color here, it's an estimated token count (based on a very crude estimate of character count / 4). There's no contradiction in having the *estimate* be a float that gets rounded at the end (so that competing fractions can cancel each other out in the meantime).

I'd argue this is the "correct" way to handle estimation, although for practical purposes it probably makes little difference.

tests/test_function.py
```diff
@@ -1319,10 +1319,155 @@ def test_udf_in_global_namespace(self, uses_db: None) -> None:
         assert process.returncode != 0  # The script should fail with an appropriate error message
         assert "Defining the UDF 'inline_udf' directly in the global namespace of a Python script" in process.stderr
 
+    def test_resource_estimator_polymorphic(self, uses_db: None) -> None:
+        """resource_estimator with _param_types works for valid overloads and fails for mismatched ones."""
+        from .utils import get_image_files, get_video_files
```
aaron-siegel: Move to top of file

tests/test_function.py
```diff
@@ -1319,10 +1319,155 @@ def test_udf_in_global_namespace(self, uses_db: None) -> None:
...
+
+        # Text overload: estimator param 'content' matches the resolved signature
+        t_text = pxt.create_table('test_est_text', {'content': pxt.String})
+        t_text.add_computed_column(emb=mock_embed(t_text.content))
+        status = t_text.insert([{'content': 'hello world'}, {'content': 'foo bar'}])
+        assert status.num_excs == 0
```
aaron-siegel: Use validate_update_status() helper function instead of manual asserts (here and below)

tests/test_function.py
```diff
@@ -1319,10 +1319,155 @@ def test_udf_in_global_namespace(self, uses_db: None) -> None:
...
+
+        # polymorphic resolved to a sync overload
+        t3 = pxt.create_table('test_sync_poly_rp', {'num': pxt.Int})
+        t3.add_computed_column(result=sync_poly_udf_with_rp(t3.num))
+        with pytest.raises(pxt.Error, match='resource_pool requires an async function'):
+            t3.insert([{'num': 1}])
+
+
+def init_test_pool(pool_name: str) -> None:
+    import datetime
```
aaron-siegel: Move these to top of file

tests/test_function.py
```diff
@@ -1319,10 +1319,155 @@ def test_udf_in_global_namespace(self, uses_db: None) -> None:
...
+
+    from pixeltable import env
+
+    pool_info = env.Env.get().get_resource_pool_info(pool_name, env.RateLimitsInfo)
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
+    pool_info.record(now, requests=(100, 99, now))
+
+
+# Ruff ignore: async is required for the rate-limits scheduler path we could put some trival
```
aaron-siegel: Please elaborate on the explanation for why it's necessary (this wasn't immediately clear to me, though now I understand)

Also, it might be easier to just put `# ruff: noqa: RUF029` at the top of the file, rather than repeating `noqa` so many times

aaron-siegel (changes requested):

tests/test_function.py
christopherpestano: Added to the comment that it's the FnCall Evaluator that restricts the definition of rate limited udfs to be strictly those that are async and moved ruff ignore to top of file.

tests/test_function.py
christopherpestano: moved

tests/test_function.py
christopherpestano: moved

tests/test_function.py
christopherpestano: changed to use validate_update_status

pixeltable/func/function.py
christopherpestano: Agreed, we can leave this comment here for reference as it is not exposed in docs. We should only be using this internally for things such as gemini embeddings.

PR #1195, opened 2026-03-10
[PXT-1067] Disable sqlalchemy INFO logging by default

sergey-mkhitaryan:
1. Set the `sqlalchemy.engine` level to WARNING during Pixeltable init. It's pretty chatty at INFO, and users don't need to see that by default.
2. `pxt.configure_logging` sets logging level not only for the `pixeltable` logger but also for sqlalchemy
3. Logging-related cleanup and improvements

### DIFF
--- a/pixeltable/env.py
+++ b/pixeltable/env.py
@@ -907,7 +907,7 @@ class Env:
             try:
                 logger.removeHandler(handler)
             except Exception as e:
-                _logger.warning(f'Error closing handler: {e}')
+                _logger.warning(f'Error removing handler: {e}')
 
         handlers_set = {fh for _, fh in self._managed_logging_handlers}
         for handler in handlers_set:


pixeltable/env.py
```diff
@@ -893,14 +903,20 @@ def _clean_up(self) -> None:
...
-        for handler in self._logger.handlers[:]:
+        for logger, handler in self._managed_logging_handlers:
             try:
-                handler.close()
-                self._logger.removeHandler(handler)
+                logger.removeHandler(handler)
             except Exception as e:
                 _logger.warning(f'Error removing handler: {e}')
 
+        for handler in {fh for _, fh in self._managed_logging_handlers}:
```
aaron-siegel: I'd find this clearer if broken up:
```
handlers = {fh for _, fh in self._managed_logging_handlers}:  # set of distinct handlers
for handler in handlers:
    ...
```

aaron-siegel (approved):

pixeltable/env.py
sergey-mkhitaryan: @copilot can you apply this suggestion?

sergey-mkhitaryan: @claude can you apply this suggestion?

PR #1204, opened 2026-03-13
PXT-968: More bounding box udfs

mkornacker:


pixeltable/functions/vision.py
```diff
@@ -809,44 +1046,220 @@ def bboxes_crop_canvas(
     Returns:
         List of adjusted bounding boxes in the same format as the input. They can extend beyond the canvas boundaries.
     """
-    raise NotImplementedError()
+    if len(bboxes) == 0:
+        return []
+
+    is_absolute = _validate_bboxes(bboxes, 'bboxes_crop_canvas()', validate_range=False)
+
```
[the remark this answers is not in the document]
mkornacker: Done.

pixeltable/functions/vision.py
```diff
@@ -428,15 +428,21 @@ def bboxes_draw(
...
     if not all(len(b) == 4 for b in bboxes):
         raise pxt.Error(f'{error_prefix}: each bounding box must have exactly 4 coordinates')
-    is_absolute = all(isinstance(x, int) and x >= 0 for x in itertools.chain.from_iterable(bboxes))
-    is_relative = all(isinstance(x, float) and 0.0 <= x <= 1.0 for x in itertools.chain.from_iterable(bboxes))
+    is_absolute = all(
+        isinstance(x, int) and (not validate_range or x >= 0) for x in itertools.chain.from_iterable(bboxes)
+    )
+    is_relative = all(
+        isinstance(x, float) and (not validate_range or (0.0 <= x <= 1.0))
+        for x in itertools.chain.from_iterable(bboxes)
```
aaron-siegel: `itertools.chain.from_iterable()` works here, but you can also do
```
all(predicate for bbox in bboxes for x in bbox)
```
which might be clearer

pixeltable/functions/vision.py
```diff
@@ -719,7 +725,89 @@ def bboxes_scale(
     Returns:
         List of scaled bounding boxes in the same format as the input.
     """
-    raise NotImplementedError()
+    if len(bboxes) == 0:
+        return []
+
+    # Parameter validation
+    has_factor = factor is not None
```
aaron-siegel, line 729: It'd be a more intuitive API to me to have two signatures, one with
```
scale_factor: float | None
```
and the other with
```
scale_factor: tuple[float, float]
```
This would be more consistent with other libraries such as PIL, which tend to take tuple parameters rather than separate x and y parameters.

pixeltable/functions/vision.py
```diff
@@ -719,7 +725,89 @@ def bboxes_scale(
...
+    arr = np.array(bboxes, dtype=np.float64)
+    assert arr.ndim == 2 and arr.shape[1] == 4
+    c0, c1, c2, c3 = arr[:, 0], arr[:, 1], arr[:, 2], arr[:, 3]
+
+    # Convert to cx, cy, w, h
+    w: np.ndarray
+    h: np.ndarray
+    cx: np.ndarray
+    cy: np.ndarray
+    if format == 'xyxy':
```
aaron-siegel: Preferred style in Python 3.10+ is to use a matcher:
```
match format:
    case 'xyxy':
        ...
    case 'xywh':
        ...
    ...
    case _:
        raise pxt.Error ...
```

pixeltable/functions/vision.py
```diff
@@ -738,20 +826,88 @@ def bboxes_pad(
...
-            coordinates in [0, 1].
+        bboxes: List of bounding boxes in absolute pixel coordinates.
         format: Format of the bounding box coordinates, one of 'xyxy', 'xywh', 'cxcywh'.
-        top: Amount to pad at the top.
-        bottom: Amount to pad at the bottom.
-        left: Amount to pad at the left.
-        right: Amount to pad at the right.
-        x: Amount to pad at the left and right.
-        y: Amount to pad at the top and bottom.
+        top: Amount to pad at the top, in absolute pixels.
```
aaron-siegel, line 829: Likewise, `padding: tuple[int, int, int, int]` would be most logical to me here.

PR #1211, opened 2026-03-19
[PXT-565] Llama CPP tool call invocation integration

sergey-mkhitaryan:
Limitations:
1. Only the specific combinations of model, `chat_format`, and `tool_choice` result in a response that is compatible with the OpenAI's tool call format. The support may be improved for Qwen models, but in order to do that, the UDF needs to attempt to parse the XML model output, and convert it to the expected format.
2. I did not use `run_tool_invocations_test` for testing because I couldn't make work most of the scenarios in it. E.g. multiple (aka parallel) tool calls is one thing that I couldn't get to work with none of the four models I've tried. Also, the random non-tool question did not work -- all four models were too eager to hallucinate a tool call when none was necessary.

### DIFF
--- a/tests/functions/test_llama_cpp.py
+++ b/tests/functions/test_llama_cpp.py
@@ -54,7 +54,7 @@ class TestLlamaCpp:
         assert len(result2['choices'][0]['message']['content']) > 0
 
     @pytest.mark.expensive  # downloads large models
-    @pytest.mark.parametrize('model', ['mistral', 'gemma', 'qwen', 'salesforce'])
+    @pytest.mark.parametrize('model', ['mistral', 'gemma', 'qwen'])
     def test_tool_invocations(self, uses_db: None, model: str) -> None:
         skip_test_if_not_installed('llama_cpp')
         from pixeltable.functions import llama_cpp
@@ -74,12 +74,6 @@ class TestLlamaCpp:
                 # If this limitation is unacceptable, we can implement a special case in the UDF that parses Qwen's
                 # XML output to a dict.
                 tool_choice = tools.choice(tool=weather)
-            case 'salesforce':
-                # file size: 3.7GB
-                repo_id = 'Salesforce/Llama-xLAM-2-8b-fc-r-gguf'
-                repo_filename = '*Q3_K_S.gguf'
-                chat_format = 'chatml-function-calling'
-                tool_choice = tools.choice(auto=True)
             case 'mistral':
                 # file size: 2.5GB
                 repo_id = 'mistralai/Ministral-3-3B-Instruct-2512-GGUF'


pixeltable/functions/llama_cpp.py
```diff
@@ -56,13 +62,27 @@ def create_chat_completion(
...
+        model_kwargs['tools'] = [{'type': 'function', 'function': tool} for tool in tools]
+
+    # Note: parallel_tool_calls doesn't seem to be supported
+    if tool_choice is not None:
+        if tool_choice['auto']:
+            model_kwargs['tool_choice'] = 'auto'
+        elif tool_choice['required']:
+            model_kwargs['tool_choice'] = 'required'
+        else:
+            assert tool_choice['tool'] is not None
```
aaron-siegel, line 77: Are we sure it's correct for this to be an assertion? It's user input, right? Is it guaranteed by the underlying Pydantic models?

Note we'll be able to clean this up once the Json typing logic is merged.

tests/functions/test_llama_cpp.py
```diff
@@ -1,6 +1,19 @@
...
+import pytest
+
 import pixeltable as pxt
 
 from ..utils import rerun, skip_test_if_not_installed, validate_update_status
+from .tool_utils import stock_price, weather
+
+
+@pytest.fixture(autouse=True)
+def cleanup_llama_cpp() -> Generator:
```
aaron-siegel: return type can be `Iterator[None]`

tests/functions/test_llama_cpp.py
```diff
@@ -1,6 +1,19 @@
...
+    yield
+    from pixeltable.functions import llama_cpp
```
aaron-siegel, line 14: I think this may fail on test configurations that don't include llama_cpp as a library.

tests/functions/test_llama_cpp.py
```diff
@@ -40,4 +53,62 @@ def test_create_chat_completions(self, uses_db: None) -> None:
...
+                repo_filename = '*Q8_0.gguf'
+                # The 'chatml-function-calling' chat format results in an extremely verbose prompt from this model
+                chat_format = None
+                # 'auto' tool choice causes Qwen to generate the tool call in its own native XML format, so specify
+                # the tool choice explicitly instead.
+                # If this limitation is unacceptable, we can implement a special case in the UDF that parses Qwen's
+                # XML output to a dict.
+                tool_choice = tools.choice(tool=weather)
+            case 'salesforce':
+                # file size: 4.9GB
```
aaron-siegel: These are massive models (from a pytest point of view) and will impose a substantial tax on CI (additional 14GB I/O for every test configuration that installs llama_cpp). Consider marking with `@pytest.mark.expensive` which limits them to a single CI configuration.

tests/functions/test_llama_cpp.py
```diff
@@ -40,4 +53,62 @@ def test_create_chat_completions(self, uses_db: None) -> None:
...
+                repo_id = 'Salesforce/Llama-xLAM-2-8b-fc-r-gguf'
+                repo_filename = '*Q4_K_M.gguf'
+                chat_format = 'chatml-function-calling'
+                tool_choice = tools.choice(auto=True)
+            case 'mistral':
+                # file size: 4.4GB
+                repo_id = 'NousResearch/Hermes-2-Pro-Mistral-7B-GGUF'
```
aaron-siegel: This is a third-party finetune of an ancient version of Mistral. Consider using an official, more recent model, such as Ministral 3.
https://huggingface.co/mistralai/Ministral-3-3B-Instruct-2512-GGUF
(It has the added advantage of being considerably smaller for testing.)

tests/functions/test_llama_cpp.py
```diff
@@ -40,4 +53,62 @@ def test_create_chat_completions(self, uses_db: None) -> None:
...
+                repo_filename = '*Q4_K_M.gguf'
+                chat_format = 'chatml-function-calling'
+                tool_choice = tools.choice(auto=True)
+            case 'gemma':
+                # file size: 4.1GB
+                repo_id = 'lmstudio-community/gemma-3-4b-it-GGUF'
+                repo_filename = '*Q8_0.gguf'
```
aaron-siegel: No need for such a large quant. Q4_K_M (for example) would put less strain on CI.

tests/functions/test_llama_cpp.py
```diff
@@ -40,4 +53,62 @@ def test_create_chat_completions(self, uses_db: None) -> None:
...
+        from pixeltable.functions import llama_cpp
+        from pixeltable.functions.openai import invoke_tools
+
+        tools = pxt.tools(stock_price, weather)
+
+        match model:
+            case 'qwen':
+                # file size: 639MB
+                repo_id = 'Qwen/Qwen3-0.6B-GGUF'
+                repo_filename = '*Q8_0.gguf'
```
aaron-siegel: No need for such a large quant.

pixeltable/functions/llama_cpp.py
sergey-mkhitaryan: This was just copied from the openai UDF. The pydoc here directs to use an instance of `pxt.func.tools.ToolChoice`. `Tools.choice()` which builds it, does the validation. However in all other scenarios (ToolChoice constructor, or tool_choice as a plain dict) there's no validation until this point. Let me know if you think this needs to be an Error.

tests/functions/test_llama_cpp.py
sergey-mkhitaryan: I don't think so, there's no `import llama_cpp` in this path, or at the module level. Let me know what you have in mind.

tests/functions/test_llama_cpp.py
sergey-mkhitaryan: that's the only option for that model, and it's <1GB https://huggingface.co/Qwen/Qwen3-0.6B-GGUF/tree/main
is it still a problem?

tests/functions/test_llama_cpp.py
sergey-mkhitaryan: Done. Note: updated llama-cpp-python for mistral 3 support.

pixeltable/functions/llama_cpp.py
aaron-siegel: Ok you might be right (my memory is hazy). Let's leave as is and fix (all the tool calling UDF params) with proper Json typing as a follow up.

tests/functions/test_llama_cpp.py
aaron-siegel: Ok, you might be right. We'll know for sure when the minimal test configs run in merge queue.

tests/functions/test_llama_cpp.py
aaron-siegel: You can use this one
`https://huggingface.co/unsloth/Qwen3-0.6B-GGUF`
(not a 3rd party model - it's a 3rd party quant of the original Qwen model.)

1G wouldn't kill us but smaller is better for unit testing.

tests/functions/test_llama_cpp.py
```diff
@@ -40,4 +53,63 @@ def test_create_chat_completions(self, uses_db: None) -> None:
...
+                # The 'chatml-function-calling' chat format results in an extremely verbose prompt from this model
+                chat_format = None
+                # 'auto' tool choice causes Qwen to generate the tool call in its own native XML format, so specify
+                # the tool choice explicitly instead.
+                # If this limitation is unacceptable, we can implement a special case in the UDF that parses Qwen's
+                # XML output to a dict.
+                tool_choice = tools.choice(tool=weather)
+            case 'salesforce':
+                # file size: 4.9GB
```
aaron-siegel: the file sizes are no longer accurate?

aaron-siegel (approved):

pixeltable/functions/llama_cpp.py
```diff
@@ -56,13 +62,27 @@ def create_chat_completion(
     if (repo_id is None) and (repo_filename is not None):
         raise excs.Error('`repo_filename` can only be provided along with `repo_id`.')
 
+    if tools is not None:
+        model_kwargs['tools'] = [{'type': 'function', 'function': tool} for tool in tools]
```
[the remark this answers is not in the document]
sergey-mkhitaryan: 0.3.18 is only required for mistral3 architecture support. tools, choice, chat_format do not necessarily require it. 

Updated the requirement to 0.3.9 when tools are present. The test passes except for mistral model.

tests/functions/test_llama_cpp.py
sergey-mkhitaryan: no, it was accurate. updated to a slightly smaller quant this time. can't go lower.

PR #1212, opened 2026-03-19
[PXT-1065 + PXT-1066] Add export_csv and export_json

christopherpestano:
Add export_csv and export_json functions for exporting Pixeltable tables and queries to CSV and JSON files.

Move CSV import/export to pixeltable/io/csv.py and JSON import/export to pixeltable/io/json.py. Trim datarows.py and pandas.py to only contain their respective concerns.

Add create_all_types_tbl test helper covering all exportable types (String, Int, Float, Bool, Timestamp, Date, UUID, Json, Array, Image, Video, Audio, Document). Tests cover media types, nulls, filtered queries, column subsets, round-trip import, pretty-printing, and custom delimiters.

Also migrated all _exec() export usage to instead use cursor, and refactored functions so they are in appropriate locations.


pixeltable/io/csv.py
```diff
@@ -0,0 +1,102 @@
...
+        writer = csv.writer(f, delimiter=delimiter, quoting=quoting)  # type: ignore[arg-type]
+        writer.writerow(col_types.keys())
+
+        for row in result:
+            csv_row: list[Any] = []
+            for col_name, col_type in col_types.items():
+                val = row[col_name]
+                if val is None:
+                    csv_row.append('')
+                elif col_type.is_image_type():
```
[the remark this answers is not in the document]
christopherpestano: Bad, image is a special case in pixel table

pixeltable/io/pandas.py
```diff
@@ -12,6 +16,11 @@
 import pixeltable.type_system as ts
 from pixeltable.env import Env
 
+if typing.TYPE_CHECKING:
+    import pixeltable as pxt
+
+_logger = logging.getLogger('pixeltable')
```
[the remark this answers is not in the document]
christopherpestano: Fixed.

pixeltable/io/datarows.py
```diff
@@ -1,10 +1,17 @@
...
 
 import pixeltable as pxt
+import pixeltable.exceptions as excs
 import pixeltable.type_system as ts
-from pixeltable import exceptions as excs
+
+if typing.TYPE_CHECKING:
+    import pixeltable as pxt
+
+_logger = logging.getLogger('pixeltable')
```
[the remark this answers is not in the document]
christopherpestano: Fixed.

pixeltable/io/json.py
```diff
@@ -0,0 +1,113 @@
...
+
+    rows: list[dict[str, Any]] = []
+    for row in result:
+        row_dict: dict[str, Any] = {}
+        for col_name, col_type in col_types.items():
+            val = row[col_name]
+
+            if val is None:
+                row_dict[col_name] = None
+            elif col_type.is_image_type():
```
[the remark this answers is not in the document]
christopherpestano: Image is a special case in Pixeltable (PIL.Image objects). Video/Audio/Document are already strings and fall through to the default branch correctly.

pixeltable/io/json.py
```diff
@@ -0,0 +1,113 @@
...
+            elif col_type.is_uuid_type():
+                row_dict[col_name] = str(val)
+            elif col_type.is_array_type():
+                row_dict[col_name] = val.tolist()
+            else:
+                row_dict[col_name] = val
+        rows.append(row_dict)
+
+    with open(file_path, 'w', encoding='utf-8') as f:
+        json.dump(rows, f, indent=indent, default=str, ensure_ascii=False, allow_nan=False)
```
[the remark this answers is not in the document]
christopherpestano: Too much complexity for the use case.

pixeltable/io/csv.py
```diff
@@ -0,0 +1,102 @@
...
+                    csv_row.append(str(val.filename) if hasattr(val, 'filename') and val.filename else '')
+                elif col_type.is_json_type():
+                    csv_row.append(json.dumps(val))
+                elif col_type.is_array_type():
+                    csv_row.append(json.dumps(val.tolist()))
+                else:
+                    csv_row.append(val)
```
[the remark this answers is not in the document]
christopherpestano: Fixed.

pixeltable/io/__init__.py
```diff
@@ -1,16 +1,18 @@
 """Functions for importing and exporting Pixeltable data."""
 # ruff: noqa: F401
 
-from .datarows import import_json, import_rows
+from .csv import export_csv, import_csv
```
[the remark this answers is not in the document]
christopherpestano: Fixed.

tests/io/test_csv.py
```diff
@@ -0,0 +1,75 @@
...
+        t, _ = create_all_types_tbl('test_csv_all')
+        original = t.collect()
+
+        csv_path = tmp_path / 'all_types.csv'
+        pxt.io.export_csv(t, csv_path)
+
+        t2 = pxt.io.import_csv('test_csv_all_reimported', str(csv_path))
+        reimported = t2.collect()
+
+        assert len(reimported) == len(original)
```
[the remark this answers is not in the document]
christopherpestano: Fixed, test now asserts ResultSet equality after round-trip.

pixeltable/io/csv.py
```diff
@@ -0,0 +1,104 @@
...
+        writer.writerow(col_types.keys())
+
+        for row in result:
+            csv_row: list[Any] = []
+            for col_name, col_type in col_types.items():
+                val = row[col_name]
+                if val is None:
+                    csv_row.append('')
+                elif col_type.is_image_type():
+                    csv_row.append(str(val.filename) if hasattr(val, 'filename') and val.filename else '')
```
aaron-siegel: I'm not convinced it actually makes sense to accept media types here.

Another option would be to disallow media types, or anything that can't be canonically represented inside the CSV file, and require the user to select on `expr.localpath` or `expr.fileurl` and be explicit about the string they're converting to.

A particular risk with the present implementation is that `val.filename` (or `val` in the case of non-image media types) could be a path to a temporary (cached) file that later gets expunged. In that scenario, the fact that it's being exported suggests a degree of permanence that isn't necessarily there.

Other comment: the use of hasattr is suspicious here. If it remains, please leave a comment explaining why (i.e., why is there not a typesafe way to get the same result)

pixeltable/io/csv.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Pixeltable column types are mapped to CSV values as follows:
+
+    - String, Int, Float, Bool: native CSV representation
+    - Timestamp, Date: ISO 8601 string representation
+    - UUID: string representation
+    - Json: JSON-encoded string
+    - Array: JSON-encoded string (via ``tolist()``)
+    - Binary: excluded from export (not representable in CSV)
+    - Image, Video, Audio, Document: file path or URL string
```
aaron-siegel, line 63: The way it is currently implemented, I don't believe you'll ever get a URL.

pixeltable/io/csv.py
```diff
@@ -0,0 +1,104 @@
...
+) -> None:
+    """
+    Exports a query result or table to a CSV file.
+
+    Pixeltable column types are mapped to CSV values as follows:
+
+    - String, Int, Float, Bool: native CSV representation
+    - Timestamp, Date: ISO 8601 string representation
+    - UUID: string representation
+    - Json: JSON-encoded string
```
aaron-siegel, line 60: What happens if it contains a non-serializable object (image or numpy array)?

pixeltable/io/csv.py
```diff
@@ -0,0 +1,104 @@
...
+    - Array: JSON-encoded string (via ``tolist()``)
```
aaron-siegel: We conventionally use single backticks for code style in docstrings.

pixeltable/io/json.py
aaron-siegel: What does `allow_nan=False` do? What if there's a float column with nans?

tests/io/test_csv.py
```diff
@@ -0,0 +1,75 @@
+import csv
+import pathlib
+
+import pixeltable as pxt
+
+from ..utils import create_test_tbl, validate_update_status
+
+
+class TestCsv:
+    def test_export_round_trip(self, uses_db: None, tmp_path: pathlib.Path) -> None:
```
aaron-siegel, line 54: I'd prefer to see this supplemented by a direct test that verifies output against an exact (expected) file.

tests/io/test_csv.py
```diff
@@ -0,0 +1,75 @@
...
+            assert len(list(csv.DictReader(f))) == 5
+
+        csv_path2 = tmp_path / 'subset.csv'
+        pxt.io.export_csv(t.select(t.c_string), csv_path2)
+        with open(csv_path2, encoding='utf-8') as f:
+            exported = list(csv.DictReader(f))
+        assert len(exported) == 10
+        assert list(exported[0].keys()) == ['c_string']
+
+    def test_export_custom_delimiter(self, uses_db: None, tmp_path: pathlib.Path) -> None:
```
aaron-siegel: Why does this need to be a separate test, rather than just parameterizing on delimiter (or iterating over delimiters?)

tests/utils.py
```diff
@@ -150,6 +150,61 @@ def create_table_data(t: pxt.Table, col_names: list[str] | None = None, num_rows
     return rows
 
 
+def create_all_types_tbl(name: str, num_rows: int = 5) -> tuple[pxt.Table, list[dict[str, Any]]]:
```
aaron-siegel: This seems redundant with the existing create_all_datatypes_tbl. Is there a reason that one can't be adapted for the new tests?

tests/utils.py
```diff
@@ -150,6 +150,61 @@ def create_table_data(t: pxt.Table, col_names: list[str] | None = None, num_rows
...
+    """Create a table with every exportable Pixeltable type and insert rows with varied data.
+
+    Includes: String, Int, Float, Bool, Timestamp, Date, UUID, Json, Array, Image, Video, Audio, Document.
+    """
+    from zoneinfo import ZoneInfo
```
aaron-siegel: Move to top of file

aaron-siegel (changes requested):

pixeltable/io/json.py
christopherpestano: It gets converted to None. The issue here is that although python supports NaN in json, the official RFC for json (RFC 8259) explicitly says that NaN is not valid json. Meaning that if we were to allow NaN, we would be outputting json that we can only guarantee python would be able to parse. So for better compatibility with languages such as Javascript, this is disabled and NaN values are converted to none.

tests/utils.py
christopherpestano: moved

tests/utils.py
christopherpestano: Fair enough, originally I made the additional one to also have the document field, should just add document to the main table

pixeltable/io/csv.py
christopherpestano: Sure, switched to single backtick

pixeltable/io/json.py
aaron-siegel: Got it. Please leave a code comment explaining this (doesn't have to be a novella - something as simple as "convert NaN to None in accordance with the RFC 8259 spec")

christopherpestano:
Need to switch how we handle media files here, currently we are using files from the temporary ~/.pixeltable directory however there is a remote url attribute that contains the actual authoritative location of the file. We should add a helper function that reaches into the users query for both export_csv and export_json and manually for each media column switches it out with its remote url attribute. Then we should validate in unit tests that we are no longer getting dirs within ~/.pixeltable but rather that we are properly getting both local file paths as well as remote urls.

pixeltable/io/csv.py
christopherpestano: Per offline discussion switched it to use ColumnPropertyRef.Property.FILEURL.

pixeltable/io/json.py
christopherpestano: Added it in addition with the reasoning being we want to be javascript compatible.

tests/io/test_csv.py
christopherpestano: Fair enough, changed it so it iterates over several delimiters

tests/io/test_csv.py
christopherpestano: Sure added test csv file to data and added test that validates round trip completeness that when we export it is exactly like the file we imported from.

pixeltable/io/csv.py
christopherpestano: Image will give you an error if it has no destination (intentionally, so that users know that to export a media column it requires a destination), otherwise it just gives you the url/filepath. For np arrays they get converted to lists.

pixeltable/io/csv.py
christopherpestano: fixed

pixeltable/io/utils.py
```diff
@@ -98,3 +103,19 @@ def normalize_schema_names(
...
+    URL stored in the database for each media column.
+    """
+    for i, expr in enumerate(select_list_exprs):
+        if isinstance(expr, ColumnRef) and expr.col_type.is_media_type():
+            if expr.col.is_computed and expr.col.destination is None:
+                raise excs.Error(
+                    f'Cannot export computed media column {expr.col.name!r} without a destination. '
+                    f'Set a destination on the column or select specific non-media columns instead.'
+                )
+            select_list_exprs[i] = ColumnPropertyRef(expr, ColumnPropertyRef.Property.FILEURL)
```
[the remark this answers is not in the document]
christopherpestano: This is fine, queries are one and done.

pixeltable/io/json.py
```diff
@@ -0,0 +1,117 @@
...
+        query = table_or_query.select()
+    else:
+        query = table_or_query
+
+    replace_media_with_fileurl(query._select_list_exprs)
+
+    col_types: dict[str, ts.ColumnType] = {name: ct for name, ct in query.schema.items() if not ct.is_binary_type()}
+
+    result = query.collect()
+
```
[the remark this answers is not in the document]
christopherpestano: Needless optimization, we should avoid mutating the select list any more than we have to

tests/io/test_csv.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def test_export_exact_output(self, uses_db: None, tmp_path: pathlib.Path) -> None:
+        """Verify exported CSV matches an expected file exactly."""
+        expected_path = get_csv_file('expected_export.csv')
+        t = pxt.io.import_csv('test_csv_exact', expected_path, primary_key='c_int')
+
+        csv_path = tmp_path / 'exact.csv'
+        pxt.io.export_csv(t.order_by(t.c_int), csv_path)
+
+        assert csv_path.read_text(encoding='utf-8') == pathlib.Path(expected_path).read_text(encoding='utf-8')
```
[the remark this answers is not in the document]
christopherpestano: Eh this is fine, if the file moves it's expected these tests start failing

tests/io/test_json.py
```diff
@@ -0,0 +1,190 @@
...
+        pxt.io.export_json(t, pretty_path, indent=2)
+        pretty = pretty_path.read_text(encoding='utf-8')
+        assert pretty.count('\n') > 1
+        assert 'Manwë' in pretty
+
+    def test_round_trip(self, uses_db: None, tmp_path: pathlib.Path) -> None:
+        """Export JSON, re-import, and verify data matches."""
+        source_path = get_json_file('example.json')
+        t = pxt.io.import_json('test_json_rt', source_path)
+
```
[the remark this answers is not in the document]
christopherpestano: bad

pixeltable/io/csv.py
```diff
@@ -0,0 +1,106 @@
...
+    if isinstance(table_or_query, pxt.catalog.Table):
+        query = table_or_query.select()
+    else:
+        query = table_or_query
+
+    replace_media_with_fileurl(query._select_list_exprs)
+
+    col_types: dict[str, ts.ColumnType] = {name: ct for name, ct in query.schema.items() if not ct.is_binary_type()}
+
+    result = query.collect()
```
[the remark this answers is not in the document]
christopherpestano: Needless optimization, we would need to reach into the select statement and delete the key for the binary column, ends up being too complicated for the perf gain.

tests/utils.py
```diff
@@ -444,6 +454,30 @@ def get_documents() -> list[str]:
...
+def get_json_files() -> list[str]:
+    json_dir = TESTS_DIR / 'data' / 'json'
+    return glob.glob(f'{json_dir}/*', recursive=True)
+
+
+def get_json_file(name: str) -> str | None:
+    json_dir = TESTS_DIR / 'data' / 'json'
+    file_path = json_dir / name
+    glob_result = glob.glob(f'{file_path}', recursive=True)
+    return glob_result.pop(0) if len(glob_result) > 0 else None
```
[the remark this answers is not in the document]
christopherpestano: bad

tests/io/test_json.py
```diff
@@ -0,0 +1,190 @@
...
+        with open(json_path2, encoding='utf-8') as f:
+            exported = json.load(f)
+        assert len(exported) == 10
+        assert list(exported[0].keys()) == ['c_string']
+
+    def test_export_indent_and_encoding(self, uses_db: None, tmp_path: pathlib.Path) -> None:
+        """Test compact vs pretty output and non-ASCII preservation."""
+        source_path = get_json_file('example.json')
+        t = pxt.io.import_json('test_json_fmt', source_path)
+
```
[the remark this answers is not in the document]
christopherpestano: bad

pixeltable/io/json.py
```diff
@@ -0,0 +1,117 @@
...
+    schema_overrides: dict[str, Any] | None = None,
+    primary_key: str | list[str] | None = None,
+    num_retained_versions: int = 10,
+    comment: str = '',
+    **kwargs: Any,
+) -> pxt.Table:
+    """
+    Creates a new base table from a JSON file. This is a convenience method and is
+    equivalent to calling `import_data(table_path, json.loads(file_contents, **kwargs), ...)`, where `file_contents`
+    is the contents of the specified `filepath_or_url`.
```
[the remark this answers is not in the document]
christopherpestano: This wasn't introduced by me but yeah that should be fixed

pixeltable/io/csv.py
```diff
@@ -0,0 +1,106 @@
...
+
+    file_path = Path(file_path)
+    file_path.parent.mkdir(parents=True, exist_ok=True)
+
+    with open(file_path, 'w', newline='', encoding='utf-8') as f:
+        writer = csv.writer(f, delimiter=delimiter, quoting=quoting)  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
christopherpestano: ignore

pixeltable/io/csv.py
```diff
@@ -0,0 +1,106 @@
...
+    - Binary: excluded from export (not representable in CSV)
+    - Image, Video, Audio, Document: file path or URL string
+
+    Args:
+        table_or_query: Table or Query to export.
+        file_path: Path to the output CSV file.
+        delimiter: Field delimiter character. Default `','`.
+        quoting: CSV quoting style (a `csv.QUOTE_*` constant). Default `csv.QUOTE_MINIMAL`.
+    """
+    from pixeltable.io.utils import replace_media_with_fileurl
```
aaron-siegel: Can this go at top of file?

pixeltable/io/csv.py
```diff
@@ -0,0 +1,106 @@
...
+    Exports a query result or table to a CSV file.
+
+    Pixeltable column types are mapped to CSV values as follows:
+
+    - String, Int, Float, Bool: native CSV representation
+    - Timestamp, Date: ISO 8601 string representation
+    - UUID: string representation
+    - Json: JSON-encoded string
+    - Array: JSON-encoded string (via `tolist()`)
+    - Binary: excluded from export (not representable in CSV)
```
aaron-siegel, line 62: We could in theory replace a binary by its repr. (Just a thought)

pixeltable/io/csv.py
```diff
@@ -0,0 +1,106 @@
...
+    - Image, Video, Audio, Document: file path or URL string
+
+    Args:
+        table_or_query: Table or Query to export.
+        file_path: Path to the output CSV file.
+        delimiter: Field delimiter character. Default `','`.
+        quoting: CSV quoting style (a `csv.QUOTE_*` constant). Default `csv.QUOTE_MINIMAL`.
+    """
+    from pixeltable.io.utils import replace_media_with_fileurl
+
```
aaron-siegel, line 71: Put `query: Query` here (it's good practice to use an explicit type decl if first assignment can happen via branching code paths)

pixeltable/io/csv.py
```diff
@@ -0,0 +1,106 @@
...
+        for row in result:
+            csv_row: list[Any] = []
+            for col_name, col_type in col_types.items():
+                val = row[col_name]
+                if val is None:
+                    csv_row.append('')
+                elif col_type.is_timestamp_type() or col_type.is_date_type():
+                    csv_row.append(val.isoformat())
+                elif col_type.is_json_type():
+                    csv_row.append(json.dumps(val))
```
aaron-siegel, line 98: I think this may fail if `val` is a dict or list containing an image, array, or binary.

pixeltable/io/csv.py
```diff
@@ -0,0 +1,106 @@
...
+                elif col_type.is_array_type():
```
aaron-siegel: Probably safer to look at the type of val rather than the col_type. Note that (for example) an ndarray can appear inside a column of type `JsonType`.

I'd replace all of these by appropriate analogs `isinstance(val, np.ndarray)`, `isinstance(val, datetime)`, etc.

pixeltable/io/json.py
```diff
@@ -0,0 +1,118 @@
...
+    file_path = Path(file_path)
+    file_path.parent.mkdir(parents=True, exist_ok=True)
+
+    rows: list[dict[str, Any]] = []
+    for row in result:
+        row_dict: dict[str, Any] = {}
+        for col_name, col_type in col_types.items():
+            val = row[col_name]
+
+            if val is None:
```
aaron-siegel: Code duplication with export_csv, and they're not identical (uuid is present here, but not in export_csv - is that intentional?)

Best to factor out the shared logic to reduce chance of bugs creeping in.

Same comments apply here as to export_csv.

pixeltable/io/json.py
```diff
@@ -0,0 +1,118 @@
...
+            elif col_type.is_array_type():
+                row_dict[col_name] = val.tolist()
+            else:
+                row_dict[col_name] = val
+        rows.append(row_dict)
+
+    with open(file_path, 'w', encoding='utf-8') as f:
+        # Python's json module allows NaN by default, which is non-compliant with RFC 8259
+        # and would produce output that JavaScript's JSON.parse() rejects.
+        json.dump(rows, f, indent=indent, default=str, ensure_ascii=False, allow_nan=False)
```
aaron-siegel: Are we sure this is what we want? It will emit a single object for all rows, rather than one object per row.

I'd think a `.jsonl` file would be preferred, with one object per row.

cc @mkornacker @pierrebrunelle

pixeltable/io/utils.py
```diff
@@ -98,3 +103,19 @@ def normalize_schema_names(
...
+
+def replace_media_with_fileurl(select_list_exprs: list[Expr]) -> None:
+    """Mutate select list in-place, replacing media ColumnRefs with their .fileurl property.
+
+    This avoids Pixeltable's file caching pipeline and instead returns the authoritative
+    URL stored in the database for each media column.
+    """
+    for i, expr in enumerate(select_list_exprs):
+        if isinstance(expr, ColumnRef) and expr.col_type.is_media_type():
+            if expr.col.is_computed and expr.col.destination is None:
```
aaron-siegel: Didn't we say we want to emit `file://` URLs for outputs that are on the local filesystem, as long as they're in a permanent location (such as in the media store)? Even so, checking `col.destination is None` won't really disambiguate because the destination could be a directory on the local filesystem anyway.

If we only want to support exports for which the url is a remote location, the preferred way to do that is to look at the URI scheme of the actual URLs. But I would advocate for supporting local file exports too.

pixeltable/io/utils.py
aaron-siegel: Looking at the code above, this does an in-place modification of an internal variable of a `Query` object. I'm not sure this is wise - in addition to violating the class contract (it's a protected member of `Query`) it could lead to inconsistent state in the Query (`select_list` won't match `_select_list_exprs`, for example) which could be a particular problem if it's a user-supplied Query that the user then tries to reuse.

I see Copilot had the identical comment, which you dismissed, but I agree with Copilot here. It'd be safest to construct a new Query. Or simply skip this step and require the user to explicitly specify fileurl in their query (in fact I recall us agreeing on that approach in an earlier meeting - did that consensus change at some point?)

tests/io/test_csv.py
```diff
@@ -0,0 +1,88 @@
...
+import pathlib
+
+import pytest
+
+import pixeltable as pxt
+
+from ..utils import create_test_tbl, get_csv_file, validate_update_status
+
+
+class TestCsv:
```
aaron-siegel, line 15: Make sure we are testing:
- ALL table datatypes (including binary, video, uuid, etc)
- Json fields containing primitive objects such as int/str/np.ndarray/Image
- Json fields with lists/dicts containing images, arrays, and binaries
- All export parameters (including `quoting`)

and that the tests validate we're doing something intelligent in all cases. (i.e., either validating the documented behavior on a successful export, or emitting an intelligible error if it's data that can't be exported for some reason)

tests/io/test_json.py
```diff
@@ -0,0 +1,190 @@
...
+    create_all_datatypes_tbl,
+    get_image_files,
+    get_json_file,
+    skip_test_if_not_installed,
+    validate_update_status,
+)
+
+
+class TestJson:
+    def test_export_all_types(self, uses_db: None, tmp_path: pathlib.Path) -> None:
```
aaron-siegel, line 14: Great - we need this for TestCsv too.

tests/io/test_json.py
```diff
@@ -0,0 +1,190 @@
...
+        """Export a table with every supported type and verify the JSON output."""
+        t = create_all_datatypes_tbl()
```
aaron-siegel, line 16: create_all_datatypes_tbl() is not comprehensive - in particular, it doesn't capture the full breadth of the json type column. Maybe this is a good opportunity to enhance it to ensure all valid object types are represented in its json type column, as well as some cases of lists/dicts with image/array/binary inside.

aaron-siegel (changes requested):

pixeltable/io/csv.py
christopherpestano: Added

pixeltable/io/csv.py
christopherpestano: Sure

pixeltable/io/csv.py
christopherpestano: I feel like exporting binaries is a bad ideaa, maybe I could see it for inlining images, but in this case who is inlining an image into csv?

pixeltable/io/csv.py
christopherpestano: In convert rows we now check that the json is serializable and raise an error if we are not able to serialize the json properly.

pixeltable/io/utils.py
christopherpestano: I had thought that the thing we were going with was that if the user explicitly placed the files in some location like via using destination, then that meant that the files were good for export, otherwise they are dependent on the original table and internal pixeltable media storage. This does emit `file://` urls, so long as it is an explicitly declared location, which can be seen in the output json when running export all datatypes for test json.

tests/utils.py
```diff
@@ -82,38 +82,57 @@ def make_tbl(name: str = 'test', col_names: list[str] | None = None) -> pxt.Tabl
     return pxt.create_table(name, schema)  # type: ignore[arg-type]
 
 
-def create_table_data(t: pxt.Table, col_names: list[str] | None = None, num_rows: int = 10) -> list[dict[str, Any]]:
+def create_table_data(
+    t: pxt.Table, col_names: list[str] | None = None, num_rows: int = 10, non_serializable_json: bool = True
```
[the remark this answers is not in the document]
christopherpestano: Fair enough

tests/io/test_json.py
```diff
@@ -0,0 +1,208 @@
...
+        t = pxt.create_table('test_json_encoding', {'name': pxt.String})
+        t.insert([{'name': 'Manwe'}, {'name': 'Feanor'}])
+
+        json_path = tmp_path / 'encoding.jsonl'
+        pxt.io.export_json(t, json_path)
+
+        with open(json_path, encoding='utf-8') as f:
+            exported = [json.loads(line) for line in f]
+        names = [row['name'] for row in exported]
+        assert 'Manwe' in names
```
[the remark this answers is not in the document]
christopherpestano: fixed

pixeltable/io/json.py
```diff
@@ -0,0 +1,97 @@
...
+
+    query: pxt.Query
+    if isinstance(table_or_query, pxt.catalog.Table):
+        query = table_or_query.select()
+    else:
+        query = table_or_query
+
+    query = query._with_new_select_list(replace_media_with_fileurl(query._select_list_exprs))
+
+    col_types: dict[str, ts.ColumnType] = {name: ct for name, ct in query.schema.items() if not ct.is_binary_type()}
```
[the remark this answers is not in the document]
christopherpestano: This is fine since we don't materialize all rows into memory at once,

pixeltable/io/csv.py
```diff
@@ -0,0 +1,102 @@
...
+if typing.TYPE_CHECKING:
+    import pixeltable as pxt
+
+
+def import_csv(
+    tbl_name: str,
+    filepath_or_buffer: str | os.PathLike,
+    schema_overrides: dict[str, Any] | None = None,
+    primary_key: str | list[str] | None = None,
+    num_retained_versions: int = 10,
```
aaron-siegel: This can be removed now, correct?

pixeltable/io/csv.py
```diff
@@ -0,0 +1,102 @@
...
+    comment: str = '',
```
aaron-siegel: Did we decide to make this `comment: str | None = None`?

pixeltable/io/json.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def import_json(
+    tbl_path: str,
+    filepath_or_url: str,
+    *,
+    schema_overrides: dict[str, Any] | None = None,
+    primary_key: str | list[str] | None = None,
+    num_retained_versions: int = 10,
+    comment: str = '',
```
aaron-siegel: same comments as above

aaron-siegel (approved):

pixeltable/io/csv.py
christopherpestano: My bad, hadn't merged latest main

pixeltable/io/csv.py
christopherpestano: Fixed

pixeltable/io/json.py
christopherpestano: fixed

pixeltable/_query.py
```diff
@@ -689,6 +689,25 @@ def bind(self, args: dict[str, Any]) -> Query:
             offset=offset_val,
         )
 
+    def _with_new_select_list(self, new_exprs: list[exprs.Expr]) -> Query:
```
mkornacker: _replace_select_list() is less awkward.

pixeltable/io/sql.py
```diff
@@ -137,12 +136,10 @@ def export_sql(
...
         batch: list[dict] = []
-        with get_runtime().catalog.begin_xact(for_write=False), engine.connect() as target_conn:
-            for data_row in query._exec():
-                row_dict: dict[str, Any] = {}
-                for col_name, e in zip(query.schema.keys(), query._select_list_exprs):
-                    row_dict[col_name] = data_row[e.slot_idx]
-                batch.append(row_dict)
+        with engine.connect() as target_conn:
+            for data_row in query.cursor():
+                # this works equivalent to the previous code; we already preclude images
```
mkornacker: "equivalent to the previous code" - the code you replaced?

pixeltable/io/utils.py
```diff
@@ -107,3 +120,82 @@ def normalize_schema_names(
...
+    try:
+        with os.fdopen(fd, **open_kwargs) as f:
+            yield f
+        os.replace(tmp_path, file_path)
+    except BaseException:
+        tmp_path.unlink(missing_ok=True)
+        raise
+
+
+def convert_rows(cursor: Iterable[Any], col_types: dict[str, ts.ColumnType]) -> Iterator[dict[str, Any]]:
```
mkornacker: Shouldn't this live in Row.to_json()? Looks like it would be generally useful.

pixeltable/io/utils.py
```diff
@@ -107,3 +120,82 @@ def normalize_schema_names(
...
+    """Convert raw cursor rows into export-ready dicts.
```
mkornacker: More specifically, they are json-serializable dicts.

mkornacker (approved):

pixeltable/_query.py
christopherpestano: Changed naming and references to _replace_select_list()

tests/io/test_json.py
christopherpestano: done

tests/io/test_json.py
christopherpestano: added

pixeltable/io/sql.py
christopherpestano: In this case it was more a hint to reviewers that in isolation this code looks unsafe because this section doesn't handle media types, but with more context on the export_sql it is actually safe. Changed the comment to just be a simple `# we already preclude images so there isn't a need to convert media types here`

pixeltable/io/utils.py
christopherpestano: Changed comment

pixeltable/io/utils.py
christopherpestano: Makes sense, this does mean however that Row needs a reference to _schema now otherwise normal users wouldn't be able to use the function (since ts.ColumnType is private). This is fine since the cost of having an extra reference per row pointing at the original dict is low and the schema is read only anyways. Made the change

PR #1213, opened 2026-03-20
PXT-1054 Allow R2 Home Bucket as Destination Blob Storage

amithadke:
adds pxt store that manages org:db mapping to client
SOA handles new pxt://org:db/prefix mapping (also internally resolving to physical bucket name)
removed client lock from s3store not needed with runtime providing client per thread


pixeltable/utils/cloud_utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def _api_headers() -> dict[str, str]:
+    headers = {'Content-Type': 'application/json'}
+    api_key = Env.get().pxt_api_key
+    if api_key is None:
+        raise excs.Error(
+            'A Pixeltable API key is required for home bucket access. '
+            'Set it with `os.environ["PIXELTABLE_API_KEY"] = "your-key"`, '
+            f'or add `api_key = "your-key"` to the `[pixeltable]` section in {Config.get().config_file}.\n'
```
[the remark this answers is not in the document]
amithadke: ignoring - following existing pattern in publish.sh

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,227 @@
...
+class PxtStore(S3Store):
+    """Home bucket store via per-thread auto-refreshing credentials."""
+
+    _pxt_store_entry: _PxtStoreCacheEntry
+
+    def __init__(self, soa: StorageObjectAddress) -> None:
+        if soa.storage_target != StorageTarget.PIXELTABLE_STORE:
+            raise AssertionError(f'PxtStore requires PIXELTABLE_STORE, got {soa.storage_target}')
+        org, db = soa.account, soa.account_extension
+        self._pxt_store_entry = _get_or_create_pxt_store_entry(org, db)
```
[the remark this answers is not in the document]
amithadke: not true anymore we added check that uri netlocs must have org and db (exactly 2 parts)

pixeltable/share/protocol/home_bucket.py
```diff
@@ -0,0 +1,42 @@
...
+from pydantic import BaseModel
+
+from .common import PxtUri, RequestBaseModel
+from .operation_types import PixeltableStoreOperationType
+
+
+class GetHomeBucketCredentialsRequest(RequestBaseModel):
+    """Request for temporary credentials to access a home bucket."""
+
+    operation_type: Literal[PixeltableStoreOperationType.GET_HOME_BUCKET_CREDENTIALS] = (
```
aaron-siegel: why is the explicit type hint needed?

pixeltable/share/protocol/operation_types.py
aaron-siegel: I'm confused. How can this ever be false?

pixeltable/share/protocol/operation_types.py
```diff
@@ -24,10 +23,21 @@ def is_replica_operation(self) -> bool:
...
+
+
 REPLICA_OPERATIONS: set[ReplicaOperationType] = {
     ReplicaOperationType.PUBLISH_REPLICA,
     ReplicaOperationType.FINALIZE_REPLICA,
     ReplicaOperationType.CLONE_REPLICA,
     ReplicaOperationType.DELETE_REPLICA,
 }
+
+PIXELTABLE_STORE_OPERATIONS: set[PixeltableStoreOperationType] = {
```
aaron-siegel: This is not necesary, just do

PIXELTABLE_STORE_OPERATIONS = set(PixeltableStoreOperationType)

or better yet, frozenset.

pixeltable/utils/cloud_utils.py
```diff
@@ -0,0 +1,82 @@
...
+        GetHomeBucketCredentialsResponse with temporary credentials
+    """
+    request = GetHomeBucketCredentialsRequest(org_slug=org, db_slug=db, prefix=prefix)
+    try:
+        response = requests.post(PIXELTABLE_API_URL, data=request.model_dump_json(), headers=_api_headers(), timeout=15)
+        if response.status_code != 200:
+            raise excs.Error(f'Failed to get home bucket credentials: {response.text}')
+        body = response.json()
+        if isinstance(body, dict) and 'body' in body:
+            import json
```
aaron-siegel: Move to top of file

pixeltable/utils/cloud_utils.py
```diff
@@ -0,0 +1,82 @@
...
+) -> str:
+    """
+    Request a presigned URL from Pixeltable Cloud for a key in the org/db home bucket.
+    Uses backend credentials on the cloud so URL expiry is independent of temp credential TTL.
+    """
+    request = GetPresignedUrlRequest(org_slug=org_slug, db_slug=db_slug, key=key, method=method, expiration=expiration)
+    try:
+        response = requests.post(PIXELTABLE_API_URL, data=request.model_dump_json(), headers=_api_headers(), timeout=30)
+        if response.status_code != 200:
+            raise excs.Error(f'Failed to get presigned URL from Pixeltable Cloud: {response.text}')
```
aaron-siegel: "Cloud" is capitalized here but lowercase in the preceding message; which is it? "Pixeltable cloud" or "Pixeltable Cloud"? Let's pick one and stick to it

pixeltable/utils/object_stores.py
```diff
@@ -249,6 +260,24 @@ def parse_object_storage_addr1(cls, src_addr: str) -> StorageObjectAddress:
             else:
                 account_extension = parsed.netloc
             key = key.lstrip('/')
+        elif scheme == 'pxt':
+            # pxt://org:db/home[/optional/prefix]
```
aaron-siegel: Why "home"? Is there a product doc/discussion that talks about this?

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def _make_credential_refresher(org: str, db: str, entry: _PxtStoreCacheEntry) -> Callable[[], dict[str, str]]:
+    """Return a credential refresher for use with RefreshableCredentials.
+
+    Updates the cache entry's no_space_left and bucket_name on each refresh.
+    """
+
+    def _refresh() -> dict[str, str]:
+        from pixeltable.utils.cloud_utils import get_home_bucket_credentials
```
aaron-siegel: Move to top of file

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,231 @@
...
+            'Only read and delete operations are allowed.',
+            category=excs.PixeltableWarning,
+            stacklevel=3,
+        )
+        entry.no_space_warned = True
+    else:
+        entry.no_space_warned = False
+
+
+def _make_credential_refresher(org: str, db: str, entry: _PxtStoreCacheEntry) -> Callable[[], dict[str, str]]:
```
aaron-siegel: If I understand things right, since there's no local context used in the body of _refresh(), there's no particular reason to use a nested def. I'd find it easier to follow to simply have a single method

```pythong
def _refresh_credentials(org: str, db: str, entry: _PxtStoreCacheEntry)
```

and then use a lambda when we need to pass it to RefreshableCredentials:

```python
refresh_using=lambda: _refresh_credentials(org, db, entry)
```

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,231 @@
...
+    # pxt_store_client is thread-local (Runtime is thread-local), so no lock is needed here.
+    pxt_store_client_dict = get_runtime().get_client('pxt_store')
+    entry = pxt_store_client_dict.clients.get(cache_key)
+    if entry is None:
+        entry = _build_pxt_store_entry(org, db)
+        pxt_store_client_dict.clients[cache_key] = entry
+    return entry
+
+
+class PxtStore(S3Store):
```
aaron-siegel: I'd like to understand the architecture decisions better here. Perhaps naively, I wasn't expecting a separate Store implementation - I was expecting that a `pxt://` URI would simply be resolved to an `s3://` URI, after which S3Store would be used.

I guess the complication is that there's credentials indirection? I think a half-page design note is warranted (if one doesn't already exist), and I'd welcome a verbal explanation also.

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,231 @@
...
+        org, db = soa.account, soa.account_extension
+        self._pxt_store_entry = _get_or_create_pxt_store_entry(org, db)
+        super().__init__(soa, resolved_physical_bucket_name=self._pxt_store_entry.bucket_name)
+
+    @property
+    def bucket_name(self) -> str:
+        """Physical bucket name (updated when credentials refresh)."""
+        return self._pxt_store_entry.bucket_name
+
+    def client(self) -> BaseClient:
```
aaron-siegel: Use a property

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,231 @@
...
+        assert self._pxt_store_entry.client is not None
+        return self._pxt_store_entry.client
+
+    def get_resource(self) -> ServiceResource:
```
aaron-siegel: Use a property

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,231 @@
...
+    def copy_local_file(self, col: 'Column', src_path: Path) -> str:
+        if self._pxt_store_entry.no_space_left:
+            raise excs.Error('No space left in Pixeltable store. Only read and delete operations are allowed.')
+        return super().copy_local_file(col, src_path)
+
+    def create_presigned_url(self, soa: StorageObjectAddress, expiration_seconds: int) -> str:
+        """Request a presigned GET URL from the control plane (lifetime independent of temp credentials)."""
+        if not soa.has_object:
+            raise excs.Error(f'StorageObjectAddress does not contain an object name: {soa}')
+        from pixeltable.utils.cloud_utils import get_presigned_url_from_cloud
```
aaron-siegel: Move to top of file

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,231 @@
...
+            'token': creds.session_token,
+            'expiry_time': expiry_time.isoformat(),
+        }
+
+    return _refresh
+
+
+def _build_pxt_store_entry(org: str, db: str) -> _PxtStoreCacheEntry:
+    """Fetch credentials and build a boto3 session for the home bucket."""
+    from pixeltable.utils.cloud_utils import get_home_bucket_credentials
```
aaron-siegel: Move to top of file

pixeltable/utils/s3_store.py
```diff
@@ -129,44 +128,43 @@ def _get_s3_client_with_region(self) -> Any:
         """
         cd = get_runtime().get_client('s3')
         default_key = 'default'
-        with client_lock:
```
aaron-siegel, line 132: Why get rid of the lock?

pixeltable/utils/s3_store.py
```diff
@@ -98,28 +93,32 @@ class S3Store(ObjectStoreBase):
 
     soa: StorageObjectAddress
 
-    def __init__(self, soa: StorageObjectAddress):
+    def __init__(self, soa: StorageObjectAddress, *, resolved_physical_bucket_name: str | None = None) -> None:
```
aaron-siegel: Add docstring explaining the meaning and purpose of `resolved_physical_bucket_name`. Also, why is it needed - why not simply swap in the soa of the resolved physical bucket? Do we need to preserve the original soa in this case?

tests/test_destination.py
```diff
@@ -152,6 +155,42 @@ def test_dest_parser(self, uses_db: None) -> None:
...
+        assert soa.container == 'home'
+        assert soa.has_object
+        assert soa.object_name == o_name
+        assert soa.prefix == f'{p_name2}/'
+        assert soa.account == 'org'
+        assert soa.account_extension == 'db'
+
+        # Negative cases for pxt:// uris
+        with pytest.raises(ValueError, match='Invalid pxt:// store URI'):
+            ObjectPath.parse_object_storage_addr('pxt://orgonly/home', allow_obj_name=False)
```
aaron-siegel: Why isn't this allowed? Shouldn't it default to `main` the same way table URIs do?

tests/test_pxt_store.py
```diff
@@ -0,0 +1,117 @@
...
+
+class TestPxtStore:
+    """Tests for Pixeltable-managed storage (pxt:// home buckets)."""
+
+    @staticmethod
+    def _resolve_pxt_destination() -> str:
+        skip_test_if_not_installed('boto3')
+        skip_test_if_no_pxt_credentials()
+        try:
+            ObjectOps.validate_destination(PXT_DEST_URI)
```
aaron-siegel: Remove (whitebox call; if it's a false destination then the test should fail via user code)

tests/test_pxt_store.py
```diff
@@ -0,0 +1,117 @@
...
+            return PXT_DEST_URI
+        except Exception as exc:
+            pytest.skip(f'Pixeltable store not reachable: {exc}')
```
aaron-siegel: It's dangerous to put stealthy pytest.skip() calls into tests that could mask a real error. What is the testing scenario where we would want to skip this, rather than fail?

tests/test_pxt_store.py
```diff
@@ -0,0 +1,117 @@
...
+
+        validate_update_status(
+            t.insert([{'img': 'tests/data/imagenette2-160/ILSVRC2012_val_00000557.JPEG'}]), expected_rows=1
+        )
+        assert ObjectOps.count(t._id, dest=dest1_uri) == 1
+
+        # Simulate quota exhaustion by injecting an entry with no_space_left=True.
+        # We patch _get_or_create_pxt_store_entry so every thread (including insert's
+        # worker pool) gets the controlled entry regardless of its local cache state.
+        from pixeltable.utils import pxt_store as pxt_store_mod
```
aaron-siegel: Move to top of file. Also, why the rename?

aaron-siegel (changes requested):

pixeltable/share/protocol/operation_types.py
amithadke: deleting it not used likely old version where we had all op types under one enum

tests/test_destination.py
amithadke: im thinking since this is a storage destination we ask specifically which db's bucket should this point to than letting it be just org and <default db>, we may land objects in different places if default db changes

tests/test_pxt_store.py
amithadke: need to skip tests until cloud side code is in, probably can just mark whole test class to skip than checking for home bucket credentials apis (with valid api key)

pixeltable/share/protocol/home_bucket.py
amithadke: this one is used as discriminator on serving side to match with request model, pydantic needs it as literal.
We also have multiple enums under one union PxtOperationType(cloud side apis + list/get table apis), ReplicaOperationType and PixeltableStoreOperationType. We have to use literal to combine them together for pydantic to match request

pixeltable/utils/object_stores.py
amithadke: thats how we are presenting bucket path on ui today pxt://org:db/bucketname/.. 
 (home for default r2 bucket)
 
 /home also helps to know that uri is for storage not table like pxt:/org:db/A/B/C is pxt dir A/B/C where 
 pxt:/org:db/home/A/B/C is object on home bucket.
 
 another option could be we drop home and use pxtfs://org:db/

pixeltable/utils/pxt_store.py
amithadke: I thought about doing all in s3 store there are few specific cases, caching org:db to client, presigned url and refreshing credentials that i felt should be in pxtstore. Pxtstore is different during init, copy_local_file, create_presigned_url and client, resoruce properties

pixeltable/utils/s3_store.py
amithadke: after runtime changes each thread gets its own s3 store that is not shared with other threads. so client_lock isn't needed

pixeltable/utils/pxt_store.py
amithadke: s3store has client and resource are methods, i dont think we can have property in pxtstore class

pixeltable/utils/s3_store.py
amithadke: i made the change but needed to change _base_uri for SOA to not use physical bucket otherwise fileurls get pxt://org:db/<physical_bucket> which will get stored and they will be invalid as we expect /home for these urls as of today

pixeltable/utils/object_stores.py
aaron-siegel: I kind of like `pxtfs://`

`home` doesn't fully disambiguate, because there's nothing stopping a user from having a directory named `'home'`.

But it's really a product decision, maybe ask about it on `#product`?

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+        entry.no_space_warned = True
+    else:
+        entry.no_space_warned = False
+
+
+def _refresh_credentials(org: str, db: str, entry: _PxtStoreCacheEntry) -> dict[str, str]:
+    """Fetch fresh credentials and update the cache entry"""
+    creds = get_home_bucket_credentials(org, db)
+    expiry_time = datetime.now(tz=timezone.utc) + timedelta(seconds=creds.ttl_seconds)
```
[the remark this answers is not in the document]
amithadke: good catch

pixeltable/share/protocol/home_bucket.py
```diff
@@ -0,0 +1,42 @@
...
+        PixeltableStoreOperationType.GET_HOME_BUCKET_CREDENTIALS
+    )
+    org_slug: str
+    db_slug: str
+    prefix: Optional[str] = None  # Optional path prefix within the home bucket
+
+    def get_pxt_uri(self) -> PxtUri:
+        return PxtUri(f'pxt://{self.org_slug}:{self.db_slug}')
```
aaron-siegel: Is this really where this belongs? It feels like the logic for composing org+db into a pxt URI should live in some more central place.

pixeltable/share/protocol/operation_types.py
```diff
@@ -13,21 +13,18 @@
...
-    ReplicaOperationType.PUBLISH_REPLICA,
-    ReplicaOperationType.FINALIZE_REPLICA,
-    ReplicaOperationType.CLONE_REPLICA,
-    ReplicaOperationType.DELETE_REPLICA,
-}
+    GET_HOME_BUCKET_CREDENTIALS = 'get_home_bucket_credentials'
+    GET_PRESIGNED_URL = 'get_presigned_url'
+
+
+REPLICA_OPERATIONS: frozenset[ReplicaOperationType] = frozenset(ReplicaOperationType)
```
aaron-siegel: Why are all these explicit type hints necessary? I'd expect mypy to be able to infer them. We generally strive to avoid extraneous type hints (if mypy can infer it and it's obvious to a human what the type is, why do we need the type hint?)

pixeltable/utils/object_stores.py
```diff
@@ -249,6 +260,26 @@ def parse_object_storage_addr1(cls, src_addr: str) -> StorageObjectAddress:
...
```
aaron-siegel: Change to pxtfs?

pixeltable/utils/object_stores.py
```diff
@@ -249,6 +260,26 @@ def parse_object_storage_addr1(cls, src_addr: str) -> StorageObjectAddress:
...
+            storage_target = StorageTarget.PIXELTABLE_STORE
+            netloc_parts = parsed.netloc.split(':')
+            if len(netloc_parts) != 2 or not netloc_parts[0] or not netloc_parts[1]:
+                raise ValueError(
+                    f"Invalid pxt:// store URI '{parsed.geturl()}': netloc must be 'org:db', got '{src_addr}'"
```
aaron-siegel: Change to pxtfs? Also below

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,223 @@
...
+
+    _logger.info('Initialized home bucket session for %s:%s', org, db)
+    return entry
+
+
+def _get_or_create_pxt_store_entry(org: str, db: str, prefix: str) -> _PxtStoreCacheEntry:
+    """Return the home bucket entry for org:db:prefix, building it on first use per thread."""
+    cache_key = f'{org}:{db}:{prefix}'
+    # pxt_store_client is thread-local (Runtime is thread-local), so no lock is needed here.
+    pxt_store_client_dict = get_runtime().get_client('pxt_store')
```
aaron-siegel, line 174: Here we do need a type hint for `pxt_store_client_dict`.

pixeltable/utils/pxt_store.py
aaron-siegel: Or we could fix S3Store now.

aaron-siegel (approved):

pixeltable/share/protocol/home_bucket.py
amithadke: this is used only to get org+db from request payload (used for auth purposes) 
RequestBaseModel
class RequestBaseModel(BaseModel, ABC):
    """Abstract base model for protocol requests that must have a PxtUri."""

    @abstractmethod
    def get_pxt_uri(self) -> PxtUri:
        """Get the PxtUri from this request. Must be implemented by subclasses."""
        pass
        
 most request models have pxturi in body some of them we have to put in there perhaps next round we can move this from uri to get_org and get_db (optional)

tests/test_pxt_store.py
```diff
@@ -0,0 +1,170 @@
...
+from pixeltable.utils.object_stores import ObjectOps, ObjectPath
+from pixeltable.utils.s3_store import S3Store
+
+from .utils import skip_test_if_no_pxt_credentials, skip_test_if_not_installed, validate_update_status
+
+PXT_DEST_URI = 'pxtfs://pixeltable:main/home/pytest'
+
+
+@pytest.mark.skip('Skip tests until pxt store changes are in the cloud')
+class TestPxtStore:
```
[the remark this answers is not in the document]
amithadke: this is intentional this is not an external service but pxt cloud service

pixeltable/utils/pxt_store.py
```diff
@@ -0,0 +1,299 @@
...
+                excs.ErrorCode.UNSUPPORTED_OPERATION,
+                f'Storage target {storage_target.value!r} not supported. Supported: {supported}',
+            )
+
+        return S3Store(soa, client=self._pxt_store_entry.client, resource=self._pxt_store_entry.resource)
+
+    def _to_logical_uri(self, physical_uri: str) -> str:
+        """Create object uri with logical bucket name"""
+        if not physical_uri.startswith('pxtfs://'):
+            return physical_uri
```
aaron-siegel: pxtfs:// is a physical URI?

aaron-siegel (approved):

pixeltable/utils/pxt_store.py
amithadke: we still pass pxtfs:// uri(soa) to underlying s3 store with custom client/resource
i will change name of variable

PR #1215, opened 2026-03-22
Improved type system for JsonType

aaron-siegel:
Revamps the type system for JsonType. `Json[]` can now be parameterized by any of:
- A `TypedDict` subclass
- A Pydantic model
- A `tuple[T1, T2, T3]` or `tuple[T, ...]` expression
- `list[T]` as a synonym for `tuple[T, ...]`
- A "convenience" structure of the form: `(T1, T2, T3)` or `(T1, T2, T3, ...)` or `{'key1': T1, 'key2': T2}` where in the convenience structures, `Ti` may be either a type or a nested convenience structure. `[T]` is a synonym for `(T, ...)`.

In all cases, the type argument is converted to an internal `JsonType.TypeSchema` representation. Literal type inference and supertype() give the expected results.

### DIFF
--- a/pixeltable/exprs/json_path.py
+++ b/pixeltable/exprs/json_path.py
@@ -68,8 +68,8 @@ class JsonPath(Expr):
     def __resolve_type(cls, col_type: ts.ColumnType, path_elements: list[str | int | slice]) -> ts.ColumnType:
         if len(path_elements) == 0:
             # JsonPath expressions always have `nullable=True`, regardless of the schema. This is because
-            # schema validation is optional in some runtime contexts (such as data import), so it's
-            # possible to encounter data at runtime that doesn't match the schema.
+            # schema validation is optional in some runtime contexts, so it's possible to encounter data
+            # at runtime that doesn't match the schema.
             return col_type.copy(nullable=True)
 
         el = path_elements[0]
--- a/tests/test_exprs.py
+++ b/tests/test_exprs.py
@@ -670,7 +670,12 @@ class TestExprs:
         t.add_computed_column(slice_range=t.c6.f5[3:7])
         t.add_computed_column(slice_range_step=t.c6.f5[3:7:2])
         t.add_computed_column(slice_range_step_item=t['c6'].f5[3:7:2])
-        res = t.collect()
+        t.add_computed_column(
+            list_of_dicts=[{'a': t.c2, 'b': t.c2}, {'a': t.c2 + 1, 'b': t.c1}, {'a': t.c2 + 2, 'b': t.c3}]
+        )
+        t.add_computed_column(item_of_list=t.list_of_dicts[:].a)
+        t.add_computed_column(item_of_vartype_list=t.list_of_dicts[:].b)
+        res = t.order_by(t.c2).collect()
         orig = res['attr']
         assert all(res['item'][i] == orig[i] for i in range(len(res)))
         assert all(res['index'][i] == orig[i][2] for i in range(len(res)))
@@ -680,6 +685,10 @@ class TestExprs:
         assert all(res['slice_range'][i] == orig[i][3:7] for i in range(len(orig)))
         assert all(res['slice_range_step'][i] == orig[i][3:7:2] for i in range(len(orig)))
         assert all(res['slice_range_step_item'][i] == orig[i][3:7:2] for i in range(len(orig)))
+        assert all(res['item_of_list'][i] == [i, i + 1, i + 2] for i in range(len(res)))
+        assert all(
+            res['item_of_vartype_list'][i] == [res['c2'][i], res['c1'][i], res['c3'][i]] for i in range(len(res))
+        )
 
     def test_json_path_types(self, uses_db: None) -> None:
         spec = {
@@ -697,6 +706,7 @@ class TestExprs:
                 ...,
             ),
             'f4': ({'f4a': int, 'f4b': str}, ...),
+            'f5': ({'f5a': int}, {'f5a': str}, {'f5a': float}, ...),
         }
         t = pxt.create_table('test', {'col': pxt.Json[spec]})
         cases: tuple[tuple[Expr, type], ...] = (
@@ -723,6 +733,8 @@ class TestExprs:
             # negative slice on variadic tuple
             (t.col.f3[-9:], pxt.Json[[pxt.Array[(2, None, None, None), np.float32]]]),
             (t.col.f4[7:14].f4a, pxt.Json[[int | None]]),  # dict resolution applied to list
+            # dict resolution applied to heterogeneous tuple
+            (t.col.f5[:].f5a, pxt.Json[(int | None, str | None, float | None, ...)]),
             (t.col.f4['*'].f4b, pxt.Json[[str | None]]),  # special '*' operator
         )
         for expr, expected_type in cases:


pixeltable/type_system.py
```diff
@@ -383,10 +391,73 @@ def from_python_type(
...
+            ),
+            nullable=nullable_default,
+        )
+
+    @classmethod
+    def __from_pydantic_model_type(cls, nullable_default: bool, t: type[pydantic.BaseModel]) -> JsonType:
+        fields: dict[str, ColumnType] = {}
+        optional_keys: list[str] = []
+        for name, info in t.model_fields.items():
+            fields[name] = cls.from_python_type(info.annotation)
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/type_system.py
```diff
@@ -851,15 +992,37 @@ def print_value(self, val: Any) -> str:
...
+        if not cls.__is_valid_json(val):
+            return None
+
+        type_schema: JsonType.TypeSchema
+        if isinstance(val, (list, tuple)):
+            type_schema = JsonType.TypeSchema(type_spec=[cls.infer_literal_type(item) for item in val])
+        else:
+            type_schema = JsonType.TypeSchema(
+                type_spec={key: cls.infer_literal_type(value) for key, value in val.items()}
+            )
```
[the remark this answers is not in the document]
aaron-siegel: The real issue was that `__is_valid_json()` needs to consider the dtype of numpy arrays. It should be the case that a valid type is returned via from_literal() any time __is_valid_json() is True. Fixed.

tests/test_types.py
```diff
@@ -154,21 +137,82 @@ def test_from_python_type(self, init_env: None) -> None:
...
                 ArrayType((5, None, 3), dtype=FloatType(), nullable=False),
                 'Array[(5, None, 3), float32]',
             ),
-            Image[100, 200]: (ImageType(width=100, height=200, mode=None, nullable=False), 'Image[(100, 200)]'),  # type: ignore[misc]
-            Image[100, None]: (ImageType(width=100, height=None, mode=None, nullable=False), 'Image[(100, None)]'),  # type: ignore[misc]
-            Image[None, 200]: (ImageType(width=None, height=200, mode=None, nullable=False), 'Image[(None, 200)]'),  # type: ignore[misc]
-            Image[(100, 200), 'RGB']: (  # type: ignore[misc]
+            Json[list[int]]: (JsonType(JsonType.TypeSchema([], variadic_type=IntType())), 'Json[(Int, ...)]'),
+            Json[tuple[int, str]]: (JsonType(JsonType.TypeSchema([IntType(), StringType()])), 'Json[(Int, String)]'),
+            Json[tuple[int, ...]]: (JsonType(JsonType.TypeSchema([], variadic_type=StringType())), 'Json[(Int, ...)]'),
```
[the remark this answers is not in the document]
aaron-siegel: Good catch. It was getting masked by a duplicate dictionary key in the test cases. Fixed.

pixeltable/type_system.py
```diff
@@ -805,44 +876,114 @@ def _validate_literal(self, val: Any) -> None:
...
+            for key, value in json_type_arg.items():
+                if not isinstance(key, str):
+                    raise excs.Error(
+                        f'Invalid type schema: expected keys of type `str`; got type `{type(key).__name__}`'
+                    )
+                type_spec[key] = cls.from_json_type_arg_r(value)
+            return JsonType(JsonType.TypeSchema(type_spec))
+
+        # Anything else: Convert it to a ColumnType instance in the usual fashion.
+        return ColumnType.from_python_type(json_type_arg)
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/type_system.py
```diff
@@ -805,44 +876,114 @@ def _validate_literal(self, val: Any) -> None:
...
+                fixed_types = json_type_arg[:-2]
+
+            if Ellipsis in fixed_types:
+                raise excs.Error('Invalid type schema: `...` allowed only in last position')
+
+            return JsonType(
+                JsonType.TypeSchema(
+                    type_spec=[cls.from_json_type_arg_r(item) for item in fixed_types],
+                    variadic_type=cls.from_json_type_arg_r(variadic_type),
+                )
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/type_system.py
```diff
@@ -1454,11 +1717,7 @@ def __class_getitem__(cls, item: Any) -> _AnnotatedAlias:
         """
         `item` (the type subscript) must be a `dict` representing a valid JSON Schema.
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/type_system.py
```diff
@@ -383,10 +391,73 @@ def from_python_type(
...
+            nullable=nullable_default,
+        )
+
+    @classmethod
+    def __from_typed_dict(cls, nullable_default: bool, t: type) -> JsonType:
+        # It's a subclass of `TypedDict`.
+        return JsonType(
+            JsonType.TypeSchema(
+                type_spec={key: cls.from_python_type(value) for key, value in t.__annotations__.items()},
+                optional_keys=list(getattr(t, '__optional_keys__', [])),
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/type_system.py
```diff
@@ -383,10 +391,73 @@ def from_python_type(
...
+        if Ellipsis in type_args:
+            raise excs.Error('Invalid type schema: `...` allowed only in last position')
+
+        return JsonType(
+            JsonType.TypeSchema(
+                type_spec=[cls.from_python_type(type_arg) for type_arg in type_args],
+                variadic_type=cls.from_python_type(variadic_type) if variadic_type is not None else None,
+            ),
+            nullable=nullable_default,
+        )
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/type_system.py
```diff
@@ -383,10 +391,73 @@ def from_python_type(
...
+    @classmethod
+    def __from_list_type(cls, nullable_default: bool, type_args: tuple) -> JsonType:
+        if len(type_args) == 0:
+            return JsonType(nullable=nullable_default)  # treat unparameterized list as untyped JSON
+        if len(type_args) > 1:
+            raise excs.Error('Invalid type schema: `list` or `Sequence` must have at most one type argument')
+        return JsonType(
+            JsonType.TypeSchema(type_spec=[], variadic_type=cls.from_python_type(type_args[0])),
+            nullable=nullable_default,
+        )
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

tests/test_types.py
```diff
@@ -108,76 +88,201 @@ def test_infer(self, init_env: None) -> None:
...
-        }
-        for py_type, (pxt_type, string) in test_cases.items():
+            (Image['RGB'], ImageType(height=None, width=None, mode='RGB', nullable=False), "Image['RGB']"),
+            (Literal['a', 'b', 'c'], StringType(nullable=False), 'String'),
+            (Literal[1, 2, 3], IntType(nullable=False), 'Int'),
+            (Literal[1, 2.0, 3], FloatType(nullable=False), 'Float'),
+            (Literal['a', 'b', None], StringType(nullable=True), 'String'),  # noqa: PYI061
+        )
+        for py_type, pxt_type, string in test_cases:
             print(py_type)
```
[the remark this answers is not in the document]
aaron-siegel: Doesn't matter; the printed values will only be shown if the test fails.

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+        if self.type_schema is None:
             return 'Json'
-        elif 'title' in self.json_schema:
-            return f'Json[{self.json_schema["title"]}]'
         else:
-            return f'Json[{self.json_schema}]'
+            return f'Json[{self.type_schema!r}]'
+
+    @dataclasses.dataclass(frozen=True)
+    class TypeSchema:
```
mkornacker, line 1115: Missing a class docstring explaining the scope of this abstraction.

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+        type_spec: list[ColumnType] | dict[str, ColumnType]
+        variadic_type: ColumnType | None = None
```
mkornacker, line 1129: How does this relate to type_spec?

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+        optional_keys: list[str] = dataclasses.field(default_factory=list)
```
mkornacker: Same here. Also, what about a list[int | None]?

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+
+        def __post_init__(self) -> None:
```
mkornacker, line 1132: What's the need for this? It simply reflects the member var decls.

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
             return 'Json'
-        elif 'title' in self.json_schema:
-            return f'Json[{self.json_schema["title"]}]'
         else:
-            return f'Json[{self.json_schema}]'
+            return f'Json[{self.type_schema!r}]'
+
+    @dataclasses.dataclass(frozen=True)
+    class TypeSchema:
+        type_spec: list[ColumnType] | dict[str, ColumnType]
```
mkornacker, line 1128: Separate into list_spec and dict_spec? I think the code would look cleaner.

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+                assert all(col_type is not None for col_type in self.type_spec), self.type_spec
+            else:
+                assert all(col_type is not None for col_type in self.type_spec.values()), self.type_spec
+            assert self.variadic_type is None or isinstance(self.variadic_type, ColumnType), self.variadic_type
+
+        def validate_literal(self, val: Any) -> None:
+            if isinstance(self.type_spec, list):
+                if not isinstance(val, list):
+                    raise TypeError(f'Expected a list; got `{val.__class__.__name__}`: {val!r}')
+                if len(val) < len(self.type_spec):
```
mkornacker, line 1144: What if val == []?

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+                    raise TypeError(
+                        f'Too few items in list: expected {qualifier} {len(self.type_spec)}; got {len(val)}'
+                    )
+                for i, item in enumerate(val):
+                    if i < len(self.type_spec):
+                        expected_type = self.type_spec[i]
+                    elif self.variadic_type is not None:
+                        expected_type = self.variadic_type
+                    else:
+                        raise TypeError(
```
mkornacker: The logic would be a bit easier to follow if the structural validation happened first, followed by the traversal you're doing here. Just in terms of seeing that it's complete.

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+                    reprs.append(repr(self.variadic_type))
+                    reprs.append('...')
+                type_spec_str = ', '.join(reprs)
+                if len(reprs) == 1:
+                    type_spec_str += ','  # add trailing comma for single-item tuples
+                return f'({type_spec_str})'
+            else:
+                r = repr(self.type_spec)
+                if len(self.optional_keys) > 0:
+                    r += f', optional_keys={self.optional_keys}'
```
mkornacker: Why not inline '| None'?

pixeltable/type_system.py
```diff
@@ -805,44 +899,120 @@ def _validate_literal(self, val: Any) -> None:
...
+                )
+            return JsonType(JsonType.TypeSchema(type_spec=[], variadic_type=cls.from_json_type_arg_r(json_type_arg[0])))
+
+        if isinstance(json_type_arg, tuple):
+            # Convenience tuple, such as `(String, Int, Float)` or `(String, Int, ...)`. A single ellipsis is
+            # allowed only in last position.
+            variadic_type = None
+            if len(json_type_arg) > 0 and json_type_arg[-1] is Ellipsis:
+                if len(json_type_arg) == 1:
+                    raise ValueError('Invalid type schema: tuple with only `...` is not allowed')
```
mkornacker, line 962: Don't we already have that validation logic elsewhere?

pixeltable/type_system.py
```diff
@@ -851,20 +1021,44 @@ def print_value(self, val: Any) -> str:
...
             val = val.model_dump()
         if not self.__is_valid_json(val):
             raise TypeError(f'That literal is not a valid Pixeltable JSON object: {val}')
-        if self.__validator is not None:
-            self.__validator.validate(val)
+        if self.type_schema is not None:
+            self.type_schema.validate_literal(val)
 
     @classmethod
     def __is_valid_json(cls, val: Any) -> bool:
```
mkornacker: Misleading method name.

pixeltable/type_system.py
```diff
@@ -851,20 +1021,44 @@ def print_value(self, val: Any) -> str:
             return str(val)
         return val_type.print_value(val)
 
+    @classmethod
+    def from_literal(
```
mkornacker, line 1035: What if the literal contains None? Shouldn't that be an error?

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
-        # schemas as (for example) {'anyOf': [{'type': 'string'}, {'type': 'null'}]}. When finding the supertype
-        # of schemas that might be nullable, we first unpack the 'anyOf's, find the supertype of the underlyings,
-        # then reapply the 'anyOf' if appropriate. The top-level schema (i.e., JsonType.json_schema) is presumed
-        # to NOT be in this form (since nullability is indicated by the `nullable` field of the JsonType object),
-        # so this subtlety is applicable only to types that occur in subfields.
-        #
-        # There is currently no special handling of lists; distinct schemas with type 'array' will union to the
-        # generic {'type': 'array'} schema. This could be a TODO item if there is a need for it in the future.
-
+    def __superschema(cls, a: TypeSchema | None, b: TypeSchema | None) -> TypeSchema | None:
```
mkornacker: Having a TypeSchema.supertype() seems a bit cleaner.

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+                if b.variadic_type is not None:
+                    variadic_type = variadic_type.supertype(b.variadic_type)
+                    if variadic_type is None:
+                        return None  # incompatible variadic types
+            else:
+                variadic_type = b.variadic_type
+
+            joined_list: list[ColumnType] = []
+            for item_a, item_b in itertools.zip_longest(a.type_spec, b.type_spec):
+                if item_a is None:
```
mkornacker: How about always making the shorter list the first param, to simplify the logic below?

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+                    assert isinstance(item_b, ColumnType)
+                    # We're past the end of a's type args. The supertype will still be valid as long as the type of
+                    # item_b can be incorporated into the variadic type. This may result in a variadic parameter in
+                    # the supertype where none exists in either subtype.
```
mkornacker: Why?

pixeltable/type_system.py
```diff
@@ -890,102 +1084,210 @@ def supertype(self, other: ColumnType, for_inference: bool = False) -> JsonType
...
+                            return None  # existing variadic type is incompatible with the extra item_a
+                else:
+                    item_supertype = item_a.supertype(item_b)
+                    if item_supertype is None:
+                        return None  # incompatible types in this position
+                    joined_list.append(item_supertype)
+
+            return JsonType.TypeSchema(type_spec=joined_list, variadic_type=variadic_type)
+
+        if isinstance(a.type_spec, dict) and isinstance(b.type_spec, dict):
```
mkornacker: I think this logic would be easier to follow:
- joined_dict = a
- mark all keys not in b as optional
- merge in b

pixeltable/func/signature.py
```diff
@@ -191,7 +192,7 @@ def validate_args(self, bound_args: dict[str, 'exprs.Expr' | None], context: str
             # non-nullable parameters can still accept nullable arguments (since in that event, FunctionCall.eval()
             # detects the Nones and skips evaluation).
             if not (
-                param.col_type.is_supertype_of(arg.col_type, ignore_nullable=True)
+                param.col_type.is_supertype_of(arg.col_type, ignore_nullable=True, strict_json=False)
```
mkornacker, line 195: I don't think this flag makes sense. If both column and parameter have json typing information, it should be checked by is_supertype_of().

pixeltable/type_system.py
aaron-siegel: Done.

pixeltable/type_system.py
aaron-siegel: Clarified in the docstring.

pixeltable/type_system.py
aaron-siegel: Clarified in the docstring.

pixeltable/type_system.py
aaron-siegel: It's there to help catch bugs in the implementation (mypy isn't perfect). I can remove it, but I tend to take the view that if an assertion was needed at some point for debugging, then it has value as a guard. Do you feel strongly about removing it?

pixeltable/type_system.py
aaron-siegel: I'm not convinced it would be cleaner - let's discuss this one before I make this change (which would not be trivial).

pixeltable/type_system.py
aaron-siegel: If len(self.type_spec) > 0 then the fixed component of the tuple has positive length, so this should properly fail.

That is, the empty list [] does not match a type specifier such as `JsonType[(int, int)]`, which expects a fixed 2-tuple. But it *would* match `JsonType[[int]]` (equivalently, `JsonType[(int, ...)]`), which is variadic.

pixeltable/type_system.py
aaron-siegel: I don't follow this comment. Can you be more specific about what you're proposing?

pixeltable/type_system.py
aaron-siegel: I don't follow. Wouldn't this munge the reprs of "optional key" and "required nullable key" (which are not the same thing)?

pixeltable/type_system.py
aaron-siegel: I don't think so? Where?

(This logic is validating that the type argument to `Json[]` is well-formed)

pixeltable/type_system.py
aaron-siegel: Legacy issue. And it merely reflects the fact that JsonType is a misleading type name.

I can change it to something else (`__is_valid_pxt_json()`? does that work? if not, please give some examples of names you'd be happy with). But I don't think that addresses the real issue, which is that JsonType isn't actually json.

pixeltable/type_system.py
aaron-siegel: Please explain - What does it mean for a literal to "contain None"? And why would it be an error?

pixeltable/type_system.py
aaron-siegel: Done.

pixeltable/func/signature.py
aaron-siegel: It's there to address the typing issues with UDF calls that we discussed previously. (We had agreed to do Json type erasure in the case of UDF call validation - if we want to revisit this let's do it as a product decision.)

pixeltable/type_system.py
aaron-siegel: Added an example to the comment to clarify.

pixeltable/type_system.py
aaron-siegel: Good idea. Done.

pixeltable/type_system.py
aaron-siegel: Done, though I'm not sure it actually makes it any easier to follow.

pixeltable/type_system.py
```diff
@@ -383,10 +408,91 @@ def from_python_type(
...
+        return JsonType(
+            JsonType.TypeSchema(type_spec=[], variadic_type=cls.__from_python_type_or_exc(type_args[0])),
+            nullable=nullable_default,
+        )
+
+    @classmethod
+    def __from_python_type_or_exc(cls, t: type | _GenericAlias | None) -> ColumnType:
+        col_type = cls.from_python_type(t)
+        if col_type is None:
+            raise excs.Error(f'Python type found in type argument is not a valid Pixeltable type: {t.__name__}')
```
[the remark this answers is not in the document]
aaron-siegel: Simplified it to just {t!r}

pixeltable/type_system.py
```diff
@@ -383,10 +408,91 @@ def from_python_type(
...
+    def __from_typed_dict(cls, nullable_default: bool, t: type) -> JsonType:
+        # It's a subclass of `TypedDict`.
+        type_spec: dict[str, ColumnType] = {}
+        for key, value in t.__annotations__.items():
+            col_type = cls.from_python_type(value)
+            if col_type is None:
+                raise excs.Error(f'Field {key!r} in TypedDict `{t.__name__}` is not a valid Pixeltable type: {value!r}')
+            type_spec[key] = col_type
+        return JsonType(
+            JsonType.TypeSchema(type_spec=type_spec, optional_keys=list(getattr(t, '__optional_keys__', []))),
```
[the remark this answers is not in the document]
aaron-siegel: I changed optional_keys to a frozenset and used `sorted` on serialization.

pixeltable/type_system.py
aaron-siegel: Actually, I made some further simplifications and now I think it's great. Thanks.

pixeltable/exprs/json_path.py
```diff
@@ -36,17 +36,90 @@ def __init__(
...
+        if anchor is not None:
+            self.components = [anchor]
         self.scope_idx = scope_idx
         # NOTE: the _create_id() result will change if set_anchor() gets called;
         # this is not a problem, because _create_id() shouldn't be called after init()
         self.id = self._create_id()
         self.file_handles = {}
 
+    @classmethod
+    def __infer_type(cls, col_type: ts.ColumnType, path_elements: list[str | int | slice]) -> ts.ColumnType:
```
mkornacker: Is __resolve_type() more appropriate? Infer usually indicates you're inferring from values.

pixeltable/type_system.py
mkornacker, line 217: It's not clear at all what that means, and seems to anticipate the caller logic (which to me would indicate it should be handled in the caller).

mkornacker, line 1037: The docstring should indicate that this can return a jsontype with an InvalidType somewhere in it.

pixeltable/type_system.py
mkornacker: _is_valid_literal()

pixeltable/type_system.py
mkornacker: Doing all the checking that can result in pxt.Error() first, followed by the traversal logic.

pixeltable/type_system.py
```diff
@@ -1258,9 +1209,81 @@ def __repr__(self) -> str:
...
+                    assert item_b is not None  # since `a` was chosen to be the shorter type_spec
+                    if item_a is None:
+                        assert isinstance(item_b, ColumnType)
+                        # We're past the end of a's type args. The supertype will still be valid as long as
+                        # the type of item_b can be incorporated into the variadic type. This may result in a
+                        # variadic parameter in the supertype where none exists in either subtype.
+                        # Example of how this can happen:
+                        #     schema of a: (int, int)
+                        #     schema of b: (int, int, int, int)
+                        #     superschema: (int, int, int, ...)  (representing a variadic int of minimum length 2)
```
mkornacker, line 1246: Leave a todo to revisit. It's not clear this is good behavior despite being technically correct, maybe we need to see what other systems do here.

pixeltable/type_system.py
```diff
@@ -1270,7 +1293,7 @@ def as_dict(self) -> dict[str, Any]:
             return {
                 'type_spec': type_spec_d,
                 'variadic_type': self.variadic_type.as_dict() if self.variadic_type is not None else None,
-                'optional_keys': self.optional_keys,
+                'optional_keys': sorted(self.optional_keys),
```
mkornacker, line 1300: Why sorted()? If it's a correctness requirement, explain.

pixeltable/exprs/json_path.py
```diff
@@ -36,17 +36,90 @@ def __init__(
...
+        if len(path_elements) == 0:
+            return col_type.copy(nullable=True)
```
mkornacker, line 73: Why always nullable?

pixeltable/exprs/json_path.py
```diff
@@ -36,17 +36,90 @@ def __init__(
...
+
+        if not isinstance(col_type, ts.JsonType):
+            # There are more path elements, but we've arrived at something other than JsonType;
```
mkornacker, line 78: Why wouldn't this return InvalidType? In Postgres, if you give it a non-existent path, it returns null.

tests/functions/test_huggingface.py
```diff
@@ -174,6 +175,8 @@ def test_detr_for_object_detection(self, uses_db: None) -> None:
         assert 'orange' in label_text
         assert 'bowl' in label_text
         assert 'broccoli' in label_text
+        # Test appropriate typing
+        assert t.featured_object.col_type.is_string_type()
```
mkornacker: Use t.get_metadata() instead.

tests/test_exprs.py
```diff
@@ -677,6 +680,60 @@ def test_json_path(self, test_tbl: pxt.Table) -> None:
...
+                pxt.Array[(2, 5, 6, 8), np.float32],
+                pxt.Array[(2, 4, 7, 8), np.float32],
+                pxt.Array[(2, 4, 6, 9), np.float32],
+                ...,
+            ),
+        }
+        t = pxt.create_table('test', {'col': pxt.Json[spec]})
+        cases: tuple[tuple[Expr, type], ...] = (
+            (t.col.f1, pxt.String),
+            (t.col.not_a_field, pxt.Json),  # "invalid" field defaults to untyped json
```
mkornacker: Why? Returning a valid type here would turn dict key misspellings into jsontype (for computed columns defined by jsonpath).

pixeltable/exprs/json_path.py
aaron-siegel: We're inferring the type of the expr. But I don't feel strongly about it - changed to resolve_type.

pixeltable/type_system.py
aaron-siegel: Legacy issue & unrelated to the current PR. Happy to address this separately though - pls create a rough-edges ticket.

aaron-siegel: Ok, added, but note that ColumnType.from_literal() can also (directly) return an InvalidType, e.g., ColumnType.from_literal(None)

pixeltable/type_system.py
aaron-siegel: That isn't an accurate name, since a datetime is a valid literal, but not a valid part of JsonType, according to the present implementation.

pixeltable/type_system.py
aaron-siegel: Ok, done.

pixeltable/type_system.py
aaron-siegel: Can you give me some more understanding of why it might not be good behavior?

What would you prefer to do instead? Fall back on schemaless Json?

pixeltable/type_system.py
aaron-siegel: It's not. It's purely so that the dict is a canonical representation. Nothing would go wrong if `sorted()` were replaced by `list()` here.

pixeltable/exprs/json_path.py
aaron-siegel: JsonPath always has to be able to resolve to None if it receives input that doesn't match the schema, right?

pixeltable/exprs/json_path.py
aaron-siegel: Because we're not doing strict type-checking of json schemas in columns - so the path might resolve to something non-null at runtime.

I'm not sure the postgres analogy is accurate here - you're talking about non-existent paths in the objects at runtime. Pixeltable does the same thing *at runtime*. But here we're talking about type schema interpretation, which doesn't have a Postgres analog. And the issue is that because we're doing loose typing, it's possible for the path to be invalid according to the type schema, but then at runtime to see an object for which it *is* a valid path.

(But see the related comment below)

tests/functions/test_huggingface.py
aaron-siegel: Why?

tests/test_exprs.py
aaron-siegel: Ok. I thought that's what we wanted - "loose" interpretation of json schemas and json paths, so that at runtime, `t.json_col.any_key` (for example) always resolves based on the observed values, independent of the type schema on the column.

Of course it would be fairly trivial to adjust this behavior. It probably merits another discussion, but: I don't think returning InvalidType is a good idea here, because of the inconsistency with what one would observe at runtime. cf. the preceding comment.

But it'd be reasonable to simply throw an exception here (at expression construction time), i.e., to say it's illegal to construct that JsonPath at all: then a JsonPath would have to be consistent with the type schema in order to make it to runtime in the first place.

pixeltable/type_system.py
aaron-siegel: re: "maybe we need to see what other systems do here." -

I did a test in Python. mypy agrees that `tuple[int, int]` and `tuple[int, int, int, int]` are incompatible, and it also agrees that `tuple[int, ...]` can accept either one. So it agrees with the principle: variadic tuples are supertypes of fixed tuples, as long as their type signatures align.

(Of course in Python, the variadic tuple is *a* supertype, but not "the" supertype - that would be `tuple[int, int] | tuple[int, int, int, int]`. But our type system doesn't have union types.)

tests/test_exprs.py
mkornacker: Yes, let's go with that (exception here).

tests/functions/test_huggingface.py
mkornacker: Because that's the public API. col_type is an implementation detail of ColumnRef.

pixeltable/type_system.py
mkornacker: Leave comment?

pixeltable/exprs/json_path.py
mkornacker: Explain in comment?

pixeltable/exprs/json_path.py
mkornacker: As per discussion below, let's raise here.

pixeltable/type_system.py
mkornacker: In your example: it might be correct to have a 2-tuple and a 4-tuple, but turning this into a variadic type now allows data that may no longer be correct (5-tuple?).

I can see how this may well be application-dependent.

pixeltable/type_system.py
aaron-siegel: Done

pixeltable/exprs/json_path.py
aaron-siegel: Done

pixeltable/type_system.py
aaron-siegel: I changed it to __is_valid_json_literal()

tests/functions/test_huggingface.py
aaron-siegel: Ok sure, done

pixeltable/exprs/json_path.py
aaron-siegel: Done

pixeltable/exprs/json_path.py
```diff
@@ -50,30 +50,62 @@ def __init__(
...
+            step_str = '' if el.step is None else f':{el.step}'
+            return f'[{start_str}:{stop_str}{step_str}]'
+        else:
+            raise AssertionError(f'Invalid JsonPath element: {el}')
+
     @classmethod
     def __resolve_type(cls, col_type: ts.ColumnType, path_elements: list[str | int | slice]) -> ts.ColumnType:
         if len(path_elements) == 0:
+            # JsonPath expressions always have `nullable=True`, regardless of the schema. This is because
+            # schema validation is optional in some runtime contexts (such as data import), so it's
```
mkornacker: Remove data import reference, we don't know whether we'll ever do schema validation at that point (and we don't do it today).

aaron-siegel: Done

PR #1221, opened 2026-03-26
[PXT-970 + PXT-1074] Gemini video reference images, TTS, and transcription

christopherpestano:
Adds Veo 3.1 reference image support to generate_videos via an overloaded signature that accepts a list of up to 3 reference images with corresponding reference types (style, subject, asset). The existing single-image signature is unchanged.

Also adds two new Gemini UDFs: generate_speech for text-to-speech using Gemini's TTS models, and transcribe for speech-to-text audio understanding.


pixeltable/functions/gemini.py
```diff
@@ -271,6 +271,37 @@ def _(model: str) -> str:
...
+        await asyncio.sleep(3)
+        operation = await _genai_client().aio.operations.get(operation)
+
+    if operation.error:
+        raise Exception(f'Video generation failed: {operation.error}')
+
+    video = operation.response.generated_videos[0]
+
+    video_bytes = await _genai_client().aio.files.download(file=video.video)  # type: ignore[arg-type]
+    assert video_bytes is not None
```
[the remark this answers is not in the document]
christopherpestano: Nah this is fine udf failures aren't user visible

pixeltable/functions/gemini.py
```diff
@@ -271,6 +271,37 @@ def _(model: str) -> str:
...
+
+async def _generate_videos_impl(
+    model: str, prompt: str | None, image: 'genai.types.Image | None', config: 'genai.types.GenerateVideosConfig | None'
+) -> str:
+    """Shared implementation for video generation: submit request, poll for completion, download result."""
+    operation = await _genai_client().aio.models.generate_videos(model=model, prompt=prompt, image=image, config=config)
+    while not operation.done:
+        await asyncio.sleep(3)
+        operation = await _genai_client().aio.operations.get(operation)
+
```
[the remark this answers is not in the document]
christopherpestano: added timeout of 5 minutes, realistically this isn't likely to be hit but nice to have just in case

pixeltable/functions/gemini.py
```diff
@@ -317,40 +362,210 @@ async def generate_videos(
...
 @generate_videos.resource_pool
 def _(model: str) -> str:
     return f'rate-limits:gemini:{model}'
 
 
+@pxt.udf(is_deterministic=False)
+async def generate_speech(text: str, *, model: str, voice: str = 'Kore', config: dict | None = None) -> pxt.Audio:
+    """
+    Generates speech audio from text using Gemini's text-to-speech capability. For additional details, see:
+    <https://ai.google.dev/gemini-api/docs/speech-generation>
```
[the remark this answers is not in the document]
christopherpestano: added

pixeltable/functions/gemini.py
```diff
@@ -317,40 +362,210 @@ async def generate_videos(
...
-    video_bytes = await _genai_client().aio.files.download(file=video.video)  # type: ignore[arg-type]
-    assert video_bytes is not None
+    Args:
+        prompt: A text description of the videos to generate.
+        image: A list of up to 3 reference images to guide style or asset appearance.
+        model: The model to use.
+        config: Configuration for generation.
+        reference_types: A list of reference types corresponding to each image. Each must be one of
+            `'style'` or `'asset'`. `'style'` is only supported on Vertex AI. If not provided, defaults to
+            `'asset'` for all images.
```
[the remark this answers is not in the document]
christopherpestano: removed subject

pixeltable/functions/gemini.py
```diff
@@ -317,40 +362,210 @@ async def generate_videos(
...
 
-    if operation.error:
-        raise Exception(f'Video generation failed: {operation.error}')
 
-    video = operation.response.generated_videos[0]
+@generate_videos.overload
+async def _(
+    prompt: str | None = None,
+    image: list[PIL.Image.Image] = [],  # noqa: B006
+    *,
```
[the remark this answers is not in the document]
christopherpestano: already removed

pixeltable/functions/gemini.py
```diff
@@ -317,40 +367,284 @@ async def generate_videos(
...
+
+    if reference_types is None:
+        reference_types = ['asset'] * len(image)
+    elif len(reference_types) != len(image):
+        raise excs.Error(f'`reference_types` length ({len(reference_types)}) must match `image` length ({len(image)}).')
+
+    reference_images = [
+        types.VideoGenerationReferenceImage(image=_pil_to_gemini_image(img), reference_type=ref_type)
+        for img, ref_type in zip(image, reference_types)
+    ]
```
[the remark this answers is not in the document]
christopherpestano: Hard to validate, can't really be done on client side

tests/functions/test_gemini.py
```diff
@@ -204,6 +204,84 @@ def test_generate_videos(self, uses_db: None) -> None:
...
+
+        t = pxt.create_table('test_tbl', {'prompt': pxt.String, 'ref1': pxt.Image, 'ref2': pxt.Image})
+        t.add_computed_column(
+            output=generate_videos(
+                t.prompt,
+                image=[t.ref1, t.ref2],
+                reference_types=['asset', 'asset'],
+                model='veo-3.1-generate-preview',
+                config={'duration_seconds': 8},
+            )
```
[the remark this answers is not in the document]
christopherpestano: Bad

pixeltable/functions/gemini.py
```diff
@@ -317,40 +370,287 @@ async def generate_videos(
...
+    env.Env.get().get_resource_pool_info(resource_pool_id, GeminiRateLimitsInfo)
+
+    config_ = types.GenerateContentConfig(**config) if config else None
+    client = _genai_client()
+
+    try:
+        size_bytes = os.stat(audio).st_size
+    except OSError as exc:
+        raise excs.Error(f"Error accessing audio file '{audio}': {exc.strerror or exc}") from exc
+    if size_bytes > GEMINI_INLINE_LIMIT_BYTES:
```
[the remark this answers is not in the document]
christopherpestano: This behavior is useful for debugging shared file access issues.

pixeltable/functions/gemini.py
```diff
@@ -271,6 +271,45 @@ def _(model: str) -> str:
...
+
+
+async def _generate_videos_impl(
+    model: str, prompt: str | None, image: 'genai.types.Image | None', config: 'genai.types.GenerateVideosConfig | None'
+) -> str:
+    """Shared implementation for video generation: submit request, poll for completion, download result."""
+    operation = await _genai_client().aio.models.generate_videos(model=model, prompt=prompt, image=image, config=config)
+
+    async def _poll() -> None:
+        nonlocal operation
```
aaron-siegel: Cleaner would be for `_poll()` to take a parameter `operation: genai.types.GenerateVideosOperation`, and return the `operation` that's left at the end. This would eliminate the need for the `nonlocal`, and `_poll()` could even be moved out into module scope (and renamed something like `_poll_gemini_op()`). I think this would make the logic easier to follow.

pixeltable/functions/gemini.py
```diff
@@ -317,40 +370,287 @@ async def generate_videos(
...
+    return await _generate_videos_impl(model, prompt, image_, config_)
 
-    if operation.error:
-        raise Exception(f'Video generation failed: {operation.error}')
 
-    video = operation.response.generated_videos[0]
+@generate_videos.overload
+async def _(
+    prompt: str | None = None,
+    image: list[PIL.Image.Image] | None = None,
```
aaron-siegel: Call this `images` in this signature?

pixeltable/functions/gemini.py
```diff
@@ -317,40 +370,287 @@ async def generate_videos(
...
+    Examples:
+        Add a computed column that generates speech from text:
+
+        >>> tbl.add_computed_column(
+        ...     audio=generate_speech(
+        ...         tbl.text, model='gemini-2.5-flash-preview-tts', voice='Kore'
+        ...     )
+        ... )
+    """
+    import wave
```
aaron-siegel: Move below require_package()

pixeltable/functions/gemini.py
```diff
@@ -317,40 +370,287 @@ async def generate_videos(
...
+    __Requirements:__
+
+    - `pip install google-genai`
+
+    Args:
+        text: The text to synthesize into speech.
+        model: The model to use (e.g. `'gemini-2.5-flash-preview-tts'`).
+        voice: The voice profile to use. Supported voices include `'Kore'`, `'Puck'`, `'Charon'`,
+            `'Fenrir'`, `'Aoede'`, `'Leda'`, `'Orus'`, `'Zephyr'`, and others. See the
+            `speech generation docs <https://ai.google.dev/gemini-api/docs/speech-generation>`_ for the full list.
```
aaron-siegel: Why the backticks and underscore here?

The usual markdown syntax would be
```
[speech generation docs](https://ai.google.dev/gemini-api/docs/speech-generation)
```

pixeltable/functions/gemini.py
```diff
@@ -317,40 +370,287 @@ async def generate_videos(
...
+    return await _generate_videos_impl(model, prompt, None, config_)
 
 
 @generate_videos.resource_pool
 def _(model: str) -> str:
     return f'rate-limits:gemini:{model}'
 
 
+@pxt.udf(is_deterministic=False)
+async def generate_speech(text: str, *, model: str, voice: str = 'Kore', config: dict | None = None) -> pxt.Audio:
```
aaron-siegel: Do we really want a default for `voice` (versus requiring the user to specify it explicitly)?

pixeltable/functions/gemini.py
```diff
@@ -317,40 +370,287 @@ async def generate_videos(
...
+        wf.setnchannels(1)
+        wf.setsampwidth(2)
+        wf.setframerate(24000)
+        wf.writeframes(data)
+    return output_path
+
+
+@generate_speech.overload
+async def _(text: str, *, model: str, voices: dict[str, str], config: dict | None = None) -> pxt.Audio:
+    """
```
aaron-siegel, line 505: Note this docstring won't be separately processed. (Perhaps it should be, but that would require updates to `pixeltable-doctools`.) Currently, any documentation of overloaded signatures needs to be in the primary docstring.

pixeltable/functions/gemini.py
```diff
@@ -317,40 +370,287 @@ async def generate_videos(
...
+
+
+@generate_speech.resource_pool
+def _(model: str) -> str:
+    return f'rate-limits:gemini:{model}'
+
+
+@pxt.udf(is_deterministic=False)
+async def transcribe(
+    audio: pxt.Audio, *, model: str, prompt: str = 'Generate a transcript of the speech.', config: dict | None = None
```
aaron-siegel: Here I would argue we definitely do not want a default prompt.

tests/functions/test_gemini.py
```diff
@@ -204,6 +204,84 @@ def test_generate_videos(self, uses_db: None) -> None:
...
+                    }
+                ]
+            ),
+            expected_rows=1,
+        )
+        results = t.collect()
+        file_path = results['output'][0]
+        assert Path(file_path).exists()
+
+    def test_generate_speech(self, uses_db: None) -> None:
```
aaron-siegel, line 242: These should be marked `expensive` (but not skipped).

aaron-siegel (changes requested):

tests/functions/test_gemini.py
christopherpestano: Hmm I thought speech gen/transcribe was relatively cheap. Marked as expensive.

pixeltable/functions/gemini.py
christopherpestano: Moved

pixeltable/functions/gemini.py
christopherpestano: Hmm but for transcription what other prompt would the user put? Regardless removed it since it's a bit misaligned with the overall style in this file.

pixeltable/functions/gemini.py
christopherpestano: Moved the important part (the aspect about images being an alternative parameter) to the main docstring. That's a good point it probably should be if we end up leveraging this heavily, I think once there's more than 2 function signatures the main docstring is going to get confusing.

pixeltable/functions/gemini.py
christopherpestano: Fair enough, removed.

pixeltable/functions/gemini.py
christopherpestano: I was just copy pasting from the top of the docstring where we do it with <>, removed.

pixeltable/functions/gemini.py
christopherpestano: Changed

pixeltable/functions/gemini.py
christopherpestano: Sure split it out and made it generic over the base operation rather than the video generation operation subclass.

pixeltable/functions/gemini.py
christopherpestano: Looked at docs and realized transcription can also include audio understanding like "Describe what happens in this audio" removed default prompt and added doc to indicate this behavior.

aaron-siegel: Sometimes it's useful to put more specific instructions when prompting understanding operations. "Here is an audio recording of a conversation about an insurance claim. Make a list of all place names that are mentioned in the audio recording." etc.

pixeltable/functions/gemini.py
```diff
@@ -265,6 +266,46 @@ def _(model: str) -> str:
...
+        operation = await asyncio.wait_for(_poll_gemini_operation(operation), timeout=300)  # type: ignore[arg-type]
+    except asyncio.TimeoutError as exc:
+        raise excs.Error(f'Video generation timed out after 300 seconds for Gemini model {model!r}.') from exc
+
+    if operation.error:
+        raise excs.Error(f'Video generation failed: {operation.error}')
+
+    video = operation.response.generated_videos[0]
+
+    video_bytes = await _genai_client().aio.files.download(file=video.video)  # type: ignore[arg-type]
```
aaron-siegel: The `# type: ignore` shouldn't be necessary here. Either make the type of _poll_gemini_operation into `genai.types.GenerateVideosOperation`; or, if your goal is to make it more general-purpose, use a type variable:
```
T = TypeVar('T', bound='genai.types.Operation')

async def _poll_gemini_operation(operation: T) -> T: ...
```
Then we'll have a typesafe implementation.

(Unless there's something else going on that I'm missing here)

christopherpestano: This one is technically necessary since the async overload for files.download doesn't accept GeneratedVideo as a type parameter whereas the sync version does. More of an oversight on the library itself. Though adding the typevar does remove a type ignore further up so still put it anyways.

PR #1225, opened 2026-03-28
[PXT-1002] Catalog refactoring in begin_xact

sergey-mkhitaryan:
1. extract a part of `begin_xact()` logic into a separate function
2. `_acquire_path_locks` and `_acquire_tbl_lock` return the locked table UUIDs, so no additional traversal is necessary later to collect locked table ids.

### DIFF
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -354,11 +354,11 @@ class Catalog:
                             raise
                         except sql_exc.DBAPIError as e:
                             # Handle retriable errors
+                            has_exc = True
                             if isinstance(
                                 e.orig, (psycopg.errors.SerializationFailure, psycopg.errors.LockNotAvailable)
                             ) and (num_retries < _MAX_RETRIES or _MAX_RETRIES == -1):
                                 _logger.debug(f'Retriable error {type(e.orig)} on attempt {num_retries}')
-                                has_exc = True
                                 num_retries += 1
                                 time.sleep(random.uniform(0.1, 0.5))
                                 conn.rollback()  # attempt failed -- don't try to commit the transaction before retrying


pixeltable/catalog/catalog.py
```diff
@@ -341,38 +334,21 @@ def begin_xact(
...
+                                finalize_pending_ops=finalize_pending_ops,
+                                num_retries=num_retries,
+                            )
+                            if not success:
+                                assert not self._undo_actions  # We should not have any undo actions at this point
+                                has_exc = True
+                                num_retries += 1
+                                time.sleep(random.uniform(0.1, 0.5))
+                                conn.rollback()  # attempt failed -- don't try to commit the transaction before retrying
+                                continue
```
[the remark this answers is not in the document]
sergey-mkhitaryan: verified in a test (conn.rollback inside a context manager with no exception), this is not a problem, and it makes the intention clearer.

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +406,66 @@ def begin_xact(
...
+                        self.validate()
+                else:
+                    # didn't get the write lock (e.g. table is a snapshot)
+                    for_write = False
+
+        except sql_exc.DBAPIError as e:
+            if isinstance(e.orig, (psycopg.errors.SerializationFailure, psycopg.errors.LockNotAvailable)) and (
+                num_retries < _MAX_RETRIES or _MAX_RETRIES == -1
+            ):
+                _logger.debug(f'Retriable error {type(e.orig)} on attempt {num_retries}')
```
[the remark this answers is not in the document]
sergey-mkhitaryan: don't care enough

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +406,66 @@ def begin_xact(
...
+                    if lock_mutable_tree:
+                        self._compute_column_dependents(locked_ids)
+                    if _logger.isEnabledFor(logging.DEBUG):
+                        # validate only when we don't see errors
+                        self.validate()
+                else:
+                    # didn't get the write lock (e.g. table is a snapshot)
+                    for_write = False
+
+        except sql_exc.DBAPIError as e:
```
mkornacker: What about operationalerror?

Add comment explaining which types of errors we need to catch here.

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +406,66 @@ def begin_xact(
...
+        self,
+        tbl: TableVersionPath | None,
+        tbl_id: UUID | None,
+        for_write: bool,
+        lock_mutable_tree: bool,
+        finalize_pending_ops: bool,
+        num_retries: int,
+    ) -> tuple[bool, bool]:
+        """
+        Acquires transactional locks on the specified tables, and updates self._x_locked_tbl_ids accordingly.
```
mkornacker, line 422: I'm assuming this should only get called with either tbl or tbl_id? Also assert that and point out in the comment.

_acquire_locks() is better, to avoid confusion with _acquire_tbl_lock().

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +406,66 @@ def begin_xact(
...
+                        tbl_id=tbl_id,
+                        for_write=for_write,
+                        lock_mutable_tree=lock_mutable_tree,
+                        raise_if_not_exists=True,
+                        check_pending_ops=finalize_pending_ops,
+                    )
+                )
+
+            if for_write:
+                if locked_ids:
```
mkornacker: Explicitly len(locked_ids) > 0

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +406,66 @@ def begin_xact(
...
+        tbl_id: UUID | None,
+        for_write: bool,
+        lock_mutable_tree: bool,
+        finalize_pending_ops: bool,
+        num_retries: int,
+    ) -> tuple[bool, bool]:
+        """
+        Acquires transactional locks on the specified tables, and updates self._x_locked_tbl_ids accordingly.
+
+        Returns (success, for_write). Success is False if a retriable error happened. When success is True, for_write
```
mkornacker: Returns (success, is_x_locked)...

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +406,66 @@ def begin_xact(
...
+        can be False if the table is not mutable.
+
+        Raises an error if a non-retriable error happened, or if the number of attempts exceeded the limit.
+        """
+        try:
+            locked_ids: set[UUID] = set()
```
mkornacker: x_locked_ids is even clearer.

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +406,66 @@ def begin_xact(
...
+                if locked_ids:
+                    self._x_locked_tbl_ids = locked_ids
+                    if lock_mutable_tree:
+                        self._compute_column_dependents(locked_ids)
+                    if _logger.isEnabledFor(logging.DEBUG):
+                        # validate only when we don't see errors
+                        self.validate()
+                else:
+                    # didn't get the write lock (e.g. table is a snapshot)
+                    for_write = False
```
mkornacker: Best to introduce a new variable (is_x_locked?) than repurposing a parameter.

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +406,66 @@ def begin_xact(
...
+
+        except sql_exc.DBAPIError as e:
+            if isinstance(e.orig, (psycopg.errors.SerializationFailure, psycopg.errors.LockNotAvailable)) and (
+                num_retries < _MAX_RETRIES or _MAX_RETRIES == -1
+            ):
+                _logger.debug(f'Retriable error {type(e.orig)} on attempt {num_retries}')
+                return (False, for_write)
```
mkornacker: Why isn't this (False, False)?

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +406,66 @@ def begin_xact(
...
+                        lock_mutable_tree=lock_mutable_tree,
+                        raise_if_not_exists=True,
+                        check_pending_ops=finalize_pending_ops,
+                    )
+                )
+
+            if for_write:
+                if locked_ids:
+                    self._x_locked_tbl_ids = locked_ids
+                    if lock_mutable_tree:
```
mkornacker: This has nothing to do with locking and should still be done in the caller.

pixeltable/catalog/catalog.py
sergey-mkhitaryan: `OperationalError` is a sublclass of `DBAPIError`.

Left a comment that this attempts to handle retriable errors.

pixeltable/catalog/catalog.py
sergey-mkhitaryan: > I'm assuming this should only get called with either tbl or tbl_id? Also assert that and point out in the comment.

There's already an assert in `begin_xact()` for that, but really I don't think that it's a fundamental limitation. With maybe a few minor tweaks, we should be able to lock multiple tables for one transaction, and that is coming in the future change.

> _acquire_locks() is better, to avoid confusion with _acquire_tbl_lock().

renamed

mkornacker: I meant: you probably want to specify the lock target either as TVP or as an id.

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +411,56 @@ def begin_xact(
...
+                    self._acquire_tbl_lock(
+                        tbl_id=tbl_id,
+                        for_write=for_write,
+                        lock_mutable_tree=lock_mutable_tree,
+                        raise_if_not_exists=True,
+                        check_pending_ops=finalize_pending_ops,
+                    )
+                )
+
+            self._x_locked_tbl_ids = x_locked_ids if for_write else set()
```
mkornacker: Could x_locked_ids be empty if for_write==False? Why not just a copy here?

pixeltable/catalog/catalog.py
```diff
@@ -581,9 +600,11 @@ def _acquire_tbl_lock(
         if row is None:
             if raise_if_not_exists:
                 raise excs.Error(self._dropped_tbl_error_msg(tbl_id))
-            return None  # nothing to lock
+            return set()  # nothing to lock
         tbl_md = schema.md_from_dict(schema.TableMd, row.md)
+        locked: set[UUID] = set()
         if for_write and tbl_md.is_mutable:
```
mkornacker, line 590: Should we instead have an 
assert tbl_md.is_mutable or not for_write
?

sergey-mkhitaryan: it breaks view creation, snapshot deletion, and some other more edgy scenarios

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +410,55 @@ def begin_xact(
...
+            if isinstance(lock_target, TableVersionPath):
+                x_locked_ids.update(
+                    self._acquire_path_locks(
+                        tbl=lock_target,
+                        for_write=for_write,
+                        lock_mutable_tree=lock_mutable_tree,
+                        check_pending_ops=finalize_pending_ops,
+                    )
+                )
+            if isinstance(lock_target, UUID):
```
mkornacker: elif

pixeltable/catalog/catalog.py
```diff
@@ -314,13 +314,6 @@ def begin_xact(
...
             return
 
-        # tv_msg = '\n'.join(
-        #     [
-        #         f'{tv.id}:{tv.effective_version} : tv={id(tv):x} sa_tbl={id(tv.store_tbl.sa_tbl):x}'
-        #         for tv in self._tbl_versions.values()
-        #     ]
-        # )
-        # _logger.debug(f'begin_xact(): {tv_msg}')
         num_retries = 0
```
mkornacker, line 317: Given that this is now being used in multiple places, let's make it a member var.

pixeltable/catalog/catalog.py
```diff
@@ -443,6 +410,55 @@ def begin_xact(
                 self._undo_actions.clear()
                 self._modified_tvs.clear()
 
+    def _acquire_locks(
+        self,
+        lock_target: TableVersionPath | UUID,
+        for_write: bool,
+        lock_mutable_tree: bool,
+        finalize_pending_ops: bool,
+        num_retries: int,
```
mkornacker: Remove parameter.

pixeltable/catalog/catalog.py
sergey-mkhitaryan: I didn't want to do that, and instead moved the retryable error handling back to where it was.

pixeltable/catalog/catalog.py
```diff
@@ -595,21 +600,23 @@ def _acquire_tbl_lock(
 
         # TODO: properly handle concurrency for replicas with live views (once they are supported)
         if for_write and not tbl_md.is_mutable:
-            return None  # nothing to lock
+            return set()  # nothing to lock
```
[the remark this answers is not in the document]
sergey-mkhitaryan: This is true, but it is the pre-existing behavior as well, and it doesn't seem to be causing any issues so far.

PR #1226, opened 2026-03-28
More video udfs

mkornacker:
Plus some refactoring.

### DIFF
--- a/pixeltable/utils/av.py
+++ b/pixeltable/utils/av.py
@@ -229,14 +229,7 @@ def run_ffmpeg_cmdline(
     """
     cmd = ['ffmpeg', *ffmpeg_args]
     if encode_video:
-        # append video encoder-related args
-        if video_encoder is None:
-            video_encoder = Env.get().default_video_encoder
-        if video_encoder is not None:
-            cmd.extend(('-c:v', video_encoder))
-        if video_encoder_args is not None:
-            for k, v in video_encoder_args.items():
-                cmd.extend((f'-{k}', str(v)))
+        append_video_encoder(cmd, video_encoder, video_encoder_args)
     # loglevel=error: avoid excessive logging
     cmd += ['-loglevel', 'error', output_path]
 


pixeltable/utils/av.py
```diff
@@ -46,21 +46,27 @@ def __get_stream_metadata(stream: av.stream.Stream) -> dict:
...
+        'duration_seconds': duration_seconds,
         'frames': stream.frames,
         'metadata': stream.metadata,
         'codec_context': codec_context_md,
     }
 
     if stream.type == 'audio':
         # Additional metadata for audio
-        channels = getattr(stream.codec_context, 'channels', None)
+        channels = cast(av.AudioCodecContext, stream.codec_context).channels
```
aaron-siegel: Safer:
```python
assert isinstance(stream.codec_context, av.AudioCodecContext)
channels = stream.codec_context.channels
```

pixeltable/utils/av.py
```diff
@@ -107,6 +129,75 @@ def has_audio_stream(path: str) -> bool:
...
 
+def get_segment_duration(path: str, approx_decoded_bytes: int) -> float | None:
+    """
+    Return the length of a segment for which the combined in-memory size of all its decoded frames is roughly
+    approx_decoded_bytes.
+
+    Returns None f the frame rate or dimensions cannot be determined.
+    """
+    # bytes of memory per pixel of decoded frames, by pixel format
+    bytes_per_pixel = {
```
aaron-siegel: Move to global scope?

pixeltable/utils/av.py
```diff
@@ -107,6 +129,75 @@ def has_audio_stream(path: str) -> bool:
     return any(stream['type'] == 'audio' for stream in md['streams'])
 
 
+def get_segment_duration(path: str, approx_decoded_bytes: int) -> float | None:
```
aaron-siegel: estimate_segment_duration() would be a more honest name. (has a generic fallback for unknown codecs; handles vfr in a crude way)

pixeltable/utils/av.py
```diff
@@ -107,6 +129,75 @@ def has_audio_stream(path: str) -> bool:
...
+        pix_fmt = video_stream.codec_context.pix_fmt
+
+        if width <= 0 or height <= 0:
+            return None
+
+        if video_stream.average_rate is None or video_stream.average_rate == 0:
+            return None
+        fps = float(video_stream.average_rate)
+
+    bpp = bytes_per_pixel.get(pix_fmt, 3.0)  # conservative fallback for unknown formats
```
aaron-siegel: 3.0 seems like a "magic constant"; best to declare it as a global constant (FALLBACK_BITS_PER_PIXEL or some such)

pixeltable/functions/video.py
```diff
@@ -271,6 +271,32 @@ def _handle_ffmpeg_error(e: subprocess.CalledProcessError) -> NoReturn:
...
+
+
+def _append_video_encoder(
+    cmd: list[str], video_encoder: str | None = None, video_encoder_args: dict[str, Any] | None = None
+) -> None:
+    """Append video encoder-related args to ffmpeg cmdline."""
+    if video_encoder is None:
+        video_encoder = Env.get().default_video_encoder
+    if video_encoder is not None:
+        cmd.extend(['-c:v', video_encoder])
```
aaron-siegel: It's preferred to use tuples for immutable sequences, as in:
```python
cmd.extend(('-c:v', video_encoder))
```
It's slightly more efficient since it guarantees Python will allocate exactly as much memory as needed to hold the sequence. Unlikely this will be a performance bottleneck but there are so many places that lists are being used instead of tuples, perhaps it eventually adds up. And in any case it's good style.

pixeltable/functions/video.py
```diff
@@ -271,6 +271,32 @@ def _handle_ffmpeg_error(e: subprocess.CalledProcessError) -> NoReturn:
     raise pxt.Error(error_msg) from e
 
 
+def _run_ffmpeg(cmd: list[str], output_path: str) -> str:
```
aaron-siegel: Good idea. It would seem logical to just take `args` rather than the full `cmd`, and then prepend `ffmpeg` within `_run_ffmpeg`, to localize the logic for discovering the ffmpeg binary. (e.g., perhaps in the future we'll want to give users a way to specify the location of their ffmpeg binary explicitly, via config; factoring that out here would simply that)

I'd also advocate relocating some more boilerplate here, such as appending `['-loglevel', 'error', output_path]`, and also appending the video encoder args (that is, _run_ffmpeg() could take optional video_encoder and video_encoder_args parameters, obviating the need for a separate call to _append_video_encoder()).

Also move this and _handle_ffmpeg_error to `utils/av.py`?

pixeltable/functions/video.py
```diff
@@ -1126,33 +1145,758 @@ def resize(
...
+    """
+    Reverse a video using ffmpeg's reverse filter.
+
+    __Requirements:__
+
+    - `ffmpeg` needs to be installed and in PATH
+
+    Args:
+        video: Input video.
+        audio: what to do with audio streams
```
aaron-siegel: Proper punctuation would make this appear less rough.

`audio: Specifies what to do with any audio streams:`

pixeltable/functions/video.py
```diff
@@ -1126,33 +1145,758 @@ def resize(
...
+    cmd = ['ffmpeg', '-i', str(video), '-vf', flip_filter, '-c:a', 'copy']
+    _append_video_encoder(cmd, video_encoder, video_encoder_args)
+    cmd.extend(['-loglevel', 'error', output_path])
+    _logger.debug(f'_flip({orientation}): {" ".join(cmd)}')
+
+    return _run_ffmpeg(cmd, output_path)
+
+
+@pxt.udf(is_method=True)
+def mirror_x(
```
aaron-siegel, line 1414: Why not a single `def mirror()` with an `axis: Literal['x', 'y']` parameter?

pixeltable/functions/video.py
```diff
@@ -1126,33 +1145,758 @@ def resize(
...
+        while remaining < 0.5:
+            atempo_parts.append('atempo=0.5')
+            remaining /= 0.5
+        while remaining > 100.0:
+            atempo_parts.append('atempo=100.0')
+            remaining /= 100.0
+        atempo_parts.append(f'atempo={remaining}')
+        cmd.extend(['-af', ','.join(atempo_parts)])
+    else:
+        cmd.extend(['-an'])
```
aaron-siegel: cmd.append('-an')

pixeltable/functions/video.py
```diff
@@ -1126,33 +1145,758 @@ def resize(
...
+    cmd = ['ffmpeg', '-i', str(video), '-vf', fade_filter, '-c:a', 'copy']
+    _append_video_encoder(cmd, video_encoder, video_encoder_args)
+    cmd.extend(['-loglevel', 'error', output_path])
+    _logger.debug(f'fade_{direction}(): {" ".join(cmd)}')
+
+    return _run_ffmpeg(cmd, output_path)
+
+
+@pxt.udf(is_method=True)
+def fade_in(
```
aaron-siegel, line 1242: It seems "free" to expose an optional start_time parameter, which I think is potentially useful?

pixeltable/utils/av.py
mkornacker: Done.

pixeltable/utils/av.py
mkornacker: Done.

pixeltable/utils/av.py
mkornacker: Done.

pixeltable/utils/av.py
mkornacker: Done.

pixeltable/functions/video.py

pixeltable/functions/video.py

pixeltable/functions/video.py

pixeltable/functions/video.py
mkornacker: This was chosen to emulate Moviepy filters. Their FadeIn doesn't have a start parameter.

pixeltable/functions/video.py

pixeltable/functions/video.py
mkornacker: Also chosen to emulate Moviepy

https://zulko.github.io/moviepy/reference/reference/moviepy.video.fx.MirrorX.html

pixeltable/functions/video.py
```diff
@@ -1047,32 +1029,22 @@ def crop(
     else:
         raise pxt.Error(f"bbox_format must be one of ['xyxy', 'xywh', 'cxcywh'], got {bbox_format!r}")
 
-    if video_encoder is None:
-        video_encoder = Env.get().default_video_encoder
-
+    cmd = ['-i', str(video), '-vf', f'crop={w}:{h}:{x}:{y}', '-c:a', 'copy']
```
aaron-siegel, line 1032: can be a tuple now? (here and in a bunch of other places)

pixeltable/utils/av.py
```diff
@@ -107,16 +135,146 @@ def has_audio_stream(path: str) -> bool:
...
+        # append video encoder-related args
+        if video_encoder is None:
+            video_encoder = Env.get().default_video_encoder
+        if video_encoder is not None:
+            cmd.extend(('-c:v', video_encoder))
+        if video_encoder_args is not None:
+            for k, v in video_encoder_args.items():
+                cmd.extend((f'-{k}', str(v)))
+    # loglevel=error: avoid excessive logging
+    cmd += ['-loglevel', 'error', output_path]
```
aaron-siegel, line 234: Or just build this into the original definition of cmd
```python
cmd = ['ffmpeg', *ffmpeg_args, 'loglevel', 'error', output_path]
```
?

pixeltable/utils/av.py
```diff
@@ -107,16 +135,146 @@ def has_audio_stream(path: str) -> bool:
...
+    video_encoder_args: dict[str, Any] | None = None,
+) -> str:
+    """
+    Create and run an ffmpeg commandline, verify the output file, and return its path.
+
+    If encode_video==True, command is expected to encode video and requires video encoder args
+    """
+    cmd = ['ffmpeg', *ffmpeg_args]
+    if encode_video:
+        # append video encoder-related args
```
aaron-siegel: Just call append_video_encoder() here?

mkornacker: Done.

pixeltable/utils/av.py
mkornacker: That would change the logic.

PR #1231, opened 2026-04-01
list_iterator

aaron-siegel:
Introduces a new JSON-focused iterator to flatten typed JSON lists into component views, and adjusts type/schema inference so Pixeltable can reliably derive iterator output schemas from Json list/tuple types.

Changes:

- Added pixeltable.functions.json.list_iterator (@pxt.iterator) to iterate over typed JSON lists (positional list-of-dicts mode, or keyword “zip lists” mode with strict/truncated/padded behavior).
- Updated iterator schema inference plumbing so conditional_output_schema() receives bound exprs.Expr arguments (not only literal args), and updated affected iterators/tests accordingly.
- Changed Python list/tuple type conversion to produce typed JSON array schemas (Json[(Json, ...)]) rather than untyped Json, and added tests + public API registration.

### DIFF
--- a/pixeltable/type_system.py
+++ b/pixeltable/type_system.py
@@ -1167,19 +1167,25 @@ class JsonType(ColumnType):
                         raise TypeError(f'Unexpected key: {key}')
 
         def to_json_schema(self) -> dict[str, Any]:
+            json_schema: dict[str, Any]
+
             if isinstance(self.type_spec, list):
-                json_schema: dict[str, Any] = {'type': 'array'}
+                json_schema = {'type': 'array'}
                 if len(self.type_spec) > 0:
                     json_schema['prefixItems'] = [t.to_json_schema() for t in self.type_spec]
                 json_schema['items'] = False if self.variadic_type is None else self.variadic_type.to_json_schema()
-                return json_schema
+
             else:
-                properties = {k: t.to_json_schema() for k, t in self.type_spec.items()}
+                json_schema = {
+                    'type': 'object',
+                    'properties': {k: t.to_json_schema() for k, t in self.type_spec.items()},
+                    'additionalProperties': False,
+                }
                 required = [k for k in self.type_spec if k not in self.optional_keys]
-                json_schema: dict[str, Any] = {'type': 'object', 'properties': properties, 'additionalProperties': False}
                 if len(required) > 0:
                     json_schema['required'] = required
-                return json_schema
+
+            return json_schema
 
         def __eq__(self, other: object) -> bool:
             return (


pixeltable/functions/json.py
```diff
@@ -59,6 +61,115 @@ def value(self) -> list[Any]:
...
+                not isinstance(expr.col_type, ts.JsonType)
+                or expr.col_type.type_schema is None
+                or not isinstance(expr.col_type.type_schema.type_spec, list)
+            ):
+                raise excs.Error(
+                    f'list_iterator(): Expected a type for `{name}` matching `list`; got `{expr.col_type}`'
+                )
+            common_supertype = ts.ColumnType.common_supertype(
+                [*expr.col_type.type_schema.type_spec, expr.col_type.type_schema.variadic_type]
+            )
```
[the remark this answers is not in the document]
aaron-siegel: Done.

pixeltable/functions/json.py
```diff
@@ -59,6 +61,115 @@ def value(self) -> list[Any]:
...
+
+            - `'strict'`: Raises an error if the input lists have different lengths.
+            - `'truncated'`: Iterates until the shortest input list is exhausted, ignoring any remaining elements in
+                longer lists.
+            - `'padded'`: Iterates until the longest input list is exhausted, yielding `None` for any missing
+                elements from shorter lists.
+        **kwargs: One or more lists to iterate over. The kwarg names will be used as column names in the output.
+            Cannot be specified together with `elements`.
+    """
+    assert (elements is None) != (len(kwargs) == 0)
```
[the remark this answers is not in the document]
aaron-siegel: Assertion is correct; if this condition is violated then validation will fail upstream.

pixeltable/func/iterator.py
```diff
@@ -415,17 +422,17 @@ def from_dict(cls, d: dict[str, Any]) -> 'GeneratingFunctionCall':
...
             # part of the schema migration. In that case we just query the iterator for it, but we lose the ability to
             # sanity check against any schema evolution of the iterator.
             if it.is_legacy_retrofit:
+                literal_args = {k: v.val for k, v in bound_args.items() if isinstance(v, exprs.Literal)}
+                for param_name, param in it.py_sig.parameters.items():
+                    if param_name not in bound_args and param.default is not inspect.Parameter.empty:
+                        literal_args[param_name] = param.default
                 _, unstored_cols = it.decorated_callable.output_schema(literal_args)  # type: ignore[attr-defined]
             else:
                 unstored_cols = it.unstored_cols
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/functions/json.py
```diff
@@ -59,6 +61,116 @@ def value(self) -> list[Any]:
...
+    else:  # bound_args.get('element') is None
+        mode = bound_args.get('mode')
+        kwargs = bound_args.get('kwargs', {})  # type: ignore[var-annotated]
+        if len(kwargs) == 0:
+            raise excs.Error('list_iterator(): No inputs provided')
+        type_schema: dict[str, ts.ColumnType] = {}
+        for name, expr in kwargs.items():
+            if (
+                not isinstance(expr.col_type, ts.JsonType)
+                or expr.col_type.type_schema is None
```
mkornacker, line 159: There's a lot of 'expr.col_type.type_schema'. If you assign that to a variable and use that, it'll make the code easier to read.

tests/functions/test_json.py
```diff
@@ -0,0 +1,154 @@
...
+                'col_1': pxt.Json[[{'a': pxt.Int}]],
+                'col_2': pxt.Json[[int]],
+                'col_3': pxt.String,  # primitive type
+                'col_4': pxt.Json,  # untyped json
+                'col_5': pxt.Json[list],  # typed json, but with an untyped subscript
+            },
+        )
+
+        with pytest.raises(pxt.Error, match=r'list_iterator\(\): `mode` argument cannot be used with `elements`'):
+            pxtf.json.list_iterator(t.col_1, mode='truncated')
```
mkornacker, line 133: Let's use that in the context of create_view(), to make sure we see the error there (and below).

pixeltable/functions/json.py
aaron-siegel: Done.

tests/functions/test_json.py
aaron-siegel: I'm not sure I see the point of this one. The `list_iterator()` call never completes, so wrapping it in a `create_view()` doesn't actually alter the test coverage, it just makes the test more confusing.

Put another way: if I have a call `f()` that raises an error, and I wrap it in another call like
```
g('x', y=f())
```
The expression evaluation will fail at the time `f()` is called, and `g()` will never be involved. So the only additional thing being tested beyond `f()`, is that Python expression evaluation happens according to the language spec.

PR #1233, opened 2026-04-02
[PXT-1096] Introduce unversioned tables

sergey-mkhitaryan:
1. Introduces a new, private `_versioned` parameter on table create.
2. Updates the metadata to include the new `versioned` and backfills existing catalogs.
3. Updates `StoreBase` to support unversioned tables without the row history.
4. Adds a basic test that verifies insert and select on an unversioned table. Update is not yet implemented.
5. Asserts on various codepaths that do not yet support, or do not make any sense with the new unversioned tables.


pixeltable/metadata/schema.py
```diff
@@ -208,10 +208,13 @@ class TableMd:
...
     name: str
     is_replica: bool
+    # Versioned tables keep their full schema and row history, and support time travel and rollback.
+    is_versioned: bool
 
     user: str | None
 
-    # monotonically increasing w/in Table for both data and schema changes, starting at 0
+    # for versioned tables, current_version monotonically increases for both data and schema changes, starting at 0
+    # not used for unversioned tables
```
mkornacker, line 215: not used but still maintained by unversioned tables?

pixeltable/metadata/schema.py
```diff
@@ -208,10 +208,13 @@ class TableMd:
     tbl_id: str  # uuid.UUID
     name: str
     is_replica: bool
+    # Versioned tables keep their full schema and row history, and support time travel and rollback.
+    is_versioned: bool
```
mkornacker: You can also add it at the end with a default, and then avoid having to backfill.

pixeltable/_query.py
```diff
@@ -353,6 +353,15 @@ async def _aexec(self) -> AsyncIterator[exprs.DataRow]:
                     yield row
 
     def _create_query_plan(self) -> exec.ExecNode:
```
mkornacker, line 502: You need to check for the mixed versioning case in join() and return a pxt.Error.

pixeltable/store.py
```diff
@@ -72,12 +74,16 @@ def storage_name(cls, tbl_id: UUID, is_view: bool) -> str:
...
+        if self.tbl_version.get().is_versioned:
+            return [*self._pk_cols, self.v_max_col]
+        else:
+            return self._pk_cols
 
     def pk_columns(self) -> list[sql.Column]:
         return self._pk_cols
 
     def rowid_columns(self) -> list[sql.Column]:
+        assert self.tbl_version.get().is_versioned, 'PXT-975 not implemented for unversioned tables'
```
mkornacker: Why not?

pixeltable/store.py
```diff
@@ -96,19 +102,33 @@ def _create_system_columns(self) -> list[sql.Column]:
                     f'SELECT column_name FROM information_schema.columns WHERE table_name = {self._storage_name()!r} '
                     'ORDER BY ordinal_position'
                 )
+                # System columns on a versioned table: rowid, [pos_0, pos_1, ...], v_min, v_max
```
mkornacker, line 135: This is now getting more complicated. Move a comprehensive description to the class-level docstring.

pixeltable/store.py
```diff
@@ -447,8 +470,15 @@ def insert_rows(
...
-                    rowid = (next(rowids),) if rowids is not None else row.pk[:-1]
-                    pk = (*rowid, v_min)
+                    if versioned:
+                        rowid = (next(rowids),) if rowids is not None else row.pk[:-1]
+                        pk = (*rowid, v_min)
+                    else:
+                        # pick a random rowid between 1 and max bigint
+                        # bigint range is [-9223372036854775808, 9223372036854775807], but negative rowids can be
+                        # confusing and difficult to work with.
+                        # For 100M rows, the probability of a collision is ~0.05%
```
mkornacker: Feels a bit high.

pixeltable/store.py
```diff
@@ -505,10 +536,36 @@ def _versions_clause(self, versions: list[int | None], match_on_vmin: bool) -> s
 
     def delete_rows(
```
mkornacker: Maybe best to split this into mark_rows_deleted() and delete_rows().

pixeltable/catalog/table_version.py
```diff
@@ -255,8 +255,10 @@ def __init__(
         self.idxs = {}
         self.idxs_by_name = {}
         self.idxs_by_col = {}
-        self.supports_idxs = self.effective_version is None or (
-            self.is_replica and self.effective_version == self.tbl_md.current_version
+        self.supports_idxs = (
+            not self.is_versioned
+            or self.effective_version is None
```
aaron-siegel: I wonder if we can simplify the logic by reducing the number of times we have to explicitly say `not self.is_versioned`.

For example, isn't it the case that `self.effective_version` is always `None` for unversioned tables? (Or if not, could we make that convention?) If so, the `not self.is_versioned` clause is superfluous here.

pixeltable/catalog/table_version.py
```diff
@@ -272,8 +274,10 @@ def versioned_name(self) -> str:
 
     def __repr__(self) -> str:
         return (
-            f'TableVersion(id={self.id!r}, name={self.name!r}, effective_version={self.effective_version}, '
-            f'anchor_tbl_id={self.anchor_tbl_id}; version={self.version})'
+            f'TableVersion(id={self.id!r}, name={self.name!r}, '
+            f'effective_version={self.effective_version if self.is_versioned else None}, '
```
aaron-siegel: Similarly - if `not self.is_versioned` implies `self.effective_version is None`, then this expression just reverts to `self.effective_version`.

pixeltable/catalog/table_version.py
```diff
@@ -370,6 +375,7 @@ def create_initial_md(
             external_stores=[],
             view_md=view_md,
             additional_md={},
+            is_versioned=versioned,
```
aaron-siegel: Why `versioned` in one place but `is_versioned` in the other?

pixeltable/catalog/table_version.py
```diff
@@ -606,6 +613,7 @@ def _store_idx_name(self, idx_id: int) -> str:
         return f'idx_{self.id.hex}_{idx_id}'
 
     def add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) -> UpdateStatus:
+        assert self.is_versioned, 'PXT-975 not implemented for unversioned tables'
```
aaron-siegel: this makes it sound like PXT-975 is the thing that's not implemented. clearer would be 'TODO: implement for unversioned tables [PXT-975]' (if indeed that's what's meant)

tests/test_unversioned_table.py
```diff
@@ -0,0 +1,103 @@
+import sqlalchemy as sql
+
+import pixeltable as pxt
+from pixeltable.runtime import get_runtime
+
+from .utils import ReloadTester, validate_update_status
+
+
+class TestUnversionedTable:
```
aaron-siegel, line 11: I guess you're anticipating adding significantly many more tests to this file, and that's why it's in a separate file rather than part of `test_table.py`?

pixeltable/catalog/table_version.py
sergey-mkhitaryan: renamed to `is_`

tests/test_unversioned_table.py
sergey-mkhitaryan: correct

pixeltable/catalog/table_version.py
sergey-mkhitaryan: The way I did it, `self.effective_version` will raise an error if `not self.is_versioned` because effective version doesn't make any sense at all for unversioned tables. Happy to discuss, but that's my preference because it's clearer even if at the expense of brevity.

pixeltable/metadata/converters/convert_50.py
```diff
@@ -0,0 +1,17 @@
...
+
+from pixeltable.metadata import register_converter
+from pixeltable.metadata.converters.util import convert_table_md
+
+
+@register_converter(version=50)
+def _(engine: sql.engine.Engine) -> None:
+    """Backfill the 'is_versioned' field in the pre-existing Table metadata"""
+    convert_table_md(engine, table_md_updater=_update_table_md)
+
```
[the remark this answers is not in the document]
sergey-mkhitaryan: good catch

pixeltable/catalog/catalog.py
```diff
@@ -235,12 +235,15 @@ def validate(self) -> None:
         for (tbl_id, effective_version, anchor_tbl_id), tbl_version in self._tbl_versions.items():
             assert tbl_id == tbl_version.id, f'{tbl_id} != {tbl_version.id}'
             assert effective_version is None or anchor_tbl_id is None
-            assert tbl_version.effective_version == tbl_version.version or tbl_version.effective_version is None, (
-                f'{tbl_version.effective_version} != {tbl_version.version} for id {tbl_id}'
-            )
-            assert effective_version == tbl_version.effective_version, (
-                f'{effective_version} != {tbl_version.effective_version} for id {tbl_id}'
-            )
+            assert (
```
mkornacker: Easier to follow with two branches:
if tbl_version.is_versioned:
...
else:
...

pixeltable/catalog/catalog.py
```diff
@@ -1856,8 +1860,12 @@ def get_tbl_version(
 
                 if tv.anchor_tbl_id is None:
                     # live non-replica table; compare our cached TableMd.current_version/view_sn to what's stored
-                    current_version, view_sn = row.md['current_version'], row.md['view_sn']
-                    if current_version != tv.version or view_sn != tv.tbl_md.view_sn:
+                    versioned, current_version, view_sn = (
```
mkornacker: It's called is_versioned in some places. Call is is_versioned everywhere?

Also: easier to read as 3 separate assignments, since it now requires a multi-line statement.

pixeltable/catalog/table_version.py
mkornacker: I agree with Aaron's logic: if effective_version == None means it's the current/live version, then it should also apply to unversioned tables.

pixeltable/catalog/table_version.py
```diff
@@ -272,8 +274,10 @@ def versioned_name(self) -> str:
 
     def __repr__(self) -> str:
         return (
-            f'TableVersion(id={self.id!r}, name={self.name!r}, effective_version={self.effective_version}, '
-            f'anchor_tbl_id={self.anchor_tbl_id}; version={self.version})'
+            f'TableVersion(id={self.id!r}, name={self.name!r}, '
```
mkornacker: I think repr() for unversioned tables should leave out everything that only applies to versioned tables.

pixeltable/catalog/table_version.py
```diff
@@ -401,6 +407,7 @@ def create_replica(cls, md: TableVersionMd, create_store_tbl: bool = True) -> Ta
 
         assert get_runtime().in_xact
         assert md.tbl_md.is_replica
+        assert md.tbl_md.is_versioned, 'TODO: implement for unversioned tables [PXT-975]'
```
mkornacker: I don't think that makes sense. We can only have replicas of specific table versions, never of live tables.

pixeltable/catalog/table_version.py
```diff
@@ -1029,16 +1044,20 @@ def insert(
         else:
             plan = Planner.create_query_insert_plan(self, query, ignore_errors=not fail_on_exception)
 
-        # this is a base table; we generate rowids during the insert
-        def rowids() -> Iterator[int]:
-            while True:
-                rowid = self.next_row_id
-                self.next_row_id += 1
-                yield rowid
+        rowids_gen = None
```
mkornacker: Missing type decl.

pixeltable/catalog/table_version.py
```diff
@@ -1029,16 +1044,20 @@ def insert(
...
+        if self.is_versioned:
+            # this is a versioned base table; we generate rowids during the insert
```
mkornacker: We always generate rowids during insert.

pixeltable/catalog/table_version_path.py
```diff
@@ -108,6 +105,7 @@ def tbl_id(self) -> UUID:
 
     def version(self) -> int:
```
mkornacker: Maybe this should return int | None

mkornacker (changes requested):

pixeltable/store.py
sergey-mkhitaryan: could be UUID, see the discussion here https://www.notion.so/Unversioned-Tables-33854609e49080d5b0b3e2d96d9f0bfe?d=33b54609e49080dfbc2c001c77c47c2a&source=copy_link#33854609e49080d28760e63254921598

pixeltable/catalog/table_version.py
sergey-mkhitaryan: alright

pixeltable/metadata/schema.py
sergey-mkhitaryan: it's just going to always be 0 for unversioned tables. do you want me to document that specifically?

pixeltable/metadata/schema.py
sergey-mkhitaryan: I actually prefer backfilling -- the migration script will be pruned away eventually, but the code and the metadata will stay forever. So it feels like a better choice to migrate and end up with cleaner code (`is_versioned` has no default value and is located where I want to see it and not at the bottom) and more predictable metadata (`is_versioned` is either True or False, no third option to be on a lookout for).

pixeltable/catalog/table_version.py
sergey-mkhitaryan: rewrote

aaron-siegel (approved):

pixeltable/store.py
```diff
@@ -680,8 +728,8 @@ def __init__(self, tbl_version: catalog.TableVersion):
         super().__init__(tbl_version)
 
     def _create_rowid_columns(self) -> list[sql.Column]:
-        self.rowid_col = sql.Column('rowid', sql.BigInteger, nullable=False)
-        return [self.rowid_col]
+        rowid_col = sql.Column('rowid', self._rowid_col_type(), nullable=False, autoincrement=False)
```
mkornacker, line 732: Comment why we need the explicit autoincrement=False.

mkornacker (approved):

PR #1237, opened 2026-04-04
Clean up bridge.py

mkornacker:
Replaces use of internal classes and methods with that of our public API.

Also extends ColumnMetadata to match what the local UI needs.

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -786,7 +786,7 @@ class TestTable:
                 'is_snapshot': False,
                 'is_replica': False,
                 'base': 'test',
-                'iterator_call': 'DummyIterator(limit=n)',
+                'iterator_call': 'DummyIterator(n)',
                 'version': 1,
                 'schema_version': 1,
                 'comment': None,


pixeltable/catalog/table.py
```diff
@@ -126,6 +136,8 @@ def _get_metadata(self) -> TableMetadata:
                 media_validation=col.media_validation.name.lower() if col.media_validation is not None else None,  # type: ignore[typeddict-item]
                 is_computed=col.is_computed,
                 computed_with=col.value_expr.display_str(inline=False) if col.value_expr is not None else None,
+                is_builtin=(not col.calls_custom_udf) if col.value_expr is not None else None,
```
aaron-siegel, line 139: Isn't this just equivalent to `is_builtin=(not col.calls_custom_udf)` (since the `value_expr` check is done also in the property?)

pixeltable/catalog/table_metadata.py
```diff
@@ -16,11 +16,15 @@ class ColumnMetadata(TypedDict):
     is_primary_key: bool
     """`True` if this column is part of the table's primary key."""
     media_validation: Literal['on_read', 'on_write'] | None
-    """The media validation policy for this column."""
+    """The media validation policy for this column. `None` if this column does not have a media type."""
```
aaron-siegel: "`None` if the type of this column is not a media type."

pixeltable/dashboard/bridge.py
```diff
@@ -322,42 +325,12 @@ def search(query: str, limit: int = 50) -> dict[str, Any]:
...
-            name = tv.iterator_call.it.name
-            iter_cols = {c.name for c in tv.cols if c.name and tv.is_iterator_column(c)}
-            return name, iter_cols
-    except Exception:
-        pass
-    return None, set()
+# Matches the topmost function name in a display expression, e.g.
+#   'my_udf(c1.len())'  -> 'my_udf'
+#   'c1.upper()'         -> 'upper'
+#   'DummyIterator(n=3)' -> 'DummyIterator'
```
aaron-siegel: This doesn't follow the modern convention for iterator names, and anyway, why use "DummyIterator" as an example?

frame_iterator(t.video) -> frame_iterator

pixeltable/dashboard/bridge.py
```diff
@@ -322,42 +325,12 @@ def search(query: str, limit: int = 50) -> dict[str, Any]:
...
+# For bare operators like 'c1 + 1' there is no match.
+_TOPMOST_FUNC_RE = re.compile(r'(?:^|.*\.)(\w+)\(')
```
aaron-siegel: I don't understand why we would choose to use a regex over the string representation of an object, when it's something that can easily be programmatically determined. It's general poor practice, and it invites bugs. Why did you revert the change to use a programmatic check and go back to using a regex?

pixeltable/dashboard/bridge.py
```diff
@@ -371,55 +344,69 @@ def get_pipeline() -> dict[str, Any]:
...
-                cw_str = str(cw)[:200] if cw else None
-                salient_fn = col.value_expr.get_first_udf() if col.value_expr is not None else None
-                func_name = salient_fn.display_name if salient_fn is not None else None
+                value_expr = value_expr[:200] if value_expr is not None else None
+                func_type: str | None
+                if not is_computed and not is_iter_col:
+                    func_type = None
+                elif is_iter_col:
+                    func_type = 'iterator'
+                elif info['is_builtin'] is not None and not info['is_builtin']:
```
aaron-siegel: I find this very confusing. So there are 3 possible values for `is_builtin` - `None`, `True`, `False`? And `None` and `True` both map to `'builtin'`, while `False` maps to `'custom_udf'`? Please explain the logic here with a code comment

pixeltable/exprs/expr.py
```diff
@@ -390,9 +379,9 @@ def list_contains(
         expr_class: type[Expr] | None = None,
         filter: Callable[[Expr], bool] | None = None,
     ) -> bool:
-        return any(e._contains(expr_class, filter) for e in expr_list)
+        return any(e.contains_(expr_class, filter) for e in expr_list)
 
-    def _contains(self, cls: type[Expr] | None = None, filter: Callable[[Expr], bool] | None = None) -> bool:
+    def contains_(self, cls: type[Expr] | None = None, filter: Callable[[Expr], bool] | None = None) -> bool:
```
aaron-siegel, line 384: Why `contains_` and not `contains`?

pixeltable/catalog/table_metadata.py
mkornacker: Done.

pixeltable/catalog/table.py
mkornacker: This sets is_builtin to None if this isn't a computed column.

pixeltable/dashboard/bridge.py
mkornacker: Done.

pixeltable/dashboard/bridge.py
mkornacker: Because the specific operation (pick up the first call) has no meaning outside of this particular local use case, so we might as well keep the logic here. Unless we change the way we print function calls in the future (highly unlikely), I don't see a problem with this regex.

pixeltable/dashboard/bridge.py
mkornacker: The preceding branches ensure that you're looking at a udf call at this point (= it wouldn't be None here).

pixeltable/exprs/expr.py
mkornacker: Clarified in comment.

pixeltable/dashboard/bridge.py
aaron-siegel: So why include the `info['is_builtin'] is not None` clause if it's automatically True?

pixeltable/catalog/table.py
aaron-siegel: I see.

pixeltable/dashboard/bridge.py
aaron-siegel: The regex contains a bug; the intended output of `1 + my_udf(x)` is `my_udf`, but the regex as written fails to match. (Note that `1 + x.my_udf()` *does* match, an odd inconsistency.)

I continue to assert that using regexes like this, when an easy alternative is available, is a very bad idea and invites bugs. I strongly advise using a programmatic approach instead. (If you don't believe "first udf call" is a logical thing to expose, let's think about what other metadata might make more sense, e.g., a list of all udfs invoked by the expression, from which the dashboard could pluck out the first one.)

If you absolutely insist on using a regex here, please ensure that it has robust test coverage so that we catch edge cases like the above and protect them against regressions.

pixeltable/exprs/expr.py
```diff
@@ -390,11 +379,13 @@ def list_contains(
...
         """
         Returns True if any subexpr is an instance of cls and/or matches filter.
+
+        This is named 'contains_' to avoid a name conflict with the contains(s: str, ...) udf.
```
aaron-siegel, line 388: Maybe call it `contains_expr` or `contains_expr_type` then?

mkornacker: We have a list_contains() which applies contains() to a list, should we also rename that? It can also accept a filter, so contains_expr or contains_expr_type aren't a great match.

What's wrong with contains_()? Sqlalchemy does that as well (they have an and_()/or_()).

pixeltable/dashboard/bridge.py
mkornacker: Done.

pixeltable/dashboard/bridge.py
mkornacker: Done.

pixeltable/exprs/expr.py
aaron-siegel: it's fine.

PR #1241, opened 2026-04-05
Fix for https://github.com/pixeltable/pixeltable/issues/1239

mkornacker:


pixeltable/functions/huggingface.py
```diff
@@ -211,6 +211,21 @@ class DetrForObjectDetectionResponse(TypedDict):
     boxes: list[list[float]]
 
 
+def _detr_config(model_id: str, revision: str) -> Any:
```
aaron-siegel: Avoid `Any` as a return type - it's easy enough to determine the correct type. (in this case it's just DetrConfig, I believe)

aaron-siegel (approved):

pixeltable/functions/huggingface.py

PR #1244, opened 2026-04-07
PXT-1048 Expand Bedrock API to support media input and output

amithadke:
- Add invoke_model_async with async context manager for S3 polling and cleanup
- Add recursive media input conversion (_process_media_input) replacing per-model field maps
- Add response image decoding for image-generation models (Stability, Nova Canvas, Titan Image Generator)
- Add conditional_return_type for invoke_model_async (pxt.Video for Nova Reel, Luma)
- Add Cohere embed v3/v4 support in embed UDF (text and image)
- Add Cohere embed-v4 image support via data URI encoding
- Add data URI input encoding for Mistral, Gemma, NVIDIA, Moonshot, Qwen, Cohere embed-v4
- Add _apply_converse_request_conversions for automatic media-to-bytes conversion in converse
- Expand _embedding_dimensions with Cohere models

### DIFF
--- a/tests/functions/test_bedrock.py
+++ b/tests/functions/test_bedrock.py
@@ -20,7 +20,7 @@ from .tool_utils import run_tool_invocations_test
 
 
 @pytest.fixture()
-def bedrock_us_east_1(uses_db: None) -> Iterator[str]:
+def bedrock_us_east_1(uses_db: None) -> Iterator[None]:
     """Configure the Bedrock client for us-east-1. Yields the temp_location S3 URI."""
     Config.init(
         config_overrides={
@@ -29,11 +29,11 @@ def bedrock_us_east_1(uses_db: None) -> Iterator[str]:
         },
         reinit=True,
     )
-    yield 's3://pxt-test-us-east-1/bedrock_outputs/'
+    yield
 
 
 @pytest.fixture()
-def bedrock_us_west_2(uses_db: None) -> Iterator[str]:
+def bedrock_us_west_2(uses_db: None) -> Iterator[None]:
     """Configure the Bedrock client for us-west-2. Yields the temp_location S3 URI."""
     Config.init(
         config_overrides={
@@ -42,7 +42,7 @@ def bedrock_us_west_2(uses_db: None) -> Iterator[str]:
         },
         reinit=True,
     )
-    yield 's3://pxt-test-us-west-2/bedrock_outputs/'
+    yield
 
 
 def _tbl_name(model_id: str) -> str:
@@ -129,6 +129,7 @@ def _converse_image_messages(t: pxt.Table) -> list:
 
 @pytest.mark.remote_api
 @rerun(reruns=3, reruns_delay=8)
+@pytest.mark.usefixtures('bedrock_us_east_1')
 class TestBedrock:
     def test_invoke_model_twelvelabs_marengo_image(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
@@ -163,7 +164,7 @@ class TestBedrock:
         validate_update_status(t.insert({'image': p} for p in image_filepaths), expected_rows=len(image_filepaths))
         _assert_twelvelabs_embedding(t.select(t.response).collect())
 
-    def test_invoke_model_twelvelabs_marengo_audio(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_invoke_model_twelvelabs_marengo_audio(self, uses_db: None) -> None:
         # Audio auto-routes to StartAsyncInvoke.
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import invoke_model
@@ -179,7 +180,7 @@ class TestBedrock:
         validate_update_status(t.insert({'audio': p} for p in audio_filepaths), expected_rows=len(audio_filepaths))
         _assert_twelvelabs_embedding(t.select(t.response).collect())
 
-    def test_invoke_model_twelvelabs_marengo_video(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_invoke_model_twelvelabs_marengo_video(self, uses_db: None) -> None:
         # Video auto-routes to StartAsyncInvoke.
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import invoke_model
@@ -362,7 +363,7 @@ class TestBedrock:
         validate_update_status(t.insert({'video': p} for p in video_filepaths), expected_rows=len(video_filepaths))
         assert t.collect()[0]['output']['output']['message']['content'][0]['text']
 
-    def test_invoke_model_nova_reel(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_invoke_model_nova_reel(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import invoke_model
 
@@ -425,7 +426,7 @@ class TestBedrock:
             t.insert([{'image': p} for p in img_paths])
             _assert_image_similarity(t, img_paths)
 
-    def test_embed_nova_multimodal_text(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_embed_nova_multimodal_text(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import embed
 
@@ -436,7 +437,7 @@ class TestBedrock:
         t.insert(_TEXT_ROWS)
         _assert_text_similarity(t)
 
-    def test_embed_nova_multimodal_image(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_embed_nova_multimodal_image(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import embed
 
@@ -448,7 +449,7 @@ class TestBedrock:
         t.insert([{'image': p} for p in img_paths])
         _assert_image_similarity(t, img_paths)
 
-    def test_invoke_model_nova_multimodal_audio(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_invoke_model_nova_multimodal_audio(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import invoke_model
 
@@ -470,7 +471,7 @@ class TestBedrock:
         validate_update_status(t.insert({'audio': p} for p in audio_filepaths), expected_rows=len(audio_filepaths))
         assert t.select(t.response).collect()[0]['response']['embeddings']
 
-    def test_invoke_model_nova_multimodal_video(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_invoke_model_nova_multimodal_video(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import invoke_model
 
@@ -496,7 +497,7 @@ class TestBedrock:
         validate_update_status(t.insert({'video': p} for p in video_filepaths), expected_rows=len(video_filepaths))
         assert t.select(t.response).collect()[0]['response']['embeddings']
 
-    def test_converse_ai21(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_converse_ai21(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         _run_converse_text(['ai21.jamba-1-5-mini-v1:0', 'ai21.jamba-1-5-large-v1:0'])
 
@@ -510,7 +511,7 @@ class TestBedrock:
             t.insert(_TEXT_ROWS)
             _assert_text_similarity(t)
 
-    def test_embed_cohere_v4_text(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_embed_cohere_v4_text(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import embed
 
@@ -519,7 +520,7 @@ class TestBedrock:
         validate_update_status(t.insert(text='Hello, world!'), expected_rows=1)
         assert len(t.collect()[0]['embedding']) == 1536
 
-    def test_embed_cohere_v4_text_custom_dimensions(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_embed_cohere_v4_text_custom_dimensions(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import embed
 
@@ -528,7 +529,7 @@ class TestBedrock:
         validate_update_status(t.insert(text='Hello, world!'), expected_rows=1)
         assert len(t.collect()[0]['embedding']) == 512
 
-    def test_embed_cohere_v4_image(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_embed_cohere_v4_image(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import embed
 
@@ -538,7 +539,7 @@ class TestBedrock:
         t.insert([{'image': p} for p in img_paths])
         _assert_image_similarity(t, img_paths)
 
-    def test_invoke_model_cohere_v4_image(self, uses_db: None, bedrock_us_east_1: str) -> None:
+    def test_invoke_model_cohere_v4_image(self, uses_db: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import invoke_model
 
@@ -558,7 +559,7 @@ class TestBedrock:
         validate_update_status(t.insert({'image': p} for p in image_filepaths), expected_rows=len(image_filepaths))
         assert t.select(t.response).collect()[0]['response']['embeddings']['float'][0]
 
-    def test_converse_deepseek(self, uses_db: None, bedrock_us_west_2: str) -> None:
+    def test_converse_deepseek(self, uses_db: None, bedrock_us_west_2: None) -> None:
         skip_test_if_no_aws_credentials()
         _run_converse_text(
             [
@@ -705,7 +706,7 @@ class TestBedrock:
 
         _run_converse_text(['qwen.qwen3-next-80b-a3b'])
 
-    def test_invoke_model_stability_image(self, uses_db: None, bedrock_us_west_2: str) -> None:
+    def test_invoke_model_stability_image(self, uses_db: None, bedrock_us_west_2: None) -> None:
         skip_test_if_no_aws_credentials()
         from pixeltable.functions.bedrock import invoke_model
 
@@ -719,6 +720,6 @@ class TestBedrock:
         validate_update_status(t.insert(prompt='a cat sitting on a mat'), expected_rows=1)
         assert isinstance(t.collect()[0]['response']['images'][0], PIL.Image.Image)
 
-    def test_converse_writer(self, uses_db: None, bedrock_us_west_2: str) -> None:
+    def test_converse_writer(self, uses_db: None, bedrock_us_west_2: None) -> None:
         skip_test_if_no_aws_credentials()
         _run_converse_text(['us.writer.palmyra-x4-v1:0', 'us.writer.palmyra-x5-v1:0'])


pixeltable/functions/bedrock.py
```diff
@@ -58,15 +70,430 @@ def _bedrock_client() -> Any:
...
+            return PIL.Image.open(io.BytesIO(obj['Body'].read()))  # type: ignore[return-value]
+        else:
+            obj = await asyncio.to_thread(store.client().get_object, Bucket=store.bucket_name, Key=result_key)
+            return json.loads(obj['Body'].read().decode('utf-8'))
+
+
+@invoke_model_async.conditional_return_type
+def _(*, model_id: str) -> ts.ColumnType:
+    if any(model_id.startswith(p) for p in _ASYNC_VIDEO_OUTPUT_MODEL_PREFIXES):
+        return ts.VideoType()
```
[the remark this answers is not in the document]
amithadke: we dont have audio generating models 
no Bedrock models produce image output via StartAsyncInvoke. Image generation models (Stability, Nova Canvas, Titan Image Generator) all use the synchronous InvokeModel

ignoring

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +70,435 @@ def _bedrock_client() -> Any:
     return get_runtime().get_client('bedrock')
 
 
-# Default embedding dimensions for models
+def _is_media(v: Any) -> bool:
+    """Return True if v is a PIL image or a file path (str) to a media file."""
+    return isinstance(v, PIL.Image.Image) or (isinstance(v, str) and not isinstance(v, bool) and Path(v).exists())
```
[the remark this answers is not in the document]
amithadke: not using puremagic here just file extension based check like gemini.py

tests/functions/test_bedrock.py
```diff
@@ -1,69 +1,295 @@
...
+def s3_us_east1() -> Iterator[str]:
+    """Configure the Bedrock client to use us-east-1. Yields the matching S3 async output location."""
+    yield 's3://pxt-test-us-east-1/bedrock_outputs/'
+
+
+@pytest.fixture()
+def bedrock_us_east_1(uses_db: None) -> Iterator[None]:
+    """Configure the Bedrock client to use us-east-1 for region-restricted models (no async output needed)."""
+    Config.init(config_overrides={'bedrock.region_name': 'us-east-1'}, reinit=True)
+    yield
```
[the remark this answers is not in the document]
amithadke: s3_us_east1 depends on bedrock_us_east_1 now to make sure Config.init is called

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+            raise pxt.Error(f'Async invocation {invocation_id} timed out after {timeout_secs}s')
+
+    soa = ObjectPath.parse_object_storage_addr(output_location, allow_obj_name=False)
+    store = ObjectOps.get_store(soa, allow_obj_name=False)
+    if not isinstance(store, S3Store):
+        raise pxt.Error('output_location must be an s3:// URI for Bedrock async invocation output')
+
+    result_prefix = f'{soa.prefix}{invocation_id}/'
+    listed = await asyncio.to_thread(store.client().list_objects_v2, Bucket=store.bucket_name, Prefix=result_prefix)
+    keys = [obj['Key'] for obj in listed.get('Contents', [])]
```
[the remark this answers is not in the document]
amithadke: not valid, don't think any model is generating 1000s of objects on bedrock

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+    mime = mimetypes.guess_type(v)[0]
+    if not mime or not mime.startswith(('image/', 'video/', 'audio/')):
+        return False
+    try:
+        return Path(v).exists()
+    except (OSError, ValueError):
+        return False
+
+
+def _to_base64_str(media: PIL.Image.Image | str) -> str:
```
aaron-siegel: I'd suggest upgrading `utils.image.to_base64()` rather than introducing something bespoke here.

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+    if isinstance(media, PIL.Image.Image):
+        return to_base64(media)
+    with open(media, 'rb') as f:
+        return b64encode(f.read()).decode('utf-8')
+
+
+def _to_binary(media: PIL.Image.Image | str) -> bytes:
```
aaron-siegel: Doesn't this duplicate logic already present in to_base64()?

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+    """Convert PIL image or file path to raw bytes for the Converse API (expects bytes, not base64)."""
+    if isinstance(media, PIL.Image.Image):
+        buf = io.BytesIO()
+        media.save(buf, format=media.format or 'JPEG')
```
aaron-siegel: This will fail for images with an alpha channel. Those edge cases are already handled by the logic in utils.image (another good reason not to duplicate it)

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+        return buf.getvalue()
+    with open(media, 'rb') as f:
+        return f.read()
+
+
+def _to_data_uri(media: PIL.Image.Image | str) -> str:
+    if isinstance(media, PIL.Image.Image):
+        mime = PIL.Image.MIME.get(media.format, 'image/jpeg')
```
aaron-siegel: Why default to image/jpeg? Pls leave a code comment explaining why we know that's safe

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+        return f'data:{mime};base64,{_to_base64_str(media)}'
+    mime = mimetypes.guess_type(media)[0] or 'application/octet-stream'
```
aaron-siegel: Why default to application/octet-stream? Please explain in a code comment

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+    return f'data:{mime};base64,{_to_base64_str(media)}'
+
+
+#####################################
```
aaron-siegel: We don't use ASCII art anywhere else in the codebase; let's not introduce it now (or if you feel that a module is impossible to navigate without ASCII art, consider breaking up the module)

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+        'nvidia.nemotron-nano-12b',
+        'moonshot.',
+        'moonshotai.',
+        'qwen.qwen3-vl',
+        'cohere.embed-v4',
+    }
+)
+
+
+def _process_media_input(obj: Any, converter: Callable) -> Any:
```
aaron-siegel: Use a subscripted `Callable` for the type hint

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+    if isinstance(obj, list):
+        return [_process_media_input(v, converter) for v in obj]
+    if _is_media(obj):
+        return converter(obj)
+    return obj
+
+
+def _apply_invoke_model_request_conversions(body: dict, model_id: str) -> dict:
+    """Convert all PIL images and media file paths in a request body to the encoding expected by the target model."""
+    if any(model_id.startswith(p) for p in _DATA_URI_INPUT_MODEL_PREFIXES):
```
aaron-siegel: Why is `_DATA_URI_INPUT_MODEL_PREFIXES` a frozenset, if we're not doing lookups? Tuple is cleaner if we just need to iterate over it

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+        return _process_media_input(body, _to_data_uri)
+    return _process_media_input(body, _to_base64_str)
+
+
+def _apply_converse_request_conversions(messages: list[dict]) -> list[dict]:
+    # The Converse API always expects raw bytes regardless of model.
+    return _process_media_input(copy.deepcopy(messages), _to_binary)
```
aaron-siegel: Why do we need a deepcopy here? _process_media_input() doesn't seem to modify its input

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+# All other models return text, embeddings, or metadata with no media fields.
+_IMAGE_GENERATION_MODEL_PREFIXES = frozenset(
+    {'stability.', 'us.stability.', 'amazon.nova-canvas', 'amazon.titan-image-generator'}
+)
+
+# Models whose invoke_model_async output is a video file.
+_ASYNC_VIDEO_OUTPUT_MODEL_PREFIXES = frozenset({'amazon.nova-reel', 'luma.'})
+
+
+def _decode_base64_images(value: str) -> str | PIL.Image.Image:
```
aaron-siegel: Shouldn't this be `_decode_base64_image()` (singular)?

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+    'cohere.embed-v4:0': 1536,
 }
 
 
+@pxt.udf(is_deterministic=False)
+async def invoke_model(
+    body: dict,
+    *,
+    model_id: str,
+    performance_config_latency: Literal['standard', 'optimized'] | None = None,
```
aaron-siegel: Do these make sense as UDF parameters? Naively it seems they are execution/runtime parameters.

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +72,435 @@ def _bedrock_client() -> Any:
...
+
+    response = await asyncio.to_thread(_bedrock_client().invoke_model, **kwargs)
+    result = json.loads(response['body'].read())
+    return _decode_invoke_model_response(result, model_id)
+
+
+@asynccontextmanager
+async def _bedrock_async_invocation(
+    model_id: str, body: dict, output_location: str, poll_interval_secs: float, timeout_secs: float
+) -> AsyncIterator[tuple['S3Store', StorageObjectAddress, str]]:
```
aaron-siegel: It's awkward to be returning `soa` and `store` as context manager artifacts when they're not actually managed by the context manager. Consider returning just `result_key` here.

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+        job: dict[str, Any] = await asyncio.to_thread(_bedrock_client().get_async_invoke, invocationArn=invocation_arn)
+        status: str = job['status']
+        if status == 'Completed':
+            break
+        if status == 'Failed':
+            raise pxt.Error(f'Async invocation {invocation_id} failed: {job.get("failureMessage", "unknown error")}')
+
+    soa = ObjectPath.parse_object_storage_addr(output_location, allow_obj_name=False)
+    store = ObjectOps.get_store(soa, allow_obj_name=False)
+    if not isinstance(store, S3Store):
```
aaron-siegel: Do this validation up front

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+    'cohere.embed-english-v3': 1024,
+    'cohere.embed-multilingual-v3': 1024,
+    'cohere.embed-v4:0': 1536,
 }
 
 
+@asynccontextmanager
+async def _bedrock_async_invocation(
+    model_id: str, body: dict, output_location: str, poll_interval_secs: float
+) -> AsyncIterator[tuple['S3Store', StorageObjectAddress, str]]:
```
aaron-siegel: It's awkward to be returning `soa` and `store` as context manager artifacts when they're not actually managed by the context manager. Consider returning just `result_key` here.

pixeltable/functions/bedrock.py
```diff
@@ -146,85 +601,34 @@ async def converse(
...
     __Requirements:__
 
     - `pip install boto3`
 
     Args:
         text: Input text to embed.
-        model_id: The embedding model identifier.
-        dimensions: Output embedding dimensions. For Titan V2: 256, 512, 1024. For Nova: 256, 512, 1024, 3072.
+        model_id: The embedding model identifier. Supported models:
+            - `amazon.titan-embed-text-v1`
```
aaron-siegel, line 604: Bulleted lists in docstrings must be preceded by a blank line or they will render incorrectly in the docs.

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+
+    For additional details, see:
+    <https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model.html>
+
+    PIL images and media file paths anywhere in the request body are converted automatically
+    to the encoding expected by the target model. For image-generation models, base64-encoded
+    images in the response are automatically decoded into `PIL.Image` objects.
+
+    For models that require async invocation, `bedrock.temp_location` must be configured
+    (set environment variable `BEDROCK_TEMP_LOCATION` or add `temp_location` to the `[bedrock]` section of config.toml).
```
aaron-siegel: s/config.toml/your Pixeltable configuration file/

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+
+    Examples:
+        Invoke Amazon Titan text embeddings:
+
+        >>> body = {'inputText': tbl.text, 'dimensions': 512, 'normalize': True}
+        >>> tbl.add_computed_column(
+        ...     response=invoke_model(body, model_id='amazon.titan-embed-text-v2:0')
+        ... )
+
+        Invoke TwelveLabs Marengo with an image column:
```
aaron-siegel: Clarify?

Invoke TwelveLabs Marengo with an image column (note that the image can be included directly in the invoke body, and will be automatically base64-encoded by Pixeltable):

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+
+    async with _bedrock_async_invocation(model_id, body, output_location, poll_interval_secs) as (
+        store,
+        soa,
+        result_key,
+    ):
+        result_uri = f'{soa.prefix_free_uri}{result_key}'
+        content_type = mimetypes.guess_type(result_key)[0] or 'application/json'
+
+        if content_type.startswith('video/') or content_type.startswith('audio/'):
```
aaron-siegel: It's dangerous to use different logic for determining the return value structure from what's used in `conditional_return_type`. This could lead to very confusing/unexpected behavior in the event one of our assumptions breaks down.

I'd strongly advise aligning the logic between the UDF body and `conditional_return_type`.

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+
+@pxt.udf(is_deterministic=False)
+async def invoke_model(
+    body: dict,
+    *,
+    model_id: str,
+    poll_interval_secs: float = 5.0,
+    performance_config_latency: Literal['standard', 'optimized'] | None = None,
+    service_tier: Literal['priority', 'default', 'flex', 'reserved'] | None = None,
+) -> dict:
```
aaron-siegel: Use pxt.Json as the return type here (it's still a little bit of a lie since technically `pxt.Video` isn't (yet) part of `pxt.Json`... but it's less of a lie than `dict`)

pixeltable/functions/bedrock.py
```diff
@@ -277,10 +687,43 @@ async def embed(text: str, *, model_id: str, dimensions: int | None = None) -> p
 
 @embed.overload
 async def _(image: PIL.Image.Image, *, model_id: str, dimensions: int | None = None) -> pxt.Array[(None,), np.float32]:
+    """
+    Generate image embeddings using Amazon Titan, Amazon Nova, or Cohere Embed v4.
```
aaron-siegel: Note that overloaded docstrings won't currently be rendered in the docs - only the primary docstring will be rendered.

tests/functions/test_bedrock.py
```diff
@@ -1,69 +1,300 @@
...
+    skip_test_if_not_installed('boto3')
+    skip_test_if_no_aws_credentials()
+
+
+def _tbl_name(model_id: str) -> str:
+    """Sanitize a model ID into a valid Pixeltable table name."""
+    return 'tbl_' + model_id.replace('.', '_').replace(':', '_').replace('-', '_')
+
+
+# ---------------------------------------------------------------------------
```
aaron-siegel: Remove ASCII art

tests/functions/test_bedrock.py
```diff
@@ -1,69 +1,300 @@
...
+        config_overrides={
+            'bedrock.region_name': 'us-west-2',
+            'bedrock.temp_location': 's3://pxt-test-us-west-2/bedrock_outputs/',
+        },
+        reinit=True,
+    )
+    yield 's3://pxt-test-us-west-2/bedrock_outputs/'
+
+
+def _skip_no_aws() -> None:
```
aaron-siegel: Rather than introduce this, we might as well just incorporate skip_test_if_not_installed('boto3') into skip_test_if_no_aws_credentials(). (I can't envision any test scenario where we'd want AWS credentials, but not care if boto3 is installed)

docs/public_api.opml
```diff
@@ -283,6 +283,7 @@
         <outline text="module|pixeltable.functions.bedrock">
           <outline text="udf|pixeltable.functions.bedrock.converse" />
           <outline text="func|pixeltable.functions.bedrock.invoke_tools" />
+          <outline text="udf|pixeltable.functions.bedrock.invoke_model" />
```
aaron-siegel, line 330: Missing `embed`?

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
     return get_runtime().get_client('bedrock')
 
 
-# Default embedding dimensions for models
+def _get_temp_location() -> str:
+    """Get the bedrock.temp_location config value, raising a clear error if not set."""
+    location = Config.get().get_value('temp_location', str, section='bedrock')
```
aaron-siegel: Use get_string_value()

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+    if not location:
+        raise pxt.Error(
+            'bedrock.temp_location must be configured for async model invocation. '
+            'Set the environment variable BEDROCK_TEMP_LOCATION or add temp_location to the '
+            '[bedrock] section of config.toml.'
```
aaron-siegel: s/config.toml/your Pixeltable configuration file/

Enclose symbols in backticks in error messages.

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+        )
+    return location
+
+
+def _requires_async(body: dict, model_id: str) -> bool:
+    """Return True if this model and input combination requires StartAsyncInvoke."""
+    if model_id.startswith(tuple(_ASYNC_VIDEO_OUTPUT_MODEL_PREFIXES)):
```
aaron-siegel: See comment below - why are these frozensets to begin with?

pixeltable/functions/bedrock.py
aaron-siegel: Also, can we pass `startswith` a tuple as above?

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+        return True
+    if model_id.startswith(tuple(_ASYNC_EMBEDDING_MODEL_PREFIXES)):
+        input_type = body.get('inputType', '')
```
aaron-siegel: Is the explicit `''` default necessary here?

pixeltable/functions/bedrock.py
amithadke: added to_bytes as primitive and to_base64 use it


def to_base64(image: PIL.Image.Image | str, format: str | None = None) -> str:
    return base64.b64encode(to_bytes(image, format)).decode('utf-8')


def to_bytes(image: PIL.Image.Image | str, format: str | None = None) -> bytes:
    if isinstance(image, str):
        with open(image, 'rb') as f:
            return f.read()
    buffer = BytesIO()
    image.save(buffer, format=format or default_format(image))
    return buffer.getvalue()

pixeltable/functions/bedrock.py
amithadke: using pil image's default format now

pixeltable/functions/bedrock.py
amithadke: going back to image as data uris are explictly getting called for image using image/jpeg now 
added comment

pixeltable/functions/bedrock.py
amithadke: yes i started as if i will do a direct lookup soon realized that model ids need a prefix search 
all of them are tuple and just use model_id.startswith(tuple)

pixeltable/functions/bedrock.py
amithadke: in this case likely not as only amazon nova models gives these options.
but we should consider runtime configs per computed column than across the entire pixeltable instance.

pixeltable/functions/bedrock.py
amithadke: done

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +79,429 @@ def _bedrock_client() -> Any:
...
+def _apply_invoke_model_request_conversions(body: dict, model_id: str) -> dict:
+    """Convert all PIL images and media file paths in a request body to the encoding expected by the target model."""
+    if any(model_id.startswith(p) for p in _DATA_URI_INPUT_MODEL_PREFIXES):
+        return _process_media_input(body, _to_data_uri)
+    return _process_media_input(body, _to_base64_str)
+
+
+def _apply_converse_request_conversions(messages: list[dict]) -> list[dict]:
+    # The Converse API always expects raw bytes regardless of model.
+    return _process_media_input(copy.deepcopy(messages), _to_binary)
```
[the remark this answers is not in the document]
amithadke: walking input is done in purpose w/o going into specific structure of api or model
ignoring this - we will improve this once we have pixeltable type info in input dict

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +102,361 @@ def _bedrock_client() -> Any:
...
+
+    result_key = next((k for k in keys if not k.endswith('manifest.json')), None)
+    if result_key is None:
+        raise pxt.Error(f'No output file (only manifest) found for job {invocation_id}')
+
+    try:
+        yield result_key
+    finally:
+        for key in keys:
+            await asyncio.to_thread(store.client().delete_object, Bucket=store.bucket_name, Key=key)
```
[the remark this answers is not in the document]
amithadke: done dont need pagination here in cleanup

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +102,369 @@ def _bedrock_client() -> Any:
...
+def _apply_invoke_model_request_conversions(body: dict, model_id: str) -> dict:
+    """Convert all PIL images and media file paths in a request body to the encoding expected by the target model."""
+    if model_id.startswith(_DATA_URI_INPUT_MODEL_PREFIXES):
+        return _process_media_input(body, _to_data_uri)
+    return _process_media_input(body, to_base64)
+
+
+def _apply_converse_request_conversions(messages: list[dict]) -> list[dict]:
+    # The Converse API always expects raw bytes regardless of model.
+    return _process_media_input(messages, to_bytes)
```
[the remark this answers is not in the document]
amithadke: doing customized walk for converse as media input specification is same across all models

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +102,385 @@ def _bedrock_client() -> Any:
...
+        return False
+
+
+def _to_data_uri(media: PIL.Image.Image | str) -> str:
+    if isinstance(media, PIL.Image.Image):
+        fmt = default_format(media)
+        mime = f'image/{fmt}'
+        return f'data:{mime};base64,{to_base64(media)}'
+    mime = mimetypes.guess_type(media)[0] or 'image/jpeg'
+    # All models using data URIs are vision models; default to image/jpeg for unknown extensions.
```
aaron-siegel, line 148: Is that safe to do?

pixeltable/functions/bedrock.py
```diff
@@ -58,15 +102,385 @@ def _bedrock_client() -> Any:
...
+    """Convert all PIL images and media file paths in a request body to the encoding expected by the target model."""
+    if model_id.startswith(_DATA_URI_INPUT_MODEL_PREFIXES):
+        return _process_media_input(body, _to_data_uri)
+    return _process_media_input(body, to_base64)
+
+
+def _apply_converse_request_conversions(messages: list[dict]) -> list[dict]:
+    """Schema-aware walk for the Bedrock Converse API."""
+
+    def walk(obj: Any) -> Any:
```
aaron-siegel: No need for a nested `def` here.

tests/functions/test_bedrock.py
```diff
@@ -75,63 +281,444 @@ def make_table(tools: pxt.Tools, tool_choice: pxt.ToolChoice) -> pxt.Table:
...
+    def test_embed_cohere_v4_text(self, uses_db: None) -> None:
+        skip_test_if_no_aws_credentials()
+        from pixeltable.functions.bedrock import embed
+
+        t = pxt.create_table('tbl', {'text': pxt.String})
+        t.add_computed_column(embedding=embed(t.text, model_id='cohere.embed-v4:0'))
+        validate_update_status(t.insert(text='Hello, world!'), expected_rows=1)
+        assert len(t.collect()[0]['embedding']) == 1536
+
+    def test_embed_cohere_v4_text_custom_dimensions(self, uses_db: None) -> None:
```
aaron-siegel: Some of these tests seem nearly identical. Why have separate tests here with 90%+ code duplication versus a single test with multiple cases (or parameterization)?

I think it's worth making a few passes through the tests to ensure we're minimizing code duplication when practical.

amithadke: I chose it this way so we can track results on coverage and can disable failing tests in future. 

another approach was to group all invoke_models or converse api calls into single tests with model id, input dicts as parameters but models responses have subtle differences (hence different validators in tests).

pixeltable/functions/bedrock.py
amithadke: in this case we don't have type info to create data uri so if its not image call will fail. We can fail here with our execution error/exception

PR #1247, opened 2026-04-07
[PXT-1131] Update Catalog table locking protocol

sergey-mkhitaryan:
1. `begin_xact()` takes a list of tables to read and write as an argument. Read tables' metadata is cached, and write tables a locked.
2. New table metadata loading is disallowed outside of the setup in `begin_xact()` and `retry_loop()`.
3. `Catalog._acquire_tbl_lock()` is split into multiple functions.
4. `Catalog.get_tbl_version()` spins up its own retry loop when run outside of a transaction.
5. Added some missing test coverage.


pixeltable/catalog/catalog.py
```diff
@@ -273,13 +290,33 @@ def mark_modified_tvs(self, *handle: TableVersionHandle) -> None:
         assert get_runtime().in_xact
         self._modified_tvs.update(handle)
 
+    @contextmanager
+    def allow_tbl_md_read(self) -> Iterator[None]:
+        """Context manager that allows reading new table metadata.
+
+        With a few rare exceptions, this should not be used outside of Catalog. Instead, the table locks and caches
```
mkornacker: "with a few rare exceptions"? It doesn't sound like this should ever be called outside of Catalog.

pixeltable/catalog/catalog.py
```diff
@@ -290,37 +327,43 @@ def begin_xact(
...
             yield get_runtime().conn
             return
 
+        assert for_write or not (tvp_write_targets or tbl_id_write_targets), (
+            'for_write must be True when write targets are specified'
+        )
         num_retries = 0
         pending_ops_tbl_id: UUID | None = None
         has_exc = False  # True if we exited the 'with ...begin_xact()' block with an exception
         while True:
```
mkornacker, line 358: Why do we need allow_tbl_md_read() as a contextmgr? Why not just set the flag to True up here and then to False right before the yield? Why does it need to be reentrant?

pixeltable/catalog/catalog.py
```diff
@@ -1287,18 +1357,19 @@ def create_fn() -> UUID:
...
             # invalidate base's TableVersion instance, so that it gets reloaded with the new mutable view
             self._clear_tv_cache(base.tbl_version.key)
-            # base_tv = self.get_tbl_version(base.tbl_id, base.tbl_version.effective_version, validate_initialized=True)
-            # view_handle = TableVersionHandle(view_id, effective_version=None)
-            # base_tv.mutable_views.add(view_handle)
 
         self._roll_forward()
-        with self.begin_xact(tbl_id=view_id, for_write=True):
+
+        @retry_loop(for_write=True, tbl_id_write_targets=[view_id])
```
mkornacker: Same logic for create_table()?

pixeltable/catalog/catalog.py
```diff
@@ -1817,94 +1878,114 @@ def get_view_ids(self, tbl_id: UUID, for_update: bool = False) -> list[UUID]:
...
-                assert key.effective_version is None
-                # _logger.debug(f'validating metadata for table {tbl_id}:{tv.version} ({id(tv):x})')
-                where_clause: sql.ColumnElement[bool]
+        conn = get_runtime().conn
+        assert conn is not None
+        tv = self._tbl_versions.get(key)
+        if (
+            # (tv is None or not tv.is_validated)
+            tv is None
+            # and check_pending_ops
```
mkornacker: ?

pixeltable/share/packager.py
```diff
@@ -438,7 +438,10 @@ def restore(self, bundle_path: Path, pxt_uri: str | None = None, explicit_versio
 
         cat = get_runtime().catalog
 
-        with cat.begin_xact(for_write=True):
+        # Allow table metadata reads anywhere in this transaction. We cannot lock the tables for write at the start of
+        # the transaction because those tables don't yet exist. And there shouldn't be any pending table ops on them
+        # for the same reason.
+        with cat.allow_tbl_md_read(), cat.begin_xact(for_write=True):
```
mkornacker: I don't understand this, why do you need to load metadata after lock acquisition? And if so, why not in a retry loop?

pixeltable/catalog/catalog.py
```diff
@@ -531,39 +602,21 @@ def _acquire_path_locks(
         read_handles = path_handles[:0:-1] if for_write else path_handles[::-1]
         for handle in read_handles:
             # update cache
-            _ = self.get_tbl_version(handle.key, validate_initialized=True)
+            _ = self._get_tbl_version(handle.key, validate_initialized=True)
         if not for_write:
             return set()  # nothing to lock
```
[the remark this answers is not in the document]
sergey-mkhitaryan: wrote a test to confirm that that is not really a problem. for write targets, we don't care if the metadata is cached right away or not, as long as those targets a transaction-locked.

```
    def test_begin_write_txn(self, uses_db: None) -> None:
        t = pxt.create_table('tbl_for_replica', {'a': pxt.Int})
        tbl_id: UUID = t._tbl_version.get().id

        for for_write in [True, False]:
            get_runtime().catalog._tbl_versions.clear()
            with get_runtime().catalog.begin_xact(for_write=for_write, tbl_id_read_targets=[tbl_id]):
                assert TableVersionKey(tbl_id, None, None) in get_runtime().catalog._tbl_versions

            get_runtime().catalog._tbl_versions.clear()
            with get_runtime().catalog.begin_xact(for_write=for_write, tvp_read_targets=[t._tbl_version_path]):
                assert TableVersionKey(tbl_id, None, None) in get_runtime().catalog._tbl_versions

        get_runtime().catalog._tbl_versions.clear()
        with get_runtime().catalog.begin_xact(for_write=True, tbl_id_write_targets=[tbl_id]):
            assert tbl_id in get_runtime().catalog._x_locked_tbl_ids

        get_runtime().catalog._tbl_versions.clear()
        with get_runtime().catalog.begin_xact(for_write=True, tvp_write_targets=[t._tbl_version_path]):
            assert tbl_id in get_runtime().catalog._x_locked_tbl_ids
```

pixeltable/catalog/table.py
```diff
@@ -612,29 +612,36 @@ def add_columns(
...
+            new_cols = [Column.create(name, spec) for name, spec in schema_copy.items()]
             for new_col in new_cols:
                 self._verify_column(new_col)
+            return new_cols
+
+        new_cols = do_add_columns()
+        if new_cols is None:
+            return UpdateStatus()
 
         assert self._tbl_version is not None
```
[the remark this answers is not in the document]
sergey-mkhitaryan: wrong, `Table.add_columns()` is done in 2 separate transactions on `main` as well.

pixeltable/catalog/catalog.py
```diff
@@ -76,7 +77,13 @@ def _unpack_row(row: sql.engine.Row | None, entities: list[type[sql.orm.decl_api
 
 
 def retry_loop(
-    *, tbl: TableVersionPath | None = None, for_write: bool, lock_mutable_tree: bool = False
+    *,
+    for_write: bool = False,
+    tvp_read_targets: Collection[TableVersionPath] | None = None,
```
mkornacker: For brevity, and with no loss of readability: read_tvps/read_tbl_ids/write_tvp/write_tbl_ids.

tests/utils.py
```diff
@@ -799,7 +800,13 @@ def run_reload_test(self, clear: bool = True) -> None:
         assert len(self.query_info) > 0, 'No queries in ReloadTester!'
         # enumerate(): the list index is useful for debugging
         for _idx, (query_dict, result_set) in enumerate(self.query_info):
-            query = pxt.Query.from_dict(query_dict)
+
+            @retry_loop()
+            def query_from_dict() -> pxt.Query:
+                return pxt.Query.from_dict(query_dict)
```
mkornacker, line 806: Put the retry loop directly into from_dict()?

pixeltable/catalog/catalog.py
```diff
@@ -1265,8 +1348,9 @@ def create_fn() -> UUID:
             if not is_snapshot and base.is_mutable():
                 # this is a mutable view of a mutable base; X-lock the base and advance its view_sn before adding
                 # the view
-                self._acquire_tbl_lock(tbl_id=base.tbl_id, for_write=True)
-                base_tv = self.get_tbl_version(TableVersionKey(base.tbl_id, None, None), validate_initialized=True)
+                # TODO: assert that the lock was acquired?
```
mkornacker: Isn't the call to _acquire_write_lock() enough? It would be confusing if that's not part of the function's contract.

pixeltable/catalog/catalog.py
```diff
@@ -1841,97 +1913,110 @@ def get_view_ids(self, tbl_id: UUID, for_update: bool = False) -> list[UUID]:
...
-            if tv is None:
-                tv = self._load_tbl_version(key, check_pending_ops=check_pending_ops)
-            elif not tv.is_validated:
-                # only live instances are invalidated
-                assert key.effective_version is None
-                where_clause: sql.ColumnElement[bool]
+        conn = get_runtime().conn
+        assert conn is not None
+        tv = self._tbl_versions.get(key)
+        if tv is None and not self._tbl_md_read_allowed and not self._in_retry_loop:
```
mkornacker: The retry loop should also set tbl_md_read_allowed. It would be better not to have to check for that separately.

pixeltable/catalog/catalog.py
```diff
@@ -276,13 +296,26 @@ def mark_modified_tvs(self, *handle: TableVersionHandle) -> None:
...
+        finally:
+            self._tbl_md_read_allowed = False
+
     @contextmanager
     def begin_xact(
         self,
         *,
-        tbl: TableVersionPath | None = None,
-        tbl_id: UUID | None = None,
         for_write: bool = False,
```
mkornacker, line 319: I don't think for_write is useful as a parameter anymore. It should be a local variable
for_write = len(write_tvps) > 0 | len(write_tbl_ids) > 0.

mkornacker (approved):

tests/utils.py
sergey-mkhitaryan: can't do that because there's a call path between `Catalog._finalize_pending_ops` and `Query.from_dict`

pixeltable/catalog/catalog.py
sergey-mkhitaryan: `_acquire_write_lock`'s contract is "acquire write lock if this table is mutable, and check pending table ops". `_acquire_write_lock` itself can't assert that the lock is always acquired. that would trip numerous operations, for example a snapshot creation. that said, likely there are opportunities for refactoring for the better around lock acquisition, and I'm open for suggestions.

pixeltable/catalog/catalog.py
sergey-mkhitaryan: done. `_allow_tbl_md_read` had to become reentrant for that to work.

pixeltable/catalog/catalog.py
sergey-mkhitaryan: we can't fully drop it because ops like create dir (write transaction but no tables to lock) won't work. what we can do is make it optional, so that the semantic is "the transaction is for write if either for_write==True or a write table target is provided", but I like the current state more. let me know.

PR #1258, opened 2026-04-09
PXT-1102 Incomplete plan generation when calling pxt.insert() on a Query

amithadke:
Fixes insert from query to correctly evaluate destination computed columns and index value columns, using a substitution dict that rewrites computed column expressions in terms of source query expressions. 

Apply substitution for index value column expr, missing computed columns, any unstored columns and  missing stored columns (mapped to None - in next version this will be default value expr for column) 

The resolved expressions are added to the query via a new add_columns() API and all destination columns are registered with the row builder after planning. Also fixes create_table(source=t) to preserve column names when building a query from a table


pierrebrunelle:
@claude review

pixeltable/_query.py
```diff
@@ -1488,3 +1488,20 @@ def to_pytorch_dataset(self, image_format: str = 'pt') -> 'torch.utils.data.Iter
...
+                If name is None, the expression is appended without adding it to the schema.
+
+        Returns:
+            a new Query with additional expressions appended to the select list.
+        """
+        q = copy.copy(self)
+        q._select_list_exprs = self._select_list_exprs + [expr for expr, _ in additional_columns]
+        q.select_list = (self.select_list or []) + additional_columns
+        # only add named columns to schema
+        q._schema = {**self._schema, **{name: expr.col_type for expr, name in additional_columns if name is not None}}
```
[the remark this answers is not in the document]
amithadke: going back to deep copy no more shallow copy. 
It does change how we detect which columns to add to table in create query insert plan specially for index value columns (using expr.id that should stay the same after Query.__init__ and select list expr mutations)

pixeltable/_query.py
```diff
@@ -1488,3 +1488,25 @@ def to_pytorch_dataset(self, image_format: str = 'pt') -> 'torch.utils.data.Iter
...
+        return Query(
+            from_clause=self._from_clause,
+            select_list=(self.select_list or []) + additional_columns,
+            where_clause=self.where_clause,
+            group_by_clause=self.group_by_clause,
+            grouping_tbl=self.grouping_tbl,
+            order_by_clause=self.order_by_clause,
+            limit=self.limit_val,
+            offset=self.offset_val,
+            sample_clause=self.sample_clause,
```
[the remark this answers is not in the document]
amithadke: not sure does init calls _normalize_select_list

pixeltable/plan.py
```diff
@@ -405,14 +405,84 @@ def create_query_insert_plan(
...
+            expr = select_exprs_by_name.get(col_name)
+            if expr is None:
+                continue
+            plan.row_builder.add_table_column(dst_col, expr.slot_idx)
+            needs_cell_materialization = needs_cell_materialization or dst_col.col_type.supports_file_offloading()
+
+        # register missing computed cols (including index val cols) via unique_exprs lookup by expr.id
+        for col, expr in zip(missing_computed_cols, missing_col_exprs):
+            matched = plan.row_builder.unique_exprs[expr]
+            assert matched is not None, f'missing expr not found in row builder {expr} for column {col.name}'
```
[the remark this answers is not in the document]
amithadke: this is an internal bug not sure if user can do anything with such error message.

pixeltable/plan.py
```diff
@@ -405,14 +405,96 @@ def create_query_insert_plan(
         cls, tbl: catalog.TableVersion, query: 'pxt.Query', ignore_errors: bool
     ) -> exec.ExecNode:
         assert not tbl.is_view
-        plan = query._create_query_plan()  # ExecNode constructed by the Query
 
-        # Modify the plan RowBuilder to register the output columns
+        query_col_names = set(query.schema.keys())
```
sergey-mkhitaryan, line 409: The amount of code it took to do this is surprising, how does `create_insert_plan` achieve the same thing with much fewer lines?
```
        computed_exprs = row_builder.output_exprs - row_builder.input_exprs
        if len(computed_exprs) > 0:
            # add an ExprEvalNode when there are exprs to compute
            plan = exec.ExprEvalNode(
                row_builder, computed_exprs, plan.output_exprs, input=plan, maintain_input_order=False
            )
```
is there any chance the same code can be used for query insert?

pixeltable/_query.py
```diff
@@ -1488,3 +1488,41 @@ def to_pytorch_dataset(self, image_format: str = 'pt') -> 'torch.utils.data.Iter
                 export_parquet(self, dest_path, inline_images=True, _write_md=True)
 
         return PixeltablePytorchDataset(path=dest_path, image_format=image_format)
+
+    def add_columns(self, additional_columns: list[tuple[exprs.Expr, str | None]]) -> 'Query':
```
sergey-mkhitaryan, line 1492: nit: "additional_columns" can be just "columns", the function name already indicates an "addition".

pixeltable/plan.py
amithadke: how? row builder needs to created with correct expressions with substitutions? 
using ExprEvalNode was earlier approach but we wanted a public query api to expand select list with set of expressions (mostly working on source query columns or in future default values)

amithadke: - removed bunch of custom code for resolving and substituting missing columns , we will still use add_column 
- also combined 2 loops into one for plan.row_builder.add_table_column
add_columns -> Query.__init__ will do deep copy of all expr (which will get mutated in place by plan.create_query_plan) so only way to match index value columns is using names or expr.id

pixeltable/plan.py
```diff
@@ -425,14 +425,61 @@ def create_query_insert_plan(
...
+
+        # map each destination column covered by the query to its corresponding source expression
```
mkornacker: Don't use destination, that has as different meaning. They're called output columns.

pixeltable/plan.py
```diff
@@ -425,14 +425,61 @@ def create_query_insert_plan(
...
+            if not col.is_stored or col.name is None or col.name in query_col_names or col.is_computed:
+                continue
+            if not col.col_type.nullable or col.is_pk:
+                raise excs.RequestError(
+                    excs.ErrorCode.MISSING_REQUIRED,
+                    f'Column {col.name!r} is required but not present in the source query',
+                )
+            substitution[exprs.ColumnRef(col)] = exprs.Literal(None, col_type=col.col_type)
+
+        # stored computed columns in destination not covered by the query
```
mkornacker: Not covered by the query? The query cannot provide values for computed columns.

pixeltable/plan.py
```diff
@@ -425,14 +425,61 @@ def create_query_insert_plan(
...
+        missing_computed_cols = [
+            col
+            for col in tbl.cols_by_id.values()
+            if col.is_stored and col.name not in query_col_names and col.is_computed
```
mkornacker: This simply needs to be all computed columns.

pixeltable/plan.py
```diff
@@ -425,14 +425,61 @@ def create_query_insert_plan(
...
+        plan = augmented_query._create_query_plan()
         needs_cell_materialization = False
-        for col_name, expr in zip(query.schema.keys(), query._select_list_exprs):
-            assert col_name in tbl.cols_by_name
-            col = tbl.cols_by_name[col_name]
-            plan.row_builder.add_table_column(col, expr.slot_idx)
+
+        # register all destination columns via unique_exprs lookup by expr.id
+        # add_columns deep copies exprs but expr.id remains same as before
+        all_dst_cols = [
```
mkornacker: Why not just take all select list items of augmented_query?

pixeltable/plan.py
```diff
@@ -425,14 +425,61 @@ def create_query_insert_plan(
...
+            (tbl.cols_by_name[name], expr)
+            for name, expr in zip(query.schema.keys(), query._select_list_exprs)
+            if name in tbl.cols_by_name
+        ] + list(zip(missing_computed_cols, missing_col_exprs))
+        for col, expr in all_dst_cols:
+            matched = plan.row_builder.unique_exprs[expr]
```
mkornacker: Unnecessary. RowBuilder.set_slot_idxs() will fix up expr lists.

tests/test_table.py
```diff
@@ -3910,3 +3910,58 @@ def test_column_comment(self, uses_db: None, do_reload_catalog: bool) -> None:
...
+        dst.add_computed_column(unstored_col=dst.c2 + 100, stored=False)
+        dst.add_computed_column(c6=dst.unstored_col * 2)
+        # c3 is missing from source so will always be none
+        dst.add_computed_column(unstored_from_missing=dst.c3 + '_suffix', stored=False)
+        dst.add_computed_column(c7=dst.unstored_from_missing + '_more')
+
+        dst.insert(src.select(src.c1, src.c2))
+
+        result = dst.order_by(dst.c2).collect()
+        assert len(result) == 2
```
mkornacker, line 3934: This can be expressed more concisely.

tests/test_table.py
```diff
@@ -3910,3 +3910,58 @@ def test_column_comment(self, uses_db: None, do_reload_catalog: bool) -> None:
         # check that raw object JSON comments are rejected for columns
         with pxt_raises(pxt.ErrorCode.TYPE_MISMATCH, match="'comment' must be a string"):
             pxt.create_table('tbl_invalid', {'c': {'type': pxt.Int, 'comment': {'comment': 'This is a test column.'}}})  # type: ignore[dict-item]
+
+    def test_insert_query_computed_cols(self, uses_db: None, e5_embed: pxt.Function) -> None:
```
mkornacker, line 3914: Missing: aggregate queries, both grouping and non-grouping.

tests/test_query.py
```diff
@@ -1049,3 +1049,24 @@ def select_inexpensive() -> None:
             assert len(res) == row_count
 
         benchmark(select_inexpensive)
+
+    def test_add_columns(self, test_tbl: pxt.Table, reload_tester: ReloadTester) -> None:
+        t = test_tbl
+        # add_columns appends named expressions to an existing select list
+        q = t.select(t.c1, t.c2).add_columns([(t.c2 + 1, 'c2_plus_1'), ((t.c2 + 1) * 2, 'c2_times_2')])
+        result = reload_tester.run_query(q)
+        assert 'c1' in result.schema
```
mkornacker: This can be expressed more concisely.

tests/test_query.py
```diff
@@ -1049,3 +1049,24 @@ def select_inexpensive() -> None:
...
+        assert 'c2' in result.schema
+        assert 'c2_plus_1' in result.schema
+        assert 'c2_times_2' in result.schema
+        for row in result:
+            assert row['c2_plus_1'] == row['c2'] + 1
```
mkornacker: assert all(...) instead of the loop

tests/test_query.py
```diff
@@ -1049,3 +1049,24 @@ def select_inexpensive() -> None:
...
+            assert row['c2_times_2'] == (row['c2'] + 1) * 2
+        reload_tester.run_reload_test()
+
+        # also works on select(*) — no explicit select list
+        q2 = t.select().add_columns([(t.c2 + 1, 'c2_plus_1')])
```
mkornacker: omit the select()

mkornacker (changes requested):

tests/test_query.py
amithadke: so t.add_columns? won't that mean something else?

pixeltable/_query.py
```diff
@@ -373,15 +379,23 @@ def __init__(
...
+        If `additional_select_list` is provided, it is appended after the base select list.
+
         Returns:
             a pair composed of the list of expressions and the list of corresponding names
         """
         if select_list is None:
             select_list = [(exprs.ColumnRef(col), None) for tbl in tbls for col in tbl.columns()]
+        if additional_select_list:
+            # rebind to a new list so the caller's select_list is not mutated in place
+            select_list = select_list + additional_select_list  # noqa: PLR6104
```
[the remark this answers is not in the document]
amithadke: we will process
base select lists, all named expressions from additional_select_list first then assign column names to all unnamed expressions from additional_select_list

pixeltable/_query.py
```diff
@@ -1728,3 +1765,43 @@ def to_pytorch_dataset(self, image_format: str = 'pt') -> 'torch.utils.data.Iter
...
+                raise excs.RequestError(
+                    excs.ErrorCode.INVALID_COLUMN_NAME, f'add_columns(): {name!r} is not a valid column name.'
+                )
+            if name in self._schema:
+                raise excs.AlreadyExistsError(
+                    excs.ErrorCode.COLUMN_ALREADY_EXISTS,
+                    f'add_columns(): column {name!r} already exists in the query. '
+                    f'Existing columns are: {list(self._schema.keys())}.',
+                )
+        new_items = [(expr, None) for expr in unnamed_columns] + [(expr, name) for name, expr in named_columns.items()]
```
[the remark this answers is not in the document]
amithadke: if self.select_list is not None:
            raise excs.RequestError(excs.ErrorCode.INVALID_STATE, 'Select list already specified')
            
            using self.select may throw above exception , checking if we can do validation better

PR #1278, opened 2026-04-16
Proposal for new exception hierarchy.

mkornacker:


pixeltable/exceptions2.py
```diff
@@ -0,0 +1,225 @@
...
+        error_code: ErrorCode,
+        message: str = '',
+        cause: BaseException | None = None,
+        retry_after: float | None = None,
+    ) -> None:
+        super().__init__(message)
+        self.error_code = error_code
+        self.retry_after = retry_after
+        if cause is not None:
+            self.__cause__ = cause
```
aaron-siegel, line 108: Can this not be passed to `super().__init__()`?

pixeltable/exceptions2.py
```diff
@@ -0,0 +1,225 @@
...
+
+class AuthorizationError(Error):
+    """Caller lacks permission for the requested operation."""
+
+
+class ExternalServiceError(Error):
+    """An upstream provider or external store returned an error."""
+
+    provider: str | None = None
+    http_status_code: int | None = None
```
aaron-siegel: Maybe call this `external_status_code` or `external_http_status_code` just to be clear?

aaron-siegel (approved):

pixeltable/exceptions2.py
mkornacker: Done.

pixeltable/exceptions2.py
mkornacker: Apparently not.

pixeltable/exceptions2.py
```diff
@@ -0,0 +1,227 @@
...
+    Error codes for Pixeltable exceptions.
+
+    Codes are assigned explicit stable values in grouped ranges.  Each code must map to a specific HTTP status code.
+    New codes must be added at the end of their group and cannot reuse or shift existing values.
+    """
+
+    http_status: int
+    is_retryable: bool
+
+    def __new__(cls, value: int, http_status: int, is_retryable: bool) -> Self:
```
sergey-mkhitaryan, line 24: nit: `is_retriable` seems to be the preferred spelling in this repo

pixeltable/exceptions2.py
```diff
@@ -0,0 +1,227 @@
...
+        obj = object.__new__(cls)
+        obj._value_ = value
+        obj.http_status = http_status
+        obj.is_retryable = is_retryable
+        return obj
+
+    # Error (0xxx)
+    INTERNAL_ERROR = 0, 500, False
+
+    # NotFoundError (1xxx)
```
sergey-mkhitaryan, line 34: It's not obvious to me that NotFoundError and AlreadyExistsError are two distinct top-level groups. Maybe they should be under one roof of "Invalid schema" errors, with perhaps different subgroups (11xx vs 12xx) for not found vs already exists.

pixeltable/exceptions2.py
```diff
@@ -0,0 +1,227 @@
...
+        """Return a JSON-serializable representation for REST error responses."""
+        d: dict[str, Any] = {'error_code': self.error_code.name, 'message': str(self), 'retryable': self.is_retryable}
+        if self.__cause__ is not None:
+            d['cause'] = str(self.__cause__)
+        if self.retry_after is not None:
+            d['retry_after'] = self.retry_after
+        return d
+
+
+class NotFoundError(Error):
```
sergey-mkhitaryan, line 128: I'm worried that if `e = pxt.Error(pxt.INSUFFICIENT_PRIVILEGES)` then catching it as `except AuthorizationError:` won't work. You can argue that pxt.Errors shouldn't be instantiated directly using Error ctor like that, but what if it was something like `e = pxt.Error.from_dict(...)`.

Should we employ some python magic so that pxt.Error constructs objects of the appropriate subclass? I think that overriding `Error.__new__` can achieve that. Or a metaclass, though that is probably an overkill.

mkornacker: We're not going to create exceptions from dicts. These won't go into the catalog or the stored data.

pixeltable/exceptions2.py
mkornacker: True, but according to Google, retryable is widely preferred in the computing world.

But we should standardize on one. Probably retryable.

pixeltable/exceptions2.py
mkornacker: The reason they're separate is that you need to deal with them differently - the exception handlers look different. The goal of the hierarchy is to differentiate based on typical handler logic (otherwise we'd just have a single class with specific error codes).

pixeltable/exceptions2.py
```diff
@@ -0,0 +1,227 @@
...
+        obj.is_retryable = is_retryable
+        return obj
+
+    # Error (0xxx)
+    INTERNAL_ERROR = 0, 500, False
+
+    # NotFoundError (1xxx)
+    COLUMN_NOT_FOUND = 1000, 404, False
+    PATH_NOT_FOUND = 1001, 404, False
+    DIRECTORY_NOT_FOUND = 1002, 404, False
```
amithadke, line 37: is this for missing ancestor in a path?

PR #1283, opened 2026-04-18
Initial pxt CLI

mkornacker:
- currently only 'pxt serve'
- also adds docs for pxt serve as well as the serving config toml format


pixeltable/serving/_config.py
```diff
@@ -0,0 +1,158 @@
...
+            raw = toml.load(f)
+    except OSError as e:
+        raise pxt.Error(f'could not read service configuration file {config_path!r}: {e}') from e
+    except toml.TomlDecodeError as e:
+        raise pxt.Error(f'invalid TOML in service configuration file {config_path!r}: {e}') from e
+    try:
+        return AppConfig.model_validate(raw)
+    except pydantic.ValidationError as e:
+        raise pxt.Error(f'invalid service configuration in {config_path}:\n{e}') from e
+
```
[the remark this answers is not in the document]
mkornacker: The config path in error messages is a user-supplied argument (not an internal ~/.pixeltable path), so exposing it in the error message is intentional and helpful. Not changing.

pixeltable/serving/_config.py
```diff
@@ -0,0 +1,177 @@
+"""TOML-driven configuration for Pixeltable HTTP services."""
+
+from __future__ import annotations
+
+import importlib
+import logging
+from typing import TYPE_CHECKING, Annotated, Any, Literal
+
+if TYPE_CHECKING:
```
aaron-siegel: This should go after all other imports (otherwise ruff's import sorting can get confused)

pixeltable/serving/_config.py
```diff
@@ -0,0 +1,177 @@
+"""TOML-driven configuration for Pixeltable HTTP services."""
```
aaron-siegel, line 1: Why are we using a separate `_config.py` for this, versus tapping into the existing Pixeltable configuration system? It seems like it will splinter our config system in a way that could be confusing to users. I think this needs more discussion.

pixeltable/serving/_config.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def create_app_from_config(config: AppConfig) -> 'fastapi.FastAPI':
+    """Build a FastAPI instance from an AppConfig"""
+    try:
+        import fastapi
+
+        from ._fastapi import FastAPIRouter
+    except ImportError as e:
+        raise pxt.Error("FastAPI is required for serving; install it with `pip install 'fastapi[standard]'`") from e
```
aaron-siegel: Use Env.require_package()

aaron-siegel (changes requested):

pixeltable/serving/_config.py
mkornacker: Done.

pixeltable/serving/_config.py
mkornacker: As discussed, follow-up work in conjunction with pxt deploy.

pixeltable/serving/_config.py
mkornacker: Done.

aaron-siegel (approved):

tests/serving/test_config.py
```diff
@@ -0,0 +1,192 @@
...
+        t.insert([{'id': 1, 'text': 'hello'}, {'id': 2, 'text': 'world'}])
+
+        # define a query function in a temporary module
+        import types
+
+        query_mod = types.ModuleType('_test_query_mod')
+        # we need to exec in the module's namespace so @pxt.query sees the right globals
+        exec(
+            """
+import pixeltable as pxt
```
aaron-siegel: Preferred style is to indent normally and then use `textwrap.dedent()` if you really need it flushed left.

PR #1287, opened 2026-04-21
PXT-1118: test_overlay_text() fails in CI

mkornacker:
The problem was the a query in the test has a Python-predicate where(), which gets evaluated prior to actual expr evaluation (incl cache prefetch, etc.).

The solution is to remove evaluation of Python-based filters in SqlNode directly and instead move this into a separate node (FilterNode), which is inserted after the CachePrefetch-/CellMaterializationNodes/ExprEvalNodes.


aaron-siegel (changes requested):
Code lgtm but:
1. Did you uncomment the skip statement for test_overlay_text()?
2. I think there should be some tests that explicitly test Python filters, including with limit clauses, and including the specific case that's failing in test_overlay_text() (yes text_overlay_text() captures it, but it'd be good to have a more isolated test that atomically exercises the query execution functionality that was broken)

tests/test_video.py
```diff
@@ -851,7 +860,6 @@ def test_video_splitter_errors(self, uses_db: None) -> None:
         with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='overlap cannot be specified with segment_times'):
             _ = pxt.create_view('s', t, iterator=video_splitter(t.video, segment_times=[1, 2], overlap=1))
 
-    @pytest.mark.skipif(IN_CI, reason='[PXT-1118] Bug involving media reference in where clause')
     # @pytest.mark.skipif('t4' in os.environ.get('PXTTEST_CI_OS', ''), reason='Fonts not available on t4 CI instances')
```
aaron-siegel: Be sure to uncomment this line or it will cause the nightly job to fail.

(3rd time I've left this same comment)

aaron-siegel (changes requested):

aaron-siegel (approved):

PR #1288, opened 2026-04-21
[PXT-1012] Fix iterator view update when base column is an iterator parameter

christopherpestano:
When a base table column used as an iterator parameter is updated, the component view now correctly re-evaluates the iterator. Previously, the view's iterator output was stale after the update because the dependency system didn't track iterator argument references and the update path didn't know to reload the view.

Two changes fix this. First, record_column_dependencies now registers iterator output columns as depending on the base columns referenced by the iterator's bound args. This causes them to appear in the recomputed columns list during cascading updates. Second, propagate_update detects when recomputed columns include iterator columns and switches from a view update plan to a view load plan, which deletes the old component rows and re-runs the iterator with the updated parameter values.

This also works correctly when the iterator view itself has dependent views.

In the future we should unify these two paths to clean up the code a bit, but that is a nice to have. We should fix the bug that already exists.


pixeltable/catalog/table_version.py
```diff
@@ -1343,11 +1343,17 @@ def propagate_update(
...
+                    col.is_iterator_col for col in recomputed_cols
+                )
                 plan = None
-                if len(recomputed_cols) > 0:
-                    plan = Planner.create_view_update_plan(view.get().path, recompute_targets=recomputed_cols)
-                status = view.get().propagate_update(
+                if needs_iterator_reload:
+                    plan, _ = Planner.create_view_load_plan(view_tv.path, propagates_insert=True)
+                elif len(recomputed_cols) > 0:
+                    plan = Planner.create_view_update_plan(view_tv.path, recompute_targets=recomputed_cols)
```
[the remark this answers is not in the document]
christopherpestano: Fixed

pixeltable/catalog/catalog.py
```diff
@@ -895,14 +895,24 @@ def _compute_column_dependents(self, mutable_tree: set[UUID]) -> None:
...
             dependencies[QColumnId(tbl_version.id, col.id)] = Expr.get_refd_column_ids(col.value_expr_dict)
+
+        if tbl_version.is_component_view:
+            iterator_arg_deps: set[QColumnId] = set()
+            for arg_expr in tbl_version.iterator_call.bound_args.values():
+                for col_ref in arg_expr.subexprs(ColumnRef):
+                    iterator_arg_deps.add(QColumnId(col_ref.col.tbl_handle.id, col_ref.col.id))
+            if len(iterator_arg_deps) > 0:
+                for col in tbl_version.iterator_columns():
+                    dependencies[QColumnId(tbl_version.id, col.id)] = iterator_arg_deps
```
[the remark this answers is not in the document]
christopherpestano: Fixed

pixeltable/catalog/table_version.py
```diff
@@ -1343,11 +1343,17 @@ def propagate_update(
...
+                    col.id <= view_tv.num_iterator_cols for col in recomputed_cols
+                )
                 plan = None
-                if len(recomputed_cols) > 0:
-                    plan = Planner.create_view_update_plan(view.get().path, recompute_targets=recomputed_cols)
-                status = view.get().propagate_update(
+                if needs_iterator_reload:
+                    plan, _ = Planner.create_view_load_plan(view_tv.path, propagates_insert=True)
+                elif len(recomputed_cols) > 0:
+                    plan = Planner.create_view_update_plan(view_tv.path, recompute_targets=recomputed_cols)
```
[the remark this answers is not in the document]
christopherpestano: we already account for this

pixeltable/catalog/table_version.py
```diff
@@ -1733,13 +1735,12 @@ def display_str(self) -> str:
...
-        # the iterator columns directly follow the pos column
-        return self.is_component_view and col.id > 0 and col.id < self.num_iterator_cols + 1
+        """Returns True if col is produced by an iterator (including the pos column)"""
+        return col.is_iterator_col
 
     def iterator_columns(self) -> list[Column]:
-        """Return all iterator-produced columns"""
-        return self.cols[1 : self.num_iterator_cols + 1]
+        """Return all iterator-produced columns (including the pos column)"""
+        return self.cols[: self.num_iterator_cols]
```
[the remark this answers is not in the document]
christopherpestano: fixed

pixeltable/exprs/column_ref.py
```diff
@@ -80,7 +80,8 @@ def __init__(
         self.reference_tbl = reference_tbl
         self.col_handle = col.handle
 
-        self.is_unstored_iter_col = col.is_iterator_col and not col.is_stored
+        # pos (id=0) is an unstored iterator column, but its value comes from the PK, not the iterator output dict
+        self.is_unstored_iter_col = col.is_iterator_col and not col.is_stored and col.id != 0
```
sergey-mkhitaryan: This attribute name no longer makes sense because of the way you changed the meaning of "iterator column". It needs to be renamed. Could be `is_iterator_evaluated` or `needs_iterator_evaluation` or something like that, based on who it is used.

tests/test_iterator.py
```diff
@@ -97,6 +97,16 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
             raise pxt.Error(pxt.ErrorCode.GENERIC_USER_ERROR, 'Parameter `str_text` must be a valid identifier.')
 
 
+class ParamEchoRow(TypedDict):
+    scol: str
+
+
+@pxt.iterator
+def param_echo_iterator(a: int) -> Iterator[ParamEchoRow]:
```
sergey-mkhitaryan: any chance you can use `simple_iterator` instead? it's nearly identical to this

aaron-siegel: agreed

christopherpestano: Ok, changed to use simple iterator

pixeltable/exprs/column_ref.py
christopherpestano: needs_iterator_evaluation sounds good, the original term was vague to begin with

pixeltable/exprs/row_builder.py
```diff
@@ -192,12 +192,7 @@ def __init__(
         #   iterated object changes
 
         col_refs = [e for e in self.unique_exprs if isinstance(e, ColumnRef)]
-
-        def refs_unstored_iter_col(col_ref: ColumnRef) -> bool:
-            tbl = col_ref.col.get_tbl()
-            return tbl.is_component_view and tbl.is_iterator_column(col_ref.col) and not col_ref.col.is_stored
-
-        unstored_iter_col_refs = [col_ref for col_ref in col_refs if refs_unstored_iter_col(col_ref)]
+        unstored_iter_col_refs = [col_ref for col_ref in col_refs if col_ref.needs_iterator_evaluation]
```
aaron-siegel: Maybe call this `col_refs_needing_iter_eval` for consistency

tests/test_iterator.py
```diff
@@ -720,18 +720,57 @@ def test_nested_iterator(self, uses_db: None) -> None:
...
+        t.update({'icol': 6})
+
+        rows = v.order_by(v.pos).collect()
+        assert len(rows) == 6, f'expected 6 rows after update, got {len(rows)}'
+        for row in rows:
+            assert row['icol'] == 6
+        scol_values = [row['scol'] for row in rows]
+        assert scol_values == ['t 0', 't 1', 't 2', 't 3', 't 4', 't 5']
+
+    def test_update_iterator_param_with_dependent_view(self, uses_db: None) -> None:
```
aaron-siegel: Let's make sure we also have tests that cover:
- Iterator views with additional columns
- Iterator views with subviews that are also iterator views

pixeltable/catalog/table_version.py
```diff
@@ -1343,11 +1339,17 @@ def propagate_update(
             base_versions = [None if plan is None else self.version, *base_versions]  # don't update in place
             # propagate to views
             for view in self.mutable_views:
+                view_tv = view.get()
                 recomputed_cols = [col for col in recomputed_view_cols if col.get_tbl().id == view.id]
+                needs_iterator_reload = view_tv.is_component_view and any(
```
aaron-siegel, line 1346: As discussed in person - this condition may be technically correct, but it is not the most natural way to reason about it (at least to me)

A code comment that explains what's going on would be useful here.

aaron-siegel (changes requested):

pixeltable/exprs/row_builder.py
christopherpestano: Renamed

tests/test_iterator.py
christopherpestano: Added a table with both of those test_update_iterator_param_with_dependent_view, including coverage for iterator view where the input is an output of a parent iterator view.

pixeltable/catalog/table_version.py
christopherpestano: Added comment explaining that all/any are equivalent here and the choice for any is for early exit

tests/test_component_view.py
```diff
@@ -410,3 +411,63 @@ def test_create_view_error(self, uses_db: None) -> None:
...
+
+    def test_update_iterator_param(self, uses_db: None) -> None:
+        """Updating a base table column used as an iterator parameter re-evaluates the iterator."""
+        t = pxt.create_table('tbl', {'n': pxt.Int})
+        v = pxt.create_view('view', t, iterator=simple_iterator(t.n, str_text='t'))
+        t.insert([{'n': 5}])
+        t.update({'n': 6})
+
+        rows = v.order_by(v.pos).collect()
+        assert len(rows) == 6, f'expected 6 rows after update, got {len(rows)}'
```
aaron-siegel, line 426: Do a query before the update also and verify that there are only 5 rows?

aaron-siegel (approved):

tests/test_component_view.py
christopherpestano: Sure can do, added assert above the update

PR #1291, opened 2026-04-22
[PXT-924] Improve error messages for incorrect udf call

sergey-mkhitaryan:
Note: only the one-signature path is being changed. I'm not touching the errors when the UDF is overloaded because there's not a single signature to blame.

New error messages:

```
missing a required argument: 'a'; expected (a: pxt.Int, b: pxt.Int), got ()
missing a required argument: 'a'; expected (a: pxt.Int, b: pxt.Int), got (x=Int)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (Int)
'a' parameter is positional only, but was passed as a keyword; expected (a: pxt.Int, b: pxt.Int), got (a=Int)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (Int, a=Int)
'b' parameter is positional only, but was passed as a keyword; expected (a: pxt.Int, b: pxt.Int), got (Int, b=Int)
missing a required argument: 'a'; expected (a: pxt.Int, b: pxt.Int), got ()
too many positional arguments; expected (a: pxt.Int, b: pxt.Int), got (Int)
missing a required argument: 'a'; expected (a: pxt.Int, b: pxt.Int), got (x=Int)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (a=Int)
missing a required argument: 'a'; expected (a: pxt.Int, b: pxt.Int), got (x=Int)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (a=Int)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (a=Int, x=Int)
multiple values for argument 'a'; expected (a: pxt.Int, b: pxt.Int), got (Int, a=Int)
missing a required argument: 'a'; expected (a: pxt.Int, b: pxt.Int), got ()
missing a required argument: 'a'; expected (a: pxt.Int, b: pxt.String | None), got ()
missing a required argument: 'a'; expected (a: pxt.Int, b: pxt.String | None), got (b=None)
got an unexpected keyword argument 'x'; expected (*args), got (x=Int)
got an unexpected keyword argument 'x'; expected (*args), got (Int, x=Int)
too many positional arguments; expected (**kwargs), got (Int)
too many positional arguments; expected (**kwargs), got (Int, x=Int)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (Int)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (Float)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (a=Float)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (a=Int)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (a=Int)
missing a required argument: 'b'; expected (a: pxt.Int, b: pxt.Int), got (a=Int)
```


pixeltable/func/function.py
```diff
@@ -197,7 +197,10 @@ def _bind_to_matching_signature(self, args: Sequence[Any], kwargs: dict[str, Any
         if len(self.signatures) == 1:
             # Only one signature: call _bind_to_signature() and surface any errors directly
             result = 0
-            bound_args = self._bind_to_signature(0, args, kwargs)
+            try:
+                bound_args = self._bind_to_signature(0, args, kwargs)
+            except TypeError as e:
+                raise TypeError(self._bind_error_msg(self.signatures[0], args, kwargs, e)) from e
```
christopherpestano: Why are we raising type errors explicitly to users? If an error is caught it makes more sense to reraise as a pixeltable error from the type error to keep parity with how the rest of the codebase handles error reraising.

pixeltable/func/function.py
```diff
@@ -217,6 +220,41 @@ def _bind_to_matching_signature(self, args: Sequence[Any], kwargs: dict[str, Any
         assert bound_args is not None
         return self._resolved_fns[result], bound_args
 
+    def _bind_error_msg(
+        self, signature: Signature, args: Sequence[Any], kwargs: dict[str, Any], cause: TypeError
+    ) -> str:
+        from pixeltable.exprs import Expr, Literal
+
+        def _val_repr(literal: Literal) -> str:
```
aaron-siegel: Feels like reinventing the wheel. Don't we have val-shortening methods elsewhere, e.g., in Formatter?

pixeltable/func/function.py
```diff
@@ -217,6 +220,41 @@ def _bind_to_matching_signature(self, args: Sequence[Any], kwargs: dict[str, Any
...
+            if literal.val is None:
+                return 'None'
+            repr_str = f'{type(literal.val).__name__}({literal.val!r})'
+            return repr_str[:100] + '...' if len(repr_str) > 100 else repr_str
+
+        def _param_desc(p: Parameter) -> str:
```
aaron-siegel: Can this be a property of `Parameter`? And can all of this be integrated with the existing logic for `Signature.__str__()`?

pixeltable/func/function.py
```diff
@@ -217,6 +220,41 @@ def _bind_to_matching_signature(self, args: Sequence[Any], kwargs: dict[str, Any
...
+            return f'{prefix}{p.name}{type_str}{default_str}'
+
+        def _arg_desc(v: Any) -> str:
+            if isinstance(v, Literal) and v.val is None:
+                return 'None'
+            if isinstance(v, Expr):
+                return str(v.col_type)
+            return type(v).__name__
+
+        param_descs = [_param_desc(p) for p in signature.parameters_by_pos]
```
aaron-siegel: spurious list creation (use a generator)

pixeltable/func/function.py
```diff
@@ -217,6 +220,41 @@ def _bind_to_matching_signature(self, args: Sequence[Any], kwargs: dict[str, Any
...
+            type_str = f': {p.col_type}' if p.col_type is not None else ''
+            default_str = f' = {_val_repr(p.default)}' if p.has_default() else ''
+            return f'{prefix}{p.name}{type_str}{default_str}'
+
+        def _arg_desc(v: Any) -> str:
+            if isinstance(v, Literal) and v.val is None:
+                return 'None'
+            if isinstance(v, Expr):
+                return str(v.col_type)
+            return type(v).__name__
```
aaron-siegel: I'd think we'd always want to use Pixeltable types here, right?

pixeltable/func/function.py
```diff
@@ -217,6 +220,41 @@ def _bind_to_matching_signature(self, args: Sequence[Any], kwargs: dict[str, Any
...
+
+        param_descs = [_param_desc(p) for p in signature.parameters_by_pos]
+        expected = ', '.join(param_descs)
+        arg_descs = [_arg_desc(a) for a in args]
+        kwarg_descs = [f'{k}={_arg_desc(v)}' for k, v in kwargs.items()]
+        provided = ', '.join(arg_descs + kwarg_descs)
+        return f'{cause}; expected ({expected}), got ({provided})'
```
aaron-siegel: Maybe `\n` instead of `; `

tests/test_function.py
```diff
@@ -152,50 +149,128 @@ def test_call(self, test_tbl: pxt.Table) -> None:
...
+    @pxt.udf()
+    def udf_variadic_kw(**kwargs: Any) -> int:
+        raise AssertionError()
 
-            @pxt.udf
-            def f1(_int_param: int) -> int:
-                return _int_param
+    def test_invalid_call(self, test_tbl: pxt.Table) -> None:
+        t = test_tbl
+        missing_a = r"missing a required argument: 'a'"
```
aaron-siegel: Elsewhere in the codebase, identifiers are not quoted in error messages if they appear after a colon.

tests/test_function.py
```diff
@@ -152,50 +149,128 @@ def test_call(self, test_tbl: pxt.Table) -> None:
...
+        missing_b = r"missing a required argument: 'b'"
+        pos_only_a = r"'a' parameter is positional only, but was passed as a keyword"
```
aaron-siegel: Let's use a consistent style for these error messages
```
parameter is positional only, but was passed as a keyword: a
```
etc.

tests/test_function.py
```diff
@@ -679,6 +753,46 @@ def _(wrong_param: str) -> ts.ColumnType:
             exc_info.value
         )
 
+        # bad default value
+        with pxt_raises(
+            pxt.ErrorCode.UNSUPPORTED_OPERATION,
```
aaron-siegel, line 812: should be INVALID_TYPE?

aaron-siegel (changes requested):

pixeltable/func/function.py
sergey-mkhitaryan: yes, that's what we've discussed at the standup

pixeltable/func/function.py
sergey-mkhitaryan: I'm not even sure that this code is reachable -- normally args and kwargs are either literals or exprs. So consider it a safety net.

pixeltable/func/function.py
sergey-mkhitaryan: I dislike multi-line errors and log statements because they make grepping the logs inconvenient. There are scenarios when you can't go without a line break, but I think that here it is tolerable enough.

tests/test_function.py
sergey-mkhitaryan: yes but it's outside of scope of this pr. this piece of diff is i'm reorganizing the already existing test scenarios.

tests/test_function.py
sergey-mkhitaryan: It comes from the built-in error raised by `Signature.bind`. It would be a substantial complication to change that. As discussed offline, it's probably not worth it.

tests/test_function.py
sergey-mkhitaryan: same as above

pixeltable/func/function.py
sergey-mkhitaryan: oh yeah, I don't know how I didn't find that

pixeltable/func/function.py
sergey-mkhitaryan: good point but this code went away

pixeltable/func/function.py
aaron-siegel: If it's a safety net, let's make it an assertion. (And if the tests cause that assertion to fail, then we'll know it's reachable.)

I'd argue it should either be an assertion, or be reached by a test.

aaron-siegel (approved):

PR #1292, opened 2026-04-23
Additional FastAPIRouter methods

mkornacker:
Added: insert_route(), add_update_route(), update_route()

Also: extended test coverage


pixeltable/serving/_fastapi.py
```diff
@@ -204,112 +201,367 @@ def add_insert_route(
...
+        Examples:
+            ```python
+            class GenerateResponse(pydantic.BaseModel):
+                caption: str
+                score: float
+
+            @router.insert_route(
+                t, path='/generate', inputs=['prompt'], outputs=['caption', 'score'], background=False
+            )
+            def format_response(*, caption: str, score: float) -> GenerateResponse:
```
aaron-siegel, line 286: It might be more flexible for the signature of the decorated function to be

```python
def format_response(outputs: dict[str, Any]) -> ...
```
This would be less boilerplate for the user to maintain if their endpoint evolves.

Or even better,
```python
def format_response(inputs: dict[str, Any], outputs: dict[str, Any]) -> ...
```
(I suspect we'll have users who want the inputs in their wrapper function as well as the outputs)

aaron-siegel (approved):

pixeltable/serving/_fastapi.py
mkornacker: Good to discuss in a follow-up, probably after we get some user feedback.

PR #1293, opened 2026-04-24
openai.responses()

aaron-siegel:


pixeltable/functions/openai.py
```diff
@@ -479,8 +502,8 @@ async def chat_completions(
         ...     response=chat_completions(messages, model='gpt-4o-mini')
         ... )
     """
-    if model_kwargs is None:
-        model_kwargs = {}
+    # Copy to avoid mutating the Literal's internal dict (which would corrupt the stored expression)
```
mkornacker: How is this a Literal?

pixeltable/functions/openai.py
```diff
@@ -479,8 +502,8 @@ async def chat_completions(
...
+    model_kwargs = dict(model_kwargs) if model_kwargs is not None else {}
```
mkornacker: model_kwargs = copy.copy(model_kwargs)?

pixeltable/functions/openai.py
```diff
@@ -536,31 +559,141 @@ def _(messages: list, model: str, model_kwargs: dict[str, Any] | None = None) ->
...
+    tool_choice: dict[str, Any] | None = None,
+    _runtime_ctx: env.RuntimeCtx | None = None,
+) -> dict:
+    """
+    Creates a model response for the given input.
+
+    Equivalent to the OpenAI `responses` API endpoint.
+    For additional details, see: <https://developers.openai.com/api/docs/guides/migrate-to-responses>
+
+    The Responses API is OpenAI's recommended API for new projects, offering built-in tools
```
mkornacker: Do we want to include this promotional paragraph?

pixeltable/functions/openai.py
```diff
@@ -536,31 +559,141 @@ def _(messages: list, model: str, model_kwargs: dict[str, Any] | None = None) ->
...
+    Request throttling:
+    Uses the rate limit-related headers returned by the API to throttle requests adaptively, based on available
+    request and token capacity. No configuration is necessary.
+
+    __Requirements:__
+
+    - `pip install openai`
+
+    Args:
+        input: A list of input items for the model, using the same role/content structure as Chat Completions
```
mkornacker: Might be better for the comment to stand on its own.

pixeltable/functions/openai.py
```diff
@@ -536,31 +559,141 @@ def _(messages: list, model: str, model_kwargs: dict[str, Any] | None = None) ->
...
+@pxt.udf(is_deterministic=False)
+async def responses(
+    input: list,
+    *,
+    model: str,
+    model_kwargs: dict[str, Any] | None = None,
+    tools: list[dict[str, Any]] | None = None,
+    tool_choice: dict[str, Any] | None = None,
+    _runtime_ctx: env.RuntimeCtx | None = None,
+) -> dict:
```
mkornacker, line 581: Add the appropriate TypedDict?

pixeltable/functions/openai.py
```diff
@@ -536,31 +559,141 @@ def _(messages: list, model: str, model_kwargs: dict[str, Any] | None = None) ->
...
+    for item in input:
+        content = item.get('content') if isinstance(item, dict) else None
+        if isinstance(content, list):
+            for part in content:
+                if (
+                    isinstance(part, dict)
+                    and part.get('type') == 'input_image'
+                    and isinstance(part.get('image_url'), PIL.Image.Image)
+                ):
+                    b64_encoded_image = to_base64(part['image_url'], format='png')
```
mkornacker, line 665: Is this always a good idea, even for large images?

pixeltable/functions/openai.py
aaron-siegel: It might have come here through a literal without copying, so we'd be mutating the internal state of the upstream literal. (This might be a more general potential issue - I haven't dug too deeply)

pixeltable/functions/openai.py
aaron-siegel: `dict(x)` is the preferred way to create a shallow copy of `x`

pixeltable/functions/openai.py
aaron-siegel: Question for @pierrebrunelle who put that in

pixeltable/functions/openai.py
aaron-siegel: Let's factor that out as a follow-up task since in general our TypedDict conversions are still prototypes, and I don't want that to hold up this PR.

pixeltable/functions/openai.py
aaron-siegel: Agreed. I changed it to mirror the docstring for chat_completions.

pixeltable/functions/openai.py
aaron-siegel: Maybe not, but it's how things are currently done - let's separately discuss if we want to revisit this (and if so, apply the changes to chat_completions as well, and any other UDFs where it makes sense)

pixeltable/functions/openai.py
aaron-siegel: Per our offline discussion, I took it out.

pixeltable/functions/openai.py
mkornacker, line 488: Might be a good idea to update the model strings to the latest, while you're here.

pixeltable/functions/openai.py
mkornacker: I'm still not following. A Literal or any other expr shouldn't get passed in here. FunctionCall.make_args() constructs the arguments from DataRow slots. Seeing an expr here would be a major bug.

pixeltable/functions/openai.py
```diff
@@ -536,31 +559,141 @@ def _(messages: list, model: str, model_kwargs: dict[str, Any] | None = None) ->
...
+            model_kwargs['tool_choice'] = 'required'
+        else:
+            assert tool_choice['tool'] is not None
+            model_kwargs['tool_choice'] = {'type': 'function', 'name': tool_choice['tool']}
+
+    if tool_choice is not None and not tool_choice['parallel_tool_calls']:
+        model_kwargs['parallel_tool_calls'] = False
+
+    # Serialize any images in input items (Responses API uses 'input_image' type)
+    input = copy.deepcopy(input)
```
mkornacker, line 655: Isn't this making copies of PIL images? Why is this copy needed? It doesn't look like input is being written.

pixeltable/functions/openai.py
mkornacker: Okay, I get it. This is a major bug. Please file a high-priority jira and turn this comment into a todo, with a note to take it out. The fix needs to happen in ExprEvalCtx.init_rows().

pixeltable/functions/openai.py
```diff
@@ -418,6 +418,29 @@ def _token_count_for_image(image: PIL.Image.Image) -> int:
...
+    num_tokens = 0.0
+    for message in messages:
+        num_tokens += 4  # every message follows <im_start>{role/name}\n{content}<im_end>\n
+        for key, value in message.items():
+            if isinstance(value, str):
+                num_tokens += len(value) / 4
+            elif isinstance(value, list):
+                for part in value:
+                    if isinstance(part, dict):
+                        if part.get('type') == text_type and isinstance(part.get('text'), str):
```
mkornacker: Can we do something stronger here than get() (and for 'image_url')? If this gets renamed, we'll never find out and it'll fail silently.

pixeltable/functions/openai.py
```diff
@@ -1039,6 +1173,13 @@ def invoke_tools(tools: Tools, response: exprs.Expr) -> exprs.InlineDict:
 
 @pxt.udf
 def _openai_response_to_pxt_tool_calls(response: dict) -> dict | None:
+    # Auto-detect: Responses API has 'output', Chat Completions has 'choices'
```
mkornacker, line 1173: assert 'output' in response or 'choices' in response? Otherwise we'll also get a silent failure.

pixeltable/functions/openai.py
mkornacker: Never mind, it's assigning to part.

mkornacker (approved):

pixeltable/functions/openai.py
aaron-siegel: I tried but our account doesn't have GPT5 access so I can't test them.

I'm not sure we should be using models that require specialized access as examples. Maybe there's something in between 4o and 5 that I can use.

pixeltable/functions/openai.py
aaron-siegel: Done.

pixeltable/functions/openai.py
aaron-siegel: Actually we won't - it will raise an exception when _chat_completions_to_pxt_tool_calls() is invoked. But anyway I added the assertion.

PR #1295, opened 2026-04-24
OpenAI image/audio/moderation UDFs: typed JSON return schema

mkornacker:
Define TypedDicts mirroring OpenAI SDK response models (ImagesResponse, ModerationCreateResponse, TranscriptionResponse, TranslationResponse and nested types). Field names and nullability match the SDK. Switch .dict() to model_dump(mode='json'). Add schema-string assertions and typed-projection assertions to image, audio, and moderations tests.

Also added SDK adherence test (PXT-1146).


aaron-siegel (commented):
General thoughts:

- You might want to update the SDK version in pyproject.toml to ensure we're working off the most recent version
- Should we have tests that validate the SDK response structure is the same? Since we're not using it directly, but instead trying to mirror it. This would ensure that if the SDK evolves, we'll be notified of it.

mkornacker:
@aaron-siegel I updated the SDK version to the latest.

I also filed a jira ticket for the tests:
https://pixeltable.atlassian.net/browse/PXT-1146

pixeltable/functions/openai.py
```diff
@@ -284,8 +284,83 @@ async def speech(input: str, *, model: str, voice: str, model_kwargs: dict[str,
...
+class TranslationResponse(TypedDict):
+    text: str | None
+    srt: str | None
+    vtt: str | None
+    duration: float | None
+    language: str | None
+    segments: list[TranscriptionSegment] | None
+
+
+def _assemble_audio_response(cls: type, result: Any, response_format: str) -> dict[str, Any]:
```
aaron-siegel: Use generics to avoid the subsequent `cast`s:

```
def _assemble_audio_response(cls: type[T], result: Any, response_format: str) -> T:
```

pixeltable/functions/openai.py
```diff
@@ -284,8 +284,83 @@ async def speech(input: str, *, model: str, voice: str, model_kwargs: dict[str,
...
+
+
+class Logprob(TypedDict):
+    token: str | None
+    bytes: list[float] | None
+    logprob: float | None
+
+
+class TranscriptionResponse(TypedDict):
+    text: str | None
```
aaron-siegel: `text` is not nullable according to the OpenAI spec

pixeltable/functions/openai.py
```diff
@@ -284,8 +284,83 @@ async def speech(input: str, *, model: str, voice: str, model_kwargs: dict[str,
...
+    srt: str | None
+    vtt: str | None
+    duration: float | None
```
aaron-siegel: `duration` is not nullable (I think you maybe confusing non-required TypedDict fields with required nullable fields?)

pixeltable/functions/openai.py
```diff
@@ -284,8 +284,83 @@ async def speech(input: str, *, model: str, voice: str, model_kwargs: dict[str,
...
+
+class Logprob(TypedDict):
+    token: str | None
+    bytes: list[float] | None
+    logprob: float | None
+
+
+class TranscriptionResponse(TypedDict):
+    text: str | None
+    srt: str | None
```
aaron-siegel: I guess we're defining these ad hoc, but they're not nullable either

pixeltable/functions/openai.py
```diff
@@ -284,8 +284,83 @@ async def speech(input: str, *, model: str, voice: str, model_kwargs: dict[str,
...
+    no_speech_prob: float
+
+
+class Logprob(TypedDict):
+    token: str | None
+    bytes: list[float] | None
+    logprob: float | None
+
+
+class TranscriptionResponse(TypedDict):
```
aaron-siegel: Not all fields are guaranteed to be present, so declare the `TypedDict` with `total=False`

pixeltable/functions/openai.py
```diff
@@ -284,8 +284,83 @@ async def speech(input: str, *, model: str, voice: str, model_kwargs: dict[str,
...
+    text: str | None
+    srt: str | None
+    vtt: str | None
+    duration: float | None
+    language: str | None
```
aaron-siegel: not nullable (but I think the ones below this *are* nullable)

pixeltable/functions/openai.py
```diff
@@ -284,8 +284,83 @@ async def speech(input: str, *, model: str, voice: str, model_kwargs: dict[str,
...
+    segments: list[TranscriptionSegment] | None
+    words: list[TranscriptionWord] | None
+    logprobs: list[Logprob] | None
+    usage: dict | None
+
+
+class TranslationResponse(TypedDict):
```
aaron-siegel: Same issues with nullable vs. `total=False`

pixeltable/functions/openai.py
```diff
@@ -878,8 +957,45 @@ def _decode_image_response(result: dict) -> dict:
...
+
+class Usage(TypedDict):
+    input_tokens: int
+    input_tokens_details: UsageInputTokensDetails
+    output_tokens: int
+    total_tokens: int
+    output_tokens_details: UsageOutputTokensDetails | None
+
+
+# Nullability mirrors the SDK's pydantic model: only 'created' is always populated. The other fields are
```
aaron-siegel: That doesn't mirror the SDK's pydantic model. See comments above.

If we really want to exactly mirror which fields are always populated in the Pydantic models, the way to do that is by using `typing_extensions.Required` and/or `typing_extensions.NotRequired`.

aaron-siegel (changes requested):

pixeltable/functions/openai.py
mkornacker: Done.

pixeltable/functions/openai.py
mkornacker: Done.

pixeltable/functions/openai.py
mkornacker: Done.

pixeltable/functions/openai.py
mkornacker: Done.

pixeltable/functions/openai.py
mkornacker: Keeping `| None`. Under our normalization (`_assemble_audio_response` surfaces absent SDK fields as None for a uniform dict shape across `response_format` variants), this field is legitimately nullable on our side even though the SDK marks it non-null. Full rationale on the total=False thread above; comment block updated in 65adf069.

pixeltable/functions/openai.py
mkornacker: Done.

pixeltable/functions/openai.py
```diff
@@ -748,8 +827,45 @@ def _decode_image_response(result: dict) -> dict:
...
+# populated for a subset of model families and endpoints (DALL-E vs gpt-image-*); absent values come back
+# as None rather than missing keys, so user code can navigate with explicit None checks.
+class ImagesResponse(TypedDict):
+    created: int
+    data: list[PIL.Image.Image] | None
+    usage: Usage | None
```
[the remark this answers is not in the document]
mkornacker: Dropped `| None` on `ImagesResponse.data` in 65adf069. All three endpoints (generations / edit / variations) populate it and `_decode_image_response` requires it as a list, so the `| None` was incorrect.

tests/functions/test_openai.py
```diff
@@ -50,6 +50,24 @@ def test_audio(self, uses_db: None) -> None:
                 model_kwargs={'prompt': 'Translate the recording from Spanish into English.', 'temperature': 0.05},
             )
         )
+        # Raw-format responses: srt and vtt are surfaced as top-level string fields in the TypedDict schema.
+        t.add_computed_column(
+            transcription_srt=transcriptions(t.speech, model='whisper-1', model_kwargs={'response_format': 'srt'})
+        )
+        t.add_computed_column(
+            translation_vtt=translations(t.speech, model='whisper-1', model_kwargs={'response_format': 'vtt'})
+        )
```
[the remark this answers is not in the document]
mkornacker: Done in 65adf069. The srt and vtt raw-format checks now run on a single-row mini-table inside `test_audio`, so the new format coverage adds 2 OpenAI calls (one srt + one vtt) instead of one per inserted row.

pixeltable/functions/openai.py
mkornacker: Done.

pixeltable/functions/openai.py
mkornacker: Done.

aaron-siegel (approved):

PR #1298, opened 2026-04-25
[PXT-1130] fix a pendingtableops corruption bug in catalog

sergey-mkhitaryan:
1. Add `tbl_version` to TableOp to improve pending ops identification. When finalizing pending ops, Catalog uses table version to update only the intended pendingtableops rows. All table operations implemented via pending table ops are required to bump table version, even drop table.
3. Table state is only changed to LIVE with the table's exclusive lock held, and only after double checking the state of pendingtableops in the same transaction.


pixeltable/catalog/tbl_ops.py
```diff
@@ -49,6 +49,13 @@ class TableOp:
     num_ops: int  # total number of ops forming the update operation
     status: OpStatus
 
+    # New version and schema version of the table. Both are required for new table ops, but declared optional for
+    # backward compatibility.
+    # The chance of encountering legacy TableOps with a None value is small and diminishing over time, so this should be
+    # safe enough to make required in a few months or so.
+    tbl_version: int | None
+    tbl_schema_version: int | None
```
mkornacker: Do we need schema_version? It always changes in lockstep with version.

pixeltable/catalog/view.py
```diff
@@ -212,9 +212,31 @@ def _create(
             key = TableVersionKey(UUID(tbl_id), 0 if is_snapshot else None, None)
             view_path = TableVersionPath(TableVersionHandle(key), base=base_version_path)
             ops = [
-                CreateTableMdOp(tbl_id=tbl_id, op_sn=0, num_ops=3, status=OpStatus.PENDING),
-                CreateStoreTableOp(tbl_id=tbl_id, op_sn=1, num_ops=3, status=OpStatus.PENDING),
-                LoadViewOp(tbl_id=tbl_id, op_sn=2, num_ops=3, status=OpStatus.PENDING, view_path=view_path.as_dict()),
+                CreateTableMdOp(
```
mkornacker: This is accumulating a lot of fields that could be set programmatically.

I'm wondering if it's time to introduce a utility (class?) that can be used to build TableOp sequences (ie, that auto-assigns op_sn, num_ops, status, tbl version, etc.). Kind of like a string builder.

tests/test_unversioned_table.py
mkornacker: Now that I'm looking at this, I'm not sure it's a good idea to test this by reaching into the underlying metadata/store.

We don't do that for versioned tables. You can go through the public API to retrieve the version history. We could also add validation logic into Catalog itself (and then only Catalog would need to interact with the internals).

tests/test_unversioned_table.py
```diff
@@ -57,6 +57,12 @@ def test_basic_ops(self, uses_db: None, reload_tester: ReloadTester) -> None:
                 ).scalar()
                 assert row_count == 1, (system_tbl, row_count)
 
+        pxt.drop_table(tbl)
+
+        with get_runtime().begin_xact() as conn:
+            inspector = sql.inspect(conn)
+            assert not inspector.has_table(store_name)
```
mkornacker: Same comment here.

mkornacker (commented):
First round.

pixeltable/catalog/tbl_ops.py
sergey-mkhitaryan: I had unversioned tables in mind. My concept of unversioned tables is that they do not have a "version". Initially in my pr I even wanted `version` to raise an error on an unversioned table, but we changed that. So `version` is now always 0 or None, and schema version will be the one identifying schema changes.

pixeltable/catalog/view.py
sergey-mkhitaryan: done

pixeltable/catalog/catalog.py
```diff
@@ -948,6 +913,122 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
...
+        Updates the pending op status in the store. If is_final_op, sets table status to LIVE after additional checks.
+
+        Note: is_final_op is a hint that this may have been the last pending table op. Due to possible concurrent schema
+        changes, only the store can be the final authority on the state of the table.
+
+        This function must be called inside a transaction with an exclusive table lock held.
+
+        Returns True if all pending ops have been resolved, False otherwise.
+        """
+        from .tbl_ops import DeleteTableMdOp
```
mkornacker: Move to top.

pixeltable/catalog/catalog.py
```diff
@@ -948,6 +913,122 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
...
+        # pending ops finalizer corrupts the table by updating pending ops associated with a future schema change, not
+        # the one that it was finalizing. This issue is described in more detail in PXT-1130.
+        # Note: all schema changes except create table increment schema version. Create table is not a problem because
+        # no other schema change can precede it.
+        # Note: the only known gap that this safeguard does not cover is table revert. The way table revert is
+        # implemented, it decrements table and/or schema versions. Which means that if we do a schema change, then
+        # revert, then a schema change again, those two schema changes will share version counters, therefore, with
+        # unlucky timing, the pendingtableops table can still get corrupted. The right fix is for that is not here, it
+        # is to reimplement revert by advancing schema/data versions, not decrementing them.
+        if op.tbl_version is None:
```
mkornacker, line 958: I'm thinking a backfill would be better than having these workarounds.

pixeltable/catalog/catalog.py
```diff
@@ -948,6 +913,122 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
...
+            _logger.info(
+                f'Finalize pending ops({tbl_id}): no PendingTableOp rows matched. Another process may have already '
+                'resolved the same pending op concurrently.'
+            )
+
+        if not is_final_op:
+            _logger.info(f'Finalize pending ops({tbl_id}): not final op, more pending ops to finalize')
+            return False
+
+        if self._read_pending_table_ops(tbl_id):
```
mkornacker: Is it useful to do this here, rather than the caller doing that directly? (and _set_pending_op_status() not returning anything, because it's not actionable)

pixeltable/catalog/catalog.py
```diff
@@ -853,12 +823,8 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
                             self.mark_modified_tvs(tv.handle)
 
                         _logger.debug(f'Finalize pending ops({tbl_id}): op {op!s} done, updating status')
-                        if is_final_op:
-                            status = conn.execute(reset_tbl_state_stmt)
-                            status = conn.execute(delete_ops_stmt)
+                        if self._set_pending_op_status(tbl_id, op, new_op_status, is_final_op=is_final_op):
```
mkornacker: Having this call also check for remaining pending ops makes the code harder to read. Is there a problem with starting another iteration to find out there aren't any ops left?

pixeltable/catalog/catalog.py
sergey-mkhitaryan: We've discussed a very similar case before:
https://github.com/pixeltable/pixeltable/pull/1153#discussion_r2838682369

The tradeoff is:
* if we leave it as is, it's just a few lines of code that can go away pretty safely in a few months. It's extremely unlikely that some catalog somewhere will have old pending table ops while pixeltable is being updated.
* a migration, on the other hand, would be an extra dev and operational effort. even more effort would be to test it because db dumps do not have pending table ops in them today. whereas this piece of code I've already tested manually.

pixeltable/catalog/catalog.py
sergey-mkhitaryan: Yeah, the reason why this function should check pending ops and update table status and tell the caller if it's done or not, is `DeleteTableMdOp` (the last op of drop table). `DeleteTableMdOp` is transactional, and if we go on to the next iteration of `_finalize_pending_ops` after executing it, we'll get `TABLE_NOT_FOUND` error. Hope that makes sense. At any rate, I've just reshuffled things a bit, so PTAL and let me know if this question is still open.

pixeltable/catalog/catalog.py
sergey-mkhitaryan: I've refactored `_finalize_pending_ops` so that there's only one transaction per iteration. But what I said above about drop table is still true, we want to check other ops' and table's state on every op status update.

pixeltable/catalog/catalog.py
```diff
@@ -775,18 +777,23 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
...
-                    q = (
-                        sql.select(schema.PendingTableOp)
-                        .where(schema.PendingTableOp.tbl_id == tbl_id)
-                        .order_by(schema.PendingTableOp.op_sn)
-                        .with_for_update()
-                    )
-                    rows = conn.execute(q).fetchall()
-                    assert len(rows) > 0
-                    ops = [TableOp.from_dict(dict(row.op)) for row in rows]
+                    ops = None
```
mkornacker: missing type decl

pixeltable/catalog/catalog.py
```diff
@@ -948,6 +925,117 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
...
+        # deletes all pendingtableops.
+        if rowcount == 0 and not isinstance(op, DeleteTableMdOp):
+            _logger.info(
+                f'Finalize pending ops({tbl_id}): no PendingTableOp rows matched. Another process may have already '
+                'resolved the same pending op concurrently.'
+            )
+
+        if not is_final_op:
+            _logger.info(f'Finalize pending ops({tbl_id}): not final op, more pending ops to finalize')
+            return (False, None)
```
mkornacker: I don't understand this logic. This is not a final op, and you're not returning the remaining ops.

pixeltable/catalog/catalog.py
```diff
@@ -948,6 +925,117 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
...
+
+        tbl_ops = self._read_pending_table_ops(tbl_id)
+        if len(tbl_ops) > 0:
```
mkornacker, line 995: How does this happen? Now it's final, and you just deleted the pending ops (or someone else did before you).

pixeltable/catalog/catalog.py
```diff
@@ -948,6 +925,117 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
...
+            sql.update(schema.Table)
+            .where(schema.Table.id == tbl_id)
+            .values(md=schema.Table.md.op('||')({'tbl_state': schema.TableState.LIVE.value, 'pending_stmt': None}))
+        )
+
+        _logger.info(f'Finalize pending ops({tbl_id}): no more pending ops, resetting table state')
+        rowcount = conn.execute(reset_tbl_state_stmt).rowcount
+        if rowcount == 0 and not isinstance(op, DeleteTableMdOp):
+            _logger.info(
+                f'Finalize pending ops({tbl_id}): no Table rows matched. Another process may have deleted the table '
```
mkornacker, line 1010: Is this useful?

pixeltable/catalog/catalog.py
sergey-mkhitaryan: Notice that this function `_set_pending_op_status` is called twice from `_finalize_pending_ops`, and the 2nd time we call it, we don't care about the remaining pending ops, because we are about to commit a transaction and finish the iteration. Returning pending ops from this function is a best effort thing that helps avoid an unnecessary read around the 1st call.

pixeltable/catalog/catalog.py
sergey-mkhitaryan: See:

> Note: is_final_op is a hint that this may have been the last pending table op. Due to possible concurrent schema changes, only the store can be the final authority on the state of the table.

Basically, this is exactly what happens in the scenario described in https://pixeltable.atlassian.net/browse/PXT-1130. We may or may not have deleted _some_ pending ops, but that doesn't mean that there are no further pending ops on this table. Therefore this read and check are necessary before we proceed to declare the table as LIVE again.

pixeltable/catalog/catalog.py
sergey-mkhitaryan: yes, let's keep it. this piece of code costs us nearly nothing, but it can be critical in understanding certain race scenarios.

pixeltable/catalog/catalog.py
```diff
@@ -775,18 +777,23 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
...
-                        .where(schema.PendingTableOp.tbl_id == tbl_id)
-                        .order_by(schema.PendingTableOp.op_sn)
-                        .with_for_update()
-                    )
-                    rows = conn.execute(q).fetchall()
-                    assert len(rows) > 0
-                    ops = [TableOp.from_dict(dict(row.op)) for row in rows]
+                    ops: list[TableOp] | None = None
+                    if rollover_op_update is not None:
+                        done, ops = self._set_pending_op_status(
```
mkornacker: Is there a good reason why this should also return ops, instead of always doing a read_pending_ops() below? I'm finding this harder to reason about.

sergey-mkhitaryan: it was a (probably micro-) optimization. removed

mkornacker (approved):

PR #1300, opened 2026-04-25
Additional functionality for mix_audio() and overlay_text()

mkornacker:


pixeltable/functions/video.py
```diff
@@ -800,14 +803,40 @@ def mix_audio(
         audio_volume: Volume multiplier for the added audio track. 1.0 is original volume.
         original_volume: Volume multiplier for the video's existing audio track. 1.0 is original volume.
         audio_start_time: Time in seconds at which the added audio begins playing in the output.
+        mix_duration: Controls which input determines the length of the mixed audio stream.
+
+            - ``"longest"`` *(default)*: the mix runs until the longer of the two audio inputs ends. Use
```
aaron-siegel: We use single backticks everywhere else for code blocks in docstrings

aaron-siegel: No need to explicitly say "default" right? It's in the signature

aaron-siegel (approved):

pixeltable/functions/video.py

PR #1301, opened 2026-04-27
[PXT-1073] Add export_iceberg

christopherpestano:
Add export_iceberg for exporting Pixeltable tables and queries to Apache Iceberg tables. The function takes a pyiceberg Catalog supplied by the caller and streams data into the target table via pyarrow RecordBatches, with a configurable batch_size_bytes and the same if_exists semantics (error/overwrite/append) as the existing exporters. Append mode validates that the source schema is compatible with the existing Iceberg table before writing.

Iceberg has no analogue for pyarrow's fixed-shape tensor type, so pxt.Array columns with a known shape are rejected with a clear error directing the caller to project the column to a list before exporting. JSON columns whose values cannot be reduced to a single concrete arrow type (for example a column containing both dicts and lists across rows, lists whose element types are incompatible, or fields whose every sampled value is None) are rejected with a clear error rather than silently coerced or dropped to a string fallback.

Tests cover all-types export with caller-side array-to-list conversion, fixed-shape tensor rejection, null handling, query filtering and column subsets, the error/overwrite/append branches, append schema mismatch, automatic namespace creation, and rejection of JSON columns that cannot be reduced to a single arrow type.


pixeltable/io/__init__.py
```diff
@@ -6,13 +6,14 @@
...
+from .iceberg import export_iceberg
 from .json import export_json, import_json
 from .lancedb import export_lancedb
 from .pandas import import_excel, import_pandas
 from .parquet import export_parquet, import_parquet
 
 __default_dir = {symbol for symbol in dir() if not symbol.startswith('_')}
-__removed_symbols = {'globals', 'hf_datasets', 'pandas', 'parquet', 'datarows', 'lancedb', 'csv', 'json'}
+__removed_symbols = {'globals', 'hf_datasets', 'pandas', 'parquet', 'datarows', 'lancedb', 'iceberg', 'csv', 'json'}
 __all__ = sorted(__default_dir - __removed_symbols)
```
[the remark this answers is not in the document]
christopherpestano: added

tests/io/test_iceberg.py
```diff
@@ -0,0 +1,196 @@
...
+        catalog = self._catalog(tmp_path)
+        pxt.io.export_iceberg(query, catalog, 'pxt.all_types')
+
+        iceberg_tbl = catalog.load_table('pxt.all_types')
+        exported = iceberg_tbl.scan().to_arrow().sort_by('row_id').to_pylist()
+
+        assert len(exported) == len(rows)
+
+        for exp_row, orig_row in zip(exported, rows):
+            assert exp_row['c_string'] == orig_row['c_string']
```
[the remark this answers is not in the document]
christopherpestano: Fixed so we don't explicitly order, just compare input and output.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Iceberg's pyarrow visitor only handles canonical Arrow types. Fixed-shape tensor columns
+    # have no Iceberg analogue (the shape can't be persisted), so we reject them rather than
+    # silently dropping shape information.
+    unsupported = [f.name for f in first_batch.schema if isinstance(f.type, pa.FixedShapeTensorType)]
+    if unsupported:
+        raise excs.RequestError(
+            excs.ErrorCode.UNSUPPORTED_OPERATION,
+            f'export_iceberg(): cannot export fixed-shape tensor column(s) {unsupported}. '
+            f'Iceberg has no fixed-shape tensor type; project the column to a list before exporting.',
```
sergey-mkhitaryan: what's the problem with doing this automatically for array columns that have a known element type?

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+from typing import Literal, cast
+
+import pixeltable as pxt
+import pixeltable.exceptions as excs
+from pixeltable.env import Env
+
+
+def export_iceberg(
+    table_or_query: pxt.Table | pxt.Query,
+    catalog: 'pyiceberg.catalog.Catalog',  # type: ignore[name-defined]  # noqa: F821
```
sergey-mkhitaryan:
```
if TYPE_CHECKING:
    from pyiceberg.catalog.sql import SqlCatalog
```
then you can use SqlCatalog instead of the quoted type annotation. I believe this won't break minimal configuration.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
+from __future__ import annotations
+
+from typing import Literal, cast
```
sergey-mkhitaryan: we already have `iceberg.py` elsewhere (and, oddly, your new test in this pr is the only user of it). is it possible to combine the two or at least rename one of them?

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+
+    from pixeltable.utils.arrow import to_record_batches
+
+    if if_exists not in ('error', 'overwrite', 'append'):
+        raise excs.RequestError(
+            excs.ErrorCode.INVALID_ARGUMENT,
+            "export_iceberg(): 'if_exists' must be one of: ['error', 'overwrite', 'append']",
+        )
+
+    query: pxt.Query
```
sergey-mkhitaryan, line 75: a bit shorter:
```
    if isinstance(table_or_query, pxt.catalog.Table):
        table_or_query = table_or_query.select()
```
(and then use `table_or_query` instead of `query`)

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+                excs.ErrorCode.PATH_ALREADY_EXISTS, f'export_iceberg(): table {table_name!r} already exists'
+            )
+        if if_exists == 'overwrite':
+            catalog.drop_table(table_name)
+            iceberg_tbl = None
+
+    batch_iter = to_record_batches(query, batch_size_bytes)
+    first_batch = next(batch_iter, None)
+    if first_batch is None:
+        return
```
sergey-mkhitaryan: it's a bug, if if_exists is overwrite, you'll exit without re-creating the iceberg table. `test_if_exists` probably needs this scenario.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Iceberg's pyarrow visitor only handles canonical Arrow types. Fixed-shape tensor columns
+    # have no Iceberg analogue (the shape can't be persisted), so we reject them rather than
+    # silently dropping shape information.
+    unsupported = [f.name for f in first_batch.schema if isinstance(f.type, pa.FixedShapeTensorType)]
+    if unsupported:
+        raise excs.RequestError(
```
sergey-mkhitaryan, line 108: this needs to happen before you've dropped the iceberg table

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+    Args:
+        table_or_query: Table or Query to export.
+        catalog: An Iceberg `Catalog` instance to write the table into.
+        table_name: Fully-qualified Iceberg table identifier (e.g. `'pxt.my_table'`). If the namespace does not
+            exist, it will be created.
+        batch_size_bytes: Maximum size in bytes for each in-memory pyarrow batch.
+        if_exists: Determines the behavior if the table already exists. Must be one of the following:
+
+            - `'error'`: raise an error
+            - `'overwrite'`: drop the existing table and create a new one
```
sergey-mkhitaryan: this behavior sounds more like "replace" than "overwrite". to me, "overwrite" implies that the existing table is truncated, and new rows are inserted. that requires the existing table's schema to be compatible with the query, but drop-and-create implementation does not validate that.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+    table_name: str,
+    *,
+    batch_size_bytes: int = 128 * 2**20,
+    if_exists: Literal['error', 'overwrite', 'append'] = 'error',
+) -> None:
+    """
+    Exports a query result or table to an Apache Iceberg table.
+
+    Data is streamed into the Iceberg table via pyarrow `RecordBatches`, the size of which can be controlled with
+    the `batch_size_bytes` parameter. The destination Iceberg `Catalog` is supplied by the caller.
```
sergey-mkhitaryan: > The destination Iceberg `Catalog` is supplied by the caller.

pretty obvious from the signature, I'd just drop this.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+
+    __Requirements:__
+
+    - `pip install pyiceberg`
+
+    Args:
+        table_or_query: Table or Query to export.
```
sergey-mkhitaryan: a pixeltable table or query, just to draw the distinction with iceberg tables.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+                f'export_iceberg(): source schema is not compatible with existing table {table_name!r}: {e}\n'
+                f'Source schema:\n{first_batch.schema}\n'
+                f'Target schema:\n{target_schema}',
+            ) from e
+
+    with iceberg_tbl.transaction() as tx:
+        tx.append(pa.Table.from_batches([first_batch]))
+        for batch in batch_iter:
+            unwrapped = batch
+            for name in json_names:
```
sergey-mkhitaryan: this piece is duplicated, extract it to a function

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,127 @@
...
+        except (pa.ArrowInvalid, ValueError) as e:
+            raise excs.RequestError(
+                excs.ErrorCode.TYPE_MISMATCH,
+                f'export_iceberg(): source schema is not compatible with existing table {table_name!r}: {e}\n'
+                f'Source schema:\n{first_batch.schema}\n'
+                f'Target schema:\n{target_schema}',
+            ) from e
+
+    with iceberg_tbl.transaction() as tx:
+        tx.append(pa.Table.from_batches([first_batch]))
```
sergey-mkhitaryan: write a comment saying that json cols in the first batch were already preprocessed above

sergey-mkhitaryan (changes requested):

pixeltable/io/iceberg.py
christopherpestano: Makes sense added pixeltable

pixeltable/io/iceberg.py
christopherpestano: Dropped

pixeltable/io/iceberg.py
christopherpestano: This operation causes dataloss, before the transform you have an explicit shape afterwards you lose that shape. If intentional then that's fine, but users should be aware that the datatype changed in an unintuitive way.

pixeltable/io/iceberg.py
christopherpestano: The semantics are equivalent to how lancedb handles overwrite in io (replacement), changed to replace though since that's the more common pattern in this repo.

pixeltable/io/iceberg.py
christopherpestano: This syntax is intentional and a common pattern, in that snippet you are changing the semantics of what table_or_query is (the type has changed but the variable name has not). Having the base query case makes the code self documenting, whereas implicit type conversions are messy. (Also Aaron commented on this pattern before https://github.com/pixeltable/pixeltable/pull/1212/changes#r3025269190, though my reasoning is the above)

pixeltable/io/iceberg.py
christopherpestano: Sure works for me

pixeltable/io/iceberg.py
christopherpestano: Removed the other file, works for me. This is just an artifact of code shifting over time

pixeltable/io/iceberg.py
christopherpestano: added

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,142 @@
...
+    catalog: 'pyiceberg.catalog.Catalog',  # type: ignore[name-defined]  # noqa: F821
+    table_name: str,
+    *,
+    batch_size_bytes: int = 128 * 2**20,
+    if_exists: Literal['error', 'replace', 'append'] = 'error',
+) -> None:
+    """
+    Exports a query result or table to an Apache Iceberg table.
+
+    Data is streamed into the Iceberg table via pyarrow `RecordBatches`, the size of which can be controlled with
```
aaron-siegel: Link `RecordBatches` to the pyarrow docs?

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,142 @@
...
+        batch_size_bytes: Maximum size in bytes for each in-memory pyarrow batch.
+        if_exists: Determines the behavior if the table already exists. Must be one of the following:
+
+            - `'error'`: raise an error
+            - `'replace'`: drop the existing table and create a new one
+            - `'append'`: append to the existing table (source schema must be compatible)
+    """
+    Env.get().require_package('pyiceberg')
+
+    import pyarrow as pa
```
aaron-siegel: Can go at top of file? (I think arrow is a required dep?)

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,142 @@
...
+    from pyiceberg.exceptions import NoSuchTableError
+
+    from pixeltable.utils.arrow import to_record_batches
```
aaron-siegel: Top of file?

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,142 @@
...
+
+    if if_exists not in ('error', 'replace', 'append'):
+        raise excs.RequestError(
+            excs.ErrorCode.INVALID_ARGUMENT,
+            "export_iceberg(): 'if_exists' must be one of: ['error', 'replace', 'append']",
```
aaron-siegel: if_exists in backticks (it's a symbol)

pixeltable/io/iceberg.py
aaron-siegel: Doesn't our to_record_batches() code already do something like this in certain cases? i.e, we're already doing some type coercions under the covers, so we're saying "if we were able to do a type coercion the previous time around, run with it; but if not, raise an error rather than do a type coercion now" - seems like inconsistent behavior.

But I'm not sure. I just spent about 10 minutes with our to_record_batches() code and it's sufficiently convoluted I still don't understand all the edge cases and how they're represented in Pixeltable type -> arrow type mappings. Might be worth taking a step back and laying out all the different cases and going through:
- For each case, how does it map to an arrow type?
- What then is the observed behavior when we map to an iceberg type?

By cases I mean
- Fixed shape & dtype array column
- Fixed dtype, fixed dim but variable shape array column
- Fixed dtype, variable dim array column
- Variable everything array column
- Json column

Each with different combinations of data in the actual query (e.g. Json column containing only lists, Fixed dtype, variable dim array column that only contains arrays of a certain fixed shape, etc)

aaron-siegel (changes requested):

pixeltable/io/iceberg.py
christopherpestano: Fixed

pixeltable/io/iceberg.py
christopherpestano: arrow moved up, iceberg as optional package stays there

pixeltable/io/iceberg.py
christopherpestano: moved up

pixeltable/io/iceberg.py
christopherpestano: We are no longer doing json conversion so this code segment is gone.

pixeltable/io/iceberg.py
christopherpestano: Fixed we create the table even on empty batch

pixeltable/io/iceberg.py
christopherpestano: Had Claude help me generate this chart after I had fixed a good portion of the bugs, tl;dr basically iceberg has partial support for the array type, unless it is an array with fixed shape and dtype in which case it doesn't recognize the extension type we use with arrow. Also iceberg doesn't recognize null columns (which is how it interprets json with a always null field present) and mixed JSON where some rows are lists and some are dicts, and JSON in which it contains arrays with a non consistent value, i.e. ['this', 'is', 1, 'array'] is invalid and would fail. Not too certain what we want to do here, we could always coerce arrays to the list arrow type which is widely supported, but the JSON is a total mess depending on format, best it seems would be to raise an error to the user.

christopherpestano: Image since GitHub is bad a rendering tables
image

tests/io/test_iceberg.py
```diff
@@ -0,0 +1,222 @@
...
+        return sqlite_catalog(tmp_path / 'warehouse')
+
+    def test_export_all_types(self, uses_db: None, tmp_path: pathlib.Path) -> None:
+        """Export a table with every supported type and verify the Iceberg output."""
+        skip_test_if_not_installed('pyiceberg')
+        t = create_all_datatypes_tbl()
+
+        # Iceberg has no fixed-shape tensor type; the caller is expected to project the
+        # column to a list before exporting.
+        query = t.select(
```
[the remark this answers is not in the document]
christopherpestano: added tests

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,185 @@
...
+    if fixed_tensor_cols:
+        raise excs.RequestError(
+            excs.ErrorCode.UNSUPPORTED_OPERATION,
+            f'export_iceberg(): cannot export fixed-shape tensor column(s) {fixed_tensor_cols}. '
+            f'Iceberg has no fixed-shape tensor type; project the column to a list before exporting.',
+        )
+
+    batch_iter = to_record_batches(query, batch_size_bytes)
+    first_batch = next(batch_iter, None)
+
```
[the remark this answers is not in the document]
christopherpestano: This is intentional in a sense, the point is to create an empty table with the given schema as a best effort, while we could reject this it's better to support json schema

pixeltable/utils/arrow.py
```diff
@@ -148,10 +182,12 @@ def create_arrow_schema() -> None:
...
-                except (pa.lib.ArrowInvalid, pa.lib.ArrowTypeError):
-                    # for mixed lists e.g. json including both lists and dicts
-                    # we need to fall back to json strings
-                    pa_type = pa.json_()
+                except (pa.lib.ArrowInvalid, pa.lib.ArrowTypeError) as e:
+                    raise excs.RequestError(
+                        excs.ErrorCode.UNSUPPORTED_OPERATION,
+                        f'JSON column {col_name!r} contains mixed types (e.g. both lists and dicts), which is not '
+                        f'supported when exporting to Arrow.',
+                    ) from e
```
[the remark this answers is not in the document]
christopherpestano: Pr description changed

pixeltable/io/iceberg.py
christopherpestano: linked

pixeltable/utils/arrow.py
```diff
@@ -253,8 +308,6 @@ def to_pydict(batch: pa.Table | pa.RecordBatch) -> dict[str, list | np.ndarray]:
             if isinstance(col, pa.ChunkedArray):
                 col = col.combine_chunks()
             out[name] = list(cast(pa.FixedShapeTensorArray, col).to_numpy_ndarray())
-        elif col.type == pa.json_():
-            out[name] = [json.loads(v) if v is not None else None for v in col.to_pylist()]
         else:
             # for the rest, use pydict to preserve python types
             out[name] = col.to_pylist()
```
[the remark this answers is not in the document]
christopherpestano: This is fine, this export support is experimental to begin with

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,160 @@
...
+        # All preflight checks have passed; safe to drop the existing table now.
+        if existing_tbl is not None:
+            assert if_exists == 'replace'
+            catalog.drop_table(table_name)
+        catalog.create_namespace_if_not_exists(table_name.rsplit('.', 1)[0])
+        iceberg_tbl = catalog.create_table(table_name, schema=arrow_schema)
+
+    batches: Iterable[pa.RecordBatch] = chain([first_batch], batch_iter) if first_batch is not None else batch_iter
+    with iceberg_tbl.transaction() as tx:
+        for batch in batches:
```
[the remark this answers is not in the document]
christopherpestano: use temp table to make more atomic

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+        raise
+    except Exception as e:
+        catalog.drop_table(temp_name)
+        raise excs.RequestError(
+            excs.ErrorCode.UNSUPPORTED_OPERATION, f'export_iceberg(): failed to write Iceberg table {table_name!r}: {e}'
+        ) from e
+
+    if existing_tbl is not None:
+        catalog.drop_table(table_name)
+    catalog.rename_table(temp_name, table_name)
```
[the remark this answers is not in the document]
christopherpestano: Low risk, iceberg doesn't let you be truly atomic so this is a best effort

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+    Instantiate a sqlite Iceberg catalog at the specified path. If no catalog exists, one will be created.
+    """
+    Env.get().require_package('pyiceberg')
+
+    from pyiceberg.catalog.sql import SqlCatalog
+
+    if isinstance(warehouse_path, str):
+        warehouse_path = Path(warehouse_path)
+    warehouse_path.mkdir(exist_ok=True)
```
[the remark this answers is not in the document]
christopherpestano: Good catch. Fixed with parents=True so nested warehouse paths get created.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+            "export_iceberg(): `if_exists` must be one of: ['error', 'replace', 'append']",
+        )
+
+    # pyiceberg requires a namespace-qualified identifier; bare names raise opaque errors deep
+    # inside catalog.load_table / create_table. Reject them up front with a clear message.
+    if '.' not in table_name:
+        raise excs.RequestError(
+            excs.ErrorCode.INVALID_ARGUMENT,
+            f"export_iceberg(): table_name {table_name!r} must be namespace-qualified (e.g. 'pxt.my_table').",
+        )
```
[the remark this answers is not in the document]
christopherpestano: Fixed. Now validates that both the namespace and name segments are non-empty using rpartition, so '.tbl' and 'ns.' are rejected up front. Added two cases in test_if_exists.

tests/io/test_iceberg.py
```diff
@@ -0,0 +1,246 @@
...
+        # Invalid if_exists value
+        with pxt_raises(pxt.ErrorCode.INVALID_ARGUMENT, match='must be one of'):
+            pxt.io.export_iceberg(t, catalog, 'pxt.if_exists', if_exists='badval')  # type: ignore[arg-type]
+
+        # Replace + preflight failure: existing table must be preserved.
+        bad = pxt.create_table('test_iceberg_replace_bad', {'c_array': pxt.Array[(4,), pxt.Float]})  # type: ignore[misc]
+        bad.insert([{'c_array': np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32)}])
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='fixed-shape tensor'):
+            pxt.io.export_iceberg(bad, catalog, 'pxt.if_exists', if_exists='replace')
+        assert catalog.load_table('pxt.if_exists').scan().to_arrow().num_rows == 6
```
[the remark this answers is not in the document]
christopherpestano: Added a mid-stream failure case in test_if_exists. It wraps the catalog so create_table returns a Table whose transaction() raises after the temp table has been created in the catalog. The test asserts that the existing Iceberg table is preserved with its original row count and that no __pxt_tmp_ table is left behind in the namespace.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def export_iceberg(
+    table_or_query: pxt.Table | pxt.Query,
+    catalog: Catalog,
+    table_name: str,
+    *,
+    batch_size_bytes: int = 128 * 2**20,
+    if_exists: Literal['error', 'replace', 'append'] = 'error',
+) -> None:
```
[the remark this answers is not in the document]
christopherpestano: Aligned with export_lancedb. Renamed 'replace' to 'overwrite' in the public API, the docstring, the validation error message, and the tests.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+
+            - `'error'`: raise an error
+            - `'overwrite'`: drop the existing table and create a new one
+            - `'append'`: append to the existing table (source schema must be compatible)
+    """
+    Env.get().require_package('pyiceberg')
+
+    from pyiceberg.exceptions import NoSuchTableError
+
+    from pixeltable.utils.arrow import to_arrow_schema, to_record_batches
```
aaron-siegel: top of file? (Arrow is a required dep)

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+            f"export_iceberg(): table_name {table_name!r} must be namespace-qualified (e.g. 'pxt.my_table').",
+        )
+
+    query: pxt.Query
+    if isinstance(table_or_query, pxt.catalog.Table):
+        query = table_or_query.select()
+    else:
+        query = table_or_query
+
+    existing_tbl = None
```
aaron-siegel: add type hint

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+
+    from pyiceberg.catalog.sql import SqlCatalog
+
+    if isinstance(warehouse_path, str):
+        warehouse_path = Path(warehouse_path)
+    warehouse_path.mkdir(parents=True, exist_ok=True)
+    return SqlCatalog(name, uri=f'sqlite:///{warehouse_path}/catalog.db', warehouse=f'file://{warehouse_path}')
+
+
+def export_iceberg(
```
aaron-siegel, line 21: Given the finickiness of our type conversions, I'd argue that a schema_overrides is essential.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+import pixeltable.exceptions as excs
+from pixeltable.env import Env
+from pixeltable.utils.arrow import find_null_fields
+
+if TYPE_CHECKING:
+    from pyiceberg.catalog import Catalog
+    from pyiceberg.catalog.sql import SqlCatalog
+
+
+def sqlite_catalog(warehouse_path: str | Path, name: str = 'pixeltable') -> SqlCatalog:
```
aaron-siegel: Is this used anywhere other than tests? If not, move to tests/utils or some such

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+            f'export_iceberg(): cannot export fixed-shape tensor column(s) {fixed_tensor_cols}. '
+            f'Iceberg has no fixed-shape tensor type; project the column to a list before exporting.',
+        )
+
+    batch_iter = to_record_batches(query, batch_size_bytes)
+    first_batch = next(batch_iter, None)
+
+    arrow_schema = first_batch.schema if first_batch is not None else schema_arrow
+
+    # `pa.infer_type` produces `pa.null()` for JSON keys whose value is None in every sampled row.
```
aaron-siegel: It'd be more logical to me for this check to occur as part of pa.infer_type(); for example via an `allow_null_type` parameter.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,187 @@
...
+    # Iceberg format-version 2 cannot represent a null-only column, so reject it up front rather
+    # than letting pyiceberg fail mid-write with a less actionable error.
+    null_paths = find_null_fields(arrow_schema)
+    if null_paths:
+        raise excs.RequestError(
+            excs.ErrorCode.UNSUPPORTED_OPERATION,
+            f'export_iceberg(): cannot infer a concrete type for JSON field(s) {null_paths} because every sampled '
+            f'value is None. Iceberg has no null-only type; populate the field with at least one non-None value '
```
aaron-siegel: This is where schema_overrides would enable us to give a more friendly advisory: "Specify its type explicitly via schema_overrides".

Asking the user to "populate the field" isn't really useful feedback IMO (we're saying "modify your data so that you can do this export")

aaron-siegel (changes requested):

pixeltable/io/iceberg.py
christopherpestano: moved

pixeltable/io/iceberg.py
christopherpestano: Added type hint; since Iceberg just calls their table class "Table" ended up doing it as from iceberg.table import Table as IcebergTable to make it obvious to readers which one is the iceberg table and which one is the pixeltable table.

pixeltable/io/iceberg.py
christopherpestano: Moved my bad

pixeltable/io/iceberg.py
christopherpestano: Agreed; that's actually quite the elegant solution between having auto schema coercion vs requiring well formatted input data for arrow outputs. Implemented.

pixeltable/io/iceberg.py
christopherpestano: Yeah fair enough, the alternative was to exclude the field if you don't have anything to populate it with. Added the schema overrides and changed the error message here to specify the type.

pixeltable/io/iceberg.py
christopherpestano: I get where this is coming from, but I think the post hoc walk ends up being a lot simpler. The check is the same recursive type tree walk either way (null column, null struct field, null list element), so wrapping it inside pa.infer_type just relocates find_null_fields without removing it. It also couples the inference helper to a downstream restriction, and a future caller reading "pa.infer_type(..., allow_null_type=False)" has to go figure out which consumer that flag exists for.

The other thing is that the error message needs format specific context ("Iceberg has no null-only type"), which has to come from the caller anyway. We're not really avoiding any iceberg specific code by moving the trigger upstream, we're just splitting it across two layers.

Of the formats we currently export to (parquet, lancedb, iceberg), iceberg is the only one that actually rejects pa.null() at the spec level. Parquet/Arrow consumers are fine with it.. And with schema_override implemented, the user has a clean way to fix it on the way in.

pixeltable/io/iceberg.py
```diff
@@ -38,6 +23,7 @@ def export_iceberg(
     *,
     batch_size_bytes: int = 128 * 2**20,
     if_exists: Literal['error', 'overwrite', 'append'] = 'error',
+    schema_override: Mapping[str, pa.DataType] | None = None,
```
aaron-siegel: Elsewhere we call this `schema_overrides`, I believe.

pixeltable/io/iceberg.py
```diff
@@ -61,13 +47,15 @@ def export_iceberg(
             - `'error'`: raise an error
             - `'overwrite'`: drop the existing table and create a new one
             - `'append'`: append to the existing table (source schema must be compatible)
+        schema_override: Optional mapping from column name to an explicit pyarrow `DataType` that should be used
```
aaron-siegel: This differs from the phrasing used in other `schema_overrides` docstrings. Let's please strive for consistency - either align with existing docstrings or if you think there's room for improvement, propose new phrasing and apply it in analogous docstrings as well.

I'd argue that the "Useful for..." would be better placed in the Examples section, keeping the Args docstring focused on a factual description of what the parameter does.

pixeltable/io/iceberg.py
```diff
@@ -125,7 +113,7 @@ def export_iceberg(
         raise excs.RequestError(
             excs.ErrorCode.UNSUPPORTED_OPERATION,
             f'export_iceberg(): cannot infer a concrete type for JSON field(s) {null_paths} because every sampled '
-            f'value is None. Iceberg has no null-only type; populate the field with at least one non-None value '
+            f'value is None. Iceberg has no null-only type; provide a `schema_override` for the field '
```
aaron-siegel: I'd say "you can specify an explicit type using the `schema_overrides` parameter."

I'm not sure what "omit it from the data" actually conveys?

aaron-siegel (approved):

pixeltable/io/iceberg.py
christopherpestano: Changed to the string you gave; in this case the omit from your data is a hint to pass in a query where do some data preprocessing to the json column to remove the null field. Removed that since the canonical path is to just give a schema override.

pixeltable/io/iceberg.py
christopherpestano: Changed to align with parquet which also has an override. I think we don't need the useful for section so I just removed it entirely, it should be pretty obvious to the user (given they got an error) as to why they should use it.

pixeltable/io/iceberg.py
christopherpestano: My bad switched

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+    # Build a deterministic arrow schema up front so we can materialize the Iceberg table even
+    # when the query yields no rows, and reject fixed-shape tensor columns before we run the query.
+    # Variable-shape arrays are mapped to pa.list_(...) by to_arrow_type and are supported by Iceberg.
+    schema_arrow = to_arrow_schema(query.schema, schema_overrides)
+    fixed_tensor_cols = [f.name for f in schema_arrow if isinstance(f.type, pa.FixedShapeTensorType)]
+    if fixed_tensor_cols:
+        raise excs.RequestError(
+            excs.ErrorCode.UNSUPPORTED_OPERATION,
+            f'export_iceberg(): cannot export fixed-shape tensor column(s) {fixed_tensor_cols}. '
+            f'Iceberg has no fixed-shape tensor type; project the column to a list before exporting.',
```
sergey-mkhitaryan: What's the easiest way to perform this projection? Is it a new UDF like what you did in the test? I'd argue that having users write such trivial UDFs creates a bad experience. There should be a built-in `to_list` method UDF for arrays.

If that existed, this error message could direct the user to it instead of giving only abstract suggestions.

Can you create this UDF? Shouldn't be very hard.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+
+    batch_iter = to_record_batches(query, batch_size_bytes, schema_overrides)
+    first_batch = next(batch_iter, None)
+
+    arrow_schema = first_batch.schema if first_batch is not None else schema_arrow
```
sergey-mkhitaryan: `arrow_schema` vs `schema_arrow` is confusing. maybe combine them into one variable? you don't need both of them after this point.

sergey-mkhitaryan: FYI iceberg 3 supports `VARIANT` type which sounds a lot like JSON. The work is underway to support it in pyarrow and pyiceberg; once that is complete, I think we should be able to get rid of this awkwardness with schemaless jsons and arrays. I'd leave a TODO for that.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+            ) from e
+        try:
+            with existing_tbl.transaction() as tx:
+                for batch in batches:
+                    tx.append(pa.Table.from_batches([batch]))
+        except excs.Error:
+            raise
+        except Exception as e:
+            raise excs.RequestError(
+                excs.ErrorCode.UNSUPPORTED_OPERATION,
```
sergey-mkhitaryan: this should be an INTERNAL_ERROR, not UNSUPPORTED_OPERATION

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+    except excs.Error:
+        catalog.drop_table(temp_name)
+        raise
+    except Exception as e:
+        catalog.drop_table(temp_name)
+        raise excs.RequestError(
+            excs.ErrorCode.UNSUPPORTED_OPERATION, f'export_iceberg(): failed to write Iceberg table {table_name!r}: {e}'
+        ) from e
+
+    if existing_tbl is not None:
```
sergey-mkhitaryan: can you either write 
`assert if_exists == 'overwrite'`

or leave a comment that says so? it'll make the code easier to follow

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+        with temp_tbl.transaction() as tx:
+            for batch in batches:
+                tx.append(pa.Table.from_batches([batch]))
+    except excs.Error:
+        catalog.drop_table(temp_name)
+        raise
+    except Exception as e:
+        catalog.drop_table(temp_name)
+        raise excs.RequestError(
+            excs.ErrorCode.UNSUPPORTED_OPERATION, f'export_iceberg(): failed to write Iceberg table {table_name!r}: {e}'
```
sergey-mkhitaryan: same here re error code

pixeltable/io/iceberg.py
sergey-mkhitaryan: I think it'll be more robust if you move this replace sequence to inside the try block above. then you'll be able to drop the temp table (while ignoring NoSuchTableError) in the finally block.

It will also resolve the bug that if the user presses ctrl+c while the temporary table is being populated, the temporary table won't be dropped (because `KeyboardInterrupt` is not an `Exception`)

pixeltable/utils/arrow.py
```diff
@@ -107,31 +135,53 @@ def to_pxt_schema(
...
     for field in schema:
         assert field.name in pxt_schema
         pxt_type = pxt_schema[field.name]
-        if isinstance(field.type, pa.FixedShapeTensorType):
+        if schema_overrides is not None and field.name in schema_overrides:
+            try:
+                pa_arrays.append(pa.array(column_vals[field.name], type=field.type))
+            except (pa.ArrowInvalid, pa.ArrowTypeError, pa.ArrowNotImplementedError) as e:
+                raise excs.RequestError(
+                    excs.ErrorCode.UNSUPPORTED_OPERATION,
```
sergey-mkhitaryan: TYPE_MISMATCH

pixeltable/utils/arrow.py
```diff
@@ -107,31 +135,53 @@ def to_pxt_schema(
...
+        elif pxt_type.is_json_type():
+            # JSON columns are typed by `pa.infer_type` against the first batch's values; that can
+            # succeed (e.g. `list<int64>` inferred from `[1, 'a', 2]`) but still fail when pyarrow
+            # actually coerces the values. Catch that here and surface a clear error.
+            try:
+                pa_arrays.append(pa.array(column_vals[field.name], type=field.type))
+            except (pa.ArrowInvalid, pa.ArrowTypeError) as e:
+                raise excs.RequestError(
+                    excs.ErrorCode.UNSUPPORTED_OPERATION,
+                    f'JSON column {field.name!r} contains values that cannot be coerced to a single '
```
sergey-mkhitaryan, line 175: we should direct the user to start using `schema_overrides` for this table

pixeltable/utils/arrow.py
```diff
@@ -145,13 +195,17 @@ def create_arrow_schema() -> None:
...
                 try:
                     pa_type = pa.infer_type(batch_columns[col_name], mask=None)
-                except (pa.lib.ArrowInvalid, pa.lib.ArrowTypeError):
-                    # for mixed lists e.g. json including both lists and dicts
-                    # we need to fall back to json strings
-                    pa_type = pa.json_()
+                except (pa.lib.ArrowInvalid, pa.lib.ArrowTypeError) as e:
+                    raise excs.RequestError(
+                        excs.ErrorCode.UNSUPPORTED_OPERATION,
+                        f'JSON column {col_name!r} contains mixed types (e.g. both lists and dicts), which is not '
```
sergey-mkhitaryan, line 212: ditto

tests/utils.py
```diff
@@ -758,6 +763,20 @@ def validate_sync_status(
         assert status.num_excs == expected_num_excs, status
 
 
+def sqlite_catalog(warehouse_path: str | Path, name: str = 'pixeltable') -> 'SqlCatalog':
```
sergey-mkhitaryan: since this is no longer in an iceberg-named file, please rename it to indicate that it's an iceberg catalog.

tests/io/test_iceberg.py
```diff
@@ -0,0 +1,328 @@
...
+    from pyiceberg.catalog import Catalog
+    from pyiceberg.table import Table
+
+
+@pxt.udf
+def array_to_list(arr: pxt.Array[(10,), pxt.Float]) -> pxt.Json:
+    return arr.tolist()
+
+
+class _MidStreamFailingTable:
```
sergey-mkhitaryan: I don't know how long it'll take to merge https://github.com/pixeltable/pixeltable/pull/1326, but the fault injection that it adds would significantly simplify this.

tests/io/test_iceberg.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._real, name)
+
+    def create_table(self, *args: Any, **kwargs: Any) -> _MidStreamFailingTable:
+        return _MidStreamFailingTable(self._real.create_table(*args, **kwargs))
+
+
+class TestIceberg:
+    def _catalog(self, tmp_path: pathlib.Path) -> Any:
```
sergey-mkhitaryan: you can make it a `@classmethod` or move it out up to the module level

tests/io/test_iceberg.py
```diff
@@ -0,0 +1,328 @@
...
+            pxt.io.export_iceberg(t, catalog, 'pxt.if_exists', if_exists='badval')  # type: ignore[arg-type]
+
+        # Overwrite + preflight failure: existing table must be preserved.
+        bad = pxt.create_table('test_iceberg_replace_bad', {'c_array': pxt.Array[(4,), pxt.Float]})  # type: ignore[misc]
+        bad.insert([{'c_array': np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32)}])
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='fixed-shape tensor'):
+            pxt.io.export_iceberg(bad, catalog, 'pxt.if_exists', if_exists='overwrite')
+        assert catalog.load_table('pxt.if_exists').scan().to_arrow().num_rows == 6
+
+        # Overwrite + mid-stream failure (after the temp table is created): the existing table
```
sergey-mkhitaryan: nice!

tests/utils.py
```diff
@@ -128,19 +131,20 @@ def create_table_data(
...
                     'confidence': 0.95,
                     'iscrowd': 0,
                     'verified': True,
                 }
             ]
         },
-        {'level1': {'level2': {'level3': [1, 2.0, 'three', False, None]}}},
-        [{'key': 'val1', 'n': 1}, {'key': 'val2', 'n': 2}],
-        {'s': 'hello', 'i': 42, 'f': 2.718, 'b': False, 'n': None},
+        # Removing json_() arrow fallback means we don't support mixed-type lists and None only fields
```
sergey-mkhitaryan: This code is used in many pixeltable tests, not just pyarrow-related. You shouldn't be making this change as it reduces the test coverage for all of them. Please find some other way to avoid mixed jsons in only arrow-related tests.

sergey-mkhitaryan (changes requested):

pixeltable/utils/arrow.py
```diff
@@ -107,31 +135,53 @@ def to_pxt_schema(
...
         elif pxt_type.is_array_type() and isinstance(field.type, pa.ListType):
             # convert ragged arrays to nested lists
             list_col_vals = [val.tolist() for val in column_vals[field.name]]
             pa_arrays.append(pa.array(list_col_vals))
-        elif field.type == pa.json_():
-            serialized = [json.dumps(v) if v is not None else None for v in column_vals[field.name]]
-            pa_arrays.append(pa.array(serialized, type=pa.json_()))
+        elif pxt_type.is_json_type():
+            # JSON columns are typed by `pa.infer_type` against the first batch's values; that can
+            # succeed (e.g. `list<int64>` inferred from `[1, 'a', 2]`) but still fail when pyarrow
```
mkornacker: This comment is wrong.

pixeltable/utils/arrow.py
```diff
@@ -107,31 +135,53 @@ def to_pxt_schema(
...
+            # actually coerces the values. Catch that here and surface a clear error.
+            try:
+                pa_arrays.append(pa.array(column_vals[field.name], type=field.type))
```
mkornacker, line 171: This should have some assert on field.type.

Also, where did dumps() go?

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+        raise excs.AlreadyExistsError(
+            excs.ErrorCode.PATH_ALREADY_EXISTS, f'export_iceberg(): table {table_name!r} already exists'
+        )
+
+    # Build a deterministic arrow schema up front so we can materialize the Iceberg table even
+    # when the query yields no rows, and reject fixed-shape tensor columns before we run the query.
+    # Variable-shape arrays are mapped to pa.list_(...) by to_arrow_type and are supported by Iceberg.
+    schema_arrow = to_arrow_schema(query.schema, schema_overrides)
+    fixed_tensor_cols = [f.name for f in schema_arrow if isinstance(f.type, pa.FixedShapeTensorType)]
+    if fixed_tensor_cols:
```
mkornacker: We generally compare explicitly: len(...) > 0

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+            f'export_iceberg(): cannot export fixed-shape tensor column(s) {fixed_tensor_cols}. '
+            f'Iceberg has no fixed-shape tensor type; project the column to a list before exporting.',
+        )
+
+    batch_iter = to_record_batches(query, batch_size_bytes, schema_overrides)
+    first_batch = next(batch_iter, None)
+
+    arrow_schema = first_batch.schema if first_batch is not None else schema_arrow
+
+    # `pa.infer_type` produces `pa.null()` for JSON keys whose value is None in every sampled row.
```
mkornacker: We typically refer to functions as pa.infer_type(). Easier to read.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+        raise excs.RequestError(
+            excs.ErrorCode.UNSUPPORTED_OPERATION,
+            f'export_iceberg(): cannot infer a concrete type for JSON field(s) {null_paths} because every sampled '
+            f'value is None. Iceberg has no null-only type; you can specify an explicit type using the schema_overrides parameter.',
+        )
+
+    catalog.create_namespace_if_not_exists(namespace)
+
+    batches: Iterable[pa.RecordBatch] = chain([first_batch], batch_iter) if first_batch is not None else batch_iter
+
```
mkornacker, line 137: Use blank lines to break up larger code blocks.

pixeltable/io/iceberg.py
christopherpestano: Changed to match existing pattern

tests/utils.py
christopherpestano: Changed to iceberg_catalog seems fair

tests/io/test_iceberg.py
christopherpestano: Thanks! Tried to emulate the fault injection logic we have been iterating on.

tests/io/test_iceberg.py
christopherpestano: Fair enough self is not being used here, changed to classmethod so the strong association with the TestIceberg class is kept.

tests/io/test_iceberg.py
christopherpestano: Yeah I think so too, though to some extend though this isn't modeling a failure during the pixeltable transaction though but rather a mid stream failure during the iceberg transaction. Meaning that there would still have to be some boilerplate here to actually trigger a failure in the pyiceberg library but it would still benefit greatly from the presence of FaultManager

pixeltable/utils/arrow.py
christopherpestano: Fixed

pixeltable/io/iceberg.py
christopherpestano: Fixed error code

pixeltable/io/iceberg.py
christopherpestano: Fixed

pixeltable/io/iceberg.py
christopherpestano: My bad missed this one, fixed

pixeltable/io/iceberg.py
christopherpestano: In this case the user would still have to specify via a schema override that they want this behavior, as it would not be supported by other catalogs like lancedb and etc I presume. Rather than leaving a TODO whch is weak and likely will be forgotten, I'll put it as a work item in the jira which is more likely we'll come back and do it.

christopherpestano: Added work item PXT-1163

pixeltable/utils/arrow.py
christopherpestano: I thought we agreed that we would not be converting json to strings? Json dumps was the experimental support for json strings that we had added before that we're supposed to remove.

pixeltable/utils/arrow.py
christopherpestano: Looked into this and there actually isn't a pa.DataType the user could pass via schema_overrides that would fix the mixed lists/dicts case. The override path feeds raw Python objects into pa.array(values, type=T), and pyarrow won't coerce a dict or list into bytes, so pa.string(), pa.json_() and pa.binary() all fail with "Expected bytes, got a dict". Union types aren't supported by pa.array either. The only way to make schema_overrides actually work here would be to special case string/json_/binary overrides and run json.dumps on the values before pa.array, basically putting back the old experimental json string path but as an explicit opt in via schema_overrides. @mkornacker can you weigh in on whether we should add that opt in path, or leave the error as a hard stop here?

pixeltable/utils/arrow.py
christopherpestano: Same answer as the comment above.

pixeltable/io/iceberg.py
christopherpestano: Sure created the udf and added it to public documentation.

pixeltable/io/iceberg.py
christopherpestano: Added more blank lines to try and break up the big block

pixeltable/utils/arrow.py
christopherpestano: Changed to the failure mechanism more common here, where the first batch assumes something about the schema and then a later batch has some data that doesn't fit the inferred type

tests/utils.py
christopherpestano: That's ok, added an additional arrow_compatible_json param that gives preexisting tests the non-arrow compatible json while preserving arrow test functionality.

pixeltable/utils/arrow.py
christopherpestano: Also not too clear to me assert here should be on field.type are we simply trying to check that the inferred type of the values matches the type in the schema? That is already caught by the try except in case of coercion failure.

pixeltable/io/iceberg.py
christopherpestano: I think better yet is to just call schema_arrow -> fallback_schema that way the role it has is obvious and makes it self documenting

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,185 @@
...
+        raise
+    except Exception as e:
+        raise excs.RequestError(
+            excs.ErrorCode.INTERNAL_ERROR, f'export_iceberg(): failed to write Iceberg table {table_name!r}: {e}'
+        ) from e
+    finally:
+        # Drop temp if it still exists; covers mid-stream errors and BaseException paths
+        # like KeyboardInterrupt. On success, rename consumed temp_name, so swallow NoSuchTableError.
+        try:
+            catalog.drop_table(temp_name)
```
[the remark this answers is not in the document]
christopherpestano: This is such an unlikely mechanism of failure, under what circumstances would a rename table even fail? And even if it did, and both tables were wiped, the user would still have the data present within their pixeltable table. They could simply attempt the export again.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,185 @@
...
+    if existing_tbl is not None and if_exists == 'error':
+        raise excs.AlreadyExistsError(
+            excs.ErrorCode.PATH_ALREADY_EXISTS, f'export_iceberg(): table {table_name!r} already exists'
+        )
+
+    # Build a deterministic arrow schema up front so we can materialize the Iceberg table even
+    # when the query yields no rows, and reject fixed-shape tensor columns before we run the query.
+    # Variable-shape arrays are mapped to pa.list_(...) by to_arrow_type and are supported by Iceberg.
+    fallback_schema = to_arrow_schema(query.schema, schema_overrides)
+    fixed_tensor_cols = [f.name for f in fallback_schema if isinstance(f.type, pa.FixedShapeTensorType)]
```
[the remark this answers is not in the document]
christopherpestano: Fixed

tests/io/test_iceberg.py
sergey-mkhitaryan: Here we are, fault manager is finally in place. Please consider cleaning this up and instead adding a fault location in `export_iceberg()`, perhaps even in multiple points for better coverage.

sergey-mkhitaryan (approved):

pixeltable/utils/arrow.py
```diff
@@ -93,6 +92,35 @@ def to_arrow_type(pxt_type: ts.ColumnType) -> pa.DataType | None:
         return None
 
 
+def to_arrow_schema(
+    pxt_schema: dict[str, ts.ColumnType], schema_overrides: Mapping[str, pa.DataType] | None = None
+) -> pa.Schema:
+    """Build a deterministic `pa.Schema` from a pixeltable schema.
+
+    Mirrors the per-column type logic that `to_record_batches` uses, but does not require any
```
mkornacker: Claude loves dropping references to other pieces of code (or caller logic) into code comments. This creates a problem when the referenced code changes.

The docstring needs to stand on its own.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+    which can be controlled with the `batch_size_bytes` parameter.
+
+    __Requirements:__
+
+    - `pip install pyiceberg`
+
+    Args:
+        table_or_query: Pixeltable `Table` or `Query` to export.
+        catalog: An Iceberg `Catalog` instance to write the table into.
+        table_name: Fully-qualified Iceberg table identifier (e.g. `'pxt.my_table'`). If the namespace does not
```
mkornacker: I don't know if it's a good idea to use our standard package abbreviation pxt as part of the path, it might confuse some people.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+            - `'append'`: append to the existing table (source schema must be compatible)
+        schema_overrides: If specified, then for each (name, type) pair in `schema_overrides`, the column with
+            name `name` will be given type `type`, instead of being inferred from the Pixeltable schema. The keys in
+            `schema_overrides` should be the column names of the Pixeltable schema.
+    """
+    Env.get().require_package('pyiceberg')
+
+    from pyiceberg.exceptions import NoSuchTableError
+
+    if if_exists not in ('error', 'overwrite', 'append'):
```
mkornacker: I'm still seeing overwrite here, not sure that's intentional.

'replace' is used elsewhere.

pixeltable/io/iceberg.py
mkornacker: We use explicit comparison everywhere.

pixeltable/io/iceberg.py
mkornacker: Better: also provide an example of how to do that (my guess is most users would be stumped at this point).

pixeltable/io/iceberg.py
mkornacker: We also qualify function calls with (), eg," pa.infer_type()" consistently throughout the entire code base. Easier to read.

pixeltable/io/iceberg.py
mkornacker: Non-obvious: why not go with schema_arrow generally?

Non-obvious code is worth a brief comment.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,173 @@
...
+
+    if existing_tbl is not None and if_exists == 'append':
+        target_schema = existing_tbl.schema().as_arrow()
+        try:
+            # Cast a zero-row slice to the target schema: pyarrow validates field names match
+            # exactly and that source types can be promoted to target types (e.g. string -> large_string).
+            sample = (
+                first_batch.slice(0, 0)
+                if first_batch is not None
+                else pa.RecordBatch.from_arrays([pa.array([], type=f.type) for f in arrow_schema], schema=arrow_schema)
```
mkornacker, line 147: Also not obvious: why even look at first_batch if the fallback works just as well.

mkornacker (approved):

pixeltable/utils/arrow.py
christopherpestano: Done. Rewrote the docstring so it no longer refers to to_record_batches; it now describes the mapping directly: each column maps via to_arrow_type(), and JSON columns map to an empty pa.struct([]) since there is no data to infer from.

pixeltable/io/iceberg.py
christopherpestano: Good point. Changed the example in both the docstring and the validation error message to 'my_namespace.my_table' instead of 'pxt.my_table'.

pixeltable/io/iceberg.py
christopherpestano: Switched to 'replace' to match the convention used elsewhere (create_table, create_view, export_to_sql). Updated the Literal, the validation message, the docstring, and the tests.

pixeltable/io/iceberg.py
christopherpestano: Changed the null_paths check to len(null_paths) > 0. The fixed_tensor_cols check above was already len(...) > 0.

pixeltable/io/iceberg.py
christopherpestano: Made the error message give a concrete example the user can copy directly, using the actual offending column name, e.g. export_iceberg(t.select(embedding=t.embedding.to_list()), catalog, table_name). Verified that following it exports the column as an Iceberg list type.

pixeltable/io/iceberg.py
christopherpestano: Done. That comment now reads pa.infer_type(), and there are no remaining unqualified references in the file.

pixeltable/io/iceberg.py
christopherpestano: Added a comment. We prefer the first batch's schema because it carries the concrete nested types inferred from real JSON values (e.g. struct<x: int64>), whereas fallback_schema maps every JSON column to an empty pa.struct([]). fallback_schema is only used when the query is empty and there is nothing to infer.

pixeltable/io/iceberg.py
christopherpestano: Same reason as the adjacent comment: first_batch.schema captures the concrete JSON struct types inferred from the data, which fallback_schema cannot (it maps JSON to an empty pa.struct), so fallback_schema is only the empty-query fallback. Added a comment explaining this.

tests/utils.py
```diff
@@ -761,6 +778,20 @@ def validate_sync_status(
...
+    Instantiate a sqlite Iceberg catalog at the specified path. If no catalog exists, one will be created.
+    """
+    Env.get().require_package('pyiceberg')
+
+    from pyiceberg.catalog.sql import SqlCatalog
+
+    if isinstance(warehouse_path, str):
+        warehouse_path = Path(warehouse_path)
+    warehouse_path.mkdir(parents=True, exist_ok=True)
+    return SqlCatalog(name, uri=f'sqlite:///{warehouse_path}/catalog.db', warehouse=f'file://{warehouse_path}')
```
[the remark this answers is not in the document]
christopherpestano: Preexisting code, already works on windows

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,205 @@
...
+            raise excs.RequestError(
+                excs.ErrorCode.UNSUPPORTED_OPERATION,
+                f'export_iceberg(): some column(s) specified in `schema_overrides` are not present '
+                f'in the source: {", ".join(sorted(extraneous_overrides))}',
+            )
+
+    # Build a deterministic arrow schema up front so we can materialize the Iceberg table even
+    # when the query yields no rows, and reject fixed-shape tensor columns before we run the query.
+    # Variable-shape arrays are mapped to pa.list_(...) by to_arrow_type and are supported by Iceberg.
+    fallback_schema = to_arrow_schema(query.schema, schema_overrides)
```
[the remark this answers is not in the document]
christopherpestano: Fixed. Moved the guard into to_arrow_type itself, which is the shared conversion point used by both to_arrow_schema and to_record_batches, so an unknown-shape array column now raises a RequestError instead of a bare AssertionError across all export paths (iceberg, parquet, lancedb). Columns pinned via schema_overrides bypass to_arrow_type, so overriding such a column still works.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    Env.get().require_package('pyiceberg')
+
+    from pyiceberg.exceptions import NoSuchTableError
+
+    if if_exists not in ('error', 'replace', 'append'):
+        raise excs.RequestError(
+            excs.ErrorCode.INVALID_ARGUMENT,
+            "export_iceberg(): `if_exists` must be one of: ['error', 'replace', 'append']",
+        )
```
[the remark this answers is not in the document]
christopherpestano: We are intentionally keeping 'replace' here as a deliberate naming decision, so we will not add an 'overwrite' alias.

tests/io/test_iceberg.py
```diff
@@ -0,0 +1,328 @@
...
+            t.c_int,
+            t.c_json,
+            t.c_string,
+            t.c_timestamp,
+            t.c_uuid,
+            t.c_binary,
+            t.c_video,
+            t.c_document,
+            c_array=array_to_list(t.c_array),
+        )
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The test now uses t.c_array.to_list() and the custom array_to_list UDF was removed.

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,205 @@
...
+            - `'replace'`: drop the existing table and create a new one
+            - `'append'`: append to the existing table (source schema must be compatible)
+        schema_overrides: If specified, then for each (name, type) pair in `schema_overrides`, the column with
+            name `name` will be given type `type`, instead of being inferred from the Pixeltable schema. The keys in
+            `schema_overrides` should be the column names of the Pixeltable schema.
+    """
+    Env.get().require_package('pyiceberg')
+
+    from pyiceberg.exceptions import NoSuchTableError
+
```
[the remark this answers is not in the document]
christopherpestano: Arrow is a required pxt dependency, there is no need to gate it

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,205 @@
...
+    Exports a query result or table to an Apache Iceberg table.
+
+    Data is streamed into the Iceberg table via pyarrow
+    [`RecordBatches`](https://arrow.apache.org/docs/python/generated/pyarrow.RecordBatch.html), the size of
+    which can be controlled with the `batch_size_bytes` parameter.
+
+    __Requirements:__
+
+    - `pip install pyiceberg`
+
```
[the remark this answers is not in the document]
christopherpestano: Per previous, arrow is a required dependency

pixeltable/io/iceberg.py
```diff
@@ -0,0 +1,205 @@
...
+import pixeltable.exceptions as excs
+from pixeltable.env import Env
+from pixeltable.utils import fault_injection
+from pixeltable.utils.arrow import find_null_fields, to_arrow_schema, to_record_batches
+from pixeltable.utils.fault_injection import FaultLocation
+
+if TYPE_CHECKING:
+    from pyiceberg.catalog import Catalog
+    from pyiceberg.table import Table as IcebergTable
+
```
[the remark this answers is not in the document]
christopherpestano: Pyarrow is already a hard dependency! Three different comments stating the same thing - how odd!

PR #1306, opened 2026-05-01
Local CockroachDB environment and CockroachDB compatibility fixes

christopherpestano:
This PR adds the ability to develop and test Pixeltable against CockroachDB, and fixes the incompatibilities that surfaced when doing so.

It introduces a devcontainer and docker-compose setup that brings up Pixeltable alongside a local CockroachDB instance, so the database backend can be exercised without any external infrastructure.

It also adds a CI job that runs against a single-node CockroachDB container. This job intentionally runs only a small selection of the test modules, chosen to track regressions in core functionality rather than to provide full coverage. Note: It is possible to run against the full test suite, however doing so makes it take 90 minutes to complete on the normal runner. We expect to deprecate CockroachDB support, so this job is not meant to validate the backend exhaustively. It exists only to catch regressions in core behavior, so that returning to CockroachDB later remains feasible if the alternatives do not work out.

CockroachDB is stricter than PostgreSQL about SQL type coercion: aggregate and extraction functions return different numeric types, and some operators require their operands to share a type. Several query translations were updated to add explicit casts so that results come back with the expected types and queries no longer fail or return incorrect values on CockroachDB. These changes are transparent on PostgreSQL.

In the course of this work we also found and fixed a latent bug in the microsecond extraction logic that produced incorrect results on PostgreSQL as well, and added a regression test that exercises sub-second values. Tests covering features that are not yet supported on CockroachDB are skipped when running against that backend.

Finally, a dependency constraint was adjusted so that the environment builds on linux/aarch64.

Validated CI Job: https://github.com/christopherpestano/pixeltable/actions/runs/26482697549/job/77983290036


tool/ci_tool.py
```diff
@@ -24,6 +24,21 @@
...
+    'tests/test_env.py',
+    'tests/test_exprs.py',
+    'tests/test_function.py',
+    'tests/test_index.py',
+    'tests/test_primary_key_index.py',
+    'tests/test_snapshot.py',
+    'tests/test_table.py',
+    'tests/test_view.py',
+    'tests/share/test_packager.py',
+)
```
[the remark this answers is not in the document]
christopherpestano: Added back

scripts/start-cockroach-ci.sh
```diff
@@ -0,0 +1,32 @@
...
+    --cache=1GiB \
+    --max-sql-memory=1GiB
+
+for _ in $(seq 1 60); do
+    if docker exec "$ROACH_NAME" /cockroach/cockroach sql --insecure -e 'SELECT 1' >/dev/null 2>&1; then
+        break
+    fi
+    sleep 2
+done
+
```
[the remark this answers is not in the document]
christopherpestano: If the container is truly not ready the next command docker exec would fail anyways, there is no issue here. This just adds an unnecessary validation layer for a slightly better log.

tool/ci_tool.py
```diff
@@ -24,6 +24,23 @@
 
 COCKROACH_TEST_MODULES = ('table', 'index')
 
+# CockroachDB suite modules: the `make slimtest` set (see Makefile) plus the primary-key index tests.
+# We intentionally exclude unversioned tables as they are
```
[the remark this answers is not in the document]
christopherpestano: fixed

scripts/start-cockroach-ci.sh
```diff
@@ -0,0 +1,32 @@
...
+    start-single-node --insecure \
+    --cache=1GiB \
+    --max-sql-memory=1GiB
+
+for _ in $(seq 1 60); do
+    if docker exec "$ROACH_NAME" /cockroach/cockroach sql --insecure -e 'SELECT 1' >/dev/null 2>&1; then
+        break
+    fi
+    sleep 2
+done
```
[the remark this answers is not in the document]
christopherpestano: This is fine, it will still error and it will be fairly obvious that something is wrong with the container if the docker exec fails, I fail to see how just printing the container is not ready is more informative than the actual container state.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+        build-essential \
+        sudo \
+        libgl1 \
+        libglib2.0-0 \
+ && rm -rf /var/lib/apt/lists/*
+
+# `ubuntu` (uid 1000) is preinstalled in the base image; grant it passwordless
+# sudo for ad-hoc apt installs in the running container.
+RUN echo 'ubuntu ALL=(ALL) NOPASSWD:ALL' > /etc/sudoers.d/ubuntu \
+ && chmod 0440 /etc/sudoers.d/ubuntu
```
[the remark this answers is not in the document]
christopherpestano: hallucination, this ubuntu image comes with the user ubuntu by default.

pixeltable/catalog/catalog.py
```diff
@@ -1153,7 +1153,9 @@ def _table_error_counts(self) -> dict[UUID, int]:
         cascade_row_count_excs = sql.func.coalesce(
             update_status['cascade_row_count_stats']['num_excs'].astext.cast(sql.Integer), 0
         )
-        errors = sql.func.coalesce(sql.func.sum(row_count_excs + cascade_row_count_excs), 0).label('errors')
+        errors = sql.func.coalesce(
+            sql.cast(sql.func.sum(row_count_excs + cascade_row_count_excs), sql.Integer), 0
+        ).label('errors')
```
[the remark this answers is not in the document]
christopherpestano: In this specific scenario it is very unlikely we will hit the int_max for row count execs.

tool/ci_tool.py
```diff
@@ -24,6 +24,22 @@
 
 COCKROACH_TEST_MODULES = ('table', 'index')
 
+# CockroachDB suite modules: the `make slimtest` set + the primary-key index tests for partial index coverage.
+# test_unversioned_table.py is intentionally excluded (see TODO: implement for unversioned tables [PXT-1101])
```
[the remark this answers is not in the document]
christopherpestano: It's fine I'm not saying anything incorrect there it is the make slimiest with the unversioned table excluded

tests/share/test_packager.py
```diff
@@ -42,6 +42,52 @@
 )
 
 
+# CockroachDB applies ADD COLUMN asynchronously, so the CREATE INDEX in replica restore can't see the new column yet.
+def _skip_replica_if_cockroachdb() -> None:
```
aaron-siegel: Move to util and rename to `skip_test_if_cockroachdb()`

tests/share/test_packager.py
```diff
@@ -42,6 +42,52 @@
...
+    if Env.get().is_using_cockroachdb:
+        pytest.skip('Replica restore not yet supported on CockroachDB')
+
+
+# Tests that exercise the package() / restore() round trip; these all hit the replica restore code path.
+# Listed here so we can apply the skip up-front instead of partway through each test (which leaves the
+# worker in a half-broken state on failure).
+_REPLICA_TESTS = frozenset(
```
aaron-siegel: This is an awkward way to do this (it effectively dislocates test metadata from the actual tests).

I'd prefer to just put `skip_test_if_cockroachdb()` in each of the affected tests; or, simply omit test_packager.py from the cockroachdb tests.

tests/conftest.py
```diff
@@ -274,6 +274,22 @@ def clean_db(drop_md_tables: bool = False) -> None:
...
+            elif Env.get().is_using_cockroachdb:
+                # Truncate existing metadata tables.
+                # CockroachDB sometimes rejects TRUNCATE when other in-flight statements are
+                # dropping indexes on the same table; fall back to ordered DELETEs that respect FKs.
+                # Delete child tables first, then parents. dirs has a self-referential FK
+                # so we null out parent_id before deleting.
+                fk_order = ['pendingtableops', 'tableversions', 'tableschemaversions', 'functions', 'tables']
+                for t in fk_order:
+                    if t in existing_md_names:
+                        conn.execute(text(f'DELETE FROM "{t}"'))
```
aaron-siegel: Can idiomatic SQLAlchemy be used here?

tests/conftest.py
```diff
@@ -274,6 +274,22 @@ def clean_db(drop_md_tables: bool = False) -> None:
...
+                if 'dirs' in existing_md_names:
+                    conn.execute(text('UPDATE "dirs" SET parent_id = NULL WHERE parent_id IS NOT NULL'))
```
aaron-siegel: Likewise

tests/test_env.py
```diff
@@ -15,11 +17,17 @@ def _reset_env(reinit: bool, db_name: str) -> None:
     Env._init_env(reinit_db=reinit)
 
 
+def _skip_if_non_local() -> None:
```
aaron-siegel: Move to util and rename `skip_test_if_nonlocal()`?

tests/test_primary_key_index.py
```diff
@@ -86,6 +89,8 @@ def test_batch_with_duplicate_fails_atomically(self, uses_db: None) -> None:
 
     def test_pk_index_row_too_large(self, uses_db: None) -> None:
         """Many PK columns can exceed the btree max row size; error message should be user-friendly."""
+        if Env.get().is_using_cockroachdb:
```
aaron-siegel: Use the util method skip_test_if_cockroachdb() (it could take an optional `message` parameter if you want to keep that customization)

tool/ci_tool.py
```diff
@@ -135,6 +151,22 @@ def generate_matrix(args: argparse.Namespace) -> None:
                 for module in COCKROACH_TEST_MODULES
             )
 
+        # Full test suite against a local single-node CockroachDB container. Restricted to the slimtest modules.
+        configs.append(
+            MatrixConfig(
+                'cockroach-full',
```
aaron-siegel: Maybe call this `cockroach-local`

tool/ci_tool.py
aaron-siegel: Maybe rename this to `cockroach-live-{module}` so it's clear.

tool/ci_tool.py
```diff
@@ -24,6 +24,22 @@
 
 COCKROACH_TEST_MODULES = ('table', 'index')
```
aaron-siegel: Maybe call this `COCKROACH_LIVE_TEST_MODULES` to make it clear it's only relevant to the live tests

tool/ci_tool.py
```diff
@@ -24,6 +24,22 @@
...
 
+# CockroachDB suite modules: the `make slimtest` set + the primary-key index tests for partial index coverage.
+# test_unversioned_table.py is intentionally excluded (see TODO: implement for unversioned tables [PXT-1101])
+SLIMTEST_MODULES = (
```
aaron-siegel: `COCKROACH_LOCAL_TEST_MODULES`? It's bothering me a little that these are specified in a different format from the preceding ones - maybe try to unify them (in one direction or the other)

tool/ci_tool.py
```diff
@@ -135,6 +151,22 @@ def generate_matrix(args: argparse.Namespace) -> None:
...
 
+        # Full test suite against a local single-node CockroachDB container. Restricted to the slimtest modules.
+        configs.append(
+            MatrixConfig(
+                'cockroach-full',
+                'py',
+                MAIN_PLATFORM,
+                '3.10',
+                pytest_options=f'{DEFAULT_PYTEST} {" ".join(SLIMTEST_MODULES)}',
+                pre_test_cmd=(
```
aaron-siegel, line 167: Nice, I like the way this is set up.

pyproject.toml
```diff
@@ -99,6 +99,9 @@ override-dependencies = [
     "ruff>=0.14",  # Required because 'fireworks-ai' hard-codes 'ruff<10'
     "protobuf>=5.29.6",  # Required because 'fireworks-ai' hard-codes 'protobuf==5.29.4'
     "huggingface-hub>=1",  # Required due to 'whisperx'
+    # torchcodec has no aarch64 wheels until 0.11, but our pin is <0.8 (torch 2.8 compat).
+    # Override pyannote-audio's unconditional torchcodec dep to skip linux-aarch64.
+    "torchcodec<0.8 ; python_version < '3.14' and (sys_platform != 'linux' or platform_machine != 'aarch64')",
```
aaron-siegel: FYI this is changing in PR #1385 which upgrades torchcodec to 0.13

aaron-siegel (changes requested):

tests/share/test_packager.py
christopherpestano: Done, added skip_test_if_cockroachdb() to tests/utils.py next to the other skip_test_if_* helpers. It takes an optional reason string so call sites can keep their specific messages.

tests/share/test_packager.py
christopherpestano: Agreed, removed the fixture. Each replica test now calls skip_test_if_cockroachdb() directly. Verified against a live cockroach container, 5 passed 26 skipped.

tests/conftest.py
christopherpestano: Yep, now iterates reversed(base_metadata.sorted_tables) so sqlalchemy computes the FK delete order.

tests/conftest.py
christopherpestano: Same fix covers this. The parent_id nulling turned out unnecessary anyway since FKs are checked at end of statement, verified live.

tests/test_env.py
christopherpestano: Done, added it as skip_test_if_not_local() in tests/utils.py, also with an optional reason string, and dropped the local helper here.

tests/test_primary_key_index.py
christopherpestano: Done, switched to skip_test_if_cockroachdb() with the cockroach specific reason passed as the optional message.

tool/ci_tool.py
christopherpestano: Done, renamed the job to cockroach-local so it lines up with the cockroach-live-* jobs.

tool/ci_tool.py
christopherpestano: Done, renamed to cockroach-live-{module}, and the module short name is now derived from the test file path.

tool/ci_tool.py
christopherpestano: Done, renamed to COCKROACH_LIVE_TEST_MODULES and the local list to COCKROACH_LOCAL_TEST_MODULES to match.

tool/ci_tool.py
christopherpestano: Done, renamed, and both lists now use full test file paths so they read the same way.

tool/ci_tool.py
christopherpestano: Thanks!

pyproject.toml
christopherpestano: Good to know

aaron-siegel (approved):

PR #1308, opened 2026-05-03
Dashboard performance and correctness fixes

mkornacker:
Performance

- /api/dirs: ~17.5s -> ~45ms on a 155-table catalog. Sidebar tree no longer instantiates Tables (which triggers transformer model loads); reads kind/version straight from the catalog.
- /api/pipeline gained a path query parameter that scopes to the connected component. TableDetailView's per-table lineage fetch went from ~20s to ~85ms.
- Per-table error counts come from a single GROUP BY over tableversions JSONB, not a per-table sweep.
- Data-table sort is restricted to B-tree-indexed columns. Avoids unbounded sorts on Bool/JSON/unstored columns.

Correctness

- Views with unstored or iterator-output columns no longer break the data tab. The bridge skips unstored columns from value/error fetches and from sort, surfacing them as a muted placeholder in the UI.
- The pipeline endpoint no longer falls into the error-stub branch on tables with unstored computed columns.

API

- New pxt.get_dir_tree() with TreeNode TypedDicts (DirectoryNode | TableNode).
- TableMetadata gained id; IndexMetadata covers B-tree indexes too; ColumnMetadata via the bridge gains is_stored / is_sorted.
- get_pipeline(tbl_path) for scoped lookups; error counts removed from its response.

Frontend

- Sidebar/main and schema/data dividers are draggable (react-resizable-panels), with persistence. Schema panel auto-fits on first table visit, then remembers per-table.
- Schema column widths are drag-resizable with min-width clamps and double-click reset.
- Long expressions open a click-to-expand modal with Python syntax highlighting.
- All error-count UI is consolidated to the sidebar tree.
- Default B-tree indexes are filtered out of the schema-panel index list.

### DIFF
--- a/docs/public_api.opml
+++ b/docs/public_api.opml
@@ -20,6 +20,7 @@
           <outline text="func|pixeltable.create_dir" />
           <outline text="func|pixeltable.drop_dir" />
           <outline text="func|pixeltable.get_dir_contents" />
+          <outline text="func|pixeltable.get_dir_tree" />
           <outline text="func|pixeltable.home" />
           <outline text="func|pixeltable.ls" />
           <outline text="func|pixeltable.move" />
--- a/docs/public_api.opml
+++ b/docs/public_api.opml
@@ -126,8 +126,11 @@
 
           <outline text="class|pixeltable.ColumnMetadata" />
           <outline text="class|pixeltable.DirContents" />
+          <outline text="class|pixeltable.DirectoryNode" />
           <outline text="class|pixeltable.IndexMetadata" />
           <outline text="class|pixeltable.TableMetadata" />
+          <outline text="class|pixeltable.TableNode" />
+          <outline text="class|pixeltable.TreeNode" />
           <outline text="class|pixeltable.UpdateStatus" />
           <outline text="class|pixeltable.VersionMetadata" />
         </outline>
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -1068,7 +1068,7 @@ class Catalog:
         dir_entries: dict[str, Catalog.DirEntry]
         table: schema.Table | None
 
-        # Only populated for table entries when get_dir_contents() was called with error_counts=True;
+        # Only populated for table entries when get_dir_contents() was called with with_error_counts=True;
         # None otherwise (including for directory entries).
         table_error_count: int | None = None
 
--- a/pixeltable/catalog/catalog.py
+++ b/pixeltable/catalog/catalog.py
@@ -1041,18 +1041,44 @@ class Catalog:
             dir_id = dir.parent_id
         return Path.parse('/'.join(names), allow_empty_path=True, allow_system_path=True)
 
+    def _table_error_counts(self) -> dict[UUID, int]:
+        """Returns map from table id to the sum of tableversions.num_excs"""
+
+        # JSONB path mirrors UpdateStatus.row_count_stats / cascade_row_count_stats -> RowCountStats.num_excs
+        # (see pixeltable/catalog/update_status.py). If those dataclass fields are renamed, this query
+        # silently returns 0.
+        stmt = sql.text("""
+            SELECT tbl_id, COALESCE(SUM(
+                COALESCE((md -> 'update_status' -> 'row_count_stats' ->> 'num_excs')::int, 0)
+              + COALESCE((md -> 'update_status' -> 'cascade_row_count_stats' ->> 'num_excs')::int, 0)
+            ), 0) AS errors
+            FROM tableversions
+            GROUP BY tbl_id
+        """)
+        rows = get_runtime().conn.execute(stmt).all()
+        return {r.tbl_id: r.errors for r in rows}
+
     @dataclasses.dataclass
     class DirEntry:
         dir: schema.Dir | None
         dir_entries: dict[str, Catalog.DirEntry]
         table: schema.Table | None
 
-    @retry_loop(for_write=False)
-    def get_dir_contents(self, dir_path: Path, recursive: bool = False) -> dict[str, DirEntry]:
-        dir = self._get_schema_object(dir_path, expected=Dir, raise_if_not_exists=True)
-        return self._get_dir_contents(dir._id, recursive=recursive)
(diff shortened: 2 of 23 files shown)


docs/public_api.opml
```diff
@@ -125,8 +126,11 @@
 
           <outline text="class|pixeltable.ColumnMetadata" />
           <outline text="class|pixeltable.DirContents" />
+          <outline text="class|pixeltable.DirectoryNode" />
           <outline text="class|pixeltable.IndexMetadata" />
           <outline text="class|pixeltable.TableMetadata" />
+          <outline text="class|pixeltable.TableNode" />
```
aaron-siegel, line 132: These are public facing classes?

pixeltable/catalog/catalog.py
```diff
@@ -1041,18 +1041,48 @@ def get_dir_path(self, dir_id: UUID) -> Path:
...
+        """Returns map from table id to the sum of num_excs across that table's versions.
+
+        Sums tableversions.md -> update_status -> row_count_stats / cascade_row_count_stats -> num_excs
+        for each row, grouped by tbl_id.
+        """
+
+        # JSONB path mirrors UpdateStatus.row_count_stats / cascade_row_count_stats -> RowCountStats.num_excs
+        # (see pixeltable/catalog/update_status.py). If those dataclass fields are renamed, this query
+        # silently returns 0.
+        stmt = sql.text("""
```
aaron-siegel: I feel like `sql.text()` is a crutch here and it's usually better to use idiomatic SQLAlchemy, which definitely seems possible here - there's nothing especially exotic happening. Maybe ask Claude if it can convert the SQL text to idiomatic SQLAlchemy?

pixeltable/catalog/catalog.py
```diff
@@ -1041,18 +1041,48 @@ def get_dir_path(self, dir_id: UUID) -> Path:
...
+            SELECT tbl_id, COALESCE(SUM(
+                COALESCE((md -> 'update_status' -> 'row_count_stats' ->> 'num_excs')::int, 0)
```
aaron-siegel: Even if you're using `sql.text()`, the `->` and `->>` notation is outdated - bracket notation has been preferred since Postgres 14 (we're on 16).
```
md['update_status']['row_count_stats']['num_excs']::int
```
etc.

pixeltable/globals.py
```diff
@@ -752,6 +753,91 @@ def _assemble_dir_contents(
...
+        A list of [`TreeNode`][pixeltable.TreeNode] dicts. Each node is either a `DirectoryNode` or a `TableNode`.
+    """
+    path_obj = catalog.Path.parse('', allow_empty_path=True)
+    catalog_entries = get_runtime().catalog.get_dir_contents(path_obj, recursive=True, with_error_counts=True)
+    path_by_id: dict[UUID, str] = {}
+    _create_path_map('', catalog_entries, path_by_id)
+    return _get_subtree('', catalog_entries, path_by_id)
+
+
+def _create_path_map(dir_path: str, catalog_entries: dict[str, Catalog.DirEntry], path_map: dict[UUID, str]) -> None:
```
aaron-siegel, line 769: This makes `globals` a bit crowded; maybe we need a new `dirtree` module or some such?

pixeltable/globals.py
```diff
@@ -1112,6 +1198,34 @@ class DirContents(TypedDict):
...
+class DirectoryNode(TypedDict):
+    """A directory entry in a [`TreeNode`][pixeltable.TreeNode] tree."""
+
+    name: str
+    path: str
+    kind: Literal['directory']
+    entries: list['TreeNode']
+
+
+TableKind = Literal['table', 'view', 'snapshot', 'replica']
```
aaron-siegel: There are some other places where this could be useful. Move to `types.py`?

pixeltable/globals.py
```diff
@@ -1112,6 +1198,34 @@ class DirContents(TypedDict):
     """List of table paths contained in this directory."""
 
 
+class DirectoryNode(TypedDict):
```
aaron-siegel: If we create a `dirtree` module or equivalent, move these there? If not, to `types.py`? (See next comment)

tests/share/test_packager.py
```diff
@@ -235,9 +235,17 @@ def __check_table(self, bundle_info: BundleInfo, tbl_name: str) -> None:
 
         # Certain metadata properties must be identical.
         metadata = t.get_metadata()
-        for property in ('indices', 'version', 'version_created', 'schema_version', 'comment', 'media_validation'):
+        for property in ('version', 'version_created', 'schema_version', 'comment', 'media_validation'):
             assert metadata[property] == bundle_info.metadata[property]
 
+        # Compare embedding indices only: btree indices aren't surfaced in a snapshot's
+        # `idxs_by_name`, but they are in a live replica's, producing a benign asymmetry
```
aaron-siegel: I find this counterintuitive. A replica is a kind of snapshot; why shouldn't it have the same properties as a snapshot?

Similarly, what happens if I do
```
t = get_table('my_tbl:784')
```
to get an anonymous snapshot. Do the btree indices show up in `t.get_metadata()`?

It'd be very strange to see btree indices in anonymous snapshots but not named snapshots. And it'd be equally strange (in my opinion) to see them in replicas but not anonymous snapshots or vice versa.

aaron-siegel (approved):

docs/public_api.opml
mkornacker: So far, yes, via get_dir_tree(). Should we hide them?

pixeltable/globals.py
mkornacker: Just for the new get_dir_tree() and related, or anything else?

tests/share/test_packager.py
mkornacker: Done.

pixeltable/globals.py

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/globals.py
mkornacker: Done.

pixeltable/globals.py

PR #1312, opened 2026-05-05
[PXT-1135] Nano Banana image generation support

christopherpestano:
After discussion with the rest of the developer team, this PR is a compromise on PXT-1135: it makes Nano Banana image generation easier to use without introducing an additional abstraction over the Gemini SDK. The original ticket proposed routing Nano Banana models through generate_images so that swapping from Imagen would be a one-line model name change, but we opted not to add that layer.

Inline image bytes returned by generate_content are now decoded into PIL images automatically, so the generated image can be extracted using JSON subscripting (response.candidates[0].content.parts[0].inline_data.data) without writing a custom UDF to base64-decode and pad-correct the data.

generate_images now raises a clear error when called with a Gemini image-generation model name and points the caller at generate_content with an example. The external doc link in the generate_images docstring is also corrected to the Imagen page.


pixeltable/functions/gemini.py
```diff
@@ -154,7 +172,17 @@ async def generate_content(
...
+        result = response.model_dump(mode='json')
+        for candidate, result_candidate in zip(response.candidates or [], result.get('candidates', [])):
+            if candidate.content is None:
+                continue
+            for part, result_part in zip(
+                candidate.content.parts or [], result_candidate.get('content', {}).get('parts', [])
+            ):
+                blob = part.inline_data
+                if blob is not None and blob.mime_type and blob.mime_type.startswith('image/') and blob.data:
+                    result_part['inline_data']['data'] = PIL.Image.open(io.BytesIO(blob.data))
```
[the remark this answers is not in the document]
christopherpestano: That seems expensive if the user gets a response with multiple images and they only extract one reject

pixeltable/functions/gemini.py
```diff
@@ -256,6 +287,14 @@ async def generate_images(prompt: str, *, model: str, config: dict | None = None
         ...     response=generate_images(tbl.prompt, model='imagen-4.0-generate-001')
         ... )
     """
+    if model.startswith('gemini-'):
+        raise excs.RequestError(
+            excs.ErrorCode.INVALID_ARGUMENT,
+            f'Model {model!r} is a Gemini image-generation model (Nano Banana) and must be used via '
+            f'`generate_content`, not `generate_images` (which is for Imagen models). See the '
+            f'`generate_content` docstring for an example.',
+        )
```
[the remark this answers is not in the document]
christopherpestano: Huh who would pass a text only model to generate_image? This doesn't make sense at all, regardless the user should be instead using generate_content

pixeltable/functions/gemini.py
aaron-siegel: I believe copilot is correct here; failure to call `img.load()` could result in the image handle being disposed of  before it gets read, leading to strange bugs at an undefined point in the future. (I've been bitten by this one before.)

pixeltable/functions/gemini.py
```diff
@@ -154,7 +172,17 @@ async def generate_content(
     async with _gemini_file_uploads(large_video_paths) as uploaded:
         contents = _replace_upload_placeholders(contents, uploaded)
         response = await client.aio.models.generate_content(model=model, contents=contents, config=config_)
-        return response.model_dump(mode='json')
+        result = response.model_dump(mode='json')
+        for candidate, result_candidate in zip(response.candidates or [], result.get('candidates', [])):
```
aaron-siegel: use zip(..., strict=True)? Since this is what we expect?

Why do we need to zip the original Pydantic model against the dict? That wasn't clear to me. Add a code comment to explain?

aaron-siegel (changes requested):

pixeltable/functions/gemini.py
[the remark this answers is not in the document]
christopherpestano: Added section in test_generate_content_nano_banana that validates that this raises

pixeltable/functions/gemini.py
christopherpestano: Interesting I was not aware of that being possible, added load to prevent this.

pixeltable/functions/gemini.py
christopherpestano: The difference between here and openai is that openai gives you the b64 string as an actual string, whereas gemini's genai will decode the b64 into actual bytes before giving you the model. This means that when you attempt to round trip naively it will raise since you have to account for how Pydantic serializes bytes by default.  That is url safe b64 which replaces some characters. You can parse that string with urlsafe_b64decode which while it works, is very possible to break if pydantic ever decides to change their default behavior. I believe the better choice is to directly pull the bytes from the original model which is faster (by a small margin since you don't have to parse anything) and more "correct" to me.

christopherpestano: The zip strict thing was a miss by me though, fixed to be True

aaron-siegel (approved):

pixeltable/functions/gemini.py
christopherpestano: Changed to instead use python dumps format instead of json dumps format

PR #1313, opened 2026-05-05
Fixing LC_CTYPE to use legacy locale on MacOS

christopherpestano:
A bug was introduced in https://github.com/pixeltable/pixeltable/pull/1286, in which we assume that all MacOS systems have C.UTF-8 available as a locale option for Postgres. This was mistaken, as C.UTF-8 was a recent addition to MacOS, and the legacy locale available since MacOS 12+ is en_US.UTF-8, which is required regardless of system language.

### DIFF
--- a/pixeltable/utils/dbms.py
+++ b/pixeltable/utils/dbms.py
@@ -47,12 +47,13 @@ class PostgresqlDbms(Dbms):
         return f'DROP DATABASE {database}'
 
     def create_db_stmt(self, database: str) -> str:
-        if platform.system() == 'Windows':
-            lc_ctype = '.UTF-8'
-        elif platform.system() == 'Darwin':
-            lc_ctype = 'en_US.UTF-8'
-        else:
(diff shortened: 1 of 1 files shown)


pixeltable/utils/dbms.py
```diff
@@ -47,7 +47,12 @@ def drop_db_stmt(self, database: str) -> str:
         return f'DROP DATABASE {database}'
 
     def create_db_stmt(self, database: str) -> str:
-        lc_ctype = '.UTF-8' if platform.system() == 'Windows' else 'C.UTF-8'
+        if platform.system() == 'Windows':
```
aaron-siegel: I might use a matcher:
```python
match platform.system() {
    case 'Windows':
        ...
}
```

aaron-siegel (approved):

pixeltable/utils/dbms.py
christopherpestano: Sure that works

PR #1318, opened 2026-05-06
Docs & README refresh: serving integration, dependency fixes, style improvements

pierrebrunelle:
Lots of changes...


docs/release/howto/deployment/infrastructure.mdx
```diff
@@ -146,7 +131,7 @@ def search_documents(query_text: str, limit: int = 5):
 from pydantic import BaseModel
 from fastapi import FastAPI
 import pixeltable as pxt
-from setup_pixeltable import search_documents
+import schema  # noqa: F401 — import triggers table creation
```
aaron-siegel: I thought `schema.py` was set up as a script, not a module. It seems to be serving a dual purpose here (which is what's causing the issues)

Why not declare the `@query` here instead so we don't have to import `schema`?

If you really want to import `schema.py`, use an `if __name__ == "__main__":` guard to make sure the script code doesn't run when it's imported.

docs/release/integrations/cloud-storage.mdx
```diff
@@ -96,6 +99,56 @@ Destinations are resolved in this order:
...
+
+    t = pxt.create_table('app/images', {'photo': pxt.Image})
+
+    t.add_computed_column(
+        thumbnail=t.photo.resize((256, 256)),
+        destination='pxtfs://myorg:mydb/home'
+    )
+    ```
+
+    Or set it as your global default in `config.toml`:
```
aaron-siegel, line 140: Or as an environment variable `PIXELTABLE_OUTPUT_MEDIA_DEST`

docs/release/migrate/from-hand-written-endpoints.mdx
```diff
@@ -0,0 +1,258 @@
...
+        results = t.select(t.title, t.document).collect()
+        return DocumentsResponse(
+            items=[DocumentItem(**row) for row in results.to_dicts()]
+        )
+    ```
+  </Tab>
+  <Tab title="FastAPIRouter">
+    ```python
+    from pixeltable.serving import FastAPIRouter
+    import schema  # import triggers table creation
```
aaron-siegel: Same comment as above - an `import` with this type of side effect is not generally desirable.

aaron-siegel (approved):

docs/release/howto/deployment/infrastructure.mdx
pierrebrunelle: I'm having conflicted thoughts and also decisions here with the new fastapi support that's why, I'll let you review with @mkornacker:
- https://github.com/pixeltable/pixeltable-starter-kit/blob/main/backend/routers/data.py alongside https://github.com/pixeltable/pixeltable-starter-kit/blob/main/backend/setup_pixeltable.py as an example and yeah decide if import schema(or whatever it's called) is not necessary cause I declare the @query in the schema-defined setup which therefore needs to be import as a module for the fastapi to work?

PR #1322, opened 2026-05-07
Thread concurrency-related bug fixes

mkornacker:
- Validate that Table public methods are called on the constructing thread; raise on cross-thread use
- Replace cached Column in ColumnRef with ColumnHandle indirection (re-resolves through catalog on every access; stale references after drop/re-add now raise instead of silently returning stale metadata)                                                                                                      
- Track origin thread + origin catalog on TableVersionHandle and TableVersionPath, with asserts to catch cross-thread/cross-catalog leaks; deepcopy resets thread-specific state                                                                                                                             
- Query defensively deepcopies on cross-catalog use; grouping table stored as a handle
- Add Runtime.context_inherited (set by copy_db_context) so ExecNode._thread_iter worker threads are sanctioned delegates and pass the origin-thread asserts                                                                                                                                                  
- Refactor object-store writes: caller thread resolves FileDestination from catalog, worker threads perform I/O without catalog access (new put_file_resolved/copy_local_file/move_local_file path)                                                                                                  
- ObjectOps.put_file and LocalStore/TempStore.save_media_object take primitives (tbl_id, col_id, tbl_version, col_name) instead of Column, decoupling the storage layer from catalog internals

### DIFF
--- a/pixeltable/_query.py
+++ b/pixeltable/_query.py
@@ -616,6 +616,9 @@ class Query:
         return len(self._from_clause.join_clauses) > 0
 
     def show(self, n: int = 20) -> ResultSet:
+        # See _output_row_iterator for the rationale on the defensive copy.
+        if self._origin_catalog is not get_runtime().catalog:
+            return copy.deepcopy(self).show(n)
         if self.sample_clause is not None:
             raise excs.RequestError(excs.ErrorCode.UNSUPPORTED_OPERATION, 'show() cannot be used with sample()')
         assert n is not None
@@ -636,6 +639,9 @@ class Query:
             Error: If the Query is the result of a join or
                 if the Query has an order_by clause.
         """
+        # See _output_row_iterator for the rationale on the defensive copy.
+        if self._origin_catalog is not get_runtime().catalog:
+            return copy.deepcopy(self).head(n)
         if self.order_by_clause is not None:
             raise excs.RequestError(excs.ErrorCode.UNSUPPORTED_OPERATION, 'head() cannot be used with order_by()')
         if self._has_joins():
@@ -663,6 +669,9 @@ class Query:
             Error: If the Query is the result of a join or
                 if the Query has an order_by clause.
         """
+        # See _output_row_iterator for the rationale on the defensive copy.
+        if self._origin_catalog is not get_runtime().catalog:
+            return copy.deepcopy(self).tail(n)
         if self.order_by_clause is not None:
             raise excs.RequestError(excs.ErrorCode.UNSUPPORTED_OPERATION, 'tail() cannot be used with order_by()')
         if self._has_joins():
@@ -860,6 +869,9 @@ class Query:
         Returns:
             The number of rows in the Query.
         """
+        # See _output_row_iterator for the rationale on the defensive copy.
+        if self._origin_catalog is not get_runtime().catalog:
+            return copy.deepcopy(self).count()
         if self.limit_val is not None and self.limit_val.val == 0:
             return 0
         with get_runtime().catalog.begin_xact(read_tbl_ids=self.referenced_tbl_ids()) as conn:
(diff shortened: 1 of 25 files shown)


tests/test_concurrent.py
```diff
@@ -52,3 +73,252 @@ def worker(tid: int) -> None:
             assert row['offset'] == row['value'] + 100
 
         assert all(c >= 0 for c in select_counts)
+
+    NUM_THREADS = 4
+    ITERATIONS = 20
+
+    def test_shared_query(self, uses_db: None) -> None:
```
aaron-siegel, line 80: Will these scenarios also be covered by random_ops?

pixeltable/_query.py
```diff
@@ -821,6 +860,8 @@ def count(self) -> int:
         Returns:
             The number of rows in the Query.
         """
+        if self._origin_catalog is not get_runtime().catalog:
```
aaron-siegel, line 863: Why not use `self = self._rebind()` here?

Incidentally I actually prefer this pattern to "self-munging". It's 2 lines of code instead of 1 but makes the intent a little clearer (and preserves the sanctity of "self").
```python
def my_op(...):
    if not self._is_thread_local():
        self._rebase().my_op(...)
```

pixeltable/exprs/column_ref.py
```diff
@@ -106,6 +107,11 @@ def __init__(
             self.components = [non_validating_col_ref]
         self.id = self._create_id()
 
+    @property
+    def col(self) -> catalog.Column:
+        # re-resolve via the current Catalog when we need the actual Column
```
aaron-siegel: clarify that this is in order to ensure the ColumnRef is thread-safe (I'm aware that's mentioned in the above comment but useful to put it here too: "(in order to ensure ColumnRef is thread-safe)"

pixeltable/catalog/table_version_handle.py
```diff
@@ -13,22 +14,39 @@
 if TYPE_CHECKING:
     from pixeltable.catalog import Column
 
+    from .catalog import Catalog
```
aaron-siegel: Merge with previous import

pixeltable/catalog/table.py
```diff
@@ -208,6 +231,17 @@ def __hash__(self) -> int:
...
     def __getattr__(self, name: str) -> 'exprs.ColumnRef':
         """Return a ColumnRef for the given name."""
+        # __getattr__ is called when normal attribute lookup fails. Avoid recursion by reading
+        # _origin_thread_id from the instance dict directly. If it's missing (Table not fully
+        # constructed yet), skip the check.
+        origin = self.__dict__.get('_origin_thread_id')
+        if origin is not None and origin != threading.get_ident():
+            raise excs.RequestError(
+                excs.ErrorCode.INVALID_STATE,
+                f'Table {self.__dict__.get("_name")!r} was accessed from a thread other than the one that '
```
aaron-siegel: Why `self.__dict__.get("_name")` instead of `self._name`? The only advantage of the former is that it will return `None` if `_name` is not defined, right? But then we'd end up with an odd error message. If that's an important scenario then best to put in a default (`get("_name", "unknown")` or some such?)

aaron-siegel (approved):

tests/test_concurrent.py
mkornacker: They should be, eventually.

pixeltable/_query.py
mkornacker: I'll use _rebind().

pixeltable/exprs/column_ref.py
mkornacker: Clarified that it's *not* thread-safe.

pixeltable/catalog/table_version_handle.py
mkornacker: Done.

pixeltable/catalog/table.py

PR #1323, opened 2026-05-07
Get rid of later Column value expr initialization

sergey-mkhitaryan:
The process of initializing a `Column` from metadata today is 2-step: first create all `Column` instances with a `value_expr_dict`, then, after initializing indexes and other stuff, call `init_value_expr` on each `Column` to finish the initialization of its value expression. That 2nd step includes retargeting of `ColumnRefs`.

This change folds everything in a single step in which `TableVersion` first fully prepares a value expr for each column, and only then creates a `Column` instance with it. Due to dependencies between columns and indexes, this is only possible if columns and indexes are initialized in lock-step.

The broader context of this change is that it will make it easier to fix retargeting of `ColumnRefs` which is a blocker to my schema evolution work [PXT-969](https://pixeltable.atlassian.net/browse/PXT-969).

[PXT-969]: https://pixeltable.atlassian.net/browse/PXT-969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pixeltable/catalog/table_version.py
```diff
@@ -569,14 +518,108 @@ def _init_cols_from_md(self, indexes: list[tuple[schema.IndexMd, index.IndexBase
...
+        if not value_expr.is_valid:
+            col_name = schema_col_md.name if schema_col_md is not None else '<unnamed>'
+            message = '\n'.join(
+                [
+                    f'The computed column {col_name!r} in table {self.name!r} is no longer valid.',
+                    value_expr.validation_error,
+                    'You can continue to query existing data from this column, but evaluating it on new data will raise an error.',  # noqa: E501
+                ]
+            )
+            warnings.warn(message, category=excs.PixeltableWarning)  # noqa: B028
```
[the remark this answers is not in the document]
sergey-mkhitaryan: Depending on the codepath that leads here, the stack distance from the user code can be different. There isn't a reasonable `stacklevel` that we could use that works in all cases.

pixeltable/catalog/table_version.py
```diff
@@ -576,14 +526,110 @@ def _init_cols_from_md(self, indexes: list[tuple[schema.IndexMd, index.IndexBase
...
```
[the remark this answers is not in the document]
sergey-mkhitaryan: Depending on the codepath that leads here, the stack distance from the user code can be different. There isn't a reasonable stacklevel that we could use that works in all cases.

pixeltable/catalog/table_version.py
```diff
@@ -576,14 +526,110 @@ def _init_cols_from_md(self, indexes: list[tuple[schema.IndexMd, index.IndexBase
                 schema_version_add=col_md.schema_version_add,
                 schema_version_drop=col_md.schema_version_drop,
                 stores_cellmd=col_md.stores_cellmd,
+                computed_with=value_expr,
                 value_expr_dict=col_md.value_expr,
                 tbl_handle=self.handle,
```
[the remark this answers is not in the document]
sergey-mkhitaryan: Left a TODO to get rid of `value_expr_dict`, which is not hard but also not exactly simple enough to do it here.

pixeltable/catalog/table_version.py
```diff
@@ -525,27 +479,21 @@ def _init_schema(self) -> None:
...
-            col.init_value_expr(tvp)
 
-        # create the sqlalchemy schema, after instantiating all Columns
-        if self.is_component_view:
-            self.store_tbl = StoreComponentView(self)
-        elif self.is_view:
-            self.store_tbl = StoreView(self)
-        else:
-            self.store_tbl = StoreTable(self)
+        assert len(self.cols) == 0
```
mkornacker: This seems gratuitous.

pixeltable/catalog/table_version.py
```diff
@@ -576,14 +526,110 @@ def _init_cols_from_md(self, indexes: list[tuple[schema.IndexMd, index.IndexBase
...
+                    value_expr.validation_error,
+                    'You can continue to query existing data from this column, but evaluating it on new data will raise an error.',  # noqa: E501
+                ]
+            )
+            warnings.warn(message, category=excs.PixeltableWarning)  # noqa: B028
+        if tvp is not None:
+            value_expr = value_expr.retarget(tvp)
+        return value_expr
+
+    def _build_idx_trigger_map(
```
mkornacker: Trigger map? Let's stick with standard terminology.

pixeltable/catalog/table_version.py
```diff
@@ -576,14 +526,110 @@ def _init_cols_from_md(self, indexes: list[tuple[schema.IndexMd, index.IndexBase
...
+        self,
+    ) -> tuple[dict[int, index.IndexBase], dict[int, list[tuple[index.IndexBase, schema.IndexMd]]]]:
+        """Build lookup structures needed to initialize indexes from metadata in the column loop.
+
+        Returns:
+            idx_col_to_idx: maps both index value and undo column ids -> IndexBase
+            col_id_trigger_to_idx: maps the highest-dependency column id for each visible index to that index,
```
mkornacker: What is a highest-dependency column?

'visible index'? (aren't they either all visible or not?)

pixeltable/catalog/table_version.py
```diff
@@ -576,14 +526,110 @@ def _init_cols_from_md(self, indexes: list[tuple[schema.IndexMd, index.IndexBase
...
+                so _init_schema can initialize each index as soon as all its dependency columns are ready.
```
mkornacker: This is commenting on the caller logic, generally not a good idea.

pixeltable/catalog/table_version.py
```diff
@@ -576,14 +526,110 @@ def _init_cols_from_md(self, indexes: list[tuple[schema.IndexMd, index.IndexBase
...
-            cols.append(col)
-        return cols
+
+            self.cols.append(col)
+            if is_visible:
+                if col.name is not None:
+                    self.cols_by_name[col.name] = col
+                self.cols_by_id[col.id] = col
+
+            # Finally initialize indexes that this column unlocks
```
mkornacker: Unlocks?

Why not do that in a separate phase? What's the purpose of interleaving it?

pixeltable/catalog/table_version.py
```diff
@@ -576,14 +526,110 @@ def _init_cols_from_md(self, indexes: list[tuple[schema.IndexMd, index.IndexBase
...
+
+        # create the sqlalchemy schema, after instantiating all Columns
+        if self.is_component_view:
+            self.store_tbl = StoreComponentView(self)
+        elif self.is_view:
+            self.store_tbl = StoreView(self)
+        else:
+            self.store_tbl = StoreTable(self)
+
+    def _init_col_value_expr_from_md(
```
mkornacker, line 560: Why not have that be a classmethod on Column? It doesn't need self other than for the error message.

mkornacker (changes requested):

pixeltable/catalog/table_version.py
sergey-mkhitaryan: deleted all but 2, moved those remaining to the top. leaving them just in case, they should no longer get in the way when reading the code.

pixeltable/catalog/table_version.py
sergey-mkhitaryan: reworked this stuff, ptal

pixeltable/catalog/table_version.py
sergey-mkhitaryan: yeah, but it seems pretty important that we print table name. or did you mean make it a classmethod that takes table name? in that case not sure what's the point of it.

mkornacker (approved):

PR #1324, opened 2026-05-08
[PXT-1094] Import RDBMS sources into Pixeltable tables

christopherpestano:
Adds pxt.io.import_sql(), which streams rows from any SQLAlchemy Selectable (a Table, a select() statement, or text(...).columns(...)) executed against a user-supplied Engine or Connection into a Pixeltable table. The same call covers create+insert (if_exists='error') and append into an existing table (if_exists='append'). replace is intentionally omitted, because create+insert is not atomic and a failed insert would leave the user with an empty replica of the source schema; callers wanting overwrite semantics should drop_table first and then call import_sql(..., if_exists='error').

All schema-side validation lives in import_sql() itself: unnamed or duplicate source columns, columns whose SQLAlchemy type does not map to a Pixeltable type (and is not resolved via schema_overrides), schema_overrides keys that do not appear in the source, computed-column collisions on append, type incompatibility between source and destination, and missing required destination columns. The reason for doing all of this preflight work in import_sql() is to keep SqlDataNode purely a runtime data path. By the time the plan is built the source is already known to be valid, so the node only has to build a slot mapping from source column names to row-builder slots and stream rows; nothing in the execution path has to re-derive or revalidate the schema.

SqlDataNode lives in its own module (pixeltable/exec/sql_data_node.py) rather than in sql_node.py. The existing nodes in sql_node.py (SqlScanNode, SqlAggregationNode, SqlJoinNode, etc.) all push parts of a Pixeltable query plan down into the embedded Postgres against catalog-managed tables. SqlDataNode does the opposite: it reads from a foreign, user-supplied database and feeds those rows into the insert pipeline. Its output contract (output_exprs = row_builder.input_exprs, populates user-column slots by name, sets unmapped slots to None) mirrors InMemoryDataNode, which is the right analogue for streaming external rows into a table. Keeping it in sql_node.py would have suggested it participates in plan pushdown, which it does not.

Planner.create_insert_plan now takes either rows or a sql_source (exactly one), and selects InMemoryDataNode or SqlDataNode accordingly. InsertableTable.insert_sql_source is the internal entry point used by import_sql() and is not part of the public Table API.


pixeltable/io/sql.py
```diff
@@ -74,3 +76,119 @@ def export_sql(
...
+    inferred_schema: dict[str, Any] = {}
+    for i, sa_col in enumerate(sa_cols):
+        col_name = getattr(sa_col, 'name', None) or getattr(sa_col, 'key', None)
+        if col_name is None:
+            raise excs.RequestError(
+                excs.ErrorCode.INVALID_SCHEMA,
+                f'SQL source has an unnamed output column at position {i}; alias it via '
+                f"`expr.label('name')` so it can be matched to a Pixeltable column.",
+            )
+        nullable = bool(getattr(sa_col, 'nullable', True))
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The bool() wrap was the bug. Now treats None as nullable.

pixeltable/io/sql.py
```diff
@@ -74,3 +76,119 @@ def export_sql(
...
+    callers who want overwrite semantics should `pxt.drop_table(tbl_name)` themselves first and then call
+    `import_sql(..., if_exists='error')`.
+
+    Args:
+        selectable: A SQLAlchemy `Selectable` (a `Table`, a `select()` statement, or `text(...).columns(...)`)
+            describing the source rows.
+        conn: A SQLAlchemy `Engine` or `Connection` to execute `selectable` against.
+        tbl_name: Pixeltable path of the destination table.
+        schema_overrides: Optional per-column overrides applied on top of the inferred schema. Keys are column
+            names; values are Pixeltable `ColumnType` instances.
```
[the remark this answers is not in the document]
christopherpestano: Updated to mention type specs like pxt.Image and pxt.Required[pxt.String].

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,160 @@
...
+            self._owns_conn = True
+        else:
+            self._conn = self.sql_source.conn
+            self._owns_conn = False
+
+        # Tables, subqueries, and aliases are not directly executable; wrap them in `select(...)`.
+        src = self.sql_source.selectable
+        stmt: sql.Executable = src if isinstance(src, sql.Executable) else sql.select(src)  # type: ignore[call-overload]
+        self._result = self._conn.execute(stmt)
+
```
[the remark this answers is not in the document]
christopherpestano: Fixed. Close owned conn on execute() failure.

pixeltable/io/sql.py
```diff
@@ -74,3 +76,119 @@ def export_sql(
...
+        inferred_schema[col_name] = pxt_type  # may be None; overridden below
+
+    if schema_overrides is not None:
+        unknown = set(schema_overrides) - set(inferred_schema)
+        if unknown:
+            raise excs.RequestError(
+                excs.ErrorCode.INVALID_ARGUMENT,
+                f'`schema_overrides` references column(s) not in the SQL source: {", ".join(sorted(unknown))}',
+            )
+        inferred_schema.update(schema_overrides)
```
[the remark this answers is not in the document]
christopherpestano: Intentional. Overrides pass through to create_table so the user spec wins.

tests/io/test_sql.py
```diff
@@ -208,3 +211,357 @@ def test_errors(self, uses_db: None) -> None:
...
+        assert schema['c_str'].is_string_type() and not schema['c_str'].nullable
+        assert schema['c_float'].is_float_type() and schema['c_float'].nullable
+        assert schema['c_bool'].is_bool_type() and schema['c_bool'].nullable
+        assert schema['c_ts'].is_timestamp_type() and schema['c_ts'].nullable
+        assert schema['c_date'].is_date_type() and schema['c_date'].nullable
+        assert schema['c_json'].is_json_type() and schema['c_json'].nullable
+        assert schema['c_bytes'].is_binary_type() and schema['c_bytes'].nullable
+
+        # Single-version-bump claim: streaming N=2500 rows still results in version 1
+        assert tbl._tbl_version.get().version == 1
```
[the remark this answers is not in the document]
christopherpestano: Switched to meta[version].

tests/io/test_sql.py
```diff
@@ -208,3 +211,357 @@ def test_errors(self, uses_db: None) -> None:
...
+        schema = tbl._get_schema()
+        assert set(schema) == {'c_int', 'c_str', 'c_float', 'c_bool', 'c_ts', 'c_date', 'c_json', 'c_bytes'}
+        assert schema['c_int'].is_int_type() and not schema['c_int'].nullable
+        assert schema['c_str'].is_string_type() and not schema['c_str'].nullable
+        assert schema['c_float'].is_float_type() and schema['c_float'].nullable
+        assert schema['c_bool'].is_bool_type() and schema['c_bool'].nullable
+        assert schema['c_ts'].is_timestamp_type() and schema['c_ts'].nullable
+        assert schema['c_date'].is_date_type() and schema['c_date'].nullable
+        assert schema['c_json'].is_json_type() and schema['c_json'].nullable
+        assert schema['c_bytes'].is_binary_type() and schema['c_bytes'].nullable
```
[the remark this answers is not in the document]
christopherpestano: Switched to get_metadata()[columns].

pixeltable/io/sql.py
```diff
@@ -74,3 +76,119 @@ def export_sql(
...
+    on_error: Literal['abort', 'ignore'] = 'abort',
+) -> pxt.Table:
+    """Import a SQL source into a Pixeltable table by streaming through a single insert plan.
+
+    `replace` is intentionally not supported. The existing `pxt.create_table(source=..., if_exists='replace')`
+    path has an unsolved create+insert atomicity gap (see the TODO at `pixeltable/globals.py` line 244): if the
+    create succeeds but the subsequent insert fails, the original table has already been destroyed and the user
+    is left with an empty replica named like the destination - silent failure trap. Until that gap is closed,
+    callers who want overwrite semantics should `pxt.drop_table(tbl_name)` themselves first and then call
+    `import_sql(..., if_exists='error')`.
```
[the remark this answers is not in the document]
christopherpestano: Removed. Shortened to two sentences.

pixeltable/exec/sql_data_node.py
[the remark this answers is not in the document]
christopherpestano: Fixed. Close owned conn on execute() failure.

tests/io/test_sql.py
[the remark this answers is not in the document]
christopherpestano: Switched to get_metadata()[columns] and meta[version].

tests/io/test_sql.py
```diff
@@ -208,3 +211,357 @@ def test_errors(self, uses_db: None) -> None:
...
+        # mix of nullable and non-nullable columns; nullable columns include NULL values
+        src = _seed_source(
+            engine,
+            'src_full',
+            [
+                sql.Column('c_int', sql.Integer, nullable=False),
+                sql.Column('c_str', sql.String, nullable=False),
+                sql.Column('c_float', sql.Float, nullable=True),
+                sql.Column('c_bool', sql.Boolean, nullable=True),
+                sql.Column('c_ts', sql.DateTime, nullable=True),
```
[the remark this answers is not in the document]
christopherpestano: Added c_uuid with NULL and roundtrip assertions.

pixeltable/io/sql.py
```diff
@@ -74,3 +76,119 @@ def export_sql(
...
+        inferred_schema,
+        primary_key=primary_key,
+        comment=comment,
+        custom_metadata=custom_metadata,
+        if_exists='error',
+    )
+    try:
+        tbl.insert(sql_source=sql_data_source, on_error=on_error)
+    except BaseException:
+        pxt.drop_table(tbl, if_not_exists='ignore')
```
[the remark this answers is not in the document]
christopherpestano: Added a comment noting __context__ preserves the original via PEP 3134. Not adding suppression.

tests/io/test_sql.py
```diff
@@ -208,3 +211,357 @@ def test_errors(self, uses_db: None) -> None:
...
+
+
+class TestImportSql:
+    """import_sql() tests, parametrized across SQLite (file) and the embedded Postgres."""
+
+    @pytest.mark.parametrize('dialect', _IMPORT_DIALECTS)
+    def test_import_full_table(self, uses_db: None, tmp_path: pathlib.Path, dialect: str) -> None:
+        """End-to-end import of a full SA Table: type inference for all common SA types, nullable vs non-nullable
+        propagation, NULL-to-None value roundtrip, exact value preservation, and the single-version-bump claim
+        across batch boundaries (rows > BATCH_SIZE so streaming actually engages)."""
```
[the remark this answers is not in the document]
christopherpestano: Added value assertion. Tz-stripped since pxt returns tz-aware local.

pixeltable/io/sql.py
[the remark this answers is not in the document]
christopherpestano: Fixed. Now treats None as nullable.

pixeltable/io/sql.py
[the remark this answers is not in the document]
christopherpestano: Updated to mention type specs like pxt.Image and pxt.Required[pxt.String].

pixeltable/io/__init__.py
```diff
@@ -10,9 +10,10 @@
 from .lancedb import export_lancedb
 from .pandas import import_excel, import_pandas
 from .parquet import export_parquet, import_parquet
+from .sql import export_sql, import_sql
 
 __default_dir = {symbol for symbol in dir() if not symbol.startswith('_')}
-__removed_symbols = {'globals', 'hf_datasets', 'pandas', 'parquet', 'datarows', 'lancedb', 'csv', 'json'}
+__removed_symbols = {'globals', 'hf_datasets', 'pandas', 'parquet', 'datarows', 'lancedb', 'csv', 'json', 'sql'}
 __all__ = sorted(__default_dir - __removed_symbols)
```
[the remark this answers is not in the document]
christopherpestano: added

pixeltable/catalog/insertable_table.py
```diff
@@ -151,9 +164,24 @@ def insert(
...
+                raise excs.RequestError(
+                    excs.ErrorCode.INVALID_ARGUMENT,
+                    '`sql_source` is mutually exclusive with `source`, `source_format`, `schema_overrides`, '
+                    'and ad-hoc row keyword arguments.',
+                )
+            return self.insert_sql_source(
+                sql_source=sql_source,
+                fail_on_exception=fail_on_exception,
+                print_stats=print_stats,
+                return_rows=return_rows,
```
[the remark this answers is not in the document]
christopherpestano: Removed

pixeltable/exec/sql_source_node.py
```diff
@@ -0,0 +1,109 @@
...
+            col_ref.col.name: col_ref.slot_idx for col_ref in self.output_exprs if col_ref.col.name is not None
+        }
+        all_output_slot_idxs = {e.slot_idx for e in self.output_exprs}
+
+        sa_cols = selectable_columns(self.sql_source.selectable)
+        source_names = [getattr(c, 'name', None) or getattr(c, 'key', None) for c in sa_cols]
+        # Caller (eg, import_sql) is responsible for validating that all source names are non-None and resolve
+        # to a destination column.
+        self._mapped_slot_idxs = [user_cols_by_name[n] for n in source_names]
+        self._unmapped_slot_idxs = list(all_output_slot_idxs - set(self._mapped_slot_idxs))
```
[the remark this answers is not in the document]
christopherpestano: Removed from public api only import_sql can insert sql.

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(
+        self, tbl: catalog.TableVersionHandle, sql_source: SqlDataSource, row_builder: exprs.RowBuilder
+    ) -> None:
+        output_exprs = list(row_builder.input_exprs)
+        super().__init__(row_builder, output_exprs, [], None)
+        assert tbl.get().is_insertable
+        self.tbl = tbl
+        self.sql_source = sql_source
+        self._owns_conn = False
```
mkornacker: All member vars require a class-level decl

pixeltable/io/data_sources.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+import sqlalchemy as sql
+
+
+@dataclass
+class SqlDataSource:
```
mkornacker, line 10: This is a **very** slender abstraction. Maybe this should just be inlined for now (ie, SqlSourceNode would take two parameters instead of an instance of this).

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,109 @@
...
+class SqlSourceNode(ExecNode):
+    """
+    Streams a SqlDataSource (a SQLAlchemy Selectable executed against an Engine or Connection) into DataRowBatches.
+
+    Same output contract as InMemoryDataNode:
+      - output_exprs = row_builder.input_exprs
+      - populates user-column slots by name; sets unmapped slots to None.
+
+    Source/destination column compatibility (unnamed columns, duplicates, computed-column collisions,
+    missing required columns, unknown destination columns) is validated by the caller (eg, `import_sql`)
```
mkornacker: No need to comment in detail on what the caller is doing, those things can drift.

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,109 @@
...
+_logger = logging.getLogger('pixeltable')
+
+
+class SqlSourceNode(ExecNode):
+    """
+    Streams a SqlDataSource (a SQLAlchemy Selectable executed against an Engine or Connection) into DataRowBatches.
+
+    Same output contract as InMemoryDataNode:
+      - output_exprs = row_builder.input_exprs
+      - populates user-column slots by name; sets unmapped slots to None.
```
mkornacker: What are user columns?

pixeltable/io/sql.py
```diff
@@ -74,3 +77,168 @@ def export_sql(
...
+    tbl_name: str,
+    *,
+    schema_overrides: dict[str, Any] | None = None,
+    primary_key: str | list[str] | None = None,
+    comment: str | None = None,
+    custom_metadata: Any = None,
+    if_exists: Literal['error', 'append'] = 'error',
+    on_error: Literal['abort', 'ignore'] = 'abort',
+) -> pxt.Table:
+    """Import a SQL source into a Pixeltable table by streaming through a single insert plan.
```
mkornacker: No need to comment on the implementation in the public API (the user probably doesn't care about insert plans).

It would be interesting to mention that the data is read in batches (of x rows), because

pixeltable/io/sql.py
```diff
@@ -74,3 +77,168 @@ def export_sql(
...
+
+    `replace` is intentionally not supported, because create+insert is not atomic and a failed insert would
```
mkornacker: Also TMI.

pixeltable/catalog/insertable_table.py
```diff
@@ -208,6 +209,37 @@ def insert_table_data_source(
         FileCache.get().emit_eviction_warnings()
         return status
 
+    def insert_sql_source(
```
sergey-mkhitaryan: If this is for internal use only, let's call it `_insert_sql_source`. There are other functions here that follow the same pattern.

pixeltable/plan.py
```diff
@@ -383,9 +386,15 @@ def create_count_stmt(cls, query: 'pxt.Query') -> sql.Select:
 
     @classmethod
     def create_insert_plan(
-        cls, tbl: catalog.TableVersion, rows: list[dict[str, Any]], ignore_errors: bool
+        cls,
+        tbl: catalog.TableVersion,
+        rows: list[dict[str, Any]] | None,
+        ignore_errors: bool,
+        *,
+        sql_source: SqlDataSource | None,
```
sergey-mkhitaryan: `rows` and `sql_source` should be next to each other

pixeltable/io/data_sources.py
```diff
@@ -0,0 +1,13 @@
...
+    """A user-supplied SQL source: a SQLAlchemy `Selectable` and an `Engine` or `Connection` to run it against."""
+
+    selectable: sql.Selectable
+    conn: sql.Engine | sql.Connection
```
sergey-mkhitaryan: can you write a comment saying this is never a connection to pixeltable store -- it's always some other sql database.

pixeltable/utils/sql.py
```diff
@@ -106,6 +106,62 @@ def get_sa_type(col_type: 'ts.ColumnType', engine: sql.Engine) -> sql.types.Type
...
+    if dialect_resolver is not None:
+        result = dialect_resolver(sa_type, nullable)
+        if result is not None:
+            return result
+    return _default_pxt_type(sa_type, nullable=nullable)
+
+
+def selectable_columns(selectable: sql.Selectable) -> list[sql.ColumnElement]:
+    """Return the output columns of a Selectable in their SELECT-clause order."""
+    if hasattr(selectable, 'selected_columns'):
```
sergey-mkhitaryan: any chance we can do this via `isinstance` checks instead? this seems a bit fragile

pixeltable/exec/sql_data_node.py
sergey-mkhitaryan: any particular reason why only some of these attrs are declared at the class level?

pixeltable/exec/sql_source_node.py
```diff
@@ -0,0 +1,109 @@
...
+from .data_row_batch import DataRowBatch
+from .exec_node import ExecNode
+
+if TYPE_CHECKING:
+    from pixeltable.io.data_sources import SqlDataSource
+
+_logger = logging.getLogger('pixeltable')
+
+
+class SqlSourceNode(ExecNode):
```
sergey-mkhitaryan: Just a heads up that https://github.com/pixeltable/pixeltable/pull/1328 changes the lifecycle semantics of `ExecNode`, so if it goes in first, you may need to update something here.

pixeltable/io/data_sources.py
sergey-mkhitaryan: why not make it always `sql.Connection`?

pixeltable/exec/sql_source_node.py
```diff
@@ -0,0 +1,109 @@
...
+        if isinstance(self.sql_source.conn, sql.Engine):
+            self._conn = self.sql_source.conn.connect()
+            self._owns_conn = True
+        else:
+            self._conn = self.sql_source.conn
+            self._owns_conn = False
+
+        # Tables, subqueries, and aliases are not directly executable; wrap them in `select(...)`.
+        src = self.sql_source.selectable
+        stmt: sql.Executable = src if isinstance(src, sql.Executable) else sql.select(src)  # type: ignore[call-overload]
```
sergey-mkhitaryan: I'd move this logic to sql.py and require that `SqlDataSource.selectable` is an `sql.Executable` or whatever is most convenient. This way if the user input is of the unsupported kind, we can raise the error earlier.

pixeltable/exec/sql_source_node.py
```diff
@@ -0,0 +1,109 @@
...
+                self._conn.close()
+                self._conn = None
+            raise
+
+    async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
+        assert self._result is not None
+        output_batch = DataRowBatch(self.row_builder)
+        for sa_row in self._result:
+            output_row = self.row_builder.make_row()
+            for slot_idx, val in zip(self._mapped_slot_idxs, sa_row):
```
sergey-mkhitaryan: `strict=True`

pixeltable/exec/sql_source_node.py
```diff
@@ -0,0 +1,109 @@
...
+                output_row[slot_idx] = val
+            for slot_idx in self._unmapped_slot_idxs:
+                output_row[slot_idx] = None
+            output_batch.add_row(output_row)
+            if len(output_batch) == self.BATCH_SIZE:
```
sergey-mkhitaryan: we typically do `>=` in such cases even though `>` is technically impossible

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # Tables, subqueries, and aliases are not directly executable; wrap them in `select(...)`.
+        src = self.sql_source.selectable
+        stmt: sql.Executable = src if isinstance(src, sql.Executable) else sql.select(src)  # type: ignore[call-overload]
+        try:
+            self._result = self._conn.execute(stmt)
+        except BaseException:
+            if self._owns_conn:
+                self._conn.close()
+                self._conn = None
```
sergey-mkhitaryan: `if not self._owns_conn`, are we keeping the connection around? what for? would it make more sense to call `self._close()` from here?

pixeltable/utils/sql.py
```diff
@@ -106,6 +106,62 @@ def get_sa_type(col_type: 'ts.ColumnType', engine: sql.Engine) -> sql.types.Type
...
+}
+
+
+def get_pxt_type(sa_type: sql.types.TypeEngine, engine: sql.Engine, *, nullable: bool) -> 'ts.ColumnType | None':
+    """Resolve a SQLAlchemy type to a Pixeltable ColumnType for the engine's dialect.
+
+    Returns None if no mapping is available; the caller is expected to either fall back to a
+    user-supplied override or raise.
+    """
+    dialect_resolver = _PXT_DIALECT_TYPE.get(engine.dialect.name)
```
sergey-mkhitaryan: let's not do this and drop `_PXT_DIALECT_TYPE`. it's easy enough to add later, but for now it's just dead code.

tests/io/test_sql.py
```diff
@@ -208,3 +212,364 @@ def test_errors(self, uses_db: None) -> None:
...
+        return sql.create_engine(Env.get().db_url)
+    raise AssertionError(dialect)
+
+
+def _seed_source(engine: sql.Engine, table_name: str, columns: list[sql.Column], rows: list[dict]) -> sql.Table:
+    """Create `table_name` with `columns` in `engine` and insert `rows`. Returns the SA Table."""
+    meta = sql.MetaData()
+    src = sql.Table(table_name, meta, *columns)
+    meta.create_all(engine)
+    if rows:
```
sergey-mkhitaryan: `if len(rows)>0` more typically in this codebase

tests/io/test_sql.py
```diff
@@ -208,3 +212,364 @@ def test_errors(self, uses_db: None) -> None:
...
+        with engine.begin() as conn:
+            conn.execute(src.insert(), rows)
+    return src
+
+
+class TestImportSql:
+    """import_sql() tests, parametrized across SQLite (file) and the embedded Postgres."""
+
+    @pytest.mark.parametrize('dialect', _IMPORT_DIALECTS)
```
sergey-mkhitaryan: it's not really a "dialect", it's a dbms

tests/io/test_sql.py
```diff
@@ -208,3 +212,364 @@ def test_errors(self, uses_db: None) -> None:
...
+    meta = sql.MetaData()
+    src = sql.Table(table_name, meta, *columns)
+    meta.create_all(engine)
+    if rows:
+        with engine.begin() as conn:
+            conn.execute(src.insert(), rows)
+    return src
+
+
+class TestImportSql:
```
sergey-mkhitaryan: it's pretty convenient when there's one test class per file (if I see `TestImportSql`, I know that it's in `test_import_sql.py`). Please either merge it with the above or move to a separate class.

tests/io/test_sql.py
```diff
@@ -208,3 +212,364 @@ def test_errors(self, uses_db: None) -> None:
...
+                assert row['c_date'] is None
+                assert row['c_uuid'] is None
+                assert row['c_json'] is None
+                assert row['c_bytes'] is None
+            else:
+                assert row['c_float'] == float(i) * 1.5
+                assert row['c_bool'] == (i % 2 == 0)
+                # Pixeltable returns Timestamp as tz-aware (local); the SA DateTime source is naive, so compare
+                # the wall-clock components only.
+                assert row['c_ts'].replace(tzinfo=None) == datetime.datetime(2024, 1, 1) + datetime.timedelta(seconds=i)
```
sergey-mkhitaryan: not good that this logic is duplicated. extract the rows that were inserted in the source table to a local variable, and reuse those values for validation here.

tests/io/test_sql.py
```diff
@@ -208,3 +212,364 @@ def test_errors(self, uses_db: None) -> None:
...
+        comp_src = _seed_source(
+            engine,
+            'comp_src',
+            [sql.Column('c_int', sql.Integer, nullable=False), sql.Column('c_doubled', sql.Integer, nullable=False)],
+            [{'c_int': 5, 'c_doubled': 10}],
+        )
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='computed column'):
+            import_sql(comp_src, engine, 'comp_dest', if_exists='append')
+
+        # --- 5. SqlSourceNode leaf: required destination column missing from source -> MISSING_REQUIRED.
```
sergey-mkhitaryan: `5.`? a claude artifact?

pixeltable/exec/sql_source_node.py
sergey-mkhitaryan: `ExecNode` subclasses are named by what they do, which in this case means `SqlImportNode` is probably the best

tests/io/test_sql.py
```diff
@@ -208,3 +212,364 @@ def test_errors(self, uses_db: None) -> None:
...
+        # Pre-create a destination where c_required is non-nullable; append a source that doesn't supply it.
+        pxt.create_table('req_dest', {'c_int': pxt.Int, 'c_required': pxt.Required[pxt.String]})
+        partial_src = _seed_source(
+            engine, 'partial_src', [sql.Column('c_int', sql.Integer, nullable=False)], [{'c_int': 1}]
+        )
+        with pxt_raises(pxt.ErrorCode.MISSING_REQUIRED, match='c_required'):
```
sergey-mkhitaryan, line 624: would be good to add a test case with `on_error='ignore'` import

sergey-mkhitaryan (changes requested):

pixeltable/catalog/insertable_table.py
christopherpestano: The analog here is insert_table_data_source though, which isn't exposed as public either. This function is called from outside of its respective class as well, it would be odd to have `tbl._insert_sql_source()` in import_sql.

tests/io/test_sql.py
christopherpestano: My bad, removed

pixeltable/exec/sql_source_node.py
christopherpestano: Fixed.

pixeltable/exec/sql_source_node.py
christopherpestano: Hmm odd, the > case should really be an assert since that means something is broken. Don't want to make that change though since the preexisting pattern is already to accept it. Might be worth bringing up with the team

pixeltable/exec/sql_data_node.py
christopherpestano: The reasoning being if the user gave us a connection they "own" the lifecycle of that connection. As we don't exactly know what other things the user is doing with that connection, for example imagine a function written by a user where they import their sql data into a pixeltable table then immediately truncate their original table upon success using the same connection. Even here they might have a try catch that when the import fails they do some random wait then retry the import with the same connection.

tests/io/test_sql.py
christopherpestano: Fixed

tests/io/test_sql.py
christopherpestano: Hmm yeah fair enough changed to _IMPORT_DBMS

tests/io/test_sql.py
christopherpestano: Made it such that we are just asserting the row equality to inserted data.

tests/io/test_sql.py
christopherpestano: Merged

pixeltable/utils/sql.py
christopherpestano: Sure in that case I'll remove the engine param as well as it goes unused otherwise. Put in the docstring that it should be reintroduced in the future.

pixeltable/utils/sql.py
christopherpestano: Sure, changed it so from the start we normalize it so it's always an executable SelectBase which merges the logic together quite a bit

pixeltable/plan.py
christopherpestano: Fixed

pixeltable/exec/sql_data_node.py
sergey-mkhitaryan: I see. In that case it seems appropriate if `SqlDataSource` unconditionally supplied a connection that SqlDataNode doesn't own. `import_sql` would be the one opening and closing the connection if necessary (i.e. if the user provided an engine). Are there any issues with it that I fail to see?

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,130 @@
...
+            col_ref.col.name: exprs.ColumnSlotIdx(col_ref.col, col_ref.slot_idx)
+            for col_ref in self.output_exprs
+            if col_ref.col.name is not None
+        }
+        all_output_slot_idxs = {e.slot_idx for e in self.output_exprs}
+
+        source_names = [
+            getattr(c, 'name', None) or getattr(c, 'key', None) for c in self.sql_source.select_stmt.selected_columns
+        ]
+        # Caller (eg, import_sql) is responsible for validating that all source names are non-None and resolve
```
sergey-mkhitaryan: since the caller must work out this mapping and validate it anyway, why not put it to `SqlDataSource` and avoid doing the same thing here?

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,130 @@
...
+            self._conn = self.sql_source.conn
+            self._owns_conn = False
+
+        try:
+            self._result = (
+                self._conn.execution_options(stream_results=True)
+                .execute(  # type: ignore[call-overload]
+                    self.sql_source.select_stmt
+                )
+                .yield_per(self.BATCH_SIZE)
```
sergey-mkhitaryan: From docs:
```
If
          :meth:`_engine.Result.yield_per` is not used,
          the :paramref:`_engine.Connection.execution_options.stream_results`
          mode of operation will instead use a dynamically sized buffer
          which buffers sets of rows at a time, growing on each batch
          based on a fixed growth size up until a limit which may
          be configured using the
          :paramref:`_engine.Connection.execution_options.max_row_buffer`
          parameter.
```

This sounds like a reasonable default strategy, I think we should defer to it instead of a fixed batch size.

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,130 @@
...
+        if isinstance(self.sql_source.conn, sql.Engine):
+            self._conn = self.sql_source.conn.connect()
+            self._owns_conn = True
+        else:
+            self._conn = self.sql_source.conn
+            self._owns_conn = False
+
+        try:
+            self._result = (
+                self._conn.execution_options(stream_results=True)
```
sergey-mkhitaryan: This modifies `self._conn` object which may not be owned by Pixeltable. I don't have specific suggestions on what to do instead, but that seems not ideal. Let's think/discuss it some more.

pixeltable/functions/timestamp.py
```diff
@@ -287,6 +287,16 @@ def replace(
     return self.replace(**kwargs)
 
 
+@pxt.udf(is_method=True)
+def strip_timezone(self: datetime) -> datetime:
```
sergey-mkhitaryan, line 294: why not add timestamp to `replace()` UDF above?

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,130 @@
...
+            raise
+
+    async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
+        assert self._result is not None
+        output_batch = DataRowBatch(self.row_builder)
+        for sa_row in self._result:
+            output_row = self.row_builder.make_row()
+            for col_info, val in zip(self._mapped_cols, sa_row, strict=True):
+                col = col_info.col
+                if col.col_type.is_image_type() and isinstance(val, bytes):
```
sergey-mkhitaryan, line 85: why only image? do we need special processing for other media types?

pixeltable/io/sql.py
```diff
@@ -74,3 +77,162 @@ def export_sql(
...
+            names; values accept any Pixeltable type spec recognized by `pxt.create_table` (eg, `pxt.Image`,
+            `pxt.Required[pxt.String]`).
+        primary_key: Forwarded to `pxt.create_table` when creating a new table.
+        comment: Forwarded to `pxt.create_table`.
+        custom_metadata: Forwarded to `pxt.create_table`.
+        if_exists: How to handle the destination table.
+
+            - `'error'`: fail if the table already exists.
+            - `'append'`: require the table to exist; verify the source schema is compatible; insert only.
+        on_error: Forwarded to `Table.insert`.
```
sergey-mkhitaryan: `on_error` is forwarded to `insert_sql_source`, but `insert_sql_source` is an implementation detail so we shouldn't be talking about it. A more proper explanation is needed here.

pixeltable/io/sql.py
```diff
@@ -74,3 +77,162 @@ def export_sql(
...
+        schema_overrides: Optional per-column overrides applied on top of the inferred schema. Keys are column
+            names; values accept any Pixeltable type spec recognized by `pxt.create_table` (eg, `pxt.Image`,
+            `pxt.Required[pxt.String]`).
+        primary_key: Forwarded to `pxt.create_table` when creating a new table.
+        comment: Forwarded to `pxt.create_table`.
+        custom_metadata: Forwarded to `pxt.create_table`.
+        if_exists: How to handle the destination table.
+
+            - `'error'`: fail if the table already exists.
+            - `'append'`: require the table to exist; verify the source schema is compatible; insert only.
```
sergey-mkhitaryan: I think `append` behavior is confusing: `if_exists` applies ONLY if it exists, therefore it should not have effect if it doesn't exist.

Consider mirroring the pandas options and behavior https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_sql.html

pixeltable/utils/sql.py
```diff
@@ -106,6 +107,51 @@ def get_sa_type(col_type: 'ts.ColumnType', engine: sql.Engine) -> sql.types.Type
...
+        return ts.JsonType(nullable=nullable)
+    if isinstance(sa_type, sql.types.String):
+        return ts.StringType(nullable=nullable)
+    return None
+
+
+def get_pxt_type(sa_type: sql.types.TypeEngine, *, nullable: bool) -> 'ts.ColumnType | None':
+    """Resolve a SQLAlchemy type to a Pixeltable `ColumnType`
+
+    TODO: Add dialect-specific type mappings as needed"""
```
sergey-mkhitaryan: I don't think that this TODO is useful -- we don't know if those mappings will be needed or not. File a jira instead, or just drop it.

pixeltable/utils/sql.py
```diff
@@ -106,6 +107,51 @@ def get_sa_type(col_type: 'ts.ColumnType', engine: sql.Engine) -> sql.types.Type
...
+        return ts.BinaryType(nullable=nullable)
+    if isinstance(sa_type, sql.types.JSON):
+        return ts.JsonType(nullable=nullable)
+    if isinstance(sa_type, sql.types.String):
+        return ts.StringType(nullable=nullable)
+    return None
+
+
+def get_pxt_type(sa_type: sql.types.TypeEngine, *, nullable: bool) -> 'ts.ColumnType | None':
+    """Resolve a SQLAlchemy type to a Pixeltable `ColumnType`
```
sergey-mkhitaryan: And returns None if no compatible type was found

pixeltable/utils/sql.py
```diff
@@ -106,6 +107,51 @@ def get_sa_type(col_type: 'ts.ColumnType', engine: sql.Engine) -> sql.types.Type
...
+
+    TODO: Add dialect-specific type mappings as needed"""
+
+    return _default_pxt_type(sa_type, nullable=nullable)
```
sergey-mkhitaryan: inline `_default_pxt_type()` code here

pixeltable/plan.py
sergey-mkhitaryan: actually, I just realized: they should just be one union type of argument

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,130 @@
...
+                    filepath, _ = TempStore.save_media_object(
+                        val, col.tbl_handle.id, col.id, col.get_tbl().version, format=None
+                    )
+                    output_row[col_info.slot_idx] = str(filepath)
+                else:
+                    output_row[col_info.slot_idx] = val
+            for slot_idx in self._unmapped_slot_idxs:
+                output_row[slot_idx] = None
+            output_batch.add_row(output_row)
+            if len(output_batch) >= self.BATCH_SIZE:
```
sergey-mkhitaryan: shouldn't we be using `self.ctx.batch_size`?

pixeltable/io/data_sources.py
sergey-mkhitaryan: reopening this

pixeltable/catalog/insertable_table.py
sergey-mkhitaryan: It would be even more odd to expose this to users even though it's for internal use only. In the spirit of putting the customers first, `insert_sql_source` should be `_insert_sql_source` and `insert_table_data_source` should be `_insert_table_data_source` instead (please rename it as well).

sergey-mkhitaryan (changes requested):

pixeltable/catalog/insertable_table.py
```diff
@@ -208,6 +209,37 @@ def insert_table_data_source(
...
+        self,
+        sql_source: SqlDataSource,
+        *,
+        on_error: Literal['abort', 'ignore'] = 'abort',
+        print_stats: bool = False,
+        return_rows: bool = False,
+    ) -> pxt.UpdateStatus:
+        """Stream a SqlDataSource into this table through a single insert plan.
+
+        Internal entry point used by `pxt.io.import_sql`; not part of the public `Table` API. Source-side
```
mkornacker: Don't comment on caller logic, those things can change.

Code comments need to stand on their own. Also, the generally accepted way of indicating it's not part of the public API is to name it '_...'.

mkornacker (commented):
Some comments that I had meant to submit earlier.

pixeltable/utils/sql.py
```diff
@@ -106,6 +107,51 @@ def get_sa_type(col_type: 'ts.ColumnType', engine: sql.Engine) -> sql.types.Type
...
+    if isinstance(sa_type, sql.types.LargeBinary):
+        return ts.BinaryType(nullable=nullable)
+    if isinstance(sa_type, sql.types.JSON):
+        return ts.JsonType(nullable=nullable)
+    if isinstance(sa_type, sql.types.String):
+        return ts.StringType(nullable=nullable)
+    return None
+
+
+def get_pxt_type(sa_type: sql.types.TypeEngine, *, nullable: bool) -> 'ts.ColumnType | None':
```
mkornacker: We have ColumnType.to_sa_type(). Why shouldn't the other direction be in ColumnType?

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,130 @@
...
+    Source/destination column compatibility (unnamed columns, duplicates, computed-column collisions,
+    missing required columns, unknown destination columns) is validated by the caller (eg, `import_sql`)
+    before the plan is constructed; this node only builds the runtime slot mapping.
+    """
+
+    BATCH_SIZE: ClassVar[int] = 1024
+
+    tbl: catalog.TableVersionHandle
+    sql_source: SqlDataSource
+    output_exprs: list[exprs.ColumnRef]
```
mkornacker, line 33: ExecNode.output_exprs already exists. This is confusing.

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,130 @@
...
+        self._conn = None
+        self._result = None
+        self._mapped_cols = []
+        self._unmapped_slot_idxs = []
+
+    def _open(self) -> None:
+        user_cols_by_name = {
+            col_ref.col.name: exprs.ColumnSlotIdx(col_ref.col, col_ref.slot_idx)
+            for col_ref in self.output_exprs
+            if col_ref.col.name is not None
```
mkornacker, line 57: Why is this if necessary?

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,130 @@
...
+        all_output_slot_idxs = {e.slot_idx for e in self.output_exprs}
+
+        source_names = [
+            getattr(c, 'name', None) or getattr(c, 'key', None) for c in self.sql_source.select_stmt.selected_columns
+        ]
+        # Caller (eg, import_sql) is responsible for validating that all source names are non-None and resolve
+        # to a destination column.
+        self._mapped_cols = [user_cols_by_name[n] for n in source_names]
+        mapped_slot_idxs = {c.slot_idx for c in self._mapped_cols}
+        self._unmapped_slot_idxs = list(all_output_slot_idxs - mapped_slot_idxs)
```
mkornacker, line 65: I lost track at this point of the relationship between row_builder.input_exprs and the named select list items of the input sql statement.

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if isinstance(self.sql_source.conn, sql.Engine):
+            self._conn = self.sql_source.conn.connect()
+            self._owns_conn = True
+        else:
+            self._conn = self.sql_source.conn
```
mkornacker: assert the expected type here - useful for the reader.

pixeltable/io/sql.py
```diff
@@ -74,3 +77,162 @@ def export_sql(
...
+        # If drop_table raises, Python's implicit exception chaining (PEP 3134) preserves the original
+        # insert error on `__context__` and the traceback still shows it.
+        pxt.drop_table(tbl, if_not_exists='ignore')
+        raise
+    return tbl
+
+
+def _validate_append_compatibility(tbl: pxt.InsertableTable, tbl_name: str, inferred_schema: dict[str, Any]) -> None:
+    """Verify the SQL source schema can append into an existing destination table."""
+    tv = tbl._tbl_version.get()
```
mkornacker: Use the public API here: get_metadata().

TableVersion instances are strictly catalog-internal.

mkornacker (changes requested):

pixeltable/catalog/insertable_table.py
christopherpestano: Changed both to _insert_sql_source() & _insert_table_data_source(), changed the docstring to instead just refer to input parameter requirements for the function.

pixeltable/io/data_sources.py
christopherpestano: Done. SqlDataSource.conn is always a Connection now. import_sql opens one from the engine and closes it if the user passed an engine, otherwise it uses the user's connection as is.

pixeltable/exec/sql_data_node.py
christopherpestano: Yep, did exactly that. The node never owns the connection anymore, dropped the _owns_conn logic entirely. import_sql owns the lifecycle and opens/closes it when given an engine.

pixeltable/exec/sql_data_node.py
christopherpestano: Fixed. We pass stream_results per execute() call now instead of setting it on the connection, so we never leave options on a connection we don't own.

pixeltable/exec/sql_data_node.py
christopherpestano: Done, dropped yield_per so it uses the dynamically sized buffer.

pixeltable/exec/sql_data_node.py
christopherpestano: Switched to self.ctx.batch_size.

pixeltable/exec/sql_data_node.py
christopherpestano: Done. import_sql validates and stores the column names on SqlDataSource (col_names) and the node just consumes those instead of recomputing them.

pixeltable/plan.py
christopherpestano: Done. create_insert_plan and TableVersion.insert take a single source: list[dict] | SqlDataSource arg now. Kept query separate since it goes through a different plan path.

pixeltable/io/sql.py
christopherpestano: Fair, changed it so append creates the table if it doesn't exist, matching pandas to_sql. if_exists only has an effect when the table actually exists now.

pixeltable/io/data_sources.py
christopherpestano: It's less slender now, per Sergey's comment below I changed it so it also carries the validated col_names so the node doesn't redo the mapping. Left it as a dataclass for that reason, but happy to inline if you'd still prefer.

pixeltable/io/sql.py
christopherpestano: The TableVersion.get() call is gone, the computed-column check uses get_metadata()[columns] now. Still using _get_schema() for the nullability and type-compatibility checks since get_metadata only exposes types as strings and I need the actual ColumnTypes for is_supertype_of().

pixeltable/exec/sql_data_node.py
christopherpestano: This is intentionally narrowing the type in the same way that InMemoryDataNode does, as later in the code here we are calling .col on the output_exprs which isn't an attribute of the more generic Expr.

pixeltable/exec/sql_data_node.py
christopherpestano: Already fixed, all member vars have class-level declaration

pixeltable/exec/sql_data_node.py
christopherpestano: Trimmed comment then

pixeltable/exec/sql_data_node.py
christopherpestano: It's just any non-system column on the destination table, regardless changed it to plainly be destination table columns

pixeltable/io/sql.py
christopherpestano: Removed block

pixeltable/io/sql.py
christopherpestano: Removed the reference to the insert plan, the docstring now just notes the rows are streamed via a server-side cursor and inserted in batches

pixeltable/utils/sql.py
christopherpestano: Good question, I was mirroring _default_sa_type, since I presumed that there was a special reason we separated the internal pixeltable representation of types from what we coerce those types for our export/import_sql functions. I'm going to merge it back and add the from_sa_type() to ts.

pixeltable/catalog/insertable_table.py
christopherpestano: Changed both to _insert_sql_source() & _insert_table_data_source(), changed the docstring to instead just refer to input parameter requirements for the function.

pixeltable/io/data_sources.py
christopherpestano: Added comment

pixeltable/exec/sql_data_node.py
christopherpestano: Not really I just felt it looked cluttered with all of them at the top, changed to instead have all declared at the class level.

pixeltable/exec/sql_source_node.py
christopherpestano: Thanks for the heads up, in this case like InMemoryDataNode we are compatible with all the changes present in that PR. We also don't benefit from caching since we don't take parameters as pxt.var but regardless there is no issue.

pixeltable/exec/sql_source_node.py
christopherpestano: Moved logic to import_sql makes sense to me

pixeltable/exec/sql_source_node.py
christopherpestano: The closest analog here is actually InMemoryDataNode which is source named. The import in this circumstance refers to the entire insert plan with all the nodes whereas the SqlSourceNode is only really responsible for loading sql rows into DataRows. Instead changed the name to SqlDataNode to make the relationship between the two ExecNodes clearer.

tests/io/test_sql.py
christopherpestano: Added test_import_on_error_ignore below

pixeltable/catalog/insertable_table.py
christopherpestano: Ok, switched to make both private then by renaming to their _ equivalents.

pixeltable/exec/sql_data_node.py
christopherpestano: We are only doing image here since we are trying to keep parity with InMemoryDataNode which only does this logic for images as well. In theory the user could load a pdf into memory then try to store it, but unlike PIL/pure image bytes documents come in too many formats to cleanly materialize here.

pixeltable/utils/sql.py
christopherpestano: Completely removed

pixeltable/utils/sql.py
christopherpestano: Added in from_sa_column

pixeltable/utils/sql.py
christopherpestano: Removed the function

pixeltable/functions/timestamp.py
christopherpestano: In this case the semantics around tzinfo are different from the other parameters present in replace(). Specifically, in replace a value of `None` being passed means that the value is to be ignored when doing the replacement, however in the specific case of timezone `None` is what we use to indicate that we are stripping the timezone entirely.

For example
`def replace(
    self: datetime,
    year: int | None = None,
    month: int | None = None,
    day: int | None = None,
    hour: int | None = None,
    minute: int | None = None,
    second: int | None = None,
    microsecond: int | None = None,
    timezone: _TzInfo | None = True
)`

would be the canonical way to do it, but would likely be very confusing to anyone that just wants to strip timezones.

pixeltable/io/sql.py
```diff
@@ -74,3 +77,167 @@ def export_sql(
...
+    `replace` is intentionally not supported, because create+insert is not atomic and a failed insert would
+    leave the user with an empty replica of the original table. To overwrite, call `pxt.drop_table(tbl_name)`
+    first and then `import_sql(..., if_exists='error')`.
+
+    Args:
+        selectable: A SQLAlchemy `Selectable` (a `Table`, a `select()` statement, or `text(...).columns(...)`)
+            describing the source rows.
+        conn: A SQLAlchemy `Engine` or `Connection` to execute `selectable` against. If a `Connection` is
+            passed, it must remain open and untouched (no commits, rollbacks, or other statements) for the
+            duration of the import; rows are streamed directly from the server-side cursor.
```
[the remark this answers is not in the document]
christopherpestano: Added ut.

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,111 @@
...
+_logger = logging.getLogger('pixeltable')
+
+
+class SqlDataNode(ExecNode):
+    """
+    Streams a SqlDataSource (a normalized SELECT executed against an Engine or Connection) into DataRowBatches.
+
+    Same output contract as InMemoryDataNode:
+      - output_exprs = row_builder.input_exprs
+      - populates user-column slots by name; sets unmapped slots to None.
```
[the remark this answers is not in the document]
christopherpestano: Fixed pr description

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,111 @@
...
+    async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
+        assert self._result is not None
+        output_batch = DataRowBatch(self.row_builder)
+        for sa_row in self._result:
+            output_row = self.row_builder.make_row()
+            for slot_idx, val in zip(self._mapped_slot_idxs, sa_row, strict=True):
+                output_row[slot_idx] = val
+            for slot_idx in self._unmapped_slot_idxs:
+                output_row[slot_idx] = None
+            output_batch.add_row(output_row)
```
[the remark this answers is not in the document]
christopherpestano: This path is unlikely to ever be hit, but I added a fix for it regardless. For this to actually occur FYI the user would have to be storing raw image bytes in their RDBMs which is an anti-pattern

pixeltable/io/data_sources.py
```diff
@@ -0,0 +1,18 @@
...
+import sqlalchemy as sql
+from sqlalchemy.sql.expression import SelectBase
+
+
+@dataclass
+class SqlDataSource:
+    """A user-supplied SQL source: a normalized SELECT statement and an `Engine` or `Connection` to run it against.
+
+    `conn` is always a handle to an external SQL database that the user wants to import from; it is never a
+    connection to Pixeltable's own metadata store.
```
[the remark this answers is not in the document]
christopherpestano: No bad, the point is that a user should see that and know to NEVER point it at the pixeltable db

pixeltable/exec/sql_data_node.py
christopherpestano: If else no longer exists as type is now always a connection.

pixeltable/catalog/insertable_table.py
christopherpestano: Also removed comment on caller logic

pixeltable/exec/sql_data_node.py
christopherpestano: I removed all the Engine/Connection logic here and added some comments, the code is significantly shorter and should be easier to follow.

pixeltable/exec/sql_data_node.py
christopherpestano: This filters out the undo cols that otherwise would be stored as {None: undo_col} in the dictionary. Technically, we could go without it as we access the dict using col_names from the sql source, and since there is no "None" column name they entry never gets touched. But it seems more logically consistent to filter it out before since we aren't going to be doing anything with it here.

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,103 @@
...
+from pixeltable import catalog, exprs
+from pixeltable.utils.local_store import TempStore
+
+from .data_row_batch import DataRowBatch
+from .exec_node import ExecNode
+
+if TYPE_CHECKING:
+    from pixeltable.io.data_sources import SqlDataSource
+
+_logger = logging.getLogger('pixeltable')
```
sergey-mkhitaryan: `__name__`

pixeltable/exec/sql_data_node.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # col_names are the SELECT's output column names (validated by the caller), positionally aligned with each
+        # result row; each resolves to a destination column of the same name.
+        self._mapped_cols = [dest_cols_by_name[n] for n in self.sql_source.col_names]
+        mapped_slot_idxs = {c.slot_idx for c in self._mapped_cols}
+        self._unmapped_slot_idxs = list(all_output_slot_idxs - mapped_slot_idxs)
+
+        # Pass stream_results per-execution rather than via the connection's execution_options: the connection
+        # is owned by the caller, so we must not leave execution options on it.
+        self._result = self.sql_source.conn.execute(  # type: ignore[call-overload]
```
sergey-mkhitaryan, line 69: my understanding of the `ExecNode` contract is that everything here except `self._result` can be initialized once in `init_bindings()` and simply reused on every iteration.

pixeltable/io/sql.py
```diff
@@ -74,3 +77,170 @@ def export_sql(
...
+        tbl_name: Pixeltable path of the destination table.
+        schema_overrides: Optional per-column overrides applied on top of the inferred schema. Keys are column
+            names; values accept any Pixeltable type spec recognized by `pxt.create_table` (eg, `pxt.Image`,
+            `pxt.Required[pxt.String]`).
+        primary_key: Forwarded to `pxt.create_table` when creating a new table.
```
sergey-mkhitaryan: just copy from create_table docstring. saying that it's forwarded to is not helping much. Also please list out somewhere which arguments only apply when a new table is to be created and ignored otherwise.

pixeltable/io/sql.py
```diff
@@ -74,3 +77,170 @@ def export_sql(
...
+            - `'error'`: create the table; fail if it already exists.
+            - `'append'`: append into the table if it already exists (verifying the source schema is
+              compatible); otherwise create it.
+        on_error: How to handle errors encountered while inserting source rows.
+
+            - `'abort'`: any row error aborts the entire import.
+            - `'ignore'`: rows that error are skipped; the rest are inserted.
+
+    Returns:
+        The destination `pxt.Table`.
```
sergey-mkhitaryan: just `` `Table` ``

pixeltable/io/sql.py
```diff
@@ -74,3 +77,170 @@ def export_sql(
...
+        raise excs.RequestError(
+            excs.ErrorCode.INVALID_ARGUMENT,
+            f"`if_exists` must be one of 'error', 'append'; got {if_exists!r}. To overwrite an existing table, "
+            f"`pxt.drop_table(tbl_name)` first and then call `import_sql(..., if_exists='error')`.",
+        )
+
+    stmt = as_select(selectable)
+    sa_cols = list(stmt.selected_columns)
+    source_names: list[str] = []
+    seen: set[str] = set()
```
sergey-mkhitaryan: `seen_cols` or `seen_col_names`

pixeltable/io/sql.py
```diff
@@ -74,3 +77,170 @@ def export_sql(
...
+    for i, sa_col in enumerate(sa_cols):
+        col_name = getattr(sa_col, 'name', None) or getattr(sa_col, 'key', None)
+        if col_name is None:
+            raise excs.RequestError(
+                excs.ErrorCode.INVALID_SCHEMA,
+                f'SQL source has an unnamed output column at position {i}; alias it via '
+                f"`expr.label('name')` so it can be matched to a Pixeltable column.",
```
sergey-mkhitaryan: unnecessary `f` qualifier

pixeltable/io/sql.py
```diff
@@ -74,3 +77,170 @@ def export_sql(
...
+        nullable_attr = getattr(sa_col, 'nullable', True)
+        nullable = True if nullable_attr is None else nullable_attr
+        pxt_type = ts.ColumnType.from_sa_type(sa_col.type, nullable=nullable)
+        if pxt_type is None and (schema_overrides is None or col_name not in schema_overrides):
+            raise excs.RequestError(
+                excs.ErrorCode.INVALID_TYPE,
+                f'Cannot infer a Pixeltable type for SQL source column {col_name!r} (SQLAlchemy type '
+                f'`{sa_col.type}`); provide an entry in `schema_overrides` to resolve it.',
+            )
+        inferred_schema[col_name] = pxt_type  # may be None; overridden below
```
sergey-mkhitaryan: I think it would be easier to follow if col type override+inference was all right here. I.e. If schema override is present, use it. If not, infer. If inferred type is None, fail immediately. No need to update `inferred_schema` (which can be called `schema` at that point) later. Still need to check for unknown columns in schema_overrides though.

pixeltable/io/sql.py
```diff
@@ -74,3 +77,170 @@ def export_sql(
...
+    """Verify the SQL source schema can append into an existing destination table."""
+    column_md = tbl.get_metadata()['columns']
+    existing_schema = tbl._get_schema()
+    computed_col_names = {name for name, md in column_md.items() if md['is_computed']}
+    # A column is required for insert if it is non-nullable and not computed.
+    required_col_names = {
+        name for name, col_type in existing_schema.items() if not col_type.nullable and name not in computed_col_names
+    }
+
+    for col_name, src_type_any in inferred_schema.items():
```
sergey-mkhitaryan: maybe call it `src_type_raw` instead, `src_type_any` is not unreasonable but slightly odd

pixeltable/type_system.py
```diff
@@ -761,6 +761,33 @@ def from_np_dtype(cls, dtype: np.dtype, nullable: bool) -> ColumnType | None:
...
+
+        The inverse of `to_sa_type()`; used to infer a schema from an external SQL source.
+        """
+        if isinstance(sa_type, sql.types.Boolean):
+            return BoolType(nullable=nullable)
+        if isinstance(sa_type, sql.types.Integer):
+            return IntType(nullable=nullable)
+        if isinstance(sa_type, (sql.types.Numeric, sql.types.Float)):
+            return FloatType(nullable=nullable)
+        if isinstance(sa_type, sql.types.Date) and not isinstance(sa_type, sql.types.DateTime):
```
sergey-mkhitaryan: DateTime is not a subclass of Date, so the second part is unnecessary. Besides you can move the `DateTime` branch higher than this if you want  to be extra careful.

tests/io/test_sql.py
```diff
@@ -208,3 +236,407 @@ def test_errors(self, uses_db: None) -> None:
         t_img2 = pxt.create_table('test_img2', {'img': pxt.Image})
         with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match=r"column 'img' of source type Image"):
             export_sql(t_img2, 'img_target', db_connect_str=connection_string, if_exists='insert')
+
+    @pytest.mark.parametrize('dialect', _IMPORT_DBMS)
```
sergey-mkhitaryan: Some comments from Claude regarding the test coverage. I didn't verify any of this (and some of it obviously irrelevant to the pr), but if it's correct about the coverage gaps, I'd address at least 1 and 2. For 1 you can use fault injection or manufacture a request that fails inside `_insert_sql_source`.

Coverage gaps (untested code paths):

1. New-table rollback on insert failure — the except BaseException: pxt.drop_table(...) block in import_sql ([sql.py:198-199](vscode-webview://0stiaclmb3hoc8ven7td18kbd4p7fpsiob8ff7q8lvcjeabvggk2/pixeltable/io/sql.py#L198-L199)) is never exercised. If on_error='abort' and a computed column raises during the first batch, the freshly-created table should be cleaned up. This path is dead in the test suite.

2. Append into a non-insertable target (View) — the isinstance(existing, pxt.InsertableTable) check at [sql.py:179](vscode-webview://0stiaclmb3hoc8ven7td18kbd4p7fpsiob8ff7q8lvcjeabvggk2/pixeltable/io/sql.py#L179) raises UNSUPPORTED_OPERATION. No test covers the case where the named destination is a View rather than a base table.

3. as_select with an unsupported selectable type — as_select() raises INVALID_ARGUMENT for anything that isn't a SelectBase or FromClause (e.g., a bare sql.text(...) without .columns()). This is a realistic user mistake that has no test.

4. on_error='abort' (the default) — only on_error='ignore' is covered by test_import_on_error_ignore. The default abort path, including the rollback behavior above, has no dedicated test.

Redundancy in existing tests:

test_if_exists lines 562–569 test 'replace' and 'garbage' as two separate assertions, but both hit the exact same if if_exists not in ('error', 'append') guard with the same error message. The comment implies 'replace' is "specially" handled — it isn't. One of the two cases can be dropped.

Minor: export suite row count

_run_export_suite creates 100,000 rows per dialect run. With two dialects (sqlite + postgresql), test_export_* inserts ~400,000 rows total. Neither test carries @pytest.mark.expensive, so they'll run in the default make test pass and could be slow on CI.

sergey-mkhitaryan (approved):

pixeltable/exec/sql_data_node.py
christopherpestano: Fixed

pixeltable/io/sql.py
christopherpestano: Fixed

pixeltable/io/sql.py
christopherpestano: Changed to `seen_cols`

pixeltable/io/sql.py
christopherpestano: Removed unnecessary `f`, holdover from when the entire string was one line

pixeltable/type_system.py
christopherpestano: My bad, removed unneeded protective if and moved the DateTime check further up.

pixeltable/io/sql.py
christopherpestano: Fair enough, that makes this slightly easier to read, changed to `src_type_any`

pixeltable/io/sql.py
christopherpestano: Agreed, thats a bit cleaner. Made it such that inferred_schema is now `schema` and changed it such that we get the type directly from the override and only attempt to infer the type when no override is present. Still need to keep the unknown block though for `schema_overrides` in order to catch typos in schema_override.

pixeltable/exec/sql_data_node.py
christopherpestano: Looked into this one. `init_bindings()` only ever gets called from `ExecPlan.__init__`, and insert plans never go through an `ExecPlan`. `create_insert_plan` returns a bare node chain that `StoreBase.insert_rows` drives directly, so if the mapping setup lived in `init_bindings` it would never run and `_mapped_cols` would be empty by the time we iterate. We'd have to add an `init_bindings` call into the insert plan driver to make it work. Since the node gets built and opened exactly once per insert (insert plans aren't cached) there's no recomputation to save, and the existing `init_bindings` overrides are all there to register `bind_sources` for Variable binding, which this node can't have. Keeping the setup in `_open` also matches `InMemoryDataNode`, which builds the same name to slot map there.

pixeltable/io/sql.py
christopherpestano: Copied the descriptions over from `create_table` and noted on all three that they only apply when a new table is created, ignored on append.

tests/io/test_sql.py
christopherpestano: Went back and finished this one off, the param and parametrize argname were still `dialect` in `_import_engine` and the four import tests. Renamed everything to `dbms`. Left `_DialectSpec` on the export side alone since that one really is about SQLAlchemy dialect type mappings.

pixeltable/io/sql.py
christopherpestano: Reworded this to match the `insert` docstring. The previous ignore bullet was actually wrong, rows don't get skipped, they still get inserted with `None` in the erroring cells and the error info lands in `errortype`/`errormsg`.

tests/io/test_sql.py
christopherpestano: Added `test_import_abort_and_view_errors` to cover 1 and 2. For 1 it imports a source where the second row is a corrupt image, asserts the import raises under the default `on_error='abort'` and that the freshly created table is gone from the catalog afterwards, which exercises the rollback block. For 2 it appends into a view and asserts the `UNSUPPORTED_OPERATION` error. Since the test runs with the default `on_error` it covers 4 as well. Skipped 3, the `as_select` guard is a single isinstance check and didn't seem worth a dedicated test.

pixeltable/io/data_sources.py
mkornacker: I would still call this a very slender abstraction, and now it's also referencing something (destination schema) that's not visible from here.

mkornacker (approved):

pierrebrunelle:
@christopherpestano switch title for the notebook: "Working with data from SQL databases" is unclear that it covers Import/Export just be straightforward: Import/Export from/to SQL Databases.

PR #1326, opened 2026-05-08
[PXT-1138] Fault injection framework for testing

sergey-mkhitaryan:
Introduces fault injection framework and a multi-threaded test coordinator for testing races. Adds new `test_catalog.py` tests that demonstrate the use of both. Also simplifies `test_rate_limits_scheduler_exception_before_pool` with fault injection.

### Fault injection

A lightweight fault injection framework for testing failure and concurrency scenarios in instrumented code paths.

Production code calls `process_fault(loc)` at each `FaultLocation`. In production this is a
no-op (`ProdFaultManager`). In tests, a `TestFaultManager` can register a `Fault` at any
location before triggering the code under test:

- `ExceptionFault`: raises a given exception
- `BlockFault`: pauses the thread at the fault point, letting the test inspect or manipulate shared state before calling
`unblock()`. Useful for reproducing race conditions.

### MultiThreadedScenario

A builder for deterministic multi-threaded test scenarios. Steps are added with `then_run` / `then_run_until` and executed in strict sequential order across threads, with each step gated on the previous one completing.


pixeltable/utils/fault_injection.py
```diff
@@ -0,0 +1,161 @@
...
+        _logger.info('Waiting until fault point is reached...')
+        self.reached.wait()
+        _logger.info('BlockFault reached')
+
+    def unblock(self) -> None:
+        _logger.info('Unblocking')
+        self.unblocked.set()
+
+
+def process_fault(loc: FaultLocation) -> None:
```
aaron-siegel, line 23: I'd simply define this here as
```
def process_fault(loc: FaultLocation) -> None:
    pass
```
to keep it as lightweight as possible in production. Then monkey-patch it in the tests.

And I think that might be the only thing that actually needs to live in the prod codebase.

pixeltable/utils/fault_injection.py
```diff
@@ -0,0 +1,161 @@
...
+    SCHEDULER_RATE_LIMITS_AEXEC = auto()
+
+
+class Fault(Protocol):
+    recurring: bool
+
+    def __call__(self, loc: FaultLocation) -> Any: ...
+
+
+class FaultManager(abc.ABC):
```
aaron-siegel: I don't see a need for this abstraction? (See below.)

pixeltable/utils/fault_injection.py
```diff
@@ -0,0 +1,161 @@
...
+        raise excs.Error(excs.ErrorCode.INTERNAL_ERROR, 'Fault injection not allowed outside of tests')
+
+    def process_fault(self, loc: FaultLocation) -> None:
+        return
+
+    def clear_faults(self) -> None:
+        return
+
+
+class TestFaultManager(FaultManager):
```
aaron-siegel: I'd keep all this logic out of the production codebase entirely. Move it to `tests`, then monkey-patch during test initialization to inject the testing logic. (See below.)

tests/coordinator.py
```diff
@@ -0,0 +1,128 @@
+import logging
+import time
+from dataclasses import dataclass
+from threading import Event, Lock, Thread, current_thread
+from typing import Any, Callable
+
+_logger = logging.getLogger('pixeltable')
+
+
+@dataclass
```
aaron-siegel, line 10: frozen=True?

tests/coordinator.py
```diff
@@ -0,0 +1,128 @@
...
+class _Step:
+    thread_id: int
+    name: str
+    fn: Callable[[], Any]
+    gate: Event | None = None
+    next_gate: Event | None = None
+
+
+class MultiThreadedScenario:
```
aaron-siegel, line 19: This is fairly complex logic, so I'd propose adding some tests that validate the operation of this testing module (i.e., they validate that faults actually are getting injected where they're intended to be).

Reasoning being, the functional tests aren't guaranteed to distinguish between "catalog successfully recovered from injected faults" vs. "faults never got properly injected". A simple test that runs a function with process_fault() and verifies that the expected exception is thrown, would plug this hole. (And whatever other tests would help verify the correct operation of the fault injection system.)

aaron-siegel (changes requested):

pixeltable/utils/fault_injection.py
sergey-mkhitaryan: good point, done

tests/coordinator.py
sergey-mkhitaryan: Can't be, `MultiThreadedScenario` sets `gate` and `next_gate` on these to stitch them together. Sure, those gates can be kept somewhere separately, but that will make `MultiThreadedScenario.execute` uglier and more difficult to follow.

tests/coordinator.py
sergey-mkhitaryan: Let me know if I misunderstood because you posted this comment on a piece of code that has nothing to do with fault injection.

I've added `test_inject_exception` to address this alleged gap. However I disagree that the gap exists because the tests that use fault injection self-verify by calling `fault.assert_count(1)`. `test_concurrent_add_column_insert` self-verifies by using `fault.reached` as a gate for one of the steps -- this test would freeze and timeout if fault injection doesn't work.

If that changed your mind, I can revert the latest commit.

pixeltable/utils/fault_injection.py
```diff
@@ -0,0 +1,25 @@
...
+class FaultLocation(Enum):
+    """Instrumented locations in the codebase where faults can be injected."""
+
+    CATALOG_FINALIZE_PENDING_OPS_NON_XACT = auto()
+    CATALOG_LOAD_VIEW_OP_EXEC = auto()
+    SCHEDULER_RATE_LIMITS_AEXEC = auto()
+    TEST = auto()
+
+
+def create_fault_manager() -> Any:
```
aaron-siegel, line 18: Do we really need this abstraction in the production code? See below.

pixeltable/runtime.py
```diff
@@ -41,6 +42,7 @@ class Runtime:
     _progress: Progress | None
     _event_loop: asyncio.AbstractEventLoop | None  # event loop for this thread
     _run_coro_executor: concurrent.futures.ThreadPoolExecutor | None
+    fault_manager: Any
```
aaron-siegel, line 45: I'm not sure this actually needs to be in `Runtime`. The only reason to put it there is to have one instance per thread, right? So - could it simply be a separately managed thread-local abstraction that exists only in tests?

Just trying to keep the production code as uncluttered as feasible.

tests/conftest.py
```diff
@@ -158,6 +158,22 @@ def init_env(tmp_path_factory: pytest.TempPathFactory, worker_id: int) -> None:
             _logger.warning(f'Failed to cleanup test schema {schema_name}: {e}')
 
 
+@pytest.fixture(autouse=True)
+def fault_injection() -> Iterator[None]:
+    """Enables fault injection"""
+    import pixeltable.utils.fault_injection as prod_fault_injection
```
aaron-siegel: Move to top of file

tests/coordinator.py
aaron-siegel: Ok.

tests/coordinator.py
aaron-siegel: I saw the fault.assert_count() calls, but I'm skeptical that's enough. I'd be most comfortable if there were an exception that actually propagated up due to an uncaught fault injection. I believe the new test does this.

tests/conftest.py
```diff
@@ -158,6 +158,22 @@ def init_env(tmp_path_factory: pytest.TempPathFactory, worker_id: int) -> None:
...
+    import tests.fault_injection as test_fault_injection
+
+    # Monkey patch fault injection to product
+    prod_fault_injection.process_fault = test_fault_injection.process_fault
+    prod_fault_injection.create_fault_manager = test_fault_injection.create_fault_manager
+
+    try:
+        yield
+    finally:
+        get_runtime().fault_manager.clear_faults()
```
aaron-siegel, line 174: See comments above - instead of this, we could define
```
TEST_TLS = threading.local()
```
somewhere in the `tests` tree (e.g., in `tests/utils.py`), and then put `fault_manager` there (`TEST_TLS.fault_manager`), keeping it out of `Runtime`.

aaron-siegel (approved):

pixeltable/runtime.py
sergey-mkhitaryan: `ExecNode` spins up threads and propagates its `Runtime` state to them by calling `Runtime.copy_db_context`. Fault manager being a part of `Runtime` is pretty convenient for this reason. Let's discuss in Slack.

PR #1328, opened 2026-05-10
PXT-1140: parameterized queries + plan caching + thread-safe catalog

mkornacker:
- Variable to SQL bindparam pipeline. Variable.prepare() and SimilarityExpr.prepare() register values; SqlNode.bind_params() passes them to conn.execute. One compiled plan is reused across distinct bind values, letting Postgres reuse its prepared-statement plan cache.                                      
- Parameterized limit and offset. Query.limit() and Query.offset() accept Literal or Variable Exprs.                                                     
- ExecPlan is the unit of cached compiled output. Planner deepcopies its inputs internally and stores planner-mutated select-list exprs and the exec tree on ExecPlan. Plans are cached on per-thread Runtime.plan_cache, keyed by Query identity, with a version-aware staleness check.
- TableVersionPath and TableVersionHandle are thread-safe. Per-thread cache slots via threading.local; cross-thread access no longer asserts. Query is genuinely immutable and threadsafe; _rebind and per-thread Query deepcopies are gone.                                                                    
- v_max = MAX_VERSION for live reads. Matches the partial PK index condition; live-row equality lookups now hit the index instead of seq-scanning.                                                                                                                             
- Comparison numeric promotion. Comparison.sql_expr whitelists same-SA-type, both-numeric, and both-date operands and lets Postgres handle implicit casts; other shapes still fall back to FilterNode.
                                                                                                                                                           
Bench (FastAPIRouter q1 lookup vs raw SQLAlchemy baseline), rps:
                                                                                                                                                           
    conc   pre-branch   this branch   pg-realistic
     1        89          484           1412                                                                                                               
     8       256          470           1793
    32       230          426            859


pixeltable/catalog/table_version_handle.py
```diff
@@ -21,30 +21,25 @@ class TableVersionHandle:
...
-        self._origin_thread_id = threading.get_ident()
-        self._origin_catalog = get_runtime().catalog
+        self._local = threading.local()
+        self._local.tbl_version = tbl_version
+        self._local.origin_catalog = get_runtime().catalog
 
     def __deepcopy__(self, memo: dict[int, object] | None = None) -> TableVersionHandle:
-        # reset _tbl_version, we might end up in a different thread
-        return TableVersionHandle(self.key)
+        # Thread-safe and structurally immutable: callers can share a single instance.
```
aaron-siegel: Add a comment to class docstring mentioning instances are immutable?

pixeltable/catalog/table_version_handle.py
```diff
@@ -77,27 +72,26 @@ def is_snapshot(self) -> bool:
...
-        # guard against incorrect cross-thread access; inherited-context threads are allowed
-        assert self._origin_thread_id == threading.get_ident() or get_runtime().context_inherited
-
         cat = get_runtime().catalog
-        cached = self._tbl_version
-        needs_refresh = self._origin_catalog is not cat or cached is None or not cached.is_validated
+        # getattr(), not attribute access: threads other than the originating one will have an empty _local
+        cached: TableVersion | None = getattr(self._local, 'tbl_version', None)
+        origin_catalog: Catalog | None = getattr(self._local, 'origin_catalog', None)
+        needs_refresh = origin_catalog is not cat or cached is None or not cached.is_validated
```
aaron-siegel: I'd actually find this a lot easier to grasp by eliminating the double negatives:
```python
if origin_catalog is cat and cached is not None and cached.is_validated:
    return cached
```

pixeltable/catalog/table_version_path.py
```diff
@@ -91,22 +80,25 @@ def as_md(self) -> schema.TableVersionPath:
...
         cat = get_runtime().catalog
-        cached = self._cached_tbl_version
+        # getattr(), not attribute access: threads other than the originating one will have an empty _local
+        cached: TableVersion | None = getattr(self._local, 'cached_tbl_version', None)
+        origin_catalog: Catalog | None = getattr(self._local, 'origin_catalog', None)
         needs_refresh = (
-            self._origin_catalog is not cat or cached is None or (get_runtime().in_xact and not cached.is_validated)
+            origin_catalog is not cat or cached is None or (get_runtime().in_xact and not cached.is_validated)
         )
         if not needs_refresh:
```
aaron-siegel: Same comment as above: the double negatives make this harder for me to reason about

pixeltable/catalog/view.py
```diff
@@ -119,10 +121,17 @@ def _create(
                     excs.ErrorCode.UNSUPPORTED_OPERATION,
                     f'View sample clause cannot be computed in the context of the base table {base.tbl_name()!r}',
                 )
-            # create a copy that we can modify and store
-            sc = sample_clause
+
+            # create a copy that we can modify and store; materialize a seed when the user did not
```
aaron-siegel: Hmmm. This is a significant change in behavior that's unrelated to the PR content.

Maybe it should be a separate PR that we can discuss independently?

pixeltable/exec/exec_node.py
```diff
@@ -6,10 +6,12 @@
 import queue
 import threading
 from types import TracebackType
-from typing import AsyncIterator, Iterable, Iterator, TypeVar
+from typing import Any, AsyncIterator, Iterable, Iterator, TypeVar
 
 from typing_extensions import Self
 
+import pixeltable.exceptions as excs
```
aaron-siegel, line 13: Not `from pixeltable import ...`?

pixeltable/exec/sql_node.py
```diff
@@ -167,16 +172,18 @@ def __init__(
...
+        self._stmt = None
+        # SqlNode subclasses with embedded CTEs assign their inputs here so this node's finalize() sees the full
+        # structure
+        self._cte_inputs: list[SqlNode] = []
 
         if self.tbl is not None:
-            tv = self.tbl.tbl_version._tbl_version
-            if tv is not None:
-                assert tv.is_validated
+            assert self.tbl.tbl_version.get().is_validated
```
aaron-siegel: Now it can be just
```python
assert self.tbl is None or self.tbl...
```

pixeltable/exec/sql_node.py
```diff
@@ -228,15 +235,33 @@ def _create_stmt(self) -> sql.Select:
...
 
         return stmt
 
     def _ordering_tbl_ids(self) -> set[UUID]:
         return exprs.Expr.all_tbl_ids(e for e, _ in self.order_by_clause)
 
+    def finalize(self) -> None:
+        self.bind_sources.extend(list(self.select_list))
+        self.bind_sources.extend(list(self.cell_md_refs))
+        if self.where_clause is not None:
```
aaron-siegel: Use the same trick as you did above?
```python
self.bind_sources.extend(e for e in (self.where_clause, self.limit, self.offset) if e is not None)
```

pixeltable/_query.py
```diff
@@ -970,7 +948,6 @@ def select(self, *items: Any, **named_items: Any) -> Query:
             >>> query = person.select(t.name, is_adult=(t.age >= 18))
 
         """
-        self = self._rebind()  # noqa: PLW0642
```
aaron-siegel, line 973: 👍

aaron-siegel (approved):

pixeltable/exec/sql_node.py
mkornacker: Done

pixeltable/exec/exec_node.py
mkornacker: A bit shorter this way. Is there a downside?

pixeltable/catalog/table_version_handle.py
mkornacker: The docstring already covers this. Removed.

pixeltable/catalog/table_version_handle.py
mkornacker: Done.

pixeltable/catalog/table_version_path.py
mkornacker: Done.

pixeltable/exec/sql_node.py
mkornacker: Done.

pixeltable/catalog/view.py
mkornacker: Done.

pixeltable/catalog/table_version_handle.py
sergey-mkhitaryan: @mkornacker did you make changes but didn't push? I was going to write the same comment.

pixeltable/catalog/table_version_handle.py
```diff
@@ -77,27 +72,26 @@ def is_snapshot(self) -> bool:
...
             new_tv.is_validated = True
         else:
-            # either no cached instance, or the catalog changed
+            # either no cached instance on this thread, or the catalog changed
             new_tv = cat.get_tbl_version(self.key)
             assert new_tv.key == self.key
-        self._tbl_version = new_tv
-        self._origin_catalog = cat
+        self._local.tbl_version = new_tv
+        self._local.origin_catalog = cat
```
sergey-mkhitaryan, line 92: can you call this `owning_catalog`? I feel like "origin" is immutable by definition.

pixeltable/catalog/table_version_path.py
```diff
@@ -91,22 +80,25 @@ def as_md(self) -> schema.TableVersionPath:
...
             return
 
         with get_runtime().catalog.begin_xact(for_write=False, read_tbl_ids=[self.tbl_version.id]):
             new_tv = self.tbl_version.get()
-        self._cached_tbl_version = new_tv
-        self._origin_catalog = cat
+        self._local.cached_tbl_version = new_tv
+        self._local.origin_catalog = cat
+
+    def _cached_md(self) -> TableVersion:
```
sergey-mkhitaryan: why "md"? more like cached_tv, no?

also, you don't need it. you can just make `refresh_cached_md` private, and have it return tbl version.

pixeltable/catalog/view.py
```diff
@@ -119,10 +121,17 @@ def _create(
...
+            # supply one, so that subsequent reads of this view/snapshot return the same rows
+            seed = sample_clause.seed if sample_clause.seed is not None else random.randint(0, 1 << 63)
             sample_clause = SampleClause(
-                sc.version, sc.n, sc.n_per_stratum, sc.fraction, sc.seed, sc.stratify_exprs.copy()
+                sample_clause.version,
```
sergey-mkhitaryan: `sample_clause = dataclasses.replace(...)`

pixeltable/catalog/table_version_path.py
mkornacker: Done.

pixeltable/catalog/table_version_handle.py
mkornacker: It's the origin of tbl_version.

pixeltable/catalog/view.py
mkornacker: Done.

pixeltable/exec/exec_node.py
```diff
@@ -21,7 +23,16 @@
...
 class ExecNode(abc.ABC):
     """
-    Base class of all execution nodes
+    Base class of all execution nodes.
+
+    Lifecycle:
+    1. The immutable node structure (output_exprs, input node, etc.) can be created incrementally
+       (ie, after init(); the planner might need to make adjustments after instance construction)
+    2. finalize() assumes the immutable structure is complete and sets bind_sources/vars
+    3. bind_params()/iter() can then be called repeated to execute the same plan with different parameters
```
sergey-mkhitaryan: typo: repeatedly

pixeltable/exec/exec_node.py
```diff
@@ -21,7 +23,16 @@
...
+    4. _open() initializes per-iteration execution state
```
sergey-mkhitaryan, line 33: what does iteration mean in this context?

pixeltable/exec/exec_node.py
```diff
@@ -48,12 +68,46 @@ def __init__(
...
+        self.bind_sources = []
+        self.vars = []
+        self.bound_args = {}
 
     def set_ctx(self, ctx: ExecContext) -> None:
         self.ctx = ctx
         if self.input is not None:
             self.input.set_ctx(ctx)
 
+    def finalize(self) -> None:
```
sergey-mkhitaryan: I feel like this is a misleading name, in python and other languages finalize often refers to the step that happen before the object is GCd. "prepare"?

pixeltable/exec/exec_node.py
```diff
@@ -159,13 +213,25 @@ def get_node(self, node_class: type[T]) -> T | None:
...
         if self.input is not None:
             self.input.set_limit(limit)
 
-    def set_offset(self, offset: int) -> None:
+    def set_offset(self, offset: exprs.Expr) -> None:
         """Default implementation propagates to input"""
         if self.input is not None:
             self.input.set_offset(offset)
+
+    def _resolve_limit_offset(self, e: exprs.Expr, role: str) -> int:
```
sergey-mkhitaryan: call it `_resolve_expr_to_int`, or `_resolve_to_int`. limit or offset happen to be what it is currently used for, but it is not fundamentally limited to only resolving those parameters.

the error message at the bottom refers to limit/offset, but it doesn't have to, because you have `role` argument

pixeltable/exec/exec_plan.py
```diff
@@ -0,0 +1,150 @@
...
+
+    All resources needed to execute a plan (sequence of ExecNodes) are included here.
+
+    Not thread-safe.
+    """
+
+    exec_root: ExecNode
+    ctx: ExecContext
+
+    # the parameter signature, merged across nodes (one entry per distinct Variable name)
```
sergey-mkhitaryan: can just say that it's variable name -> type. the "one entry per" is already obvious from it being a dict.

pixeltable/exec/exec_plan.py
```diff
@@ -0,0 +1,150 @@
...
+    param_types: dict[str, ts.ColumnType]
+    # nodes that declared parameters; used to dispatch bind_params()
```
sergey-mkhitaryan: nit: `used to dispatch bind_params()` is not a useful comment, I'd just drop it. it doesn't tell me enough to free me from having to find its usages and reading though them. (Also you've renamed `bind_params` since then)

pixeltable/exec/exec_plan.py
```diff
@@ -0,0 +1,150 @@
...
+    param_nodes: list[ExecNode]
```
sergey-mkhitaryan: many of these (`param_types`, `param_nodes`, `compile_versions`, etc.) are not used outside of this class. should they be prefixed with `_`?

pixeltable/exec/exec_plan.py
```diff
@@ -0,0 +1,150 @@
...
+            except TypeError as e:
+                raise excs.RequestError(
+                    excs.ErrorCode.INVALID_ARGUMENT,
+                    f'bind_params: argument {name!r} is not a valid {expected_type}: {e}',
+                ) from e
+
+        for node in self.param_nodes:
+            node.bind_params(coerced)
+
+    def is_stale(self, current_versions: dict[UUID, int]) -> bool:
```
sergey-mkhitaryan: `matches_table_versions`? `compile_versions` don't need to be newer or older, they need to be the same

pixeltable/exec/exec_plan.py
```diff
@@ -0,0 +1,150 @@
...
+            node = node.input
+
+    def _bind_params(self, args: dict[str, Any]) -> None:
+        """Assign values to plan parameters"""
+        # extra args are silently ignored (Variables may have been substituted away at compile time)
+        missing = self.param_types.keys() - args.keys()
+        if len(missing) > 0:
+            raise excs.RequestError(excs.ErrorCode.INVALID_ARGUMENT, f'bind_params: missing: {sorted(missing)}')
+
+        # coerce each value to the canonical Pixeltable representation for its declared type
```
sergey-mkhitaryan, line 94: As a further optimization, can coercing be done in compile time instead?

sergey-mkhitaryan (changes requested):

pixeltable/exec/exec_node.py

pixeltable/exec/exec_plan.py
mkornacker: done.

pixeltable/exec/exec_plan.py
mkornacker: The args don't show up until runtime.

pixeltable/exec/exec_node.py

pixeltable/exec/exec_node.py
mkornacker: Referring to point 3: one round of iter()

pixeltable/exec/exec_plan.py
mkornacker: Done.

pixeltable/exec/exec_node.py

sergey-mkhitaryan (approved):
rubber stamping

PR #1333, opened 2026-05-12
Agent-friendly CLI

mkornacker:
This PR turns `pxt` into a low-latency, daemon-backed CLI for catalog inspection and management, and folds the previous `pxt serve` / `pxt deploy` functionality in as subcommands.

- **Replacement, not alongside.** `pxt` (`pyproject.toml:[project.scripts]`) is repointed from `pixeltable.cli:main` to `pxt_cli.client.main:main`. The new CLI is a superset: `serve` and `deploy` continue to work, plus `ls`, `describe`, `rows`, `get`, `count`, `errors`, `history`, `columns`, `computed`, `idxs`, `status`, `config`, `drop`, `rm`, `rename`, `mv`, `revert`, `shell`, `health`, `daemon`, `dashboard`.
- **Daemon-backed.** Catalog commands route over local HTTP to a long-lived daemon (`pxt_cli/server/`), so cold start is ~40 ms instead of ~1 s. Auto-spawn on first command; identity-fingerprinted per `(PIXELTABLE_HOME, port)`; restarts on identity drift. Stdlib `http.server.ThreadingHTTPServer` + custom router
- **Dashboard.** The daemon also hosts the dashboard SPA. The SPA bundle ships in the wheel.
- **Docs.** `docs/release/platform/cli.mdx` rewritten to cover both catalog operations and serve/deploy under one unified `pxt` page; the old `pcli.mdx` was merged in and removed. `README.md` and `dashboard/ARCHITECTURE.md` updated to reflect the unified `pxt` entry point and the stdlib server stack.

Proposal: https://www.notion.so/pixeltable/Agent-friendly-CLI-35d54609e49080869315ce8a597f7cee


pierrebrunelle (commented):
## Review: played with pcli hands-on, ran all 21 tests (pass), read every file

### Bugs

**B1. Missing table → 500 instead of 400.** `describe`, `rows`, `count`, `get`, `errors`, `history`, `columns`, `idxs` all return `pcli: 500 Internal Server Error` for nonexistent tables. `drop` and `revert` handle this correctly. Fix: add a global exception handler in `app.py`:

```python
from pixeltable import exceptions as excs
from fastapi.responses import JSONResponse

@app.exception_handler(excs.Error)
async def pxt_error_handler(request, exc):
    return JSONResponse(status_code=400, content={"detail": str(exc)})
```

**B2. `pcli ls nonexistent_dir` silently returns empty output, exit 0.** `_descend` returns `[]` for missing paths — indistinguishable from an empty dir. Should exit non-zero or print a warning.

**B3. `probe.py:33` leaks file descriptor.** `log = open(log_path, 'a')` is never closed in the parent process after `Popen`.

**B4. Cold start with port conflict gives a 15s timeout then `RuntimeError`.** When another process holds port 22089, the daemon log shows the bind error but the client doesn't surface it. Reading the last line of `~/.pixeltable/logs/pcli-daemon.log` on timeout would help.

### Design feedback

**D1. `--json` on `ls` returns `num_cols: null, flags: ""` unless you also pass `-l`.** Since `--json` is the agent codepath, it should always return full metadata (or document the interaction). Tested: `pcli ls --json` has no `num_cols` or meaningful `flags` on any entry.

**D2. `pcli idxs` is very noisy.** 45 btree (auto-created) vs 10 embedding indexes in my catalog. An `--embedding` filter would help agents skip the noise.

**D3. `env` only shows `PIXELTABLE_*` vars.** Misses `PCLI_PORT` and doesn't indicate presence of `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` — the exact vars agents debug.

### Nits

- Three `.md` plan files at repo root (`pxt-cli-design.md`, `pcli-v0-plan.md`, `pcli-errors-and-metadata-plan.md`) — shouldn't land on main.
- `pcli health` ignores argv entirely (no `--help`). Every other command supports it.
- `--json` envelope inconsistency: `ls` wraps in `{"entries": [...], "tree": null}`, most others return the array directly.
- `main.py` if/elif chain (18 branches) — a dict dispatch would scale better.

### Performance (warm, PCLI_PORT=22090)

| Command | Latency |
|---------|---------|
| health | ~245ms |
| ls | ~200ms |
| ls -l app | ~200ms |
| ls --counts app | ~16s |
| describe | ~390ms |
| count | ~425ms |
| rows -n 2 | ~350ms |

~200ms floor is Python client startup. Acceptable for v0.

mkornacker:
@copilot review

pierrebrunelle (commented):
### Remaining items

1. **`ls -l` details mode calls `get_metadata()` per table.** The old version used `TreeNode` fields from `get_dir_tree()`. The new version is more accurate but adds one round-trip per table. For small catalogs this is fine (~360ms for 9 tables). For catalogs with 100+ tables it could be slow. Worth noting in a follow-up — maybe cache or batch.

2. **`ls --json` envelope still wraps in `{"entries": [...], "tree": null}`** while most other commands return the array directly. Minor inconsistency but not a blocker.

3. **`errors` route now filters `is_stored` columns** (line 23: `c.get('is_stored')`). Good fix — unstored computed columns would have failed on `errortype` access.

4. **No `shell` command yet** — it's in the `COMMANDS` dict (`'shell': ('shell', 'interactive REPL')`) but no `pcli/client/commands/shell.py` exists. Running `pcli shell` will crash with `ModuleNotFoundError`. Either add a stub or remove from `COMMANDS` until implemented.

### When the pcli merges into pxt, a few to consider if easy to add!?!

**`pxt recompute <table> [--col NAME] [--where-errors]`** — completes the error-recovery loop. Agents already use `pxt errors` to find failures; today they must drop to Python for `t.recompute_columns()`. This closes the loop entirely in the CLI.

**`pxt plan <table>`** — computed column dependency DAG. The starter kit's agent table has an 8-step chain (11 computed cols). When one fails, agents need to see execution order and dependencies. Even a flat topological-sort output would be valuable.

**`pxt lint <paths>`** — static analysis catching the exact mistakes the skill warns about: missing `if_exists=`, similarity on unindexed column, slow API calls without batching. Highest single-command value for AI-generated code.

**`pxt udfs [--module PATH]`** — discover UDFs/queries with name, kind, signature, file:line. Agents joining a project currently grep for `@pxt.udf`; this replaces that.

**`pxt search <table> <column> "query" [-n 5]`** — quick embedding index validation from the CLI. Agents verifying embeddings work after schema setup would use this constantly.

**`pxt endpoints` / `pxt logs`** — already in the design doc. Critical for debugging `pxt serve` deployments.

Just suggestions - feel free to skip

aaron-siegel (commented):
A full review would take a long time, so I'm going to stick to high-level comments. Main ones are:
- Can you help me understand the decision to use fastapi rather than a simple http service (like we did with the frontend)? It's not clear to me what fastapi gains us, given that the only thing we need it for is local interprocess communication. And there's a cost in user experience (the CLI won't work until the user installs the fastapi dependency).
- I think the package should be `pixeltable_cli` to be consistent with the existing `pixeltable_pgserver`.

pixeltable/config.py
```diff
@@ -447,6 +462,32 @@ def get_bool_value(self, key: str, section: str = 'pixeltable') -> bool | None:
     def get_list_value(self, key: str, section: str = 'pixeltable') -> list[Any] | None:
         return self.get_value(key, list, section)
 
+    def config_keys(self) -> Iterable[ConfigKey]:
+        """Return all configuration setting from the known-schema registry."""
```
aaron-siegel: typo

pixeltable/config.py
```diff
@@ -447,6 +462,32 @@ def get_bool_value(self, key: str, section: str = 'pixeltable') -> bool | None:
     def get_list_value(self, key: str, section: str = 'pixeltable') -> list[Any] | None:
         return self.get_value(key, list, section)
 
+    def config_keys(self) -> Iterable[ConfigKey]:
```
aaron-siegel: Why use a generator for this? I'd think it'd be cleaner to just statically materialize it.

pixeltable/config.py
```diff
@@ -447,6 +462,32 @@ def get_bool_value(self, key: str, section: str = 'pixeltable') -> bool | None:
...
+
+    def get_value_source(self, key: str, section: str = 'pixeltable') -> Literal['env', 'file', 'unset']:
+        """Return source of config value returned by get_value():
+        - env: environment variable (or programmatic config override) is set
+        - file: only the config.toml has it
+        - unset: neither layer carries it
+        """
+        if self.lookup_env(section, key) is not None:
+            return 'env'
+        lookup_elems = [*section.split('.'), key]
```
aaron-siegel: What's an example where `section.split('.')` has more than one component?

pixeltable/config.py
```diff
@@ -447,6 +462,32 @@ def get_bool_value(self, key: str, section: str = 'pixeltable') -> bool | None:
...
+        value: Any = self.__config_dict
+        for el in lookup_elems:
```
aaron-siegel: I don't see why it's necessary to do it this way; just check whether
```
self.__config_dict.get(section, {}).get(key)
```
is `None` should be sufficient?

pixeltable/env.py
```diff
@@ -670,6 +672,19 @@ def create_client(self, name: str) -> Any:
...
         return client
 
+    def get_client_credential_params(self) -> dict[str, list[str]]:
+        """Return the credential-bearing init params for each registered API client with credentials.
+
+        Returns: Mapping from provider name to its credential param names in registration order.
+        """
+        out: dict[str, list[str]] = {}
+        with _client_factories_lock:
+            for name, factory in _client_factories.items():
```
aaron-siegel, line 585: This seems really hacky and not guaranteed to generalize well... one straightforward way to make it more robust would be to augment the `env.register_client()` decorator with something like:
```
def register_client(name: str, auth_param: str | None = None)
```

pixeltable/config.py
```diff
@@ -447,6 +462,32 @@ def get_bool_value(self, key: str, section: str = 'pixeltable') -> bool | None:
...
+        """Return all configuration setting from the known-schema registry."""
+        for section, options in KNOWN_CONFIG_OPTIONS.items():
+            for key, info in options.items():
+                if isinstance(info, tuple):
+                    description, expected_type = info
+                else:
+                    description, expected_type = info, str
+                yield ConfigKey(section=section, key=key, description=description, expected_type=expected_type)
+
+    def get_value_source(self, key: str, section: str = 'pixeltable') -> Literal['env', 'file', 'unset']:
```
aaron-siegel: Note that this won't tell you *which* config file a key comes from (my impression of the way it's used in the CLI is that it always implies it comes from `config.toml` or equivalent, which isn't necessarily true; it could be coming from project-level config).

aaron-siegel (changes requested):

pixeltable/config.py
mkornacker: Done.

pixeltable/config.py
mkornacker: Done.

mkornacker:
@aaron-siegel See the latest changes:
- switched to ThreadingHTTPServer to remove FastAPI dependency
- pxt_cli is not a separate package, unlike pixeltable_pgserver
- it's a top-level directory instead of pixeltable/cli in order to avoid importing pixeltable/__init__.py

pixeltable/config.py
mkornacker: Done.

pixeltable/config.py
mkornacker: Done.

pixeltable/env.py
mkornacker: Done.

pixeltable/config.py
mkornacker: Done.

pixeltable/env.py
```diff
@@ -669,20 +660,23 @@ def create_client(self, name: str) -> Any:
                 )
 
         client = client_factory.init_fn(**init_kwargs)
-        self._logger.info(f'Initialized `{name}` client with parameters: {init_kwargs}.')
+        redacted = {**init_kwargs}
```
aaron-siegel: I believe `dict(init_kwargs)` is the preferred way to do this

pixeltable_cli/client/commands/daemon.py
```diff
@@ -0,0 +1,149 @@
...
+    width = max(len(k) for k, _ in rows)
+    for k, v in rows:
+        print(f'{k.ljust(width)}  {v}')
+
+    env_snapshot = health.get('pixeltable_env') or {}
+    if len(env_snapshot) > 0:
+        print()
+        print('PIXELTABLE_* env vars at daemon startup:')
+        for k in sorted(env_snapshot):
+            print(f'  {k}={env_snapshot[k]}')
```
aaron-siegel, line 149: Does this risk leaking a secret?

pxt_cli/client/commands/rm.py
```diff
@@ -1,23 +1,23 @@
...
-from ..http import post
+from ..http import post, quote_path
 from ..parser import Parser
 
 EPILOG = """\
 Examples:
-  pcli rm my_dir -f                # remove an empty directory
-  pcli rm my_dir -r -f             # recursive: also remove contained tables/views/subdirs
-  pcli rm my_dir -n -r             # dry-run
+  pxt rm my_dir -f                # remove an empty directory
```
aaron-siegel: Maybe `pxt drop-dir` would be a better name, to align with the SDK.

pyproject.toml
```diff
@@ -238,7 +238,7 @@ exclude = [
 ]
 
 [tool.hatch.build.targets.wheel]
-packages = ["pixeltable", "pcli"]
+packages = ["pixeltable", "pxt_cli"]
```
aaron-siegel: `pixeltable_cli` would better align with existing auxiliaries as a package name (e.g. `pixeltable_pgserver`)

I don't think the more verbose name would inconvenience anyone, since no one except us ever has to refer to the CLI package, right?

aaron-siegel (approved):

pixeltable_cli/client/commands/daemon.py
mkornacker: No, everything is being redacted by the daemon.

pxt_cli/client/commands/rm.py
mkornacker: Done.

pyproject.toml
mkornacker: Done.

pixeltable/env.py

PR #1336, opened 2026-05-12
[PXT-326] Support (HF) SAM V3 segmentation model

christopherpestano:
Adds support for the SAM 3 (Segment Anything Model 3) promptable concept segmentation model from Hugging Face.

The new sam_for_segmentation UDF in pixeltable.functions.huggingface takes an image and a concept prompt (a text noun phrase, a set of bounding boxes, or both) and returns every matching object instance as a typed Sam3ForSegmentationResponse with per-instance confidence scores, bounding boxes, and binary masks.

facebook/sam3 is a gated model that requires Hugging Face authentication. The accompanying tests skip automatically when no token is configured.

Note: We are bumping up the transformer's version since pre 5.6.0 there is a bug that prevents samv3 automatic segmentation.


tests/functions/test_huggingface.py
```diff
@@ -201,6 +201,54 @@ def test_detr_for_segmentation(self, uses_db: None) -> None:
...
+        skip_test_if_not_installed('transformers')
+        from pixeltable.functions.huggingface import sam_for_segmentation
+
+        t = pxt.create_table('test_tbl', {'img': pxt.Image})
+        t.insert(img=SAMPLE_IMAGE_URL)
+
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='At least one of'):
+            t.add_computed_column(seg=sam_for_segmentation(t.img))
+
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='input_boxes_labels'):
```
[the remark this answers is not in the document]
christopherpestano: Correct as written: add_computed_column wraps the UDF's INVALID_ARGUMENT as UNSUPPORTED_OPERATION (store.py write_column), verified against facebook/sam3.

tests/functions/test_huggingface.py
```diff
@@ -201,6 +201,54 @@ def test_detr_for_segmentation(self, uses_db: None) -> None:
...
         assert 'label_text' in result['segments_info'][0]
 
+    @pytest.mark.expensive
+    def test_sam_for_segmentation(self, uses_db: None) -> None:
+        skip_test_if_not_installed('transformers')
+        from pixeltable.functions.huggingface import sam_for_segmentation
+
+        t = pxt.create_table('test_tbl', {'img': pxt.Image})
+        t.add_computed_column(seg=sam_for_segmentation(t.img, text='orange', threshold=0.3))
+        status = t.insert(img=SAMPLE_IMAGE_URL)
```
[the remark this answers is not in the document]
christopherpestano: Done; now skips via huggingface_hub.get_token() when no token is configured, and dropped the expensive marker for consistency with test_detr_for_segmentation.

pixeltable/functions/vision.py
```diff
@@ -1390,34 +1438,42 @@ def overlay_segmentation(
...
+        id_map = (segmentation.astype(np.int32) * ids).max(axis=0)
+    return _overlay_id_map(
+        img,
+        id_map,
+        alpha=alpha,
+        background=background,
+        segment_colors=segment_colors,
+        draw_contours=draw_contours,
+        contour_thickness=contour_thickness,
+    )
```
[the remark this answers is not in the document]
christopherpestano: Fixed by removing background from the 3D overload and forcing background=0, since uncovered pixels are always 0 in the per-instance form.

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +376,154 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+        ```python
+        {
+            # list of confidence scores for each detected instance
+            'scores': [0.92, 0.88],
+            # list of bounding boxes, [x1, y1, x2, y2] in absolute pixel coordinates
+            'boxes': [[51.9, 356.1, 181.4, 413.9], [383.2, 58.6, 605.6, 361.3]],
+            # binary masks for each detected instance, shape (num_instances, H, W)
+            'masks': np.ndarray,
+        }
+        ```
```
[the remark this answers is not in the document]
christopherpestano: Done; the UDF now returns a typed Sam3ForSegmentationResponse and the Returns section uses prose instead of a fenced block.

pixeltable/functions/vision.py
```diff
@@ -1390,34 +1438,49 @@ def overlay_segmentation(
...
+    segment_colors: list[str] | None = None,
+    draw_contours: bool = True,
+    contour_thickness: int = 1,
+) -> PIL.Image.Image:
+    num_instances = segmentation.shape[0]
+    if ids is not None and len(ids) != num_instances:
         raise ValueError(
-            f'Segmentation shape {segmentation.shape} does not match image dimensions ({img.height}, {img.width})'
+            f'`ids` (length {len(ids)}) must have the same length as the number of instance masks ({num_instances})'
         )
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The 3D overload now validates that the mask stack's (H, W) matches the image and raises a Pixeltable RequestError with code INVALID_ARGUMENT and a clear message. Covered by a new case in test_overlay_segmentation_instance_masks.

pixeltable/functions/image.py
```diff
@@ -560,6 +560,61 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
...
+        self.canvas = None
+
+    def update(self, tile: PIL.Image.Image, tile_box: tuple[int, int, int, int], width: int, height: int) -> None:
+        if tile is None:
+            return
+        if self.canvas is None:
+            self.canvas = PIL.Image.new('RGB', (width, height))
+        # Tiles that extend past the original image (the padding `tile_iterator` adds to edge tiles) are clipped
+        # by `paste`, so the padding never appears in the result.
+        self.canvas.paste(tile, (int(tile_box[0]), int(tile_box[1])))
```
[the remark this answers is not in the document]
christopherpestano: Fixed. update() now validates width and height against the canvas created from the first row and raises RequestError with code INVALID_ARGUMENT on mismatch. Covered in test_stitch_tiles_edge_cases.

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +377,268 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+    masks: pxt.Array[(None, None, None), pxt.Bool]
+
+
+@pxt.udf
+def sam_for_segmentation(
+    image: PIL.Image.Image,
+    *,
+    model_id: str = 'facebook/sam3',
+    text: str | None = None,
+    input_boxes: list[list[float]] | None = None,
```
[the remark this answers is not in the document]
christopherpestano: The huggingface module entry in public_api.opml intentionally has no children; the doc generator then auto discovers all public UDFs in the module, so sam_for_segmentation is already documented. Adding individual entries would switch the module to explicit mode and drop the other huggingface functions from the docs. The iterator is a different story: auto discovery only picks up UDFs, so we fixed local_public_names to include iterators in module __all__ by default; rendering iterators for auto discovered modules additionally requires a doctools change, which is tracked separately.

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +377,268 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+@pxt.iterator
+def sam_for_video_segmentation(
+    video: pxt.Video,
+    *,
+    text: str,
+    model_id: str = 'facebook/sam3',
+    max_frame_num_to_track: int | None = None,
+    image_size: int = 1008,
+    revision: str | None = None,
+) -> Iterator[Sam3VideoSegmentationFrame]:
```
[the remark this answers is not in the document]
christopherpestano: Fixed. Added test_sam_for_video_segmentation, a smoke test over a short sample video with max_frame_num_to_track bounding the cost. It checks per frame alignment of object_ids, scores, boxes and masks, mask dtype, and that object ids are stable across frames. It skips when transformers is not installed or no Hugging Face token is configured, same as the image SAM test.

pixeltable/functions/vision.py
```diff
@@ -1356,6 +1356,45 @@ def _get_contours(mask: np.ndarray, thickness: int = 1) -> np.ndarray:
...
+    if segment_colors is None:
+        segment_colors = []
+    missing_ids = segment_ids[len(segment_colors) :]
+    auto_colors = __create_label_colors(missing_ids)
+    color_map = {**auto_colors, **dict(zip(segment_ids, segment_colors))}
+
+    segment_alpha = int(alpha * 255)
+
+    overlay_array = np.zeros((img.height, img.width, 4), dtype=np.uint8)
+    rgb_by_id = {id: PIL.ImageColor.getrgb(color_map[id]) for id in segment_ids}
```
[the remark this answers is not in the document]
christopherpestano: Fixed, renamed to sid.

pixeltable/functions/image.py
```diff
@@ -560,6 +560,61 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
...
+
```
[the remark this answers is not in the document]
christopherpestano: Good catch on the crash, though the mechanism is different: the executor never calls update() with None for a parameter that is not declared nullable, it skips the whole row, so the early return was dead code and has been removed. A group consisting entirely of None tiles reaches value() with no canvas; value() now returns None for that case instead of crashing. Covered in test_stitch_tiles_edge_cases.

pixeltable/functions/image.py
```diff
@@ -560,6 +560,61 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
...
+    def value(self) -> PIL.Image.Image:
+        assert self.canvas is not None
+        return self.canvas
```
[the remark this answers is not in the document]
christopherpestano: Fixed, in a slightly different way: rather than raising, value() now returns None when no tiles were aggregated, matching the convention of the builtin aggregates sum and mean for empty input. The assert is gone and the docstring documents the None result.

pixeltable/functions/huggingface.py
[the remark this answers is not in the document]
christopherpestano: See the reply on the earlier identical comment: the empty huggingface module entry means the doc generator auto discovers its UDFs, and enumerating entries would drop the unlisted ones. sam_for_segmentation is documented via auto discovery; iterator rendering needs a doctools change, tracked separately.

pixeltable/functions/image.py
```diff
@@ -560,6 +560,61 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
...
+
+        >>> tiles.group_by(tiles.image).select(
+        ...     stitched=stitch_tiles(
+        ...         tiles.overlay,
+        ...         tiles.tile_box,
+        ...         tiles.image.width,
+        ...         tiles.image.height,
+        ...     )
+        ... ).collect()
+    """
```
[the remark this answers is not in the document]
christopherpestano: Fixed. Added test_stitch_tiles, which tiles the sample image with and without overlap, stitches back, and asserts byte exact reconstruction, plus test_stitch_tiles_edge_cases covering the width and height mismatch error and a group of all None tiles.

pixeltable/functions/vision.py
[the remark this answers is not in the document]
christopherpestano: Duplicate.

pixeltable/functions/image.py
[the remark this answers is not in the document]
christopherpestano: Duplicate.

pixeltable/functions/image.py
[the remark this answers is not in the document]
christopherpestano: Duplicate.

pixeltable/functions/image.py
```diff
@@ -560,6 +560,61 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
             )
 
 
+@pxt.uda
+class stitch_tiles(pxt.Aggregator):
+    """
```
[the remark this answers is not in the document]
christopherpestano: Duplicate.

pixeltable/functions/huggingface.py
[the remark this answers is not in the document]
christopherpestano: Duplicate.

pixeltable/functions/image.py
```diff
@@ -560,6 +560,64 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
...
+
+    canvas: PIL.Image.Image | None
+
+    def __init__(self) -> None:
+        self.canvas = None
+
+    def update(self, tile: PIL.Image.Image, tile_box: tuple[int, int, int, int], width: int, height: int) -> None:
+        if self.canvas is None:
+            self.canvas = PIL.Image.new('RGB', (width, height))
+        elif self.canvas.size != (width, height):
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The canvas is now created with the first tile's mode, and the palette is carried over for palette images. update() also validates that subsequent tiles match the canvas mode and palette, raising RequestError with code INVALID_ARGUMENT on mismatch, since paste would otherwise silently convert mismatched tiles and corrupt pixels. Covered by round trip tests parameterized over RGB, L and RGBA plus a palette preservation check.

pixeltable/functions/vision.py
```diff
@@ -1356,6 +1356,46 @@ def _get_contours(mask: np.ndarray, thickness: int = 1) -> np.ndarray:
...
+        )
+    segment_ids = sorted(int(sid) for sid in np.unique(segmentation) if sid != background)
+    if segment_colors is None:
+        segment_colors = []
+    missing_ids = segment_ids[len(segment_colors) :]
+    auto_colors = __create_label_colors(missing_ids)
+    color_map = {**auto_colors, **dict(zip(segment_ids, segment_colors))}
+
+    segment_alpha = int(alpha * 255)
+
```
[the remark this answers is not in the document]
christopherpestano: Not worthwhile

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +377,338 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+                ('image_size', self.image_size),
+            ),
+        )
+        processor = _lookup_processor(
+            self.model_id, Sam3VideoProcessor.from_pretrained, revision=self.revision, size=self.image_size
+        )
+
+        # The tracker needs the whole clip up front, so drain the wrapped `frame_iterator`; `frame_idx` then
+        # indexes directly into `extracted`.
+        extracted = list(self.frames)
```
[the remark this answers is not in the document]
christopherpestano: Fixed

pixeltable/functions/vision.py
```diff
@@ -1390,34 +1439,55 @@ def overlay_segmentation(
...
+            pxt.ErrorCode.INVALID_ARGUMENT,
+            f'Instance mask shape {segmentation.shape[1:]} does not match image dimensions ({img.height}, {img.width})',
+        )
+    # Each instance is painted with its id, which determines its color. When `ids` is given (e.g. persistent
+    # tracking ids), a given id maps to the same color in every frame; otherwise instances are numbered 1..N in
+    # order. Where masks overlap, later instances win (which is highest-id-wins for the default 1..N case).
+    id_map = np.zeros((img.height, img.width), dtype=np.int32)
+    instance_ids = np.arange(1, num_instances + 1) if ids is None else np.asarray(ids)
+    for instance_id, mask in zip(instance_ids.astype(np.int32), segmentation, strict=True):
+        id_map[mask] = instance_id
```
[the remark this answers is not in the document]
christopherpestano: It's highly unlikely people are going to be segmenting something with 2^32 - 1 objects being tracked, that would require an absurd computer to even process that segmentation.

scripts/prepare-nb-tests.sh
```diff
@@ -37,6 +37,7 @@ EXPENSIVE_NOTEBOOKS=(
     doc-chunk-for-rag
     embedding-indexes
     img-image-to-image
+    img-promptable-segmentation
     multimodal_backend
     queries-and-expressions
```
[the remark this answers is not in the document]
christopherpestano: Hallucination

pixeltable/functions/vision.py
```diff
@@ -1376,12 +1416,21 @@ def overlay_segmentation(
...
     Args:
         img: Input image.
-        segmentation: 2D array of the same shape as `img` where each pixel value is a segment id.
+        segmentation: Either a 2D `(H, W)` int32 array of segment ids, or a 3D `(num_instances, H, W)`
+            boolean array of per-instance masks.
         alpha: Blend factor for the overlay (0.0 = only original image, 1.0 = only segmentation colors).
         background: Segment id to treat as background (not overlaid with color, showing the original
-            image through).
+            image through). Only meaningful for the 2D id-map form.
         segment_colors: List of colors, one per segment id. If the list is shorter than the number of segments, the
```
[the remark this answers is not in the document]
christopherpestano: The ids are baked into the segmentation

pixeltable/exprs/function_call.py
```diff
@@ -374,7 +374,11 @@ def make_args(self, data_row: DataRow) -> tuple[list[Any], dict[str, Any]] | Non
 
         kwargs: dict[str, Any] = {}
         parameters = self.fn.signature.parameters
+        # Aggregator init parameters are passed to the constructor (via agg_init_args), not to update().
+        init_param_names = self.fn.init_param_names[0] if isinstance(self.fn, func.AggregateFunction) else ()
```
mkornacker, line 381: This doesn't deal with polymorphic functions.

pixeltable/functions/image.py
```diff
@@ -560,6 +560,83 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
             )
 
 
+@pxt.uda
+class stitch_tiles(pxt.Aggregator):
+    """
+    Reconstructs a single image from the tiles produced by `tile_iterator`, pasting each tile back at its
```
mkornacker: Hyperlink.

pixeltable/functions/image.py
```diff
@@ -560,6 +560,83 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
...
+    original position. This is the inverse of `tile_iterator`: aggregate the tiles of an image (grouped by a
+    column that identifies the source image) to obtain the full image, optionally after transforming each tile
+    first (e.g. drawing a segmentation overlay on it).
```
mkornacker: e.g.,

pixeltable/functions/image.py
```diff
@@ -560,6 +560,83 @@ def validate(cls, bound_args: dict[str, Any]) -> None:
...
+
+    The reconstruction is naive: tiles are pasted in the order they arrive, so where tiles overlap, later tiles
```
mkornacker, line 572: This should require ordering. See make_video().

pixeltable/functions/vision.py
```diff
@@ -1356,6 +1356,46 @@ def _get_contours(mask: np.ndarray, thickness: int = 1) -> np.ndarray:
...
+def _overlay_id_map(
+    img: PIL.Image.Image,
+    segmentation: np.ndarray,
+    *,
+    alpha: float,
+    background: int,
+    segment_colors: list[str] | None,
+    draw_contours: bool,
+    contour_thickness: int,
+) -> PIL.Image.Image:
```
mkornacker, line 1368: comment: what does this do?

pixeltable/functions/vision.py
```diff
@@ -1376,12 +1416,21 @@ def overlay_segmentation(
 
     If no colors are specified, this function randomly assigns each segment a specific color based on a hash of its id.
 
+    Two input formats are supported:
+
+    - A 2D `(H, W)` int32 id-map where each pixel value is a segment id. This matches the panoptic
+        output of `detr_for_segmentation`.
```
mkornacker: Hyperlinks, and ().

pixeltable/functions/vision.py
```diff
@@ -1376,12 +1416,21 @@ def overlay_segmentation(
...
+    - A 3D `(num_instances, H, W)` boolean stack with one binary mask per instance. This matches the
+        per-instance mask output of `sam_for_segmentation`. Instances are assigned ids `1..num_instances`
```
mkornacker: Same.

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,477 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
 
 
+class Sam3ForSegmentationResponse(TypedDict):
+    scores: pxt.Array[(None,), pxt.Float]
+    boxes: pxt.Array[(None, 4), pxt.Float]
+    masks: pxt.Array[(None, None, None), pxt.Bool]
+
+
+@pxt.udf
+def sam_for_segmentation(
```
mkornacker: Isn't this signature SAM3-specific? If so, call it sam3_for_segmentation?

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,477 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+    masks_np = result['masks'].cpu().numpy()
+    if masks_np.dtype != np.bool_:
+        masks_np = masks_np.astype(bool)
+
+    return Sam3ForSegmentationResponse(
+        scores=result['scores'].cpu().numpy(), boxes=result['boxes'].cpu().numpy(), masks=masks_np
+    )
+
+
+class Sam3AutomaticMaskGenerationResponse(TypedDict):
```
mkornacker: It's identical to the previous response model, unify them.

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,477 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+    stability_score_thresh: float = 0.95,
+    crops_n_layers: int = 0,
+    revision: str | None = None,
+) -> Sam3AutomaticMaskGenerationResponse:
+    """
+    Segments every object in an image with SAM 3 (Segment Anything Model 3), with no prompt. `model_id` should
+    be a reference to a pretrained
+    [SAM 3 Model](https://huggingface.co/docs/transformers/model_doc/sam3) such as `facebook/sam3`.
+
+    Unlike `sam_for_segmentation`, which requires a concept prompt, this samples a grid of points across the
```
mkornacker: Hyperlink.

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,477 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+        self.frames = frame_iterator.decorated_callable(video=video, fps=fps)  # type: ignore[attr-defined]
+        self.text = text
+        self.model_id = model_id
+        self.max_frame_num_to_track = max_frame_num_to_track
+        self.image_size = image_size
+        self.revision = revision
+        self.rows = None
+
+    def _track(self) -> Iterator[Sam3VideoSegmentationFrame]:
+        env.Env.get().require_package('transformers')
```
mkornacker: Do that in init().

tests/functions/test_huggingface.py
```diff
@@ -209,6 +210,128 @@ def test_detr_for_segmentation(self, uses_db: None) -> None:
...
+        from huggingface_hub import get_token
+
+        if get_token() is None:
+            pytest.skip('Skipping SAM 3 test: facebook/sam3 is gated and no Hugging Face token is configured')
+        from pixeltable.functions.huggingface import sam_for_video_segmentation
+
+        video_path = next(f for f in get_video_files() if f.endswith('bangkok_half_res.mp4'))
+        t = pxt.create_table('test_tbl', {'video': pxt.Video})
+        v = pxt.create_view(
+            'test_view', t, iterator=sam_for_video_segmentation(t.video, text=['car'], max_frame_num_to_track=2)
```
mkornacker: Also have a test scenario with an explicit fps arg.

mkornacker (changes requested):
Initial comments.

pixeltable/functions/huggingface.py
christopherpestano: Added link

pixeltable/functions/huggingface.py
christopherpestano: Renamed to `sam3_for_segmentation` and `sam3_for_video_segmentation`, kept `sam_automatic_mask_generation` as version agnostic as it actually does accept samv2 as a model arg.

pixeltable/functions/huggingface.py
christopherpestano: Unified into generic `SamForSegmentationResponse`, which is model agnostic since `sam_automatic_mask_generation()` can use multiple sam versions.

pixeltable/functions/image.py
christopherpestano: Added

pixeltable/functions/image.py
christopherpestano: My bad I thought that contract was implicit, added `(requires_order_by=True)`. I suppose there isn't a real reason for users to want to stitch something out of order.

pixeltable/functions/vision.py
christopherpestano: Added both

pixeltable/functions/vision.py
christopherpestano: Added

pixeltable/exprs/function_call.py
christopherpestano: FunctionCall is never polymorphic see `__init__()` where there is an explicit assert not fn.is_polymorphic. The signature is resolved at the udf call site.

pixeltable/functions/huggingface.py
christopherpestano: Good point, missed this when I converted the function version of the iterator I had just copied the entire body into _track(). Moved the dep check into init

christopherpestano: Moved

pixeltable/functions/vision.py
christopherpestano: Added docstring outlining that this is an image overlay.

pixeltable/functions/image.py
christopherpestano: Added trailing comma

tests/functions/test_huggingface.py
christopherpestano: Added test scenario that passes fps and asserts on number of output frames.

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,473 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
 
 
+class SamForSegmentationResponse(TypedDict):
+    scores: pxt.Array[(None,), pxt.Float]
+    boxes: pxt.Array[(None, 4), pxt.Float]
+    masks: pxt.Array[(None, None, None), pxt.Bool]
+
+
+@pxt.udf
+def sam3_for_segmentation(
```
aaron-siegel, line 397: Just `sam3()`? (We usually follow the Hugging Face class names verbatim, so that we're not improsiving/making things up in ad hoc ways)

EDIT: After reading the entire PR, I understand the reasons for calling this `sam3_for_segmentation()`. Still, worth another pass through the names for our UDFs and response dicts to ensure we're being consistent (both internally consistent and consistent w/ the nomenclature in the HF docs).

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,473 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+    __Requirements:__
+
+    - `pip install torch transformers`
+    - `facebook/sam3` is a gated repository. Request access on its
+        [model page](https://huggingface.co/facebook/sam3), then authenticate with
+        `huggingface-cli login` (or set the `HF_TOKEN` environment variable) before calling this UDF.
+
+    Args:
+        image: The image to segment.
+        model_id: The pretrained SAM 3 model to use (default `facebook/sam3`).
```
aaron-siegel: No need to state the defaults in docstrings; the doc builder will note them automatically.

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,473 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+        input_boxes_labels: Optional list of integer labels for each box: `1` for positive
+            (include this region), `0` for negative (exclude this region). If `input_boxes` is
+            provided and this is `None`, all boxes default to positive.
+        threshold: Confidence threshold for filtering detections.
+        mask_threshold: Threshold applied to mask logits to produce binary masks.
+        revision: The specific model revision to use (e.g., a branch, tag, or git identifier). If not
+            specified, uses the default revision for the model.
+
+    Returns:
+        A `SamForSegmentationResponse` containing:
```
aaron-siegel, line 439: cross-link to doc page for response dict

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,473 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
     return output_list
 
 
+class SamForSegmentationResponse(TypedDict):
```
aaron-siegel, line 381: Missing docstring

Also: `Sam3ForSegmentationResponse`

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,473 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+        >>> tbl.add_computed_column(
+        ...     seg=sam3_for_segmentation(
+        ...         tbl.image,
+        ...         text='handle',
+        ...         input_boxes=[[40, 183, 318, 204]],
+        ...         input_boxes_labels=[0],
+        ...     )
+        ... )
+    """
+    env.Env.get().require_package('transformers')
```
aaron-siegel, line 472: 'torch', 'transformers' ?

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,473 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+    if masks_np.dtype != np.bool_:
+        masks_np = masks_np.astype(bool)
+
+    return SamForSegmentationResponse(
+        scores=result['scores'].cpu().numpy(), boxes=result['boxes'].cpu().numpy(), masks=masks_np
+    )
+
+
+@pxt.udf
+def sam_automatic_mask_generation(
```
aaron-siegel, line 531: Ok, now I see why `sam3` wasn't entirely sufficient; but, there's an inconsistency here; this one starts with `sam_` but the previous one `sam3_`. Is that because this is explicitly compatible with SAM1/2 models? If so, state that in the docs.

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,473 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+    stability_score_thresh: float = 0.95,
+    crops_n_layers: int = 0,
+    revision: str | None = None,
+) -> SamForSegmentationResponse:
+    """
+    Segments every object in an image with SAM (Segment Anything Model), with no prompt. `model_id` should
+    be a reference to a pretrained
+    [SAM Model](https://huggingface.co/docs/transformers/model_doc/sam3) such as `facebook/sam3`.
+
+    Unlike [`sam3_for_segmentation`][pixeltable.functions.huggingface.sam3_for_segmentation], which requires a concept
```
aaron-siegel: The docs are assuming `sam3_for_segmentation` as a prior reference point, which doesn't necessarily reflect the order in which the user encounters them.

I'd advocate for using these docstrings to simply describe the behavior of each UDF. If it's important to compare/contrast different UDFs, put an overview of the various UDFs in the module docstring (and optionally link to it from individual UDF docstrings).

pixeltable/functions/huggingface.py
```diff
@@ -376,6 +378,473 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
...
+            discarded.
+        stability_score_thresh: Quality threshold on mask stability under mask-logit threshold changes; masks
+            scoring below it are discarded.
+        crops_n_layers: Number of additional image-crop layers to run. Values above `0` improve recall on
+            small objects at the cost of speed.
+        revision: The specific model revision to use (e.g., a branch, tag, or git identifier). If not
+            specified, uses the default revision for the model.
+
+    Returns:
+        A `Sam3AutomaticMaskGenerationResponse` containing:
```
aaron-siegel: That's not what it returns, according to the UDF signature.

tests/functions/test_image.py
```diff
@@ -95,6 +97,74 @@ def test_tile_iterator(self, uses_db: None) -> None:
                 tile = image.crop(box)
                 assert list(result['tile'].getdata()) == list(tile.getdata())
 
+    @pytest.mark.parametrize('overlap', [(0, 0), (10, 10)])
+    @pytest.mark.parametrize('mode', ['RGB', 'L', 'RGBA'])
+    def test_stitch_tiles(self, mode: str, overlap: tuple[int, int], uses_db: None) -> None:
```
aaron-siegel: It'd be great for this E2E roundtrip test to insert multiple images.

aaron-siegel (changes requested):

pixeltable/functions/huggingface.py
christopherpestano: My bad, was not aware of that feature from the builder.

pixeltable/functions/huggingface.py
christopherpestano: Fixed, I had assumed that transformers had torch as a required dependency.

pixeltable/functions/huggingface.py
christopherpestano: My bad I had unified it with sam3 without fixing the docstring.

pixeltable/functions/huggingface.py
christopherpestano: The _for_segmentation suffix is there to distinguish it from sam_automatic_mask_generation: a bare sam3() next to sam_automatic_mask_generation would read as two different models, even though both can run facebook/sam3.

pixeltable/functions/huggingface.py
christopherpestano: This does not currently work for the huggingface module: its UDFs are auto discovered by the doc generator, and auto discovery does not pick up classes, so the target page is never generated and the link comes out broken. Filed PXT-1235 to render response dicts properly: https://pixeltable.atlassian.net/browse/PXT-1235

pixeltable/functions/huggingface.py
christopherpestano: Added a docstring.

pixeltable/functions/huggingface.py
christopherpestano: Correct, sam_automatic_mask_generation works with any SAM version that supports automatic mask generation, including SAM 1 and SAM 2 checkpoints. The docstring now states that explicitly and mentions facebook/sam2-hiera-large as an example.

pixeltable/functions/huggingface.py
christopherpestano: Agreed. The docstrings no longer use sam3_for_segmentation as a prior reference point; sam_automatic_mask_generation and the video iterator now describe their behavior standalone.

tests/functions/test_image.py
christopherpestano: Done, the roundtrip test now inserts the full test image set via get_image_files and verifies byte exact reconstruction of every image.

docs/release/howto/cookbooks/images/img-promptable-segmentation.ipynb
```diff
@@ -37,7 +37,7 @@
    "source": [
     "## Solution\n",
     "\n",
-    "Use the `sam_for_segmentation` UDF in `pixeltable.functions.huggingface`. It returns a typed result with one entry per detected instance: a confidence `score`, a bounding `box`, and a binary `mask`. The `overlay_segmentation` vision UDF accepts those masks directly, so you can visualize results without any glue code.\n",
+    "Use the `sam3_for_segmentation` UDF in `pixeltable.functions.huggingface`. It returns a typed result with one entry per detected instance: a confidence `score`, a bounding `box`, and a binary `mask`. The `overlay_segmentation` vision UDF accepts those masks directly, so you can visualize results without any glue code.\n",
```
mkornacker: Use hyperlinks and correct formatting. See string.py for examples.

pixeltable/functions/huggingface.py
```diff
@@ -378,14 +378,22 @@ def detr_for_segmentation(image: Batch[PIL.Image.Image], *, model_id: str, thres
     return output_list
 
 
-class Sam3ForSegmentationResponse(TypedDict):
+class SamForSegmentationResponse(TypedDict):
+    """
+    Instance segmentation results, as returned by
+    [`sam3_for_segmentation()`][pixeltable.functions.huggingface.sam3_for_segmentation] or
+    [`sam_automatic_mask_generation()`][pixeltable.functions.huggingface.sam_automatic_mask_generation]:
```
mkornacker: We typically inline the comments into the field declarations. See pixeltable/types.py.

mkornacker (approved):

pixeltable/functions/huggingface.py
christopherpestano: Sure I can do that, changed from top level docstring to per field.

docs/release/howto/cookbooks/images/img-promptable-segmentation.ipynb
christopherpestano: Changed to use the same hyperlinks formatting as `working-with-fiftyone.ipynb`.

aaron-siegel (approved):

PR #1338, opened 2026-05-12
feat: add LiteLLM as AI gateway provider

RheagalFire:
## Summary

- Adds LiteLLM as a new provider integration, giving users access to 100+ LLM providers through a single `chat_completions` UDF.
- Follows the existing provider UDF pattern (mirrors `deepseek.py` / `openrouter.py`). Drop-in, no changes to existing providers.

## Motivation

Pixeltable already has excellent coverage of major providers (OpenAI, Anthropic, Gemini, Bedrock, Groq, Deepseek, Mistral, etc.) and OpenRouter as a routing service. LiteLLM complements these by providing a lightweight Python SDK that calls provider APIs directly using your own keys. This is useful for:

- **Providers without dedicated Pixeltable integrations** (Cohere, AI21, Perplexity, SambaNova, Cerebras, etc.)
- **Swapping providers without changing code** - just change the model string (e.g. `anthropic/claude-sonnet-4-5` to `openai/gpt-4o`)
- **Self-hosted endpoints** (vLLM, TGI, custom OpenAI-compatible servers)

Unlike OpenRouter (which routes through a third-party service requiring an OpenRouter API key), LiteLLM calls provider APIs directly.

## Changes

- `pixeltable/functions/litellm.py` - new provider module with `chat_completions` async UDF
- `pixeltable/functions/__init__.py` - registered in import list
- `pyproject.toml` - added `litellm>=1.55.0,<1.85` as dependency

## Implementation details

- **Async**: Uses `litellm.acompletion()` to match Pixeltable's async UDF pattern.
- **`drop_params=True`**: Silently drops provider-unsupported kwargs, preventing cross-provider failures.
- **OpenAI-compatible response**: `result.model_dump(mode='json')` returns the same dict format as the existing OpenAI/Groq/Deepseek UDFs.
- **Rate limiting**: Uses `resource_pool='request-rate:litellm'` for Pixeltable's built-in request throttling.
- **Provider auth**: LiteLLM reads provider-specific env vars automatically (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `COHERE_API_KEY`, etc.).
- **No client registration needed**: Unlike other providers, LiteLLM doesn't require a persistent client object. The `litellm.acompletion()` function manages connections internally.

## Example usage

```python
import pixeltable as pxt
from pixeltable.functions.litellm import chat_completions

t = pxt.create_table('my_table', {'prompt': pxt.String})

# Use Anthropic Claude via LiteLLM
messages = [{'role': 'user', 'content': t.prompt}]
t.add_computed_column(
    response=chat_completions(messages, model='anthropic/claude-sonnet-4-5')
)

# Use a provider without a dedicated Pixeltable integration
t.add_computed_column(
    cohere_response=chat_completions(messages, model='cohere/command-r-plus')
)

# With additional parameters
t.add_computed_column(
    response2=chat_completions(
        messages,
        model='openai/gpt-4o',
        model_kwargs={'temperature': 0.7, 'max_tokens': 500},
    )
)
```

## Tests

**Live E2E test** against Anthropic Claude (`anthropic/claude-sonnet-4-5`):

```python
>>> result = await litellm.acompletion(
...     model='anthropic/claude-sonnet-4-5',
...     messages=[{'role': 'user', 'content': 'What is 2+2? Answer with just the number.'}],
...     drop_params=True,
... )
>>> d = result.model_dump(mode='json')
>>> d['choices'][0]['message']['content']
'4'
>>> d['usage']
{'completion_tokens': 5, 'prompt_tokens': 20, 'total_tokens': 25, ...}
```

Response format is OpenAI-compatible with standard `choices`, `usage`, and `model` keys.

## Risk / Compatibility

- Additive only. Existing providers untouched.
- `litellm>=1.55.0,<1.85` added as direct dependency, consistent with how other provider SDKs are declared.
- Response format is OpenAI-compatible, so downstream code processing `chat_completions` output works unchanged.

### DIFF
--- a/pixeltable/functions/litellm.py
+++ b/pixeltable/functions/litellm.py
@@ -1,0 +1,92 @@
+"""
+Pixeltable UDFs that wrap the LiteLLM SDK.
+
+LiteLLM provides a unified interface to 100+ LLM providers (Anthropic, Bedrock, Vertex AI,
+Cohere, Mistral, etc.) through a single `completion()` call. The provider is specified via
+the model string (e.g. `anthropic/claude-sonnet-4-5`, `bedrock/anthropic.claude-v2`).
+
+In order to use these UDFs, you must first `pip install litellm` and set the appropriate
+provider-specific environment variables (e.g. `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`).
+See https://docs.litellm.ai/docs/providers for the full list of supported providers.
+"""
+
+from typing import Any
+
+import pixeltable as pxt
+from pixeltable.env import Env
+from pixeltable.utils.code import local_public_names
+
+
+@pxt.udf(is_deterministic=False, resource_pool='request-rate:litellm')
+async def chat_completions(
+    messages: list,
+    *,
+    model: str,
+    model_kwargs: dict[str, Any] | None = None,
+) -> dict:
+    """
+    Creates a model response for the given chat conversation using LiteLLM.
+
+    LiteLLM routes the request to the correct provider based on the model string.
+    For example, `anthropic/claude-sonnet-4-5` routes to Anthropic, `bedrock/anthropic.claude-v2`
+    routes to AWS Bedrock, and `vertex_ai/gemini-pro` routes to Google Vertex AI.
+
+    For a full list of supported models and providers, see: <https://docs.litellm.ai/docs/providers>
+
+    Request throttling:
+    Applies the rate limit set in the config (section `litellm`, key `rate_limit`). If no rate
+    limit is configured, uses a default of 600 RPM.
+
+    __Requirements:__
+
+    - `pip install litellm`
+
+    Args:
+        messages: A list of messages comprising the conversation so far.
+        model: The model to use, prefixed with the provider name
+            (e.g. `'anthropic/claude-sonnet-4-5'`, `'openai/gpt-4o'`, `'bedrock/anthropic.claude-v2'`).
+        model_kwargs: Additional keyword args passed to `litellm.acompletion()`.
+            For details, see: <https://docs.litellm.ai/docs/completion/input>
+
+    Returns:
+        A dictionary containing the response and other metadata in OpenAI-compatible format.
+
+    Examples:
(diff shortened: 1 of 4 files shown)


pixeltable/functions/litellm.py
```diff
@@ -0,0 +1,92 @@
...
+        ... )
+    """
+    if model_kwargs is None:
+        model_kwargs = {}
+
+    Env.get().require_package('litellm')
+
+    import litellm
+
+    result = await litellm.acompletion(
```
christopherpestano, line 79: One thing to note here is that this goes around the internal pixeltable rate limiting infrastructure. LiteLLM manages token estimation, multi consumer, and retries internally. Worth a further look to see if this fits in architecturally with pixeltable's overall design, we may just be externalizing something we already implement (though with the benefit of not having to maintain the model list ourselves).

pixeltable/functions/litellm.py
```diff
@@ -0,0 +1,92 @@
...
+
+from typing import Any
+
+import pixeltable as pxt
+from pixeltable.env import Env
+from pixeltable.utils.code import local_public_names
+
+
+@pxt.udf(is_deterministic=False, resource_pool='request-rate:litellm')
+async def chat_completions(
```
christopherpestano, line 21: Code should provide unit tests in the format of tests/functions/test_litellm.py, as well as those tests should be marked with remote_api (but not expensive). Test coverage of any new code (especially public facing code) is required in this repo.

christopherpestano: It should be easy enough to do at least an example showing it doing the same sort of stuff in test_openai/gemini

pixeltable/functions/litellm.py
```diff
@@ -0,0 +1,92 @@
...
+    messages: list,
+    *,
+    model: str,
+    model_kwargs: dict[str, Any] | None = None,
+) -> dict:
```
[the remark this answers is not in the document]
christopherpestano: This is correct by copilot

PR #1339, opened 2026-05-13
Serving config updates

aaron-siegel:
- Use standard `module:attr` pattern for query specs in config
- Remove boilerplate `modules` config section


pixeltable/serving/_config.py
mkornacker: Rename appropriately.

mkornacker (approved):

PR #1345, opened 2026-05-15
[PXT-1158][PXT-1162] Fix retargeting to stop creating ghost columns

sergey-mkhitaryan:
When `TableVersion` initializes column value expressions from metadata, instead of initializing ColumnRefs and SimilarityExprs using the latest table version, and later retargeting them, pass the target table versions right to `Expr.from_dict()` so that it can deserialize metadata into Exprs that immediately target the correct table version.

Also fixes PXT-1162: `SimilarityExpr` support for snapshots


pixeltable/catalog/table_version.py
```diff
@@ -482,6 +482,7 @@ def _init_schema(self) -> None:
             # "anchored" TableVersionPath.
             assert self.path is not None
             tvp = self.path
+        target_tbl_versions = {tvh.id: tvh.get() for tvh in tvp.get_tbl_versions()} if tvp is not None else None
```
aaron-siegel, line 485: This seems to be deterministically recoverable from a `TableVersionPath`? If so, why not keep the `TableVersionPath` as the reference object and do this unpacking inside Expr.from_dict()?

aaron-siegel (approved):

pixeltable/catalog/table_version.py
sergey-mkhitaryan: `tvp.get_tbl_versions()` returns a list, but retargeting looks up by table id, so something somewhere needs to convert it to a map. We probably don't want to do that in every Expr that requires retargeting.

PR #1357, opened 2026-05-18
PXT-1068: recompile SELECT * query

mkornacker:
- Query resolves select list at execution time if no explicit select list
- FastAPIRouter.add_query_route() freezes the select list of the @pxt.query template if it doesn't have one
- Table.limit()/distinct()/sample() now fail fast on dropped tables


pixeltable/_query.py
```diff
@@ -285,10 +285,20 @@ class ResultCursor(Iterable[Row]):
...
     def __init__(self, query: Query):
         self._query = query
-        self._row_iterator: Generator[list[Any], None, None] | None = None
-        self._columns: dict[str, int] = {name: i for i, name in enumerate(query.schema)}
+        self._row_iterator = None
+        # TODO: fix the race condition:
+        # - we set _schema/_columns here
+        # - a schema change takes places subsequently, but before open()
+        self._schema = query.schema
+        self._columns = {name: i for i, name in enumerate(self._schema)}
```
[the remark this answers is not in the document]
mkornacker: Already left a todo.

pixeltable/catalog/catalog.py
```diff
@@ -263,6 +263,16 @@ def _active_tbl_clause(
     def _dropped_tbl_error_msg(self, tbl_id: UUID) -> str:
         return f'Table was dropped (no record found for {tbl_id})'
 
+    def assert_tbls_exist(self, tbl_ids: Collection[UUID]) -> None:
```
aaron-siegel: `raise_unless_tables_exist()` would be a clearer name (it's user-level error checking, not assertion logic)

aaron-siegel (approved):

pixeltable/catalog/catalog.py
mkornacker: Good point. I'll rename to validate_..., which we have used elsewhere to indicate user-level error checking.

PR #1358, opened 2026-05-19
Twelvelabs async file uploads

aaron-siegel:


pixeltable/functions/twelvelabs.py
```diff
@@ -17,20 +20,51 @@
...
+        A list of asset IDs corresponding to the uploaded files.
+    """
+    client = _twelvelabs_client()
+    uploaded: list[str] = []
+
+    try:
+        if len(files) > 0:
+            tasks: list[Coroutine[Any, Any, 'UploadResult']] = []
+            for file in files:
+                tasks.append(client.multipart_upload.upload_file(file_path=file, file_type=input_type))  # type: ignore[attr-defined]
```
sergey-mkhitaryan: should there be a global semaphore that limits the concurrency of these uploads?

pixeltable/functions/twelvelabs.py
```diff
@@ -17,20 +20,51 @@
...
+    after use.
+
+    Returns:
+        A list of asset IDs corresponding to the uploaded files.
+    """
+    client = _twelvelabs_client()
+    uploaded: list[str] = []
+
+    try:
+        if len(files) > 0:
```
sergey-mkhitaryan: I'd special case this at the top of the function, immediately `yield []` and return. Get rid of this condition, and the one in `finally`.

pixeltable/functions/twelvelabs.py
```diff
@@ -162,27 +184,68 @@ async def _(
...
-    with open(video, 'rb') as fp:
-        b64_str = b64encode(fp.read()).decode('utf-8')
-        res = await cl.embed.v_2.create(
-            input_type='video',
-            model_name=model_name,
-            video=twelvelabs.VideoInputRequest(
+    size_bytes = os.stat(file_path).st_size
+    res: twelvelabs.EmbeddingSuccessResponse
+
+    if size_bytes > TWELVELABS_INLINE_LIMIT_BYTES:
```
sergey-mkhitaryan, line 215: I believe there are good reasons to go much lower than 4MB here:
* loading a 4MB file, encoding it to base64 and putting it on the wire creates some memory pressure
* depending on the implementation and the transport configuration of the asset uploads, they can be more reliable than a somewhat large HTTP post request.
* it's not clear if the 5MB limit they have in their docs applies to raw data or base64 encoding. If the latter, 4MB+base64 overhead can exceed the limit. but I guess it's more likely the former.

my feeling is that the limit for inline should be 1-2MB tops

pixeltable/functions/twelvelabs.py
```diff
@@ -17,20 +20,51 @@
...
+    client = _twelvelabs_client()
+    uploaded: list[str] = []
+
+    try:
+        if len(files) > 0:
+            tasks: list[Coroutine[Any, Any, 'UploadResult']] = []
+            for file in files:
+                tasks.append(client.multipart_upload.upload_file(file_path=file, file_type=input_type))  # type: ignore[attr-defined]
+            upload_results = await asyncio.gather(*tasks)
+            uploaded = [u.asset_id for u in upload_results]
```
sergey-mkhitaryan: if one of the uploads fails, none of them will be cleaned up. are we ok with that?

sergey-mkhitaryan (approved):

pixeltable/functions/twelvelabs.py
aaron-siegel: Right now `files` always has length 1 for twelvelabs. (I was mirroring the pattern in Gemini for which there can be multiple files, and left in support for that because there was no extra work needed for it, even if we don't end up using it.)

We could think about some unified framework for these uploaders that's smarter about that sort of thing, as a future project.

pixeltable/functions/twelvelabs.py
aaron-siegel: Done.

pixeltable/functions/twelvelabs.py
aaron-siegel: Good point about the 5 MB ambiguity. I switched it to 2 MB. We could lower it further but let's maybe ask if they have a recommendation.

pixeltable/functions/twelvelabs.py
aaron-siegel: I think that's fine for now. The main reason for the cleanup is to ensure the storage host doesn't get slammed during normal operation. When failures occur they should be sporadic and if this causes a few orphaned files to be left over, they'll get TTL'ed.

We can revisit if it becomes a problem or if/when we implement more sophisticated shared logic

PR #1363, opened 2026-05-21
Fix tests that depend on non-deterministic order

sergey-mkhitaryan:
A few tests that I decide not to change are `TestPackager`, `TestTimestamp::test_timestamp_zones`, and `TestBridge`. Those are not trivial to fix, and that's not worth doing yet.


sergey-mkhitaryan:
Found more of these. Will fix them all in one pr.

mkornacker:
Something to consider: change ReloadTester.run_query() to check that the query has an order_by. Protects against future backsliding.

mkornacker (approved):

PR #1364, opened 2026-05-22
Class-based Table declarators

aaron-siegel:
Introduces class-based ("dataclass-style") Table declarators via a new `TableModelMetaclass`. The metaclass has two instances, `TableModel` and `ViewModel`, that are used to declare tables and views, respectively. A model gets resolved to a `Table` instance when any action is taken on it, and most of the functionality on `TableModel` is a pass-through to that underlying `Table`.

Columns are referenced via a public-facing `pxt.Column` declarator that creates references to "abstract columns" whose type is `InvalidType`. The types get filled in appropriately when the model resolves to a `Table`.

`Expr.substitute()` has been reimplemented to permit type-changing substitutions in all expression classes.

Table methods (`select()`, `collect()`, `where()`, etc.) can be invoked directly on `TableModel`. In order to avoid having to maintain 20+ method signatures in parallel, a mechanism is introduced where (i) selected table methods are passed through to the underlying `Table` class via `TableModelMetaclass.getattr()`; and (ii) a `model.pyi` file is auto-generated via a new `generate_type_stubs.py` tool, based on the signatures in `pxt.Table`, ensuring that the signatures are both auto-synced and are cognizable to static type checkers.

Currently, the following are *not* implemented:

- Validation of the model against *existing* tables
- Auto-migration of schemas in response to model changes

That will be follow-up work.


pixeltable/catalog/model.py
```diff
@@ -0,0 +1,388 @@
...
+
+        # Now process the declarations to build column and index dictionaries.
+        columns: dict[str, ColumnSpec] = {}
+        indexes: dict[str, EmbeddingIndex] = {}
+        for attr_name, (annotation, value) in decls.items():
+            if value is None:
+                assert annotation is not None
+                # Pure type annotation; no value
+                columns[attr_name] = ColumnSpec(type=annotation)
+            elif isinstance(value, EmbeddingIndex):
```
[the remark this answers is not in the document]
aaron-siegel: this is completely wrong, dude

pixeltable/catalog/model.py
```diff
@@ -0,0 +1,388 @@
...
+
+from .catalog import retry_loop
+from .globals import is_valid_identifier
+from .table import Table
+
+if TYPE_CHECKING:
+    import pixeltable as pxt
+
+# Table methods exposed as class-level operations on the model.
+FORWARDED_TABLE_METHODS: frozenset[str] = frozenset(
```
sergey-mkhitaryan, line 19: I assume there's a reason why this isn't determined at runtime? What is it?

pixeltable/catalog/model.py
```diff
@@ -0,0 +1,388 @@
...
+        'sync',
+        'tail',
+        'unlink_external_stores',
+        'update',
+        'where',
+    )
+)
+
+# Sanity check to guard against drift in the SDK surface.
+assert all(hasattr(Table, method) for method in FORWARDED_TABLE_METHODS)
```
sergey-mkhitaryan, line 53: I think this should be a test instead. Also we may want to test the opposite: that the Table functions are all here.

tests/test_table_model.py
```diff
@@ -0,0 +1,422 @@
...
+            .order_by(ExampleTableModel.id)
+            .collect()
+        )
+        assert list(results) == [
+            {'id': 1, 'descr': 'Name: Alice'},
+            {'id': 2, 'descr': 'Name: Bob'},
+            {'id': 3, 'descr': 'Name: Charlie'},
+        ]
+
+    def test_all_table_exprs(self, uses_db: None) -> None:
```
mkornacker, line 223: This test would be a lot more meaningful if it
- created the same table via add_computed_column()
- inserted the same rows into both tables
- verified that they contain the same data

tests/test_table_model.py
```diff
@@ -0,0 +1,422 @@
...
+            clip_idx = EmbeddingIndex(Column.img, embedding=dummy_embedding.using(n=768))
+
+        _ = ExampleTableModel.create()
+
+        spec: Any
+        match spec_type:
+            case 'model':
+                spec = ExampleTableModel
+            case 'name':
+                spec = 'test_table'
```
mkornacker, line 308: Is this really useful, over just the model?

tests/test_table_model.py
```diff
@@ -0,0 +1,422 @@
...
+            case 'query':
+                spec = ExampleTableModel.select(ExampleTableModel.value, ExampleTableModel.img).where(
+                    ExampleTableModel.value > 0.5  # type: ignore[arg-type]
+                )
+            case 'table':
+                spec = ExampleTableModel.table
```
mkornacker, line 314: Same here.

tests/test_table_model.py
```diff
@@ -0,0 +1,422 @@
...
+                spec = ExampleViewModel.table
+
+        class ExampleViewModel2(pxt.ViewModel):
+            __table_name__ = 'test_view_2'
+            __base_table__ = spec
+
+            subview_col_1 = Column.img.rotate(180)
+            subview_col_2 = Column.view_col_1.rotate(270)
+
+        _ = ExampleViewModel2.create()
```
mkornacker, line 345: Also missing evidence that this leads to the same outcome as creating everything via create_table/view/add_computed_column

tests/test_table_model.py
```diff
@@ -0,0 +1,422 @@
...
+                __table_name__ = 'bad_iter_ref'
+                __base_table__ = 'unused'
+                __iterator__ = 'not a generating function call'
+
+        # Type annotation conflicts with the `type=` argument in `Column()`.
+        with pxt_raises(excs.ErrorCode.INVALID_SCHEMA, match='conflicts with the `type=` argument'):
+
+            class TypeConflict(pxt.TableModel):
+                __table_name__ = 'type_conflict'
+                name: pxt.Int = Column(type=pxt.String)  # type: ignore[assignment]
```
mkornacker, line 382: Not sure what I'm looking at.

pixeltable/catalog/table.py
```diff
@@ -1036,25 +1039,35 @@ def add_embedding_index(
         rows are inserted into the table.
 
         To add an embedding index, specify the column to be indexed and, if the column is not an `Array` column, an
-        embedding UDF. `String`, `Image`, `Video`, `Audio` and `Array` columns are currently supported.
+        embedding UDF. `String`, `Image`, `Audio`, `Video`, `Document`, and `Array` columns are currently supported.
```
mkornacker, line 1042: Why is this here?

pixeltable/catalog/model.py
```diff
@@ -0,0 +1,388 @@
...
+                expr = col_spec.get('value')
+                if expr is not None:
+                    realized_expr: exprs.Expr = expr.copy().substitute(subst_dict)
+                    residual_placeholders = list(realized_expr.subexprs(_PlaceholderColumnRef))
+                    if len(residual_placeholders) > 0:
+                        raise excs.RequestError(
+                            excs.ErrorCode.INVALID_SCHEMA,
+                            f'Computed column {col_name!r} references undefined column(s): {residual_placeholders}',
+                        )
+                    tbl.add_computed_column(
```
mkornacker, line 338: The entire table should be created in a single transaction, with a single version, via TableVersion.create_initial_md().

pixeltable/catalog/model.py
```diff
@@ -0,0 +1,388 @@
...
+        'update',
+        'where',
+    )
+)
+
+# Sanity check to guard against drift in the SDK surface.
+assert all(hasattr(Table, method) for method in FORWARDED_TABLE_METHODS)
+
+
+class _PlaceholderColumnRef(exprs.Expr):
```
mkornacker, line 56: You can use Variable for this purpose.

pixeltable/catalog/model.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def __getattr__(self, key: str) -> _PlaceholderColumnRef:
+        if not isinstance(key, str) or not is_valid_identifier(key):
+            raise AttributeError(f'Invalid column name: {key}')
+        return _PlaceholderColumnRef(key)
+
+    def __hasattr__(self, key: str) -> bool:
+        return isinstance(key, str) and is_valid_identifier(key)
+
+    def __call__(
```
mkornacker, line 103: I found that confusing - that Column.<> is a column reference but Column(...) is a column definition.

mkornacker (changes requested):

PR #1368, opened 2026-05-22
Typed JSON return values for audio.get_metadata() and video.get_metadata()

aaron-siegel:
- Introduces a `ContainerMetadata` typed dict for `audio.get_metadata()` and `video.get_metadata()`
- Relocates helper function to `pixeltable.functions.util`
- Docstrings for all new TypedDicts
- Doc update: pretty-print JSON return values when they're built from a TypedDict (to avoid massive doc clutter for complex JSON types)
- Fixes a related bug in doctools (signatures for UDFs with typed json parameters were not rendering properly)

### DIFF
--- a/docs/public_api.opml
+++ b/docs/public_api.opml
@@ -348,6 +348,9 @@
           <outline text="func|pixeltable.functions.vision.bboxes_resize_canvas" />
           <outline text="func|pixeltable.functions.vision.overlay_segmentation" />
         </outline>
+        <outline text="class|pixeltable.functions.CodecContextMetadata" />
+        <outline text="class|pixeltable.functions.ContainerMetadata" />
+        <outline text="class|pixeltable.functions.StreamMetadata" />
       </outline>
 
       <!-- AI Integrations group -->
--- a/pixeltable/functions/util.py
+++ b/pixeltable/functions/util.py
@@ -1,3 +1,8 @@
+from typing import TypedDict
+
+import av
+import av.container
+import av.stream
 import PIL.Image
 
 from pixeltable.config import Config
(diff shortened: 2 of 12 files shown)


pixeltable/functions/util.py
```diff
@@ -31,3 +36,127 @@ def normalize_image_mode(image: PIL.Image.Image) -> PIL.Image.Image:
     if image.mode == 'LA':
         return image.convert('RGBA')
     return image
+
+
+class CodecContextMetadata(TypedDict, total=False):
```
mkornacker, line 41: Should this live in the top level or a new av (or media) module?

docs/public_api.opml
```diff
@@ -348,6 +348,9 @@
           <outline text="func|pixeltable.functions.vision.bboxes_resize_canvas" />
           <outline text="func|pixeltable.functions.vision.overlay_segmentation" />
         </outline>
+        <outline text="class|pixeltable.functions.CodecContextMetadata" />
```
mkornacker, line 351: Are we going to add each new typeddict we introduce for a udf to the docs? Is this going to overwhelm the docs? Or is there some other way of including them in the docs (eg, the docs-generating code that processes udfs could auto-generate a Returns section with the typeddict).

pixeltable/functions/util.py
```diff
@@ -31,3 +36,127 @@ def normalize_image_mode(image: PIL.Image.Image) -> PIL.Image.Image:
...
+    """Overall bit rate of the container in bits per second, or `None` if unknown."""
+    size: int | None
+    """Size of the container in bytes, or `None` if unknown."""
+    metadata: dict[str, str]
+    """Additional container-level metadata tags (e.g. title, encoder)."""
+    streams: list[StreamMetadata]
+    """Per-stream metadata for each stream in the container."""
+
+
+def get_metadata(path: str) -> ContainerMetadata:
```
mkornacker, line 103: So far, the source files under functions/ contained udfs and related and we've left shared utility functions in av.py. That seems like a good setup to avoid circular includes.

pixeltable/type_system.py
```diff
@@ -960,9 +965,15 @@ def _validate_literal(self, val: Any) -> None:
 class JsonType(ColumnType):
     type_schema: TypeSchema | None
 
-    def __init__(self, type_schema: TypeSchema | None = None, nullable: bool = False):
+    # pretty_print_name is used only for generating clean documentation; it is not stored in the catalog.
```
mkornacker, line 968: What is it?

pixeltable/utils/av.py
```diff
@@ -28,71 +25,6 @@
 }
 
 
-def get_metadata(path: str) -> dict:
```
mkornacker, line 31: I would really prefer to keep these here to be available to other utility functions. How about we move the typeddicts?

mkornacker (changes requested):

mkornacker:
Overall comment: we should probably have a broader discussion about the mechanics of documenting udfs with typed json return types (and where those types are allowed to reside in the code base).

mkornacker (approved):

docs/public_api.opml
aaron-siegel: The usual approach here would be to include them in the docs.

We could potentially move them to their own section (or a subsection) so the clutter is self-contained?

pixeltable/functions/util.py
aaron-siegel: I'm not sure I agree it's a good idea to be using `get_metadata()` as a shared utility function. Usually when it's used as a utility function we're using it to get one piece of metadata - and this is an awfully inefficient way to do that. I'd argue that the `ContainerMetadata` TypedDict and associated `get_metadata()` really are there to serve the UDF - and that in cases where we need some simpler bit of information it's preferable to query `av` directly (which is now a fairly clean typesafe API, so I'm not sure what's lost by doing that).

pixeltable/utils/av.py
aaron-siegel: See comment above - I'm not sure I agree it's a good idea to dual-purpose these as utility functions.

pixeltable/utils/av.py
```diff
@@ -183,8 +115,9 @@ def get_max_volume_db(path: str) -> float | None:
 
 def has_audio_stream(path: str) -> bool:
     """Check if video has audio stream using PyAV."""
-    md = get_metadata(path)
-    return any(stream['type'] == 'audio' for stream in md['streams'])
+    with av.open(path) as container:
```
aaron-siegel, line 118: Example here: I'd argue this is the preferred way to resolve container properties on non-UDF codepaths.

pixeltable/type_system.py
aaron-siegel: I expanded the comment.

pixeltable/functions/util.py
aaron-siegel: Alternatively, we could introduce a `pxtf.types`.

PR #1377, opened 2026-05-27
Cross reference HTTP serving page from SDK page on FastAPIRouter

apreshill:
Based on user feedback


pixeltable/serving/_fastapi.py
```diff
@@ -156,6 +156,10 @@ def url_for_media(rel_path: str) -> str:
 class FastAPIRouter(fastapi.APIRouter):
     """
     A FastAPI `APIRouter` that exposes Pixeltable table operations as HTTP endpoints.
+
+    `FastAPIRouter` is for apps that already have a FastAPI server. To start an
+    all-in-one Pixeltable HTTP service, use `pxt serve` from the CLI. Learn more
```
pierrebrunelle: not sure what all-in-one means for users? should we remove or be specific?

apreshill: I was using your words from Slack 😉 but I agree- I'll edit to be more specific

pierrebrunelle (approved):

PR #1379, opened 2026-05-27
Cleanup Pixeltable's logging infrastructure

sergey-mkhitaryan:
Use Python built-in logging more idiomatically. Each Pixeltable module gets its own logger name derived from the module path (e.g. `pixeltable.env`, `pixeltable.catalog.catalog`).

`pxt.configure_logging()` is gone, and so is ad-hoc log filtering that existed in `Env`. If the user needs to adjust the log levels, they can do it via logging:

```
logging.getLogger('pixeltable').setLevel(logging.WARNING)
logging.getLogger('pixeltable.env').setLevel(logging.INFO)
logging.getLogger('sqlalchemy.engine').setLevel(logging.DEBUG)
```

The same goes for stdout logging, etc.

Verified manually:
- SQLAlchemy stays quiet by default — `sqlalchemy.engine` is at `WARNING`
- When interacting with Pixeltable in Python REPL, it looks the same as before
- When running in a notebook (using vscode or jupyter), it looks the same as before
- Logging behavior in `pytest` and `pytest -s` is like before
- Stress tests (aka random ops) stdout and file logs look good as before
- `create_test_db_dump.py` stdout looks the same as before


tests/functions/test_gemini.py
```diff
@@ -25,7 +25,7 @@
 )
 from .tool_utils import run_tool_invocations_test
 
-_logger = logging.getLogger('pixeltable')
+_logger = logging.getLogger('pixeltable.test')
```
aaron-siegel: Maybe `pixeltable_tests` so as not to give the illusion that it's actually a subpackage?

tool/create_test_db_dump.py
```diff
@@ -16,13 +16,13 @@
...
 from pixeltable import functions as pxtf, metadata, type_system as ts
-from pixeltable.env import Env
+from pixeltable.env import LOG_FMT_STR, Env
 from pixeltable.func import Batch
 from pixeltable.io.external_store import Project
 from pixeltable.iterators.base import ComponentIterator
 from tool.udfs_for_db_dump import test_array_udf, test_binary_udf, test_date_udf, test_timestamp_udf, test_uuid_udf
 
-_logger = logging.getLogger('pixeltable')
+_logger = logging.getLogger('pixeltable.create_test_db_dump')
```
aaron-siegel: Or `tool.create_test_db_dump`, or just `__name__`

pixeltable/metadata/__init__.py
```diff
@@ -14,8 +14,8 @@
 
 from .schema import SystemInfo, SystemInfoMd
 
-_console_logger = ConsoleLogger(logging.getLogger('pixeltable'))
-_logger = logging.getLogger('pixeltable')
+_logger = logging.getLogger(__name__)
+_console_logger = ConsoleLogger(_logger)
```
aaron-siegel, line 18: Why is this module using its own ConsoleLogger but everyone else is using Env's?

aaron-siegel (approved):
Really glad we're cleaning this up.

pixeltable/metadata/__init__.py
sergey-mkhitaryan: Because it is needed before Env is fully initialized, and calling `Env.get()` can get us in infinite recursion. There is a way to get rid of this -- `upgrade_md()` just needs to take a `ConsoleLogger` as a parameter -- but that's not very elegant either.

PR #1384, opened 2026-06-02
PXT-1183: race in Catalog.create_user()

mkornacker:
- create_user() now runs as a serializable transaction
- adds isolation_level parameter to begin_xact()
- also: converted some concurrency tests to share Table handles


pixeltable/catalog/catalog.py
```diff
@@ -87,6 +87,7 @@ def retry_loop(
     write_tvps: Collection[TableVersionPath] | None = None,
     write_tbl_ids: Collection[UUID] | None = None,
     lock_mutable_tree: bool = False,
+    isolation_level: str | None = None,
```
aaron-siegel: Make this (and related params) a `Literal`?

aaron-siegel (approved):

pixeltable/runtime.py
```diff
@@ -135,13 +135,15 @@ def run(coro: Coroutine[Any, Any, _T]) -> _T:
...
-    def begin_xact(self, *, for_write: bool = False) -> Iterator[sql.Connection]:
+    def begin_xact(self, *, for_write: bool = False, isolation_level: str | None = None) -> Iterator[sql.Connection]:
         """Start or join a database transaction.
 
         Prefer Catalog.begin_xact() unless there is a specific reason to call this directly.
 
         Args:
             for_write: unused (TODO use or remove)
+            isolation_level: if specified, the isolation level for the new transaction; defaults to
+                _XACT_ISOLATION_LEVEL. Can only be set when starting the outermost transaction.
```
sergey-mkhitaryan: `_XACT_ISOLATION_LEVEL` is an internal detail. Just say READ COMMITTED.

pixeltable/runtime.py
```diff
@@ -135,13 +135,15 @@ def run(coro: Coroutine[Any, Any, _T]) -> _T:
         return self._run_coro_executor.submit(run, coro).result()
 
     @contextmanager
-    def begin_xact(self, *, for_write: bool = False) -> Iterator[sql.Connection]:
+    def begin_xact(self, *, for_write: bool = False, isolation_level: str | None = None) -> Iterator[sql.Connection]:
```
sergey-mkhitaryan: should it be `= _XACT_ISOLATION_LEVEL` so that the default behavior is encoded right in the signature?

pixeltable/catalog/catalog.py
```diff
@@ -2897,19 +2903,29 @@ def create_user(self, user: str | None) -> None:
...
         Creates a catalog record (root directory) for the specified user, if one does not already exist.
         """
-        with get_runtime().begin_xact():
+        # - we need to run this as SERIALIZABLE in order to avoid a race when two processes are started against
+        #   an empty store (they both see a count of 0 for the root dir and both create a new root dir)
+        # - the retry loop will handle the serialization failure
+        # - this can only be run inside a new transaction, to ensure the isolation level
+        assert not get_runtime().in_xact, 'create_user() must run as the outermost transaction'
+
+        @retry_loop(for_write=True, isolation_level='SERIALIZABLE')
```
sergey-mkhitaryan: What I don't like about this is that now we have 2 distinct locking/serialization mechanisms for metadata changes:
- drop dir, move, etc. lock the parent dir
- create user relies on serializable

This of course complicates things, makes them difficult to reason about, etc.

Can we try to solve this with just one mechanism? I can suggest a few options:
1. Serializable for all `schema.Dir` ops. Sure there's an overhead, but these ops are not in hot path, so that's probably fine.
2. Parent dir locking (`_acquire_dir_xlock`) for all ops, which is for the most part how it works today. Creating a named user locks the root (null) user. This doesn't solve the race to create the root user, but there are options for that edge case as well, including to do nothing about it, or do it in `pxt.init()`.

pixeltable/catalog/catalog.py
```diff
@@ -238,6 +240,9 @@ def __init__(self) -> None:
         self._tbl_md_read_allowed = False
         self._column_dependencies = {}
         self._column_dependents = None
+        # Register as the runtime's catalog singleton before _init_store(): _init_store() runs through
+        # retry_loop, which resolves get_runtime().catalog and would otherwise re-enter Catalog() here.
+        get_runtime()._catalog = self
```
sergey-mkhitaryan: This is not great:
- a partially initialized Catalog instance escapes its constructor
- it breaks encapsulation by reaching in and modifying other object's internal state
- now there's multiple places where `Runtime._catalog` is set, which makes it harder to understand

I would much rather you roll your own retry loop in `Catalog.create_user()`. Most of the stuff in `@retry_loop` doesn't apply to it anyway. Another alternative is to pass an optional `Catalog` instance to `@retry_loop`, but that's kinda crappy too.

sergey-mkhitaryan (changes requested):

pixeltable/runtime.py
mkornacker: Done.

pixeltable/runtime.py
mkornacker: We typically use None to indicate the default.

pixeltable/catalog/catalog.py
mkornacker: Good idea but out of scope here.

I already have something else in the works for pxt-438.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/utils/fault_injection.py
```diff
@@ -17,6 +17,7 @@ class FaultLocation(Enum):
     IO_EXPORT_ICEBERG_APPEND = auto()
     SCHEDULER_RATE_LIMITS_AEXEC = auto()
     STORE_CREATE_SA_TBL_AFTER_EXISTS_CHECK = auto()
+    CATALOG_CREATE_USER_AFTER_EXISTS_CHECK = auto()
```
sergey-mkhitaryan: please order lexicographically

pixeltable/catalog/catalog.py
sergey-mkhitaryan: I'm not quite buying that this is out of scope, but approving with the assumption that this is temporary.

sergey-mkhitaryan (approved):

pixeltable/utils/fault_injection.py
mkornacker: Done.

PR #1385, opened 2026-06-02
vLLM integration

aaron-siegel:


tests/functions/test_vllm.py
```diff
@@ -0,0 +1,58 @@
+import pixeltable as pxt
+
+from ..utils import rerun, skip_test_if_not_installed, validate_update_status
+
+
+@rerun(reruns=3, reruns_delay=15)
+class TestVllm:
```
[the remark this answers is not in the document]
aaron-siegel: It's not possible to install vllm outside of a CUDA environment

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,187 @@
...
+        _model_cache[key] = vllm.LLM(model=model, **engine_args)
+    return _model_cache[key]
+
+
+def _request_output_to_dict(output: 'vllm.RequestOutput') -> dict:
+    """Convert a vLLM RequestOutput to a JSON-serializable dict, preserving native structure."""
+    return {
+        'request_id': output.request_id,
+        'prompt': output.prompt,
+        'prompt_token_ids': list(output.prompt_token_ids) if output.prompt_token_ids else None,
```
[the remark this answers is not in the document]
aaron-siegel: fixed

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,187 @@
...
+    messages: list[dict],
+    *,
+    model: str,
+    engine_args: dict[str, Any] | None = None,
+    sampling_params: dict[str, Any] | None = None,
+) -> VllmRequestOutput:
+    """
+    Generate a chat completion from a list of messages using vLLM.
+
+    Uses vLLM's high-throughput inference engine for efficient local LLM serving.
```
mkornacker: "high-throughput/efficient" sounds like marketing, which we typically avoid, and aside from that the sentence doesn't convey anything useful.

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,187 @@
...
+
+import pixeltable as pxt
+from pixeltable.env import Env
+from pixeltable.utils.code import local_public_names
+
+if TYPE_CHECKING:
+    import vllm
+
+
+class VllmRequestOutput(TypedDict):
```
mkornacker, line 32: Nice!

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,187 @@
...
+    request_id: str
+    prompt: str | None
+    prompt_token_ids: list[int] | None
+    outputs: list[dict]
```
mkornacker: Does this have more structure?

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,187 @@
...
+    return _model_cache[key]
+
+
+def _request_output_to_dict(output: 'vllm.RequestOutput') -> dict:
+    """Convert a vLLM RequestOutput to a JSON-serializable dict, preserving native structure."""
+    return {
+        'request_id': output.request_id,
+        'prompt': output.prompt,
+        'prompt_token_ids': list(output.prompt_token_ids) if output.prompt_token_ids is not None else None,
+        'outputs': [
```
mkornacker, line 174: Can we also get a typeddict for this? This is presumably the field that is most important to the user (and avoid casts would be much more user-friendly).

scripts/prepare-nb-tests.sh
```diff
@@ -12,6 +12,14 @@ SKIP_NOTEBOOKS=(
     working-with-twelvelabs         # [PXT-1119] Exceeds rate limit
 )
 
+# Check if `nvidia-smi` returns success; if not, skip GPU notebooks too
+if ! nvidia-smi > /dev/null 2>&1; then
+    echo "nvidia-smi not found or not working; skipping GPU notebooks."
+    SKIP_NOTEBOOKS+=(
+        working-with-vllm           # vLLM requires a CUDA environment
```
mkornacker, line 19: Are you sure this is the only one? Any HF model I run locally runs on my gpu.

tests/functions/test_vllm.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+        )
+
+        validate_update_status(t.insert(input='What are some edible species of fish?'), expected_rows=1)
+        result = t.collect()['output'][0]
+        result2 = t.collect()['output2'][0]
+        print(result)
+        assert result['finished'] is True
+        assert len(result['outputs']) > 0
+        assert len(result['outputs'][0]['text']) > 0
```
mkornacker, line 41: Maybe this is out of scope here, but it would be good to validate that the returned type matches what's declared. As it is, the test validates less than what the udf promises.

pixeltable/functions/vllm.py
```diff
@@ -0,0 +1,187 @@
...
+    outputs = llm.generate([prompt], **gen_kwargs)
+    return _request_output_to_dict(outputs[0])
+
+
+def _lookup_model(model: str, engine_args: dict[str, Any]) -> 'vllm.LLM':
+    import vllm
+
+    args_key = json.dumps(engine_args, sort_keys=True, default=str)
+    key = (model, args_key)
+    if key not in _model_cache:
```
mkornacker: This is not thread-safe, but would be with a lock.

Suggested fix:
  _model_cache_lock = threading.Lock()

  def _lookup_model(model: str, engine_args: dict[str, Any]) -> 'vllm.LLM':
      import vllm
      key = (model, json.dumps(engine_args, sort_keys=True, default=str))
      with _model_cache_lock:
          if key not in _model_cache:
              _model_cache[key] = vllm.LLM(model=model, **engine_args)
          return _model_cache[key]


Medium-term, we should probably manage these resources in Env. We have a number of these time bombs sitting around.

mkornacker (changes requested):

pixeltable/functions/vllm.py
aaron-siegel: Removed

pixeltable/functions/vllm.py
aaron-siegel: Done.

scripts/prepare-nb-tests.sh
aaron-siegel: HF models can run on CPU (hence do not require a CUDA environment)

tests/functions/test_vllm.py
aaron-siegel: Do you mean validate the 'text' field is a string, or validate the entire dict? The former is trivial, the latter needs some thought (maybe some new test utilities)

pixeltable/functions/vllm.py
aaron-siegel: Fixed. (Incidentally, code snippets in github comments are much easier to read if you fence them)

mkornacker (approved):

PR #1391, opened 2026-06-03
Verify all necessary locks on retry_loop re-entry

sergey-mkhitaryan:
- If already in a retry loop, validate that the outer one acquired all write locks. Same check that already exists in `begin_xact()`
- Make validation stronger by taking into account `lock_mutable_tree`
- On the happy path, this check should add no new I/O


pixeltable/catalog/catalog.py
```diff
@@ -711,6 +708,26 @@ def _acquire_write_lock(
...
+    def _check_write_locks(
+        self, write_tvps: Collection[TableVersionPath], write_tbl_ids: Collection[UUID], lock_mutable_tree: bool
+    ) -> None:
+        """Asserts that all specified write targets (and their mutable trees, if lock_mutable_tree) are locked."""
+
+        def assert_tree_locked(tbl_id: UUID) -> None:
+            tree = self._get_mutable_tree(tbl_id)
+            assert tree.issubset(self._x_locked_tbl_ids), (
+                f'mutable tree of {tbl_id} not fully locked: {tree - self._x_locked_tbl_ids}'
+            )
```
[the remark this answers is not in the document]
sergey-mkhitaryan: not worth it IMO

pixeltable/catalog/catalog.py
```diff
@@ -711,6 +708,26 @@ def _acquire_write_lock(
                 )
         return locked
 
+    def _check_write_locks(
+        self, write_tvps: Collection[TableVersionPath], write_tbl_ids: Collection[UUID], lock_mutable_tree: bool
```
mkornacker, line 718: re Collection: is this gratuitous generality? set[] is easier to understand. Would set[] have any practical downsides?

mkornacker (approved):

pixeltable/catalog/catalog.py
sergey-mkhitaryan: `@retry_loop` and `begin_xact()` are using `Collection`. While there are no strong reasons against `set` there, some minor reasons include:
* some callers like `Query._output_row_iterator()` or `FastAPIRouter.add_query_route()` pass an already existing `list` object -- they would need to copy it to a `set`
* `Collection` better communicates to the caller how this is being used. No `set` semantics are used.
* most callers will pass a `list` which is marginally faster to create

All in all I think it's a minor inconvenience in `Catalog` that unlocks a more significant convenience for its callers.

PR #1398, opened 2026-06-05
Skip HF-dependent notebooks and unit tests in pull requests

sergey-mkhitaryan:
Github secrets are not available in pr checks, and HuggingFace downloads consistently fail due to rate limiting without a token.

Merge queue, on the other hand, runs in the context of the main repo, which means HF token is available, and we can expect much higher limits.

This pr disables HF-dependent notebooks and tests in pull requests (i.e. when a commit is pushed), but leaves them enabled in merge queue tests. Those tests are now called "expensive", and what used to be "expensive" is now "very expensive". This terminology mirrors the existing tests terminology and its behavior.

Additionally, fixes `test_dir_size_skips_stat_errors` that leaks an injected errors to other tests.


scripts/prepare-nb-tests.sh
```diff
@@ -20,12 +20,29 @@ VERY_EXPENSIVE_NOTEBOOKS=(
     working-with-together           # Poor reliability
 )
 
+# Notebooks that are skipped unless --include-merge-queue is passed.
```
aaron-siegel: Calling this `--include-merge-queue` introduces an unnecessary new CI abstraction (which isn't entirely accurate, since these run in any non-PR context, not just merge queue). In order to keep things combinatorially manageable, I suggest:
- call these EXPENSIVE_NOTEBOOKS
- change the existing `--include-expensive` to `--include-very-expensive`
- change `--include-merge-queue` to `--include-expensive`
That will exactly match the semantics of the existing terminology.

aaron-siegel (changes requested):

mkornacker (approved):

aaron-siegel (approved):

PR #1400, opened 2026-06-07
Generalization of TableVersionPath

mkornacker:
- introduces TablePath as the metadata-only base class for TableVersionPath
- introduces TableMdPath as a concrete metadata-only sibling to TVP
- adds ColumnVersionMd, analogous to the existing TableVersionMd
- converts ColumnRef from Column to ColumnVersionMd

This doesn't fully remove the use of TableVersion instances in expr classes outside of execution-related methods (such as eval()), which we'll need for delegated execution.


aaron-siegel (approved):
My one thought would be to make sure @sergey-mkhitaryan looks at the changes (in particular, are the ColumnVersionMd changes consistent with where #1090 is heading).

PR #1402, opened 2026-06-09
Keep HuggingFace out of PR and merge CI

mkornacker:
Removes HuggingFace model/dataset downloads and notebook runs from PR and merge-queue CI, moving them to the scheduled (weekly) tier, while preserving embedding-index coverage in normal CI.

Changes
- Mark all HuggingFace-using tests very_expensive so they are excluded from PR and merge CI and run only on the scheduled tier
- Stop running notebook tests on PR and merge CI; they now run only on the scheduled tier. The matrix entries are commented out (not deleted) so they are easy to restore.
- Preserve embedding-index coverage without HF downloads

Verification
- Full suite under coverage (PR/merge marker filter): index machinery coverage preserved (embedding_index 95%, similarity_expr 93%); functions/huggingface.py drops to 18% as intended (now weekly-only).
- The only failures are pre-existing environment issues (cv2, yolox, cloud credentials), confirmed to reproduce on a clean tree.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,8 +1,10 @@
 import datetime
 import glob
+import hashlib
 import itertools
 import json
 import logging
+import math
 import os
 import random
 import re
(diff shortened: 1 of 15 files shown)


tests/utils.py
```diff
@@ -1019,3 +1021,47 @@ def cleanup(r: str) -> str:
...
+def _local_image_vec(image: PIL.Image.Image, dim: int) -> np.ndarray:
+    side = math.isqrt(dim)
+    grid = image.convert('L').resize((-(-dim // side), side))  # downsample to a grid of >= dim pixels
+    v = np.asarray(grid, dtype=np.float32).reshape(-1)[:dim]
+    norm = float(np.linalg.norm(v))
+    return v / norm if norm > 0 else v
+
+
+@pxt.udf
+def local_embedding(text: str, *, dim: int = LOCAL_EMBED_DIM) -> pxt.Array[(None,), np.float32]:
```
aaron-siegel, line 1056: Consider switching to use this in contexts that currently use a bespoke dummy_embedding?

aaron-siegel (approved):

tests/utils.py
mkornacker: Done.

PR #1404, opened 2026-06-09
Catalog.mark_modified_tvs is never passed more than one TV handle

sergey-mkhitaryan:


aaron-siegel (approved):
sure, but I don't see what harm the old pattern was doing

PR #1408, opened 2026-06-10
Fix two related issues in Catalog

sergey-mkhitaryan:
Issue 1: Start with a base table and a view. Thread 0 loads the view md, and is about to initialize it when Thread 1 drops it. Thread 0 then proceeds to initialize the view, which involves loading the base table, and fails with AssertionError due to an inconsistent state of the base table and the view -- the view is live, but the base table has no views.

Fix: switch to REPEATABLE READ isolation level. At READ COMMITTED, the base table read, and the view read happen at different snapshots, therefore can be inconsistent with one another.

Issue 2: Start with a base table and a view. Thread 0 begins to drop the view, but pauses inside finalize pending ops (without the exclusive lock). Thread 1 swoops in in the meantime to insert a row into the base table, and finalizes view drop as a side effect. However when it attempts the insert, it gets "table not found".

This happens because view_sn is incremented on the base table when the view deletion is starting, not when the view metadata actually goes away from the database. Therefore view_sn does not perfectly track the metadata state of the views.

Fix: update the base table's view_sn in lock step with the view metadata deletion.


pixeltable/catalog/tbl_ops.py
```diff
@@ -168,12 +169,28 @@ def undo(self, tv: TableVersion | None) -> None:
...
         assert get_runtime().in_xact
-        get_runtime().catalog.delete_tbl_md(uuid.UUID(self.tbl_id))
+        cat = get_runtime().catalog
+        if self.base_tbl_id is not None:
+            try:
+                cat._incr_view_sn(uuid.UUID(self.base_tbl_id))
+            except excs.NotFoundError as e:
+                if e.error_code != excs.ErrorCode.TABLE_NOT_FOUND:
+                    raise
+                # The base may have already been dropped if we're tearing down the entire hierarchy at once.
```
mkornacker: Move above the if, that's what it refers to.

pixeltable/catalog/tbl_ops.py
```diff
@@ -168,12 +169,28 @@ def undo(self, tv: TableVersion | None) -> None:
...
+    """
+    If it's a mutable view that's being dropped, this op will also advance the base table's view_sn. base_tbl_id must be
+    provided in that case.
+    """
+
     needs_tv: ClassVar[bool] = False
     needs_xact: ClassVar[bool] = True
 
+    # Defaults to None so that already existing pending ops can be deserialized. TODO: clean this up later
+    base_tbl_id: str | None = None
```
mkornacker: mutable_base_tbl_id for ultimate clarity (that's an important part of the semantics).

pixeltable/utils/fault_injection.py
```diff
@@ -12,12 +12,12 @@ class FaultLocation(Enum):
     CATALOG_CREATE_USER_AFTER_EXISTS_CHECK = auto()
     CATALOG_CREATE_VIEW_BEFORE_MD_COMMITTED = auto()
     CATALOG_FINALIZE_PENDING_OPS_NON_XACT = auto()
+    CATALOG_LOAD_TBL_VERSION_BEFORE_INIT = auto()
     CATALOG_LOAD_VIEW_OP_EXEC = auto()
     IO_EXPORT_ICEBERG_WRITE_TEMP = auto()
     IO_EXPORT_ICEBERG_BEFORE_SWAP = auto()
     IO_EXPORT_ICEBERG_APPEND = auto()
     SCHEDULER_RATE_LIMITS_AEXEC = auto()
-    STORE_CREATE_SA_TBL_AFTER_EXISTS_CHECK = auto()
```
mkornacker, line 20: Why remove this?

tests/test_catalog.py
```diff
@@ -329,35 +329,61 @@ def create_view() -> None:
         # Verify that the insert was propagated to vb.
         assert pxt.get_table('vb').count() == 1
 
-    def test_load_table_concurrent_drop_store_tbl(self, uses_db: None, fault_injection: None) -> None:
```
mkornacker, line 332: What's wrong with this test?

mkornacker (approved):

pixeltable/utils/fault_injection.py
sergey-mkhitaryan: the test that needed it goes away, see below

tests/test_catalog.py
sergey-mkhitaryan: With REPEATABLE READ it's no longer validating anything interesting. Under READ COMMITTED there was a scenario in which a table could have been observed as existing, and later as non-existing within one transaction. With the isolation change, that condition disappeared.

PR #1414, opened 2026-06-12
Remove data sharing and replica support

mkornacker:
- removes table sharing and the replica-related logic, simplifying the catalog
- migrates pxtfs:// related logic into a new module

Removed
- pxt.publish(), pxt.replicate(), Table.push(), Table.pull(), and the pixeltable/share/ package
- Replica internals: is_replica/is_fragment/anchor_tbl_id, anonymous system base tables, and system paths

Db migration
- Replica-free databases upgrade automatically (obsolete fields dropped).
- A database that still contains replicas is not migrated: the upgrade aborts with the replica paths and instructions for manually converting existing replicas.


pixeltable/catalog/catalog.py
```diff
@@ -1181,7 +1168,7 @@ def get_dir_path(self, dir_id: UUID) -> Path:
                 break
             names.insert(0, dir.md['name'])
             dir_id = dir.parent_id
-        return Path.parse('/'.join(names), allow_empty_path=True, allow_system_path=True)
```
aaron-siegel, line 1184: Do we really want to drop support for system paths? It seems like a useful abstraction independent of data sharing.

aaron-siegel (approved):

pixeltable/catalog/catalog.py
mkornacker: As discussed offline: we can add it back later if there's a concrete use case for it.

PR #1419, opened 2026-06-16
Refactor of Table and Catalog

mkornacker:
New classes:
- Table: now an abstract base class declaring the full public table API
- LocalTable: the concrete local-execution table (former Table implementation); InsertableTable and View now subclass it.
- TableProxy / InsertableTableProxy: handles for hosted tables; all operations currently raise NotImplementedError.
- CatalogBase: abstract catalog API; Catalog now subclasses it.
- CatalogProxy: RPC-delegated catalog; stub for now.

Refactored:
- catalog.Path now carries the catalog location (org/db) parsed from pxt://<org>[:<db>]/... URIs, with .catalog_uri and .is_local; subsumes and replaces the old PxtUri.
- Runtime: single _catalog replaced by _catalogs: dict[str, CatalogBase] keyed by catalog URI, plus get_catalog(path) that routes to the local Catalog or a CatalogProxy.
- pxt.* globals now route every catalog operation through get_runtime().get_catalog(path); they no longer touch table internals


pixeltable/catalog/local_table.py
```diff
@@ -0,0 +1,1274 @@
...
+                n=n, n_per_stratum=n_per_stratum, fraction=fraction, seed=seed, stratify_by=stratify_by
+            )
+
+    def collect(self) -> 'pxt._query.ResultSet':
+        return self.select().collect()
+
+    def cursor(self) -> 'pxt._query.ResultCursor':
+        return self.select().cursor()
+
+    def show(self, *args: Any, **kwargs: Any) -> 'pxt._query.ResultSet':
```
aaron-siegel: I know this isn't a new issue but it'd be great if these had proper signatures. Maybe this refactor would be a good opportunity to sneak that in?

pixeltable/catalog/path.py
```diff
@@ -1,54 +1,134 @@
...
+
+    Identifies the catalog it lives in (org/db) plus the path within that catalog (components) and an
+    optional version. Org/db is None for local paths.
+
+    Construct via parse() or from_components(), which apply context-specific rules.
+    """
+
+    org: str | None = None  # None => in-process catalog (catalog_uri ''); a slug => remote/proxied catalog
+    db: str | None = None  # database within the org; always None when org is None, optional otherwise
+    components: tuple[str, ...] = ('',)  # ('',) denotes the catalog root
```
aaron-siegel: Note: it'd be more logical (and probably admit some code simplification) to use the empty tuple `()` for the root

pixeltable/catalog/path.py
```diff
@@ -1,54 +1,134 @@
...
+        if self.org is not None and not is_valid_identifier(self.org, allow_hyphens=True):
+            raise excs.RequestError(excs.ErrorCode.INVALID_PATH, f'Invalid organization name: {self.org!r}')
+        if self.db is not None and not is_valid_identifier(self.db, allow_hyphens=True):
+            raise excs.RequestError(excs.ErrorCode.INVALID_PATH, f'Invalid database name: {self.db!r}')
+        if self.version is not None and self.version < 0:
+            raise excs.RequestError(excs.ErrorCode.INVALID_PATH, f'Version must be non-negative: {self.version}')
+        if len(self.components) == 0:
+            raise excs.RequestError(excs.ErrorCode.INVALID_PATH, 'Path must have at least one component')
+        # ('',) is the root sentinel; any other component must be a valid identifier
+        if self.components != ('',) and not all(is_valid_identifier(c, allow_hyphens=True) for c in self.components):
```
aaron-siegel: (cf. preceding comment - the special case could be omitted here)

pixeltable/catalog/table_proxy.py
```diff
@@ -0,0 +1,257 @@
...
+class TableProxy(Table):
+    """Abstract base for handles to hosted tables, reached over RPC (delegated execution)."""
+
+    _path_obj: CatalogPath
+
+    def __init__(self, path: CatalogPath):
+        super().__init__(uuid4())  # placeholder id; the real id is assigned by the dispatch server
+        self._path_obj = path
+
+    def _name(self) -> str:
```
aaron-siegel, line 39: Why not simply declare this an `@abstractclass` too and omit the method signatures that are already defined in the parent? It feels like pure boilerplate to have to redeclare them.

Is the issue that the methods need a concrete implementation somewhere that's visible to the type checker, because their "true" implementations are auto-generated proxies? If it's a type checker issue, I might argue it's better to just selectively disable the type checker, than to redeclare all the method signatures, which duplicates code and is going to be a pain to maintain.

Another option is to auto-generate `.pyi` stubs that the type checker can ingest (I've set up the infra to do this in the class-based schema PR; it's a convenient way to "copy" APIs to unrelated structures in a statically checkable way, without having to separately maintain them).

aaron-siegel (approved):

pixeltable/catalog/path.py
mkornacker: Done.

pixeltable/catalog/local_table.py
mkornacker: Done.

pixeltable/catalog/table_proxy.py
mkornacker: I'll leave this in place for now. The current implementation doesn't use auto-generated stubs. Once that work is done, we can see what the scope for simplification is.

PR #1421, opened 2026-06-16
[PXT-1169] various improvements to pxt serve output

sergey-mkhitaryan:
1. Wait until the server is actually up before printing that we are listening.
2. Bug fix: with `--port 0`, print the actual port that Uvicorn found rather than `:0`
3. Skip the traceback on `KeyboardInterrupt`
4. Suppress the uvicorn logging from stdout/stderr

Implementation details:
* To print the correct status at the right time, we subclass `uvicorn.Server` and override its `startup()`. This is the most elegant way to do this correctly that I found.
* Before this change, `pxt serve` would print the successful status before Uvicorn runs. That behavior was wrong, but easy to test because it didn't need Uvicorn to actually bind in a test. Once it was fixed, testing the stdout of `pxt serve` end-to-end became very difficult. For that reason I narrowed down the affected tests to only validate the function that produces the output string, and do nothing else.

Sample `uv run pxt serve pipeline --port 0` output:
```
...
Created table 'documents'.
Created table 'images'.
Starting Pixeltable service pipeline...
Pixeltable is running on http://localhost:53511
  Routes: 7
  API docs at http://localhost:53511/docs
```

Sample `uv run pxt serve pipeline --port 0 --json` output:
```
...
Created table 'documents'.
Created table 'images'.
{"status": "started", "host": "0.0.0.0", "port": 53560, "url": "http://localhost:53560", "docs_url": "http://localhost:53560/docs", "routes": 7}
```


pierrebrunelle:
It's way better! Thanks.

@sergey-mkhitaryan I wonder if the ideal output should be solely:

```
Starting Pixeltable service pipeline...
Pixeltable is running on http://localhost:63076
  Routes: 7
  API docs at http://localhost:63076/docs
 ```
 
 And could show uvicorn traceback only only on failure/requesting verbose/logs etc?
 
 The fact that we use uvicorn internally for our serve, not really meaningful to users. It's not like they have built their own FastAPI app here using Uvicorn as well. They just want to serve that endpoint using our toolkit.

**Comment by pierrebrunelle** [comment]
✅

File: pixeltable/env.py [hunk 1]
```diff
@@ -356,6 +356,14 @@ def _set_up(self, echo: bool = False, reinit_db: bool = False) -> None:
         self._managed_logging_handlers.append((_http_server_logger, http_fh))
         _http_server_logger.propagate = False
 
+        # Route uvicorn's loggers to the main pxt log file
```
aaron-siegel, line 359: Are we sure this is what we want? Note that pxt already has a couple 'auxiliary' logs (although maybe it shouldn't; logging architecture is something we've never really given much thought to)

pixeltable_cli/client/commands/serve.py
```diff
@@ -325,42 +344,37 @@ def _run(cfg: config.ServiceConfig, app: Any, json_output: bool = False) -> None
...
+    if not json_output:
+        print(f'Starting Pixeltable service {cfg.name}...')
+    # log_config=None keeps uvicorn from installing its own stderr handlers, we don't want its logging in the console.
+    # Env routes uvicorn loggers to a file.
+    server = PxtServer(uvicorn.Config(app, host=cfg.host, port=cfg.port, log_config=None))
     try:
-        uvicorn.run(app, host=host, port=port)
+        server.run()
+    except KeyboardInterrupt:
+        print('KeyboardInterrupt', file=sys.stderr)
```
aaron-siegel: Maybe print a more user-friendly message here? "Shutting down Pixeltable server." or some such? (Jupyter, uvicorn, etc. could be good models for the UX here)

pixeltable_cli/client/commands/serve.py
```diff
@@ -316,6 +317,24 @@ def _create_route_from_args(args: argparse.Namespace) -> config.RouteConfig:
     raise AssertionError(f'unknown serve mode: {args.mode}')
 
 
+def _started_status(host: str | None, port: int, is_ssl: bool, n_routes: int, json_output: bool) -> str:
+    if host is None:
```
aaron-siegel: Is this necessary? Might be cleaner to just have `host: str` as the parameter. It's coming from `uvicorn.Config` anyway which has type `str`.

One might argue the default should be '127.0.0.1', not '0.0.0.0' (which I believe means something different in uvicorn) - 127.0.0.1 is the default in uvicorn.Config

aaron-siegel (approved):

pixeltable/env.py
sergey-mkhitaryan: Yeah, I don't know, I think unless uvicorn turns out way too chatty, it's fine to be in the main log file.

pixeltable_cli/client/commands/serve.py
sergey-mkhitaryan: Updated the error message. Jupyter notebook asks to confirm shutdown on ctrl+C. I don't know if we need that, but that can be added later anyway.

pixeltable_cli/client/commands/serve.py
sergey-mkhitaryan: you're right, host cannot be None here

pixeltable/env.py
sergey-mkhitaryan: Regardless, this is very easy to change in the future, let's keep it simple for now and see how it goes.

PR #1425, opened 2026-06-20
Delegated execution

mkornacker:
This is a functionally complete implementation of delegated execution, without any performance optimization or atomicity guarantees (all communication is synchronous and doesn't store request state/makes request state recoverable).

## New classes

- **`ViewProxy`** (`catalog/view_proxy.py`) — client handle for a hosted view/snapshot.
- **`ProxyClient`** (`service/proxy_client.py`) — abstract transport for proxy RPCs.
- **`InProcessProxyClient`** (`service/proxy_client.py`) — in-process transport (no HTTP).
- **`ProxyHttpClient`** (`service/proxy_client.py`) — HTTP transport to a proxy daemon.
- **`ProxyRequest`** (`service/proxy_protocol.py`) — serialized proxy request.
- **`ProxyResponse`** (`service/proxy_protocol.py`) — serialized proxy response.
- **`TablePathKey`** (`catalog/table_path.py`) — version identity of a table path, used for CAS-gated mutations.
- **`ProxyResultCursor`** (`_query.py`) — cursor over a materialized proxy result set.

New function-based modules:
- **`service/proxy_daemon.py`** — daemon entrypoint and lifecycle/discovery API.
- **`service/proxy_dispatch.py`** — server-side request dispatch.

## Significantly changed classes

- **`TableProxy`** (`catalog/table_proxy.py`) — full hosted-table API with client-side validation.
- **`InsertableTableProxy`** (`catalog/insertable_table_proxy.py`) — hosted insert from rows/Query/Table sources.
- **`CatalogProxy`** (`catalog/catalog_proxy.py`) — delegated catalog create/get/move/drop.
- **`TableMdPath`** (`catalog/table_path.py`) — metadata-backed path supports queries, indexes, and snapshots without a live `TableVersion`.
- **`Table`** (`catalog/table.py`) — shared client-side argument/mutability validation.
- **`Query` / `ResultCursor` / `ResultSet`** (`_query.py`) — proxy execution and a public result schema.
- **`ColumnRef`** / **`SimilarityExpr`** (`exprs/`) — proxy-aware embedding/index resolution.
- **`Error`** (`exceptions.py`) — error codes that round-trip across the wire.
- **`Runtime`** (`runtime.py`) — per-catalog routing for local vs proxy.

## Remaining functional gaps

- doesn't deal with local media correctly (tests only work because the test proxy daemon is running on the same machine)
- **Snapshots**: metadata reports a version-qualified path/name (`test_snap:0`) instead of the bare path; snapshot-mutation errors carry the daemon's bare name, not the full `pxt://` path; `rename_column` on a snapshot raises an internal error instead of a clean one; pinned-version / snapshot-only column resolution differs.
- **Dropped-table detection**: query builders (`select`/`group_by`/`where`) and reused handles don't detect a dropped table (the check only runs daemon-side at dispatch); args are serialized before that check, so a bad arg can mask the error.
- **Dropped-index detection**: a serialized query referencing a since-dropped index doesn't error on reload.
- **Base-table column refs**: a base-table column referenced through a view/snapshot query can fail to resolve over proxy.

## Test conversion (`tests/test_*.py`)

The dual-mode harness is new (master has 0 proxy tests). **253 of 472** `test_` methods in tests/test_*.py. now also run against the proxy.


pixeltable/catalog/catalog_proxy.py
```diff
@@ -61,29 +93,70 @@ def create_view(
...
         if_exists: IfExistsParam,
     ) -> Table:
-        self._unsupported()
+        args = {
+            'path': path,
+            'base': base,
+            'select_list': select_list,
+            'where': where,
+            'sample_clause': sample_clause,
+            'additional_columns': additional_columns if additional_columns is not None else {},
```
aaron-siegel: I find `additional_columns or {}` easier to follow here

pixeltable/catalog/catalog_proxy.py
```diff
@@ -61,29 +93,70 @@ def create_view(
...
         self, dir_path: Path, recursive: bool = False, with_error_counts: bool = False
     ) -> dict[str, DirEntry]:
-        self._unsupported()
+        return self._client.send_request(
+            'CatalogBase',
+            'get_dir_contents',
+            {'dir_path': dir_path, 'recursive': recursive, 'with_error_counts': with_error_counts},
+        )
 
     def create_dir(self, path: Path, if_exists: IfExistsParam, parents: bool) -> Dir:
```
aaron-siegel, line 154: Lots of boilerplate here. Can't these all be autogened? Explicit boilerplate like this increases the maintenance burden as we end up having to keep method signatures in sync in N different places. And they all follow the identical pattern `send_request('CatalogBase', method_name, kwargs)`.

pixeltable/catalog/globals.py
```diff
@@ -217,3 +224,29 @@ def is_system_column_name(name: str) -> bool:
...
+    from .column import Column
+
+    result: dict[str, ColumnSpec] = {}
+    for name, spec in schema.items():
+        if isinstance(spec, exprs.Expr):
+            result[name] = {'value': spec}
+            continue
+        if isinstance(spec, dict):
+            col_spec: dict[str, Any] = dict(spec)
+            Column._validate_column_spec(name, cast('ColumnSpec', col_spec))
```
aaron-siegel: Why is `ColumnSpec` quoted? I'm not sure what it means to `cast` to a `TypedDict` instance (note that `TypedDict` doesn't exist at runtime)

pixeltable/catalog/globals.py
```diff
@@ -217,3 +224,29 @@ def is_system_column_name(name: str) -> bool:
...
+
+    from .column import Column
+
+    result: dict[str, ColumnSpec] = {}
+    for name, spec in schema.items():
+        if isinstance(spec, exprs.Expr):
+            result[name] = {'value': spec}
+            continue
+        if isinstance(spec, dict):
+            col_spec: dict[str, Any] = dict(spec)
```
aaron-siegel, line 241: Why not just `col_spec = ColumnSpec(spec)`? Does that not work?

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -1,10 +1,165 @@
...
+            RowDataTableDataConduit,
+            TableDataConduit,
+        )
+
+        self._validate_insert_source(source)
+        if source is None:
+            # the kwargs form (t.insert(col=val, ...)) is a single row
+            source = [kwargs]
+            kwargs = None
+        if isinstance(source, collections.abc.Iterator):
```
aaron-siegel: typing.Iterator

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -1,10 +1,165 @@
...
+
+    def compute(
+        self,
+        source: collections.abc.Sequence[dict[str, Any]] | collections.abc.Sequence[pydantic.BaseModel],
+        /,
+        *,
+        on_error: Literal['abort', 'ignore'] = 'abort',
+    ) -> list[dict[str, Any]]:
+        # str/bytes are technically Sequences; reject them explicitly rather than treating them as paths/URLs.
+        if isinstance(source, (str, bytes)) or not isinstance(source, collections.abc.Sequence) or len(source) == 0:
```
aaron-siegel: typing.Sequence (etc - and all over - it should rarely be necessary to refer directly to collections.abc)

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -1,10 +1,165 @@
...
+                excs.ErrorCode.UNSUPPORTED_OPERATION,
+                f'Hosted insert does not support a {type(source).__name__} source yet.',
+            )
+        return self._dispatch(
+            'insert', {'rows': rows, 'on_error': on_error, 'print_stats': print_stats, 'return_rows': return_rows}
+        )
+
+    def compute(
+        self,
+        source: collections.abc.Sequence[dict[str, Any]] | collections.abc.Sequence[pydantic.BaseModel],
```
aaron-siegel: typing.Sequence

pixeltable/catalog/insertable_table_proxy.py
aaron-siegel: Wouldn't the special case of `source` being an instance of `(str, bytes)` also be rejected by the subsequent check?

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -1,10 +1,165 @@
...
+        # str/bytes are technically Sequences; reject them explicitly rather than treating them as paths/URLs.
+        if isinstance(source, (str, bytes)) or not isinstance(source, collections.abc.Sequence) or len(source) == 0:
+            raise excs.RequestError(
+                excs.ErrorCode.UNSUPPORTED_OPERATION,
+                f'compute() requires a non-empty sequence of dicts or pydantic models; got {type(source).__name__}',
+            )
+        if not all(isinstance(row, (dict, pydantic.BaseModel)) for row in source):
+            raise excs.RequestError(
+                excs.ErrorCode.UNSUPPORTED_OPERATION,
+                f'compute() requires a sequence of dicts or pydantic models; got {type(source).__name__}',
```
aaron-siegel: The `got` here won't tell the user much in the case where this is a list containing something other than a dict/model

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -1,10 +1,165 @@
...
+            schema[col_md.name] = col_md.col_type
+            if col_md.is_computed:
+                converter.computed_col_names.add(col_md.name)
+            elif not col_md.col_type.nullable:
+                converter.reqd_col_names.add(col_md.name)
+        converter.pxt_schema = schema
+        converter.prepare_for_insert_into_table()
+        return converter.pxt_rows
+
+    def delete(self, where: 'exprs.Expr' | None = None) -> UpdateStatus:
```
aaron-siegel, line 161: What about update() / batch_update() ?

pixeltable/catalog/table.py
```diff
@@ -815,6 +825,69 @@ def compute(
             ... # contains {'errortype': ..., 'errormsg': ...}.
         """
 
+    def _validate_update_value_spec(self, value_spec: dict[str, Any]) -> None:
+        from pixeltable import exprs
+
+        for col_name, val in value_spec.items():
+            if not isinstance(col_name, str):
```
aaron-siegel, line 842: Also check is_valid_identifier()?

pixeltable/catalog/table_path.py
```diff
@@ -62,16 +86,51 @@ def is_snapshot(self) -> bool: ...
...
     def is_mutable(self) -> bool: ...
 
     @abc.abstractmethod
     def is_versioned(self) -> bool: ...
 
+    @property
+    @abc.abstractmethod
+    def catalog_uri(self) -> Path:
+        """The catalog this table belongs to."""
+        ...
```
aaron-siegel: spurious ...

pixeltable/catalog/table_path.py
```diff
@@ -502,32 +621,45 @@ def get_column_md(self, qcolid: QColumnId) -> ColumnVersionMd:
...
+                        f'Column {col_name!r} has multiple {idx_class.display_name()} indices; '
+                        'specify `idx_name` instead',
+                    )
+                if name is not None:
+                    named = [idx_md for idx_md in candidates if idx_md.name == name]
+                    if len(named) == 0:
+                        raise excs.NotFoundError(
+                            excs.ErrorCode.INDEX_NOT_FOUND, f'Index {name!r} not found for column {col_name!r}'
+                        )
+                    return named[0]
```
aaron-siegel, line 660: assert len(named) == 1 for good measure?

pixeltable/catalog/table_proxy.py
```diff
@@ -190,7 +276,8 @@ def drop_index(
         idx_name: str | None = None,
         if_not_exists: Literal['error', 'ignore'] = 'error',
     ) -> None:
-        raise NotImplementedError
+        self._check_mutable('drop an index from')
+        self._dispatch('drop_index', {'column': column, 'idx_name': idx_name, 'if_not_exists': if_not_exists})
```
aaron-siegel: Again, lots of boilerplate here (cf. preceding comment; although in this case it's not quite as simple due to the varying validation logic).

At minimum, it would be great to avoid repeating all the kwargs in the call to `_dispatch()`, which is error prone (e.g., if we add a new kwarg and forget to also update the _dispatch() call). I think this should be achievable with something like

```python
def drop_index(
    self,
    *,
    column: str | ColumnRef | None = None,
    idx_name: str | None = None,
    if_not_exists: Literal['error', 'ignore'] = 'error',
) -> None:
    bound_args = {k: v for k, v in locals().items() if k != 'self'}
    ...  # validation logic and whatnot
    self._dispatch('drop_index', bound_args)
```
etc.

pixeltable/catalog/view_proxy.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def __init__(self, id: UUID, is_anon_snapshot: bool, tbl_md_path: TableMdPath, client: 'ProxyClient'):
+        effective_version = tbl_md_path.effective_version() if is_anon_snapshot else None
+        super().__init__(id, effective_version, tbl_md_path, client)
+
+    def _display_name(self) -> str:
+        return 'snapshot' if self._tbl_md_path.is_snapshot() else 'view'
+
+    def _get_base_table(self) -> Table | None:
+        return None
```
aaron-siegel, line 33: ? Is this intentional here? Needs a code comment explaining what's going on

pixeltable/catalog/view_proxy.py
```diff
@@ -0,0 +1,56 @@
...
+
+if TYPE_CHECKING:
+    from pixeltable import exprs, type_system as ts
+    from pixeltable.service.proxy_client import ProxyClient
+
+    from ..globals import TableDataSource
+    from .update_status import UpdateStatus
+
+
+class ViewProxy(TableProxy):
```
aaron-siegel, line 20: I can see value in extending the hierarchy with a `LocalView` and `LocalInsertableTable`, so that `View` and `InsertableTable` become abstracts with shared functionality. Then
```
class ViewProxy(TableProxy, View):
```
and so on.

pixeltable/func/callable_function.py
```diff
@@ -183,17 +184,27 @@ def overload(self, fn: Callable) -> CallableFunction:
...
         return super()._as_dict()
 
     @classmethod
     def _from_dict(cls, d: dict) -> Function:
+        if 'binary' in d:
+            try:
+                return cls.from_store(d['name'], d['store_md'], base64.b64decode(d['binary']))
+            except Exception:
+                error_msg = (
+                    f'the locally defined UDF {d["name"]!r} could not be deserialized. This probably means the UDF\n'
```
aaron-siegel, line 201: We need to change to storing notebook UDFs as global references rather than pickled code (pickled code basically never belongs in the DB).

pixeltable/func/expr_template_function.py
```diff
@@ -153,12 +153,23 @@ def __str__(self) -> str:
     def _as_dict(self) -> dict:
         if self.self_path is not None:
             return super()._as_dict()
-        assert not self.is_polymorphic
-        assert len(self.templates) == 1
-        return {'expr': self.template.expr.as_dict(), 'signature': self.signature.as_dict(), 'name': self.name}
+        if not self.is_polymorphic:
+            return {'expr': self.template.expr.as_dict(), 'signature': self.signature.as_dict(), 'name': self.name}
```
aaron-siegel: Why two different dict signatures here? Why not simply normalize everything to the polymorphic-friendly structure?

pixeltable/func/function.py
```diff
@@ -513,14 +513,14 @@ def as_dict(self) -> dict[str, Any]:
...
-        assert not self.is_polymorphic
         classpath = f'{self.__class__.__module__}.{self.__class__.__qualname__}'
         return {'_classpath': classpath, **self._as_dict()}
 
     def _as_dict(self) -> dict:
-        """Default serialization: store the path to self (which includes the module path) and signature."""
+        """Default serialization: store the path to self (which includes the module path) and signature(s)."""
         assert self.self_path is not None
+        if self.is_polymorphic:
+            return {'path': self.self_path, 'signatures': [sig.as_dict() for sig in self.signatures]}
```
aaron-siegel: Same question as above - why the branching logic?

pixeltable/metadata/schema.py
```diff
@@ -433,6 +430,15 @@ class Function(Base):
     against version mismatches.
     """
 
+    # TODO: deprecate this table. Anonymous CallableFunctions now serialize their pickled body by value, inline
```
aaron-siegel, line 433: They shouldn't be serializing pickled code at all (see above comment) - that's the real TODO, imo.

pixeltable/service/proxy_daemon.py
```diff
@@ -0,0 +1,378 @@
+"""Proxy daemon: a Pixeltable process that serves the proxy protocol over HTTP, running catalog
+operations against its own database in direct mode.
+
+This module is both the daemon entrypoint (``python -m pixeltable.service.proxy_daemon``) and the
```
aaron-siegel: double backticks is different style from docstring style everywhere else in the codebase

pixeltable/_query.py
```diff
@@ -1296,13 +1350,13 @@ def join(
...
             group_by_clause=self.group_by_clause,
-            grouping_tbl=self.grouping_tbl,
+            grouping_tbl_key=self.grouping_tbl_key,
             order_by_clause=self.order_by_clause,
             limit=self.limit_val,
             offset=self.offset_val,
         )
 
-    def group_by(self, *grouping_items: Any) -> Query:
+    def group_by(self, *grouping_items: exprs.Expr | catalog.Table) -> Query:
```
aaron-siegel, line 1368: As long as we're polishing the API here it'd be even better would be to use overloads since there are really two signatures:

```python
def group_by(self, *grouping_items: exprs.Expr) -> Query:
```

```python
def group_by(self, grouping_tbl: catalog.Table) -> Query:
```

and the existing signature is a bit of a lie

tests/conftest.py
```diff
@@ -217,6 +218,94 @@ def uses_db(init_env: None, request: pytest.FixtureRequest) -> Iterator[None]:
...
+    clean_db()
+    Config.init({}, reinit=True)
+    Env.get().default_time_zone = None
+    Env.get().user = None
+    reload_catalog()
+    FileCache.get().validate()
+    FileCache.get().set_capacity(10 << 30)  # 10 GiB
+
+    if catalog_mode == 'proxy':
+        from pixeltable.service import proxy_daemon
```
aaron-siegel, line 302: move to top of file

tests/conftest.py
```diff
@@ -217,6 +218,94 @@ def uses_db(init_env: None, request: pytest.FixtureRequest) -> Iterator[None]:
...
+    yield p
+
+    Env.get().user = None
+    get_runtime().catalog.validate_store()
+
+
+@pytest.fixture(scope='function')
+def make_local_path(init_env: None) -> Iterator[Callable[[str], str]]:
+    """Stand-in for make_catalog_path() for tests that fail in proxy mode."""
+    clean_db()
```
aaron-siegel: This code is repeated numerous times now; pls factor it out (I'll need to modify it for class-based schemas, for example)

tests/conftest.py
```diff
@@ -395,6 +499,11 @@ def small_img_tbl(uses_db: None) -> pxt.Table:
     return create_img_tbl('small_img_tbl', num_rows=40)
 
 
+@pytest.fixture(scope='function')
+def small_img_tbl_dual(make_catalog_path: Callable[[str], str]) -> pxt.Table:
```
aaron-siegel: I feel like we are unnecessarily duplicating many abstractions here.

Rather than bloat conftest with `*_dual()` methods for everything, I'd rather just upgrade all the existing ones with a `make_catalog_path` (or even better imo, `catalog_root`; the `Callable` seems heavyweight and fuss for this purpose, but I don't feel as strongly about that). It would default to a single-resolution local catalog but be selectively upgradeable to a parameterized local/proxy resolution.

tests/test_array_type.py
```diff
@@ -35,20 +38,24 @@ def test_array_dtypes(self, do_reload_catalog: bool, init_env: None, uses_db: No
...
             except Exception as e:
                 raise type(e)(f'Failed for col_dtype={col_dtype}') from e
 
     def _test_array_dtype(
-        self, col_dtype: type[np.generic] | ColumnType, acceptable_dtypes: list, do_reload_catalog: bool
+        self,
+        col_dtype: type[np.generic] | ColumnType,
+        acceptable_dtypes: list,
+        do_reload_catalog: bool,
+        p: Callable[[str], str],
```
aaron-siegel: `make_catalog_path`, please.

tests/test_array_type.py
```diff
@@ -101,8 +108,11 @@ def _validate_dtype(self, val: np.ndarray, literal_dtype: type[np.generic] | Col
             case _:
                 raise ValueError(f'Unsupported dtype: {literal_dtype}')
 
-    def test_non_parameterized_array_accepts_all_dtypes(self, init_env: None, uses_db: None) -> None:
-        t = pxt.create_table('test_numpy_dtypes', {'array': pxt.Array})
+    def test_non_parameterized_array_accepts_all_dtypes(
+        self, init_env: None, make_catalog_path: Callable[[str], str]
+    ) -> None:
+        p = make_catalog_path
```
aaron-siegel, line 114: What's the point, if it only gets called once?

aaron-siegel (changes requested):

pixeltable/catalog/catalog_proxy.py
mkornacker: Done.

pixeltable/catalog/catalog_proxy.py
mkornacker: There are only 5 methods occupying 25 lines of code that are boilerplate. Creating a new abstraction for that feels like overkill at this point, but happy to revisit this setup at a future point in time.

re N different places: I don't foresee that going beyond 2.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Removed.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: That's in TableProxy/LocalTable.

pixeltable/catalog/table_path.py
mkornacker: Done.

pixeltable/catalog/table_path.py
mkornacker: Done.

pixeltable/catalog/view_proxy.py
mkornacker: Done.

pixeltable/metadata/schema.py
mkornacker: Agreed.

pixeltable/service/proxy_daemon.py
mkornacker: Done.

tests/test_array_type.py
mkornacker: Uniformity. It doesn't hurt anything.

pixeltable/catalog/globals.py
mkornacker: Done.

pixeltable/catalog/globals.py

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Done.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Done.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Done.

mkornacker: Done.

tests/conftest.py
mkornacker: Done.

tests/conftest.py
mkornacker: Done.

pixeltable/_query.py
mkornacker: Done.

pixeltable/func/expr_template_function.py
mkornacker: Done.

pixeltable/func/function.py
mkornacker: Done.

pixeltable/catalog/view_proxy.py
mkornacker: I'm not sure what that would look like. Can we defer that?

pixeltable/catalog/table.py
mkornacker: Done.

pixeltable/catalog/table_proxy.py
mkornacker: Done.

tests/test_array_type.py
mkornacker: Done.

pixeltable/func/callable_function.py
mkornacker: Agreed. Follow-on?

tests/conftest.py
mkornacker: Done.

aaron-siegel (approved):

PR #1427, opened 2026-06-22
[PXT-923] OpenTelemetry instrumentation for Pixeltable (part 1)

christopherpestano:
Adds opt-in OpenTelemetry support for Pixeltable. Core Pixeltable carries no OpenTelemetry dependency and never activates telemetry on its own; instrumentation lives in a separate distribution, opentelemetry-instrumentation-pixeltable, installable on its own or via the pixeltable[otel] extra.

Pixeltable core exposes a small, dependency-free hooks layer that reports spans and events. When no subscriber is registered every call is a near-free no-op, so there is no overhead for users who do not opt in. The instrumentation package subscribes to those hooks and translates them into OpenTelemetry spans, metrics, and logs.

Applications that already own an OpenTelemetry SDK can attach Pixeltable to their global providers with PixeltableInstrumentor().instrument(), and the standard opentelemetry-instrument CLI discovers it automatically via its entry point. Applications that want Pixeltable to build the providers and OTLP exporter can call init() once at startup. Configuration is read from the [otel] section of the Pixeltable config and overridden by the standard OTEL_EXPORTER_OTLP_* environment variables. The pxt serve command gains an --otel flag to enable instrumentation for a served app.

This is part 1 of two. It establishes the hooks layer, the instrumentation bridge, and the initial span and metric call sites for the insert path (operation, planning, media, store, row, and UDF spans). Part 2 will extend coverage to the remaining operations (update, delete, queries) across the repo.


pixeltable/catalog/catalog.py
```diff
@@ -448,6 +453,10 @@ def begin_xact(
                 raise
 
             finally:
+                # still non-None here means this attempt failed; sys.exc_info() carries the escaping exception
+                # (None on retry `continue`s, which consumed it); no-op after the yield
+                hooks.span_end(xact_span, exc=sys.exc_info()[1])
+                xact_span = None
```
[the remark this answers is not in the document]
christopherpestano: Already fixed

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -193,6 +197,13 @@ def _dispatch_input_rows(self) -> None:
 
         self.eval_ctx.init_rows(rows)
         self.set_var_slots(rows)
+        if hooks.active():
+            # per-row spans nest their UDF cell spans; suppressed (None) unless DEBUG is enabled under an
+            # operation span, so a bare query with no operation span stays dark
+            for row in rows:
+                if row.parent_row is None and self._row_spans_opened < self.MAX_ROW_SPANS:
+                    row.span = hooks.span_start('pixeltable.row', level=hooks.DEBUG, parent=hooks.current_span())
+                    self._row_spans_opened += 1
```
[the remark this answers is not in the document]
christopherpestano: Fixed. Rows are now appended to a list only when span_start actually emitted a span, and the cap checks the length of that list, so it counts emitted spans rather than attempts.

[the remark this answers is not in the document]
christopherpestano: Fixed. ExprEvalNode now tracks rows with open spans and ends any span still open, with the operation error attached, in the main loop's finally block. This covers exception aborts, cancellation, and abandoned cursors. Added test_failed_insert_ends_row_spans to cover this.

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -193,6 +197,13 @@ def _dispatch_input_rows(self) -> None:
 
         self.eval_ctx.init_rows(rows)
         self.set_var_slots(rows)
+        if hooks.active():
+            # per-row spans nest their UDF cell spans; suppressed (None) unless DEBUG is enabled under an
+            # operation span, so a bare query with no operation span stays dark
+            for row in rows:
+                if row.parent_row is None and self._row_spans_opened < self.MAX_ROW_SPANS:
+                    row.span = hooks.span_start('pixeltable.row', level=hooks.DEBUG, parent=hooks.current_span())
```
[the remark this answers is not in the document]
christopherpestano: Fixed. ExprEvalNode now tracks rows with open spans and ends any span still open, with the operation error attached, in the main loop's finally block. This covers exception aborts, cancellation, and abandoned cursors. Added test_failed_insert_ends_row_spans to cover this.

docs/release/platform/observability.ipynb
```diff
@@ -0,0 +1,269 @@
...
+      "\u001b[0;31mImportError\u001b[0m: cannot import name 'hooks' from 'pixeltable' (/opt/homebrew/Caskroom/miniforge/base/envs/pxt/lib/python3.10/site-packages/pixeltable/__init__.py)"
+     ]
+    }
+   ],
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The notebook was re-executed in a clean environment and committed with real outputs; the error traceback is gone.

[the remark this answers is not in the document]
christopherpestano: Fixed. The notebook was re-executed in a clean environment and committed with real outputs; the error traceback is gone.

docs/release/platform/observability.ipynb
```diff
@@ -0,0 +1,269 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "obs-init",
+   "metadata": {},
+   "outputs": [
+    {
+     "ename": "ImportError",
+     "evalue": "cannot import name 'hooks' from 'pixeltable' (/opt/homebrew/Caskroom/miniforge/base/envs/pxt/lib/python3.10/site-packages/pixeltable/__init__.py)",
+     "output_type": "error",
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The notebook was re-executed in a clean environment and committed with real outputs; the error traceback is gone.

docs/release/platform/observability.ipynb
```diff
@@ -0,0 +1,269 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "obs-intro",
+   "metadata": {},
+   "source": [
+    "# Observability with OpenTelemetry\n",
```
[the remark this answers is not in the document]
christopherpestano: Keeping as is for consistency: every existing notebook in docs/release/platform (embedding-indexes, udfs-in-pixeltable, external-files) starts with an H1 markdown cell and has no YAML frontmatter raw cell. The guidelines document describes the older docs/notebooks layout.

pixeltable/hooks.py
```diff
@@ -0,0 +1,305 @@
...
+def current_span() -> SpanHandle | None:
+    return _current_span.get()
+
+
+def _log_subscriber_error(subscriber: Subscriber, method: str, exc: Exception) -> None:
+    # warn once per (subscriber, method), then drop to debug to avoid log storms from per-row failures
+    key = (id(subscriber), method)
+    level = logging.DEBUG if key in _logged_error_keys else logging.WARNING
+    _logged_error_keys.add(key)
+    _logger.log(level, f'instrumentation subscriber {type(subscriber).__name__}.{method}() failed: {exc!r}')
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The check and add of _logged_error_keys now run under the registry lock.

tests/test_tracing_insert.py
```diff
@@ -0,0 +1,122 @@
...
+            hooks.set_span_level(hooks.INFO)
+
+    def test_failed_insert_records_exc(self, uses_db: None) -> None:
+        t = pxt.create_table('tracing_test', {'c': pxt.Int}, if_exists='replace')
+        t.add_computed_column(out=fail_on_three(t.c))
+        sub = RecordingSubscriber()
+        hooks.subscribe(sub)
+        try:
+            with pytest.raises(excs.ExprEvalError):
+                t.insert([{'c': i} for i in range(10)])
```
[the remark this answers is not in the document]
christopherpestano: Fixed, the assertion now uses match='ValueError'. Note pxt_raises is not applicable here since ExprEvalError is not a pxt.Error and carries no error code.

pixeltable/exec/expr_eval/evaluators.py
```diff
@@ -99,6 +100,13 @@ def reset(self) -> None:
         if self.call_args_queue is not None:
             self.call_args_queue = asyncio.Queue[FnCallArgs]()
 
+    def _cell_span(self, row: exprs.DataRow) -> Any:
+        """Span for one row's UDF call, nested under the row's span; no-op when the row has no span."""
+        if row.span is None:
+            return contextlib.nullcontext()
+        # DEBUG so cell spans emit/suppress in lockstep with the row span they nest under
+        return hooks.span(f'udf.{self.fn.display_name}', level=hooks.DEBUG, parent=row.span)
```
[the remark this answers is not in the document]
christopherpestano: Fixed. The per-row UDF cell span now passes set_current=True, so provider instrumentors parent their spans to the UDF span. This is safe under concurrency: each asyncio task copies the context at creation, so the ambient span is task-specific and overlapping UDF calls never see each other's span.

pixeltable/exec/expr_eval/evaluators.py
```diff
@@ -187,15 +195,18 @@ def _create_batch_call_args(self, call_args: list[FnCallArgs]) -> FnCallArgs:
...
-                )
-            else:
-                # check for cancellation before starting something potentially long-running
-                if asyncio.current_task().cancelled() or self.dispatcher.exc_event.is_set():
-                    return
-                result_batch = self.fn.exec_batch(batched_call_args.batch_args, batched_call_args.batch_kwargs)
+            # batched calls process many rows in one invocation, so they can't nest under a single row
+            # span; emit one span under the ambient operation span instead (see per-row span design)
+            with hooks.span(f'udf.{self.fn.display_name}', batch_size=len(batched_call_args.rows)):
+                if self.fn.is_async:
```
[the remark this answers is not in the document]
christopherpestano: Batched spans intentionally stay non-current: without set_current the span is already suppressed when there is no ambient operation span, so bare queries stay dark with no extra guard needed. Making batched and scheduler path calls current is deferred to part 2 together with the scheduler instrumentation.

pixeltable/catalog/insertable_table.py
```diff
@@ -158,18 +158,20 @@ def insert(
...
             kwargs = None
 
-        data_source = TableDataConduit.create(
-            source, source_format=source_format, src_schema_overrides=schema_overrides, extra_fields=kwargs
-        )
-        data_source.add_table_info(self)
-        data_source.prepare_for_insert_into_table()
+        with hooks.span('pixeltable.insert', set_current=True):
+            with hooks.span('pixeltable.data_source.prepare'):
+                data_source = TableDataConduit.create(
```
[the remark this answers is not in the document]
christopherpestano: The PR description has been updated: part 1 includes the hooks layer, the bridge, and the initial insert path span and metric call sites; part 2 extends coverage to the remaining operations.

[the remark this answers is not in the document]
christopherpestano: The PR description has been updated: part 1 includes the hooks layer, the bridge, and the initial insert path span and metric call sites; part 2 extends coverage to the remaining operations.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+    def _instrument(self, **kwargs: Any) -> None:
+        self._subscriber = _OtelSubscriber(kwargs.get('tracer_provider'), kwargs.get('meter_provider'))
+        hooks.subscribe(self._subscriber)
+        _install_record_factory()
+
+    def _uninstrument(self, **kwargs: Any) -> None:
+        if self._subscriber is not None:
+            hooks.unsubscribe(self._subscriber)
+            self._subscriber = None
+        _uninstall_record_factory()
```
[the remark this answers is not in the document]
christopherpestano: This works as intended because BaseInstrumentor makes each instrumentor class a singleton: its __new__ caches cls._instance, so every PixeltableInstrumentor() call returns the same instance and instrument and uninstrument always share state. A double record factory install is also prevented since BaseInstrumentor.instrument() is a no-op when already instrumented.

pixeltable/hooks.py
```diff
@@ -0,0 +1,306 @@
...
+    with _registry_lock:
+        if subscriber not in _SUBSCRIBERS:
+            _SUBSCRIBERS = (*_SUBSCRIBERS, subscriber)
+
+
+def unsubscribe(subscriber: Subscriber) -> None:
+    """Remove a previously registered subscriber; a no-op if it isn't registered."""
+    global _SUBSCRIBERS  # noqa: PLW0603
+    with _registry_lock:
+        _SUBSCRIBERS = tuple(s for s in _SUBSCRIBERS if s is not subscriber)
```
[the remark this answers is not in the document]
christopherpestano: Fixed. subscribe() now uses an identity check (all(s is not subscriber for s in _SUBSCRIBERS)) so both registration and removal use identity consistently.

pixeltable/telemetry.py
```diff
@@ -0,0 +1,306 @@
...
+    if level < _span_level or (parent is None and not set_current):
+        return None
+    attrs = _resolve_attrs(attrs)
+    tokens: list[Any] = []
+    for s in subs:
+        parent_token: Any = None
+        if parent is not None and s in parent.subs:
+            # a subscriber that registered after the parent started has no parent token (its span is a root)
+            parent_token = parent.tokens[parent.subs.index(s)]
+        try:
```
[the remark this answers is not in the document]
christopherpestano: Keeping as is. The subscriber count is 1 or 2 in practice, so the two scans over a tiny tuple amount to a handful of pointer comparisons. Building or caching a dict per span_start() would cost more than it saves at this scale and adds state to SpanHandle for no measurable benefit.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+        return record
+
+    logging.setLogRecordFactory(factory)
+
+
+def _uninstall_record_factory() -> None:
+    global _prev_record_factory  # noqa: PLW0603
+    if _prev_record_factory is not None:
+        logging.setLogRecordFactory(_prev_record_factory)
+        _prev_record_factory = None
```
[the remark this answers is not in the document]
christopherpestano: Keeping as is. _install_record_factory() has a single caller, _instrument(), and BaseInstrumentor is a singleton that refuses double instrumentation (it warns and returns without calling _instrument again), so a second install without an intervening uninstall is unreachable through the supported API.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -0,0 +1,290 @@
...
+    # or on an explicit otel.metrics/otel.logs = true
+    metrics_env = std_traces_env or 'OTEL_EXPORTER_OTLP_METRICS_ENDPOINT' in os.environ
+    export_metrics = metrics is True or (metrics is not False and (metrics_env or cfg_endpoint is not None))
+    owns_mp = False
+    mp = meter_provider
+    if mp is None and export_metrics:
+        existing_mp = otel_metrics.get_meter_provider()
+        if 'Proxy' not in type(existing_mp).__name__:
+            mp = existing_mp
+        else:
```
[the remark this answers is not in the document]
christopherpestano: Keeping as is. Unlike ProxyTracerProvider, which is public in opentelemetry.trace and which we already check with isinstance, the metrics API has no public proxy class; the only candidate is opentelemetry.metrics._internal._ProxyMeterProvider. An isinstance check against a private _internal class is just as version fragile as the name check, and it fails at import time rather than degrading gracefully. Verified against opentelemetry-api 1.39.1.

packages/opentelemetry-instrumentation-pixeltable/tests/test_otel.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def test_protocol_grpc(self, tmp_path: Path) -> None:
+        _run_py(
+            'import opentelemetry.instrumentation.pixeltable as o\n'
+            'o.init()\n'
+            'from opentelemetry import trace\n'
+            'tp = trace.get_tracer_provider()\n'
+            'exporter = tp._active_span_processor._span_processors[0].span_exporter\n'
+            "assert type(exporter).__module__.startswith('opentelemetry.exporter.otlp.proto.grpc'), "
+            'type(exporter).__module__\n',
```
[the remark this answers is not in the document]
christopherpestano: Keeping as is. The point of this test is end to end verification that init() actually wires a grpc exporter into the provider; a unit test of _span_exporter() alone would not cover the wiring. The SDK is a pinned test dependency of this package, so a rename of those internals would surface as a test failure at upgrade time, which is exactly when we would want to notice.

docs/release/platform/observability.ipynb
```diff
@@ -0,0 +1,264 @@
...
+   "id": "obs-init",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "OTel bridge active -> Grafana Cloud\n"
+     ]
+    }
```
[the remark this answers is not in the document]
christopherpestano: Keeping as is. Every existing notebook in docs/release/platform (embedding-indexes, external-files, udfs-in-pixeltable) is committed with real outputs and execution counts, and the repo notebook checks pass with outputs present. This notebook follows that convention; the earlier error traceback was already removed by re-running in a clean environment.

pixeltable_cli/client/commands/serve.py
```diff
@@ -236,6 +241,16 @@ def _serve(args: argparse.Namespace) -> None:
...
+    if args.otel:
+        try:
+            import opentelemetry.instrumentation.pixeltable as pxt_otel
+        except ImportError as e:
+            raise excs.RequestError(
+                excs.ErrorCode.MISSING_REQUIRED,
+                "--otel requires the instrumentation package; install it with `pip install 'pixeltable[otel]'`",
+            ) from e
+        pxt_otel.init()
+
```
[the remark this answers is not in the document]
christopherpestano: This is user error, the command line already tells you in both help and the error above to install pixeltable[otel], it should be fairly obvious to the user what to do.

pixeltable/catalog/insertable_table.py
```diff
@@ -158,18 +158,20 @@ def insert(
...
-            fail_on_exception=fail_on_exception,
-            print_stats=print_stats,
-            return_rows=return_rows,
-        )
+            return self._insert_table_data_source(
+                data_source=data_source,
+                fail_on_exception=fail_on_exception,
+                print_stats=print_stats,
+                return_rows=return_rows,
+            )
```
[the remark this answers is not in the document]
christopherpestano: That's part 2

pixeltable/hooks.py
```diff
@@ -0,0 +1,306 @@
...
+def span_end(handle: SpanHandle | None, *, exc: BaseException | None = None, attrs: HookAttrs = None) -> None:
+    """End a span started with span_start(); a no-op for None handles."""
+    if not isinstance(handle, SpanHandle):
+        return
+    if handle.cv_token is not None:
+        try:
+            _current_span.reset(handle.cv_token)
+        except Exception as e:
+            _logger.debug(f'failed to reset ambient span: {e!r}')
+        handle.cv_token = None
```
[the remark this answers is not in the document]
christopherpestano: This should fail, we shouldn't encourage "soft" failure where wrong behavior silently goes through

pixeltable/exec/cache_prefetch_node.py
```diff
@@ -223,16 +223,23 @@ def __process_input_batch(self, input_batch: DataRowBatch, executor: futures.Thr
...
             self.in_flight_requests[f] = url
 
-    def __fetch_url(self, url: str) -> tuple[Path | None, Exception | None]:
+    def __fetch_url(self, url: str, hooks_ctx: Any) -> tuple[Path | None, Exception | None]:
+        """Runs on a worker thread of the ThreadPoolExecutor."""
+        hooks_token = hooks.restore_context(hooks_ctx)
         try:
-            return fetch_url(url), None
+            with hooks.span('pixeltable.media.fetch', level=hooks.DEBUG, url=url):
+                return fetch_url(url), None
```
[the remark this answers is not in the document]
christopherpestano: fixed

.github/workflows/pytest.yml
```diff
@@ -261,7 +270,7 @@ jobs:
           cd target/docs && mint broken-links
       - name: Lint
         if: ${{ matrix.test-category == 'lint' }}
-        run: ruff check pixeltable tests tool
+        run: ruff check pixeltable tests tool packages/opentelemetry-instrumentation-pixeltable/src packages/opentelemetry-instrumentation-pixeltable/tests
```
aaron-siegel: How about
`packages/**/src packages/**/tests`
?
More extensible and less of a mouthful

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+
+def _install_record_factory() -> None:
+    """Stamp pixeltable log records with otelTraceID/otelSpanID so logs are filterable by trace.
+
+    A record factory (the mechanism opentelemetry-instrumentation-logging uses) rather than a filter on
+    the 'pixeltable' logger: logging runs logger-level filters only on the originating logger, so a filter
+    there never sees records created on child loggers (`pixeltable.exec...` etc.), which is where nearly
+    all pixeltable records originate.
+    """
+    global _prev_record_factory  # noqa: PLW0603
```
aaron-siegel: I might argue for encapsulating in a class (`OtelEnv` or `RecordFactoryManager` or some such) rather than using globals.

Edit (after reading code below): why aren't they just methods of the instrumentor?

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -0,0 +1,359 @@
...
+        >>> import opentelemetry.instrumentation.pixeltable as pxt_otel
+        ...
+        ... pxt_otel.init(endpoint='http://localhost:4318')
+        ... pxt_otel.instrument_fastapi(app)
+    """
+    try:
+        from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor  # type: ignore[import-untyped]
+    except ImportError as exc:
+        raise excs.RequestError(
+            excs.ErrorCode.UNSUPPORTED_OPERATION,
```
aaron-siegel: Use Env.get().require_package() for this (it's the standardized way of checking for optional packages in pixeltable)

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -0,0 +1,359 @@
...
+            'instrument_fastapi() requires the `opentelemetry-instrumentation-fastapi` package. '
+            'To install it, run: `pip install -U opentelemetry-instrumentation-fastapi`',
+        ) from exc
+    kwargs.setdefault('tracer_provider', _state.tracer_provider)
+    FastAPIInstrumentor.instrument_app(app, **kwargs)
+
+
+def instrument_sqlalchemy(**kwargs: Any) -> None:
```
aaron-siegel: As we discussed in our 1:1, I'm not sure we need this - SQLAlchemy is an implementation detail of Pixeltable and I'm not convinced there's justification or value in exposing its internals to the user.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -0,0 +1,359 @@
...
+    cfg_headers = headers if headers is not None else config.get_string_value('headers', section='otel')
+    if metrics is None:
+        metrics = config.get_bool_value('metrics', section='otel')
+    if logs is None:
+        logs = config.get_bool_value('logs', section='otel')
+
+    use_grpc = _use_grpc(cfg_protocol)
+
+    # standard env vars beat pixeltable config: operators reconfigure deployed apps through them
+    std_traces_env = 'OTEL_EXPORTER_OTLP_ENDPOINT' in os.environ or 'OTEL_EXPORTER_OTLP_TRACES_ENDPOINT' in os.environ
```
aaron-siegel: we could align these with pixeltable config by naming our config parameters in a like manner.

```
[otel]
exporter_oltp_endpoint = '...'
```
etc.

Then rely on Pixeltable config resolution as the unified mechanism. This is how other integrations handle this. (In the Pixeltable config resolution chain, env vars always get highest priority, so the behavior would be as intended.)

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -0,0 +1,359 @@
...
+    tp = tracer_provider
+    if tp is None:
+        existing = trace.get_tracer_provider()
+        if not isinstance(existing, ProxyTracerProvider):
+            # the embedding application already configured a tracer provider; never clobber it
+            tp = existing
+        elif std_traces_env or cfg_endpoint is not None:
+            # plain SDK with an OTLP exporter; config-derived kwargs are withheld whenever the standard
+            # env vars are set, so the exporter's native env resolution wins
+            from opentelemetry.sdk.trace import TracerProvider
```
aaron-siegel: Move to top of file

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -0,0 +1,359 @@
...
+    metrics_env = std_traces_env or 'OTEL_EXPORTER_OTLP_METRICS_ENDPOINT' in os.environ
+    export_metrics = metrics is True or (metrics is not False and (metrics_env or cfg_endpoint is not None))
+    owns_mp = False
+    mp = meter_provider
+    if mp is None and export_metrics:
+        existing_mp = otel_metrics.get_meter_provider()
+        if 'Proxy' not in type(existing_mp).__name__:
+            mp = existing_mp
+        else:
+            from opentelemetry.sdk.metrics import MeterProvider
```
aaron-siegel: Is there a reason we need all these inline imports?

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -0,0 +1,359 @@
...
+    if export_logs:
+        logger_provider, log_handler = _build_log_export(use_grpc, cfg_endpoint, cfg_headers, cfg_service, logs_env)
+
+    # construction succeeded; apply global state
+    if owns_tp:
+        trace.set_tracer_provider(tp)
+    if owns_mp:
+        otel_metrics.set_meter_provider(mp)
+    if logger_provider is not None:
+        from opentelemetry._logs import set_logger_provider
```
aaron-siegel: same question

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -0,0 +1,359 @@
...
+
+        set_logger_provider(logger_provider)
+        logging.getLogger('pixeltable').addHandler(log_handler)
+    _state.owns_tracer_provider = owns_tp
+    _state.owns_meter_provider = owns_mp
+    _state.tracer_provider = tp
+    _state.meter_provider = mp
+    _state.logger_provider = logger_provider
+
+    from . import PixeltableInstrumentor
```
aaron-siegel: same question

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -0,0 +1,359 @@
...
+    """
+    protocol = (os.environ.get('OTEL_EXPORTER_OTLP_PROTOCOL') or cfg_protocol or 'http/protobuf').strip().lower()
+    return protocol == 'grpc'
+
+
+def _span_exporter(use_grpc: bool, endpoint: str | None, headers: str | None, *, withhold: bool) -> Any:
+    ep = None if withhold else endpoint
+    hdrs = _parse_headers(headers) if headers is not None and not withhold else None
+    if use_grpc:
+        from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter as GrpcSpanExporter
```
aaron-siegel, line 278: same question here as well (I think all these packages are guaranteed to be present via `pip install pixeltable[otel]`?) and in various other places below

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/version.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.1.0'
```
aaron-siegel, line 1: Do we not want to use dynamic versioning?

pixeltable/config.py
```diff
@@ -415,7 +415,9 @@ def __merge_config(
 
     def lookup_env(self, section: str, key: str, default: Any = None) -> Any:
         override_var = f'{section}.{key}'
-        env_var = f'{section.upper()}_{key.upper()}'
+        # the OTEL_* env namespace is reserved by the OpenTelemetry spec; keep derived names pixeltable-specific
```
aaron-siegel: I don't think this is a good idea. Our usual methodology is to align our config with our integrations. So, for example, the [openai] section uses OPENAI_ env prefix, even though that's "reserved" by Open AI. The overrides mentioned below are then handled automatically by the env -> cfg file resolution chain, in a way that's consistent with how the integration naturally resolves them.

Either way, let's please preserve the canonical mapping of cfg file sections to env prefixes and not start introducing ad hoc exceptions.

pixeltable/config.py
```diff
@@ -582,6 +584,15 @@ def get_value_source(self, key: str, section: str = 'pixeltable') -> Path | Lite
         'app_name': 'Optional name for your application (for OpenRouter analytics)',
         'rate_limit': 'Rate limit for OpenRouter API requests',
     },
+    'otel': {
+        'endpoint': 'OTLP collector endpoint (eg http://localhost:4318); overridden by OTEL_EXPORTER_OTLP_ENDPOINT',
```
aaron-siegel: cf. above comments: simply call this cfg parameter 'exporter_oltp_endpoint' and keep the 'otel' <-> 'OTEL_' mapping as intended, and the cfg resolution chain will naturally take care of things. As it is, we're basically reimplementing the cfg resolution logic in the otel code.

pixeltable/telemetry.py
```diff
@@ -0,0 +1,304 @@
+"""Dependency-free instrumentation hooks.
```
aaron-siegel, line 1: Let's call this module something like `otel.py`? There are already a lot of "hooks" floating around.

pixeltable/telemetry.py
```diff
@@ -0,0 +1,304 @@
...
+_logger = logging.getLogger(__name__)
+
+TRACE = 5
+DEBUG = 10
+INFO = 20
+
+HookAttrs = dict[str, Any] | Callable[[], dict[str, Any]] | None
+
+
+class Subscriber:
```
aaron-siegel: Is this user-facing? If so, all methods need docstrings

pixeltable/hooks.py
```diff
@@ -0,0 +1,304 @@
...
+_registry_lock = threading.Lock()
+_SUBSCRIBERS: tuple[Subscriber, ...] = ()
+_span_level = INFO
+_current_span: ContextVar[SpanHandle | None] = ContextVar('pxt_current_span', default=None)
+_logged_error_methods: WeakKeyDictionary[Subscriber, set[str]] = WeakKeyDictionary()
+
+
+def subscribe(subscriber: Subscriber) -> None:
+    """Register a subscriber to receive instrumentation callbacks; idempotent."""
+    global _SUBSCRIBERS  # noqa: PLW0603
```
aaron-siegel: To me this is another sign it'd be useful to have an OtelEnv abstraction (see above)

pixeltable_cli/client/commands/serve.py
```diff
@@ -236,6 +241,16 @@ def _serve(args: argparse.Namespace) -> None:
         _print_dry_run(cfg, args.json)
         return
 
+    if args.otel:
+        try:
+            import opentelemetry.instrumentation.pixeltable as pxt_otel
```
aaron-siegel: use Env.get().require_package()

Makefile
```diff
@@ -182,18 +183,21 @@ typecheck: install
...
 	@mypy pixeltable/catalog/model.py
+	@echo 'Running `mypy` on $(OTEL_PKG) ...'
+	@MYPYPATH=$(OTEL_PKG)/src mypy --explicit-package-bases --namespace-packages \
+		$(OTEL_PKG)/src/opentelemetry/instrumentation/pixeltable $(OTEL_PKG)/tests
 
 .PHONY: lint
 lint: install
 	@echo 'Running `ruff check` ...'
-	@ruff check pixeltable pixeltable_cli tests tool
+	@ruff check pixeltable pixeltable_cli tests tool $(OTEL_PKG)/src $(OTEL_PKG)/tests
```
aaron-siegel, line 193: Would `packages/**/src` etc. work here?

pyproject.toml
```diff
@@ -78,6 +78,15 @@ dependencies = [
 # `pxt serve` hosts a user-defined FastAPI app and shells out to uvicorn; the daemon
 # behind `pxt ls`/`pxt describe`/etc. is stdlib-only and does not need this extra.
 serve = ["fastapi[standard]>=0.120.0"]
+# OpenTelemetry instrumentation. Core pixeltable carries no OTEL dependency; the instrumentation is a
+# separate distribution, opt-in via opentelemetry.instrumentation.pixeltable. The SDK and exporters live
+# here rather than in that package so it stays api-only, per contrib convention.
+otel = [
+    "opentelemetry-instrumentation-pixeltable",
```
aaron-siegel: Pin this to an explicit version the same way we do with `pixeltable_pgserver`

pyproject.toml
```diff
@@ -271,7 +286,16 @@ strict_optional = false
 warn_return_any = true
 warn_unused_ignores = true
 
+[tool.pyright]
+typeCheckingMode = "off"
```
aaron-siegel, line 291: What does this do?

aaron-siegel (changes requested):

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: Changed to a private method, we should keep this for debugging without exposing this to users then.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: My bad, I had thought since it's a separate package we would avoid importing too many things to avoid creating unneeded interdependencies. But I guess this is fine, fixed.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: The sdk isn't a required dependency of this package so we can't put it at the top of file, moved to top of function. As for the specific reason that we can't put it as a required dependency of `opentelemetry-instrumentation-pixeltable` it is common convention for all `opentelemetry-instrumentation-*` packages to purely import the otel api as a required package.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: The otel sdk is not a required dependency so it cannot go at the top of file, moved to top of function and added an Env.get().require_package() to guard.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: This one can go at top of file, checked and it's a part of the Otel-api which is a required dependency

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: This one can go top of file moved up

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: These are not guaranteed though if the person using pixeltable is instrumenting otel the canonical way of `opentelemetry-instrument` which auto installs the `opentelemetry-instrumentation-pixeltable` but does not guarantee that the sdk is installed. In this case the exports have to remain where they are since we can't even guarantee that the user has both grpc and http exports both installed. Added guard to make sure the correct exporter being requested is installed.

pixeltable/telemetry.py
christopherpestano: No these are not user-facing, users should not be writing their own Subscribers to pixeltable hooks, the only thing they should be aware of are `init()` and `instrument_fastapi()` from the instrumentation library. Everything else is an implementation detail, they shouldn't even really be making their own spans or events using the hooks.py that is only for internal consumption.

pixeltable_cli/client/commands/serve.py
christopherpestano: Fixed, changed it to Env.get().require_package() with custom error message for `pip install pixeltable[otel]`

.github/workflows/pytest.yml
christopherpestano: Yeah that works, changed it as such

pyproject.toml
christopherpestano: Once we added the otel package under packages/, Pylance couldn't resolve its imports, so I added a [tool.pyright] section with extraPaths pointing at its src folder.

The catch is that just having a [tool.pyright] section flips Pylance out of its default. With no config it defaults to typeCheckingMode "off", but the moment a pyright config exists it switches to pyright's default (basic/standard), which turned typechecking on across the whole repo and lit up errors everywhere.

So typeCheckingMode = "off" is just restoring the behavior we already had before. Nothing was ever explicitly turned off; it was off by default, and adding the config section is what forced us to set it back.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: Nevermind, actually moving this to top of file creates a circular import issue, moved back to top of function

Makefile
christopherpestano: In make, packages/**/src is passed to the shell literally where ** acts as a plain * (one level, not recursive) so it only works by luck with one package, and it breaks the mypy line because the shell doesn't glob-expand the right-hand side of MYPYPATH=..., so keeping the explicit $(OTEL_PKG) variable ends up working out better for the Makefile at least.

pyproject.toml
christopherpestano: Sure pinned version

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/__init__.py
christopherpestano: Done, moved the record factory install/uninstall into PixeltableInstrumentor as methods and the previous-factory reference is now an instance attribute, so there are no module-level globals anymore. Originally I was avoiding adding any methods not present in the base BaseInstrumentor class, but this reads better.

pixeltable/config.py
christopherpestano: Done, renamed the keys to exporter_otlp_endpoint, exporter_otlp_protocol, and exporter_otlp_headers so they map onto the standard OTEL_EXPORTER_OTLP_* env vars through the config chain, and removed the reimplemented resolution from the otel code. Two behavior notes: an explicit init() kwarg now beats the env var, matching pixeltable's config override precedence, and only the generic OTEL_EXPORTER_OTLP_ENDPOINT is honored now, not the signal-specific traces/metrics/logs vars.

pixeltable/config.py
christopherpestano: Done, the PIXELTABLE_OTEL special case is gone and otel uses the canonical OTEL_ prefix like every other section; see the exporter_otlp_endpoint thread for the full change.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: Done, aligned as suggested; see the exporter_otlp_endpoint config thread for details.

pixeltable/hooks.py
christopherpestano: Done, added a TelemetryEnv singleton following the Env convention that owns the subscriber registry, with subscribe/unsubscribe/active as methods on it. I named it TelemetryEnv rather than OtelEnv since hooks.py is provider agnostic. The span emission API stays module-level and the current-span pointer stays a ContextVar, since it has to be per-context rather than process-global.

pixeltable/telemetry.py
christopherpestano: Done, renamed it to telemetry.py. I went with telemetry rather than otel since the module is provider agnostic (OTEL is just one subscriber), but this gets it out of the crowded hooks namespace like you wanted.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/version.py
christopherpestano: I think it's best to keep this static. The opentelemetry-python-contrib contributing guidelines note that a new instrumentation should carry an explicit X.Y.dev version, and all the contrib packages (plus third party ones like Traceloop's) set their version from a static version.py through hatch rather than from git tags. Going dynamic would tie this package's version to pixeltable's git tags, which is the coupling the contrib convention avoids, since these packages version independently of the library they instrument.

Guidelines: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/CONTRIBUTING.md

pixeltable/telemetry.py
```diff
@@ -0,0 +1,330 @@
...
+        self.pending_attrs: dict[str, Any] | None = None
+
+
+_current_span: ContextVar[SpanHandle | None] = ContextVar('pxt_current_span', default=None)
+
+
+class TelemetryEnv:
+    """Process-global instrumentation state: the subscriber registry and the span-level threshold.
+
+    Singleton accessed via TelemetryEnv.get(), mirroring pixeltable.env.Env. Subscriber management lives
```
mkornacker: If this is part of the global environment, it should be in Env.

If this is thread-specific, it should be in Runtime.

pixeltable/telemetry.py
```diff
@@ -0,0 +1,330 @@
...
+    _span_level: int
+    _logged_error_methods: WeakKeyDictionary[Subscriber, set[str]]
+
+    @classmethod
+    def get(cls) -> TelemetryEnv:
+        if cls._instance is not None:
+            return cls._instance
+        with cls._lock:
+            if cls._instance is None:
+                cls._instance = TelemetryEnv()
```
mkornacker: Why not call this class Telemetry and add an instance to Env?

pixeltable/telemetry.py
```diff
@@ -0,0 +1,330 @@
...
+        return None
+
+    def exit_context(self, token: Any) -> None:
+        pass
+
+
+class SpanHandle:
+    """Opaque handle for an emitted span; created by span_start() and consumed by span_end()."""
+
+    __slots__ = ('cv_token', 'pending_attrs', 'subs', 'tokens')
```
mkornacker, line 76: Comment: why slots

pixeltable/exec/expr_eval/evaluators.py
```diff
@@ -99,6 +100,17 @@ def reset(self) -> None:
         if self.call_args_queue is not None:
             self.call_args_queue = asyncio.Queue[FnCallArgs]()
 
+    def _cell_span(self, row: exprs.DataRow) -> Any:
```
mkornacker: Is there a more meaningful return type?

pixeltable/exec/expr_eval/evaluators.py
```diff
@@ -99,6 +100,17 @@ def reset(self) -> None:
...
+        """Span for one row's UDF call, nested under the row's span; no-op when the row has no span."""
```
mkornacker: Should state that this is a contextmgr and a drop-in for telemetry.span().

pixeltable/exec/cache_prefetch_node.py
```diff
@@ -223,16 +224,31 @@ def __process_input_batch(self, input_batch: DataRowBatch, executor: futures.Thr
...
+        # carry the ambient instrumentation span onto the worker threads so fetch spans nest correctly
+        hooks_ctx = telemetry.capture_context()
         for url in cache_misses:
-            f = executor.submit(self.__fetch_url, url)
+            f = executor.submit(self.__fetch_url, url, hooks_ctx)
             _logger.debug(f'submitted {url} for idx {url_pos[url]}')
             self.in_flight_requests[f] = url
 
-    def __fetch_url(self, url: str) -> tuple[Path | None, Exception | None]:
+    def __fetch_url(self, url: str, hooks_ctx: Any) -> tuple[Path | None, Exception | None]:
```
mkornacker: Can you type that more meaningfully?

pixeltable/exec/object_store_save_node.py
```diff
@@ -300,26 +300,32 @@ def __process_input_batch(self, input_batch: DataRowBatch, executor: futures.Thr
...
             file_size = work_item.src_path.stat().st_size
-            f = executor.submit(self.__persist_media_file, work_item)
+            f = executor.submit(self.__persist_media_file, work_item, hooks_ctx)
             self.in_flight_requests[f] = ObjectStoreSaveNode.WorkDesignator(
                 str(work_item.src_path), work_item.destination, file_size
             )
             _logger.debug(f'submitted {work_item}')
 
-    def __persist_media_file(self, work_item: WorkItem) -> tuple[str | None, Exception | None]:
+    def __persist_media_file(self, work_item: WorkItem, hooks_ctx: Any) -> tuple[str | None, Exception | None]:
```
mkornacker: Anything better than Any?

pixeltable/telemetry.py
christopherpestano: Added a comment. One handle is allocated per emitted span, which at DEBUG level means per row and per UDF cell, so slots drops the per instance dict to cut allocation cost and memory on that hot path.

pixeltable/exec/expr_eval/evaluators.py
christopherpestano: Changed to contextlib.AbstractContextManager[telemetry.SpanHandle | None], which covers both the nullcontext and telemetry.span() cases.

pixeltable/exec/expr_eval/evaluators.py
christopherpestano: Done. The docstring now states it is a context manager with telemetry.span() semantics and describes the no span case.

pixeltable/exec/cache_prefetch_node.py
christopherpestano: Made the snapshot class public as CtxSnapshot and typed hooks_ctx as telemetry.CtxSnapshot | None.

pixeltable/exec/object_store_save_node.py
christopherpestano: Same as the prefetch node, typed as telemetry.CtxSnapshot | None.

pixeltable/store.py
```diff
@@ -548,26 +548,29 @@ def insert_rows(
...
-                        pk = (*rowid, v_min)
-                    else:
-                        # UUID7 appears to be a more performant choice for Postgresql, however for CockroachDB UUID4 is
-                        # recommended.
-                        assert not Env.get().is_using_cockroachdb, 'TODO: implement for unversioned tables [PXT-1101]'
-                        pk = (uuid7(),)
-                    assert len(pk) == len(self._pk_cols)
-                    table_row, num_row_exc = row_builder.create_store_table_row(row, cols_with_excs, pk)
-                    num_excs += num_row_exc
+                with telemetry.span('pixeltable.store.build_rows', level=telemetry.DEBUG, rows=len(row_batch)):
```
mkornacker, line 551: Why not name it after the method?

pixeltable/telemetry.py
christopherpestano: It is process wide, but unlike Env its contents cannot be rebuilt. Subscribers are pushed in by outside callers, usually before Pixeltable initializes, and Pixeltable has no way to find them again after a teardown since core has no knowledge of the otel side. The instrumentor holding the subscriber is a process lifetime singleton whose instrumented flag stays set, so it will not instrument a second time and tracing would silently stop after a reinit. The registry has to match that lifetime, like handlers in the logging module.

pixeltable/telemetry.py
christopherpestano: Renamed it to SubscriberRegistry and moved active() out to a module level function, so the class now holds only the registry. Putting an instance on Env runs into the lifecycle problem from the other thread: subscribe() happens in instrument(), typically before Env exists, and an Env rebuild would silently drop the subscriber.

pixeltable/store.py
christopherpestano: The section under this span is not the insert itself, it only converts a batch of computed DataRows into store table rows before anything is written, so naming it after insert_rows would be misleading. The actual write is reported separately by the pixeltable.sa.insert_rows span in sql_insert. In general I tried to avoid naming spans after internal code and instead describe a unit of work someone reading a trace would understand, since these names show up in trace UIs.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+            telemetry.set_span_level(_resolve_span_level(span_level))
+        self._subscriber = _OtelSubscriber(kwargs.get('tracer_provider'), kwargs.get('meter_provider'))
+        SubscriberRegistry.get().subscribe(self._subscriber)
+        self._install_record_factory()
+
+    def _uninstrument(self, **kwargs: Any) -> None:
+        if self._subscriber is not None:
+            SubscriberRegistry.get().unsubscribe(self._subscriber)
+            self._subscriber = None
+        self._uninstall_record_factory()
```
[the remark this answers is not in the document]
christopherpestano: Minor, not worth fixing

pixeltable/utils/filecache.py
```diff
@@ -241,7 +242,7 @@ def add(self, tbl_id: UUID, col_id: int, url: str, path: Path) -> Path:
         new_path = entry.path
         os.rename(str(path), str(new_path))
         new_path.touch(exist_ok=True)
-        _logger.debug(f'FileCache: cached url {url} with file name {new_path}')
+        _logger.debug(f'FileCache: cached url {redact_url(url)} with file name {new_path}')
```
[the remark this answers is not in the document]
christopherpestano: This is fine for now, low cost

pixeltable/utils/filecache.py
```diff
@@ -216,7 +217,7 @@ def lookup(self, url: str) -> Path | None:
         self.cache.move_to_end(key, last=True)
         self.num_hits += 1
         self.keys_retrieved.add(key)
-        _logger.debug(f'file cache hit for {url}')
+        _logger.debug(f'file cache hit for {redact_url(url)}')
         return path
```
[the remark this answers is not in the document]
christopherpestano: Relatively low cost we can optimize in part 2

docs/release/platform/observability.ipynb
```diff
@@ -0,0 +1,257 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "obs-intro",
+   "metadata": {},
+   "source": [
+    "# Observability with OpenTelemetry\n",
```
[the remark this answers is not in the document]
christopherpestano: Basically every notebook does this same thing, ignore

pixeltable/utils/filecache.py
```diff
@@ -206,7 +207,7 @@ def lookup(self, url: str) -> Path | None:
         key = self._url_hash(url)
         entry = self.cache.get(key, None)
         if entry is None:
-            _logger.debug(f'file cache miss for {url}')
+            _logger.debug(f'file cache miss for {redact_url(url)}')
             return None
```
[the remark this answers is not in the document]
christopherpestano: Ignore, minor cost

pixeltable/utils/http.py
```diff
@@ -150,7 +161,8 @@ def fetch_url(url: str, allow_local_file: bool = False) -> Path:
     from .local_store import TempStore
     from .object_stores import ObjectOps
 
-    _logger.debug(f'fetching url={url} thread_name={threading.current_thread().name}')
+    safe_url = redact_url(url)
+    _logger.debug(f'fetching url={safe_url} thread_name={threading.current_thread().name}')
     parsed = urllib.parse.urlparse(url)
 
     if len(parsed.scheme) <= 1:
```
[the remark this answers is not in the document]
christopherpestano: Bad

pixeltable/utils/http.py
```diff
@@ -35,9 +35,19 @@
...
+        # A malformed URL is unsafe to reproduce because we cannot reliably identify its credential-bearing parts.
+        return '<redacted-url>'
+    netloc = parsed.netloc.rpartition('@')[2]
+    return urllib.parse.urlunsplit((parsed.scheme, netloc, parsed.path, '', ''))
+
+
 def is_retriable_error(exc: Exception) -> tuple[bool, float | None]:
-    """Attempts to guess if the exception indicates a retriable eror. If that is the case, returns True
-    and the retry delay in seconds."""
+    """Attempts to guess if the exception indicates a retriable error. If that is the case, returns True"""
```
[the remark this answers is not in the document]
christopherpestano: Fixed

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -193,6 +197,16 @@ def _dispatch_input_rows(self) -> None:
...
+        if telemetry.active():
+            # per-row spans nest their UDF cell spans; suppressed (None) unless DEBUG is enabled under an
+            # operation span, so a bare query with no operation span stays dark
+            for row in rows:
+                if row.parent_row is None and len(self._span_rows) < self.MAX_ROW_SPANS:
+                    row.span = telemetry.span_start(
+                        'pixeltable.row', level=telemetry.DEBUG, parent=telemetry.current_span()
+                    )
+                    if row.span is not None:
+                        self._span_rows.append(row)
```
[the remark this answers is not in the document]
christopherpestano: Minor optimization doesn't make sense once part 2 is in since it'll always have that parent span in that case

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+from typing import Any, Callable
+
+from opentelemetry import context as otel_context, metrics as otel_metrics, trace
+from opentelemetry.context import Context, Token
+
+# opentelemetry-instrumentation marks instrumentor.py with a module-level `# type: ignore`, so mypy
+# can't see BaseInstrumentor despite the package shipping py.typed
+from opentelemetry.instrumentation.instrumentor import BaseInstrumentor  # type: ignore[attr-defined]
+from opentelemetry.trace import StatusCode, set_span_in_context
+
```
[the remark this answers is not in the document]
christopherpestano: Huh no it's not?

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+    def on_span_end(self, token: Any, exc: BaseException | None, attrs: dict[str, Any] | None) -> None:
+        assert isinstance(token, _SpanToken)
+        span = token.span
+        for k, v in _clean_attrs(attrs).items():
+            span.set_attribute(k, v)
+        if exc is not None:
+            if isinstance(exc, Exception):
+                span.record_exception(exc)
+            span.set_status(StatusCode.ERROR, str(exc))
+        span.end()
```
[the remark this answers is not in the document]
christopherpestano: Huh yes it does, weird hallucination set status actually DOES accept code + error.

pixeltable/telemetry.py
```diff
@@ -0,0 +1,345 @@
...
+        with self._lock:
+            self._subscribers = tuple(s for s in self._subscribers if s is not subscriber)
+
+    def _log_subscriber_error(self, subscriber: Subscriber, method: str, exc: Exception) -> None:
+        # warn once per (subscriber, method), then drop to debug to avoid log storms from per-row failures
+        with self._lock:
+            logged = self._logged_error_methods.setdefault(subscriber, set())
+            level = logging.DEBUG if method in logged else logging.WARNING
+            logged.add(method)
+        _logger.log(level, f'instrumentation subscriber {type(subscriber).__name__}.{method}() failed: {exc!r}')
```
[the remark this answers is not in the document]
christopherpestano: Explicitly added parent class Hashable to Subscriber, as well as default __hash__ implementation

mkornacker (approved):

aaron-siegel (approved):

pixeltable/telemetry.py
```diff
@@ -0,0 +1,347 @@
...
```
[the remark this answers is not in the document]
christopherpestano: Subs are always hashable

pixeltable/exec/object_store_save_node.py
```diff
@@ -300,26 +300,34 @@ def __process_input_batch(self, input_batch: DataRowBatch, executor: futures.Thr
...
+        hooks_ctx = telemetry.capture_context()
         for work_item in work_to_do:
             # determine size before file gets moved
             file_size = work_item.src_path.stat().st_size
-            f = executor.submit(self.__persist_media_file, work_item)
+            f = executor.submit(self.__persist_media_file, work_item, hooks_ctx)
             self.in_flight_requests[f] = ObjectStoreSaveNode.WorkDesignator(
                 str(work_item.src_path), work_item.destination, file_size
             )
             _logger.debug(f'submitted {work_item}')
```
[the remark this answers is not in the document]
christopherpestano: When each worker attaches that context they get a thread local so the shared state isn't mutated.

pixeltable/utils/filecache.py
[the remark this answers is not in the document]
christopherpestano: Minor optimization, later we are going to switch how we handle redaction anyway

pixeltable/telemetry.py
```diff
@@ -0,0 +1,347 @@
...
+
+from __future__ import annotations
+
+import contextlib
+import dataclasses
+import logging
+import threading
+from contextvars import ContextVar, Token
+from typing import Any, Callable, Hashable, Iterator
+from weakref import WeakKeyDictionary
```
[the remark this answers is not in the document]
christopherpestano: What kind of hallucination is this? We didn't replace WeakKeyDictionary

PR #1429, opened 2026-06-24
PXT-1224: add_embedding_index() uses index definition

mkornacker:
For unnamed indices, uses the actual definition, not the auto-generated name, to detect duplicates.

### DIFF
--- a/pixeltable/catalog/local_table.py
+++ b/pixeltable/catalog/local_table.py
@@ -51,6 +51,8 @@ if TYPE_CHECKING:
     import pixeltable.plan
     from pixeltable.globals import TableDataSource
 
+    from .table_version import TableVersion
+
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


pixeltable/catalog/local_table.py
```diff
@@ -820,10 +823,39 @@ def add_embedding_index(
                 column=col,  # Pass column for shape validation
             )
             _ = idx.create_value_expr(col)  # validation only; result discarded
+
+            if idx_name is None:
+                # Unnamed index: duplicate detection is by index definition on this column.
```
aaron-siegel, line 828: Just a thought; I wonder if we should simply be preventing multiple unnamed index creation (i.e., reject unnamed index creation if there is already an index on the column; require a name to add 2 or more indices)

I'm worried this might still lead to surprising behavior, e.g., if someone calls add_embedding_index(if_exists='ignore') with *slightly* different index configuration, not anticipating a second index being created.

aaron-siegel (approved):

pixeltable/catalog/local_table.py
mkornacker: A good question to discuss.

PR #1434, opened 2026-06-26
Class-based table declarators

aaron-siegel:
Introduces class-based ("dataclass-style") Table declarators via a new `TableModelMetaclass`. The metaclass has two instances, `TableModel` and `ViewModel`, that are used to declare tables and views, respectively. A model gets resolved to a `Table` instance when any action is taken on it, and most of the functionality on `TableModel` is a pass-through to that underlying `Table`. Columns are added promptly to the model namespace at declaration time, and can be referenced in the usual way for subsequent computed columns.

Table methods (`select()`, `collect()`, `where()`, etc.) can be invoked directly on `TableModel`. In order to avoid having to maintain 20+ method signatures in parallel, a mechanism is introduced where (i) selected table methods are passed through to the underlying `Table` class via `TableModelMetaclass.getattr()`; and (ii) a `model.pyi` file is auto-generated via a new `generate_type_stubs.py` tool, based on the signatures in `pxt.Table`, ensuring that the signatures are both auto-synced and are cognizable to static type checkers.

Major TODOs:

- Extend the validation logic of the model against existing tables
- Auto-merge of schemas in response to model changes
- Embedding index creation is currently a separate DB transaction; it should be brought under the transaction umbrella of table creation


tests/test_table_model.py
```diff
@@ -0,0 +1,736 @@
...
+        assert str(tbl._path()) == expected_path
+
+        if root != '':
+            return  # Exact metadata comparison only applies to the '' case
+
+        metadata: dict[str, Any] = dict(tbl.get_metadata())
+
+        metadata.pop('id')
+        metadata.pop('version_created')
+        print(metadata)
```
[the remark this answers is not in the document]
aaron-siegel: it's not debug logging - it's there to facilitate test evolution when the metadata changes.

pixeltable/catalog/catalog.py
```diff
@@ -1457,32 +1457,20 @@ def create_table(
...
-                path.name,
-                schema,
-                primary_key=primary_key,
-                comment=comment,
-                custom_metadata=custom_metadata,
-                media_validation=media_validation,
-                create_default_idxs=create_default_idxs,
-                is_versioned=is_versioned,
+        columns = [Column.create(name, spec) for name, spec in schema.items()]
+        create_fn = retry_loop(for_write=True)(
```
mkornacker: Comment: why not the normal @retry_loop pattern

pixeltable/catalog/catalog.py
```diff
@@ -1512,69 +1543,103 @@ def create_view(
...
-                predicate=where,
-                sample_clause=sample_clause,
-                is_snapshot=is_snapshot,
-                create_default_idxs=create_default_idxs,
-                iterator_call=iterator,
-                comment=comment,
-                custom_metadata=custom_metadata,
-                media_validation=media_validation,
+        additional_columns_ = [Column.create(name, spec) for name, spec in additional_columns.items()]
+        create_fn = retry_loop(for_write=True)(
```
mkornacker: Same here.

pixeltable/catalog/view.py
```diff
@@ -125,17 +126,19 @@ def _create(
...
-            if not col.is_computed:
-                continue
-            # make sure that the value can be computed in the context of the base
-            if col.value_expr is not None and not col.value_expr.is_bound_by([base]):
-                raise excs.RequestError(
-                    excs.ErrorCode.UNSUPPORTED_OPERATION,
-                    f'Column {col.name!r}: Value expression cannot be computed in the context of the '
-                    f'base table {base.tbl_name()!r}',
-                )
+        # same for value exprs; but we can skip this check if an explicit tbl_id is provided, as in the case of
```
mkornacker: Just looking at this code doesn't make it clear why the check can be skipped. Does anything break if we don't skip it?

tests/pixeltable_cli/test_bridge.py
```diff
@@ -1,13 +1,12 @@
...
 
-import numpy as np
 import pytest
 
 import pixeltable as pxt
 from pixeltable.functions.video import frame_iterator
 from pixeltable_cli.server import bridge
 
-from ..utils import get_test_video_files
+from ..utils import dummy_embedding, get_test_video_files
```
mkornacker, line 9: Why not  just local_embedding, which we already have?

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+            clip_idx = EmbeddingIndex(img, embedding=dummy_embedding.using(n=768))
+
+        expected_path = 'test_table' if root == '' else f'{root}/test_table'
+        if root != '':
+            pxt.create_dir(root, parents=True)
+
+        with capture_console_output(match=rf'Created {expected_path!r} from TableModel `ExampleTableModel`.'):
+            tbl = ExampleTableModel.create(root)
+
+        assert str(tbl._path()) == expected_path
```
mkornacker: get_metadata().path

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+        if root != '':
+            return  # Exact metadata comparison only applies to the '' case
+
+        metadata: dict[str, Any] = dict(tbl.get_metadata())
+
+        metadata.pop('id')
+        metadata.pop('version_created')
+        print(metadata)
+
+        assert {name: info['type_'] for name, info in metadata['columns'].items()} == {
```
mkornacker: A better test: create an analogous table with pxt.create_table(), then assert that the get_metadata() output matches. That would be far easier to inspect for correctness.

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+                'base': None,
+                'iterator_call': None,
+            },
+            tbl.get_metadata(),
+        )
+
+    def test_table_model_query(self, uses_db: None) -> None:
+        class ExampleTableModel(pxt.TableModel, name='test_table'):
+            id: pxt.Required[pxt.Int]
+            name: pxt.String
```
mkornacker: Let's check with our design partners whether they'd prefer 'pxt.String | None', which would match pydantic. Class-based schemas look like pydantic models, and flipping the meaning of a plain '<field>: <type>' might confuse people.

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+            string_add = name + ' suffix'
+            string_radd = 'prefix ' + name
+            string_mul = name * 3
+            string_rmul = 3 * name
+            type_cast = arr.astype(pxt.Array[(2, 3), np.float32])  # type: ignore[misc]
+
+        tbl = AllExprsTableModel.create()
+
+        metadata = tbl.get_metadata()
+        assert {name: info['type_'] for name, info in metadata['columns'].items()} == {
```
mkornacker: Here too: it would be far easier to validate correctness with an analogous table from pxt.create_table().

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+from .utils import (
+    assert_dicts_eq,
+    assert_table_metadata_eq,
+    capture_console_output,
+    dummy_embedding,
+    pxt_raises,
+    validate_update_status,
+)
+
+pytestmark = pytest.mark.local('Models')
```
mkornacker: Why? All these tests absolutely need to work for proxy tables.

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+            computed_with_special_props_2 = Column(value=img.rotate(90), destination='.')
+
+            clip_idx = EmbeddingIndex(img, embedding=dummy_embedding.using(n=768))
+
+        expected_path = 'test_table' if root == '' else f'{root}/test_table'
+        if root != '':
+            pxt.create_dir(root, parents=True)
+
+        with capture_console_output(match=rf'Created {expected_path!r} from TableModel `ExampleTableModel`.'):
+            tbl = ExampleTableModel.create(root)
```
mkornacker: Why not always pxt.create_all()?

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+            match=rf"Created '{prefix}test_table' from TableModel `ExampleTableModel`.\n"
+            rf"Created '{prefix}test_view' from ViewModel `ExampleViewModel`.\n"
+            rf"Created '{prefix}test_subview' from ViewModel `ExampleSubviewModel`.\n"
+            rf"Created '{prefix}test_view_from_query' from ViewModel `ExampleViewModelFromQuery`.\n"
+            rf"Created '{prefix}test_subview_from_query' from ViewModel `ExampleSubviewModelFromQuery`."
+        ):
+            if create_all:
+                pxt.create_all(root)
+            else:
+                _ = ExampleTableModel.create(root)
```
mkornacker: Why test this separately?

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+                _ = ExampleViewModel.create(root)
+                _ = ExampleSubviewModel.create(root)
+                _ = ExampleViewModelFromQuery.create(root)
+                _ = ExampleSubviewModelFromQuery.create(root)
```
mkornacker: Missing:
- an analogous view hierarchy created with create_table/_view()
- an insert and query, to demonstrate that the tables are now functional.

I think that should be part of every test here.

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+        ):
+            pass
+
+        ExampleTableModel.create()  # should succeed; schema matches existing table
+
+        with pxt_raises(
+            excs.ErrorCode.SCHEMA_MISMATCH,
+            match=r"TableModel `BadTableModel` is defined as a table, but the existing 'test_view' is a view.",
+        ):
+            BadTableModel.create()
```
mkornacker: Why this instead of create_all()? Isn't create_all() going to be the recommended best practice? I don't think we should even expose .create(), it would be difficult for users to track what exactly needs a .create() call. It goes against the spirit of the schema declaration itself, which always contains all tables. With .create you'd need to add/run code somewhere that knows exactly which tables already exist and which don't. A create_all/bind_all() works as a generic preamble in an app server.

tests/test_table_model.py
```diff
@@ -0,0 +1,738 @@
...
+
+        with capture_console_output(
+            match=rf"Created '{prefix}test_table' from TableModel `ExampleTableModel`.\n"
+            rf"Created '{prefix}test_view' from ViewModel `ExampleViewModel`.\n"
+            rf"Created '{prefix}test_subview' from ViewModel `ExampleSubviewModel`.\n"
+            rf"Created '{prefix}test_view_from_query' from ViewModel `ExampleViewModelFromQuery`.\n"
+            rf"Created '{prefix}test_subview_from_query' from ViewModel `ExampleSubviewModelFromQuery`."
+        ):
+            if create_all:
+                pxt.create_all(root)
```
mkornacker: An unscoped create_all() seems very dangerous to me, you might accidentally pull in some table classes via imports. I think this is really going to trip up users.

Why not follow the sqlalchemy model and have an explicit Base, and then do Base.create_all()? That feels a lot safer/predictable.

mkornacker (changes requested):

pixeltable/catalog/catalog.py
aaron-siegel: ? Isn't that what it is? (The normal retry_loop pattern)

```python
@retry_loop
def fn() -> ...

fn()
```

is absolutely identical to

```python
def fn() -> ...

retry_loop(fn)()
```

so I'm not sure what the comment would say? Just that one syntax happens to be more convenient than the other in this location? Choice of syntax isn't something I'd ordinarily comment on, but happy to leave it if you think it'd be helpful

pixeltable/catalog/catalog.py
aaron-siegel: (see above)

pixeltable/catalog/view.py
aaron-siegel: The issue is that we can now have additional_columns that refer to other columns *in the view*. Those cannot be evaluated in the context of the base table (nor in the context of any already-existing table; they depend on columns that haven't been created yet).

But, I took another look and I've reinstituted the check for TableModels and broadened it to allow for sibling columns if a tbl_id is specified - PTAL.

The column creation logic for views could use some refactoring, but I'd like to make that a follow-up. Validation should be functionally correct now.

tests/pixeltable_cli/test_bridge.py
aaron-siegel: Tests were written before local_embedding was merged in. Can we make this a follow-up?

tests/test_table_model.py
aaron-siegel: Proxy logic has been implemented and local marker removed.

tests/test_table_model.py
aaron-siegel: Done

tests/test_table_model.py
aaron-siegel: Done

tests/test_table_model.py
aaron-siegel: Ok

tests/test_table_model.py
aaron-siegel: Done.

tests/test_table_model.py
aaron-siegel: I'd been assuming we'd want individual create() for models, but per your comments I've taken it out for now and made create() protected (_create()). We should discuss this as a product question though.

tests/test_table_model.py
aaron-siegel: I switched to create_all() in all the functional tests (tests that reproduce user flows).

For the `errors` tests though, it's really a lot easier to use individual `create()` (now `_create()`) to cleanly trap the error generation. I think this is a harmless use of unit/whitebox testing.

pixeltable/catalog/catalog.py
mkornacker: As discussed, let's just keep things uniform.

pixeltable/catalog/catalog.py
```diff
@@ -1512,69 +1545,275 @@ def create_view(
...
+            self._clear_tv_cache(base_tvp.tbl_version.key)
+        self._roll_forward()
+
+        @retry_loop(read_tbl_ids=[tbl_id])
+        def _get_tbl() -> LocalTable:
+            return self.get_table_by_id(tbl_id)
+
+        return _get_tbl(), True
+
+    def _validate_model_against_existing(
```
mkornacker: Chatty name, and the fact that the table already exists is caller semantics and doesn't affect the logic here. 
_validate_model? _validate_table?

pixeltable/catalog/catalog.py
```diff
@@ -1512,69 +1545,275 @@ def create_view(
...
+                    media_validation=media_validation,
+                    create_default_idxs=create_default_idxs,
+                    is_versioned=True,
+                    tbl_id=tbl_id,
+                )
+            )
+            base_tvp: TableVersionPath | None = None
+        else:
+            base_tvp = base._first_tbl
+            create_fn = retry_loop(for_write=True)(
```
mkornacker: Same retry_loop comment, let's just stick with the existing conventions.

pixeltable/catalog/view.py
```diff
@@ -125,52 +168,20 @@ def _create(
             # create a copy that we can modify and store
             sample_clause = dataclasses.replace(sample_clause, stratify_exprs=copy.copy(sample_clause.stratify_exprs))
 
-        # same for value exprs
-        for col in columns:
-            if not col.is_computed:
-                continue
+        # verify that computed column expressions can be evaluated in the context of the base
+        # If a tbl_id is provided, then we also include sibling columns that are antecedent in the given column
```
mkornacker: Why only if a tbl_id is provided? It sounds like that's something you should always do (we just didn't have a way to construct this before).

pixeltable/exprs/expr.py
```diff
@@ -282,10 +282,15 @@ def resolve_computed_cols(self, resolve_cols: set[catalog.Column] | None = None)
                 return result
             result = result.substitute({ref: ref.col.value_expr for ref in target_col_refs})
 
-    def is_bound_by(self, tbls: list[catalog.TablePath]) -> bool:
+    def is_bound_by(self, tbls: list[catalog.TablePath], siblings: list[catalog.Column] | None = None) -> bool:
```
mkornacker, line 285: siblings -> cols (the method doesn't need to know the caller semantics).

Also, this needs to use ColumnVersionMd (the method only needs to reference metadata).

tests/test_table_model.py
mkornacker: I really think this would substantially improve the tests. As it is, eyeballing the full metadata dump for correctness isn't a good way to understand whether the test itself is correct.

This would require a test utility assert_model_metadata_eq(model_tbl, tbl) to ignore schema versions.

tests/test_table_model.py
```diff
@@ -461,165 +473,178 @@ class ExampleSubviewModelFromQuery(
...
-            else:
-                _ = ExampleTableModel.create(root)
-                _ = ExampleViewModel.create(root)
-                _ = ExampleSubviewModel.create(root)
-                _ = ExampleViewModelFromQuery.create(root)
-                _ = ExampleSubviewModelFromQuery.create(root)
-
-    def test_view_model_with_iterator(self, uses_db: None) -> None:
-        class ExampleTableModel(pxt.TableModel, name='test_table'):
+            TableModel.create_all(p(root))
```
mkornacker, line 444: Missing: validation of the schema of the created tables/views.

pixeltable/catalog/model.py
```diff
@@ -0,0 +1,742 @@
...
+    )
+)
+
+# Sanity check to guard against drift in the SDK surface.
+for method in FORWARDED_TABLE_METHODS:
+    assert hasattr(Table, method), method
+
+
+@dataclass(frozen=True)
+class Column:
```
mkornacker, line 60: We already have a catalog.Column, can we call this something else like ModelColumn?

pixeltable/catalog/model.py
```diff
@@ -0,0 +1,742 @@
...
+            column_spec['destination'] = self.destination
+        if self.custom_metadata is not None:
+            column_spec['custom_metadata'] = self.custom_metadata
+        if self.comment is not None:
+            column_spec['comment'] = self.comment
+        return column_spec
+
+
+@dataclass(frozen=True)
+class EmbeddingIndex:
```
mkornacker, line 94: ModelEmbeddingIndex, to avoid confusion with index.EmbeddingIndex?

pixeltable/catalog/model.py
```diff
@@ -0,0 +1,742 @@
...
+
+def _col_type_from_spec(column_spec: ColumnSpec) -> ts.ColumnType:
+    """The ColumnType that a column defined by `column_spec` will have."""
+    if 'type' in column_spec:
+        return ts.ColumnType.normalize_type(column_spec['type'], nullable_default=True, allow_builtin_types=False)
+    assert 'value' in column_spec
+    return column_spec['value'].col_type
+
+
+class _PlaceholderColumnRef(exprs.Expr):
```
mkornacker: Awkward name. ModelColumnRef/ModelQuery (or are they used outside of that context)? Template... (that's effectively what models are)? And why the _ prefix, it's used in Catalog.

pixeltable/catalog/model.py
```diff
@@ -0,0 +1,742 @@
...
+            )
+        stratify_exprs: list[exprs.Expr] = []
+        if stratify_by is not None:
+            if isinstance(stratify_by, exprs.Expr):
+                stratify_by = [stratify_by]
+            stratify_exprs = list(stratify_by)
+        sample_clause = SampleClause(None, n, n_per_stratum, fraction, seed, stratify_exprs)
+        return dataclasses.replace(self, sample_clause=sample_clause)
+
+    def _bind(self, binding_root: str) -> 'pxt.Query':
```
mkornacker, line 279: Why _bind and not bind?

mkornacker (approved):

tests/test_table_model.py
aaron-siegel: Done

pixeltable/catalog/catalog.py
```diff
@@ -1512,69 +1545,275 @@ def create_view(
...
+                iterator.it, subst_args, subst_kwargs, subst_bound_args, iterator.outputs, iterator.validation_error
+            )
+            # Build substitutions for the iterator's output columns.
+            for name, output in iterator.outputs.items():
+                catalog_col = Column.create(name, {'type': output.col_type, 'stored': output.is_stored})  # type: ignore[arg-type]
+                catalog_col.tbl_handle = tbl_handle
+                catalog_col.id = next(next_col_id)
+                subst_dict[_PlaceholderColumnRef(name)] = exprs.ColumnRef(
+                    catalog_col.column_version_md(), perform_validation=(media_validation == 'on_read')
+                )
```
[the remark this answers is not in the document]
aaron-siegel: Great catch. I created a test to trap this error and fixed it.

pixeltable/catalog/catalog.py
aaron-siegel: Changed to _validate_model()

pixeltable/catalog/catalog.py
aaron-siegel: Ok

pixeltable/catalog/view.py
aaron-siegel: Added a TODO for this.

pixeltable/exprs/expr.py
aaron-siegel: Changed to `antecedent_cols`

tests/test_table_model.py
aaron-siegel: I've upgraded all tests in test_table_model.py to create tables using both a model and a traditional create_table()/create_view(), and verify that both the schemas and the insertion behavior line up.

pixeltable/catalog/model.py
aaron-siegel: It's user-facing, so I'd prefer to keep it concise. ModelColumn would be awkward for the user, and it'd really be putting the cart before the horse to make decisions about our SDK surface based on preexisting internal names.

I don't think it should cause any problems, since only `model.Column` is user-facing. If we really want to disambiguate, I'd argue changing `catalog.Column` to `catalog.CatalogColumn` is the way to go.

pixeltable/catalog/model.py
aaron-siegel: Exact same comment as above

pixeltable/catalog/model.py
aaron-siegel: On this one I don't really care, so ModelColumnRef/ModelQuery it is.

pixeltable/catalog/model.py
aaron-siegel: Same reason as _create() - to hide it from the user and make bind_all() the preferred way of doing things. Or did you want to keep individual MyModel.bind() calls, just not MyModel.create()? It seems like we'd want SDK parity there.

PR #1437, opened 2026-06-30
Proxy tables: FastAPIRouter and import_<>()

mkornacker:


pixeltable/catalog/globals.py
```diff
@@ -250,3 +255,13 @@ def normalize_schema(schema: Mapping[str, type | ColumnSpec | exprs.Expr]) -> di
             )
         result[name] = cast(ColumnSpec, col_spec)
     return result
+
+
+def is_hf_dataset(source: 'TableDataSource' | None) -> bool:
```
aaron-siegel, line 260: I question whether globals is the right place for this. If it goes in globals, it should be `_is_hf_dataset()`?

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -60,27 +75,76 @@ def insert(
...
+                # Normalize the overrides to ColumnType (schema inference requires instances, not type markers).
+                data_source = TableDataConduit.create(
+                    source,
+                    source_format=source_format,
+                    src_schema_overrides=self._normalize_schema_overrides(schema_overrides),
+                    extra_fields=kwargs,
+                )
+                data_source.src_pk = []
+                data_source.infer_schema()
+                rows = self._convert_local_paths([row for batch in data_source.valid_row_batch() for row in batch])
```
aaron-siegel: spurious list creation

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -60,27 +75,76 @@ def insert(
...
+                schema_overrides=schema_overrides,
+                extra_fields=kwargs,
+                on_error=on_error,
+                print_stats=print_stats,
+                return_rows=return_rows,
+            )
+
+        # a HuggingFace dataset is sent to the daemon in its own on-disk form and reconstructed there, rather
+        # than materialized into rows on the client
+        if is_hf_dataset(source):
```
aaron-siegel, line 109: So this is the reason for `is_hf_dataset()`. Hmm. I wonder if something like `is_direct_proxy()` would be more appropriate, since that really captures the intent. (It just so happens that we're only doing that with HF datasets currently, but that might not always be the case?)

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -60,27 +75,76 @@ def insert(
...
-        # a Table or Query source runs on the server against the same hosted catalog
         if isinstance(data_source, QueryTableDataConduit):
             return self._insert_query(
                 data_source.pxt_query, on_error=on_error, print_stats=print_stats, return_rows=return_rows
             )
-        # dict/pydantic rows are shipped to the daemon, which validates and inserts them through the same conduit
+
+        # in-memory data sources are sent as row dicts
         if isinstance(data_source, PydanticTableDataConduit):
             rows = self._pydantic_to_rows(cast('list[Any]', source))
```
aaron-siegel, line 131: `'list[Any]'` and `list` have the identical meaning here, I believe.

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -112,18 +176,245 @@ def _insert_query(
...
+                'connect_url': engine.url.render_as_string(hide_password=False),
+                'sql_text': sql_text,
+                'col_names': list(sql_source.col_names),
+                'on_error': on_error,
+                'print_stats': print_stats,
+                'return_rows': return_rows,
+            },
+        )
+
+    def _local_path(self, val: str) -> str | None:
```
aaron-siegel: Don't we already have this in `utils.parse_local_file_path()`? Nearly verbatim?

Is this an example of Claude writing a bunch of code without checking if it has already been written? :)

pixeltable/catalog/table_proxy.py
```diff
@@ -166,6 +166,9 @@ def list_views(self, *, recursive: bool = True) -> list[str]:
     def columns(self) -> list[str]:
         return [col_md.name for col_md in self._tbl_path.column_md() if col_md.name is not None]
 
+    def _get_schema(self) -> dict[str, 'ts.ColumnType']:
```
aaron-siegel: Should this be defined on `Table`? Seems like it doesn't depend on anything specific to `TableProxy`.

pixeltable/io/sql.py
```diff
@@ -189,13 +199,13 @@ def run(connection: sql.Connection) -> pxt.Table:
 
         existing = pxt.get_table(tbl_name, if_not_exists='ignore')
         if if_exists == 'append' and existing is not None:
-            if not isinstance(existing, pxt.InsertableTable):
+            if existing._display_name() != 'table':
```
aaron-siegel: Is this really the best way to do this check? It seems fragile.

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -112,18 +176,245 @@ def _insert_query(
...
+        from pixeltable import type_system as ts
+
+        if schema_overrides is None:
+            return None
+        return {
+            name: t if isinstance(t, ts.ColumnType) else ts.ColumnType.normalize_type(t, nullable_default=True)
+            for name, t in schema_overrides.items()
+        }
+
+    def _convert_local_paths(self, rows: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
aaron-siegel: No reason not to take an Iterable as input here and return a generator

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -112,18 +176,245 @@ def _insert_query(
...
+          and re-instantiate it on the daemon
+        - datasets.save_to_disk() writes a self-contained Arrow copy (media bytes embedded, full feature metadata)
+        - this avoids materialization of the dataset in Python
+        """
+        from pixeltable.service.proxy_protocol import encode_dir_tree
+
+        # a streaming dataset (IterableDataset/IterableDatasetDict) has no on-disk serialization; supporting it
+        # would mean reading the entire stream into client memory, which is exactly what sending the on-disk
+        # form avoids
+        dataset = cast(Any, source)
```
aaron-siegel, line 340: ? what's the point of this?

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -112,18 +176,245 @@ def _insert_query(
...
+        from pixeltable.service.proxy_protocol import LocalFile
+
+        media_cols = self._media_column_names()
+        if len(media_cols) == 0:
+            return rows
+
+        converted: list[dict[str, Any]] = []
+        for row in rows:
+            new_row = dict(row)
+            for name in media_cols & new_row.keys():
```
aaron-siegel: Is it ever possible for the row to have an incomplete list of keys?

pixeltable/catalog/insertable_table_proxy.py
```diff
@@ -112,18 +176,245 @@ def _insert_query(
...
+        """Convert local file path values of media columns to LocalFile for correct request serialization."""
+        from pixeltable.service.proxy_protocol import LocalFile
+
+        media_cols = self._media_column_names()
+        if len(media_cols) == 0:
+            return rows
+
+        converted: list[dict[str, Any]] = []
+        for row in rows:
+            new_row = dict(row)
```
aaron-siegel: Perhaps this should be copied lazily to avoid unnecessary memcopies

e.g. change this line to `new_row = row`, and under `if p is not None`, add something like
```python
if new_row is row:
    new_row = dict(row)
```

something like that?

pixeltable/service/proxy_dispatch.py
```diff
@@ -40,58 +52,132 @@ def handle(request_json: str) -> str:
...
+        # best-effort removal of this request's uploaded temp files; missing_ok covers a file that a handler moved
+        # into a reassembled directory (those directories are removed by the handler that creates them)
+        for temp_path in request._uploaded_names:
+            try:
+                pathlib.Path(temp_path).unlink(missing_ok=True)
+            except OSError:
+                pass
+
+
+def _result_local_path(val: str) -> pathlib.Path | None:
```
aaron-siegel: Yet another copy of this same logic? Why?

pixeltable/service/proxy_dispatch.py
```diff
@@ -359,6 +516,27 @@ def build() -> Query:
     return build().count()
 
 
+def _encode_row_media(rows: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
aaron-siegel, line 508: Should we be using list[Row] in these various places where we're using list[dict]?

pixeltable/serving/_fastapi.py
```diff
@@ -1674,7 +1687,7 @@ def _validate_dml_args(
         background: bool,
         error_prefix: str,
         route_type: Literal['insert', 'update', 'compute'],
-    ) -> tuple[list[str], list[str], list[str], dict[str, catalog.Column]]:
+    ) -> tuple[list[str], list[str], list[str], dict[str, catalog.ColumnVersionMd]]:
```
aaron-siegel: I know it isn't a new issue, but this would really benefit from a NamedTuple

tests/test_table_model.py
```diff
@@ -48,7 +48,7 @@ class ExampleTableModel(TableModel, name='test_table'):
                 comment='This is a column with special properties',
             )
             computed_with_special_props = Column(value=(value / 3), stored=False)
-            computed_with_special_props_2 = Column(value=img.rotate(90), destination='.')
```
aaron-siegel, line 51: We should find some way to test the destination parameter here (just removing it weakens our test coverage)

aaron-siegel (changes requested):

pixeltable/catalog/insertable_table_proxy.py
mkornacker: I don't follow. What is is_direct_proxy()? An HF dataset needs custom serialization.

tests/test_table_model.py
mkornacker: It can't be a local path, and an s3 uri will make the test fail randomly.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Looks that way.

pixeltable/catalog/globals.py
mkornacker: Why? It's not local to this file, and globals.py contains a collection of utility functions.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Done.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: What's wrong with list[Any]? We use that elsewhere.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Why return a generator? That has no benefit, there is no streaming use case here.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: What do you mean by keys? Primary key column values?

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Done.

pixeltable/catalog/table_proxy.py
mkornacker: Done.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Avoiding subsequent mypy errors.

pixeltable/io/sql.py
mkornacker: Done.

pixeltable/serving/_fastapi.py
mkornacker: Done.

pixeltable/service/proxy_dispatch.py
mkornacker: Not a bad idea, but we'd need to do a larger redesign and introduce a RowBatch abstraction (you still need access to the column names), and it's not clear this should be tied to query results. I'd prefer to do that in a separate PR.

pixeltable/service/proxy_dispatch.py
mkornacker: Done.

pixeltable/catalog/insertable_table_proxy.py
aaron-siegel: I meant: Why do we need to say
```python
for name in media_cols & new_row.keys():
```
rather than
```python
for name in media_cols:
```
The `& new_row.keys()` is a defensive clause, but against what exactly?

pixeltable/catalog/globals.py
aaron-siegel: Oh sorry I missed that it was `catalog/globals` on the first pass.

pixeltable/catalog/insertable_table_proxy.py
aaron-siegel: Maybe that's the wrong name for it but my point is, it's not the fact that it's a HF dataset specifically that needs special treatment, it's the fact that it needs custom serialization. How about `needs_custom_serialization()` or some such?

pixeltable/catalog/insertable_table_proxy.py
aaron-siegel: Ok but it should certainly take an Iterable. There's nothing about the input that requires it to be a list.

pixeltable/service/proxy_dispatch.py
aaron-siegel: Confused as to what "Done" means; the function is still there.

aaron-siegel (approved):

pixeltable/catalog/insertable_table_proxy.py
mkornacker: It needs to know it's an HF dataset so it can call _insert_hf_dataset(). There is no generic "custom serialization".

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Non-required columns need not be present.

pixeltable/catalog/insertable_table_proxy.py
mkornacker: Happy to do that, when we have a use case for it. As it stands, the first thing would be to convert it to a list.

pixeltable/service/proxy_dispatch.py
mkornacker: Done.

PR #1439, opened 2026-07-06
PXT-1225/1229: Thread safety

mkornacker:
Thread- and multi-process-concurrency fixes for FileCache.

Thread-safety for model caches and Env.


pixeltable/utils/file_lock.py
```diff
@@ -0,0 +1,64 @@
...
+from pathlib import Path
+from typing import Iterator
+
+if sys.platform == 'win32':
+    import msvcrt
+else:
+    import fcntl
+
+
+class FileLock:
```
aaron-siegel: Explain reasons not to use `fasteners` package? It's already in our minimal install and does more or less the same thing, as I understand it.

https://pypi.org/project/fasteners/

aaron-siegel (changes requested):

pixeltable/utils/file_lock.py
mkornacker: It's not a good idea to add dependencies to replace 64 lines of easy-to-follow code.

I'm not seeing fasteners in our core dependencies, but if it were there, it would be good to get rid of.

aaron-siegel: We're not adding dependencies. It's already there transitively.

And it's not just about replacing 64 lines of code, it's about using battle tested libraries that may address edge cases that our code doesn't (this type of logic is susceptible to obscure platform-specific edge cases, for example).

mkornacker: Done.

pixeltable/exec/expr_eval/expr_eval_node.py
```diff
@@ -524,7 +524,13 @@ def _done_cb(self, t: asyncio.Task) -> None:
...
-            # TODO: find a better error class
-            self.error = excs.Error(excs.ErrorCode.GENERIC_USER_ERROR, f'Exception in task: {exc}\n{stack_trace}')
+            if isinstance(exc, excs.Error):
+                # keep the specific error so its error_code survives; stash the traceback as detail for diagnostics
+                if exc.detail is None:
+                    exc.detail = traceback.format_exc()
+                self.error = exc
+            else:
+                self.error = excs.Error(
+                    excs.ErrorCode.GENERIC_USER_ERROR, f'Exception in task: {exc}\n{traceback.format_exc()}'
```
sergey-mkhitaryan: why not attach `exc` as cause?

pixeltable/utils/filecache.py
```diff
@@ -104,183 +138,305 @@ def init(cls) -> None:
...
+        entries: list[CacheEntry] = []
+        for path_str in glob.glob(str(Env.get().file_cache_dir / '*')):
+            path = Path(path_str)
+            if _CACHE_ENTRY_FILE_RE.fullmatch(path.name) is None:
+                # not a cache entry file
+                continue
+            try:
+                entries.append(CacheEntry.from_file(path))
+            except (ValueError, OSError):
+                # the file matched the name pattern but vanished between the glob and the stat (another process
```
sergey-mkhitaryan, line 171: if a file is unreadable, perhaps we should delete it

pixeltable/utils/filecache.py
```diff
@@ -104,183 +138,305 @@ def init(cls) -> None:
...
+                _logger.debug('clearing all entries from file cache')
+                failures: list[str] = []
+                for path_str in glob.glob(str(Env.get().file_cache_dir / '*')):
+                    if _CACHE_ENTRY_FILE_RE.fullmatch(os.path.basename(path_str)) is None:
+                        continue
+                    try:
+                        os.remove(path_str)
+                    except FileNotFoundError:
+                        pass  # already removed by another process
+                    except OSError as exc:
```
sergey-mkhitaryan: Claude tells me that a `PermissionError` will be raised on Windows if the file to be removed is open by another process. Does this break anything?

pixeltable/utils/filecache.py
```diff
@@ -104,183 +138,305 @@ def init(cls) -> None:
...
+                # remove every cache file on disk, not just the entries this instance tracks: another FileCache
+                # instance or process may have written files we never adopted. The lock file and any foreign
+                # files do not match the entry-name pattern and are left in place.
+                _logger.debug('clearing all entries from file cache')
+                failures: list[str] = []
+                for path_str in glob.glob(str(Env.get().file_cache_dir / '*')):
+                    if _CACHE_ENTRY_FILE_RE.fullmatch(os.path.basename(path_str)) is None:
+                        continue
+                    try:
+                        os.remove(path_str)
```
sergey-mkhitaryan: why not call `self._remove_file()` and drop the first except clause

pixeltable/utils/filecache.py
```diff
@@ -104,183 +138,305 @@ def init(cls) -> None:
...
-                except OSError as exc:
-                    name = os.path.basename(path)
-                    origin = 'tracked' if name in tracked_names else 'untracked'
-                    failures.append(f'{name} ({origin}): {exc}')
-            self.cache.clear()
-            self.total_size = 0
-            self.num_requests, self.num_hits, self.num_evictions = 0, 0, 0
-            self.keys_retrieved.clear()
-            self.keys_evicted_after_retrieval.clear()
+        with self._lock:
```
sergey-mkhitaryan, line 201: should this also acquire `_evict_lock`?

pixeltable/utils/filecache.py
```diff
@@ -104,183 +138,305 @@ def init(cls) -> None:
...
+                    raise RuntimeError(
+                        f'FileCache.clear(): could not remove {len(failures)} file(s) from '
+                        f'{Env.get().file_cache_dir}:\n  {detail}'
+                    )
+                return
+
+            entries_to_remove = [e for e in self.cache.values() if e.tbl_id == tbl_id]
+            _logger.debug(f'clearing {self.num_files(tbl_id)} entries from file cache for table {tbl_id}')
+            for entry in entries_to_remove:
+                self._remove_file(entry.path)
```
sergey-mkhitaryan: there is a couple of inconsistencies between the `tbl_id is None` branch and this:
* the former will accumulate failures and raise only after attempting to remove all files. the latter fails fast.
* the former lists files on disk and goes over them, the latter purely from the state in memory.

these differences don't feel intentional, are they? should these two branches both call to a helper function that removes the provided list of files?

pixeltable/utils/filecache.py
```diff
@@ -104,183 +138,305 @@ def init(cls) -> None:
...
+                    self.cache[key].size for key in self.evicted_working_set_keys if key in self.cache
+                )
+                suggested_cache_size = self.capacity_bytes + extra_capacity_needed + (1 << 30)
+                warnings.warn(
+                    f'{len(self.evicted_working_set_keys)} media file(s) had to be downloaded multiple times '
+                    'this session, because they were evicted\nfrom the file cache after their first access. The '
+                    f'total size of the evicted file(s) is {round(extra_capacity_needed / (1 << 30), 1)} GiB.\n'
+                    f'Consider increasing the cache size to at least {round(suggested_cache_size / (1 << 30), 1)} '
+                    f'GiB (it is currently {round(self.capacity_bytes / (1 << 30), 1)} GiB).\n'
+                    f'You can do this by setting the value of `file_cache_size_g` in: {Config.get().config_file}',
```
sergey-mkhitaryan, line 252: should we also suggest adjusting `file_cache_lease_s` as an alternative?

scripts/filecache_multiprocess_stress.py
```diff
@@ -0,0 +1,216 @@
...
+            return True
+        nxt = cur.__cause__ or cur.__context__
+        cur = nxt if isinstance(nxt, Exception) else None
+    return False
+
+
+def _worker(worker_id: int, port: int, n_images: int, rows: int, rounds: int, n_threads: int) -> None:
+    """One process: runs n_threads threads, each driving its own table through the insert/query fetch path.
+
+    Exit code: 1 if any thread hit a non-capacity error (a real failure); 2 if the only errors were the
```
sergey-mkhitaryan, line 98: exit codes 1 and 2 are reserved by `python`. apparently [3-63] are for custom use like this.

scripts/filecache_multiprocess_stress.py
```diff
@@ -0,0 +1,216 @@
...
+        ]
+
+        def run(thread_id: int) -> None:
+            try:
+                capacity_hits[thread_id] = _thread_body(
+                    tbls[thread_id], worker_id, thread_id, port, n_images, rows, rounds
+                )
+            except Exception as exc:
+                traceback.print_exc()
+                fatal.append(exc)
```
sergey-mkhitaryan: `fatal` and `capacity_hits` need to be guarded by a lock

scripts/filecache_multiprocess_stress.py
```diff
@@ -0,0 +1,216 @@
...
+    """One process: runs n_threads threads, each driving its own table through the insert/query fetch path.
+
+    Exit code: 1 if any thread hit a non-capacity error (a real failure); 2 if the only errors were the
+    acceptable 'out of capacity' contention signal; 0 if every thread completed cleanly.
+    """
+    fatal: list[Exception] = []
+    # per-thread slots: each thread writes only its own index, so no cross-thread synchronization is needed
+    capacity_hits = [0] * n_threads
+    try:
+        tbls = [
```
sergey-mkhitaryan, line 106: why not have everyone hit the same table? if every thread of every process gets its own table, they will never contend on the same cache file, which seems like a big gap in coverage.

sergey-mkhitaryan (changes requested):

pixeltable/exec/expr_eval/expr_eval_node.py
aaron-siegel: Agreed - this would be preferable to appending the stack trace as a string

aaron-siegel (approved):

scripts/filecache_multiprocess_stress.py
mkornacker: That's the point: the contention is not on cache files, it's on cache capacity.

Also: FileCache hashes on the url, not on tbl id/col id. In this test, the tables all share the same urls.

pixeltable/utils/filecache.py
mkornacker: No. _evict_lock is needed for concurrent lookups. clear() removes all cache files and doesn't care about lookups - that is solved by a different part of the runtime system (clear() is called in response to dropping a table).

pixeltable/utils/filecache.py
mkornacker: That's fine. clear(None) is just a test utility. Clarified in the comment.

pixeltable/utils/filecache.py
mkornacker: This is a different scenario (test utility).

pixeltable/utils/filecache.py
mkornacker: Done.

pixeltable/utils/filecache.py
mkornacker: No. Exposing more knobs isn't a good idea. If this becomes a problem, the right thing to do is to increase capacity.

scripts/filecache_multiprocess_stress.py
mkornacker: Done.

pixeltable/utils/filecache.py
mkornacker: Clarified in comment.

pixeltable/exec/expr_eval/expr_eval_node.py
mkornacker: Done.

scripts/filecache_multiprocess_stress.py
mkornacker: This script controls how _worker is invoked, so it's not a problem.

sergey-mkhitaryan (approved):

PR #1441, opened 2026-07-07
PXT-1220: keep pickled functions out of computed columns

mkornacker:
…kled) functions


pixeltable/exprs/expr.py
```diff
@@ -422,6 +422,18 @@ def contains_(self, cls: type[Expr] | None = None, filter: Callable[[Expr], bool
...
+    def validate_storable(self, context: str) -> None:
+        """Raise if this expr references a Function that cannot be persisted."""
+        from .function_call import FunctionCall
+
+        for fn_call in self.subexprs(FunctionCall):
+            if not fn_call.fn.is_storable:
+                raise excs.RequestError(
+                    excs.ErrorCode.UNSUPPORTED_OPERATION,
+                    f'{context} uses `{fn_call.fn.display_name}()`, which was created with `.apply()` or defined '
+                    f'as a local function. Define it as a module-level `@pxt.udf` and use that instead.',
```
aaron-siegel: We should consider adding a URL that links to the relevant doc pages here.

pixeltable/func/function.py
```diff
@@ -576,6 +581,12 @@ def __init__(self, self_path: str, fn_dict: dict[str, Any], error_msg: str):
         self.fn_dict = fn_dict
         self.error_msg = error_msg
 
+    @property
+    def is_storable(self) -> bool:
+        # _as_dict() re-serializes fn_dict verbatim, so a pickle-backed original (an inline 'binary' body or an
+        # 'id' reference into the legacy functions table) would be persisted again and must be rejected
```
aaron-siegel: Are we sure about this? The idea behind `InvalidFunction` is that it won't block catalog manipulations; it simply keeps invalid functions in their "invalid" state until the user does something about them, but permits re-serialization.

I'd favor this behavior here. Essentially, we'd be "quarantining" the legacy pickled function so it can't be executed, but in a way that doesn't force the user to take action immediately (catalog stays usable for any operations that don't depend on the quarantined function)

aaron-siegel (changes requested):
I'm not sure I agree with this approach - I think it burdens the user unnecessarily. Let's discuss in person.

pixeltable/func/mcp.py
```diff
@@ -34,24 +35,153 @@ async def mcp_udfs_async(url: str) -> list['pxt.func.Function']:
...
+        # over time, so the interface is part of the identity; _tool_interface is name-aware (MCP calls by name).
+        return (
+            isinstance(other, McpFunction)
+            and self.url == other.url
+            and self.tool_name == other.tool_name
+            and _tool_interface(self.signature) == _tool_interface(other.signature)
+        )
+
+    def __hash__(self) -> int:
+        # coarser than __eq__ (the interface is not hashable), which is a valid hash: equal instances share a bucket
```
aaron-siegel: Is this a claude generated comment? :)

seems like something any cs undergrad would already know... "let me just copy/paste the wikipedia entry on hashtables into the user's codebase because why not?"

pixeltable/func/mcp.py
```diff
@@ -34,24 +35,153 @@ async def mcp_udfs_async(url: str) -> list['pxt.func.Function']:
 
 
 def mcp_tool_to_udf(url: str, mcp_tool: 'mcp.types.Tool') -> 'pxt.func.Function':
-    Env.get().require_package('mcp')
-    import mcp
-    from mcp.client.streamable_http import streamablehttp_client
+    return McpFunction(url, mcp_tool.name, [_signature_from_tool(mcp_tool)], mcp_tool.name, mcp_tool.description)
```
aaron-siegel, line 39: Don't we still need the require_package('mcp')?

pixeltable/metadata/converters/convert_54.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # embedding-index init_args, and view predicates / iterator args, live in Table.md
+    for tbl_id, table_md in tbl_mds.items():
+        for idx_md in (table_md.get('index_md') or {}).values():
+            if _contains_pickled_fn(idx_md):
+                offending.add(f'{tbl_names[tbl_id]}: embedding index {idx_md.get("name")!r}')
+        if _contains_pickled_fn(table_md.get('view_md')):
+            offending.add(f'{tbl_names[tbl_id]}: view definition')
+
+    if len(offending) > 0:
```
aaron-siegel: per our in-person conversation, I think we can load the pickled UDFs as InvalidFunction instances rather than failing the migration

tests/test_function.py
```diff
@@ -84,6 +104,16 @@ def test_serialize_anonymous(self, init_env: None) -> None:
         assert isinstance(td['templates'], list) and len(td['templates']) == 1
         assert isinstance(Function.from_dict(td), func.ExprTemplateFunction)
 
+    def test_invalid_function_is_storable(self) -> None:
```
aaron-siegel: I'm confused about this. Why would an InvalidFunction ever need to be storable?

pixeltable/exprs/expr.py
mkornacker: Done.

pixeltable/func/function.py
mkornacker: Done.

tests/test_function.py
mkornacker: Done.

pixeltable/metadata/converters/convert_54.py
mkornacker: Done.

pixeltable/func/mcp.py
mkornacker: Done.

pixeltable/func/mcp.py
mkornacker: Done.

pixeltable/func/mcp.py
```diff
@@ -35,23 +36,154 @@ async def mcp_udfs_async(url: str) -> list['pxt.func.Function']:
...
+        # over time, so the interface is part of the identity; _tool_interface is name-aware (MCP calls by name).
+        return (
+            isinstance(other, McpFunction)
+            and self.url == other.url
+            and self.tool_name == other.tool_name
+            and _tool_interface(self.signature) == _tool_interface(other.signature)
+        )
+
+    def __hash__(self) -> int:
+        # we leave the signature out of this, because it's no immediately hashable (and collisions here have no perf
```
aaron-siegel: s/no/not/

tests/test_migration.py
```diff
@@ -95,6 +96,10 @@ def test_convert_53_normalize(self) -> None:
     @rerun(reruns=3, reruns_delay=8)  # Deal with occasional concurrency issues
     @pytest.mark.skipif(platform.system() == 'Windows', reason='Does not run on Windows')
     @pytest.mark.skipif(sys.version_info >= (3, 11), reason='Runs only on Python 3.10 (due to pickling issue)')
+    # Dumps at or below version 54 contain computed columns backed by pickled functions (apply()/stored UDFs).
+    # Migrating them past 54 leaves those references in place, so the columns load as InvalidFunction and warn;
+    # that is expected here and must not fail the test.
+    @pytest.mark.filterwarnings('ignore::pixeltable.exceptions.PixeltableWarning')
```
aaron-siegel: Hmm. A blanket `ignore` on all PixeltableWarnings seems exceedingly broad here.

Much better would be:
- if `version < 55`, *expect* the warning, and filter on just the specific warning text being caught (so we don't inadvertently suppress other warnings)
- if `version >= 55`, don't expect a warning

tests/test_migration.py
```diff
@@ -95,6 +96,10 @@ def test_convert_53_normalize(self) -> None:
     @rerun(reruns=3, reruns_delay=8)  # Deal with occasional concurrency issues
     @pytest.mark.skipif(platform.system() == 'Windows', reason='Does not run on Windows')
     @pytest.mark.skipif(sys.version_info >= (3, 11), reason='Runs only on Python 3.10 (due to pickling issue)')
```
aaron-siegel: We can get rid of this `skipif` now - hooray!!

aaron-siegel (approved):

pixeltable/func/mcp.py
mkornacker: Done.

tests/test_migration.py
mkornacker: Done.

tests/test_migration.py
mkornacker: Done.

PR #1442, opened 2026-07-07
Class-based declarator fixes

aaron-siegel:
Several upgrades to class-based declarators:

- Embedding indices are now folded into the same transaction umbrella as table creation
- Proper substitution of column references to explicit base query select lists
- Light refactoring to unify a few code paths and reduce duplication


pixeltable/catalog/catalog.py
```diff
@@ -1499,37 +1484,57 @@ def _create_table(
...
         create_default_idxs: bool,
         is_versioned: bool,
-        tbl_id: UUID | None = None,
-    ) -> tuple[UUID, bool]:
+        additional_idxs: list[tuple[Column, str | None, 'index.IndexBase']] | None = None,
+        explicit_tbl_id: UUID | None = None,
+    ) -> tuple[LocalTable, bool]:
         import pixeltable.metadata.schema
 
+        assert explicit_tbl_id is None or if_exists == IfExistsParam.ERROR
```
mkornacker: Comment on reasoning.

pixeltable/catalog/catalog.py
```diff
@@ -1499,37 +1484,57 @@ def _create_table(
         media_validation: MediaValidation,
         create_default_idxs: bool,
         is_versioned: bool,
-        tbl_id: UUID | None = None,
-    ) -> tuple[UUID, bool]:
+        additional_idxs: list[tuple[Column, str | None, 'index.IndexBase']] | None = None,
+        explicit_tbl_id: UUID | None = None,
```
mkornacker, line 1487: Isn't tbl_id explicit enough?

pixeltable/catalog/catalog.py
```diff
@@ -1667,50 +1682,93 @@ def create_from_model(
...
+                    col_name: str | None
+                    if select_name is not None:
+                        # The select list has an explicit name for this expression as a kwarg; use it.
+                        col_name = select_name
+                    elif isinstance(expr, exprs.ColumnRef):
+                        # It's an unnamed column reference; use the name of the referenced column as a fallback.
+                        col_name = expr.column_md.name
+                    else:
+                        # It's a compound expression with no explicit name. A name will be assigned when the table
+                        # is created, but it's anonymous to the TableModel.
```
mkornacker: This looks like a questionable choice, I don't think it makes sense to auto-generate column names when you're creating a view (or any table). I didn't realize we're doing that.

Leave todo to clean up later?

pixeltable/catalog/catalog.py
```diff
@@ -1667,50 +1682,93 @@ def create_from_model(
...
-                    elif isinstance(expr, exprs.ColumnRef):
-                        subst_dict[ModelColumnRef(expr.column_md.name)] = expr
-
         # We allocate the table id up front so that self-referential ColumnRefs (built below) point at it; since
         # this runs in the catalog that owns the table, no such reference ever needs to be serialized.
         tbl_id = uuid4()
         tbl_handle = TableVersionHandle(TableVersionKey(tbl_id, None))
 
-        initial_col_id = 0
+        # View columns always go in a specific order:
```
mkornacker: There is a fair amount of logic here that doesn't strictly require Catalog. In the interest of not having Catalog balloon too much, let's move that into LocalTable. Something like _prepare_create_from_model() that hands back a (named?) tuple with the missing args to _create_table/_view().

pixeltable/globals.py
```diff
@@ -242,6 +246,10 @@ def create_table(
         )
     )
 
+    if was_created:
```
mkornacker, line 249: Also add an 'else' console output?

mkornacker (changes requested):

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
aaron-siegel: ? I was trying to distinguish between clearly this and the result of `explicit_tbl_id or uuid4()`... happy to choose another name but I'd kind of prefer it not be `tbl_id`, for clarity.

pixeltable/catalog/catalog.py
aaron-siegel: Ok. I think we need a more involved design discussion around
- What types of queries are acceptable table bases
- How to handle column shadowing. We should have a very clearly articulated set of rules about how namespace collisions get resolved.

For now I'll put it in a TODO but this should be a topic for a future arch/freeform meeting.

pixeltable/catalog/catalog.py
aaron-siegel: That feels a little like spaghetti to me. (Right now LocalTable doesn't know anything about models, it's purely a lower-level abstraction. And we'd be going from `model.py` to `catalog` to `LocalTable` and then back to `catalog` again (for the final `_create_view()/_create_table()`. I'm not convinced this would make things any clearer? How strongly do you feel about it?

pixeltable/globals.py
aaron-siegel: Saying what - table already exists? (I added this, pls confirm it's what you had in mind.)

pixeltable/service/proxy_protocol.py
```diff
@@ -83,6 +84,10 @@ def serialize(obj: Any) -> Any:
         return {_TAG: 'Function', 'v': obj.as_dict()}
     if isinstance(obj, func.GeneratingFunctionCall):
         return {_TAG: 'GeneratingFunctionCall', 'v': obj.as_dict()}
+    if isinstance(obj, EmbeddingIndex):
+        # A declarative model's embedding-index spec (a dataclass of an Expr column ref + embedding Functions +
+        # scalars); serialize field-by-field so the nested Exprs/Functions round-trip via their own handlers.
+        return {_TAG: 'EmbeddingIndex', 'v': {f.name: serialize(getattr(obj, f.name)) for f in dataclasses.fields(obj)}}
```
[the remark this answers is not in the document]
aaron-siegel: False (the tests in `test_table_model.py` are proxied)

pixeltable/globals.py
mkornacker: I think that's useful output - I'd want to see that if I'm running a create_table().

mkornacker (approved):

PR #1445, opened 2026-07-07
[PXT-1230] Fix insert cold start caused by huggingface import

christopherpestano:
The first insert into a table paid a ~300ms cold start even when the source was a plain list of dicts. TableDataConduit.create() runs HFTableDataConduit.is_applicable() on every insert, and that check unconditionally imports the huggingface datasets library just to build the tuple of dataset classes for an isinstance test, so the import cost lands inside the first insert.

If the application never imported datasets, the source cannot be a huggingface dataset instance, so there is nothing to check. is_applicable() now consults sys.modules and only runs the isinstance check (and the import behind it) when datasets is already loaded in the process. Actual huggingface sources are unaffected, since constructing a Dataset requires the module to be loaded.

Measured with the OTEL spans from PXT-923: on a cold insert of a single dict row, the pixeltable.data_source.prepare span drops from ~350ms to ~2ms and the whole insert from ~356ms to ~8ms, with the datasets module no longer imported at all.

### DIFF
--- a/pixeltable/io/table_data_conduit.py
+++ b/pixeltable/io/table_data_conduit.py
@@ -421,7 +421,7 @@ class HFTableDataConduit(TableDataConduit):
         t = cls(**kwargs)
         import datasets
 
-        assert isinstance(tds.source, cls._get_dataset_classes())
+        assert cls._is_dataset_class(tds.source)
         if 'column_name_for_split' in t.extra_fields:
             t.column_name_for_split = t.extra_fields['column_name_for_split']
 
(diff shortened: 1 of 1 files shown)


pixeltable/io/table_data_conduit.py
```diff
@@ -452,8 +453,8 @@ def _get_dataset_classes(cls) -> tuple[type, ...]:
     @classmethod
     def is_applicable(cls, tds: TableDataConduit) -> bool:
         try:
-            return (isinstance(tds.source_format, str) and tds.source_format.lower() == 'huggingface') or isinstance(
-                tds.source, cls._get_dataset_classes()
+            return (isinstance(tds.source_format, str) and tds.source_format.lower() == 'huggingface') or (
+                'datasets' in sys.modules and isinstance(tds.source, cls._get_dataset_classes())
```
sergey-mkhitaryan: Notice that `_get_dataset_classes()` is used in two places, and both of them is an `isinstance()` check. It would be more elegant if you replaced `_get_dataset_classes()` with `_is_dataset_class(cls_to_check)`, and had a shortcut in it that returns False if `'datasets' not in sys.modules` to avoid an unnecessarily import.

sergey-mkhitaryan (changes requested):

pixeltable/io/table_data_conduit.py
```diff
@@ -452,8 +453,8 @@ def _get_dataset_classes(cls) -> tuple[type, ...]:
...
             )
```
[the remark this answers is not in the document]
christopherpestano: Already manually validated this works, this method doesn't really catch the timing difference

pixeltable/io/table_data_conduit.py
christopherpestano: Good point, my original thought was that the other reference to _get_dataset_classes() was fine since it was after the guard for is_applicable(), but it did mean there was kind of a split in the code logic. Implemented the change.

sergey-mkhitaryan (approved):

PR #1447, opened 2026-07-08
Re-enable inlined udfs w/in computed columns in notebooks

mkornacker:


pixeltable/func/udf.py
```diff
@@ -129,6 +130,13 @@ def make_function(
         function_path = None
     elif decorated_fn.__module__ != '__main__' and decorated_fn.__name__.isidentifier():
         function_path = f'{decorated_fn.__module__}.{decorated_fn.__qualname__}'
+
+    elif from_decorator and is_interactive_env() and decorated_fn.__name__.isidentifier():
+        # Give a udf inlined in a notebook cell a __main__.<name> path to enable computed columns with inlined udfs;
+        # resolves only within the session and degrades to InvalidFunction elsewhere
+        # TODO: remove and rework the offending notebooks
+        function_path = f'{decorated_fn.__module__}.{decorated_fn.__qualname__}'
```
aaron-siegel, line 138: This is the exact same statement as above, might be clearer to combine into a single `elif`? (I had to squint to convince myself they were identical)

aaron-siegel (approved):

pixeltable/func/udf.py
mkornacker: I did that on purpose, just so it can be removed very cleanly.

PR #1448, opened 2026-07-09
feat: Add Nebius Token Factory as LLM Provider Integration

amrrs:
Adds `pixeltable/functions/nebius.py` with `chat_completions` and `embeddings` UDFs for [Nebius Token Factory](https://tokenfactory.nebius.com/)'s OpenAI-compatible API, following the same pattern as `deepseek.py`/`openrouter.py`.

- `chat_completions`: chat completion API, with support for `tools`/`tool_choice`
- `embeddings`: embedding API for `Qwen/Qwen3-Embedding-8B` (the only embedding model Nebius currently serves), with:
  - static dimension resolution (`conditional_return_type`) so `add_embedding_index()` works without manual type hints
  - `model_kwargs` support for the `dimensions` param, since the model's native 4096-dim output exceeds pgvector's indexing limits (2000 for `fp32`, 4000 for `fp16`) — a truncated size (e.g. 1024) must be requested to make it indexable
- Registers the `nebius` config section (`api_key`, `rate_limit`) and wires `NEBIUS_API_KEY` into CI (`pytest.yml`)
- Tests in `tests/functions/test_nebius.py`, including a regression test that exercises `add_embedding_index()` + similarity search end-to-end

### Test plan

- [x] `ruff check` / `ruff format --check` pass
- [x] `pytest -m remote_api tests/functions/test_nebius.py -v` — 2 passed, run live against the Nebius API


pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,185 @@
...
+    import openai
+
+
+@env.register_client('nebius', credential_param='api_key')
+def _(api_key: str) -> 'openai.AsyncOpenAI':
+    import openai
+
+    return openai.AsyncOpenAI(
+        api_key=api_key,
+        base_url='https://api.tokenfactory.nebius.com/v1/',
```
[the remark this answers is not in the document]
amrrs: Fixed in 1eda94c6 — dropped the trailing slash from `base_url`.

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,185 @@
...
+        ... ]
+        >>> tbl.add_computed_column(
+        ...     response=chat_completions(
+        ...         messages, model='meta-llama/Llama-3.3-70B-Instruct'
+        ...     )
+        ... )
+    """
+    if model_kwargs is None:
+        model_kwargs = {}
+
```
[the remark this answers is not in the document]
amrrs: Fixed in 1eda94c6 — added `Env.get().require_package('openai')` before the client call, matching the `openrouter.py` pattern.

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,185 @@
...
+        ...     'text',
+        ...     embedding=embeddings.using(
+        ...         model='Qwen/Qwen3-Embedding-8B', model_kwargs={'dimensions': 1024}
+        ...     ),
+        ... )
+    """
+    if model_kwargs is None:
+        model_kwargs = {}
+    result = await _nebius_client().embeddings.create(input=input, model=model, **model_kwargs)
+    return [np.array(data.embedding, dtype=np.float64) for data in result.data]
```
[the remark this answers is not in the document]
amrrs: Fixed in 1eda94c6 — same `require_package('openai')` check added here.

sergey-mkhitaryan:
@amrrs thanks for your pr. We will review it shortly.

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+@env.register_client('nebius', credential_param='api_key')
+def _(api_key: str) -> 'openai.AsyncOpenAI':
+    import openai
+
+    return openai.AsyncOpenAI(
+        api_key=api_key,
+        base_url='https://api.tokenfactory.nebius.com/v1',
+        http_client=httpx.AsyncClient(limits=httpx.Limits(max_keepalive_connections=100, max_connections=500)),
```
sergey-mkhitaryan: is there any particular reason why this configuration? if no, let's drop it and rely on httpx defaults

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,188 @@
...
+    if tool_choice is not None and not tool_choice['parallel_tool_calls']:
+        if 'extra_body' not in model_kwargs:
+            model_kwargs['extra_body'] = {}
+        model_kwargs['extra_body']['parallel_tool_calls'] = False
+
+    result = await _nebius_client().chat.completions.with_raw_response.create(
+        messages=messages, model=model, **model_kwargs
+    )
+
+    return json.loads(result.text)
```
sergey-mkhitaryan: please check that `result` is successful. same in embedding udf.

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+_embedding_dimensions_cache: dict[str, int] = {'Qwen/Qwen3-Embedding-8B': 4096}
```
sergey-mkhitaryan: let's call it simply `_embedding_dimensions` as it's not really a "cache"

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,188 @@
...
+        >>> tbl.add_computed_column(
+        ...     response=chat_completions(
+        ...         messages, model='meta-llama/Llama-3.3-70B-Instruct'
+        ...     )
+        ... )
+    """
+    if model_kwargs is None:
+        model_kwargs = {}
+
+    env.Env.get().require_package('openai')
```
sergey-mkhitaryan: nit: make this the first statement in the function. same below.

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,188 @@
...
+    Examples:
+        Add a computed column that applies the model `Qwen/Qwen3-Embedding-8B`
+        to an existing Pixeltable column `tbl.text` of the table `tbl`:
+
+        >>> tbl.add_computed_column(
+        ...     response=embeddings(tbl.text, model='Qwen/Qwen3-Embedding-8B')
+        ... )
+
+        `Qwen/Qwen3-Embedding-8B` produces 4096-dimensional embeddings by default, which exceed
+        pgvector's dimensionality limits for `add_embedding_index()` (2000 for `fp32` precision,
```
sergey-mkhitaryan: this is a good example, but pgvector is pixeltable's implementation detail, not a part of the api, so let's not mention it here. fp16 and fp32 are parts of api, but I don't feel like we need to talk about them here as they are described in the `add_embedding_index()` docstring. maybe just say that 4000 is max supported by pixeltable embedding indexes, so with this model we should specify a custom dimension.

tests/functions/test_nebius.py
```diff
@@ -0,0 +1,55 @@
...
+
+import pixeltable as pxt
+import pixeltable.type_system as ts
+
+from ..utils import rerun, skip_test_if_no_client, skip_test_if_not_installed, validate_update_status
+
+pytestmark = pytest.mark.local('UDF/integration test')
+
+
+@pytest.mark.remote_api
```
sergey-mkhitaryan, line 11: also add: `@pytest.mark.very_expensive`

tests/functions/test_nebius.py
```diff
@@ -0,0 +1,55 @@
...
+        type_info = t._get_schema()
+        assert isinstance(type_info['embedding'], ts.ArrayType)
+        assert type_info['embedding'].shape == (4096,)
+
+        validate_update_status(t.insert(input='Hello, world!'), 1)
+        result = t.collect()
+        assert len(result['embedding'][0]) == 4096
+
+        # Via add_embedding_index(), which requires a statically known embedding dimension. The default
+        # 4096 dimensions exceed pgvector's indexing limits, so request a truncated size instead.
```
sergey-mkhitaryan: again no need to mention pgvector. moreover, I don't think that this entire comment is necessary.

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,188 @@
+"""
+Pixeltable UDFs for Nebius Token Factory models.
+
+Provides integration with Nebius Token Factory's language and embedding models. In order to use
+them, you must first `pip install openai` and configure your Nebius Token Factory API key.
```
sergey-mkhitaryan: ...by setting the `NEBIUS_API_KEY` environment variable or `nebius.api_key` in the Pixeltable config file.

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,188 @@
...
+    Request throttling:
+    Applies the rate limit set in the config (section `nebius`, key `rate_limit`). If no rate
+    limit is configured, uses a default of 600 RPM.
+
+    __Requirements:__
+
+    - `pip install openai`
+
+    Args:
+        messages: A list of messages to use for chat completion, as described in the Nebius API documentation.
```
sergey-mkhitaryan, line 66: does nebius support vision? I think we should either implement image serialization similarly to `chat_completions()` in `openai.py` or explicitly document that vision is not supported.

sergey-mkhitaryan (changes requested):
looks pretty good overall, mostly minor comments. thanks!

pixeltable/functions/nebius.py
amrrs: Fixed: dropped the custom `httpx.Limits` and now just use `openai.AsyncOpenAI(api_key=..., base_url=...)` with httpx defaults.

pixeltable/functions/nebius.py
amrrs: Fixed in both UDFs: we now check the parsed JSON response for an `error` key and raise `excs.ExternalServiceError(excs.ErrorCode.PROVIDER_ERROR, ...)` instead of silently returning/mis-parsing a failed response. `embeddings()` now also goes through `with_raw_response` (matching `openai.py`'s pattern) so the same check applies there.

pixeltable/functions/nebius.py
amrrs: Renamed to `_embedding_dimensions`.

pixeltable/functions/nebius.py
amrrs: Moved `require_package('openai')` to the first statement in both `chat_completions()` and `embeddings()`.

pixeltable/functions/nebius.py
amrrs: Reworded per your suggestion: dropped the pgvector/fp16/fp32 details and now just say the model's default 4096 dims exceed the 4000-dim max supported by Pixeltable embedding indexes, so a custom `dimensions` value must be requested via `model_kwargs`.

tests/functions/test_nebius.py
amrrs: Added `@pytest.mark.very_expensive` to the test class.

tests/functions/test_nebius.py
amrrs: Removed the comment.

pixeltable/functions/nebius.py
amrrs: Added: "...by setting the `NEBIUS_API_KEY` environment variable or `nebius.api_key` in the Pixeltable config file."

pixeltable/functions/nebius.py
amrrs: Confirmed Nebius does support vision (e.g. `Qwen/Qwen2-VL-72B-Instruct`) using the same OpenAI-compatible `image_url` content format. Implemented image serialization in `chat_completions()` matching `openai.py` (base64-encode any `PIL.Image.Image` passed in an `image_url` content part) and added a docstring example.

amrrs:
> looks pretty good overall, mostly minor comments. thanks!

thanks for the comments, could you please review again @sergey-mkhitaryan

pixeltable/functions/nebius.py
sergey-mkhitaryan: When I tried the example from the docstring, I got 404 (model does not exist). use `Qwen/Qwen2.5-VL-72B-Instruct` instead. also add a unit test for vision, something like this:

```
    def test_vision(self, uses_db: None) -> None:
        skip_test_if_not_installed('openai')
        skip_test_if_no_client('nebius')
        from pixeltable.functions.nebius import chat_completions

        t = pxt.create_table('test_tbl', {'image': pxt.Image})
        msgs = [
            {
                'role': 'user',
                'content': [
                    {'type': 'text', 'text': "What's in this image?"},
                    {'type': 'image_url', 'image_url': t.image},
                ],
            }
        ]
        t.add_computed_column(response=chat_completions(msgs, model='Qwen/Qwen2.5-VL-72B-Instruct'))

        validate_update_status(t.insert(image=SAMPLE_IMAGE_URL), 1)
        result = t.collect()
        assert 'broccoli' in result['response'][0]['choices'][0]['message']['content'].lower()
```

this one passes for me locally.

amrrs: Good catch, thanks — switched the docstring example to `Qwen/Qwen2.5-VL-72B-Instruct` and added `test_vision` (using your snippet) to `test_nebius.py`.

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,229 @@
+"""
+Pixeltable UDFs for Nebius Token Factory models.
+
+Provides integration with Nebius Token Factory's language and embedding models. In order to use
+them, you must first `pip install openai` and configure your Nebius Token Factory API key, by
```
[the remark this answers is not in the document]
amrrs: Added `pixeltable.functions.nebius` (with `chat_completions` and `embeddings`) to `docs/public_api.opml`, alongside the other provider modules.

pixeltable/functions/nebius.py
amrrs: Sorry for the dumb one! fixing it now!

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,229 @@
...
+    model: str,
+    model_kwargs: dict[str, Any] | None = None,
+    tools: list[dict[str, Any]] | None = None,
+    tool_choice: dict[str, Any] | None = None,
+) -> dict:
+    """
+    Creates a model response for the given chat conversation.
+
+    Equivalent to the Nebius Token Factory `chat/completions` API endpoint.
+    For additional details, see: <https://docs.nebius.com/studio/inference/api-reference>
```
sergey-mkhitaryan: this url (it appears 3 times in this file) forwards to https://docs.tokenfactory.nebius.com/quickstart. if that is the intention, just use that. if not, let's find the appropriate page to point to.

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,229 @@
...
+
+    if model_kwargs is None:
+        model_kwargs = {}
+
+    result = await _nebius_client().embeddings.with_raw_response.create(input=input, model=model, **model_kwargs)
+
+    result_json = json.loads(result.text)
+    if 'error' in result_json:
+        raise excs.ExternalServiceError(excs.ErrorCode.PROVIDER_ERROR, str(result_json['error']), provider='nebius')
+    return [np.array(data['embedding'], dtype=np.float64) for data in result_json['data']]
```
sergey-mkhitaryan, line 216: `test_embeddings` no longer passes: `ValueError: could not convert string to float:`

sergey-mkhitaryan (changes requested):

pixeltable/functions/nebius.py
amrrs: So sorry this is messed up! fixed those urls and tested them as well on the browser, it works without redirect!

pixeltable/functions/nebius.py
amrrs: this was due to the default encoding format it seems, fixed that and ran the test suite and it's no more erroring out

pixeltable/functions/nebius.py
amrrs: @sergey-mkhitaryan also i'm generally sorry this has gone back and forth!

sergey-mkhitaryan: all good, no worries at all : )

pixeltable/functions/nebius.py
```diff
@@ -0,0 +1,231 @@
...
+    input: Batch[str], *, model: str, model_kwargs: dict[str, Any] | None = None
+) -> Batch[pxt.Array[(None,), pxt.Float]]:
+    """
+    Query an embedding model for a given string of text.
+
+    Equivalent to the Nebius Token Factory `embeddings` API endpoint.
+    For additional details, see: <https://docs.tokenfactory.nebius.com/api-reference/inference/create-embeddings>
+
+    Request throttling:
+    Applies the rate limit set in the config (section `nebius`, key `rate_limit`). If no rate
```
sergey-mkhitaryan, line 169: Let's not mention the default RPM of 600. That's the service-side default and can change without us knowing in the future.

sergey-mkhitaryan (approved):
Approving but please address one last comment from me.

sergey-mkhitaryan:
`./changelog/changelog.mdx:44:3` is known problem unrelated to this pr, we are working to fix it.

amrrs:
Thank you very much @sergey-mkhitaryan for your support on this!

sergey-mkhitaryan:
@amrrs thanks again for working on this.

PR #1453, opened 2026-07-11
Agent-friendly API changes

mkornacker:
Breaking changes:
  - Removed ColumnRef.select()/show()/head()/tail()/count()/distinct(); use t.select(t.col)... instead
  - string.count(pattern) is now callable as a method on string columns (was shadowed by ColumnRef.count())
  - On json columns, the six removed names now denote field accesses
  - vision.eval_detections() output field 'class' renamed to 'category' (matches COCO vocabulary; mean_ap updated)                                 
  
  Expression guards:
  - Expr.__iter__/__len__ now raise with guidance; iterating a json/array column previously hung in an infinite loop                         

  Operators and builtins:
  - t.x ** y is a native arithmetic operator (POW; result type Float, SQL pow())
  - abs(expr) and round(expr[, n]) work via __abs__/__round__, delegating to math.abs/math.round

  New UDFs (Python-parity gaps that trip up coding agents):
  - string: split, rsplit, splitlines (typed list[str])
  - math: sqrt, exp, ln, log10 (with SQL pushdowns); pow() now takes float parameters
  - image: size property, typed as a (width, height) pair so size[0]/size[1] work
  - timestamp: date(); date: replace()

  Method lookup:
  - Int columns now resolve Float-declared method UDFs (t.int_col.abs()/.sqrt()/.pow() etc. previously raised)

  Typed returns (structured Json column types + SDK doc shapes):
  - huggingface: TypedDicts for detr_for_segmentation, vit_for_image_classification, detr_to_coco, text_classification, token_classification, question_answering
  - yolox: yolox and yolo_to_coco; rerankers: voyageai.rerank, jina.rerank
  - date/timestamp isocalendar share an IsoCalendar TypedDict; string partition/rpartition return tuple[str, str, str]
  - document_splitter: Heading and BoundingBox TypedDicts drive the output column types; __next__ returns DocumentChunk

  Type system:
  - TypedDict detection uses is_typeddict(), recognizing typing_extensions.TypedDict and TypedDict subclasses
  - from_python_type() unwraps NotRequired[T], enabling mixed required/optional TypedDict fields


pixeltable/exprs/arithmetic_expr.py
```diff
@@ -22,7 +22,11 @@ class ArithmeticExpr(Expr):
     operator: ArithmeticOperator
 
     def __init__(self, operator: ArithmeticOperator, op1: Expr, op2: Expr):
-        if op1.col_type.is_json_type() or op2.col_type.is_json_type() or operator == ArithmeticOperator.DIV:
+        if (
+            op1.col_type.is_json_type()
+            or op2.col_type.is_json_type()
+            or operator in (ArithmeticOperator.DIV, ArithmeticOperator.POW)
```
aaron-siegel: I don't think this is correct - `a**b` is an int if `a` and `b` are both ints (only DIV is an automatic float)

pixeltable/exprs/arithmetic_expr.py
```diff
@@ -154,6 +161,9 @@ def eval_non_null(self, op1_val: float, op2_val: float) -> float:
             return op1_val % op2_val
         elif self.operator == ArithmeticOperator.FLOORDIV:
             return op1_val // op2_val
+        elif self.operator == ArithmeticOperator.POW:
+            # the result type is Float, so a float is returned even for int operands
```
aaron-siegel: But why?

pixeltable/functions/document.py
```diff
@@ -306,7 +339,7 @@ def __next__(self) -> dict[str, Any]:
                 elif md_field == ChunkMetadata.BOUNDING_BOX:
                     result[md_field.name.lower()] = section.metadata.bounding_box
 
-            return result
+            return cast(DocumentChunk, result)
```
aaron-siegel: Can this be avoided by declaring `result: DocumentChunk = {}` above? Then make the assignments explicit, e.g., instead of

```python
result[md_field.name.lower()] = self._doc_title
```
specify it as
```python
result['title'] = self._doc_title
```
(I don't see any harm in doing it this way, and it'll be typesafe)

pixeltable/functions/huggingface.py
aaron-siegel: Missing type hint here (maybe that's why the type: ignore is needed below)

aaron-siegel: Missing type hint

pixeltable/functions/image.py
```diff
@@ -451,6 +451,18 @@ def mode(self: PIL.Image.Image) -> str:
...
 
+@pxt.udf(is_property=True)
+def size(self: PIL.Image.Image) -> tuple[int, int]:
+    """
+    Return the image size as a `[width, height]` list.
+
+    Equivalent to [`PIL.Image.Image.size`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.size).
+    """
+    # the annotation declares the fixed two-element schema; the runtime value must be a json value (a list),
+    # not a tuple, so that path expressions like size[0] evaluate against it
```
aaron-siegel, line 462: Hmmm, or, we could make path expressions do the right thing on tuples?

This feels like a very Claude-y solution ("the annotation declares the fixed two-element schema; the runtime value must be a json value (a list)")... putting a band-aid on it

pixeltable/functions/jina.py
```diff
@@ -184,10 +185,23 @@ def _(model: str, dimensions: int | None) -> ts.ArrayType:
     return ts.ArrayType((dim,), dtype=np.dtype(np.float32), nullable=False)
 
 
+# TypedDict and NotRequired both come from typing_extensions: on Python 3.10 the stdlib TypedDict does not
```
aaron-siegel: Why is this comment necessary?

pixeltable/functions/jina.py
```diff
@@ -10,6 +10,7 @@
 
 import aiohttp
 import numpy as np
+from typing_extensions import NotRequired, TypedDict
```
aaron-siegel, line 16: Can't `TypedDict` come from `typing`?

pixeltable/functions/math.py
```diff
@@ -112,6 +112,66 @@ def _(self: sql.ColumnElement, other: sql.ColumnElement) -> sql.ColumnElement:
...
+    return math.exp(self)
+
+
+@exp.to_sql
+def _(self: sql.ColumnElement) -> sql.ColumnElement:
+    return sql.func.exp(self)
+
+
+@pxt.udf(is_method=True)
+def ln(self: float) -> float:
```
aaron-siegel: I'd call this `log` to align with Python.

pixeltable/functions/string.py
```diff
@@ -478,16 +478,18 @@ def pad(self: str, width: int, side: str = 'left', fillchar: str = ' ') -> str:
...
 
 @pxt.udf(is_method=True)
-def partition(self: str, sep: str = ' ') -> list:
+def partition(self: str, sep: str = ' ') -> tuple[str, str, str]:
     """
     Splits string at the first occurrence of `sep`, and returns 3 elements containing the part before the
     separator, the separator itself, and the part after the separator. If the separator is not found, return 3 elements
     containing string itself, followed by two empty strings.
     """
+    # the runtime value is a list (json arrays are lists; path expressions index into them); the annotation
```
aaron-siegel, line 487: Same comment as above.

pixeltable/functions/timestamp.py
```diff
@@ -8,12 +8,13 @@
 >>> t.select(t.timestamp_col.year, t.timestamp_col.weekday()).collect()
 """
 
-from datetime import datetime
+from datetime import date as pydate, datetime
```
aaron-siegel: Better IMO would be to just `import datetime` and reference `datetime.date`.

I find that nonstandard `import ... as ...` imports make the code harder to follow.

tests/functions/test_math.py
```diff
@@ -85,3 +85,44 @@ def test_methods(self, method_type: type, uses_db: None) -> None:
                 assert mref.fn.name == pxt_fn.name, pxt_fn
             else:
                 raise AssertionError()
+
+    def test_log_fns(self, uses_db: None) -> None:
```
aaron-siegel: Seems like unnecessary code duplication. Can't we fold it into the preceding test and restrict the test cases via a flag in test_params?

aaron-siegel (changes requested):

pixeltable/functions/document.py
mkornacker: Done.

pixeltable/functions/huggingface.py

pixeltable/functions/math.py
mkornacker: Done.

pixeltable/functions/timestamp.py
mkornacker: Done.

pixeltable/functions/jina.py
mkornacker: Done.

pixeltable/functions/jina.py
mkornacker: Done.

tests/functions/test_math.py
mkornacker: Done.

pixeltable/functions/image.py
mkornacker: Out of scope, we're doing this everywhere.

pixeltable/functions/string.py
mkornacker: Still out of scope.

pixeltable/exprs/arithmetic_expr.py
mkornacker: Done.

pixeltable/exprs/arithmetic_expr.py
mkornacker: Done.

aaron-siegel (approved):

PR #1462, opened 2026-07-15
docs: Add Nebius Token Factory provider notebook

pierrebrunelle:


docs/release/howto/providers/working-with-nebius.ipynb
```diff
@@ -0,0 +1,223 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pixeltable as pxt\n",
+    "\n",
+    "# Remove the 'nebius_demo' directory and its contents, if it exists\n",
+    "pxt.drop_dir('nebius_demo', force=True)\n",
```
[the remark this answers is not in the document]
pierrebrunelle: Disagree — `force=True` already makes a missing directory a no-op.

In `Catalog.drop_dir`, raise-on-missing is gated by `if_not_exists == ERROR and not force`, so with `force=True` a missing dir is skipped (verified locally). Same pattern as the other provider notebooks (`together`, `openrouter`, `deepseek`, etc.).

`if_not_exists='ignore'` would also work, but isn't needed here when `force=True` is already set.

docs/release/howto/providers/working-with-nebius.ipynb
```diff
@@ -0,0 +1,680 @@
...
+ "cells": [
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "vscode": {
+     "languageId": "raw"
+    }
+   },
+   "source": [
+    "---\n",
```
sergey-mkhitaryan, line 11: this cell is displayed as-is on my machine, without any formatting

pixeltable/runtime.py
```diff
@@ -182,6 +183,25 @@ def get_client(self, name: str) -> Any:
         self._clients[name] = client
         return client
 
+    async def close_clients(self) -> None:
```
mkornacker: What about reset_runtime()?

sergey-mkhitaryan: `reset_runtime()` does a bunch of stuff like invalidating the catalog's cache, re-creating the thread-local runtime, etc. that I don't think are necessarily what we need in `ExecNode`.

demianarc:
The Nebius notebook itself uses current chat and embedding IDs, but this docs PR also includes unrelated runtime/client-lifecycle changes in `pixeltable/runtime.py` and `pixeltable/exec/exec_node.py`. The branch is conflicting and currently has failed/cancelled checks.

Please split the runtime fix from the provider notebook, clear stored execution outputs/counts from the docs artifact, then rebase onto the already merged native provider (#1448). A focused notebook/source smoke fixture should validate current model IDs without requiring the runtime cleanup to land in the same PR.

mkornacker:
Claude is pointing out that this has a number of gaps in terms of runtime cleanup, and it would be worth fixing all of them here.
- the loop shutdown sequence isn't complete (missing shutdown_default_executor())
- it needs to be called in more places.

Concretely:
- new Runtime._owns_event_loop, set in _init_event_loop() (we're currently using is_running() as a signal in reset_runtime(), which is incorrect)
- Runtime.close():
```
  def close(self) -> None:
      """Close this runtime's clients, and the event loop if this runtime created it."""
      loop = self._event_loop
      self._event_loop = None
      if loop is None or loop.is_closed():
          # nothing was created on a loop, or the loop is already gone and its sockets with it
          self._clients.clear()
          return

      if not self._owns_event_loop:
          # an adopted loop (eg Jupyter's) keeps running and is not ours to tear down, but the clients are
          # still bound to it: hand them to that loop instead of dropping them
          if loop.is_running():
              # deliberately not awaited: this may be running on that very loop, and waiting would deadlock
              asyncio.run_coroutine_threadsafe(self.close_clients(), loop)
          else:
              loop.run_until_complete(self.close_clients())
          return

      try:
          tasks = [t for t in asyncio.all_tasks(loop) if not t.done()]
          for t in tasks:
              t.cancel()
          if len(tasks) > 0:
              loop.run_until_complete(asyncio.gather(*tasks, return_exceptions=True))
          loop.run_until_complete(self.close_clients())
          loop.run_until_complete(loop.shutdown_asyncgens())
          loop.run_until_complete(loop.shutdown_default_executor())
      finally:
          loop.close()
          self._clients.clear()

```
and
```
  def reset_runtime() -> None:
      """Reset the current thread's Runtime instance. Used for testing."""
      runtime = getattr(_thread_local, 'runtime', None)
      if runtime is not None:
          ...catalog invalidation, unchanged...
          # the executor's worker thread has its own Runtime, loop and clients; close them from inside it
          executor = runtime._run_coro_executor
          if executor is not None:
              executor.submit(lambda: get_runtime().close()).result()
              executor.shutdown(wait=True)
          runtime.close()
      _thread_local.runtime = Runtime()
```
-

pixeltable/runtime.py
```diff
@@ -185,6 +183,25 @@ def get_client(self, name: str) -> Any:
...
+        """Gracefully close this runtime's clients and drop them.
+
+        Must be awaited on the event loop the clients were created on, while it is still running: async
+        clients are bound to that loop, and closing them there releases their sockets in order. Otherwise
+        the sockets are left for GC to close on an already-closed loop, which raises 'Event loop is closed'.
+        """
+        for client in self._clients.values():
+            close = getattr(client, 'close', None)
```
mkornacker: This is skipping clients that today are missing a close() (bfl, jina). Would be ideal to close that gap now, it's not a lot of code.

mkornacker (changes requested):

mkornacker (approved):

PR #1464, opened 2026-07-15
[PXT-1236] get rid of caching in TableVersionHandle

sergey-mkhitaryan:
Motivation: to fix [PXT-1236](https://pixeltable.atlassian.net/browse/PXT-1236), we want to start evicting older (LRU) cached TV entries from `Catalog._tbl_versions`. The thread-local cache in `TableVersionHandle` gets in the way of that: if a table is evicted from Catalog's cache but is still in `TableVersionHandle._local`, we can end up mixing two different `sql.Table` instances in a query, which results in errors. Besides, a thread-local cache in `TableVersionHandle` is confusing and difficult to reason about.

Here's why the new code is semantically and performance-wise equivalent to the old one.

When in transaction, the new code simply relies on Catalog to obtain table version. Catalog maintains its own cache, so the new code has a similar performance (i.e. TV will only be loaded if it isn't already cached and validated).

When not in a transaction and effective version is `None` (i.e. live table): the old code would always go down the `cat.get_tbl_version()` path, same as the new code. To understand why the `return cached` branch would never be hit, observe that Catalog resets `is_validated` to `False` on all live tables at the end of each transaction. A live table is never "valid" unless it was validated in the current transaction, but in this case there's no current transaction.

When not in a transaction and effective version is not `None` (i.e. snapshot): if there is a cached instance, both the old and the new code return it without starting a new transaction. If not, in both cases a transaction is started to load the snapshot, which is then cached.

Also in this pr: `Catalog.get_tbl_version()` and `Catalog._get_tbl_version`: `validate_initialized` param defaults to True.

[PXT-1236]: https://pixeltable.atlassian.net/browse/PXT-1236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pixeltable/catalog/table_version_handle.py
```diff
@@ -61,26 +52,23 @@ def is_snapshot(self) -> bool:
...
-            new_tv.is_validated = True
-        else:
-            # either no cached instance on this thread, or the catalog changed
-            new_tv = cat.get_tbl_version(self.key)
-            assert new_tv.key == self.key
-        self._local.tbl_version = new_tv
-        self._local.origin_catalog = cat
-        return new_tv
+        if get_runtime().in_xact:
+            return cat.get_tbl_version(self.key)
```
mkornacker: This uses the default validate_initialized=False, which seems wrong here (and seems wrong in general - why is that the default)?

sergey-mkhitaryan: Good point, I don't know why `validate_initialized` was False by default (that decision probably predates me). Changed. However this call should maintain validation disabled:
```
            # disable validation because we can get here during the TableVersion initialization while the instance is
            # not yet valid.
```

pixeltable/catalog/table_version_handle.py
```diff
@@ -61,26 +52,25 @@ def is_snapshot(self) -> bool:
...
-            new_tv = cat._tbl_versions[self.key]
-            new_tv.is_validated = True
-        else:
-            # either no cached instance on this thread, or the catalog changed
-            new_tv = cat.get_tbl_version(self.key)
-            assert new_tv.key == self.key
-        self._local.tbl_version = new_tv
-        self._local.origin_catalog = cat
-        return new_tv
+        if get_runtime().in_xact:
```
mkornacker: At least skip the repeated get_runtime() calls.

pixeltable/catalog/table_version_handle.py
```diff
@@ -61,26 +52,25 @@ def is_snapshot(self) -> bool:
...
+            # disable validation because we can get here during the TableVersion initialization while the instance is
+            # not yet valid.
+            return cat.get_tbl_version(self.key, validate_initialized=False)
```
mkornacker: And call _get_tbl_version() here directly.

mkornacker (approved):

PR #1466, opened 2026-07-16
pxt schema update

mkornacker:
Plus: a fix for a bbox utility udf needed for the Backblaze webinar.


pixeltable/catalog/model.py
```diff
@@ -507,11 +505,14 @@ def __prepare__(  # type: ignore[override]
                 raise excs.RequestError(
                     excs.ErrorCode.INVALID_ARGUMENT, f'{display_name}: `name` must be a valid Pixeltable identifier.'
                 )
-            if tbl_name in mcs.registered_models:
+            # scoped to this model_base() so a long-lived process can re-load a schema without colliding with a
+            # previous load; duplicate names within one schema are still caught
+            base_models = bases[0].__dict__['__registered_models__']
```
aaron-siegel: Just `bases[0].__registered_models__` should work

pixeltable/catalog/model.py
```diff
@@ -507,11 +505,14 @@ def __prepare__(  # type: ignore[override]
                 raise excs.RequestError(
                     excs.ErrorCode.INVALID_ARGUMENT, f'{display_name}: `name` must be a valid Pixeltable identifier.'
                 )
-            if tbl_name in mcs.registered_models:
+            # scoped to this model_base() so a long-lived process can re-load a schema without colliding with a
```
aaron-siegel: This comment feels out of context; I think it can just be removed, but if we keep it I'd add a little more context. (Scoping it to model_base() is clearly correct and IMO doesn't need specific justification in the comments; the residual ClassVar was just an oversight on my part.)

aaron-siegel (approved):

pixeltable/catalog/model.py

PR #1472, opened 2026-07-17
[PXT-1240] Table.add_btree_index(), stop adding b-tree indexes by default

sergey-mkhitaryan:
* Table md stores if the table was created with default idxs enabled or not. This property is immutable. (Legacy tables assumed to have default indexes enabled).
* `create_default_idxs=True` effectively means that b-tree indexes are managed by Pixeltable. No manual b-tree index addition or drop is allowed. Creating and dropping columns also creates and drops their indexes automatically.
* New function `Table.add_btree_index()` to add a b-tree index manually when `create_default_idxs=False`. But a view cannot declare a b-tree index on the base table at the moment.
* At most one b-tree index per column is allowed
* Support for explicit b-tree indexes in class-based schemas


pixeltable/catalog/local_table.py
```diff
@@ -418,6 +418,7 @@ def add_columns(
         self,
         schema: Mapping[str, type | ColumnSpec],
         if_exists: Literal['error', 'ignore', 'replace', 'replace_force'] = 'error',
+        create_default_idxs: bool = False,
```
aaron-siegel: Make this part of `ColumnSpec` so it can be specified on a per-column basis, rather than all-or-nothing?

That would also reduce a fair amount of boilerplate passing `create_default_idxs` around separately (which is part of the point of `ColumnSpec`)

pixeltable/catalog/local_table.py
```diff
@@ -681,6 +686,86 @@ def do_alter_column() -> None:
...
+                    excs.ErrorCode.UNSUPPORTED_OPERATION,
+                    f'Cannot create a B-tree index on column {col.name!r}: it belongs to a base table. '
+                    'Add the index to the base table instead.',
+                )
+
+            err = self._tbl_version.get()._btree_index_error(col)
+            if err is not None:
+                raise err
+
+            if idx_name is not None and self._resolve_btree_index_name_collision(idx_name, if_exists_):
```
aaron-siegel: I find the semantics here a little confusing.

Is this correct?: If I add_btree_index() with if_exists='ignore', and there is already a btree index with the same name, then no error will be generated, even if the index is on a different column?

If so, I find that very counterintuitive (I'm not sure what the user scenario is where that behavior would be desirable)

pixeltable/catalog/local_table.py
```diff
@@ -681,6 +686,86 @@ def do_alter_column() -> None:
...
+        if if_exists == IfExistsParam.ERROR:
+            raise excs.AlreadyExistsError(excs.ErrorCode.INDEX_ALREADY_EXISTS, f'Duplicate index name: {idx_name}')
+        assert if_exists == IfExistsParam.IGNORE
+        if not isinstance(tv.idxs_by_name[idx_name].idx, index.BtreeIndex):
+            raise excs.RequestError(
+                excs.ErrorCode.UNSUPPORTED_OPERATION, f'Index {idx_name!r} already exists, but is not a B-tree index.'
+            )
+        return True
+
+    def _find_matching_btree_idxs(self, col: Column) -> list[TableVersion.IndexInfo]:
```
aaron-siegel: There's at most one, right? So wouldn't it be more honest for this to be
```python
_get_btree_idx(self, col: Column) -> TableVersion.IndexInfo | None
```

pixeltable/catalog/model.py
```diff
@@ -105,6 +105,17 @@ class EmbeddingIndex:
     precision: Literal['fp16', 'fp32'] = 'fp16'
 
 
+@dataclass(frozen=True)
+class BtreeIndex:
```
aaron-siegel, line 131: Just an FYI: You'll need to evolve the diff/update code as well, once that merges (later today).

pixeltable/catalog/table_version.py
```diff
@@ -546,27 +572,39 @@ def add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) ->
...
+            )
         if col.col_type.is_media_type() and col.is_iterator_col:
-            return False
-        if not col.col_type.is_scalar_type() and not (col.col_type.is_media_type() and not col.is_computed):
-            # wrong type for a B-tree
-            return False
-        if col.col_type.is_bool_type():  # noqa : SIM103 Supress `Return the negated condition directly` check
+            return excs.RequestError(
+                excs.ErrorCode.UNSUPPORTED_OPERATION,
+                f'Cannot create a B-tree index on column {col.name!r}, which is produced by an iterator.',
```
aaron-siegel: Why not?

pixeltable/catalog/table_version.py
```diff
@@ -546,27 +572,39 @@ def add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) ->
...
+            )
+        if not col.col_type.is_scalar_type() and not col.col_type.is_media_type():
+            return excs.RequestError(
+                excs.ErrorCode.TYPE_MISMATCH,
+                f'Index on column {col.name}: B-tree index requires scalar or media type, got {col.col_type}',
```
aaron-siegel: What are the semantics of a btree index on a media type? Sorted by fileurl?

pixeltable/catalog/table_version.py
```diff
@@ -546,27 +572,39 @@ def add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) ->
...
+            return excs.RequestError(
+                excs.ErrorCode.UNSUPPORTED_OPERATION,
+                f'Cannot create a B-tree index on computed media column {col.name!r}.',
+            )
+        if col.col_type.is_bool_type():
             # B-trees on bools aren't useful
-            return False
-        return True
+            return excs.RequestError(
+                excs.ErrorCode.UNSUPPORTED_OPERATION, f'A B-tree index on boolean column {col.name!r} is not supported.'
```
aaron-siegel: I'd fold this into the earlier type-checking error
(so the error becomes "B-tree index requires non-bool scalar type or a media type")

And I'd expect the error code to be the same in both instances.

pixeltable/catalog/table_version.py
```diff
@@ -576,6 +614,8 @@ def _create_index_md(
         self.next_idx_id += 1
         if idx_name is None:
             idx_name = f'idx{idx_id}'
+            while idx_name in self.idxs_by_name:
+                idx_name += '_'
```
aaron-siegel: So we could get `idx0____` or some such? Why not just increment idx_id until we find an unused one?

tests/test_table_model.py
```diff
@@ -299,6 +287,31 @@ class ExampleTableModel(TableModel, name='test_table'):
...
+    def test_btree_index_declaration(self, make_catalog_path: Callable[[str], str]) -> None:
+        p = make_catalog_path
+        TableModel = pxt.model_base()
+
+        class ExampleTableModel(TableModel, name='test_table'):
+            id: pxt.Required[pxt.Int]
+            name: pxt.String
+            img: pxt.Image
+
+            name_idx = BtreeIndex(name)  # column referenced directly
```
aaron-siegel: What's the `# column referenced directly` for?

tests/test_table_model.py
```diff
@@ -299,6 +287,31 @@ class ExampleTableModel(TableModel, name='test_table'):
...
+
+        # Verify both the model-based table and an analogous manually built one carry the same indices.
+        indices = tbl.get_metadata()['indices']
+        assert indices['name_idx']['index_type'] == 'btree' and indices['name_idx']['columns'] == ['name']
+        assert indices['img_idx']['index_type'] == 'btree' and indices['img_idx']['columns'] == ['img']
+
+        tbl2 = pxt.create_table(p('test_table_2'), {'id': pxt.Required[pxt.Int], 'name': pxt.String, 'img': pxt.Image})
+        tbl2.add_btree_index('name', idx_name='name_idx')
+        tbl2.add_btree_index('img', idx_name='img_idx')
+        assert schema_from_tbl_md(tbl.get_metadata()) == schema_from_tbl_md(tbl2.get_metadata())
```
aaron-siegel: Does `schema_from_tbl_md()` verify indexes? (I can't remember)

pixeltable/catalog/table_version.py
sergey-mkhitaryan: yes

pixeltable/catalog/table_version.py
sergey-mkhitaryan: the decision not to create such indexes by default creates a long time back. It might have been because iterator output media files are named by pixeltable, therefore an index on them is not particularly meaningful. I'm inheriting this decision not to create default indexes and promoting it to a decision not to allow b-tree indexes in this case at all, which maybe a step too far. Not sure.

tests/test_table_model.py
sergey-mkhitaryan: yes, it includes indexes

pixeltable/metadata/schema.py
```diff
@@ -249,6 +249,10 @@ class TableMd:
     # TODO when the catalog migration happens, let's backfill and get rid of the default.
     is_versioned: bool = True
 
+    # Indicates if default b-tree indexes are enabled for this table. None if unknown (i.e. the table was created before
+    # this parameter was introduced).
+    default_idxs_enabled: bool | None = None
```
mkornacker: Or simply make the default true, which reflects the current situation?

pixeltable/index/btree.py
```diff
@@ -32,12 +32,22 @@ def str_filter(s: str | None) -> str | None:
     def __init__(self) -> None:
         pass
 
+    @classmethod
+    def type_error(cls, c: 'catalog.Column') -> excs.RequestError | None:
```
mkornacker: I didn't understand what this did before reading the code.

validate_col_type()? or check_...?

pixeltable/catalog/table_metadata.py
```diff
@@ -79,6 +79,8 @@ class TableMetadata(TypedDict):
     """Index metadata for all of the indices of the table."""
     is_versioned: bool
     """`True` if this is a versioned table."""
+    default_idxs_enabled: bool
```
mkornacker: Let's go with has_default_idxs, which is an already established pattern (and adapt the other ..._enabled instances).

pixeltable/catalog/model.py
```diff
@@ -1276,12 +1314,44 @@ def op() -> dict[str, TableDiff]:
...
+                    )
+                )
+
+            # Default indexes have no counterpart in __indexes__, and create_default_idxs=True is incompatible with
+            # explicitly declared B-tree indexes. So, if the existing table has default indexes enabled, all of its
+            # B-tree indexes are default indexes, and they can all be ignored, because resolving the column diff takes
+            # care of the indexes too. Otherwise all B-tree indexes are explicitly declared, so they are diffed and
+            # compared like embedding indexes. If the two sides disagree on create_default_idxs, no meaningful diff of
+            # B-tree indexes can be computed, so they are left out.
+            include_btree_idxs = not model_default_idxs and not existing_default_idxs
```
mkornacker, line 1344: I think this should only look at existing_default_idxs (if you're trying to change it, you already get an error).

tests/pixeltable_cli/test_smoke.py
```diff
@@ -191,8 +191,8 @@ def test_listings_honor_wd(
         """'columns' and 'idxs' with no path cover the working directory, not the whole catalog."""
         p = make_catalog_path
         pxt.create_dir(p('cli_cwd_list'), if_exists='ignore')
-        pxt.create_table(p('cli_cwd_list/inside'), {'x': pxt.String}, if_exists='replace')
-        pxt.create_table(p('cli_cwd_outside'), {'y': pxt.String}, if_exists='replace')
+        pxt.create_table(p('cli_cwd_list/inside'), {'x': pxt.String}, if_exists='replace', create_default_idxs=True)
```
mkornacker: Why always True?

tests/test_dirs.py
```diff
@@ -106,10 +106,11 @@ def test_get_dir_tree_error_count(self, make_catalog_path: Callable[[str], str])
         p = make_catalog_path
         t = pxt.create_table(p('errs'), {'x': pxt.Int})
         t.add_computed_column(y=_fail_on_neg(t.x))
+        t.add_btree_index('y')
```
mkornacker: Why not leave this out and adapt the reported numbers?

pixeltable/catalog/local_table.py
```diff
@@ -749,6 +752,95 @@ def do_alter_column() -> None:
...
+        ):
+            if self._tbl_version.get().default_idxs_enabled:
+                raise excs.RequestError(
+                    excs.ErrorCode.UNSUPPORTED_OPERATION,
+                    f'{self._display_str()}: cannot add a B-tree index to a table created with '
+                    'create_default_idxs=True; its eligible columns are indexed automatically.',
+                )
+
+            col = self._resolve_column_parameter(column)
+            if col.tbl_handle.id != self._tbl_version.get().id:
```
mkornacker: Why not allow this? We do it for embedding indices.

pixeltable/catalog/local_table.py
```diff
@@ -749,6 +752,95 @@ def do_alter_column() -> None:
...
+                excs.ErrorCode.UNSUPPORTED_OPERATION, f'Index {idx_name!r} already exists, but is not a B-tree index.'
+            )
+        if info.col.qid != col.qid or if_exists == IfExistsParam.ERROR:
+            raise excs.AlreadyExistsError(
+                excs.ErrorCode.INDEX_ALREADY_EXISTS, f'Index {idx_name!r} already exists on column {info.col.name!r}.'
+            )
+        assert if_exists == IfExistsParam.IGNORE
+        return True
+
+    def _find_existing_btree_idx(self, col: Column) -> TableVersion.IndexInfo | None:
```
mkornacker: Are there other callers? If not, inline, it does very little.

pixeltable/catalog/local_table.py
```diff
@@ -749,6 +752,95 @@ def do_alter_column() -> None:
...
+                    )
+                assert if_exists_ == IfExistsParam.IGNORE
+                return
+
+            idx = index.BtreeIndex()
+            _ = self._tbl_version.get().add_index(col, idx_name=idx_name, idx=idx)
+
+        FileCache.get().emit_eviction_warnings()
+
+    def _resolve_btree_index_name_collision(self, idx_name: str, col: Column, if_exists: IfExistsParam) -> bool:
```
mkornacker: Are there other callers? If not, inlining it would make the code easier to follow. This doesn't really add up to a named piece of logic (ie, I have to read the entire function to understand what it does; the name doesn't help me).

pixeltable/catalog/table_version.py
```diff
@@ -314,10 +325,17 @@ def create_initial_md(
             index_cols.extend([val_col, undo_col])
 
             idx_id = next(index_ids)
+            if idx_name is not None:
+                resolved_idx_name = idx_name
+            else:
+                # skip index ids whose default name an explicitly named index has already claimed
```
mkornacker: I don't think that's a good idea, to have the name drive the id assignment. You can simply pick any number here that hasn't been taken.

pixeltable/catalog/table_version.py
```diff
@@ -314,10 +325,17 @@ def create_initial_md(
             index_cols.extend([val_col, undo_col])
 
             idx_id = next(index_ids)
+            if idx_name is not None:
+                resolved_idx_name = idx_name
```
mkornacker, line 325: Declare before if.

pixeltable/globals.py
```diff
@@ -60,7 +60,7 @@ def create_table(
     source: TableDataSource | None = None,
     source_format: Literal['csv', 'excel', 'parquet', 'json'] | None = None,
     schema_overrides: dict[str, Any] | None = None,
-    create_default_idxs: bool = True,
+    create_default_idxs: bool = False,
```
mkornacker: We should rename this to the actual recorded piece of metadata. I'm in favor of has_default_idxs (and then that's also what get_metadata() returns). Having two different names will create confusion (esp. w/ coding agents).

pixeltable/catalog/table_version.py
```diff
@@ -581,39 +600,90 @@ def add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) ->
         return status
 
     @classmethod
-    def _is_btree_indexable(cls, col: Column) -> bool:
+    def _btree_index_error(cls, col: Column) -> excs.RequestError | None:
```
mkornacker: Here too: I found the name '..._error' unintuitive.

Does this validation logic really belong here? It could be written against TablePath/ColumnVersionMd.

pixeltable/catalog/table_version.py
mkornacker: Correct, there is no meaningful query that would benefit from the index (a btree index on a media column is only good for file name comparisons).

pixeltable/catalog/table_version.py
```diff
@@ -581,39 +600,90 @@ def add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) ->
...
     ) -> int:
         """Create md for given index and update self._tbl_md. Returns index id."""
-        idx_id = self.next_idx_id
-        self.next_idx_id += 1
+        existing_names = {i.name for i in self._tbl_md.index_md.values()}
         if idx_name is None:
-            idx_name = f'idx{idx_id}'
+            # find the next index id whose default name isn't taken by an explicitly named index
+            while f'idx{self.next_idx_id}' in existing_names:
+                self.next_idx_id += 1
```
mkornacker: Here, too: I don't think this is a good idea.

pixeltable/catalog/table_version.py
```diff
@@ -581,39 +600,90 @@ def add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) ->
...
+                f'Cannot create a B-tree index on computed media column {col.name!r}.',
+            )
+        return None
+
+    @classmethod
+    def _is_btree_indexable(cls, col: Column) -> bool:
+        return cls._btree_index_error(col) is None
+
+    @classmethod
+    def _validate_btree_idxs(
```
mkornacker: I don't think this makes sense as it is. It assumes that names have already been assigned, which by itself doesn't really make sense. That should probably happen here.

mkornacker (changes requested):
I don't think forcing the user to name btree indices in models is a good idea (there is nothing you can do with the name; it forces you to make a decision that has no impact on anything that follows, aside from possibly leading to a collision).

pixeltable/catalog/model.py
sergey-mkhitaryan: If we do that, the diff is going to contain "drop all indexes" in addition to the unsupported entry generated above, which is not wrong but kind of confusing in my opinion.

pixeltable/catalog/table_version.py
sergey-mkhitaryan: changed: idx name are now f'idx_{uuid.hex()}'

pixeltable/catalog/local_table.py
sergey-mkhitaryan: We do but it doesn't work (https://pixeltable.atlassian.net/browse/PXT-1259).

Filed https://pixeltable.atlassian.net/browse/PXT-1260 to do this separately once PXT-1259 is fixed.

pixeltable/catalog/table_version.py
sergey-mkhitaryan: Moved to BtreeIndex.

Regarding the name, there's `Expr.validation_error()` that does the same thing, so there's a precedent. At any rate, what would you call it?

pixeltable/catalog/table_version.py
sergey-mkhitaryan: I generalized it as `_validate_idxs()` (by adding more index type-agnostic checks, including one that was completely missing), and moved it to the module level. PTAL.

pixeltable/catalog/table_version.py
```diff
@@ -2037,3 +1960,59 @@ def as_dict(self) -> dict:
     def from_dict(cls, d: dict) -> TableVersion:
         key = TableVersionKey.from_dict(d)
         return get_runtime().catalog.get_tbl_version(key)
+
+
+def _validate_idxs(
```
mkornacker: Make this an instance method of TV: it should not/will not be used outside of it, and it needs TV state.

pixeltable/catalog/table_version.py
```diff
@@ -2037,3 +1960,59 @@ def as_dict(self) -> dict:
...
+        info.col.name
+        for info in existing_idxs
+        if isinstance(info.idx, index.BtreeIndex) and info.col.tbl_handle.id == tbl_id
+    }
+    new_names: set[str] = set()
+    new_btree_col_names: set[str] = set()
+
+    for idx_col, idx_name, idx in idxs:
+        if idx_name is not None:
+            # the manifests this validates are keyed by index name, so a name cannot repeat within idxs
```
mkornacker: What are "the manifests"?

pixeltable/catalog/table_version.py
```diff
@@ -2037,3 +1960,59 @@ def as_dict(self) -> dict:
...
+    has_default_idxs: bool,
+    existing_idxs: Iterable[TableVersion.IndexInfo] = (),
+) -> None:
+    """Validate the indexes in idxs, which are about to be created on the table with id tbl_id.
+
+    existing_idxs: the table's live indexes; a new index must not collide with one of those.
+    """
+    existing_by_name = {info.name: info for info in existing_idxs}
+    # names of the columns that already have a B-tree index; a view's base columns are excluded, because the
+    # validation below rejects them as targets anyway
```
mkornacker: Why do we want to prohibit view btrees on view base columns?

pixeltable/catalog/table_version.py
```diff
@@ -2037,3 +1960,59 @@ def as_dict(self) -> dict:
...
+        if not isinstance(idx, index.BtreeIndex):
+            continue
+        if has_default_idxs:
+            raise excs.RequestError(
+                excs.ErrorCode.INVALID_ARGUMENT,
+                'Cannot combine has_default_idxs=True with an explicitly declared B-tree index.',
+            )
+        assert isinstance(idx_col, Column)
+        assert idx_col.name is not None, repr(idx_col)
+        err = index.BtreeIndex.validation_error(idx_col, tbl_id)
```
mkornacker: I'm still finding the name validation_error unhelpful, I basically need to look it up to understand what it does (it validates and returns an error as a flag).

pixeltable/catalog/table_version.py
```diff
@@ -2037,3 +1960,59 @@ def as_dict(self) -> dict:
...
+
+        if not isinstance(idx, index.BtreeIndex):
+            continue
+        if has_default_idxs:
+            raise excs.RequestError(
+                excs.ErrorCode.INVALID_ARGUMENT,
+                'Cannot combine has_default_idxs=True with an explicitly declared B-tree index.',
+            )
+        assert isinstance(idx_col, Column)
+        assert idx_col.name is not None, repr(idx_col)
```
mkornacker: This assumes that idx_col is a Column. Why is that correct here? It could also be a str. 

I'm finding the logic of this function hard to follow. I'm not sure it really stands on its own (= should be a function; a function is an abstraction; can you cleanly articulate what this one does?).

pixeltable/catalog/table_version.py
```diff
@@ -2037,3 +1960,59 @@ def as_dict(self) -> dict:
...
+        err = index.BtreeIndex.validation_error(idx_col, tbl_id)
+        if err is not None:
+            raise err
+        if idx_col.name in new_btree_col_names:
```
mkornacker: Why not also check existing idxs here?

pixeltable/catalog/table_version.py
```diff
@@ -2037,3 +1960,59 @@ def as_dict(self) -> dict:
...
+            raise excs.AlreadyExistsError(
+                excs.ErrorCode.INDEX_ALREADY_EXISTS, f'More than one B-tree index declared on column {idx_col.name!r}.'
+            )
+        if idx_col.name in btree_col_names:
+            raise excs.AlreadyExistsError(
```
mkornacker: So basically two different error messages for the same problem? That'll confuse coding agents.

docs/public_api.opml
```diff
@@ -55,6 +55,7 @@
             <outline text="method|pixeltable.Table.batch_update" />
             <outline text="method|pixeltable.Table.delete" />
             <outline text="method|pixeltable.Table.compute" />
+            <outline text="method|pixeltable.Table.add_btree_index" />
```
mkornacker, line 58: Why no drop_btree_index()?

pixeltable/catalog/local_table.py
```diff
@@ -776,34 +776,47 @@ def add_btree_index(
...
-                )
-
-            err = self._tbl_version.get()._btree_index_error(col)
+            err = index.BtreeIndex.validation_error(col, self._tbl_version.get().id)
             if err is not None:
                 raise err
 
-            if idx_name is not None and self._resolve_btree_index_name_collision(idx_name, col, if_exists_):
-                return
+            if idx_name is not None:
```
mkornacker: This code seems superfluous, given that we already know we don't want to name btree indices.

mkornacker (changes requested):

pixeltable/catalog/table_version.py
sergey-mkhitaryan: Moved it inside, but made it a `@classmethod` because `create_initial_md()` calls it. So all parameters had to stay.

pixeltable/catalog/table_version.py
sergey-mkhitaryan: should have said index specs, but I simply dropped this comment.

pixeltable/catalog/table_version.py
sergey-mkhitaryan: We don't _want_ to prohibit it, but it is not an insignificant effort to support this, and is out of scope of this pr. Some gaps:
* In `Planner._build_update_columns()`, `modified_val_cols = target.get_idx_val_columns(modified_base_cols)` misses index value columns found in views.
* Some work on the expression side will be necessary for such index to actually be used for anything.
* The last item may or may not become a no-op once we start using conditional PG indexes.

pixeltable/catalog/table_version.py
sergey-mkhitaryan: refactored

pixeltable/catalog/table_version.py
sergey-mkhitaryan: no, it's always a Column at this point. all callers resolve columns before coming here. moved the assertion and added a sentence in the docstring that explicitly states that.

sergey-mkhitaryan: regarding the function itself, it validates indexes before adding them to the table. it's used by 3 separate points in this class, so seems pretty solid. wdyt?

pixeltable/catalog/table_version.py
sergey-mkhitaryan: it does, see `existing_by_name` above

pixeltable/catalog/table_version.py
sergey-mkhitaryan: The first one is "you are adding 2 different indexes with the same name", and the 2nd is "the index name that you provided is already taken by an existing index". I can change the first error to some other code, but none of the existing codes look more appropriate than `INDEX_ALREADY_EXISTS`.

docs/public_api.opml
sergey-mkhitaryan: there's no `drop_btree_index()`. we have `drop_index()` and it's already here.

pixeltable/catalog/local_table.py
sergey-mkhitaryan: with class-based schema today there's no way to declare a b-tree index without a name. we will change that in the future, but not as part of this change.

pixeltable/catalog/table_version.py
mkornacker: It's the same thing. Both are named in the same model. The model describes the end state, which in this case means you have a duplicate name.

sergey-mkhitaryan: unified

pixeltable/catalog/local_table.py
```diff
@@ -174,13 +174,16 @@ def _get_metadata(self) -> TableMetadata:
         if any(col.is_pk for col in columns):
             primary_key = [col.name for col in columns if col.is_pk]
 
+        has_default_idxs = self._tbl_version is not None and self._tbl_version.get().has_default_idxs
```
mkornacker, line 173: why not simply tv.has_default_idxs?

pixeltable/catalog/local_table.py
```diff
@@ -752,6 +755,53 @@ def do_alter_column() -> None:
...
+
+        # A B-tree index is parameterless, so replacing one with another achieves nothing; only 'error' and
+        # 'ignore' are meaningful.
+        if if_exists not in ('error', 'ignore'):
+            raise excs.RequestError(
+                excs.ErrorCode.INVALID_ARGUMENT, f"if_exists must be one of: ['error', 'ignore']; got {if_exists!r}"
+            )
+        if_exists_ = IfExistsParam.validated(if_exists, 'if_exists')
+
+        self._check_mutable('add an index to')
```
mkornacker: Do this first, before any asserts.

Is there a test_snapshot.py test case for it?

pixeltable/catalog/model.py
```diff
@@ -127,14 +127,28 @@ def __repr__(self) -> str:
...
+    """A B-tree index specification used in a TableModel or ViewModel definition."""
+
+    column: Any
+
+    def __repr__(self) -> str:
+        return f'BtreeIndex(column={self.column})'
+
+
+# An index specification declared as a class attribute in a TableModel or ViewModel definition.
+IndexSpec = EmbeddingIndex | BtreeIndex
```
mkornacker: This collides with the existing IndexSpec. Is this user-facing? Can we just call this _Index?

pixeltable/catalog/table_version.py
```diff
@@ -659,10 +658,83 @@ def _add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) ->
...
+                if existing_info is not None:
+                    raise excs.AlreadyExistsError(
+                        excs.ErrorCode.INDEX_ALREADY_EXISTS,
+                        f'Index {idx_name!r} already exists on column {existing_info.col.name!r}.',
+                    )
+                new_names.add(idx_name)
+
+            if not isinstance(idx, index.BtreeIndex):
+                continue
+            if has_default_idxs:
```
mkornacker: Check this first. Otherwise a user might hit the name collision error first (and correct it and then get this).

pixeltable/catalog/table_version.py
```diff
@@ -659,10 +658,83 @@ def _add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) ->
...
+                raise excs.RequestError(
+                    excs.ErrorCode.UNSUPPORTED_OPERATION,
+                    'Cannot create an explicit B-tree index on a table with has_default_idxs=True; '
+                    'its eligible columns are indexed automatically.',
+                )
+            assert idx_col.name is not None, repr(idx_col)
+            if idx_col.tbl_handle.id != tbl_id:
```
mkornacker: Same here.

pixeltable/catalog/table_version.py
```diff
@@ -659,10 +658,83 @@ def _add_index(self, col: Column, idx_name: str | None, idx: index.IndexBase) ->
...
+        new_names: set[str] = set()
+        new_btree_col_names: set[str] = set()
+
+        for idx_col, idx_name, idx in idxs:
+            assert isinstance(idx_col, Column)
+
+            if idx_name is not None:
+                assert idx_name not in new_names, idx_name
+                existing_info = existing_by_name.get(idx_name)
+                if existing_info is not None:
```
mkornacker, line 712: Check this last.

mkornacker (approved):

pixeltable/catalog/local_table.py
mkornacker: Please add a test for add_btree_index() to test_snapshot.py.

pixeltable/catalog/local_table.py
sergey-mkhitaryan: if it's a pure snapshot, `tv` is the base table, and `tv.has_default_idxs` can return True whereas the snapshot of course cannot have any default indexes.

pixeltable/catalog/local_table.py
sergey-mkhitaryan: Done. `TestIndex.test_btree_index_on_view()` covers this.

sergey-mkhitaryan: moved the snapshot part of `test_btree_index_on_view()` to `TestSnapshot`

PR #1474, opened 2026-07-19
CLI commands: cwd/pwd

mkornacker:


pixeltable_cli/client/main.py
```diff
@@ -7,6 +7,8 @@
 # Each key names a module under pixeltable_cli.client.commands.* exposing run(argv: list[str]) -> None.
 COMMANDS: dict[str, str] = {
     'health': 'show daemon info',
+    'cwd': 'set or clear the working directory prepended to relative paths',
```
christopherpestano: Instead of `cwd`, wouldn't it be more conventional to call this `cd`? When I see `cwd` it normally refers to the actual directory itself not the act of switching dirs, llms + people may get confused by this.

pixeltable_cli/server/daemon.py
```diff
@@ -3,11 +3,33 @@
...
+
+
+def set_wd(session: str, uri: str) -> None:
+    with _wd_lock:
+        _wd[session] = uri
+
+
+def clear_wd(session: str) -> None:
+    with _wd_lock:
+        _wd.pop(session, None)
```
christopherpestano, line 31: The locking mechanism is unnecessary here, we aren't doing any read-modify-write here, the GIL already guarentees that these single dict ops won't result in inconsistent state.

pixeltable_cli/client/utils.py
```diff
@@ -10,14 +10,144 @@
...
+    """
+    if _IS_WINDOWS:
+        launcher_pid = os.getppid()
+        shell_pid = _win_parent_pid(launcher_pid)
+        # fall back to the trampoline if its parent can't be resolved (session won't persist, but is stable
+        # within a single command rather than crashing)
+        pid = shell_pid if shell_pid is not None else launcher_pid
+        return f'{pid}:{_win_process_creation_time(pid)}'
+    ppid = os.getppid()
+    return f'{ppid}:{_parent_start_time(ppid)}'
```
christopherpestano: Most of the bespoke logic here can be removed by using psutil, which would remove the bespoke windows logic and the subprocess.run which starts a whole new subprocess each time. Basically this entire block simplifies to the below.

```python
def session_key() -> str:
      """Stable per-terminal id: the invoking shell's pid + its creation time (creation time
      distinguishes a recycled pid). On Windows the console-script trampoline is our parent, so
      step up one more to the shell every command in the terminal shares."""
      proc = psutil.Process(os.getppid())
      if _IS_WINDOWS:
          proc = proc.parent() or proc
      return f'{proc.pid}:{proc.create_time()}'
```

pixeltable_cli/client/commands/mv.py
```diff
@@ -40,8 +40,8 @@ def run(argv: list[str]) -> None:
         print(f'would move {args.path} -> {dst}')
         return
```
christopherpestano: This is now broken by the cwd change, if the user attempts to do `pxt mv my_dir/my_table /` while cwd is set to `wd` the final destination of the table will be `/wd/my_table` and not `/my_table` due to the stripping logic here conflicting with the new path semantics.

pixeltable_cli/server/routes.py
```diff
@@ -349,7 +362,7 @@ def columns(req: Request) -> models.ColumnsResponse:
             excs.ErrorCode.INVALID_ARGUMENT, "'path' query parameter is empty; omit it entirely for all tables"
         )
     if path is not None:
-        _validate_path(path)
+        path = req.resolve_path(path)
     paths = _resolve_table_paths(path)
```
christopherpestano: Shouldn't `pxt columns` with no path specified give you the columns in the cwd rather than the root?

christopherpestano (approved):
Looks pretty good, I'll trust your judgment to decide how you want to resolve these comments, approved

pixeltable_cli/server/daemon.py
mkornacker: This was done intentionally (eventually we want to be able to run w/o the GIL). It's very little added complexity.

PR #1475, opened 2026-07-19
PXT-1227: View.compute()

mkornacker:
Also:
- moves Row from _query.py to a new row.py
- also adds row.py:RowBatch
- View.compute() does *not* populate RowBatch.index_values

Follow-ups:
- Table.insert(return_rows=True): change UpdateStatus.rows from dict[str, Any] to RowBatch
- View.compute() to populate RowBatch.index_values, if this turns out to be useful
- View.compute(select_items: dict[str, Expr]), if that turns out to be needed


pixeltable/catalog/globals.py
```diff
@@ -184,6 +184,25 @@ def validated(cls, name: str, error_prefix: str) -> MediaValidation:
             ) from None
 
 
+class OnErrorParameter(enum.Enum):
```
aaron-siegel: `Param` or `Parameter`? I'd prefer we be consistent.

[ok, I see this was just moved around, so this isn't new. It just made it very obvious to me since the next class in the file is `IfExistsParam`.]

pixeltable/row.py
```diff
@@ -0,0 +1,195 @@
...
+
+    The `errors` property holds error info (`{'errortype': ..., 'errormsg': ...}`) for each cell whose
+    evaluation failed; the `index_values` property holds the values of embedding indexes defined on the
+    row's table. Both are keyed by column or index name.
+    """
+
+    _data: tuple[Any, ...]
+    _columns: dict[str, int]
+    _col_types: dict[str, ColumnType]
+    _errors: dict[str, CellError]
```
aaron-siegel, line 31: not `_cellmd`?

pixeltable/row.py
```diff
@@ -0,0 +1,195 @@
...
+    _index_values: dict[str, Any]
+
+    def __init__(
+        self,
+        data: Iterable[Any],
+        columns: dict[str, int],
+        col_types: dict[str, ColumnType],
+        errors: dict[str, CellError] | None = None,
+        index_values: dict[str, Any] | None = None,
```
aaron-siegel, line 40: I thought we'd agreed index values wouldn't be separately included, and that if the user wants them, the way to retrieve them is by explicitly including `my_col.embedding()` in the select list?

pixeltable/row.py
```diff
@@ -0,0 +1,195 @@
...
+        return iter(self._columns)
+
+    def __contains__(self, key: object) -> bool:
+        return key in self._columns
+
+    def __len__(self) -> int:
+        return len(self._columns)
+
+    def __repr__(self) -> str:
+        return 'Row({' + ', '.join(f'{k!r}: {v!r}' for k, v in self.items()) + '})'
```
aaron-siegel: Alternatively: `f'Row({dict(self)})`?

pixeltable/catalog/globals.py

pixeltable/row.py
mkornacker: This is only for insert(); compute() doesn't return index values.

pixeltable/row.py
mkornacker: For now we only care about errors.

pixeltable/row.py
mkornacker: Done.

aaron-siegel (approved):

pixeltable/row.py
aaron-siegel: ok

PR #1477, opened 2026-07-20
nightly github action uploads test reports as github artifacts

sergey-mkhitaryan:
I want to build some kind of agentic ci monitoring, for which this will be useful.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -129,7 +129,14 @@ jobs:
       - name: Run the unit tests
         env:
           PXTTEST_CI_OS: ${{ matrix.os }}
-        run: pytest -v -m 'not very_expensive' tests
+        run: pytest -v -m 'not very_expensive' --junitxml=junit-report.xml tests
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -129,7 +129,14 @@ jobs:
       - name: Run the unit tests
         env:
           PXTTEST_CI_OS: ${{ matrix.os }}
-        run: pytest -v -m 'not very_expensive' tests
+        run: pytest -v -m 'not very_expensive' --junitxml=junit-report.xml tests
+      - name: Upload test report
```
aaron-siegel, line 133: Do we want it in the scheduled job in pytest.yml as well?

aaron-siegel (approved):

.github/workflows/nightly.yml
sergey-mkhitaryan: Let me try to do something with this one, and I'll expand to other jobs if I get it to work.

PR #1479, opened 2026-07-21
TableModel.diff_all() and TableModel.update_all()

aaron-siegel:
Introduces:
- `TableModel.diff_all()` that prints a list of differences between a `model_base` instance and the catalog
- `TableModel.update_all()` that synchronizes a `model_base` instance with the catalog

Calling `update_all()` by itself will only add columns/indices and generate an error if any are dropped. `update_all(allow_destructive=True)` will also delete columns/indices. Any alterations of existing column/table properties are currently treated as "FATAL" (must be synchronized manually); this will be improved as we add alter_table()/alter_column() functionality.

Also prohibits "anonymous" compound expressions in base queries in a view model.
- `base=MyTable.select(MyTable.col)` is ALLOWED
- `base=MyTable.select(named_col=(MyTable.col + 1))` is ALLOWED
- `base=MyTable.select(MyTable.col + 1)` is DISALLOWED


pixeltable/catalog/model.py
```diff
@@ -919,12 +1154,97 @@ def _create_all(binding_root: str = '') -> tuple[list[str], list[str]]:
...
+                    dropped_columns=list(r.dropped_columns),
+                    new_idxs={idx_name: model.__indexes__[idx_name] for idx_name in r.new_indexes},
+                    dropped_idxs=list(r.dropped_indexes),
+                )
+            )
+
+        # All models share `binding_root`, hence a single catalog; apply every table's changes in one transaction.
+        cat = get_runtime().get_catalog(updates[0]['path'])
+        cat.update_from_model(updates)
+
```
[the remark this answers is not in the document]
aaron-siegel: Done

pixeltable/catalog/catalog.py
```diff
@@ -1703,15 +1703,14 @@ def create_from_model(
...
         )
 
-        # If the table already exists, validate the model against it and rebind (the server enforces its own
-        # consistency; we never trust a client to have validated).
+        # If the table already exists, rebind to it. `create_all()` validates up front that every existing table
+        # already matches its model, so reaching here means there is nothing to create.
         existing = self.get_table(path, IfNotExistsParam.IGNORE)
         if existing is not None:
-            self._validate_model(existing, display_name, base, iterator)
             return existing, False
```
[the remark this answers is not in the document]
aaron-siegel: Done

pixeltable/catalog/catalog.py
```diff
@@ -1748,36 +1747,42 @@ def create_from_model(
...
 
+            # First drop any columns or indices that are being removed. Do this in *inverse* declaration order so that
+            # dependent columns or indices are dropped first.
+            for _, tbl_version, update in tbl_info[::-1]:
+                for col_name in update['dropped_columns']:
+                    col = tbl_version.cols_by_name[col_name]
+                    tbl_version.drop_column(col)
+                for idx_name in update['dropped_idxs']:
+                    idx = tbl_version.idxs_by_name[idx_name]
+                    tbl_version.drop_index(idx.id)
```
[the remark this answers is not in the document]

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+                )
+        catalog_col = catalog.Column.create(name, subst_spec)
+        catalog_col.tbl_handle = tbl_handle
+        catalog_col.id = next(next_col_id)
+        additional_cols.append(catalog_col)
+        visible_cols[name] = catalog_col
+        subst_dict[ModelColumnRef(name, catalog_col.col_type)] = exprs.ColumnRef(
+            catalog_col.column_version_md(),
+            perform_validation=subst_spec.get('media_validation', str(tvp.media_validation).lower()) == 'on_read',
+        )
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/catalog/model.py
```diff
@@ -104,6 +105,27 @@ class EmbeddingIndex:
     metric: Literal['cosine', 'ip', 'l2'] = 'cosine'
     precision: Literal['fp16', 'fp32'] = 'fp16'
 
+    def __repr__(self) -> str:
```
mkornacker, line 107: Would be good to get some guidance in a (file-level) docstring regarding which classes need repr(), etc. There seems to be some implicit contract.

pixeltable/catalog/catalog.py
```diff
@@ -1748,36 +1746,65 @@ def create_from_model(
...
+                    # catalog corruption.
+                    dependent_user_cols = [
+                        c for c in self.get_column_dependents(tbl_version.id, col.id) if c.name is not None
+                    ]
+                    if len(dependent_user_cols) > 0:
+                        raise excs.RequestError(
+                            excs.ErrorCode.UNSUPPORTED_OPERATION,
+                            f'Cannot drop column {col.name!r} because the following columns depend on it:\n'
+                            f'{", ".join(c.name for c in dependent_user_cols)}',
+                        )
```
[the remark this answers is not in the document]
aaron-siegel: This is unnecessary; order-dependent is fine; that's why we drop them in reverse declaration order.

pixeltable/catalog/catalog.py
```diff
@@ -1748,36 +1746,65 @@ def create_from_model(
...
+                    tbl_version.drop_column(col)
+
+                for idx_name in update['dropped_idxs']:
+                    idx_info = tbl_version.idxs_by_name[idx_name]
+                    val_col = idx_info.val_col
+                    col_dependents = get_runtime().catalog.get_column_dependents(val_col.get_tbl().id, val_col.id)
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/catalog/model.py
```diff
@@ -527,6 +553,17 @@ def __prepare__(  # type: ignore[override]
...
                     )
                 assert isinstance(base, ModelQuery)
+                if base.select_clause is not None:
+                    # Validate the select list.
+                    items, _ = base.select_clause
+                    for item in items:
+                        if not isinstance(item, ModelColumnRef):
+                            raise excs.RequestError(
+                                excs.ErrorCode.INVALID_ARGUMENT,
+                                f'{display_name}: `base` select() list may contain only direct column references '
```
mkornacker, line 562: Something to consider: only allow col refs here and force all computed cols into the class body itself. Would probably improve clarity and remove a source of syntactic variation that doesn't have any functional benefit and will probably just provide another opportunity for agent confusion.

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
     return iterator, additional_cols, resolved_idxs
 
 
+class Updates(TypedDict):
```
mkornacker: SchemaChange/-Update or something like that? Updates is very generic (and 'table update' already has a different connotation).

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+    new_idxs: dict[str, EmbeddingIndex]
+    dropped_idxs: list[str]
+
+
+def prepare_model_updates(
+    tvp: catalog.TableVersionPath,
+    display_name: str,
+    new_columns: dict[str, ColumnSpec],
+    new_idxs: dict[str, EmbeddingIndex],
+) -> tuple[list[catalog.Column], list[tuple[catalog.Column, str | None, index.IndexBase]]]:
```
mkornacker, line 951: docstring would be nice.

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+    for col in tvp.columns():
+        visible_cols[col.name] = col
+        subst_dict[ModelColumnRef(col.name)] = exprs.ColumnRef(
+            col.column_version_md(), perform_validation=(col.media_validation == MediaValidation.ON_READ)
+        )
+
+    tbl_handle = tvp.tbl_version
+    next_col_id = itertools.count(start=tvp.tbl_version.get().next_col_id())
+
+    # Process any additional columns specified in the view model body.
```
mkornacker: How do we know it's a view?

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+    additional_cols: list[catalog.Column] = []
+    for name, spec in new_columns.items():
+        subst_spec = spec.copy()
```
mkornacker: resolved_spec would be easier to follow (I initially thought this had something to do with the actual substitution dict, rather than being the resolution end product).

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+    dropped_idxs: list[str]
+
+
+def prepare_model_updates(
+    tvp: catalog.TableVersionPath,
+    display_name: str,
+    new_columns: dict[str, ColumnSpec],
+    new_idxs: dict[str, EmbeddingIndex],
+) -> tuple[list[catalog.Column], list[tuple[catalog.Column, str | None, index.IndexBase]]]:
+    visible_cols: dict[str, catalog.Column] = {}
```
mkornacker: user_cols is more appropriate.

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+        catalog_col.id = next(next_col_id)
+        additional_cols.append(catalog_col)
+        visible_cols[name] = catalog_col
+        subst_dict[ModelColumnRef(name, catalog_col.col_type)] = exprs.ColumnRef(
+            catalog_col.column_version_md(),
+            perform_validation=subst_spec.get('media_validation', str(tvp.media_validation).lower()) == 'on_read',
+        )
+
+    # Resolve each declared embedding index against the model's visible columns.
+    resolved_idxs: list[tuple[catalog.Column, str | None, index.IndexBase]] = []
```
mkornacker, line 1017: What are the tuple fields?

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+    for idx_name, idx_spec in new_idxs.items():
+        if not isinstance(idx_spec.column, ModelColumnRef):
```
mkornacker, line 1019: Comment: what else could it be?

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+            video_embed=idx_spec.video_embed,
+            document_embed=idx_spec.document_embed,
+            column=visible_cols[col_name],
+        )
+        resolved_idxs.append((visible_cols[col_name], idx_name, idx))
+
+    return additional_cols, resolved_idxs
+
+
+class ValidationResults(NamedTuple):
```
mkornacker: This would benefit from a docstring regarding scope: what does it need to capture?

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+    model_cls: TableModelMeta
+    tbl_exists: bool
+    model_kind: str  # 'table' or 'view'
+    existing_kind: str | None  # None if the table does not yet exist
+    model_iterator: str | None  # display string of the model's iterator call, or None if it has no iterator
```
mkornacker: iterator_call? This sounds like it's only the iterator name.

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+            or len(self.altered_columns) > 0
+            or len(self.altered_properties) > 0
+        )
+
+    @property
+    def has_destructive_changes(self) -> bool:
+        return self.tbl_exists and (len(self.dropped_columns) > 0 or len(self.dropped_indexes) > 0)
+
+    @property
+    def has_benign_changes(self) -> bool:
```
mkornacker: 'additive' is the term used in other db migration frameworks.

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+    altered_columns: list[tuple[str, dict[str, Any], dict[str, Any]]]
+    # Table-level properties that differ, as (property name, model value, existing value). Fatal for now.
+    altered_properties: list[tuple[str, Any, Any]]
+
+    @property
+    def has_changes(self) -> bool:
+        return self.has_fatal_changes or self.has_destructive_changes or self.has_benign_changes
+
+    @property
+    def has_fatal_changes(self) -> bool:
```
mkornacker: 'unsupported' is the term used in other db migration frameworks.

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+        'media_validation': (spec.get('media_validation') or default_media_validation)
+        if col_type.is_media_type()
+        else None,
+        'comment': comment if comment else None,
+        'custom_metadata': spec.get('custom_metadata'),
+        'destination': str(spec['destination']) if spec.get('destination') is not None else None,
+    }
+
+
+def _existing_column_properties(col_md: ColumnMetadata) -> dict[str, Any]:
```
mkornacker: Given that this is for existing columns, can't we use the actual Column instance for the comparison? Comparing strings seems fragile.

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+    @property
+    def has_destructive_changes(self) -> bool:
+        return self.tbl_exists and (len(self.dropped_columns) > 0 or len(self.dropped_indexes) > 0)
+
+    @property
+    def has_benign_changes(self) -> bool:
+        return not self.tbl_exists or len(self.new_columns) > 0 or len(self.new_indexes) > 0
+
+
+def _model_column_properties(spec: ColumnSpec, default_media_validation: str) -> dict[str, Any]:
```
mkornacker: I'd find that easier to follow  with a dataclass ColumnProperties with constructors from_spec(ColumnSpec) and from_md(ColumnMd), and that way you'd get some type safety as well. I don't think this would be more code, aside from the dataclass declaration (and that would actually be helpful to see).

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+        'primary_key': col_md['is_primary_key'],
+        'stored': col_md['is_stored'],
+        'media_validation': col_md['media_validation'],
+        'comment': col_md['comment'],
+        'custom_metadata': col_md['custom_metadata'],
+        'destination': col_md['destination'],
+    }
+
+
+def _model_table_properties(model: TableModelMeta) -> dict[str, Any]:
```
mkornacker: Same here.

pixeltable/catalog/model.py
```diff
@@ -905,6 +931,358 @@ def prepare_model(
...
+            model_sample=model_sample,
+            existing_sample=existing_md['view_sample'],
+            altered_columns=altered_columns,
+            altered_properties=altered_properties,
+        )
+
+    return results
+
+
+def _format_diff(name: str, r: ValidationResults) -> list[str]:
```
mkornacker: This would be substantially more useful if it produced a list of typeddicts, and we then layer the "human-friendly" output on top of that.

Claude's thoughts on this:
 Agent-friendly? Not really — and the return type is why

  For a human reading a notebook/CLI it's fine: clear verbs (CREATED/ADDED/DROPPED), severity tags ((FATAL)), indented grouping. But for an agent driving this
  (your "agents spin up services" case), three problems:

  1. diff_all() returns nothing. An agent's natural loop is: define models → diff_all() to see what would change → decide → update_all() or
  update_all(allow_destructive=True). With a None return it has to capture console/log text and scrape it. That's the dealbreaker, independent of formatting.
  2. The text is free-form and mixes phrasings. "kind mismatch (FATAL):", "the following columns have altered properties (FATAL):", "{col} {prop}: model=..., 
  existing=...", "... will be ADDED:" — no stable grammar; parsing needs brittle regex.
  3. It dumps Python reprs into the prose — f'{col_name!r} = {model_column_specs[col_name]}' renders a ColumnSpec dict, and indexes render an EmbeddingIndex repr.
  Noisy for humans, unparseable for agents.
  
  And the decision an agent actually needs to make — is this change additive / destructive / unsupported — exists structurally on ValidationResults
  (has_fatal_changes / has_destructive_changes / has_benign_changes), but _format_diff flattens it into "(FATAL)" substrings.

pixeltable/catalog/model.py
aaron-siegel: This is a bigger change and I'd argue we'd want to couple it with changes to create_view() so that the interpretation of base queries is aligned between the two modes. We should also maybe discuss with the team whether this is truly the behavior we want.

Let's make this a follow-up? In the meantime I doubt many people will use or even know about this feature.

pixeltable/catalog/model.py
aaron-siegel: I think there's an implicit language-level contract that all Python classes should have useful reprs :)

I'm not sure what specific guidance you're asking for here.

pixeltable/catalog/model.py
aaron-siegel: Changed to `CatalogUpdates` (to signal this is a backend/proxy interface class, as distinguished from `TableDiff`/`SchemaChange`).

pixeltable/catalog/model.py
aaron-siegel: Done.

pixeltable/catalog/model.py
aaron-siegel: Done.

pixeltable/catalog/model.py
aaron-siegel: Clarified.

pixeltable/catalog/model.py
aaron-siegel: Done. Also changed `additional_cols` to `resolved_cols` so everything is using a consistent nomenclature.

pixeltable/catalog/model.py
aaron-siegel: Same as everywhere else that indexes interact with the catalog. I don't think it's fruitful or necessary to repeat the meaning of this tuple everywhere it appears - since it's common to a lot of the code base. (I'm just piggybacking off `pixeltable.index` here.)

It should probably be made into a `NamedTuple` but that's outside the scope of this PR.

pixeltable/catalog/model.py
aaron-siegel: Nothing else is valid but it's coming from the user, so there could be some random crap there (and so an exception is more appropriate than an assertion). I'm not sure that needs a code comment?

pixeltable/catalog/model.py
aaron-siegel: Replaced it with `TableDiff`/`SchemaChange` per your suggestion.

pixeltable/catalog/model.py
aaron-siegel: I agree comparing strings isn't ideal, but
1. This is being done client side (where the models live) and Column instances aren't proxyable;
2. I'm striving to operate off the *Metadata structures where feasible, as we've discussed;
3. I think a better solution than going off Column would be to introduce more structure to ColumnMetadata - i.e. return structured, publicly consumable expressions rather than strings. That would also be more friendly to users who want to programmatically inspect their catalog structure.

For now, I don't think there's much harm in using strings; it ensures that identical expressions yield a match, which is the main goal here. If there are cases where our expressions have inaccurate repr's, that's a bug we need to fix anyway.

pixeltable/catalog/model.py
aaron-siegel: In this case, everything being compared is using its original representation, so I don't see an issue.

pixeltable/catalog/model.py
aaron-siegel: Done, via separate `get_model_diff()` and `diff_all()` as we discussed.

To be clear, I had always envisioned a separate code path for json output; I certainly did not envision anyone trying to parse the diff_all() strings.

pixeltable/catalog/model.py
```diff
@@ -1005,57 +1012,57 @@ def prepare_model_updates(
...
+
+
+class TableDiff(TypedDict):
+    """How one model differs from its catalog table."""
+
+    path: str  # catalog path of the table
+    model_cls: str  # model class name, so an agent can map back to code
+    kind: Literal['table', 'view']
+    exists: bool
+    resolution: Literal['up_to_date', 'create', 'update_additive', 'update_destructive', 'unsupported']
```
mkornacker: typedef that literal somewhere?

pixeltable/catalog/model.py
mkornacker: I still think this would make the code easier to follow.

pixeltable/catalog/model.py
```diff
@@ -1130,45 +1137,66 @@ def _visible_columns(model: TableModelMeta) -> dict[str, ColumnSpec]:
...
     Analyze each registered model against the current catalog state, summarizing the schema changes that creating
     the models would entail, along with any incompatibilities with an already-existing table of the same name.
     This is purely informational: it neither modifies the catalog nor raises on incompatibilities.
     """
     binding_root = TableModelMeta._normalize_binding_root(binding_root)
-    results: dict[str, ValidationResults] = {}
+    results: dict[str, TableDiff] = {}
 
     for name, model in registered_models.items():
         visible_columns = _visible_columns(model)
```
mkornacker: Can we stick with 'user columns'?

pixeltable/catalog/model.py
```diff
@@ -1184,103 +1212,185 @@ def validate_models(registered_models: dict[str, TableModelMeta], binding_root:
             idx_name for idx_name, info in existing_md['indices'].items() if info['index_type'] == 'embedding'
         }
 
-        # For columns present in both, flag any whose properties differ.
-        altered_columns: list[tuple[str, dict[str, Any], dict[str, Any]]] = []
+        changes = []
```
mkornacker, line 1281: type decl

tests/test_table_model.py
```diff
@@ -584,6 +584,592 @@ class ExampleViewModelFromQuery(
...
+
+        # Created as a view; V2 redeclares it as a table, producing a kind mismatch.
+        class ExampleKind(TableModel, name='test_kind', base=ExampleTable):
+            kc1 = ExampleTable.value + 1
+            kc2 = ExampleTable.value + 2
+
+        TableModel.create_all(root)
+
+        # Re-diffing the original models reports no differences (in particular, the view's iterator round-trips).
+        with capture_console_output() as out:
```
mkornacker, line 642: missing: get_diff()

mkornacker (approved):

pixeltable/catalog/model.py
```diff
@@ -919,12 +1433,109 @@ def _create_all(binding_root: str = '') -> tuple[list[str], list[str]]:
...
+
+        else:
+            binding_root = TableModelMeta._normalize_binding_root(binding_root)
+            updates: list[CatalogUpdates] = []
+            for name, d in to_update:
+                model = registered_models[name]
+                new_col_names = [c['name'] for c in d['changes'] if c['target'] == 'column' and c['op'] == 'add']
+                dropped_col_names = [c['name'] for c in d['changes'] if c['target'] == 'column' and c['op'] == 'drop']
+                new_idx_names = [c['name'] for c in d['changes'] if c['target'] == 'index' and c['op'] == 'add']
+                dropped_idx_names = [c['name'] for c in d['changes'] if c['target'] == 'index' and c['op'] == 'drop']
```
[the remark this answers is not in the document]
aaron-siegel: Done.

pixeltable/catalog/model.py
```diff
@@ -872,7 +897,7 @@ def prepare_model(
         visible_cols[name] = catalog_col
         subst_dict[ModelColumnRef(name, catalog_col.col_type)] = exprs.ColumnRef(
             catalog_col.column_version_md(),
-            perform_validation=subst_spec.get('media_validation', media_validation) == 'on_read',
+            perform_validation=subst_spec.get('media_validation', media_validation.name.lower()) == 'on_read',
         )
```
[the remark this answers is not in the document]
aaron-siegel: Done.

pixeltable/catalog/model.py
```diff
@@ -919,12 +1433,109 @@ def _create_all(binding_root: str = '') -> tuple[list[str], list[str]]:
...
+                dropped_idx_names = [c['name'] for c in d['changes'] if c['target'] == 'index' and c['op'] == 'drop']
+                # Resolve `type` annotations to ColumnTypes, mirroring `_create()`.
+                new_columns: dict[str, ColumnSpec] = {}
+                for col_name in new_col_names:
+                    spec = model.__columns__[col_name].copy()
+                    if 'type' in spec:
+                        spec['type'] = ts.ColumnType.normalize_type(  # type: ignore[typeddict-item]
+                            spec['type'], nullable_default=True, allow_builtin_types=False
+                        )
+                    new_columns[col_name] = spec
```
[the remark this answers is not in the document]
aaron-siegel: Done.

pixeltable/catalog/model.py
aaron-siegel: Done.

pixeltable/catalog/model.py
aaron-siegel: It's already defined above, so attempting to put an annotation here triggers a mypy error

I reorganized things a bit; see if you're ok with the result

tests/test_table_model.py
aaron-siegel: Done

pixeltable/catalog/model.py
aaron-siegel: Done

pixeltable/catalog/model.py
```diff
@@ -919,12 +1479,118 @@ def _create_all(binding_root: str = '') -> tuple[list[str], list[str]]:
...
+            )
+
+        # Apply column/index changes to existing tables. Brand-new tables are handled by `_create_all()` below.
+        to_update = [
+            (name, d) for name, d in diffs.items() if d['resolution'] in ('update_additive', 'update_destructive')
+        ]
+
+        if len(to_update) == 0:
+            Env.get().console_logger.info('Catalog is up to date.')
+
```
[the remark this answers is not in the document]
aaron-siegel: fixed. (The suggestion was not the correct fix, but fixed.)

PR #1480, opened 2026-07-21
Fix scheduled test failures

christopherpestano:
Fixes the failures from the July 20 nightly and July 21 scheduled test runs. The pgvector 0.5.0 migration originally bundled here has been split into its own PR.

Google removed the Veo 2.0 and 3.0 models from the Gemini API. The video generation test, the Veo notebooks, and the docstring examples now use veo-3.1-fast-generate-preview.

Together AI image generation failed on a provider-side 500. InternalServerError is added to the retryable network error patterns, so tests marked rerun-on-network-error retry it.

The audio-podcast-chapters notebook is skipped because it needs whisperx, which is disabled in the dev environment (whisperx hard-pins torch 2.8).

The data-import-huggingface notebook uses the built-in string len() instead of apply(), which can no longer be used in stored computed columns.

The nightly and pytest workflows now start an Ollama server if one is not already running. The nightly workflow also removes the local pixeltable_cli source tree alongside pixeltable so tests run against the installed package.

Bumps av to 17.


pixeltable/exec/sql_node.py
```diff
@@ -474,9 +473,10 @@ async def __aiter__(self) -> AsyncIterator[DataRowBatch]:
                             output_row[slot_idx] = sql_row[i]
                     else:
                         raise RuntimeError(f'Unexpected datetime value for {e}')
-                elif isinstance(sql_row[i], HalfVector):
-                    # All array data needs to be materialized as ndarrays
-                    output_row[slot_idx] = sql_row[i].to_numpy().astype(np.float32)
+                elif e.col_type.is_array_type() and isinstance(sql_row[i], list):
+                    # pgvector VECTOR/HALFVEC values decode to lists of floats; all array data
+                    # needs to be materialized as ndarrays
+                    output_row[slot_idx] = np.asarray(sql_row[i], dtype=np.float32)
```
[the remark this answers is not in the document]
christopherpestano: This keeps the same behavior as pgvector 0.4.x: vector results were cast to float32 by pgvector itself, and we cast halfvec results to float32 here. The stored values are fp32 or fp16, so casting to a declared float64 would invent precision that was never stored. Changing the dtype handling would be a behavior change that's out of scope for this PR.

pyproject.toml
```diff
@@ -52,8 +52,8 @@ dependencies = [
     "fasteners>=0.19",
     "psycopg[binary]>=3.1.18",
     "sqlalchemy>=2.0.23",
-    "pgvector>=0.4.0,<0.5.0",
-    "av>=10.0.0",
+    "pgvector>=0.5.0",
```
aaron-siegel: Is the change not backward compatible with pgvector < 0.5? Ideally we'd make the change in a backward compatible way.

pyproject.toml
```diff
@@ -52,8 +52,8 @@ dependencies = [
...
+    "av>=17.1.0",
```
aaron-siegel: What's the need for this? It's way too restrictive for our required dependencies.

aaron-siegel (changes requested):
THANK YOU for taking this on!

pyproject.toml
christopherpestano: Ended up being easier than I thought to make the change backwards compatible, the import I had changed to was already backwards compatible so the only logic that needed to be tweaked was the sql_node handling of HalfVector. Moving forward though `make check` needs to be done with pgvector 0.5.0 though since they added type hints to the library.

pyproject.toml
christopherpestano: Bumped av back down, I wanted to do a cleaner fix for one of our failing nightly tests but I guess it's not worth the version bump, made a minor change to test_video.py instead.

tests/test_video.py
```diff
@@ -241,6 +241,8 @@ def test_get_metadata(self, uses_db: None) -> None:
         base_t.add_computed_column(metadata=base_t.video.get_metadata())
         validate_update_status(base_t.insert({'video': p} for p in video_filepaths), expected_rows=len(video_filepaths))
         result = base_t.where(base_t.metadata.size == 2234371).select(base_t.metadata).collect()['metadata'][0]
+        # the mp4 stream metadata contains a vendor_id entry with av 15 but not with av 17
+        result['streams'][0]['metadata'].pop('vendor_id', None)
```
aaron-siegel: Just to clarify: it's fine if our TESTS require av >= 17. Just not ok to put it as a constraint in the required dependencies. But our tests depend only on the dev environment.

I'd prefer this:
- Put 'vendor_id' back in
- Keep av locked at 17.1 in our uv.lock
- But revert the required dependency to its old value (10 or 11 or whatever it was before)

We generally don't need our tests to be resilient against wide version ranges (and shouldn't unless we're specifically trying to validate version range compatibility in some way); it just complicates the tests unnecessarily.

aaron-siegel (changes requested):

tests/test_video.py
christopherpestano: Hmm, this feels like a bad practice to me. If we advertise support for a broad version range but our test suite only passes on a narrow subset of it, that's a bit misleading. But the change here is minor (a single field removed from the expected metadata), so I'll defer and make it.

.github/workflows/nightly.yml
```diff
@@ -117,7 +117,10 @@ jobs:
         # cmake is needed for certain CI environments where it is not preinstalled.
         run: ./scripts/retry.sh 3 30 conda install -y -c conda-forge 'cmake>=3.22' libiconv 'ffmpeg==6.1.1=gpl*'
       - name: Delete source tree
-        run: rm -rf pixeltable  # Ensure we use pypi pixeltable package for testing, not local source tree
+        # Ensure we use the pypi pixeltable package for testing, not the local source tree. pixeltable_cli
```
sergey-mkhitaryan, line 120: too many words, please just say that this ensures we use pypi packages for testing and not the source.

.github/workflows/nightly.yml
```diff
@@ -189,6 +192,9 @@ jobs:
           ollama serve &  # Start the ollama server if it didn't start via install.sh
           sleep 10  # Wait for the ollama server to start
           # Pull the models used for testing now, with retries, to safeguard against connection issues
+          if ! curl -fsS --max-time 2 http://127.0.0.1:11434 > /dev/null 2>&1; then
```
sergey-mkhitaryan, line 195: can you write a brief comment explaining what this does?

scripts/prepare-nb-tests.sh
```diff
@@ -2,6 +2,7 @@
 
 # Notebooks that are always skipped
 SKIP_NOTEBOOKS=(
+    audio-podcast-chapters          # Uses whisperx, which is disabled in our dev env (hard-pins torch 2.8)
```
sergey-mkhitaryan, line 5: is this forever, or is there a path forward to re-enabling this nb at some point? if the latter, please file a jira.

tests/utils.py
```diff
@@ -1002,6 +1002,7 @@ def rerun(**kwargs: Any) -> Callable:
     'timed out',
     'Timeout',
     'ExternalServiceError',
+    'InternalServerError',
```
sergey-mkhitaryan, line 1005: when and how does this come up as a network error? I'm concerned that this will mask actual errors that are meant to be caught and not retried.

sergey-mkhitaryan (changes requested):

PR #1481, opened 2026-07-21
[PXT-1236] Evict LRU entries from Catalog cache

sergey-mkhitaryan:
If the cache grows too large, Catalog evicts the LRU entries at the end of the transaction. It deliberately avoids evicting in the middle of the transaction -- that would be disruptive to the transaction state.

This change includes a number of fixes for issues that I found by aggressively lowering the new cache limit:
* `TableVersionPath._cached_tv()` should refresh the entire path when reading a table. Otherwise, if this table is a snapshot, the transaction will trip an assertion in Catalog.
* proxy dispatch: wrap `cat.get_table_by_id()` in a retry loop instead of a transaction. The table is not guaranteed to be in the cache.
* `PixeltableWarning` is now re-raised form `Catalog._finalize_pending_ops()` instead of causing the test to get stuck retrying it indefinitely.


pixeltable/catalog/catalog.py
```diff
@@ -209,8 +214,8 @@ class Catalog(CatalogBase):
     # cached TableVersion instances; key: [id, version]
     # - mutable version of a table: version == None (even though TableVersion.version is set correctly)
     # - snapshot versions: records the version of the snapshot
-    _tbl_versions: dict[TableVersionKey, TableVersion]
-    _tbls: dict[tuple[UUID, int | None], LocalTable]
+    _tbl_versions: OrderedDict[TableVersionKey, TableVersion]
```
mkornacker, line 232: Would be good to point or here or somewhere else that all get() on _tbls and _tbl_versions also require a move_to_end() for the LRU update.

mkornacker (approved):

PR #1483, opened 2026-07-22
Proxy mode test infra improvements

sergey-mkhitaryan:
Almost all changes are test-only so low risk.

1. In proxy mode, the daemon uses the same database as the pytest process. This allows `_validate_catalog_state()` (that runs in the pytest process at the end of the test) to actually verify the database that was used in the test.
2. `Catalog.validate_store()` ignores Pixeltable warnings.
3. `TestIndex.test_embedding_basic` does not attempt to revert a dropped index because that causes catalog corruption (https://pixeltable.atlassian.net/browse/).
4. `TestEnvReset` restores the `Env` state at the end.
5. `TestCatalog.test_dropped_connection` only drops the current process's connections to the store rather than all, which may include proxy daemon connections. To make that happen, each DB connection gets an "application name" derived from the pid.
6. `Env._clean_up()` does not forcibly terminate all connections to the database -- that was screwing with the daemon process. There was no good reason to do that to begin with -- `_clean_up()` does not drop the database (and if it did, "drop with force" takes care of active connections).
7. At the end of session-scoped `proxy_daemon_db` fixture, stop the daemon rather than delete it. Deleting it causes its logs to disappear, and CI run cannot retain them as artifacts.
8. Updated `pytest.yml` to correctly produce test artifacts (including daemon logs) on all platforms.

### DIFF
--- a/pixeltable/env.py
+++ b/pixeltable/env.py
@@ -534,24 +534,13 @@ class Env:
         finally:
             engine.dispose()
 
-    def _pgserver_terminate_connections_stmt(self) -> str:
-        return f"""
-                SELECT pg_terminate_backend(pg_stat_activity.pid)
-                FROM pg_stat_activity
-                WHERE pg_stat_activity.datname = '{self._db_name}'
-                AND pid <> pg_backend_pid()
-            """
-
     def _drop_store_db(self) -> None:
         assert self._db_name is not None
         engine = sql.create_engine(self._dbms.default_system_db_url(), future=True, isolation_level='AUTOCOMMIT')
         preparer = engine.dialect.identifier_preparer
         try:
             with engine.begin() as conn:
-                # terminate active connections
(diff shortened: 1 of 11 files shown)


pixeltable/env.py
```diff
@@ -48,6 +48,15 @@
 T = TypeVar('T')
 
 
+def store_app_name() -> str:
+    """The application_name that this process's connections report to the store.
```
mkornacker, line 52: What does application name mean?

mkornacker (approved):

pixeltable/env.py
sergey-mkhitaryan: `application_name` is a PG command line parameter which labels connections in `pg_stat_activity`. I can rename this function (in a follow up pr) to something more abstract like "store connection label" if you'd like.

PR #1490, opened 2026-07-27
TableModes.update_all() bug fixes

mkornacker:
## update_all()

- Apply all table changes in a single write transaction, ordered bases-first
- Compare-and-swap on table id + schema versions (self and bases); abort with
ConcurrencyError if the catalog moved since the diff was computed
- Capture the CAS baseline in the same transaction that computes the diff, so
the versions provably match what was diffed
- Validate drops against the full change set (dependents dropped in the same
call no longer block each other)
- Reject drops of columns a view predicate depends on
- Add new columns in declaration order, so one may reference an earlier one

## View updates

- Fix: updating a base column that a view predicate references never added
newly-qualifying rows; the update now also runs the load plan, anti-joined
against rows already present
- Fix: component-view rows were orphaned when the base was itself a view
- Fix: unreferenced versioned ancestors dropped their version predicate from
the query's FROM clause

## Column resolution

- New ColumnRefByName expr: model-declared exprs bind by name and resolve once
ids are assigned, on both the create and update paths
- Iterator/view columns now shadow same-named base columns instead of being
renamed; a collision within one view is an error
- Fix: a shadowing column and the base column it shadows collided as dict keys


pixeltable/catalog/catalog.py
```diff
@@ -1514,7 +1518,7 @@ def _create_table(
         media_validation: MediaValidation,
         create_default_idxs: bool,
         is_versioned: bool,
-        additional_idxs: list[tuple[Column, str | None, 'index.IndexBase']] | None = None,
+        additional_idxs: list[tuple[str, str | None, 'index.IndexBase']] | None = None,
```
sergey-mkhitaryan: here and below: please clarify somewhere what those tuple components are. it was already confusing before your change, now even more so.

pixeltable/catalog/catalog.py
```diff
@@ -1619,7 +1623,7 @@ def _create_view(
         custom_metadata: Any,
         media_validation: MediaValidation,
         if_exists: IfExistsParam,
-        additional_idxs: list[tuple[Column, str | None, 'index.IndexBase']] | None = None,
+        additional_idxs: list[tuple[str, str | None, 'index.IndexBase']] | None = None,
```
sergey-mkhitaryan: we normally allow a view to declare an index on its base table's column. but this change seems to prohibit that if shadowing takes places (the target column can only be specified by name, therefore no way to tell which of the multiple homonymous cols is meant). it is intended?

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
                 explicit_tbl_id=tbl_id,
             )
 
-    def update_from_model(self, updates: list[model.CatalogUpdates]) -> None:
+    def update_from_model(self, schema_changes: list[model.TableSchemaChange]) -> None:
         """Update a table or view from a declarative model.
```
sergey-mkhitaryan: please update this comment: "update tables and/or views", not "a table or view". plural.

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
 
         If the table does not exist, raises NotFoundError. If the model is incompatible with the existing table,
         raises RequestError.
         """
-        tbls = [self.get_table(update['path'], IfNotExistsParam.ERROR) for update in updates]
+        # fault point:
+        # - the diff that produced updates was computed in an earlier read transaction
```
sergey-mkhitaryan, line 1788: I don't think that we should have comments justifying fault locations. I can just find the test that uses it if I need to know.

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
+        tbls = [self.get_table(schema_change['path'], IfNotExistsParam.ERROR) for schema_change in schema_changes]
+
+        # A table dropped and recreated at the same path since the diff has a new tbl_id; its schema versions are not
+        # comparable to the snapshotted ones, so reject rather than apply the diff to a different table.
+        for tbl, schema_change in zip(tbls, schema_changes):
+            if tbl._tbl_version_path.tbl_id != schema_change['expected_tbl_id']:
+                raise excs.ConcurrencyError(
+                    excs.ErrorCode.CONCURRENT_MODIFICATION,
+                    f'Table {str(schema_change["path"])!r} was replaced since update_all() computed its changes; '
+                    're-run update_all().',
```
sergey-mkhitaryan: add "please" re-run

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
+                )
```
sergey-mkhitaryan: this should be inside `update_fn()`, otherwise the validation is not atomic with the update.

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
 
         @retry_loop(for_write=True, write_tvps=[tbl._tbl_version_path for tbl in tbls], lock_mutable_tree=True)
         def update_fn() -> None:
-            # (tbl_version_path, tbl_version, updates) tuple for each table in the model update
+            # (tbl_version_path, tbl_version, TableSchemaChange) tuple for each table in the model update
```
sergey-mkhitaryan, line 1816: worth adding that this list is in base-tables-first order. that's important for the loop where the changes are applied.

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
-                dropped_cols = sorted(
-                    (tbl_version.cols_by_name[name] for name in update['dropped_columns']), key=lambda c: -c.id
+            # Validate all drops against the current catalog, before making any schema changes: the validation sees a
+            # consistent state and drop order (within and across tables) is irrelevant. A column whose only remaining
+            # dependents are themselves being dropped is allowed.
+            dropped_col_set: set[Column] = set()
+            for _, tv, schema_change in tbl_info:
+                dropped_col_set.update(tv.cols_by_name[name] for name in schema_change['dropped_columns'])
+
+            def check_column_dependents(tbl_id: UUID, col_id: int, kind: str, name: str) -> None:
```
aaron-siegel: `kind: Literal['column', 'index']` would be clearer here. (I initially interpreted it as table/view until I read the code)

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
+                remaining_dependents = [
+                    c
+                    for c in self.get_column_dependents(tbl_id, col_id)
+                    if c.name is not None and c not in dropped_col_set
+                ]
+                if len(remaining_dependents) > 0:
+                    raise excs.RequestError(
+                        excs.ErrorCode.UNSUPPORTED_OPERATION,
+                        f'Cannot drop {kind} {name!r} because the following columns depend on it:\n'
```
aaron-siegel: I think this error message might be confusing here; something like "{kind} {name!r} was removed from model `{...}`, but cannot be dropped because other columns depend on it:\n{...}\nDrop these columns first, or remove them from their associated models" would be clearer

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
+                    check_column_dependents(val_col.get_tbl().id, val_col.id, 'index', idx_name)
+                for name in schema_change['dropped_columns']:
+                    col = tv.cols_by_name[name]
+                    check_column_dependents(tv.id, col.id, 'column', name)
+
+            # check for dependent view predicates; a view can only reference columns on its base path, so the
+            # candidates are the mutable views of the updated tables (deduplicated: an updated table's views are also
+            # the views of any updated ancestor)
+            mutable_views = {view_tv.id: view_tv for _, tv, _ in tbl_info for view_tv in self._mutable_view_tvs(tv)}
+            views_by_qid = {
```
sergey-mkhitaryan: I don't know if it matters, but if two views depend on one column, one of them will overwrite the other in this dict. (you mention the error message being deterministic below)

pixeltable/catalog/model.py
```diff
@@ -1227,173 +1180,193 @@ def _add_index_change(idx_name: str, idx: EmbeddingIndex) -> SchemaChange:
     )
 
 
-def validate_models(registered_models: dict[str, TableModelMeta], binding_root: str) -> dict[str, TableDiff]:
+def validate_models(
+    registered_models: dict[str, TableModelMeta], binding_root: str
+) -> tuple[dict[str, TableDiff], dict[str, UUID], dict[str, dict[UUID, int]]]:
```
aaron-siegel: dict[str, tuple[TableDiff, UUID, int]] would be more logical to me. (Or even better a NamedTuple, dict[str, ValidationData] or some such? Or just put the table id and schema version into TableDiff.)

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
+            }
+            dropped_cols_by_qid = {col.qid: col for col in dropped_col_set}
+            view_dependencies = [
+                (dropped_cols_by_qid[qid], view_tv)
+                for qid, view_tv in views_by_qid.items()
+                if qid in dropped_cols_by_qid
+            ]
+            if len(view_dependencies) > 0:
+                # sort() for deterministic error message
+                view_dependencies.sort(key=lambda d: (d[0].qid.tbl_id, d[0].qid.col_id))
```
sergey-mkhitaryan: if this is only for tests passing consistently, I'd rather you update the test's expectations rather than do this extra thing in the product code that's kind of hard to parse.

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
-            for tvp, tv, update in tbl_info:
-                resolved_cols, resolved_idxs = model.prepare_model_updates(
-                    tvp, tv.display_str(), update['new_columns'], update['new_idxs']
+            # Apply per table in forward order (base tables first), so a view's new column can reference a base's
+            # new column: the view's resolution sees the base's already-mutated columns through `tvp.columns()`.
+            updated_tbl_ids = {tvp.tbl_id for tvp, _, _ in tbl_info}
+            applied_tbl_ids: set[UUID] = set()
+            for tvp, tv, schema_change in tbl_info:
+                # make sure we're doing this in base -> view order
+                pending_ancestor_ids = (set(tvp.tbl_ids[1:]) & updated_tbl_ids) - applied_tbl_ids
```
sergey-mkhitaryan, line 1904: This would have been simpler and more performant if instead of `applied_tbl_ids` you maintained `remaining_tbl_ids`. You'd simply check that no overlap is between `set(tvp.tbl_ids[1:])` and `remaining_tbl_ids`

aaron-siegel (approved):
I didn't have time for a full review, but I did read through catalog.py, model.py, and test_table_model.py.

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
+            # dependents are themselves being dropped is allowed.
+            dropped_col_set: set[Column] = set()
+            for _, tv, schema_change in tbl_info:
+                dropped_col_set.update(tv.cols_by_name[name] for name in schema_change['dropped_columns'])
+
+            def check_column_dependents(tbl_id: UUID, col_id: int, kind: str, name: str) -> None:
+                remaining_dependents = [
+                    c
+                    for c in self.get_column_dependents(tbl_id, col_id)
+                    if c.name is not None and c not in dropped_col_set
```
sergey-mkhitaryan: because of `c.name is not None` this fails to take into account index value columns. and index value columns can be in a different table.

more specifically, start with table base_t (c0, c1), and view_t (on base_t, idx=embedding index(c0)).
then update base_t to (c1) (i.e. drop c0) -- this change won't be blocked on view_t.

pixeltable/catalog/catalog.py
```diff
@@ -1762,62 +1765,120 @@ def create_from_model(
...
+                added_cols, added_idxs = model.prepare_model_updates(
+                    tvp, tv.display_str(), schema_change['new_columns'], schema_change['new_idxs']
                 )
-                tv.add_columns(resolved_cols, print_stats=False, on_error='abort')
-                for col, idx_name, idx_base in resolved_idxs:
-                    tv.add_index(col, idx_name, idx_base)
+                dropped_cols = [tv.cols_by_name[name] for name in schema_change['dropped_columns']]
+                dropped_idx_ids = [tv.idxs_by_name[name].id for name in schema_change['dropped_idxs']]
+                expected_schema_version = schema_change['expected_versions'][schema_change['expected_tbl_id']]
+                tv.apply_schema_change(expected_schema_version, added_cols, dropped_cols, added_idxs, dropped_idx_ids)
```
sergey-mkhitaryan, line 1919: we need to log the entire batch of schema changes that is being applied. otherwise, if a schema in production becomes corrupted, we will have difficultly debugging it. I can see that `apply_schema_change()` logs when the changes have been applied, but it doesn't say which changes specifically.

pixeltable/catalog/model.py
```diff
@@ -828,12 +773,9 @@ def prepare_model(
         # Build substitutions for the iterator's output columns.
         for name, output in iterator.outputs.items():
             catalog_col = catalog.Column.create(name, {'type': output.col_type, 'stored': output.is_stored})  # type: ignore[arg-type]
-            catalog_col.id = next(next_col_id)
             catalog_col.tbl_handle = tbl_handle
             user_cols[name] = catalog_col
-            subst_dict[ModelColumnRef(name)] = exprs.ColumnRef(
-                catalog_col.column_version_md(), perform_validation=(media_validation == MediaValidation.ON_READ)
-            )
+            preceding_names.add(name)
```
sergey-mkhitaryan, line 777: `media_validation` param is now unused

pixeltable/catalog/model.py
```diff
@@ -1227,173 +1180,193 @@ def _add_index_change(idx_name: str, idx: EmbeddingIndex) -> SchemaChange:
...
+) -> tuple[dict[str, TableDiff], dict[str, UUID], dict[str, dict[UUID, int]]]:
     """
     Analyze each registered model against the current catalog state, summarizing the schema changes that creating
     the models would entail, along with any incompatibilities with an already-existing table of the same name.
     This is purely informational: it neither modifies the catalog nor raises on incompatibilities.
+    All metadata reads happen in a single transaction.
+
+    Returns: (diffs, tbl_ids, schema_versions)
     """
+    from .catalog import retry_loop  # imported here: pixeltable.catalog.catalog imports this module
```
sergey-mkhitaryan: we normally don't justify function-level imports from catalog, I don't think it's useful.

pixeltable/catalog/table_version.py
```diff
@@ -251,13 +251,32 @@ def create_initial_md(
...
             col.id = next(column_ids)
             col.schema_version_add = 0
+
+        # resolve ColumnRefByName's to ColumnRefs in computed columns
+        subst = exprs.ExprDict[exprs.Expr](
+            (
+                exprs.ColumnRefByName(col.name),
+                exprs.ColumnRef(
+                    col.column_version_md(),
+                    perform_validation=col.effective_media_validation(media_validation) == MediaValidation.ON_READ,
```
sergey-mkhitaryan: I think that this new function `effective_media_validation()` is of a very limited utility, you could just do `(col._media_validation or media_validation)`

pixeltable/catalog/table_version.py
```diff
@@ -626,12 +650,24 @@ def drop_index(self, idx_id: int) -> None:
...
         # we're creating a new schema version
         self.bump_version(bump_schema_version=True)
+        cols_to_drop = self._cascade_drop_index(idx_id)
+        self._drop_columns(cols_to_drop)
+        self._write_md(new_version=True, new_schema_version=True)
+        _logger.info(f'Dropped index {idx_name} on table {self.name}')
+
+    def _cascade_drop_index(self, idx_id: int) -> list[Column]:
+        """
+        Mark an index as dropped and remove it from the store and the in-memory/persistent lookup structures.
```
sergey-mkhitaryan: this is not entirely correct because index value and undo columns are kept. can you rephrase it?

pixeltable/catalog/model.py

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/table_version.py
```diff
@@ -1363,17 +1518,41 @@ def propagate_update(
                 needs_iterator_reload = view_tv.is_component_view and any(
                     view_tv.is_iterator_column(col) for col in recomputed_cols
                 )
-                plan = None
-                if needs_iterator_reload:
-                    plan, _ = Planner.create_view_load_plan(view_tv.path, propagates_insert=True)
+
+                # an update can move a row in or out of a filtered view, but only if it changed one of the columns
+                # the filter is evaluated over
+                membership_may_change = (
```
sergey-mkhitaryan: this doesn't work correctly when you have base_t -> v1 (with predicate) -> v2 (no predicate). perhaps there needs to be a way for propagate_update to communicate from parent to child that new rows were added. which sounds a lot like an insert plan, so maybe that calls for a significant refactoring and unification of these plans. repro:

```
    def test_blah(self, make_catalog_path: Callable[[str], str]) -> None:
        p = make_catalog_path
        t = pxt.create_table(p('tbl'), {'n': pxt.Int})
        t.insert([{'n': 1}, {'n': 9}])
        v1 = pxt.create_view(p('view1'), t.where(t.n > 2))
        v2 = pxt.create_view(p('view2'), v1, additional_columns={'own': v1.n * 10})
        v3 = pxt.create_view(p('view3'), v2, additional_columns={'own3': v2.own + 1})
        assert [r['n'] for r in v1.collect()] == [9]
        assert [r['own'] for r in v2.collect()] == [90]
        assert [r['own3'] for r in v3.collect()] == [91]

        # an insert that qualifies reaches every level
        t.insert([{'n': 5}])
        assert sorted(r['n'] for r in v1.collect()) == [5, 9]
        assert sorted(r['own'] for r in v2.collect()) == [50, 90]
        assert sorted(r['own3'] for r in v3.collect()) == [51, 91]

        # an update that makes a row qualify must reach every level too
        t.update({'n': 7}, where=t.n == 1)
        assert sorted(r['n'] for r in v1.collect()) == [5, 7, 9]
        assert sorted(r['own'] for r in v2.collect()) == [50, 70, 90] # fails: 70 is not there
        assert sorted(r['own3'] for r in v3.collect()) == [51, 71, 91]
```

pixeltable/catalog/catalog.py
mkornacker: Introduced a new NamedTuple.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: Why not? The location is critical to being able to test something useful. I would personally like to see this explanation.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: We do this elsewhere, and doing this in the test might mean parsing out strings and converting them, etc.

pixeltable/catalog/model.py
mkornacker: We do this everywhere in local_table.py.

pixeltable/catalog/table_version.py

pixeltable/catalog/model.py
mkornacker: Done.

tests/test_table_model.py
```diff
@@ -508,6 +508,152 @@ class ExampleSubviewModelFromQuery(
...
+        """A view model column may shadow a base column; references to that name resolve to the shadowing column."""
+        p = make_catalog_path
+        TableModel = pxt.model_base()
+
+        class ExampleTableModel(TableModel, name='test_table'):
+            id: pxt.Required[pxt.Int]
+            value: pxt.Float
+
+        class ExampleViewModel(TableModel, name='test_view', base=ExampleTableModel):
+            value = ExampleTableModel.value * 100.0
```
sergey-mkhitaryan, line 521: In our imperative API (create_view, add_computed_column, etc.) we do not allow duplicate column names, but through declarative API we do. That means that some table structures created through declarative API, are not possible to create using imperative API. What do you think of this asymmetry?

pixeltable/catalog/catalog.py
mkornacker: Clarified in comment.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: This has no measurable performance impact.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: Done.

pixeltable/catalog/table_version.py

pixeltable/catalog/table_version.py
mkornacker: Fixed by propagating the potential membership change via the recursive calls.

tests/test_table_model.py
mkornacker: Done.

pixeltable/exec/sql_node.py
```diff
@@ -328,11 +332,14 @@ def create_from_clause(
...
             assert len(deleted_at_current_version) == 0
+            assert exclude_deleted_at_current_version is None
         joined_tbls: list[catalog.TableVersionHandle] = [candidates[0]]
         for t in candidates[1:]:
             # the tables in the path must either all be versioned or not
             assert t.get().is_versioned == versioned
-            if t.id in refd_tbl_ids:
+            # a versioned table has to be joined even if nothing references its columns, in order to apply the version
+            # predicate (= select only visible rows)
+            if versioned or t.id in refd_tbl_ids:
```
sergey-mkhitaryan, line 342: this, together with the changes in plan.py, causes the leaf view to be added to the join list with a v_min condition, and can lead to incorrect query result. odd that none of the tests caught this, maybe our coverage of various view trees is not good. repro:

```
    @pytest.mark.local('blah')
    def test_test(self, make_catalog_path:Callable[[str] , str])->None:
        p = make_catalog_path
        t = pxt.create_table(p('tbl'), {'n': pxt.Int, 'm': pxt.Int, 'k': pxt.Int})
        t.insert([{'n': 1, 'm': 1, 'k': 1}, {'n': 2, 'm': 2, 'k': 2}, {'n': 3, 'm': 3, 'k': 3}])
        v1 = pxt.create_view(p('v1'), t, additional_columns={'c1': t.k * 2})
        v2 = pxt.create_view(p('v2'), v1, additional_columns={'c2': v1.m * 3})

        t.update({'k': 100}, where=t.n == 2)
        t.update({'m': 50}, where=t.n == 3)

        res = v2.select(v2.n).order_by(v2.n).collect()
        assert res['n'] == [1,2,3]
```

PR #1491, opened 2026-07-29
pxt schema commands

mkornacker:
Implementation of the pxt schema subcommands, as outlined in the Notion design doc.

Plus: bug fix for view update propagation.


pixeltable/plan.py
```diff
@@ -825,13 +828,17 @@ def create_view_update_plan(
 
     @classmethod
     def create_view_load_plan(
-        cls, view: catalog.TableVersionPath, propagates_insert: bool = False, exclude_existing_rows: bool = False
+        cls,
+        view: catalog.TableVersionPath,
+        created_at_current_version: list[catalog.TableVersionHandle] | None = None,
```
sergey-mkhitaryan: in `create_view_update_plan` signature, the same parameter is required. let's sync them up.

pixeltable/plan.py
sergey-mkhitaryan, line 1004: this is very outdated, I don't think that the original idea behind the TODO is possible anymore.

pixeltable/catalog/table_version.py
```diff
@@ -1493,19 +1495,36 @@ def propagate_update(
                     and len(exprs.Expr.get_refd_column_ids(view_tv.predicate.as_dict()) & changed_qids) > 0
                 )
 
+                # a base that produced no plans of its own didn't create a new version, so its rows are matched
+                # live rather than by v_min; base_versions records that as a None, and zip() drops the ancestors
+                # this update never reached
+                created_bases = [b for b, v in zip(view_tv.path.get_bases(), base_versions) if v is not None]
```
sergey-mkhitaryan: claude found that the same problem still exists in delete propagation

```
    def test_noop_delete_of_base(self, make_catalog_path: Callable[[str], str]) -> None:
        """A base delete that matches no rows leaves the views on top of it alone."""
        p = make_catalog_path
        t = pxt.create_table(p('tbl'), {'n': pxt.Int})
        t.insert([{'n': 1}])
        # no columns and no filter of its own: an update of t leaves v untouched, so it stays at its old version
        v = pxt.create_view(p('v'), t)

        t.update({'n': 2})
        t.delete(where=t.n == 999)
        assert v.select(v.n).collect()['n'] == [2]
```

pixeltable/catalog/catalog.py
```diff
@@ -2310,6 +2309,22 @@ def read_tbl_record(self, tbl_id: UUID) -> schema.Table:
...
+        """
+        paths: list[str] = []
+        for tbl_id in tbl_ids[:max_paths]:
+            record = self.read_tbl_record(tbl_id)
+            paths.append(str(self.get_dir_path(record.dir_id).append(record.md['name'])))
+        paths.sort()
+        if len(tbl_ids) <= max_paths:
+            return ', '.join(repr(p) for p in paths)
+        listed = ', '.join(repr(p) for p in paths[:max_paths])
+        return f'{listed} and {len(paths) - max_paths} more'
```
sergey-mkhitaryan, line 2367: `len(tbl_ids) - max_paths`

pixeltable/catalog/catalog.py
```diff
@@ -2310,6 +2309,22 @@ def read_tbl_record(self, tbl_id: UUID) -> schema.Table:
...
+        transaction. Paths are listed in sorted order, and beyond a fixed limit are replaced by a count.
+        """
+        paths: list[str] = []
+        for tbl_id in tbl_ids[:max_paths]:
+            record = self.read_tbl_record(tbl_id)
+            paths.append(str(self.get_dir_path(record.dir_id).append(record.md['name'])))
+        paths.sort()
+        if len(tbl_ids) <= max_paths:
+            return ', '.join(repr(p) for p in paths)
+        listed = ', '.join(repr(p) for p in paths[:max_paths])
```
sergey-mkhitaryan, line 2366: `paths` cannot be longer than `max_paths`.

pixeltable/catalog/catalog.py
```diff
@@ -2310,6 +2309,22 @@ def read_tbl_record(self, tbl_id: UUID) -> schema.Table:
             raise excs.table_was_dropped(tbl_id)
         return schema.Table(**row._mapping)
 
+    def _tbl_paths_str(self, tbl_ids: Sequence[UUID], max_paths: int = 5) -> str:
+        """The paths of the given tables, comma-separated, for use in an error message.
+
+        Reads the stored records directly rather than loading each table, so that it is usable at any point in a
+        transaction. Paths are listed in sorted order, and beyond a fixed limit are replaced by a count.
```
sergey-mkhitaryan: "Paths are listed in sorted order" is technically correct, but because you truncate before sorting, the set of paths that are included is not deterministic, and I'm guessing that determinism is what you care about here.

pixeltable/catalog/catalog.py
```diff
@@ -2310,6 +2309,22 @@ def read_tbl_record(self, tbl_id: UUID) -> schema.Table:
...
+        """
+        paths: list[str] = []
+        for tbl_id in tbl_ids[:max_paths]:
+            record = self.read_tbl_record(tbl_id)
```
sergey-mkhitaryan: you may want to catch "table not found" errors here. probably impossible today, but since there's nothing in the docstring that says that tables must exist, other users may call this when tables may not exist.

pixeltable_cli/server/bridge.py
```diff
@@ -606,20 +607,158 @@ def schema_update(schema_path: str, target: str) -> tuple[list[str], list[str]]:
...
     for base in bases:
-        base_created, base_existed = base.create_all(target)
-        created.extend(base_created)
-        existed.extend(base_existed)
-    return created, existed
+        base.update_all(catalog_dir, allow_destructive=allow_destructive, destructive_hint=_DESTRUCTIVE_HINT)
+
+    # update_all() recomputes its own diff, so what it applied is only the plan above if nothing changed the
+    # catalog in between; re-reading confirms that before every operation is reported as applied
+    if not _schema_plan(bases, schema_file, catalog_dir)['in_agreement']:
```
sergey-mkhitaryan: this would be cleaner if update_all returned the change sets it applied.

pixeltable_cli/server/bridge.py
```diff
@@ -606,20 +607,158 @@ def schema_update(schema_path: str, target: str) -> tuple[list[str], list[str]]:
...
+    plan['ops'] = [schema_types.drop_table_op(pxt_path, 'applied') for pxt_path in dropped]
+    return plan
+
+
+def schema_update(
+    schema_file: str, catalog_dir: PxtPath, *, allow_destructive: bool = False
+) -> schema_types.SchemaPlan:
+    """Reconcile the binding root with the schema file: create missing tables and migrate existing ones.
+
+    Returns the plan that was applied, each operation annotated with its status. Applying is all-or-nothing, so a
```
sergey-mkhitaryan: > Applying is all-or-nothing

it's not because `cat.update_from_model` and `_create_models` each start its own transactions (create models start N transactions).

pixeltable_cli/server/bridge.py
```diff
@@ -606,20 +607,158 @@ def schema_update(schema_path: str, target: str) -> tuple[list[str], list[str]]:
...
+            assert blocked_by is not None
+            raise blocked_by
+        remaining = deferred
+
+    plan['ops'] = [schema_types.drop_table_op(pxt_path, 'applied') for pxt_path in dropped]
+    return plan
+
+
+def schema_update(
+    schema_file: str, catalog_dir: PxtPath, *, allow_destructive: bool = False
```
sergey-mkhitaryan, line 746: my problem with `allow_destructive: bool` is that it doesn't communicate which destructive ops the user approved. in the face of concurrent catalog updates, the user may see and approve one destructive diff, but what is actually applied could be another destructive diff.

pixeltable/catalog/model.py
```diff
@@ -1476,35 +1517,42 @@ def _create_models(binding_root: str, expect_created: set[str]) -> tuple[list[st
             (created if was_created else existed).append(str(tbl._path()))
         return created, existed
 
-    def _create_all(binding_root: str = '') -> tuple[list[str], list[str]]:
-        """Returns (created, existing): absolute paths of tables created now and those that already exist."""
+    def _create_all(catalog_dir: str = '', *, mismatch_hint: str = _PY_MISMATCH_HINT) -> tuple[list[str], list[str]]:
```
sergey-mkhitaryan: no caller sets `mismatch_hint` yet, do you plan on using it in the future?

pixeltable/plan.py
mkornacker: Done.

pixeltable/catalog/catalog.py

pixeltable/catalog/catalog.py
mkornacker: Done.

pixeltable/catalog/catalog.py
mkornacker: Done.

pixeltable_cli/server/bridge.py
mkornacker: I think the answer here is that you're inviting trouble if you're doing concurrent model-based schema changes and you're explicitly asking for columns to be dropped.

An engineering solution would involve a) getting a diff and locking the schema at the same time, and then b) applying that diff. That's a lot of code and has problems of its own (now you need a mechanism to release the lock somehow if the user doesn't do it, etc.). 

Most people are very careful about dropping columns and tables. We'll have to let them know that doing these things concurrently can lead to unanticipated results.

pixeltable/catalog/catalog.py

pixeltable/catalog/table_version.py
mkornacker: I'm removing the propagation fixes into a new branch/PR.

pixeltable/plan.py
mkornacker: Also moved to a new PR.

pixeltable_cli/server/bridge.py
mkornacker: Done.

pixeltable/catalog/model.py
mkornacker: Removed.

pixeltable_cli/server/bridge.py
mkornacker: Done.

pixeltable/utils/vision.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+
+
+def find_boundaries(mask: np.ndarray) -> np.ndarray:
```
sergey-mkhitaryan: what is this file?

pixeltable/utils/timing.py
```diff
@@ -0,0 +1,63 @@
+"""Timing utilities for profiling insert path performance."""
```
sergey-mkhitaryan: do we need this on main?

pixeltable_cli/server/bridge.py
```diff
@@ -606,20 +607,156 @@ def schema_update(schema_path: str, target: str) -> tuple[list[str], list[str]]:
...
+        'destructive': op['severity'] == 'destructive',
+    }
+
+
+def schema_prune(schema_file: str, catalog_dir: PxtPath) -> schema_types.SchemaPlan:
+    """Drop the tables under catalog_dir that no model in the schema file declares.
+
+    Returns the plan, with one drop_table operation per dropped table. A view is dropped before its base, so that
+    pruning a group of related tables does not depend on the order they are listed in. Nothing is force-dropped:
+    a table that something outside the pruned set depends on is left in place and its error is raised.
```
sergey-mkhitaryan, line 713: worth calling out that drop is non-atomic, and a raised error can mask the fact that a subset of tables got dropped.

pixeltable_cli/client/confirm.py
sergey-mkhitaryan: should this also be `refused_exit_code`?

pixeltable/catalog/model.py
```diff
@@ -1581,13 +1625,25 @@ def _update_all(binding_root: str = '', *, allow_destructive: bool = False) -> N
...
-        _create_models(binding_root, {name for name, d in diffs.items() if d['resolution'] == 'create'})
+        try:
+            _create_models(catalog_dir, {name for name, d in diffs.items() if d['resolution'] == 'create'})
+        except excs.Error as e:
+            # the migrations above are already committed; name them, so that a failure here doesn't read as
+            # though the catalog were untouched. Augmenting in place keeps the exception's type and fields.
+            if len(update_diffs) > 0:
+                migrated = ', '.join(repr(d['path']) for _, d in update_diffs)
+                e.args = (
+                    f'{e}\n\nThe following table(s) were already migrated: {migrated}. '
```
sergey-mkhitaryan: `str(e)` is message + detail, and detail is something that we deliberately want to avoid sending to the client, according to its comment. so maybe just `e.message` here?

pixeltable/catalog/model.py
```diff
@@ -1527,11 +1575,7 @@ def _update_all(binding_root: str = '', *, allow_destructive: bool = False) -> N
             detail = '\n'.join(line for name, d in destructive for line in _format_diff(name, d))
             raise excs.RequestError(
                 excs.ErrorCode.SCHEMA_MISMATCH,
-                f'The following updates would result in destructive catalog changes.\n'
-                f'{detail}\n'
-                'If you wish to apply these changes, re-run `update_all()` with `allow_destructive=True`.\n'
-                'If you intended to rename columns or indexes instead of dropping them, apply those changes '
-                'directly with `pxt.move()`.',
+                f'The following updates would result in destructive catalog changes.\n{detail}\n{destructive_hint}',
```
sergey-mkhitaryan: I asked claude to review this, and it noticed that `destructive_hint` is a part of presentation logic, which has no place in model. A cleaner solution would be for the CLI to catch this SCHEMA_MISMATCH, and add details/modify wording there. Adding from myself, I like that idea, but you may need a separate error code, or some kind of flag on the error that specifically communicates that destructive op guard is what causes it.

sergey-mkhitaryan (approved):
please update pr description

pixeltable/utils/timing.py
mkornacker: Accidentally added.

pixeltable/utils/vision.py
mkornacker: Accidentally added.

pixeltable_cli/server/bridge.py
mkornacker: Done.

pixeltable_cli/client/confirm.py
mkornacker: Done.

pixeltable/catalog/model.py

pixeltable/catalog/model.py

PR #1493, opened 2026-07-30
Cloud-hosted databases: client SDK and CLI 

mkornacker:


pixeltable_cli/client/commands/db.py
```diff
@@ -0,0 +1,267 @@
...
+    )
+
+    build_failed = result.get('last_build_state') == 'FAILED'
+    build_error = result.get('last_build_error') or ''
+    if not args.json_output:
+        final_state = result.get('state', '')
+        if build_failed:
+            print(f'Runtime build failed: {build_error}', file=sys.stderr)
+        elif final_state:
+            print(f'Runtime build {final_state.lower()}.')
```
sergey-mkhitaryan, line 256: check how we exit after getting into the `final_state` or `else` branches. I don't know what final states are possible, but a timeout should not exit with 0

pixeltable/serving/deploy.py
```diff
@@ -160,89 +37,76 @@ def _collect_project_files(project_dir: Path, include: list[str] | None, exclude
...
+    if not conda_prefix:
         return None
-
-    Env.get().console_logger.info(f'Found a conda environment: {os.environ["CONDA_DEFAULT_ENV"]}')
     try:
-        result = subprocess.run((conda_exe, 'env', 'export', '--no-builds'), capture_output=True, check=True)
-    except (FileNotFoundError, subprocess.CalledProcessError) as exc:
-        Env.get().console_logger.warning(f'Failed to export conda environment: {exc}')
+        result = subprocess.run(
+            ['conda', 'env', 'export', '--no-builds', '--prefix', conda_prefix], capture_output=True, check=True
```
sergey-mkhitaryan: What if they are using mamba (like us)? I think the CONDA_EXE way was more correct and reliable

pixeltable/serving/deploy.py
```diff
@@ -160,89 +37,76 @@ def _collect_project_files(project_dir: Path, include: list[str] | None, exclude
...
+        )
+    except (subprocess.CalledProcessError, FileNotFoundError):
```
sergey-mkhitaryan: If `conda` failed, I feel like it isn't safe to proceed with building the bundle as if conda env is not set up at all. Also, we should absolutely log something here if we catch one of these errors.

pixeltable/serving/deploy.py
```diff
@@ -160,89 +37,76 @@ def _collect_project_files(project_dir: Path, include: list[str] | None, exclude
...
-        conda_export: Output of `conda env export --no-builds`, included as `conda-env.yml`
-            in the bundle when provided.
-        project_dir: Path to the project directory. Defaults to the current working directory.
-
-    Returns:
-        Path to the generated tarball.
+    filtered = [
+        line
+        for line in result.stdout.decode('utf-8').splitlines(keepends=True)
+        if not re.match(r'^\s+-\s+pixeltable', line)
```
sergey-mkhitaryan: is it intended that we exclude `pixeltable-yolox` as well?

pixeltable_cli/client/commands/db.py
```diff
@@ -0,0 +1,267 @@
...
+
+def _update_runtime(args: argparse.Namespace) -> None:
+    # imported lazily: build_db_runtime_bundle pulls in pixeltable, which the stdlib-only client avoids
+    # loading for the other db subcommands
+    from pixeltable.serving.deploy import build_db_runtime_bundle
+
+    org, db = parse_db_uri(args.db_uri, prog='pxt db update-runtime')
+
+    if args.project_dir is not None:
+        os.chdir(args.project_dir)
```
sergey-mkhitaryan: Claude says os.chdir is a bad idea:

Inside a pxt shell session, after one db update-runtime --project-dir ~/proj/app-a:

1. The wrong project gets bundled and deployed — silently. This is the one that actually costs you something. A later db update-runtime pxt://org:other-db with no --project-dir falls back to Path.cwd() ([db.py:200](vscode-webview://0stiaclmb3hoc8ven7td18kbd4p7fpsiob8ff7q8lvcjeabvggk2/pixeltable_cli/client/commands/db.py#L200)), which is now app-a. It packages app-a's sources and lockfile and rebuilds other-db's runtime image from them. No error and no summary of what went in — just Building runtime bundle... done (12.4 MB). You find out when the hosted database starts running the wrong code.

2. Relative --project-dir compounds. A second --project-dir app-b resolves against ~/proj/app-a, so you get ~/proj/app-a/app-b: usually the clean "no project files found" error, but if such a subdirectory happens to exist, it bundles it without complaint.

3. service create / service update read the wrong pixeltable.toml. Both discover the [[pixeltable.service]] block from cwd — that's precisely why the e2e test has to pass cwd=_SAMPLE_APP. So route definitions would come from app-a while you're updating a service in an unrelated database. Relative --config, serve --config, and pxt schema <file> resolve against app-a too.

4. Config-singleton pinning outlives the chdir. Whichever directory is current when Config.get() first initializes decides the config file for the rest of the session — so the first update-runtime doesn't just move the cwd, it fixes the config source for everything after it.

pixeltable/serving/deploy.py
mkornacker: Done.

pixeltable/serving/deploy.py
mkornacker: Done.

pixeltable/serving/deploy.py
mkornacker: Done.

pixeltable_cli/client/commands/db.py
mkornacker: Done.

pixeltable_cli/client/commands/db.py
mkornacker: Done.

pixeltable_cli/client/hosted.py
```diff
@@ -0,0 +1,225 @@
...
+    return result
+
+
+def exit_if_pending(result: dict[str, Any], pending_states: frozenset[str], waited_for: str) -> None:
+    """Report an unfinished transition and exit 1. waited_for names it, eg 'the database to start'.
+
+    A state still in pending_states means the transition did not complete; an empty result means no state
+    was read at all.
+    """
+    if len(result) > 0 and result.get('state') not in pending_states:
```
sergey-mkhitaryan: we want to exit 0 only if the final state is successful, per the current operation's definition. otherwise, say, if the op is to start database, but the actual final state is STOPPED or FAILED, we will still exit 0.

pixeltable_cli/client/commands/db.py
```diff
@@ -0,0 +1,274 @@
...
+    elif args.action == 'update':
+        _update(args)
+    elif args.action == 'update-runtime':
+        _update_runtime(args)
+    elif args.action == 'delete':
+        _delete(args)
+
+
+def _create(args: argparse.Namespace) -> None:
+    org, db = parse_db_uri(args.db_uri, prog='pxt db create')
```
sergey-mkhitaryan: what about db_uri from the config file?

docs/release/platform/cli.mdx
```diff
@@ -47,7 +48,7 @@ These flags work the same way across the catalog commands that support them and
...
+| `--json` | Inspection, query, and mutation commands (`ls`, `describe`, `columns`, `computed`, `idxs`, `history`, `status`, `config`, `rows`, `get`, `count`, `errors`, `drop`, `drop-dir`, `rename`, `mv`, `revert`), plus `serve`, `db`, `service`, `org`, and `daemon status` | Emit machine-readable JSON instead. Not accepted by `shell` (interactive REPL), `health` (always JSON), `dashboard` (URL launcher), the navigation commands (`cd`, `pwd`), or the lifecycle subcommands that print plain status lines (`daemon start`/`stop`). |
 | `-n`, `--dry-run` | Every catalog mutation (`drop`, `rm`, `rename`, `mv`, `revert`) | Print the intended action; don't execute. `pxt serve` accepts the long form `--dry-run` only. |
```
sergey-mkhitaryan: `rm` is still mentioned on this line, as well as elsewhere in this doc

sergey-mkhitaryan (approved):

PR #1494, opened 2026-07-31
Dashboard - multi-catalog browsing, pipeline inspector, table detail

mkornacker:
The dashboard can browse hosted catalogs alongside the local one; table and pipeline views reworked. Two catalog bug fixes come along, both hit while testing
against a real catalog.

Dashboard
- Add hosted catalogs (pxt://org:db) next to the local one; persisted in the browser (lib/catalogs.ts), each tree loaded lazily on open (CatalogTree)
- Search spans the local catalog plus every added catalog, returning fully resolvable paths
- Table detail reworked: row count, iterator columns marked, computed expressions wrap and expand
- Pipeline inspector and column flow diagram simplified; kind badges for table/view/snapshot/iterator
- Directory responses missing entries repaired client-side with a one-time warning (api/normalize.ts, unit-tested)

Daemon
- GET /api/dashboard/search takes repeated catalogs params
- GET /api/dashboard/tables/meta returns row_count
- Column metadata carries is_iterator_col

Catalog fixes
- A component view whose iterator can no longer be located crashed on len(None) unless its outputs were persisted; such views now load with their stored data
readable
- InvalidGeneratingFunction never set name, so get_metadata() and describe() raised AttributeError


dashboard/src/api/normalize.ts
```diff
@@ -0,0 +1,27 @@
+import type { TreeNode } from '@/types'
+
+const warnedPaths = new Set<string>()
+
+function warnOnce(path: string): void {
```
sergey-mkhitaryan: nothing in this file is used except by its own test

pixeltable/func/iterator.py
```diff
@@ -358,6 +358,7 @@ class InvalidGeneratingFunction(GeneratingFunction):
 
     def __init__(self, fqn: str, d: dict[str, Any], error_msg: str) -> None:
         self.fqn = fqn
+        self.name = fqn.rsplit('.', 1)[-1]
```
sergey-mkhitaryan: declare `name` at the top of the class

sergey-mkhitaryan: same problem is in `GeneratingFunction._retrofit()`

pixeltable_cli/server/bridge.py
```diff
@@ -262,25 +263,17 @@ def export_table_csv(table_path: str, limit: int = 100_000) -> bytes:
...
 
-def search(query: str, limit: int = 50) -> dict[str, Any]:
-    """
-    Search across directories, tables, and columns.
-    """
-    query_lower = query.lower()
-
-    results: dict[str, Any] = {'query': query, 'directories': [], 'tables': [], 'columns': []}
-
+def _search_catalog(root: str, query_lower: str, results: dict[str, Any], limit: int) -> None:
```
sergey-mkhitaryan: I find it awkward, both the implementation and the user experience I imagine, that limit applies separately to each type of thing that we find. Can we just apply the limit to the sum of all items, and leave earlier once it is reached?

pixeltable_cli/server/bridge.py
```diff
@@ -300,8 +293,10 @@ def search(query: str, limit: int = 50) -> dict[str, Any]:
             results['tables'].append({'path': tbl_path, 'name': tbl_name, 'kind': tbl_md['kind']})
 
         # Search columns within this table (reuse tbl_md)
-        if tbl_md and len(results['columns']) < limit:
+        if tbl_md:
```
sergey-mkhitaryan: just `assert tbl_md is not None` right before line 292 and simplify both conditions.

pixeltable_cli/server/bridge.py
```diff
@@ -311,8 +306,30 @@ def search(query: str, limit: int = 50) -> dict[str, Any]:
...
+    query_lower = query.lower()
+
+    results: dict[str, Any] = {'query': query, 'directories': [], 'tables': [], 'columns': [], 'unavailable': []}
+
+    # The local catalog is the empty root; each additional catalog is searched at its hosted-uri root.
+    roots = ['', *(additional_db_uris or [])]
+
+    for root in roots:
+        try:
+            _search_catalog(root, query_lower, results, limit)
```
sergey-mkhitaryan: not ideal that we will continue searching through catalogs even if limit was reached on all search results.

dashboard/src/api/normalize.test.ts
```diff
@@ -0,0 +1,48 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
```
sergey-mkhitaryan: we should add this to pytest.yml and maybe Makefile (but I don't know which target), otherwise nothing runs these tests

pixeltable_cli/server/bridge.py
sergey-mkhitaryan: we essentially swallow all exceptions -- that can lead to incorrect search results. I think this should raise and result in an "unavailable" catalog

dashboard/src/components/TableDetailView.tsx
```diff
@@ -1667,7 +1636,16 @@ export function TableDetailView({ tablePath }: { tablePath: string }) {
...
     <div className="flex flex-col h-full animate-fade-in">
       {/* ── Header ──────────────────────────────────────────────────── */}
-      <TableHeader metadata={metadata} onTableClick={(path) => navigate(`/table/${path}`)} />
+      {(() => {
+        const node = pipelineData?.nodes.find(n => n.path === tablePath)
+        return (
+          <TableHeader
+            metadata={metadata}
+            versions={node?.versions ?? []}
+            rowCount={metadata.row_count}
```
sergey-mkhitaryan, line 1645: claude tells me that row count is already here: `rowCount={node?.row_count ?? null}`. if that's true, you don't need the change in `routes.py`. take it fwiw.

sergey-mkhitaryan (approved):

PR #1496, opened 2026-08-03
Add client side db name validation

amithadke:
Fix states in cli.mdx

### DIFF
--- a/docs/release/platform/cli.mdx
+++ b/docs/release/platform/cli.mdx
@@ -890,7 +890,7 @@ Route type reference:
 
 Manage cloud-hosted Pixeltable databases. A database is a hosted Pixeltable instance with its own compute, storage, and Python runtime.
 
-Database URIs use the form `pxt://org:db`. Valid states: `PROVISIONING`, `AVAILABLE`, `STOPPING`, `STOPPED`, `UPDATING`, `FAILED`.
+Database URIs use the form `pxt://org:db`. Valid states: `PROVISIONING`, `STARTING`, `AVAILABLE`, `UPDATING`, `STOPPING`, `STOPPED`, `FAILED`.
 
 The URI argument is optional: with it omitted, these commands use `db_uri` from the Pixeltable config file (see [Configuration](/platform/configuration)), so a project that sets it can run `pxt db status` and friends with no argument.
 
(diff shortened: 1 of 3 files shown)


eeshsaxena (commented):
Nice addition — client-side validation here is a good guardrail. A couple of things I noticed while reading through it:

### 1. The regex accepts a trailing newline (`.match` + `$`)

```python
_HOSTED_NAME_RE = re.compile(r'^[a-z0-9]([a-z0-9-]*[a-z0-9])?$')
...
if not _HOSTED_NAME_RE.match(value):
```

In Python, `$` matches just before a trailing `\n`, and `.match` only anchors the start, so a name ending in a newline passes validation:

```python
>>> _HOSTED_NAME_RE.match('main\n')     # truthy -> accepted
>>> _HOSTED_NAME_RE.fullmatch('main\n') # None   -> rejected
```

So `CreateDbRequest(org='acme', db='main\n')` validates successfully, and the newline then flows into URI / k8s-label construction downstream where it's invalid. Using `re.fullmatch(_HOSTED_NAME_RE, value)` (the `^`/`$` then become redundant), or anchoring with `\Z` instead of `$`, closes this.

### 2. Coverage: the rename path and other requests aren't validated

Only `CreateDbRequest.db` gets the validator, but `CreateDbRequest.db_name` and `UpdateDbRequest.db_name` are not. If `db_name` is a hosted/renamable name, a db could be created with a valid `db` and then renamed to an invalid or over-length name via update, bypassing this check. The other db-mutating requests (`Update`/`Start`/`Stop`/`Delete`) also take `db` unvalidated. Since `_validate_hosted_name(value, kind)` is already generic, reusing it on the rename fields would be low-cost. (If `db_name` is a free-form display name rather than a URI/k8s identifier, a short comment noting that would clear up the ambiguity.)

### 3. Test gap

The `rejects_invalid_name` parametrization doesn't cover `'main\n'` or `''`. Adding `'main\n'` would currently fail, which is a good regression guard for point 1.

### Minor
- The comment says "valid RFC 1123 label ... at most 29 characters" — RFC 1123 labels are max 63; 29 looks like a product-specific limit, so it might read more clearly without tying it to the RFC.
- `org` appears in the same `pxt://org:db` URIs but isn't validated anywhere; worth a note on whether it's user-chosen (and should use the same rule) or derived from the API key.

Points 1 and 2 are the substantive ones. Thanks!

mkornacker (approved):

PR #1497, opened 2026-08-04
[PXT-1101] operational tables: rename "unversioned" to "operational" and "versioned" to "data versioned"

sergey-mkhitaryan:


pixeltable/catalog/local_table.py
```diff
@@ -198,7 +198,10 @@ def _get_metadata(self) -> TableMetadata:
         )
 
     def _get_version(self) -> int | None:
```
mkornacker, line 200: I still think we should increment the version numbers for operational tables as well, but we'll only be doing that for schema changes. It would simplify the code.

pixeltable/metadata/schema.py
```diff
@@ -213,8 +213,8 @@ class TableMd:
 
     user: str | None
 
-    # for versioned tables, current_version monotonically increases for both data and schema changes, starting at 0
-    # not used for unversioned tables
+    # for data-versioned tables, current_version monotonically increases for both data and schema changes, starting at 0
+    # not used for operational tables
```
mkornacker, line 217: I think we should leave this logic in place for operational tables and still increment the version numbers in lockstep (current_version/_schema_version).

pixeltable/metadata/schema.py
```diff
@@ -245,9 +245,9 @@ class TableMd:
     tbl_state: TableState = TableState.LIVE
     pending_stmt: TableStatement | None = None
 
-    # Versioned tables keep their full schema and row history, and support time travel and rollback.
+    # Data-versioned tables keep their full schema and row history, and support time travel and rollback.
```
mkornacker: Operational tables also keep their schema history.

mkornacker (changes requested):
Overall feedback: let's restore current_version for operational tables, for the sake of simplification (I had wanted to ask for that when these changes were made originally, but it got merged before I got around to it).

pixeltable/metadata/schema.py
sergey-mkhitaryan: updated wording

pixeltable/catalog/local_table.py
sergey-mkhitaryan: Makes sense, left a couple of TODOs so as not to forget. Want to keep this pr simple and focused on just the nomenclature.

mkornacker (approved):

PR #1510, opened 2026-08-11
ConfigVar + FastAPIRouter for models

mkornacker:
Config vars
- pxt.ConfigVar declares a config variable at module scope and reads its value from
  whatever target the code runs on, so one table or model declaration is portable
  across local and hosted deployments.
- Declared types are str, pxt.URI and pxt.Secret. The type selects the config section
  the binding is read from, and Secret values are redacted wherever they are shown.
- A column's destination accepts a ConfigVar[URI], so a declaration names its object
  store instead of embedding one deployment's URI. Metadata stores only the reference,
  never a secret's value.
- Reading an unset var names the var and the config section to add it under. Using one
  in an expression, or interpolating one into a string, is rejected with the reason.

Serving
- Routes can be declared against a table model, not just a Table, so an app file
  declares its routes against the same models that define its schema.
- router.bind(base_path) resolves those models to tables and pins each route to the
  schema version it resolved against. Binding the same router at a second base path
  is rejected.
- router.get_service_diff(base_path) lists what the tables cannot serve as blocked
  operations, each carrying the pxt schema update command that clears it, rather than
  raising. bind() raises the same information when it cannot proceed.
- A router with unbound model routes now fails at app startup instead of per request.

Refactor
- catalog/model.py is split into a package (declaration, resolution, diff, base) with
  no behavior change. This is most of the diff.


pixeltable/catalog/globals.py
```diff
@@ -29,155 +30,55 @@
...
-    @property
-    def is_pure_snapshot(self) -> bool:
-        return (
-            self.tbl_md.view_md is not None
-            and self.tbl_md.view_md.is_snapshot
-            and self.tbl_md.view_md.predicate is None
-            and self.tbl_md.view_md.sample_clause is None
-            and len(self.schema_version_md.columns) == 0
-        )
+    # Allow unpacking as a tuple
```
aaron-siegel: Why not just make it a NamedTuple (which is basically what it is?)

pixeltable/catalog/globals.py
```diff
@@ -29,155 +30,55 @@
...
+    table_error_count: int | None = None
 
 
 class IndexSpec(NamedTuple):
     """
+    A declared spec names its column; a resolved one identifies it. A column of the table being created is
+    identified by the Column instance the builder assigns an id to; one that already exists, in this table or a
+    base, by its metadata.
+
     TODO: once the minimum Python is 3.11, make this generic in the column type, so that a declared spec is an
```
aaron-siegel: Just a note: the minimum Python is now 3.11

pixeltable/catalog/metadata_types.py
```diff
@@ -0,0 +1,138 @@
+"""The metadata records describing a table and its columns.
```
aaron-siegel: Suggestion: Make this `catalog/types.py` and allow other pure data types here too? QColumnId, TableVersionKey etc.

pixeltable/catalog/table_metadata.py
```diff
@@ -37,7 +37,11 @@ class ColumnMetadata(TypedDict):
     is_iterator_col: bool
     """`True` if this column is produced by an iterator (only applicable to views)."""
     destination: str | None
-    """An object store reference for computed files, if one is configured."""
+    """An object store destination for computed files, if one is configured.
+
+    A destination declared as a config var reads as '$(name)' rather than as the location it currently
```
aaron-siegel: Put '$(name)' in backticks or it might render strangely in the docs.

Would benefit from a cross link to docs on config vars since not all users will be familiar with them.

pixeltable/config.py
```diff
@@ -219,22 +339,16 @@ class Config:
...
-    __config_overrides: dict[str, Any]
+
+    # modification time and size of the config file when it was last read, needed for reload_if_changed()
+    __stamp: tuple[float, int] | None
 
     # section -> key -> (value, source_path); source_path is None for settings that don't come from a file
     __config_dict: dict[str, dict[str, tuple[Any, Path | None]]]
 
-    def __init__(self, config_overrides: dict[str, Any], additional_config_files: list[str]) -> None:
+    def __init__(self) -> None:
```
aaron-siegel: I'm not sure it's wise to back out `config_overrides`; dynamic config overrides was a specific feature requested by a user at one point. The user can achieve this anyway by explicitly setting `os.environ` values before initialization, so by removing this option we're not actually exercising any more control over what the user does or doesn't do wrt dynamic configuration. Why not make it easy for them?

This also brings to mind another thought, which is, what happens if the user changes an *environment variable* (not the config file) in between CLI calls? i.e. suppose I do this:

```
pxt schema update ...
```
I get an error because it requires an OpenAI key to resolve some index spec in the schema, so then I do this:
```
os.environ['OPENAI_API_KEY'] = ...
pxt schema update ...
```
I would expect the daemon to recognize that the environment has changed; will it?

pixeltable/config.py
```diff
@@ -273,25 +372,36 @@ def config_file(self) -> Path:
...
-            elif len(config_overrides) > 0 or len(additional_config_files) > 0:
-                raise excs.RequestError(
-                    excs.ErrorCode.INVALID_STATE,
-                    'Pixeltable has already been initialized; cannot specify new config values in the same session',
-                )
+                cls.__instance = cls()
+
+    @classmethod
+    def reload_if_changed(cls) -> bool:
+        """Reload the config file if it changed since it was read. Returns True if it was reloaded."""
```
aaron-siegel: Does this guarantee the `Env` instance will pick up the new config in all cases? What if I change the time zone (for example)?

tests/test_concurrent_model.py
```diff
@@ -154,7 +154,6 @@ class Base(TM, name='test_table'):
         class BaseV2(TM2, name='test_table'):
             id: pxt.Int
             text: pxt.String | None
-
```
aaron-siegel, line 157: ?

Why?

it's much clearer to me to have whitespace separating the __indexes__ declaration from the schema fields (i.e. that whitespace was intentional)

tests/test_table_model.py
```diff
@@ -1106,8 +1163,8 @@ class ExampleNewV2(TableModelV2, name='test_new'):
                 model filter   : id > 5
                 existing filter: id > 0
               sample mismatch (FATAL):
-                model sample   : sample(n=20, n_per_stratum=None, fraction=None, seed=2, [])
-                existing sample: sample(n=10, n_per_stratum=None, fraction=None, seed=1, [])
+                model sample   : sample(n=None, n_per_stratum=None, fraction=0.25, seed=2, [])
```
aaron-siegel, line 1167: Why did these all change?

tests/test_table_model.py
```diff
@@ -1516,7 +1573,7 @@ class ExampleQueryView(
             name='test_query_view',
             base=ExampleTable.select(ExampleTable.id, ExampleTable.value, plusone=(ExampleTable.value + 1))
             .where(ExampleTable.value > 0.5)
-            .sample(n=10, seed=1),
+            .sample(fraction=1.0, seed=1),
```
aaron-siegel, line 1579: A sample with fraction 1.0? What does that even mean?

aaron-siegel (changes requested):

pixeltable/catalog/globals.py

pixeltable/catalog/table_metadata.py
mkornacker: Done.

tests/test_concurrent_model.py
mkornacker: Done.

tests/test_table_model.py
mkornacker: Explained in comment.

pixeltable/catalog/globals.py

pixeltable/catalog/metadata_types.py
mkornacker: Done.

tests/test_table_model.py
mkornacker: We only allow fractional sampling for view queries, see create_view() (which used to check for that directly). It was an oversight that we allowed this for model views.

pixeltable/config.py
mkornacker: As discussed, we're still dropping config_overrides.

pixeltable/serving/_fastapi.py
```diff
@@ -339,11 +423,160 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+            tbl = pxt.get_table(f'{catalog.Path.dir_prefix(base_path)}{tbl_name}', if_not_exists='ignore')
+            if tbl is None:
+                ops.append(
+                    blocked_route_op(
+                        spec.display_name, f'{spec.display_name}: table {tbl_name!r} does not exist', base_path
+                    )
+                )
+                continue
+
+            cols_by_name = {col.name: col for col in tbl._tbl_path.column_md() if col.name is not None}
```
sergey-mkhitaryan: P1 pixeltable/serving/_fastapi.py:466 — get_service_diff() validates only referenced column names and types. It does not validate table kind, primary keys, or iterator/cardinality semantics. For example, an insert route declared on a table model can bind successfully to a same-shaped view; every request then fails when it calls tbl.insert(). Likewise, an update route can bind to a table lacking the model’s PK. Treat these route-relevant structural properties as blocked schema mismatches before binding.

pixeltable/config.py
```diff
@@ -194,6 +205,169 @@ def _check_python_version(cls, v: str | None) -> str | None:
...
+
+    TAG = '$confvar'
+
+    # types a ConfigVar may declare, by name, so that a stored reference can be reconstituted
+    _CONFVAR_TYPES: ClassVar[dict[str, type[str]]] = {'str': str, 'URI': URI, 'Secret': Secret}
+
+    name: str
+    type_: type[ConfVarT]
+
+    def __init__(self, name: str, type_: type[ConfVarT]) -> None:
```
sergey-mkhitaryan, line 290: P2 pixeltable/config.py:294 — ConfigVar accepts arbitrary declared types, although serialization only supports str, URI, and Secret. A ConfigVar('x', SomeStrSubclass) can be persisted with an unrecognized type name and subsequently fails to load metadata. Validate type_ at construction against _CONFVAR_TYPES.values() (or serialize a canonical supported type).

pixeltable/globals.py
```diff
@@ -41,16 +41,13 @@
 _logger = logging.getLogger(__name__)
 
 
-def init(config_overrides: dict[str, Any] | None = None, additional_config_files: list[str] | None = None) -> None:
+def init() -> None:
```
aaron-siegel: I thought we agreed to keep `config_overrides`? If we're keeping environment variables around as a means of configuration, there is absolutely no reason to get rid of it (it's basically a convenience alternative to using environment variables)

Why remove a feature that (i) users have requested, (ii) has already been implemented, and (iii) creates no problems for us?

(I'm fine with getting rid of additional_config_files, as we've discussed.)

aaron-siegel (changes requested):

aaron-siegel (approved):
I'm approving so this doesn't block on another round of review, but my approval assumes `config_overrides` will be restored before committing. If that's controversial, let's please discuss one more time and try to resolve it first.

pixeltable/catalog/local_table.py
```diff
@@ -139,7 +139,7 @@ def _get_metadata(self) -> TableMetadata:
                 comment=col.comment,
                 custom_metadata=col.custom_metadata,
                 is_iterator_col=False,
-                destination=col._explicit_destination,
+                destination=col.display_destination,
```
sergey-mkhitaryan, line 142: `Column.display_destination` eventually falls back to `Env.get().default_output_media_dest`/`Env.get().default_input_media_dest`. This can result in unexpected SCHEMA_MISMATCH when the result of this is compared to the spec that has `None` destination.

pixeltable/serving/_fastapi.py
sergey-mkhitaryan: 2. Medium — pixeltable/serving/_fastapi.py:466: compute route on a view with an explicit select() can never bind
_add_dml_route registers input_cols from the base table's column_md() (:2265), but get_service_diff looks them up in tbl._tbl_path.column_md(), which only includes base columns when the view has no select list (view.py:130). Result: every base input column reports has no column 'x' (blocked) and bind() always fails. Check compute-route inputs against the root path. The new test_view_model_target only covers a base query without a select list.

pixeltable/config.py
```diff
@@ -219,47 +393,31 @@ class Config:
...
-        additional_configs = [self.__load_project_config(Path(f)) for f in additional_config_files]
-
-        self.__config_dict = {}
-
-        # Load lowest precedence first (for additional configs, last specified = highest precedence).
-        for source in (user_config, pyproject_config, project_config, *additional_configs):
-            self.__merge_config(self.__config_dict, source)
+        self.__config_dict = self.__load_user_config()
+        self.__stamp = self.__file_stamp()
+        self.__warn_about_miscased_env_vars()
```
sergey-mkhitaryan, line 435: 3. Medium — docs/release/platform/cli.mdx:620 (also 587, 639, 755, 770, 809): documented config surface removed
Config no longer reads ./pixeltable.toml or [tool.pixeltable], and --config was dropped from pxt serve / service create / service update, but the docs still document both. A user following the docs gets SERVICE_NOT_FOUND with no flag to point at the file (the e2e test had to switch to PIXELTABLE_CONFIG=). Meanwhile pxt db update-runtime still reads pixeltable.toml (serving/deploy.py:102) — the same file now behaves inconsistently.

sergey-mkhitaryan (commented):
Some LLM-generated findings that made sense to me after a quick investigation.

pixeltable/serving/_fastapi.py
mkornacker: Reworked the way models are validated in FastAPIRouter.

mkornacker: Done.

pixeltable/catalog/local_table.py
mkornacker: Reworked how metadata is reported.

pixeltable/globals.py

pixeltable/config.py
mkornacker: Done.

pixeltable/config.py
mkornacker: Done.

pixeltable/config.py
mkornacker: Clarified.

PR #1512, opened 2026-08-11
[ docs ] Consolidate agent guidance for docs and dev

apreshill:
## Summary

Documentation writing guidance was spread across three root-level files and a
cookbook style guide that no longer matched the current `docs/release/` layout.
This moves that guidance into one file that lives with the content it governs,
retires the stale guides, and stops maintaining `AGENTS.md` and `CLAUDE.md` as two
copies of the same thing.

## Heads up: your agent files change

**`CLAUDE.md` is now a thin file that imports `AGENTS.md`.** No guidance is
lost. It carries a one-line `@AGENTS.md` import plus `Required After Every Code
Change`, which was the only genuinely Claude-specific content.

- Claude Code reads `CLAUDE.md` and does *not* read `AGENTS.md`, so the import
  is Anthropic's documented way to keep one source of truth. Claude loads the
  imported file at session start and sees what it saw before.
- The two files had already drifted: `CLAUDE.md` had `Required After Every Code
  Change`, `AGENTS.md` had `Key Files Reference` and `Getting Help`. Every edit
  either had to be made twice or the files diverged further.
- The import is used rather than `ln -s AGENTS.md CLAUDE.md` because symlinks
  need Administrator or Developer Mode on Windows.
- **What to do differently:** edit `AGENTS.md`. Put something in `CLAUDE.md`
  only if it is Claude-specific.

**`AGENTS.md`** regains the "never change `integrations.telemetry.enabled`"
guard. It had been relocated into `docs/release/AGENTS.md`, which agent tooling
does not load automatically, so the guard depended on an agent choosing to go
look for it.

**`.github/copilot-instructions.md`** is additive only. Nothing is removed. It
gains a pointer to the docstring guidelines and a bullet for notebooks; the
existing inline rules stay inline, where Copilot reliably sees them.

## New: `docs/release/AGENTS.md`

- Documentation writing guidance: voice, callouts, internal link construction
  from `docs.json`, cross-referencing, new page structure, and the local build
  and preview workflow.
- Scoped to `docs/release/`, so it costs no context when working on library code.
- `docs/_guidelines/` keeps the mechanical how-to for notebooks and docstrings.
  This file covers writing craft and points at those for mechanics.

## Removed

- `docs/_old/cookbooks/{STYLE_GUIDE,README,CONTRIBUTING}.md`, which described a
  cookbook layout that no longer exists. Nothing in the repo references them.

## Fixed

- `GUIDELINES_FOR_NOTEBOOKS.md`: the Kaggle, Colab, and GitHub frontmatter links
  used the old `docs/notebooks/` prefix. Now `docs/release/`, matching what
  shipped notebooks actually contain.
- The cross-reference sweep command used grep syntax (`"a\|b"`) that matches
  nothing under ripgrep, which is what this repo tells agents to use.
- Three relative links to `docs/_guidelines/` that `mint broken-links` rejects,
  since those are repo files rather than published pages.


pierrebrunelle:
Closing this in favor of #1628, which carries the parts that still apply. Thank you for it: the diagnosis was right, and the consolidation is now done.

**What #1628 takes from here**

- **CLAUDE.md stops being a second copy.** It is now nine lines whose body is `@AGENTS.md`, the Claude Code import directive, so Claude Code reads exactly what Codex and the other agents read. That reaches the goal of this PR without deleting content.
- **AGENTS.md gains the two sections it was missing**, `Required After Every Code Change` and `Code Comments`. An agent reading AGENTS.md had never seen the close-out procedure at all.
- **`TARGET=dev` for docs deploys**, stated as a rule: dev is the only target an agent may run or suggest, stage and prod are for humans. Both files showed `TARGET=stage` as the example.
- **AGENTS.md gains an index** pointing at the instructions that govern the other surfaces: Copilot code review, the four docs guidelines, the dashboard docs, the skill, CONTRIBUTING.

**Why not just rebase this one**

It has been conflicting since August, and the conflict is not mechanical. CLAUDE.md has grown from 349 to 386 lines since this branch was cut, most of it the five-step close-out procedure and the prose rules. Merging as-is would delete that work, and this branch's version of `Required After Every Code Change` is the older three-step one. Porting forward was safer than rebasing backward.

**Not carried over:** the new `docs/release/AGENTS.md` writing-craft file. `docs/_guidelines/` already covers that ground, and adding a fifth agent-facing doc cuts against consolidating.

One thing found while porting: CLAUDE.md step 5 cited `feedback_simple_direct_prose.md`, which exists nowhere in the repo. Dropped the pointer, since the rules it named are spelled out inline.

PR #1514, opened 2026-08-11
Pixeltable daemon logs at DEBUG during tests

sergey-mkhitaryan:
new `--test` cli arg enables DEBUG logging and the `/reinitialize` endpoint.

Motivation for more verbose logging is to help investigate daemon getting stuck e.g. https://github.com/pixeltable/pixeltable/actions/runs/31444605474/job/93636063420

### DIFF
--- a/pixeltable/service/proxy_daemon.py
+++ b/pixeltable/service/proxy_daemon.py
@@ -267,7 +267,7 @@ def configure_test_logging(db: str) -> None:
     ep = endpoint(db)
     if ep is None:
         raise excs.Error(excs.ErrorCode.INTERNAL_ERROR, f'No running proxy daemon for {db!r}')
-    response = httpx.post(f'{ep}/configure_test_logging', timeout=1.0)
+    response = httpx.post(f'{ep}/configure_test_logging', timeout=10.0)
     response.raise_for_status()
 
 
--- a/pixeltable/service/proxy_daemon.py
+++ b/pixeltable/service/proxy_daemon.py
@@ -14,6 +14,7 @@ a given database, and locating a running one via the port.lock file in its home 
 
 import atexit
 import json
+import logging
 import os
 import shutil
 import signal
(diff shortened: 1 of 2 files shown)


mkornacker:
Reviewed this in the context of some follow-on work we want to do (surfacing server-side exception detail to proxy clients), so a few notes. Two things I checked first that turned out fine:

**1. Stale daemons become a confusing failure.** The docstring notes that test mode "only takes effect if this call actually launches the daemon; an already-running one is returned as is." Previously that was benign. Now a daemon left over from an earlier run has no `/reinitialize` registered, and `tests/conftest.py:345` calls it unconditionally, so you get a 404 on a call that used to work, inside a fixture, which will read as an unrelated failure. Worth having the fixture check `/health` for test mode and restart if it doesn't match, or at minimum making `reinitialize()` fail with a message that names the cause.

**2. The fixed-address branch ignores `test_mode` for uvicorn.** The loopback path sets `log_level = 'debug'` under test mode, but the fixed-address path uses `config.get_string_value('log_level') or 'info'` and never consults it. Per the docstring, fixed-address is the mode used "when an external orchestrator (e.g. a sidecar) handles routing" -- i.e. the cloud, which is where the verbosity is most useful. `_logger.info('Test mode enabled')` is also only in the loopback branch, so a fixed-address daemon gives no signal that it is in test mode at all.

**3. The NOTSET coupling is invisible.** Point one above only works because of those guards in `env.py`. Nothing in either file records that dependency, so removing or refactoring the guards would silently disable this PR's main feature. One comment at the `_serve()` site would cover it.

**4. Worth verifying:** with `log_config=None`, does uvicorn still apply `log_level` to its own loggers, or does skipping `dictConfig` make the setting inert? If the latter, the `'debug'` in the loopback branch does nothing and the real effect comes only from the `pixeltable`/`sqlalchemy.engine` levels. I did not chase this down.

**5. Ad-hoc argv parsing.** `test_mode = '--test' in sys.argv[1:]` silently ignores anything unrecognized. Fine for one flag, but it will rot as flags accumulate.

## Two requests for follow-on work

**Record the mode on `Env`.** It is currently a parameter threaded to `_build_app`, but `proxy_dispatch.dispatch()` runs inside the `/rpc` route with no access to it -- and that is where we need to read it to gate returning server-side error detail to clients. Doing this here makes the follow-up a small delta instead of a refactor.

**Reconsider the name while it is cheap.** `--test` is right for the test suite and wrong for a cloud sandbox, where you are debugging a deployed environment rather than running tests. We want to hang error passthrough on the same mode and surface it as a per-database setting, and `--dev` or `--debug` reads better for that. One call site in conftest today; more later.

mkornacker (approved):

PR #1515, opened 2026-08-11
[PXT-1269] Direct b-tree and embedding index support for operational tables

sergey-mkhitaryan:
`BtreeIndex` has 2 modes:
1. `uses_value_col==True`(the existing behavior on `main` for data-versioned tables): value column is created to hold values to index (i.e. live row versions only, NULL otherwise, and long strings are truncated). The store index is created on this value column.
2. `uses_value_col==False` (the new behavior): there is no value column, the store index is created directly on the indexed column.

The rule for undo columns is that they are only created if `uses_value_col` is True, and only in data-versioned tables. In practice, this preserves the existing behavior today: undo columns are created for all indexes in data-versioned tables.

Today the new mode `uses_value_col==False` is used with only b-tree indexes on operational tables. In the future, we can implement it for data-versioned tables. To keep the existing (and correct) behavior of only indexing the latest row versions, the store index will need to declared as [partial](https://www.postgresql.org/docs/current/indexes-partial.html).

Embedding indexes on operational are supported by using value columns but not undo columns.

Postgres limits the length of the length of a string that can be indexed using btree. The pre-existing behavior with value columns is that Pixeltable truncates long strings, and adjusts query plans that rely on the index accordingly. The new `BtreeIndex` mode without value column shifts this responsibility to the user. Creating a b-tree index on a column with strings that are too long (~2KB+), or inserting such a string in a column with b-tree index will raise a `CONSTRAINT_VIOLATION` error.


pixeltable/catalog/table_version.py
```diff
@@ -491,13 +518,9 @@ def _init_schema(self) -> None:
...
-            # Finally initialize the index whose undo column this is. Undo columns have the highest col id of all
-            # columns involved in an index, so by the time undo column is initialized, the index can be initialized
-            # as well.
-            if self.supports_idxs and col.id in undo_col_to_idx:
-                idx, idx_md = undo_col_to_idx[col.id]
-                if idx_md.is_visible_in_version(self.schema_version):
-                    self._init_idx(idx, idx_md)
+            # Finally initialize the indexes for which this is the last column they reference; all of their columns
+            # now exist.
+            self._init_visible_idxs(idxs_by_last_col_id.get(col.id, []))
```
mkornacker: Comment why we need this complicated logic, instead of simply initializing indices in a single pass after the columns

pixeltable/catalog/table_version.py
```diff
@@ -526,34 +549,49 @@ def _init_col_value_expr_from_md(
...
+            idxs_by_last_col_id.setdefault(max(col_ids, default=None), []).append((idx, md))
 
-        undo_col_to_idx: dict[int, tuple[index.IndexBase, schema.IndexMd]] = {}
-        for idx_md, idx in indexes:
-            assert idx_md.index_val_undo_col_id not in undo_col_to_idx
-            undo_col_to_idx[idx_md.index_val_undo_col_id] = (idx, idx_md)
+        return idxs_by_last_col_id
 
-        return undo_col_to_idx
+    def _init_visible_idxs(self, idxs: list[tuple[index.IndexBase, schema.IndexMd]]) -> None:
```
mkornacker: I think _init_idxs() is sufficient, this made me wonder if we're initializing any other indices.

Also, this should check supports_idxs. As it stands, it doesn't make sense on its own, you need to understand that build_idxs_by_last_col_id() was called earlier and already checked supports_idxs. That's an ordering dependency that makes the code harder to follow.

pixeltable/catalog/table_version.py
```diff
@@ -646,14 +685,52 @@ def _create_index(
...
+    def _create_index_columns_for_schema(
+        cls,
+        tbl_handle: TableVersionHandle,
+        col: Column,
+        idx: index.IndexBase,
+        schema_version: int,
+        is_data_versioned: bool,
+        next_col_id: Callable[[], int],
+    ) -> tuple[Column | None, Column | None]:
+        """Create the columns that idx needs in order to index col, as far as it needs any.
```
mkornacker: ", as far as it needs any" doesn't add anything (it already says "that idx needs" earlier).

"Create the columns needed to index col." would say the exact same thing.

pixeltable/catalog/table_version.py
```diff
@@ -646,14 +685,52 @@ def _create_index(
...
+            self.handle,
+            col,
+            idx,
+            schema_version=self.schema_version,
+            is_data_versioned=self.is_data_versioned,
+            next_col_id=self.next_col_id,
+        )
+
+    @classmethod
+    def _create_index_columns_for_schema(
```
mkornacker: I didn't understand '_for_schema'.

pixeltable/catalog/table_version.py
```diff
@@ -1991,11 +2065,22 @@ def get_computed_col_names(self) -> list[str]:
...
+        """All index value columns of all indexes in this version."""
+        return {info.val_col for info in self.idxs.values() if info.val_col is not None}
+
+    @property
+    def idx_undo_cols(self) -> set[Column]:
+        """All index undo columns of all indexes in this version."""
+        return {info.undo_col for info in self.idxs.values() if info.undo_col is not None}
+
     def get_idx_val_columns(self, cols: Iterable[Column]) -> set[Column]:
+        """Return the index value columns of the indexes on cols; indexes without one are skipped."""
```
mkornacker: The comment isn't really saying anything that's not already obvious (it's set[Column]; if an index doesn't have a value col, there's nothing to include).

pixeltable/catalog/table_version.py
```diff
@@ -646,14 +685,52 @@ def _create_index(
...
+        is_data_versioned: bool,
+        next_col_id: Callable[[], int],
+    ) -> tuple[Column | None, Column | None]:
+        """Create the columns that idx needs in order to index col, as far as it needs any.
+
+        Returns (value column, undo column), both of which are optional.
+        """
+        if not idx.uses_value_col:
+            return None, None
+        val_col = Column.create_index_value_column(
```
mkornacker: Why not have a combined Column.create_index_columns() instead of this split plus this class method plus the _create_index_columns() wrapper?

pixeltable/exprs/comparison.py
```diff
@@ -84,25 +69,36 @@ def _sql_compatible(t1: ts.ColumnType, t2: ts.ColumnType) -> bool:
...
-                left = idx_info[0].val_col.sa_col
+            assert isinstance(self._op1, ColumnRef)
+            col = self._op1.col
+            # indices don't apply to snapshots
+            tbl = col.get_tbl()
+            idx_info = None if tbl.is_snapshot else tbl.find_btree_index(col)
+            if idx_info is not None:
+                if idx_info.val_col is None:
+                    # No index value column, the store index is on the indexed column itself
+                    left = col.sa_col
```
mkornacker: I think this is exactly what it was already assigned in l89.

tests/test_index.py
```diff
@@ -980,10 +981,12 @@ def create_random_str(n: int) -> str:
         assert t.where(t.data >= s).count() == self.BTREE_TEST_NUM_ROWS - idx
         assert t.where(t.data > s).count() == self.BTREE_TEST_NUM_ROWS - idx - 1
 
-        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION) as exc_info:
-            assert len(data[56]) == 256
-            _ = t.where(t.data == data[56]).count()
-        assert 'String literal too long' in str(exc_info.value)
+        # Verify Comparison edge cases around max btree value length
```
mkornacker, line 989: How is this verifying the edge case? It looks like it dropped the edge case.

pixeltable/index/btree.py
```diff
@@ -18,6 +18,11 @@
 class BtreeIndex(IndexBase):
     """
     Interface to B-tree indices in Postgres.
+
+    Two representations, selected by `uses_value_col`:
```
mkornacker: This file doesn't use backticks in code comments, let's honor existing patterns.

pixeltable/exprs/row_builder.py
```diff
@@ -601,16 +607,22 @@ def create_row_batch(
             assert self.row_batch_col_types is None
             self.row_batch_output_map = []
             self.row_batch_col_types = {}
-            # index value columns have name=None; key their values by index name (unique within the table)
             tbls = {col.get_tbl().id: col.get_tbl() for col in self.table_columns}
-            idx_names = {info.val_col.qid: name for tbl in tbls.values() for name, info in tbl.idxs_by_name.items()}
-            undo_col_qids = {info.undo_col.qid for tbl in tbls.values() for info in tbl.idxs_by_name.values()}
+            idx_val_col_qid_to_idx_name = {
```
mkornacker: Does this really make the code more readable?

mkornacker (changes requested):

mkornacker:
I think this should also address embedding indices. Some of the logic changes also pertain to them, but there's no proof it actually works. Splitting this work up into two pieces creates more total work.

pixeltable/catalog/table_version.py
sergey-mkhitaryan: That comment was on `_build_idxs_by_last_col_id()`, moving here instead.

pixeltable/exprs/row_builder.py
sergey-mkhitaryan: renamed back to `idx_names`, but yeah I think it was better.

pixeltable/exprs/comparison.py
sergey-mkhitaryan: good catch, simplified.

tests/test_index.py
sergey-mkhitaryan: This is about the string branch in `Comparison._can_use_index_value_col()`. The first two queries get a `False` because the literal length is 256 (which equals to the limit). The next one also gets a `False` because the literal length is 257. And the last one gets a `True` because the literal length is 255. Those are the boundary conditions I'm referring to.

pixeltable/catalog/table_version.py
```diff
@@ -526,34 +555,46 @@ def _init_col_value_expr_from_md(
...
-        undo_col_to_idx: dict[int, tuple[index.IndexBase, schema.IndexMd]] = {}
-        for idx_md, idx in indexes:
-            assert idx_md.index_val_undo_col_id not in undo_col_to_idx
-            undo_col_to_idx[idx_md.index_val_undo_col_id] = (idx, idx_md)
+        return idxs_by_last_col_id
 
-        return undo_col_to_idx
+    def _init_idxs(self, idxs: list[tuple[index.IndexBase, schema.IndexMd]]) -> None:
+        """Initialize those of idxs that are visible in the current schema version."""
+        for idx, idx_md in idxs:
```
mkornacker, line 584: This needs to check supports_idxs. As it is, it's only correct to call this method after _build_...(), meaning it doesn't stand on its own and has subtle control flow dependencies that aren't directly visible.

tests/test_operational_table.py
```diff
@@ -91,6 +94,104 @@ def test_select_limit_offset(self, uses_db: None) -> None:
...
+        )
+        assert tbl.count() == 5
+
+        sim = tbl.text.similarity(string='Volcanic eruptions reshape entire coastlines in a matter of days.')
+        assert tbl.select(tbl.id).order_by(sim, asc=False).limit(1).collect()['id'] == [1]
+
+        validate_update_status(tbl.delete(where=tbl.id == 4), 1)
+
+        sim = tbl.text.similarity(string='Espresso machines build pressure in order to extract the coffee.')
+        res = tbl.select(tbl.id).order_by(sim, asc=False).collect()['id']
```
mkornacker, line 198: Missing: drop_index/drop_embedding_index().

mkornacker (approved):

tests/test_operational_table.py
sergey-mkhitaryan: Added `drop_embedding_index()`, `drop_index()` is already covered

PR #1517, opened 2026-08-12
Test issues with Yolox

sergey-mkhitaryan:
* `TestYolox` is marked as very expensive
* `TestQuery.test_to_coco()` becomes `TestYolox.test_yolox_coco_integration()`
* `TestQuery.test_to_coco()` is a simpler version that does not depend on Yolox or any non-local services


mkornacker (approved):

aaron-siegel:
I'm concerned about so many tests being marked very_expensive. No one is monitoring the nightly tests, which means tests marked as very_expensive are currently just going into a void.

What's the argument against just marking it expensive? IMO we should bias towards 'expensive' over 'very_expensive' whenever feasible.

PR #1522, opened 2026-08-14
Reduce the pr (i.e. on push) test coverage to only static checks + slimtest

sergey-mkhitaryan:
The merge queue coverage remains the same.

Note: I hardcoded the list of "slim" tests to `ci_tool.py`, so we now have 2 copies of this list, the other one being in `Makefile`. The alternative was to have a `.txt` file in the repo with this list, so that `Makefile` and ci tool both parse it, but I thought that that would be ugly. Those 2 lists diverging won't be a critical problem, if that happens.


tool/ci_tool.py
```diff
@@ -68,33 +94,49 @@ def generate_matrix(args: argparse.Namespace) -> None:
...
+    # Linting, type checking, etc.; run on every trigger.
+    configs = [MatrixConfig('static-checks', 'lint', MAIN_PLATFORM, '3.11')]
 
     if trigger == 'pull_request':
-        # Tier 1 only: Just the standard tests on MAIN_PLATFORM.
-        configs.append(MatrixConfig('standard', 'py', MAIN_PLATFORM, '3.11'))
-        # Notebook tests are not run on PRs (Hugging Face downloads are rate-limited without a token, which is
-        # unavailable on PRs). Non-HF notebooks run in the merge queue; all notebooks run on the scheduled tier.
+        # Configs selected for a PR validation, i.e. on every push to a PR.
+        # Tier 1 only: static checks plus the slim test subset, both on MAIN_PLATFORM. This is strictly a subset of
```
aaron-siegel: "static checks" is a little misleading since they're covered above, not here.

tool/ci_tool.py
```diff
@@ -68,33 +94,49 @@ def generate_matrix(args: argparse.Namespace) -> None:
...
+        # checks that are run in merge queue.
+        configs.append(
+            MatrixConfig(
+                'slim',
+                'py',
+                MAIN_PLATFORM,
```
aaron-siegel: Don't we want to keep these on all BASIC_PLATFORMS? The slim tests should be quick to run.

tool/ci_tool.py
```diff
@@ -68,33 +94,49 @@ def generate_matrix(args: argparse.Namespace) -> None:
...
+        )
 
     else:
+        # A non-PR trigger: merge queue or a schedule
+
+        # Standard configs on the basic platforms
+        configs.extend(MatrixConfig('standard', 'py', os, '3.11') for os in BASIC_PLATFORMS)
+
+        # Minimal deps tests
+        configs.append(MatrixConfig('minimal', 'py', MAIN_PLATFORM, '3.11', uv_options='--no-dev'))
```
aaron-siegel: What's the harm in keeping these around on PRs? I'd argue that if we can add more coverage to PRs without impacting speed or reliability, we should (and I wouldn't expect these configs to impact either).

aaron-siegel (changes requested):

.github/workflows/pytest.yml
```diff
@@ -1,14 +1,14 @@
...
 
-# On PR updates, the basic test configurations are run:
-#   - Python 3.11 on ubuntu-24.04, macos-15, and windows-2025 runners.
-#   - A minimal installation of Pixeltable (--no-dev) with Python 3.11 on ubuntu-24.04.
-#   - An installation of Pixeltable with the 'otel' extra and Python 3.10 on ubuntu-24.04.
-#   - An installation of Pixeltable with all required dependencies pinned to their minimums.
+# On PR updates, only fast on-push checks are run:
 #   - Linting, type-checking, docstring validation, etc.
-#   - Random ops.
+#   - The 'slim' subset of the unit tests (SLIM_TESTS in tool/ci_tool.py) with Python 3.11 on ubuntu-24.04.
```
aaron-siegel: This is no longer accurate.

aaron-siegel (approved):
I feel like the logic in ci_tool is getting complex enough that we need a unit test to verify that it's producing the right job configurations :)

(A bug in ci_tool that causes an unintended weakening of the configurations list would be a coverage disaster)

sergey-mkhitaryan:
@aaron-siegel regarding the testing, good point, let me do it in a follow up.

PR #1528, opened 2026-08-14
Schema types are now non-nullable by default, following standard type conventions

aaron-siegel:
Deprecates `pxt.Required` and makes non-nullable the default for all type specifiers, whether in UDFs, class-based schemas, or traditional `create_table()` schemas.

Instead of:
```
col1: pxt.Required[pxt.Int]
col2: pxt.Float
```
one now writes:
```
col1: pxt.Int
col2: pxt.Float | None
```

- Modifies type signatures of various schema-accepting functions to generalize from `type` to `TypeForm` in order to accommodate `| None` syntax.
- Upgrades our mypy plugin to handle the above properly in various exotic cases.
- Changes all schemas in all tests to be semantically equivalent to what they were before (`Required[T]` -> `T`; `T` -> `T | None`)
- Changes notebooks only as necessary to get the notebooks to pass (`Required[T]` -> `T` always; `T` -> `T | None` only as needed)


pixeltable_cli/client/commands/schema.py
```diff
@@ -22,7 +22,7 @@
 
 
 class Docs(TableModel, name='docs'):
-    title: pxt.Required[pxt.String]           # a stored column; without Required it is nullable
+    title: pxt.String                         # a stored column
```
mkornacker, line 25: Also include examples of nullable columns, with explanatory comment, in both examples. That'll help agents.

mkornacker (approved):

pixeltable_cli/client/commands/schema.py
aaron-siegel: Done.

pixeltable_cli/client/commands/schema.py
```diff
@@ -22,7 +22,7 @@
...
     body: pxt.String
     title_upper = pxtf.string.upper(title)    # a computed column: an assignment, not an annotation
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

PR #1533, opened 2026-08-15
Detailed error reporting from daemon in tests

sergey-mkhitaryan:
Pixeltable's daemon's test mode (`--test`) enables detailed error reporting.

Also: CLI daemon's logs are correctly produced as test artifacts.


pixeltable/service/proxy_dispatch.py
```diff
@@ -92,19 +93,21 @@ def handle(request_json: str, request_parts: list[bytes]) -> tuple[str, list[byt
...
-            # the client only gets the message; keep the diagnostic detail (e.g. an evaluation stack trace)
-            # for whoever reads the server logs
+            # Log error details in the server log
             _logger.info('Error detail handling %s.%s:\n%s', request.class_name, request.method, e.detail)
         _logger.info('%s.%s error (%.2fs)', request.class_name, request.method, time.monotonic() - t0)
         error_dict = e.to_dict()
         error_dict['message'] = _restore_upload_names(error_dict['message'], request._uploaded_names)
         if 'cause' in error_dict:
             error_dict['cause'] = _restore_upload_names(error_dict['cause'], request._uploaded_names)
+        if include_error_detail and e.detail is not None:
```
mkornacker: This directly contradicts the comment on Error.detail.

Instead of fishing it out here, it would be better to change Error.to_dict() to something like to_dict(with_detail: bool = False).

mkornacker (approved):

PR #1535, opened 2026-08-15
PXT-1297: fix async teardown

mkornacker:
…ping them

What it fixes
- Every threaded plan execution created a thread, a Runtime, an event loop and a set of SDK clients, and closed none of them. The clients' sockets were left to GC, which then reported Unclosed client session and Event loop is closed from destructors. This grew with every query in notebooks and async servers.
- reset_runtime() closed the loop first and then dropped the clients, so they were never closed at all.
- BFL registered an atexit hook per client that ran asyncio.run(session.close()) on a long-dead loop; Jina never closed its session.

Changes
- Runtime.close_clients(): closes each cached client on the loop it was created on. It is the only code that removes entries from _clients, so no client can be dropped unclosed.
- Runtime.close(): cancels pending tasks, closes the clients, then shutdown_asyncgens + shutdown_default_executor before closing the loop. A loop we adopted (Jupyter's) is left running, but its clients are still closed on it.
- ExecNode._thread_iter(): runs on an asyncio.Runner and closes the thread's clients while its loop is alive (the fix from #1462).
- reset_runtime(): closes clients before the loop, and also retires the run_coro executor thread's runtime.
- close_threadpool_runtimes(): each worker thread of a pool closes its own runtime, the only thread that safely can. Used by reset_runtime() and by FastAPIRouter shutdown, whose worker threads previously leaked a loop and a client set per pool.
- FastAPIRouter.__shutdown is now idempotent: FastAPI invokes it twice per app shutdown.
- BFL and Jina open their aiohttp session lazily and expose close(); the atexit hook is gone.


pixeltable/exec/exec_node.py
```diff
@@ -147,22 +147,25 @@ def run() -> None:
...
+                    try:
+                        runner.run(produce())
+                        result_queue.put(ExecNode._THREAD_QUEUE_SENTINEL)
+                    except BaseException as e:
+                        result_queue.put(e)
+                    finally:
+                        # the clients are bound to this loop, so they have to be closed while it is open
+                        runner.run(thread_runtime.close_clients())
             finally:
                 telemetry.exit_context(hooks_token)
```
sergey-mkhitaryan: let's get rid of a nested try, and do this in (what currently is) the nested finally. `exit_context()` is pretty well protected from raising anything. Ideally we should turn it into a ContextManager instead, but that doesn't have to be in this pr.

pixeltable/runtime.py
```diff
@@ -287,6 +354,29 @@ def get_runtime() -> Runtime:
...
+    """
+    # one task per worker slot, each waiting at the barrier until all of them run:
+    # - no thread can take a second task, so the calls land on n_workers distinct threads.
+    # - in a thread pool with a thread count below the maximum, submitting one task per *existing* thread would
+    #   not reach those that are still busy (since they're busy, the pool simply starts up more threads for the cleanup
+    #   task)
+    barrier = threading.Barrier(n_workers)
+
+    def close_runtime() -> None:
+        barrier.wait()
```
sergey-mkhitaryan: let's wait with a timeout, we don't want to get completely stuck here

pixeltable/serving/_fastapi.py
```diff
@@ -363,6 +371,14 @@ def add_api_route(self, path: str, *args: Any, **kwargs: Any) -> None:
...
 
     def __shutdown(self) -> None:
+        # FastAPI calls this more than once per app shutdown
+        if self._is_shut_down:
+            return
+        self._is_shut_down = True
+        if len(self._jobs) > 0:
+            # a background job ran, so a worker thread has a runtime to close; the calls queue behind the
+            # in-flight requests, so that the loops and clients are closed only after workers stop using them
+            close_threadpool_runtimes(self._executor, _N_BACKGROUND_WORKERS)
```
sergey-mkhitaryan, line 584: the existing behavior is that we simply cancel the ongoing tasks, if any. here you are effectively changing it to wait for all ongoing tasks to complete. I don't know if that makes a material difference for anything, but it would be safer to cancel all tasks in the thread pool before calling `close_threadpool_runtimes()`.

pixeltable/runtime.py
```diff
@@ -301,15 +391,9 @@ def reset_runtime() -> None:
...
-                loop = runtime._event_loop
-                if not loop.is_running():
-                    loop.close()
-            except Exception:
-                pass
-        runtime._clients.clear()
         if runtime._run_coro_executor is not None:
-            runtime._run_coro_executor.shutdown(wait=False)
+            close_threadpool_runtimes(runtime._run_coro_executor, _N_RUN_CORO_WORKERS)
+            runtime._run_coro_executor.shutdown(wait=True)
```
sergey-mkhitaryan: should `Runtime.close()` do this as well?

sergey-mkhitaryan (approved):

pixeltable/runtime.py
mkornacker: Done.

pixeltable/serving/_fastapi.py
mkornacker: Done.

pixeltable/runtime.py
mkornacker: Done.

pixeltable/exec/exec_node.py

PR #1537, opened 2026-08-17
Add api for patching secrets on org/db

amithadke:
goal is to add/replace/delete secrets in single api call.

Also adding simple version of pxt secret cli 
We can use this cli/api in org update or db update too

### DIFF
--- a/pixeltable_cli/client/commands/secret.py
+++ b/pixeltable_cli/client/commands/secret.py
@@ -1,0 +1,96 @@
+"""`pxt secret {list,set,rm} <uri>` - manage org- and database-scoped runtime secrets."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+
+from pixeltable_cli.utils import split_pxt_uri
+
+from ..parser import Parser
+from ..utils import get_request, post_request
+
+EPILOG = """\
+Examples:
+  pxt secret list pxt://myorg
+  pxt secret list pxt://myorg:mydb
+  pxt secret set  pxt://myorg OPENAI_API_KEY=sk-... ANTHROPIC_API_KEY=sk-...
+  pxt secret rm   pxt://myorg:mydb OLD_KEY STALE_KEY
+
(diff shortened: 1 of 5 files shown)


pixeltable_cli/client/commands/secret.py
```diff
@@ -0,0 +1,96 @@
...
+    p = sub.add_parser('list', help='list secret names in a scope (never their values)')
+    p.add_argument('uri', help='Scope URI: pxt://org or pxt://org:db')
+    p.add_argument('--json', action='store_true', dest='json_output', help='Emit JSON output')
+
+    p = sub.add_parser('set', help='add or replace secrets')
+    p.add_argument('uri', help='Scope URI: pxt://org or pxt://org:db')
+    p.add_argument('assignments', nargs='+', metavar='KEY=VALUE')
+    p.add_argument('--json', action='store_true', dest='json_output', help='Emit JSON output')
+
+    p = sub.add_parser('rm', help='remove secrets')
```
aaron-siegel, line 39: `remove` or `delete` instead of `rm`? What terminology do other secrets managers use?

pixeltable_cli/models.py
```diff
@@ -198,6 +198,13 @@ class SchemaUpdateBody(BaseModel):
     allow_destructive: bool = False
 
 
+class PatchSecretsBody(BaseModel):
+    org: str
+    db: str | None = None
+    set: dict[str, str] = {}
```
aaron-siegel, line 204: maybe call this set_secrets (and delete_secrets); `set` is a python keyword and this isn't a set in that sense :)

pixeltable_cli/models.py
```diff
@@ -198,6 +198,13 @@ class SchemaUpdateBody(BaseModel):
...
+    delete: list[str] = []
```
aaron-siegel, line 205: I'd align this with our above terminology (standardize on 'delete' or 'remove' or 'rm')

PR #1538, opened 2026-08-18
Operational tables: new schema features and better testing

sergey-mkhitaryan:
1. Basic schema evolution support for operational tables (column add and drop, alter column type)
3. Class-based schema support for operational tables
4. Primary key support for operational tables
5. A test decorator that forks the test and runs it against two table types: data-versioned and operational.
6. The existing tests that can run against an operational tables are converted; TestOperationalTable is altogether dropped.
7. Some missing test coverage added for both table types.
8. Side effect: internal Pixeltable APIs (Catalog, etc.) do not take a `primary_key` parameter in addition to `list[Column]`. The property of belonging to a PK is already encoded in `Column` as `Column.is_pk`. This makes the interfaces simpler and more honest regarding the order of columns in the actual PK: it follows the order of columns in the schema, not the order of columns in the `primary_key` list.


pixeltable/catalog/model.py
```diff
@@ -489,6 +490,7 @@ def __prepare__(  # type: ignore[override]
         media_validation: Literal['on_read', 'on_write'] = 'on_write',
         comment: str | None = None,
         custom_metadata: Any = None,
+        _is_data_versioned: bool = True,
```
mkornacker: is_data_versioned

pixeltable/catalog/table_version_handle.py
```diff
@@ -96,11 +96,13 @@ class ColumnHandle:
     def get(self) -> 'Column':
         tv = self.tbl_version.get()
         if self.col_id not in tv.cols_by_id:
-            schema_version_drop = tv._tbl_md.column_md[self.col_id].schema_version_drop
+            drop_info = ''
+            if tv.is_data_versioned:
+                schema_version_drop = tv._tbl_md.column_md[self.col_id].schema_version_drop
```
mkornacker: This makes it look like the ColumnMd record got removed, but I'm not seeing code for that. Did I miss anything?

pixeltable/catalog/model.py
```diff
@@ -549,6 +551,14 @@ def __prepare__(  # type: ignore[override]
                         f'`model_base()` as `{cls_name}`.',
                     )
 
+            # A view is always of the same kind as its base. Operational views aren't supported yet.
```
mkornacker: This needs to differentiate:
- the view uses a different mode (you can't do that)
- the view wants to be operational (we can't do that yet)

pixeltable/catalog/local_table.py
mkornacker: -> int

tests/test_table.py
```diff
@@ -3006,6 +3013,98 @@ def test_delete(
             img_t.delete(where=img_t.img.width > 100)
         assert 'not expressible' in str(excinfo.value)
 
+    @versioned_and_operational
+    def test_crud(self, make_catalog_path: Callable[[str], str], is_data_versioned: bool) -> None:
```
mkornacker: Does this add coverage beyond test_insert/update/etc.?

tests/test_table.py
```diff
@@ -3006,6 +3013,98 @@ def test_delete(
...
+        assert t.select(t.val_x2).where(t.id == 5).collect()['val_x2'] == [10.0]
+
+        # all of it survives a catalog reload
+        reload_catalog()
+        t = pxt.get_table(p('crud'))
+        assert t.count() == 4
+        assert t.select(t.id, t.val_x2).order_by(t.id).collect()['id'] == [2, 3, 4, 5]
+
+    @versioned_and_operational
+    def test_add_drop_column(self, make_catalog_path: Callable[[str], str], is_data_versioned: bool) -> None:
```
mkornacker, line 3032: Same comment here: don't we already have that covered with existing test methods?

tests/test_table_model.py
```diff
@@ -410,6 +411,88 @@ class NoDefaultsV3(TableModelV3, name='no_defaults_table', has_default_idxs=True
         ):
             TableModelV3.diff_all(root)
 
+    def test_operational_table_model(self, make_catalog_path: Callable[[str], str]) -> None:
```
mkornacker: Why is this a test instead of applied to every existing test method?

mkornacker (changes requested):
This seems to have very little added test surface, in light of what should now be enabled.
It also doesn't test that operation table's metadata still has a non-None version.

pixeltable/catalog/model.py
sergey-mkhitaryan: not a public feature yet, let's keep it "private"

tests/test_table.py
sergey-mkhitaryan: This one exercises the interaction of different operations on one table, in one test. Also note that test_insert/update/delete cannot run on operational tables yet because operational tables do not yet support PK constraint.

tests/test_table.py
sergey-mkhitaryan: this one adds coverage:
* value correctness in the face of column add and drop
* name reuse after drop

but also the existing tests that have a similar coverage cannot run on operational tables.

tests/test_table_model.py
sergey-mkhitaryan: converted other tests instead. the one with views is a bit ugly, but that's probably fine.

pixeltable/store.py
```diff
@@ -169,6 +169,10 @@ def _create_system_columns(self) -> list[sql.Column]:
             self._pk_cols = rowid_cols
             return rowid_cols
 
+    def pk_constraint_name(self, tbl_version: catalog.TableVersion) -> str:
+        """Name of the Postgres object enforcing the Pixeltable primary key."""
+        return f'pk_idx_{tbl_version.id.hex}' if tbl_version.is_data_versioned else f'pk_{tbl_version.id.hex}'
```
mkornacker: Comment: why

pixeltable/store.py
```diff
@@ -214,28 +218,34 @@ def create_sa_tbl(self, tbl_version: catalog.TableVersion | None = None) -> None
...
-                    pk_idx_exprs.append(col.sa_col)
-            idx_name = f'pk_idx_{tbl_version.id.hex}'
-            idxs.append(
-                sql.Index(
-                    idx_name,
-                    *pk_idx_exprs,
-                    unique=True,
-                    postgresql_using='btree',
-                    postgresql_where=(self.v_max_col == schema.Table.MAX_VERSION),
+            pk_name = self.pk_constraint_name(tbl_version)
```
mkornacker, line 224: Why can't we have a uniform name?

tests/test_alter_column.py
mkornacker, line 70: What about all the other tests?

tests/conftest.py
```diff
@@ -507,8 +514,17 @@ def clean_db(drop_md_tables: bool = False) -> None:
 
 
 @pytest.fixture(scope='function')
-def test_tbl(make_catalog_path: Callable[[str], str]) -> pxt.Table:
-    return create_test_tbl(make_catalog_path('test_tbl'))
+def test_tbl(make_catalog_path: Callable[[str], str], request: pytest.FixtureRequest) -> pxt.Table:
```
mkornacker: What about all the other table fixtures?

tests/test_index.py
mkornacker, line 636: Why is this test excluded from coverage?

mkornacker (approved):
Re test coverage: I was expecting basically every test that does a create_table() to get the is_data_versioned parameter.

What is still missing to get this test coverage?

pixeltable/store.py
sergey-mkhitaryan: as in, the same name for versioned and operational tables? the versioned tables, we already call it "pk_idx_XYZ" and that needs to stay for compatibility reasons. For operational tables, it shouldn't be "idx" because it's not an index.

tests/test_alter_column.py
sergey-mkhitaryan: `test_alter_column_errors` is the only one here not parametrized. It doesn't need to be cause it covers the validation that is completely independent of the table type.

tests/conftest.py
sergey-mkhitaryan: oh yes, good catch

PR #1540, opened 2026-08-19
slimtest target extended to include more core tests

sergey-mkhitaryan:
`test_operational_table` removal is intentional -- this test is going away soon.

`TestDashboard` (in `test_smoke.py`) requires the dashboard. `build-dashboard` in the test matrix became always True, so I got rid of it.

### DIFF
--- a/tool/ci_tool.py
+++ b/tool/ci_tool.py
@@ -24,47 +24,41 @@ VERY_EXPENSIVE_PYTEST = "-m 'not benchmark and not cloud_e2e'"
 
 # The core-functionality test modules that PR checks run on every push. This mirrors the `slimpytest` target in the
 # Makefile.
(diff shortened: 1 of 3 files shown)


tool/ci_tool.py
```diff
@@ -24,23 +24,41 @@
...
-        'function',
-        'index',
-        'operational_table',
-        'snapshot',
-        'table',
-        'table_model',
-        'types',
-        'view',
-    )
+SLIM_TESTS = (
```
aaron-siegel, line 27: I'd find this more maintainable / less error prone if it was defined as a tuple and then `join`ed, rather than rely on a `' '` to be present at the end of every string.

aaron-siegel (approved):

PR #1542, opened 2026-08-20
pxt service

mkornacker:
Replaces `pxt serve` and its TOML service config with a declarative CLI over an application file: a Python file whose `FastAPIRouter` instances declare services, and whose models name the tables their routes serve. `pxt serve` and `[pixeltable.serving]` config are gone.

`pxt service diff | update | run | prune | stop | list | example`

# New abstractions
- `ServiceSpec` / `RouteSpec` (`serving/_spec.py`): what a file declares about a service, without
  naming a deployment. Two equal specs serve the same contract, which is what `compare_specs()`
  in `serving/_diff.py` uses to decide up-to-date, restart, or destructive.
- `ServiceDeployment` (`serving/service_deployment.py`): the record of one running service, one
  JSON file per service in a tree mirroring the catalog directory its models bind against. Owns
  start, stop, health and liveness. Derived, not authoritative: a record's liveness is its
  process's liveness, identified by pid plus process creation time, so a crashed service is
  absent from the next read with no cleanup pass.
- `service_runner` (`serving/service_runner.py`): the worker entry point that `start()` launches,
  which binds the port, writes the record, and serves.
- `ModelQuery` (`catalog/model/query.py`): a query written against models rather than tables, so
  a query route can be declared before its tables exist and bound when the service starts.
- Application-file loading (`utils/module_loader.py`, `utils/app_module.py`): loads a file that no
  module path can reach, under a name derived from its path, and resolves its imports against
  installed modules first and its own directory second. UDFs and queries defined in such a file
  are stored as a file plus a qualname, so a computed column over one resolves in any later
  process. Moving the file makes the column read-only, and a hosted target is refused, since it
  cannot read a local file.
- `utils/process.py`: pid liveness and identity, including the Windows probe that `os.kill()`
  cannot provide.


docs/release/howto/deployment/infrastructure.mdx
```diff
@@ -66,7 +66,7 @@ project/
 │   ├── data.py            # CRUD routes + queries for data pipeline
 │   ├── search.py          # Search routes + queries
 │   └── agent.py           # Agent routes (declarative + hand-written)
-├── pyproject.toml         # Dependencies and pxt serve config
+├── app.py                 # Models and services, run by pxt service
```
aaron-siegel, line 69: Isn't `pyproject.toml` (and/or `pixeltable.toml`) still relevant?

pixeltable/catalog/model/query.py
```diff
@@ -0,0 +1,200 @@
...
+
+import pixeltable as pxt
+from pixeltable import catalog, exceptions as excs, exprs
+from pixeltable.exprs import ColumnRefByName
+from pixeltable.query_clauses import FromClause
+
+from .declaration import MODEL_BY_DECLARED_TBL_ID, TableModelMeta
+
+
+class ModelQuery(pxt.Query):
```
aaron-siegel: I don't think it's wise to inherit from pxt.Query here; they're really two different things and I'd advocate for keeping them separate (just as TableModel doesn't inherit from Table, etc., by design). In general, I find that subclasses that break their ancestors' contracts (or fulfill them only weakly) are confusing and invite bugs.

I'd like to better understand the need for this - why conflate them? If there are overlapping bits of functionality that are important to abstract away, a better approach IMO is to introduce a QueryBase or some such, that encapsulates the shared functionality.

pixeltable/catalog/model/query.py
```diff
@@ -0,0 +1,200 @@
...
+    def _collect(self, args: dict[str, Any] | None = None, *, media_as_urls: bool = False) -> Any:
+        raise self._unbound('collect()')
+
+    async def _acollect(self, args: dict[str, Any] | None = None) -> Any:
+        raise self._unbound('collect()')
+
+    def cursor(self) -> Any:
+        raise self._unbound('cursor()')
+
+    def show(self, n: int = 20) -> Any:
```
aaron-siegel: See above - better just not to subclass from pxt.Query in the first place.

pixeltable/catalog/view.py
```diff
@@ -58,8 +58,12 @@ def _display_name(self) -> str:
         return 'table'
 
     @classmethod
-    def validate_view_query(cls, query: Query, *, is_snapshot: bool = False) -> None:
-        """Verify that a view can be defined by query."""
+    def validate_view_query(cls, query: Query, *, is_snapshot: bool = False, prefix: str = '') -> None:
+        """Verify that a view can be defined by query.
+
+        Args:
+            prefix: prepended to the message of every error raised here, to identify what is being defined.
```
aaron-siegel: Call this `error_prefix`

pixeltable/service/management_protocol.py
```diff
@@ -338,7 +338,7 @@ class GetServiceRunResponse(BaseModel):
 class OrgRecord(BaseModel):
     org_id: str
     org: str
-    default_db: Optional[str] = None
+    default_db: str | None = None
```
aaron-siegel, line 341: I wonder if there is a linting rule we can enable for this.

aaron-siegel (changes requested):
Another high level design comment. I'll continue going through some of the details but with 5000 lines of code I'm going to focus on the high level and approve once we've aligned on that :)

pixeltable/catalog/model/query.py
mkornacker: Done.

pixeltable/catalog/model/query.py
mkornacker: Done.

pixeltable/catalog/view.py
mkornacker: Done.

docs/release/howto/deployment/infrastructure.mdx
mkornacker: Out of scope.

pixeltable/service/management_protocol.py
mkornacker: Something to look into.

aaron-siegel (approved):

PR #1549, opened 2026-08-20
[PXT-880] Make table, column, and database identifiers case-insensitive

sergey-mkhitaryan:
The overarching principle behind this change is that the name folding (i.e. lower-casing) happens as early as possible, ideally right at the front door. This allows the Pixeltable internals to be mostly oblivious of this matter.

However a different principle applies to class-based schema:
- Python bindings (e.g. column declaration): names stay case sensitive so that these Python declarations are still an idiomatic Python. Only when the declaration crosses the domain boundary to Pixeltable (`create_all()`, `update_all()`), the names are finally folded.
- Everything else (e.g. `name='Foo'`, `EmbeddingIndex(name='Idx')`, etc.) that at no point exists as a Python binding, is folded immediately upon arrival.

Columns are folded on import. `import_csv`, `import_pandas`, parquet, etc. lower-case column names. There is a new failure mode in those: Pixeltable is no longer able to import from a data source whose columns collide after folding, e.g. 'a' and 'A'.

Database names are also case insensitive now. `Env` implements a fallback so as not to completely break the existing dbs (i.e. the db name is `MyDb`, but suddenly Pixeltable creates an empty `mydb` in its place).

A breaking change: only ASCII characters are allowed in Pixeltable table/column/db/etc. names.

Metadata conversion that can refuse to run. It validate that all visible names in the catalog are valid according to the new rules, and that folding them will not create duplicates. An error is raised if there are any problems. If not, the converter folds the identifiers in the catalog. Except the database name itself -- it stays the same.

`to_pydantic()` now matches model fields case-insensitively.

JSON paths stay case-sensitive -- they are data rather than Pixeltable identifiers.


docs/release/platform/type-system.mdx
```diff
@@ -281,6 +281,34 @@ t.where(t.result == None).select(
 ).collect()
 ```
 
+## Identifier names
```
aaron-siegel: Is type_system.mdx really the most logical place for this?

docs/release/platform/type-system.mdx
```diff
@@ -281,6 +281,34 @@ t.where(t.result == None).select(
...
+# every casing resolves to the same table and column ...
+t = pxt.get_table('mydir.mytable')
+t.MyCol, t.mycol, t['MYCOL']       # all the same column
+
+# ... but the stored (and displayed) name is the folded one
+t.columns()                         # ['mycol']
+t.select(t.MyCol).collect()         # result key: 'mycol'
+```
+
+Names must also be **ASCII**. Column and index names must be valid Python identifiers that don't start with `_`,
```
aaron-siegel: "ASCII" is a bit jargony here. "Table, column, and index names must use alphanumeric characters and be valid Python identifiers, and cannot start with _" - something like that?

pixeltable/catalog/model/declaration.py
```diff
@@ -116,6 +133,10 @@ class EmbeddingIndex:
     precision: Literal['fp16', 'fp32'] = 'fp16'
     name: str | None = None
 
+    def __post_init__(self) -> None:
+        if isinstance(self.name, str):
+            object.__setattr__(self, 'name', fold_identifier(self.name))
```
aaron-siegel, line 156: What is `object.__setattr__()`? Is that different from plain `setattr()`?

Why wouldn't one just do `self.name = fold_identifier(self.name)` here? Needs explanation.

pixeltable/catalog/model/declaration.py
```diff
@@ -420,14 +442,17 @@ def add_reserved_column_ref(
         """Add `name` as a reserved column (it is resolvable in the class body, and its symbol cannot be reused,
         but it does not have a ColumnSpec and will not be included in the list of columns for the view to create).
         """
-        self.reserved_cols[name] = kind
-        super().__setitem__(name, ColumnRefByName(name, col_type))
+        folded_name = fold_identifier(name)
```
aaron-siegel: I think we should be case-folding in the select list / iterator columns (i.e., this function should only see inputs that have already been normalized). Otherwise we have divergent conventions (Tables are case-insensitive; Queries are case-sensitive), which could cause problems: What if I create a Query that has two select list entries that are distinct, but have identical folds, and try to instantiate a View on top of that?

(cf. our discussion at standup today.)

pixeltable/catalog/globals.py
```diff
@@ -121,18 +123,47 @@ def validated(cls, param_val: str, param_name: str) -> IfNotExistsParam:
             ) from None
 
 
+def fold_identifier(name: str) -> str:
+    """Fold an identifier to its stored form.
+
+    Note: Pixeltable only allows ASCII characters in identifiers, so we don't need to worry about any exotic characters
```
aaron-siegel: remove "Note: " (everything is a note)

pixeltable/catalog/globals.py
```diff
@@ -121,18 +123,47 @@ def validated(cls, param_val: str, param_name: str) -> IfNotExistsParam:
...
+    return name.lower()
+
+
+def fold_mapping_keys(m: Mapping[str, _T]) -> dict[str, _T]:
+    """Fold the keys of a user-supplied column mapping, raising INVALID_SCHEMA if keys collide once folded."""
+    result: dict[str, _T] = {}
+    for name, spec in m.items():
+        folded_name = fold_identifier(name)
+        if folded_name in result:
+            spellings = ', '.join(repr(n) for n in m if fold_identifier(n) == folded_name)
```
aaron-siegel: I found this hard to read because of the variable names `n` and `m` (which my brain is trained to interpret as natural numbers).

pixeltable/catalog/globals.py
```diff
@@ -121,18 +123,47 @@ def validated(cls, param_val: str, param_name: str) -> IfNotExistsParam:
...
+            )
+        result[folded_name] = spec
+    return result
+
+
 def is_valid_identifier(name: str, *, allow_hyphens: bool = False) -> bool:
+    """Identifiers are restricted to ASCII characters: on ASCII, str.lower(), str.casefold() and SQL LOWER() all behave
+    in the same way.
+    """
+    if not name.isascii():
```
aaron-siegel: Isn't this implicit in `adj_name.isidentifier()` below?

pixeltable/catalog/model/declaration.py
```diff
@@ -459,7 +484,7 @@ def set_col_value(self, name: str, value: Any) -> None:
             spec = {'value': expr}
         self.known_cols[name] = spec
         # Add the column to the namespace so that it can be referenced in subsequent expressions in the class body.
-        super().__setitem__(name, exprs.ColumnRefByName(name, col_type_from_spec(spec)))
+        super().__setitem__(name, exprs.ColumnRefByName(fold_identifier(name), col_type_from_spec(spec)))
```
aaron-siegel: Can we do the folding in the ColumnRefByName initializer, rather than at every call site?

pixeltable/catalog/model/diff.py
```diff
@@ -202,18 +203,19 @@ def user_columns(model: TableModelMeta) -> dict[str, ColumnSpec]:
             specs[item.name] = {'value': item, 'stored': False}
         for col_name, expr in named_items.items():
             specs[col_name] = {'value': expr, 'stored': not isinstance(expr, exprs.ColumnRefByName)}
-    return specs
+    return fold_mapping_keys(specs)
 
 
 def base_query_columns(model: TableModelMeta) -> set[str]:
-    """Names of the columns a model's base query projects via its `select()` clause (empty if there is none)."""
+    """Folded names of the columns a model's base query projects via its `select()` clause (empty if there is none)."""
```
aaron-siegel: Same comment as above - shouldn't the select_list_exprs already be normalized? Why do we need to do additional folding here?

pixeltable/catalog/local_table.py
```diff
@@ -206,7 +208,7 @@ def __hash__(self) -> int:
         return hash(self._tbl_version_path.tbl_id)
 
     def __getattr__(self, name: str) -> 'exprs.ColumnRef':
-        col_md = self._tbl_version_path.get_column_md_by_name(name)
+        col_md = self._tbl_version_path.get_column_md_by_name(fold_identifier(name))
```
aaron-siegel: Again, can we do the normalization in the function, rather than at the call sites?

pixeltable/catalog/local_table.py
```diff
@@ -566,6 +569,7 @@ def do_add_computed_column() -> UpdateStatus:
                 'add_computed_column', '`col_name=col_type` or `col_name=expression`', kwargs
             )
             col_name, spec = next(iter(kwargs.items()))
+            col_name = fold_identifier(col_name)
             if not is_valid_identifier(col_name):
```
aaron-siegel, line 572: I'd do this check before folding, so that the error message reflects what the user inputs. (Same comment may apply in various other places.)

That is, if I do
```
t.add_computed_column(_MyCol=...)
```
I'd expect the error to read
```
Invalid column name: _MyCol
```
not
```
Invalid column name: _mycol
```

pixeltable/catalog/local_table.py
```diff
@@ -631,7 +635,8 @@ def do_drop_column() -> None:
             if_not_exists_ = IfNotExistsParam.validated(if_not_exists, 'if_not_exists')
 
             if isinstance(column, str):
-                col = self._tbl_version_path.get_column(column)
+                col_name = fold_identifier(column)
+                col = self._tbl_version_path.get_column(col_name)
```
aaron-siegel: Move the folding from the call site into the getter?

I'm going to stop repeating this comment now; consider it general feedback that applies everywhere: normalizing inside the accepting function has two advantages:
- simpler code (we don't have to repeat `fold_identifier` at every call site)
- reduces the potential for bugs if someone forgets a `fold_identifier` somewhere or other. (likewise, every function that assumes its input is normalized should be asserting that.)

pixeltable/catalog/path.py
```diff
@@ -51,12 +51,18 @@ def __post_init__(self) -> None:
         if not all(is_valid_identifier(c, allow_hyphens=True) for c in self.components):
             raise excs.RequestError(excs.ErrorCode.INVALID_PATH, f'Invalid path: {"/".join(self.components)}')
 
+        if self.org is not None:
+            object.__setattr__(self, 'org', fold_identifier(self.org))
```
aaron-siegel, line 56: Why `__setattr__` here? Why not just assignment?

pixeltable/io/table_data_conduit.py
```diff
@@ -161,17 +162,24 @@ def add_table_info(self, table: pxt.Table) -> None:
...
+        mapped_col_to_original_col: dict[str, str] = {}
         for col_name in columns:  # FIXME
-            mapped_col_name = self.source_column_map.get(col_name, col_name)
-            col_name_set.add(mapped_col_name)
+            mapped_col_name = fold_identifier(self.source_column_map.get(col_name, col_name))
+            if mapped_col_name in mapped_col_to_original_col:
+                raise excs.RequestError(
+                    excs.ErrorCode.INVALID_SCHEMA,
+                    f'Pixeltable column names are case-insensitive; columns '
+                    f'{mapped_col_to_original_col[mapped_col_name]!r} and {col_name!r} both denote column '
```
aaron-siegel: Instead of "both denote column xyz", how about just "are the same"?

pixeltable/metadata/converters/convert_55.py
```diff
@@ -0,0 +1,215 @@
...
+
+import sqlalchemy as sql
+
+import pixeltable.exceptions as excs
+from pixeltable.metadata import register_converter
+from pixeltable.metadata.converters.util import convert_table_md, convert_table_schema_version_md
+from pixeltable.metadata.schema import Dir, Table, TableSchemaVersion
+
+
+@register_converter(version=55)
```
aaron-siegel, line 12: There's a ton of business logic in here and I didn't read it in detail; let's make sure all code paths are tested (i.e. the v55 db dump should exercise them)

pixeltable/env.py
```diff
@@ -81,6 +81,22 @@ def store_app_name() -> str:
...
+
+def validate_db_name(db_name: str) -> None:
+    from pixeltable.catalog.globals import is_valid_identifier
+
+    if not is_valid_identifier(db_name, allow_hyphens=True):
+        raise excs.RequestError(excs.ErrorCode.INVALID_CONFIGURATION, f'Invalid database name: {db_name!r}')
+    if len(db_name) > MAX_DB_NAME_LEN:
+        raise excs.RequestError(
+            excs.ErrorCode.INVALID_CONFIGURATION,
+            f'Database name is too long ({len(db_name)} characters; the limit is {MAX_DB_NAME_LEN})',
```
aaron-siegel: Suffix with `: {db_name}`

pixeltable/row.py
```diff
@@ -14,6 +17,30 @@ class CellError(TypedDict):
     errormsg: str
 
 
+class _FoldedKeyMapping(Mapping[str, _V]):
```
aaron-siegel: Why do we need this? `Row` is part of the public API and IMO this complicates it by returning a bespoke collection class rather than a simple `dict` which users can easily understand. What is the user story where they benefit from getting one of these back instead of a dict via `row.errors`?

For that matter I'm not convinced `Row` itself needs any changes. It's really meant as a more efficient `dict` with simple, `dict`-like semantics.

tests/test_index.py
```diff
@@ -361,10 +361,10 @@ def test_add_index_after_drop(
         p = make_catalog_path
         t = small_img_tbl
         sample_img = t.select(t.img).head(1)[0, 'img']
-        t.add_embedding_index('img', idx_name='clip_idx', embedding=local_embed)
+        t.add_embedding_index('img', idx_name='Clip_Idx', embedding=local_embed)
```
aaron-siegel: I'd be happier if we isolated case-insensitivity into a separate test, rather than shoe-horning it at random into other tests. (Keeps the tests clean, self contained, and easy to digest)

Same comment for the tests above. Rather than modify existing tests, create a `test_case_insensitive_dir`, `test_case_insensitive_index`, etc., like with column below.

tests/test_path.py
```diff
@@ -232,3 +232,38 @@ def test_path_delimiter_str_hash_compare(self) -> None:
         assert str(dotted_appended) == 'a/b/c/d'
         assert str(unix_appended) == 'a/b/c/d'
         assert dotted_appended.components == unix_appended.components == ('a', 'b', 'c', 'd')
+
+    def test_non_ascii_identifiers(self) -> None:
+        assert not is_valid_identifier('café')
+        assert not is_valid_identifier('Ω')
+        for path in (('Café'), ('dir/Café'), ('pxt://CAFÉ/tbl'), ('pxt://variata:CAFÉ/tbl')):
```
aaron-siegel: Let's not hard code customer names in our codebase (even in tests)

tests/test_snapshot.py
```diff
@@ -116,7 +117,9 @@ def test_basic(self, make_catalog_path: Callable[[str], str]) -> None:
         # the time of creating a snapshot will raise an error now.
         tbl = create_test_tbl(name=tbl_path)
         assert 'c1' in tbl.columns()
-        with pxt_raises(pxt.ErrorCode.COLUMN_ALREADY_EXISTS, match="Column 'c1' already exists in the base table"):
+        with pxt_raises(
+            pxt.ErrorCode.COLUMN_ALREADY_EXISTS, match=re.escape("Columns ('c1') already exist in the base table")
```
aaron-siegel: I find this error message a lot more confusing in its new form.

aaron-siegel (changes requested):
`tool/create_test_db_dump.py` needs changes, to add case-insensitive schema objects that exercise the corresponding migration paths.

pixeltable/catalog/model/declaration.py
sergey-mkhitaryan: Neither of `setattr()` and `self.xyz` works on a frozen dataclass. Added comments. The docs [recommend](https://docs.python.org/3/library/dataclasses.html#frozen-instances) `__setattr__`.

pixeltable/catalog/globals.py
sergey-mkhitaryan: not really, python allows some non-alphanumeric chars in identifiers https://docs.python.org/3/reference/lexical_analysis.html#lexical-names-nonascii

pixeltable/catalog/path.py
sergey-mkhitaryan: a frozen dataclasses, same as above. added a comment.

pixeltable/catalog/model/declaration.py
sergey-mkhitaryan: you're right, we should and do case fold in those columns. the name folding here was completely unnecessary here.

pixeltable/catalog/model/declaration.py
sergey-mkhitaryan: good call

pixeltable/catalog/local_table.py
sergey-mkhitaryan: noted

docs/release/platform/type-system.mdx
sergey-mkhitaryan: Per Alison, a big docs refactoring is coming, so it doesn't matter too much. Followed her suggestions.

pixeltable/metadata/converters/convert_55.py
sergey-mkhitaryan: Updating the dump here https://github.com/pixeltable/pixeltable/pull/1566

pixeltable/row.py
sergey-mkhitaryan: Per discussion, reverted this -- row is case sensitive again.

pixeltable/catalog/globals.py
```diff
@@ -3,7 +3,8 @@
 import dataclasses
 import enum
 import itertools
-from typing import TYPE_CHECKING, Any, Mapping, NamedTuple, cast
+from collections.abc import Mapping
```
aaron-siegel: Why this change? Is this alias somehow different from `typing.Mapping`?

pixeltable/catalog/globals.py
aaron-siegel: Ok.

In that case I'd find the logic easier to follow if `not name.isascii()` were incorporated into the return clause below

pixeltable/catalog/model/diff.py
```diff
@@ -213,7 +214,8 @@ def base_query_columns(model: TableModelMeta) -> set[str]:
         return set()
     # "anonymous" compound expressions are not allowed here, so every unnamed item names a column
     assert all(expr.is_column_ref for expr, name in base.select_list if name is None)
-    return {expr.default_column_name() if name is None else name for expr, name in base.select_list}
+    names = {expr.default_column_name() if name is None else name for expr, name in base.select_list}
```
aaron-siegel: ? Why this change?

pixeltable/catalog/globals.py
```diff
@@ -120,19 +123,47 @@ def validated(cls, param_val: str, param_name: str) -> IfNotExistsParam:
...
+    """
+    if not name.isascii():
+        return False
     # If allow_hyphens=True, we allow hyphens to appear in the name, but we still do not permit a name to start with one
     adj_name = name.replace('-', '_') if allow_hyphens else name
     return adj_name.isidentifier() and not name.startswith('-') and not name.startswith('_')
 
 
 def is_system_column_name(name: str) -> bool:
+    """name must already be folded."""
```
aaron-siegel, line 157: Is this still true?

If so, we should assert it, and why call `fold_identifier` below?

pixeltable/service/proxy_daemon.py
```diff
@@ -141,6 +143,8 @@ def start(db: str, test_mode: bool = False) -> str:
         **os.environ,
         'PIXELTABLE_HOME': str(proxy_home(db)),  # own media/tmp + port.lock
         'PIXELTABLE_PGDATA': pgdata,  # shared postmaster
+        # intentionally using as typed db name, without folding: Env has a fallback for legacy (case-sensitive)
```
aaron-siegel: It took me awhile to understand this comment; the meaning of "using as typed db name" is unclear (I eventually figured it out).

Much clearer would be "intentionally using the raw db name without case-folding" or some such.

But do we even need to worry about this? How many users have existing Pixeltable databases with custom, mixed-case overrides of `PIXELTABLE_DB`? My guess is zero.

pixeltable/service/proxy_protocol.py
```diff
@@ -295,7 +295,8 @@ def _serialize(obj: Any, sink: PartSink) -> Any:
             'v': {
                 'schema': {name: t.as_dict() for name, t in obj._col_types.items()},
                 'rows': [[_serialize(val, sink) for val in row._data] for row in obj],
-                'errors': [row.errors for row in obj],
+                # row.errors is a Mapping, but pydantic needs a dict
```
aaron-siegel: I thought we agreed to turn it back into a dict.

pixeltable/serving/_fastapi.py
```diff
@@ -40,6 +40,7 @@
 import pixeltable as pxt
 from pixeltable import catalog, exceptions as excs, exprs, func, type_system as ts
 from pixeltable.catalog import model
+from pixeltable.catalog.globals import fold_identifier
```
aaron-siegel: Import from `catalog` rather than `catalog.globals` if we're outside the `catalog` package? (otherwise why import them up to `catalog` in the first place?)

aaron-siegel (approved):
Approved with various additional minor comments

pixeltable/catalog/model/diff.py
sergey-mkhitaryan: there was folding here at some point, but then it got moved. reverting.

pixeltable/catalog/globals.py
sergey-mkhitaryan: good catch

pixeltable/service/proxy_daemon.py
sergey-mkhitaryan: Rephrased. I don't know about the existing users with mixed-cased db names, you may be right. I do know that there's a demand for supporting multiple pixeltable dbs on one host. Let's keep it for now, it's easy and even less risky to get rid of it later.

pixeltable/service/proxy_protocol.py
sergey-mkhitaryan: missed this

pixeltable/serving/_fastapi.py
sergey-mkhitaryan: fixed here + one other place

PR #1554, opened 2026-08-24
constraint the preshed version

sergey-mkhitaryan:
Motivation: fix this error `OverflowError: Python int too large to convert to C ssize_t` when using `spacy`. E.g. https://github.com/pixeltable/pixeltable/actions/runs/32755189701/job/97521083893

Trigger: appears to be a Cython release https://cython.readthedocs.io/en/latest/src/changes.html#id2

`preshed` (a transitive dependency of `spacy`) publishes no wheels for Python 3.13+ before 3.0.10. Compiled from its sdist instead, `PreshMap` rejects keys >= 2**63 with `OverflowError: Python int too large to convert to C ssize_t`, which breaks every spaCy model load. 3.0.13 provides cp313 and cp314 wheels.

Note: `preshed` is not a direct Pixeltable dependency, which is why this change uses `constraint-dependencies` instead of updating the dev group.

Note: `preshed` 3.0.13 released in March 2026 https://pypi.org/project/preshed/#history


pyproject.toml
```diff
@@ -114,6 +114,9 @@ override-dependencies = [
     "protobuf>=5.29.6",  # Required because 'fireworks-ai' hard-codes 'protobuf==5.29.4'
     # "huggingface-hub>=1",  # Required due to 'whisperx'
 ]
+constraint-dependencies = [
```
aaron-siegel: Can't we just add this as a normal dev dependency?

aaron-siegel (changes requested):

pyproject.toml
```diff
@@ -151,6 +151,7 @@ dev = [
     "notebook>=7.5.6",
     "black>=26.3.1",
     "pixeltable-doctools @ git+https://github.com/pixeltable/pixeltable-doctools",
+    "preshed>=3.0.13 ; python_version < '3.14'",
```
aaron-siegel: add a comment explaining why it's necessary?

aaron-siegel (approved):

PR #1564, opened 2026-08-28
Enable `TableModelMeta` to work on Python 3.14+ without `from __future__ import annotations`

aaron-siegel:
The change relies on CPython as the interpreter, but I expect Pixeltable is already heavily dependent on CPython irrespective of this change.


pixeltable/catalog/model/declaration.py
```diff
@@ -243,18 +244,27 @@ class _ModelNamespace(dict):
     reserved_cols: dict[str, Literal['base query', 'iterator']]
 
     # The scope in which the class body is defined; used to evaluate stringized type annotations (see
-    # set_col_type). Populated from the defining frame in TableModelMeta.__prepare__.
-    eval_globals: dict[str, Any]
-    eval_locals: dict[str, Any]
+    # set_col_type) and prebind annotations.
```
mkornacker: We typically refer to function calls with '()'.

mkornacker (approved):
This is missing the test apps in tests/pixeltable_cli/apps.

pixeltable/catalog/model/declaration.py
```diff
@@ -323,7 +337,7 @@ def set_col_type(self, name: str, type_: Any) -> None:
             # PEP 649 otherwise defers annotation evaluation entirely -- annotations arrive as strings. Evaluate
             # the string in the scope where the model class is defined to recover the actual type.
```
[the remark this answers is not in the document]
aaron-siegel: Fixed.

pixeltable/catalog/model/declaration.py
aaron-siegel: Fixed. I also put this in CLAUDE.md.

PR #1566, opened 2026-08-29
[PXT-880] Add mixed-case schema objects to the test db dump

sergey-mkhitaryan:
Pixeltable identifiers are becoming case-insensitive, stored case-folded, but the metadata converter
that folds them has nothing to fold.

Rename a directory, a component view, a table and its columns, and one computed column per table, and
name one embedding index. Add a column that is dropped again, so that a name survives only in a
historical schema version. Regenerate the version 55 dump.


aaron-siegel (changes requested):
I'd prefer that new entities be introduced with the mixed-case logic, leaving the existing entities as-is so that they're consistent across db dumps.

This would simplify things; we'd no longer have to parameterize the _run_v15_tests and _verify_v45 checks; those would operate consistently across dumps. Instead, just add a new _verify_v55(). (Best: rename _verify_c7_null() to _verify_v55() and put the mixed case validation there.)

aaron-siegel (approved):

PR #1568, opened 2026-08-30
pxt db

mkornacker:
- pxt db diff/update pxt://org:db reconciles a hosted database with its [[pixeltable.database]]
  entry: secrets first, then the artifacts, then capacity; pxt db build-image builds unconditionally
- an image and a project archive replace the single bundle: the image holds the environment and is
  keyed by image_digest, the archive holds the project files and is keyed by archive_digest, so a
  source edit ships seconds of archive and a lockfile edit rebuilds the image as well
- ProjectFingerprint is the one manifest behind both, answering image_needed/archive_needed/
  restart_needed and naming the causes per scope
- a service instance records the project it loaded; pxt service diff reports a restart when it is
  stale, and blocks on pxt db update when the database's project is behind the working copy
- new pod entry point, python -m pixeltable.serving.pod_runner: fetch the archive, verify the digest
  it was given, unpack it, serve one service, report the fingerprint it loaded
- management protocol: BUILD_IMAGE, SET_PROJECT, GET_PROJECT, GET_PROJECT_UPLOAD_URL and
  REPORT_SERVICE_INSTANCE replace UPDATE_RUNTIME and the bundle operations; the build request carries
  python_version, system_dependencies, deps_type, uv_options and image_digest, so the archive holds
  no metadata.json
- the schema, service and db verbs move out of the CLI bridge into pixeltable/catalog/model/schema.py,
  pixeltable/serving/service.py and pixeltable/service/db.py; bridge.py is dashboard data only
- one set of pydantic types in pixeltable_cli/types.py, used by core and the CLI, replacing the two
  parallel sets of TypedDicts
- tests/pixeltable_cli/test_db.py drives every pxt db verb and the pod against a hosted database,
  skipped unless PXTTEST_CLOUD_DB_URI is set
- needs a lockstep control-plane release: the deployed one predates the renamed operations, answers an
  absent database with 401 rather than 404, and does not yet store the fingerprint or serve
  REPORT_SERVICE_INSTANCE


docs/release/platform/cli.mdx
```diff
@@ -606,6 +607,62 @@ pxt service update app.py my_dir    # serve this file's services against them
...
+
+app = fastapi.FastAPI()
+
+@app.post('/notes')
+def add_note(note_id: int, text: str) -> dict[str, int]:
+    return {'rows': Notes.insert([{'note_id': note_id, 'text': text}]).num_rows}
+```
+
+```bash
+pxt schema update notes_app.py my_dir     # create the tables for the models in notes_app.py
```
aaron-siegel, line 632: I still think taking a Python "file" as an input is a big Python antipattern.

pixeltable/catalog/model/schema.py
```diff
@@ -0,0 +1,154 @@
...
+from pixeltable_cli.types import CheckReport, SchemaChangeOp, SchemaPlan, TableDiff
+from pixeltable_cli.utils import PxtPath
+
+from .declaration import TableModelMeta
+from .diff import PY_DESTRUCTIVE_HINT
+
+_DESTRUCTIVE_HINT = "Re-run 'pxt schema update' with --allow-destructive to apply these changes."
+
+
+def schema_diff(app_file: str, catalog_dir: PxtPath) -> SchemaPlan:
```
aaron-siegel, line 15: Seems like an antipattern to me; I thought we'd decided to key on modules.

aaron-siegel (approved):

PR #1573, opened 2026-09-01
New Catalog locking

sergey-mkhitaryan:
Replaces the locking mechanism in Catalog with Postgres table locks (LOCK TABLE) on store tables, for both data-versioned and operational tables. The tables-row lock (SELECT … FOR UPDATE NOWAIT + UPDATE lock_dummy) is gone in all of its uses.

**Design doc is a part of the pr: docs/design/catalog-locking.md**

# How it works

- Catalog exposes transaction context managers like `begin_read_xact()`, `begin_write_xact()`, `begin_schema_change_xact()`, and decorators `retry_schema_change_loop`, `retry_read_loop`, etc., each of which acquires the appropriate table locks. Metadata-only reads take no locks.
- Readers are never blocked by writers, and writers are never blocked by readers for either table kind.
- One global lock acquisition order — the whole lock set sorted by store table name — ensures no deadlocks. Directory locks are a second phase, `dirs` rows sorted by catalog path, and must come after the store table locks because a row lock pins the snapshot.
- A write's lock set depends on metadata that may only be read after locking, so it is guessed (from the metadata cache, or read from the store in a separate transaction), locked, then validated against the current metadata. On a mismatch, another attempt is made using the up-to-date lock set.
- Store table creation and drop are done transactionally with the table metadata creation and deletion. This establishes the invariant: a committed tables row implies an existing store table, and vice versa. `CreateStoreTableOp` and `DropStoreTableOp` are gone.

# Behavior changes

## Operational tables

- concurrent writes to one table no longer serialize on the table
- reads and writes fail fast during a schema change (new error: `SCHEMA_CHANGE_IN_PROGRESS`)

## Data-versioned tables

- write exclusion moves from the tables row to EXCLUSIVE on the store table
- contended writes wait in PG lock queue instead of polling

No metadata version bump. `Table.lock_dummy` is no longer used and removed from `schema.py`, but the column stays in already existing databases.

# Breaking changes

TODO

# Other changes

Stress test aka random ops: re-enable rename_view: it works now.

`RowidRef` serializes table effective version in addition to table id.


pierrebrunelle:
grok review

pierrebrunelle (commented):
This redesign replaces `tables.lock_dummy` with Postgres `LOCK TABLE` on store tables, with a clear global order (store tables by name, then `dirs` by path), guess-lock-validate for writes, and fail-fast vs wait split by table kind. The `TestLocking` matrix plus stale-guess / dir-phase / cold-cache scenarios are strong evidence the core protocol works.

Before merge, three gaps need closing: (1) design doc §2 claims `@pxt.query` UDF tables are locked via `begin_read_xact()`, but PXT-1343 and the reentrant `_begin_xact` early-return show they are not, and tests only `filterwarnings` the miss; (2) design doc §8 still describes `_make_lock_set()` / `assert_rows_read_locked()` that do not match the code; (3) the PR body’s Breaking changes section is still `TODO`, and `catalog.py` still has a leftover `# TODO continue from here`.

### Prioritized findings
1. **PXT-1343 vs design doc.** Nested `begin_*_xact` joins an open transaction without adding the new lock targets. A `@pxt.query` UDF therefore reads unlocked store tables during computed-column / insert evaluation. Either fix the hole or rewrite §2 and the PR body to name PXT-1343 as an explicit known limitation.
2. **Breaking changes `TODO`.** Fill before merge: `SCHEMA_CHANGE_IN_PROGRESS`, ORM `Table.lock_dummy` removal (column retained in existing DBs), removal of `CreateStoreTableOp` / `DropStoreTableOp`, and any public concurrency behavior callers must handle.
3. **Design doc §8 stale.** No `_make_lock_set()`, mixed kinds are allowed and handled by `_lock_set_blocking`, and the read check is `check_rows_read_locked` (warning), not `assert_rows_read_locked`.
4. **WIP / process markers.** Delete `# TODO continue from here` and the “try to split up” process note in the locking path.
5. **DeadlockDetected retry.** Protocol claims deadlock-freedom via global sort, yet `_is_retryable_exc` still retries `DeadlockDetected` with an open investigate TODO. Prove impossible under the order, or document the retry as intentional belt-and-suspenders.

### Test asks
- PXT-1343 concurrency: while an insert runs a `@pxt.query` UDF against table `Q`, run a schema change on `Q` and assert exclusion or the documented failure mode (not only `filterwarnings`).
- Path race under `write_paths`: move/rename of `a/t` racing with drop/create between guess and lock; assert `_StaleLockSetError` recovery.
- Mixed-kind wait policy: write an operational target with a data-versioned `read_tbl_keys` entry; assert the whole set waits when a schema change holds the operational store table.

Core lock order, guess-lock-validate, operational vs versioned wait policy, and transactional store create/drop look solid.

**Verdict: needs changes before merge.**

docs/design/catalog-locking.md
```diff
@@ -0,0 +1,257 @@
...
+
+### Metadata-only reads take no locks
+
+`pxt.get_table()`, `list_tables()`, `get_dir_contents()` and `describe()` read catalog metadata, not rows, and
+therefore lock nothing. They may observe a schema change in progress and report a schema that is about to change,
+which is better than blocking or failing; any operation the caller then performs on the returned handle locks
+properly.
+
+The distinction is per call site, not per entry point. A query's expressions can pull in tables whose data is read,
+through a `@pxt.query` UDF for instance, and those go through `begin_read_xact()`, which locks them.
```
pierrebrunelle, line 75: Delete or rewrite this claim. PXT-1343: a `@pxt.query` UDF’s table is not in the outer lock set, and a nested `begin_read_xact()` is reentrant so it acquires nothing. Either fix the protocol or state the gap here.

docs/design/catalog-locking.md
```diff
@@ -0,0 +1,257 @@
...
+
+## 8. How the protocol is enforced
+
+Asserts catch a wrong operation class, or a statement reached without the lock it needs, instead of leaving
+either to code review:
+
+- `_lock_tables()`: the targets are sorted by name, and none is already locked.
+- `_assert_md_write_locked()`: a metadata write holds a **self-conflicting** mode on the store table, `EXCLUSIVE` or
+  `ACCESS EXCLUSIVE`, depending on the table kind.
+- `assert_rows_write_locked()` / `assert_rows_read_locked()`: called from `StoreBase` and `SqlNode`, which own every
```
pierrebrunelle, line 254: Rename to `check_rows_read_locked`. It warns today (PXT-1343); it is not an assertion.

docs/design/catalog-locking.md
```diff
@@ -0,0 +1,257 @@
...
+  statement that reads or writes store rows.
+- `_make_lock_set()`: the set does not span both table kinds (§2).
```
pierrebrunelle, line 256: Delete this bullet. There is no `_make_lock_set()`, and mixed kinds are intentional: `_lock_set_blocking` waits for the whole set when any table is data-versioned.

pixeltable/catalog/catalog.py
```diff
@@ -415,51 +679,93 @@ def begin_xact(
...
         if for_write and _frozen.get():
             raise ProhibitedWriteError()
         read_tvps = read_tvps or []
         write_tvps = write_tvps or []
-        read_tbl_ids = read_tbl_ids or []
-        write_tbl_ids = write_tbl_ids or []
+        read_tbl_keys = read_tbl_keys or []
+        write_tbl_keys = write_tbl_keys or []
+        write_paths = write_paths or []
         if get_runtime().in_xact:
```
pierrebrunelle, line 694: Reentrant `begin_*_xact` yields the open connection without merging the new targets into `_locks_held`. That is the PXT-1343 hole. Escalate locks on reentry, or require callers to name every table up front and assert on nested targets.

pixeltable/catalog/catalog.py
```diff
@@ -557,47 +882,648 @@ def _try_rollback(self, conn: sql.Connection) -> None:
...
+    def assert_rows_write_locked(self, tv: TableVersion) -> None:
+        """Verifies that the transaction holds a lock appropriate for writing rows to this table."""
+        mode = _tbl_lock_mode(_TblOpClass.DATA_WRITE, tv.is_data_versioned)
+        store_tbl_name = tv.store_tbl._storage_name()
+        assert self._is_locked(store_tbl_name, mode), (store_tbl_name, mode, self._locks_held)
+
+    def check_rows_read_locked(self, tv: TableVersion) -> None:
+        """Verifies that the transaction holds a lock appropriate for reading rows from the table.
+
+        TODO(PXT-1343): once fixed, this warning should become an assertion.
```
pierrebrunelle, line 1415: Keep this ticket visible in the design doc and PR body until it is an assert. Silencing it with `filterwarnings` across tests hides a protocol miss.

pixeltable/catalog/catalog.py
```diff
@@ -557,47 +882,648 @@ def _try_rollback(self, conn: sql.Connection) -> None:
...
+                if tree_result is None:
+                    return None
+                add(tree_result[0])
+                any_data_versioned |= tree_result[1]
+        return _LockSet(
+            tbl_targets=tuple(targets[name] for name in sorted(targets)),
+            blocking=_lock_set_blocking(op_class, any_data_versioned),
+        )
+
+    # TODO continue from here
```
pierrebrunelle, line 1225: Delete this leftover `# TODO continue from here`.

pixeltable/catalog/catalog.py
```diff
@@ -557,47 +882,648 @@ def _try_rollback(self, conn: sql.Connection) -> None:
...
+        # Metadata-only reads take no locks
+        if op_class is _TblOpClass.MD_READ:
+            return _LockSet()
+
+        lock_set: _LockSet | None = None
+
+        # Attempt to build the lock set from the metadata cache. That is only possible if no write paths are requested,
+        # because Catalog does not cache paths or directory structures. Operations with write paths are create/drop
+        # table/dir, and move. They always fall back to the store read in a dedicated transaction, which is more
+        # expensive, but shouldn't happen too frequenly.
```
pierrebrunelle, line 1256: Typo: `frequenly` → `frequently`.

pixeltable/catalog/catalog.py
```diff
@@ -106,53 +107,97 @@ def _is_retryable_exc(e: BaseException) -> bool:
     # TODO: Investigate whether DeadlockDetected points to a bug in our locking protocol, which is
```
pierrebrunelle, line 107: Either show deadlocks cannot occur under the global store-name / dir-path order, or rewrite this as intentional belt-and-suspenders. An open “investigate whether the protocol is buggy” under a deadlock-free claim is not merge-ready.

pixeltable/catalog/catalog.py
```diff
@@ -557,47 +882,648 @@ def _try_rollback(self, conn: sql.Connection) -> None:
...
+            if isinstance(e.orig, psycopg.errors.LockNotAvailable):
+                assert not blocking
+                raise _LockNotAvailableError from e
+            if isinstance(e.orig, psycopg.errors.UndefinedTable):
+                raise _StaleLockSetError from e
+            raise
+        for store_tbl_name in store_tbl_names:
+            self._locks_held[store_tbl_name] = mode
+
+    # TODO this looks like 2 independent functions in one. Try to split up.
```
pierrebrunelle, line 1469: Delete this process note. Split in a follow-up if needed; do not leave review narration in the locking path.

pixeltable/exceptions.py
```diff
@@ -88,6 +88,7 @@ def __new__(cls, value: int, http_status: int, is_retryable: bool) -> Self:
     # ConcurrencyError (7xxx)
     SERIALIZATION_FAILURE = 7000, 409, True
     CONCURRENT_MODIFICATION = 7001, 409, True
+    SCHEMA_CHANGE_IN_PROGRESS = 7002, 409, True
```
pierrebrunelle, line 91: Call out `SCHEMA_CHANGE_IN_PROGRESS` in the Breaking changes section: operational reads/writes that used to wait or serialize now fail fast with this code.

pierrebrunelle (commented):
Follow-up after a deeper pass. Still needs changes before merge.

Additional blockers beyond the first COMMENT:

1. Guess-lock-validate: when a rebuilt lock set equals the failed guess, the code raises `excs.Error(INTERNAL_ERROR)` (~728–733). A concurrent create-then-drop can restore the original set legitimately, and `Error.__init__` asserts if you construct `Error` itself. Raise `InternalError` only for true impossibilities; retry when the set is still valid.
2. Pending-op upgrade: `CreateTableMdOp` / `DeleteTableMdOp` now require `is_view`, and `CreateStoreTableOp` / `DropStoreTableOp` are gone. Pre-PR pending ops cannot deserialize. Default `is_view` or keep tombstones, and write Breaking changes.
3. Create paths: `paths=[path]` is only set on replace (~2465). Ordinary create leaves the parent `dirs` X-lock to `_prepare_dir_op()` after SELECTs, so create is not under guess-lock-validate. Pass `paths=[path]` for every create (table, view, dir).
4. Leftover WIP: `# TODO continue from here` (~1225), comments naming `_lock_set()` / `_make_lock_set()` / `assert_rows_read_locked()` that do not exist, and `_prepare_dir_op()` still documenting the old parent-before-child protocol (~2251).

Also: `check_rows_read_locked()` still warns (PXT-1343) while the design doc names an assert; design doc §8 `_make_lock_set()` bullet does not match `_lock_set_blocking()`.

pixeltable/catalog/tbl_ops.py
```diff
@@ -154,43 +135,43 @@ def undo(self, tv: TableVersion | None) -> None:
...
-    """Undo-only log record"""
+    """Represents a new table metadata write and a store table creation.
+
+    The forward flow is inlined in Catalog; this serves as an undo log entry only.
+    """
 
     needs_tv: ClassVar[bool] = False
     needs_xact: ClassVar[bool] = True
 
+    is_view: bool
```
pierrebrunelle, line 146: Give `is_view` a default, or keep tombstone classes for `CreateStoreTableOp` / `DropStoreTableOp`. `TableOp.from_dict()` will raise on any pending op serialized before this PR. Put that in Breaking changes.

docs/design/catalog-locking.md
pierrebrunelle, line 256: Delete the `_make_lock_set()` bullet, or restore that check. `_lock_set_blocking()` already waits if any table is data-versioned; the code does not refuse mixed kinds.

pixeltable/catalog/catalog.py
```diff
@@ -415,51 +679,93 @@ def begin_xact(
...
+                    write_paths=write_paths,
+                    lock_path_subtree=lock_path_subtree,
+                )
+            except PendingTableOpsError as e:
+                if not finalize_pending_ops:
+                    raise
+                pending_ops_tbl_id = e.tbl_id
+                continue
+
+            if prev_failed_lock_set is not None and attempt_lock_set == prev_failed_lock_set:
```
pierrebrunelle, line 728: Retry a rebuilt lock set even when it equals the failed guess. A concurrent create then drop of a view restores the original set, and that set is now correct.

pixeltable/catalog/catalog.py
```diff
@@ -415,51 +679,93 @@ def begin_xact(
...
+                # The previous lock set was deemed stale, but re-reading the metadata produced the same lock set. This
+                # shouldn't happen, and no progress can be made by retrying.
+                raise excs.Error(
```
pierrebrunelle, line 731: Raise `InternalError`, not `Error`. `Error.__init__` asserts on `cls is not Error`.

pixeltable/catalog/catalog.py
pierrebrunelle, line 1225: Delete `# TODO continue from here`.

pixeltable/catalog/catalog.py
```diff
@@ -1598,7 +2461,10 @@ def _create_table(
         if additional_idxs is None:
             additional_idxs = []
 
-        @retry_loop(for_write=True)
+        @retry_schema_change_loop(
+            paths=[path] if if_exists in (IfExistsParam.REPLACE, IfExistsParam.REPLACE_FORCE) else [],
```
pierrebrunelle, line 2465: Pass `paths=[path]` for every create, not only replace. Same in `_create_view()` and `_create_dir()`.

pixeltable/catalog/catalog.py
```diff
@@ -557,47 +882,648 @@ def _try_rollback(self, conn: sql.Connection) -> None:
...
+        read_tbl_keys: Collection[TableVersionKey],
+        write_tvps: Collection[TableVersionPath],
+        write_tbl_keys: Collection[TableVersionKey],
+        lock_mutable_tree: bool,
+        write_paths: Collection[Path],
+        lock_path_subtree: bool,
+    ) -> _LockSet:
+        """The lock set for a transaction with these targets, warming up the metadata cache if it isn't sufficient.
+
+        Runs before the transaction opens, which is what keeps the store read out of it. _lock_set() issues no
```
pierrebrunelle, line 1240: Name `_lock_set_from_cache()` / `_lock_set_from_store()`, or drop the sentence. `_lock_set()` is gone.

PR #1575, opened 2026-09-02
[PXT-1327] Make /health endpoint async to avoid starvation under load

christopherpestano:
The /health endpoint was a sync def, so FastAPI ran it on the threadpool. Under heavy work the daemon starved these handlers, health checks timed out, and the pod was killed. Making it async lets it be scheduled directly on the event loop so it stays responsive during heavy work.

Also adds a test that simulates heavy work on the daemon by inserting and collecting 1M rows. It is marked skip for now because large collects need server-side streaming support before it can pass on cloud.

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -1119,6 +1119,15 @@ class TestTable:
         assert res2['c1'] == [r['c1'] for r in rows]
         assert res2['c2'] == [r['c2'] for r in rows]
 
+    @pytest.mark.skip(reason='Test fails on cloud without server-side streaming')
+    def test_bulk_scalar_collect(self, db_root: DatabaseRoot) -> None:
+        t = pxt.create_table(db_root.make_catalog_path('bulk'), {'id': pxt.Int, 'val': pxt.String})
+        for i in range(10):
+            t.insert([{'id': i * 100_000 + j, 'val': f'row {i}/{j}'} for j in range(100_000)])
+        assert t.count() == 1_000_000
+        res = t.collect()
+        assert len(res) == 1_000_000
+
     def test_insert_query(self, test_tbl: pxt.Table, db_root: DatabaseRoot) -> None:
         p = db_root.make_catalog_path
         t = test_tbl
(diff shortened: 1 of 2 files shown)


tests/test_table.py
```diff
@@ -1119,6 +1119,15 @@ def test_create_from_rows(self, db_root: DatabaseRoot) -> None:
         assert res2['c1'] == [r['c1'] for r in rows]
         assert res2['c2'] == [r['c2'] for r in rows]
 
+    @pytest.mark.skip(reason='Test fails on cloud without server-side streaming')
+    def test_bulk_scalar_collect(self, db_root: DatabaseRoot) -> None:
```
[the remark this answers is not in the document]
christopherpestano: It's only meaningful to test it against the cloud it's well established that an insert like this works locally.

tests/test_table.py
```diff
@@ -1119,6 +1119,15 @@ def test_create_from_rows(self, db_root: DatabaseRoot) -> None:
...
+        t = pxt.create_table(db_root.make_catalog_path('bulk'), {'id': pxt.Int, 'val': pxt.String})
+        for i in range(10):
+            t.insert(({'id': i * 100_000 + j, 'val': f'row {i}/{j}'} for j in range(100_000)))
```
aaron-siegel: unnecessary extra set of parens?

tests/test_table.py
```diff
@@ -1119,6 +1119,15 @@ def test_create_from_rows(self, db_root: DatabaseRoot) -> None:
         assert res2['c1'] == [r['c1'] for r in rows]
         assert res2['c2'] == [r['c2'] for r in rows]
 
+    @pytest.mark.skip(reason='Test fails on cloud without server-side streaming')
```
aaron-siegel: `@pytest.mark.db_roots('local', 'proxy', reason=...)` if we want this to run in local/proxy mode but skip cloud for now

aaron-siegel (approved):

tests/test_table.py
christopherpestano: Copilot suggested fix was off as usual

tests/test_table.py
christopherpestano: Sure but it's not super meaningful in local (or local proxy) considering we have perf tests that already do something similar. The regression only exists in cloud

mkornacker (approved):

PR #1576, opened 2026-09-02
Refactor `pxt db` CLI tests

aaron-siegel:
- New architecture for `pxt db` CLI tests: build a Pixeltable wheel as text fixture, include it in the archive for the test project, and refer directly to the wheel in `requirements.txt` (so that no special logic is required on the control plane)
- Reorganize `test_db.py` to avoid repeated unnecessary DB creation, reducing test running times
- Avoid overloaded use of `PXTTEST_CLOUD_DB_URI` in `test_db.py`

### DIFF
--- a/tests/pixeltable_cli/test_db.py
+++ b/tests/pixeltable_cli/test_db.py
@@ -7,7 +7,6 @@ which takes minutes. They run against the session's hosted database, the one the
 """
 
 import json
-import os
 import pathlib
 import shutil
 import uuid
@@ -16,8 +15,7 @@ from typing import Any
 import pytest
 
 from ..utils import DatabaseRoot
-
-from .conftest import PxtRunner
+from .conftest import PxtRunner, write_requirements
 
 pytestmark = [
     pytest.mark.remote_api,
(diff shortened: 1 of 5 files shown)


tests/pixeltable_cli/test_db.py
```diff
@@ -110,64 +88,58 @@ def get_target_ops(plan: dict[str, Any], target: str) -> list[dict[str, Any]]:
...
-        """build-image sends whatever the project holds and builds it."""
-        ops = cli('db', 'build-image', current_db, '--json', cwd=project, timeout=_BUILD_TIMEOUT).json
+        # `db build-image`: Sends whatever the project holds and builds it
+        ops = cli('db', 'build-image', test_db_uri, '--json', cwd=project, timeout=_BUILD_TIMEOUT).json
         assert [op['target'] for op in ops] == ['image']
         assert all(op['status'] == 'applied' for op in ops), ops
 
-    def test_errors(self, cli: PxtRunner, project: pathlib.Path, hosted_db: str) -> None:
-        create_project_config(cli, project, hosted_db)
+    def test_db_errors(self, cli: PxtRunner, project: pathlib.Path, test_db_uri: str) -> None:
```
mkornacker, line 144: Might even roll this into test_db_ops() to avoid having to create another db.

mkornacker (approved):

tests/pixeltable_cli/test_db.py
aaron-siegel: It doesn't actually create a DB (all the commands just fail fast) - that's why I left it separate.

PR #1578, opened 2026-09-03
Catalog-only daemon mode for hosted databases

amithadke:


pixeltable/config.py
```diff
@@ -891,6 +891,11 @@ def describe_setting(self, section: str, key: str) -> str:
         'daemon_port': ('Listen port for the proxy daemon in fixed-address mode (e.g. 8000)', int),
         'db_uri': 'Base pxt:// URI for remote catalog access (e.g. pxt://myorg:mydb)',
     },
+    'cli_server': {
```
mkornacker: Why do we need this in the config file?

PR #1579, opened 2026-09-03
PXT-1353: pxt service update keeps stable port

mkornacker:
- pxt service update keeps a restarted service on the port it was reached at, so its callers do not have to be
  redirected, and takes --port to pin one (it fails if the port is taken rather than moving to another);
  the port now shows up in list and diff output.
- update and diff take an optional service name, so one service of a multi-service file can be reconciled or
  compared on its own, which is also what makes --port usable there.
- Terminology: replaces 'declaration' with 'definition' where warranted


Makefile
```diff
@@ -252,7 +252,7 @@ linkscheck: docs
 
 .PHONY: clean
 clean:
-	@rm -f pixeltable/catalog/model.pyi pixeltable/catalog/model/declaration.pyi || true
+	@rm -f pixeltable/catalog/model.pyi pixeltable/catalog/model/definition.pyi || true
```
aaron-siegel: If you make this change, you also need to keep `declaration.pyi` on this line or `make clean` won't actually clean up source trees from before this commit.

aaron-siegel (changes requested):
I think "declaration" is correct for models. cf. this snippet from the Pydantic docs:
```
In order to declare a generic model, you perform the following steps:
```
For services I don't have a strong preference one way or the other.

aaron-siegel (approved):

Makefile
mkornacker: Done

PR #1581, opened 2026-09-03
PXT-1352: Performance improvements for proxy queries

mkornacker:
- collect() over a proxy now encodes its response body straight from the row iterator
  (Query._collect_content()), rather than materializing a ResultSet of Row objects and
  re-encoding it. The post-collect encoding phase drops from 53.0s to ~9.5s at 1M rows,
  with a byte-identical body.
- test_bulk_scalar_collect (1M rows of Int + String; ~30s end to end, almost all of it the
  10 x 100k inserts) now runs over the proxy: collect() takes 13.3s, of which 10.6s is
  server-side, against 8.3s local. It was previously local-only, marked "Fails on
  proxy/cloud without server-side streaming for large collects".
- head()/tail() build their rowid order-by on the client and ship it, so the daemon only
  ever runs collect. RowidRef carries each id's effective version, so a pinned path stays
  pinned, and the version is part of the ref's identity.
- Protocol: ProxyResponse is a TypedDict, encode_response() writes the body in one pass,
  run_query() lost its method parameter; PROTOCOL_VERSION 2 -> 3.


pixeltable/service/proxy_dispatch.py
```diff
@@ -666,6 +608,9 @@ def _encode_result_set(result: dict) -> dict:
     }
 )
 
+# methods whose handler returns the encoded response body, rather than a value for encode_response()
+_BODY_METHODS: frozenset[tuple[str, str]] = frozenset({('Query', 'collect')})
```
aaron-siegel, line 612: I'd find this easier to follow using standard Python entity references, e.g.,
```
'pixeltable._query.Query:collect'
```
rather than
```
`('Query', 'collect')`
```

I realize we're using those keys in the existing `_TABLE_HANDLERS` too so maybe it's not actionable in this PR, but it's something that occurred to me while reading this

pixeltable/service/proxy_protocol.py
```diff
@@ -212,18 +215,20 @@ class ProxyRequest(BaseModel):
...
+
+    A response sets either result or error, and current_md only after a mutation or a stale-md refusal.
+    """
+
+    result: Any  # return value
+    error: dict[str, Any]  # excs.Error.to_dict(), set instead of result on failure
 
     # serialized TableMdPath (list[TableVersionMd]); returned after a mutation so the client refreshes its md
-    current_md: Any = None
+    current_md: Any
```
aaron-siegel: Why isn't this one typed?

aaron-siegel (approved):

pixeltable/service/proxy_protocol.py
mkornacker: Done.

mkornacker: Done.

PR #1593, opened 2026-09-05
[PXT-1344] pxt db logs and pxt service logs

christopherpestano:
Adds two CLI verbs that read the logs of a hosted database and of a hosted service, so a failed deploy or
an opaque 500 can be diagnosed from the command line:

  pxt db logs pxt://org:db
  pxt service logs pxt://org:db/services/NAME

Both take --since (30s, 10m, 1h, 2d; default 1h), --tail N (the newest N lines in the window, default 200,
at most 10000), --include-health (keep the health probe lines, left out by default) and --json.

On the wire this is a new GET_LOGS management operation with its request and response models; the local
daemon forwards it to the control plane like the other hosted verbs. The control plane side is in
pixeltable-cloud (PR to follow), which reads the records back from CloudWatch. This PR also adds the
operation to the read-only set and gives it the longer timeout, since the server waits on a log query.


pixeltable/service/management_protocol.py
```diff
@@ -180,6 +181,40 @@ class GetArchiveUploadUrlResponse(BaseModel):
...
+class GetLogsRequest(BaseModel):
+    """Read what a database's pods logged: the database pod's, or one service's when service_name is given.
+
+    The log merges what the pods' processes logged with what they wrote to their console, by time.
+    """
+
+    operation_type: Literal[ManagementOperationType.GET_LOGS] = ManagementOperationType.GET_LOGS
+    org: str | None = None
+    db: str
+    service_name: str | None = Field(default=None, pattern=_K8S_NAME, max_length=63)
```
mkornacker: Services have paths (= names by themselves aren't unique).

pixeltable_cli/client/commands/service.py
```diff
@@ -164,7 +165,18 @@ class Docs(TableModel, name='docs'):
   'pxt service diff' answers that.
 {_OWN_APP}{_HOSTED}{_APP_FILE}"""
 
-VERBS = ('diff', 'update', 'run', 'prune', 'stop', 'list', 'check', 'example')
+LOGS_EPILOG = """\
+Examples:
+  pxt service logs pxt://acme:main/services/ingest                       # what its pods logged in the last hour
```
mkornacker: Needs to show paths, not just a name

Do we want to require the 'services' qualifier here? it's clear from context

This should work identically for locally running ones.

pixeltable_cli/client/main.py
```diff
@@ -32,8 +32,8 @@
     'daemon': 'control the daemon (start/stop/restart/status)',
     'localproxy': 'manage local proxy daemons (create/start/stop/delete)',
     'dashboard': 'print and open the dashboard URL',
-    'db': 'manage hosted databases (diff/update/create/list/status/start/stop/build-image/delete)',
-    'service': 'run the services an application file declares (diff/update/prune/stop/list/example)',
+    'db': 'manage hosted databases (diff/update/create/list/status/logs/start/stop/build-image/delete)',
```
mkornacker: While you're here, also remove 'create'.

mkornacker (changes requested):

pixeltable_cli/server/routes.py
```diff
@@ -809,6 +812,32 @@ def stop_db(req: Request) -> dict[str, Any]:
     return management_client.api_call(req.body(StopDbRequest))
 
 
+@router.get('/api/logs')
+def get_logs(req: Request) -> dict[str, Any]:
+    since = req.query_str('since', default='1h')
+    assert since is not None
+    since_seconds = parse_duration_str(since)
```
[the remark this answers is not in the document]
christopherpestano: Yeah, swapped it for `req.query_str('since') or '1h'`. The limit bounds are checked by `query_int` now too.

pixeltable/service/management_protocol.py
christopherpestano: Added base_path to disambiguate. The cloud then finds the service by path and name and then works off its id, so the name doesn't need the k8s pattern anymore also.

pixeltable_cli/client/commands/service.py
christopherpestano: Changed it to work like stop, so you can pass a bare name, dir/name, or pxt://org:db/dir/name, and an ambiguous name errors the same way. Dropped the services/ qualifier. Also, if the service is running locally it errors and tells you where the log file is, like you suggested.

pixeltable_cli/client/main.py
christopherpestano: Removed.

pixeltable/serving/service.py
```diff
@@ -146,6 +170,17 @@ def service_stop(names: list[str]) -> list[ServiceChangeOp]:
     return ops
 
 
+def service_logs(name: str, *, since_seconds: int, limit: int, include_health: bool) -> Sequence[LogRecord]:
+    """The log of the instance name addresses, the way service_stop() addresses one."""
```
mkornacker: Garbled/backwards LLM grammar, please fix.

pixeltable/serving/service.py
```diff
@@ -146,6 +170,17 @@ def service_stop(names: list[str]) -> list[ServiceChangeOp]:
     return ops
 
 
+def service_logs(name: str, *, since_seconds: int, limit: int, include_health: bool) -> Sequence[LogRecord]:
```
mkornacker: Everything else here accepts target: PxtPath. Why not this?

mkornacker (changes requested):
This is missing the logging functionality in the cli tests (ie, the basic test scenarios: db update, schema update, run some inserts, see them show up in logs; and the same for services).

pixeltable/serving/service.py
christopherpestano: Changed it to target: PxtPath.

pixeltable/serving/service.py
christopherpestano: Rewritten, and I swept the rest of the diff for the same problem.

tests/pixeltable_cli/test_service.py
```diff
@@ -635,6 +643,86 @@ def test_addressing(self, cli: PxtRunner, apps: Callable[[str], str], db_root: D
...
+
+        # the daemon validates --since and --tail before resolving the service
+        r = cli('service', 'logs', f'{target}/ingest', '--since', 'bogus', check=False)
+        assert r.returncode == 1 and 'must be a duration' in r.stderr, r.stderr
+        r = cli('service', 'logs', f'{target}/ingest', '--tail', '50000', check=False)
+        assert r.returncode == 1 and "'limit' must be <= 10000" in r.stderr, r.stderr
+        r = cli('service', 'logs', f'{target}/nosuch', check=False)
+        assert r.returncode == 1 and 'No service' in r.stderr, r.stderr
+
+        if db_root.id == 'cloud':
```
mkornacker, line 736: Check for the error message containing the log file path in the local case.

tests/pixeltable_cli/test_db.py
mkornacker, line 165: Add a test_local_logs_error() test case that tests the error msg.

mkornacker (approved):

tests/pixeltable_cli/test_service.py
christopherpestano: Done, the test now checks the message with the full log path.

tests/pixeltable_cli/test_db.py
christopherpestano: Added test_local_logs_error(). db logs on a local URI now raises an error naming the daemon log file, which the test checks.

PR #1598, opened 2026-09-07
Fix nightly MCP and optional notebook dependencies

pierrebrunelle:
## Summary

Nightly installs MCP 2.x even though the example server uses its v1 API, and sparse dependency configurations run tests without required optional packages. Cap MCP at `>=1.27.2,<2`, report child-server startup failures immediately with stderr, guard the mistune/spaCy-dependent tests, and install spaCy in the primary transformers configuration.

Fix isolated notebook dependencies with torchvision for SAM3 and `setuptools<82` for TensorFlow Hub's `pkg_resources` import. Skip observability only in pip-install notebook mode; retain source-based coverage using the workspace instrumentation package.

Clarify the public-API testing rule in AGENTS.md and GitHub agent instructions. Claude inherits AGENTS.md. Internal access is limited to focused internal unit tests or fixture setup unsupported by public APIs; the instructions do not require explanatory comments. This guidance-only follow-up passed diff checks; CI must rerun on the new head.


pierrebrunelle:
@sergey-mkhitaryan if you want to review this.

.github/instructions/tests.instructions.md
```diff
@@ -6,7 +6,8 @@ applyTo: "tests/**"
 - A test takes `db_root: DatabaseRoot` and builds paths with `db_root.make_catalog_path(...)`. Never hardcode a catalog path. 586 tests already do this.
 - A test using `uses_db` needs `@pytest.mark.db_roots('local', reason='...')`; `tests/conftest.py` raises a `UsageError` at collection without it. The reason names a PXT ticket, not a bare TODO. Delete the exclusion when the ticket lands.
 - Any test for `pxt.Error` or a subclass uses `pxt_raises()`, not `pytest.raises()`. Both always take `match=` to verify error text.
-- Assert on user-visible behavior through the public API, not `col.stored`, `ColumnRef`, or `TableVersion` internals. Use `Table.get_metadata()`, `t.describe()`, or queries.
+- Follow the Testing rule in `AGENTS.md`: exercise behavior through public SDK, CLI, or HTTP APIs and assert on public results, metadata, or errors. Do not inspect private attributes or mutate internal state when a public API covers the case.
+- Use internal APIs only for focused tests of internal components or fixture setup that public APIs cannot support; explain why they are needed and reuse shared test utilities where available.
```
sergey-mkhitaryan: > explain why they are needed

My experience with Claude, and I've heard the same from other people, is that it writes unnecessary, overly verbose comments. So asking it to explain anything by default seems like asking for a trouble.

.github/workflows/nightly.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         python-version: ["3.11", "3.13"]
         package-configs:
           - ""
-          - "lancedb pylance mcp opencv-python scenedetect"
+          - "lancedb pylance 'mcp>=1.27.2,<2' opencv-python scenedetect"
```
sergey-mkhitaryan, line 50: mcp 2.0 was released ~ 2 months ago. What is the path for us to support it? Note that it's an optional (dev) dependency, so we don't have to be very conservative with it even if it's backward incompatible.

docs/release/platform/embedding-indexes.ipynb
```diff
@@ -811,7 +811,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install -qU tensorflow tensorflow-hub tensorflow-text"
+    "%pip install -qU tensorflow tensorflow-hub tensorflow-text 'setuptools<82'"
```
sergey-mkhitaryan, line 814: same question as `mcp`. `setuptools` 81.0.0 is 7 month old already, and over time this restriction will be even less reasonable. Do we have a plan to support more recent versions?

scripts/prepare-nb-tests.sh
```diff
@@ -106,6 +106,11 @@ fi
 TARGET_DIR="$1"
 shift
 
+if [[ $DO_PIP_INSTALL == true ]]; then
+    # The instrumentation package is available in the source checkout, but is not yet published on PyPI.
```
sergey-mkhitaryan, line 110: please file a jira to address this later when the instrumentation is published

tests/test_mcp.py
```diff
@@ -102,11 +103,31 @@ def test_mcp_as_tools(self, db_root: DatabaseRoot, init_mcp_server: str) -> None
         assert res[0]['tool_calls'] == {'pixelmultiple': [str((7 + 22) * 9)], 'pixeldict': None}
 
 
-def _wait_for_port(port: int, timeout: float = 30.0) -> None:
+@contextmanager
+def _mcp_child(env: dict[str, str]) -> Iterator[tuple[subprocess.Popen[Any], Any]]:
+    # TemporaryFile, not PIPE: the server writes DEBUG logs and a full pipe stalls it.
```
sergey-mkhitaryan: Gotta love Claude answering questions that no one asked. I'd drop this comment, it's not that useful to me. What would be useful instead is to write a docstring explaining what this function yields.

tests/test_mcp.py
```diff
@@ -102,11 +103,31 @@ def test_mcp_as_tools(self, db_root: DatabaseRoot, init_mcp_server: str) -> None
         assert res[0]['tool_calls'] == {'pixelmultiple': [str((7 + 22) * 9)], 'pixeldict': None}
 
 
-def _wait_for_port(port: int, timeout: float = 30.0) -> None:
+@contextmanager
+def _mcp_child(env: dict[str, str]) -> Iterator[tuple[subprocess.Popen[Any], Any]]:
```
sergey-mkhitaryan: for the 2nd item, Claude says `IO[bytes]` is the type.

sergey-mkhitaryan (changes requested):

.github/instructions/tests.instructions.md
pierrebrunelle: Removed the blanket explanation requirement from AGENTS.md and both GitHub instruction files in 425865b14. The public-API rule and narrow internal-test/fixture exceptions remain, without asking agents to add comments.

tests/test_mcp.py
pierrebrunelle: Changed both stderr annotations to IO[bytes] and the process annotations to Popen[bytes] in 425865b14. Focused Ruff and strict mypy checks on the helpers pass.

.github/workflows/nightly.yml
pierrebrunelle: Agreed: the cap should be temporary, not our support policy. I checked MCP 2.2.0 in an isolated environment. Besides FastMCP -> MCPServer, v2 removes streamablehttp_client, which pixeltable/func/mcp.py imports in both discovery and execution. The replacement streamable_http_client yields two streams rather than three; port/stateless_http move to the server run call.

Proposed follow-up: migrate those two client paths and the example server together, run the existing public-API discovery/computation, persistence, and changed/missing-tool tests in local and proxy modes, then raise the dev requirement to v2, remove the nightly cap, and regenerate uv.lock. Keeping that compatibility change separate lets this PR repair the release nightly immediately. Migration reference: https://py.sdk.modelcontextprotocol.io/v2/migration/

docs/release/platform/embedding-indexes.ipynb
pierrebrunelle: This constraint is only in the TensorFlow Hub example install cell, not a Pixeltable runtime restriction. The latest tensorflow-hub release is still 0.16.1 and imports pkg_resources; upstream issue tensorflow/hub#934 and fix PR tensorflow/hub#935 are still open. Remove the bound when a released Hub version contains that fix, then execute the custom embedding UDF and index/query cells in fresh Python 3.11/3.13 environments with current setuptools. If Hub remains unreleased, replace this example with a maintained embedding implementation rather than expanding the pin. Upstream fix: https://github.com/tensorflow/hub/pull/935

scripts/prepare-nb-tests.sh
pierrebrunelle: The follow-up is drafted with acceptance criteria: confirm the instrumentation version required by pixeltable[otel] is published, remove the pip-mode skip, and pass isolated Python 3.11/3.13 notebook runs while retaining --no-pip coverage.

Will make a Jira Ticket

pierrebrunelle: Filed as [PXT-1427](https://pixeltable.atlassian.net/browse/PXT-1427).

Scoped it wider than the skip, because the skip is the symptom. `opentelemetry-instrumentation-pixeltable` 0.1.0 is not on PyPI at all (both `/pypi/.../json` and `/simple/` return 404), so `pip install 'pixeltable[otel]'` fails for anyone on a released wheel, which is the exact command `docs/release/platform/observability.ipynb` gives users. The `otel` CI leg runs `uv sync --no-dev --extra otel` and resolves from the uv workspace, so it never exercised the PyPI path either. This skip removes the last check that did.

Acceptance on the ticket: `pip install 'pixeltable[otel]'` succeeds from PyPI alone on 3.11 and 3.13, the notebook runs in pip mode, and this skip comes out with `--no-pip` coverage still passing.

.github/instructions/tests.instructions.md
```diff
@@ -7,6 +7,7 @@ applyTo: "tests/**"
 - A test using `uses_db` needs `@pytest.mark.db_roots('local', reason='...')`; `tests/conftest.py` raises a `UsageError` at collection without it. The reason names a PXT ticket, not a bare TODO. Delete the exclusion when the ticket lands.
 - Any test for `pxt.Error` or a subclass uses `pxt_raises()`, not `pytest.raises()`. Both always take `match=` to verify error text.
 - Assert on user-visible behavior through the public API, not `col.stored`, `ColumnRef`, or `TableVersion` internals. Use `Table.get_metadata()`, `t.describe()`, or queries.
+- Reach for an internal API only to test an internal component directly, or to set up a fixture no public API can build.
```
sergey-mkhitaryan: "Avoid using the internal API as much as possible. Only use it to test behaviors that are very difficult or impossible to reproduce using only the public API."

.github/workflows/nightly.yml
sergey-mkhitaryan: Please file a jira to adopt MCP>=2, and leave a `TODO(PXT-xyz): ...` here.

docs/release/platform/embedding-indexes.ipynb
sergey-mkhitaryan: Since this is a notebook, I'm going to leave it to your judgement @pierrebrunelle  (I mean Pierre the human, not whatever wrote the message above). We don't want to make it look like this is a limitation inherent to Pixeltable.

scripts/prepare-nb-tests.sh
sergey-mkhitaryan: Please leave a `# TODO(PXT-1427): ...` here.

pyproject.toml
```diff
@@ -192,7 +192,7 @@ dev = [
     # "whisperx>=3.8.5 ; python_version < '3.14'",
     "ollama>=0.6",
     "llama-cpp-python>=0.3.18 ; sys_platform == 'linux'",
-    "mcp>=1.27.2",
+    "mcp>=1.27.2,<2",
```
sergey-mkhitaryan, line 196: Please leave the same `TODO(PXT-...)` as in nightly.yml

AGENTS.md
```diff
@@ -111,6 +111,11 @@ make formatcheck  # ruff format --check
 
 ### Testing
 
+Exercise behavior through public SDK, CLI, or HTTP APIs and assert on public results, metadata, or errors:
+use `Table.get_metadata()`, `t.describe()`, or queries rather than `col.stored`, `ColumnRef`, or
+`TableVersion` internals. Reach for an internal API only to test an internal component directly, or to set up
```
sergey-mkhitaryan: Same as `tests.instructions.md`: replace the "Reach..." sentence with my comment.

sergey-mkhitaryan (approved):

.github/workflows/nightly.yml
pierrebrunelle: Filed [PXT-1452](https://pixeltable.atlassian.net/browse/PXT-1452) for the 2.x migration, with the two `streamablehttp_client` call sites, the example server rename, and the missing version ceiling in `require_package()` in scope. `TODO(PXT-1452)` added here in 1744c95a7.

pyproject.toml
pierrebrunelle: Same `TODO(PXT-1452)` added above the `mcp` line in 1744c95a7. `uv lock --check` still passes.

scripts/prepare-nb-tests.sh
pierrebrunelle: Done in 1744c95a7: `# TODO(PXT-1427): ...`.

.github/instructions/tests.instructions.md
pierrebrunelle: Replaced with your wording in 1744c95a7.

AGENTS.md
pierrebrunelle: Replaced with your wording in 1744c95a7, wrapped at 120.

PR #1600, opened 2026-09-08
[PXT-183][PXT-1362] Alter Computed Column support with unchanged column type

sergey-mkhitaryan:
The column type must not change, but no restrictions on what the new expression could be outside of normal computed column limitations.

`update_all()` does not recompute the altered columns deliberately. `pxt schema update` prints a notice referring the user to `pxt recompute`. Imperative API, on the other hand, defaults to `recompute=True, cascade=True`.

Table column initialization in `_init_schema()` now happens in a topological order. It is no longer true that, within one table, a computed column can only depend on columns with a lower id.

Metadata change and data recompute (including cascade) all happen in a single table version, so one `revert()` undoes all of that together.

Rearranged how the schema validation happens during a model-driven schema change. First, the schema changes are validated by and applied to each individual table involved. That validation was in Catalog and now moved to TableVersion. After that, each table and each mutable view performs a holistic schema validation once again to check that all dependencies on other tables are satisfied and that there are no dependency cycles. Only after that a potentially expensive new column and index population happens, and the transaction finally commits.


tests/pixeltable_cli/test_schema.py
```diff
@@ -241,6 +243,22 @@ def test_evolution(self, cli: PxtRunner, apps: Callable[[str], str], db_root: Da
...
+        self, cli: PxtRunner, db_root: DatabaseRoot, project_dir: pathlib.Path
+    ) -> None:
+        """Changing a computed column's expression reports that the stored values need recomputing."""
+        p = db_root.make_catalog_path
+        schema_file = project_dir / 'app_schema.py'
+        schema_file.write_text(SCHEMA_SRC)
+        altered_file = project_dir / 'app_schema_altered.py'
+        altered_file.write_text(SCHEMA_SRC.replace('pxtf.string.upper(title)', 'pxtf.string.lower(title)'))
+
+        target = p('altered')
```
mkornacker, line 258: Also check schema diff before the update and after (to verify that its empty), and also make sure --json works correctly.

pixeltable/catalog/model/diff.py
```diff
@@ -207,6 +208,59 @@ def _add_index_change(idx: IndexDefinition) -> SchemaChangeOp:
...
+    col_md: ColumnMetadata,
+) -> SchemaChangeOp | None:
+    """The op that applies a computed column's new value expression, or None if this change cannot be applied."""
+    new_value = spec.get('value')
+    # a computed column becoming a data column, or vice versa, is a different kind of change
+    if new_value is None or not col_md['is_computed']:
+        return None
+    new_value_expr = exprs.Expr.from_object(new_value)
+
+    # Adding new dependencies to the column computation is currently not supported.
```
aaron-siegel: Why not?

This creates an unfortunate scenario where I can alter a column from A to B, but then can't go back from B to A, even if it's a schema no-op.

pixeltable/catalog/model/diff.py
```diff
@@ -207,6 +207,50 @@ def _add_index_change(idx: IndexDefinition, action: str = 'will be added') -> Sc
...
+    return SchemaChangeOp(
+        target='column',
+        name=col_name,
+        op='alter',
+        severity='additive',
+        model={'value': model_props.value},
+        existing={'value': existing_props.value},
+        description=(
+            f'the value expression of computed column {col_name!r} will be updated; '
+            f'existing values will NOT be recomputed'
```
aaron-siegel: I'm not sure that "not" needs to be capitalized

pixeltable/catalog/model/diff.py
```diff
@@ -207,6 +207,50 @@ def _add_index_change(idx: IndexDefinition, action: str = 'will be added') -> Sc
     )
 
 
+def _alter_value_change(
```
aaron-siegel: Why not combine this and the succeeding into a single method that always returns a `SchemaChangeOp`? I think that'd be semantically clearer, rather than have the fallback logic in the caller.

pixeltable/catalog/model/diff.py
```diff
@@ -522,12 +566,23 @@ def by(target: str, op: str | None = None, names: tuple[str, ...] | None = None)
...
                 detail.append(f'    {c.name!r} {prop}: model={model_val!r}, existing={c.existing[prop]!r}')
 
+    supported_alters: list[SchemaChangeOp] = [c for c in altered_cols if c.severity != 'unsupported']
+    if len(supported_alters) > 0:
+        detail.append('  the following computed columns have a new value expression, and will be UPDATED:')
+        for c in supported_alters:
+            detail.append(f'    {c.name!r}: {c.existing["value"]} -> {c.model["value"]}')
+        # If any computed column's expression changed, include the recompute notice
+        if any(c.details.previous_value is not None for c in supported_alters):
+            detail.append('  the current values of these columns are NOT recomputed automatically.')
```
aaron-siegel: Use consistent language w/ the above

(e.g., "existing values will not be recomputed; use `pxt recompute` to do so")

pixeltable/catalog/local_table.py
```diff
@@ -755,6 +763,51 @@ def do_alter_column() -> None:
...
+            if not col.is_computed:
+                raise excs.RequestError(
+                    excs.ErrorCode.UNSUPPORTED_OPERATION, f'Column {col.name!r} is not a computed column'
+                )
+
+            new_value_expr = exprs.Expr.from_object(spec)
+            if new_value_expr is None:
+                raise excs.RequestError(
+                    excs.ErrorCode.INVALID_EXPRESSION,
+                    f'Column {col.name!r}: the new value needs to be a Pixeltable expression, but it is a {type(spec)}',
```
aaron-siegel: usual convention is to enclose types in backticks

pixeltable/catalog/local_table.py
```diff
@@ -755,6 +763,51 @@ def do_alter_column() -> None:
...
+                )
+            new_value_expr = new_value_expr.copy()
+            new_value_expr.bind_rel_paths()
+
+            self._verify_computed_col_value(col.name, new_value_expr)
+
+            tv = self._tbl_version.get()
+            if recompute:
+                assert tv.is_data_versioned, 'TODO: implement recompute for operational tables [PXT-1101]'
```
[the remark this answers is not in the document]
aaron-siegel: Agreed - unless we're going to fix this before the next release, put a useful error message here (with the JIRA ticket number in a TODO code comment)

pixeltable/catalog/local_table.py
```diff
@@ -755,6 +763,51 @@ def do_alter_column() -> None:
...
+        @retry_loop(for_write=True, write_tvps=[self._tbl_version_path], lock_mutable_tree=True)
+        def do_alter_computed_column() -> UpdateStatus:
+            self._check_mutable('alter columns of')
+
+            col = self._tbl_version_path.get_column(fold_identifier(col_name))
+            if col is None:
+                raise excs.NotFoundError(excs.ErrorCode.COLUMN_NOT_FOUND, f'Unknown column: {col_name}')
+            if col.get_tbl().id != self._tbl_version_path.tbl_id:
+                raise excs.RequestError(
+                    excs.ErrorCode.UNSUPPORTED_OPERATION, f'Cannot alter base table column {col.name!r}'
```
aaron-siegel, line 783: What if it's an iterator output column, or a base query column?

pixeltable/catalog/table.py
```diff
@@ -459,6 +459,43 @@ def alter_column(self, column: str | ColumnRef, *, type_: TypeForm) -> None:
...
+
+        Raises:
+            Error: If the column does not exist, is not a computed column, belongs to a base table, or if the new
+            expression has a different column type or creates a dependency cycle.
+
+        Examples:
+            Change a computed column to use a different scaling factor:
+
+            >>> tbl = pxt.create_table('my_table', {'n': pxt.Int})
+            ... tbl.add_computed_column(scaled=tbl.n * 2)
```
aaron-siegel: Separate commands should be preceded by `>>>` (use `...` only to continue a single multi-line statement)

pixeltable_cli/client/commands/schema.py
```diff
@@ -490,6 +490,28 @@ def _update_output(plan: SchemaPlan, *, as_json: bool) -> None:
...
+    altered_computed_cols: list[str] = [
+        f'  {tbl.path}.{op.name}'
+        for tbl in plan.tables
+        for op in tbl.ops
+        if op.op == 'alter' and op.details.previous_value is not None and op.status == 'applied'
+    ]
+    if len(altered_computed_cols) == 0:
+        return
+    print()
+    print('the value expressions of these columns changed, but their stored values were NOT recomputed:')
```
aaron-siegel: s/NOT/not/

pixeltable_cli/client/commands/schema.py
```diff
@@ -490,6 +490,28 @@ def _update_output(plan: SchemaPlan, *, as_json: bool) -> None:
...
+    for line in altered_computed_cols:
+        print(line)
+    print('run `pxt recompute` to recompute them using their new expressions.')
```
aaron-siegel: I'd soften this to "run `pxt recompute` if you wish to recompute them" (make it clear it's not necessary and is at the user's discretion)

tests/test_alter_column.py
```diff
@@ -117,3 +117,203 @@ def test_alter_column_errors(self, db_root: DatabaseRoot) -> None:
...
+        t.add_computed_column(c=t.n * 2)
+        v = pxt.create_view(db_root.make_catalog_path('test_view'), t, additional_columns={'d': t.c + 1})
+        validate_update_status(t.insert(n=1), 2)
+
+        # a view's computed column can be altered
+        v.alter_computed_column(d=v.c + 100)
+        assert v.select(v.d).collect()['d'] == [102]
+
+        # a base table's column cannot be altered through the view
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='Cannot alter base table column'):
```
aaron-siegel, line 303: put these in test_computed_column_errors?

tests/test_alter_column.py
```diff
@@ -117,3 +117,203 @@ def test_alter_column_errors(self, db_root: DatabaseRoot) -> None:
...
+            t.alter_computed_column(c=(t.n + t.m) / 2)
+        # a reference that makes the column depend on itself, directly or indirectly
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='circular dependency'):
+            t.alter_computed_column(c=t.c + 1)
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='circular dependency'):
+            t.alter_computed_column(c=t.n + t.m + t.d)
+        # a reference to a cell metadata property
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match="'errortype' property"):
+            t.alter_computed_column(c=t.n + (t.c.errortype != None).astype(pxt.Int))
+
```
aaron-siegel: Also add tests that try to alter a base query column and iterator output column

pixeltable/catalog/model/diff.py
sergey-mkhitaryan: I assume this is an old comment that you wrote but didn't submit until now.

tests/test_alter_column.py
sergey-mkhitaryan: I'm assuming you mean `test_alter_column_errors` above. Moving there.

pixeltable/catalog/local_table.py
sergey-mkhitaryan: everything works, added a missing test

pixeltable/catalog/catalog.py
```diff
@@ -1871,111 +1879,56 @@ def update_fn() -> None:
...
                     f'add indexes {[spec.idx_name for spec in added_idxs]}, '
                     f'drop indexes {change_set["dropped_idxs"]}'
                 )
-                tv.apply_schema_change(expected_schema_version, added_cols, dropped_cols, added_idxs, dropped_idx_ids)
+                tv.apply_schema_change_md(
+                    expected_schema_version, added_cols, altered_cols, dropped_cols, added_idxs, dropped_idx_ids
+                )
+                validation_tvs[tv.id] = tv
+                for view_tv in self._mutable_view_tvs(tv):
+                    validation_tvs[view_tv.id] = view_tv
```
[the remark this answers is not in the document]
sergey-mkhitaryan: known problem: https://pixeltable.atlassian.net/browse/PXT-1343

pixeltable/catalog/model/diff.py
```diff
@@ -207,6 +207,46 @@ def _add_index_change(idx: IndexDefinition, action: str = 'will be added') -> Sc
...
+    spec: ColumnSpec,
+    model_props: _ColumnProperties,
+    existing_props: _ColumnProperties,
+    col_md: ColumnMetadata,
+    altered: list[str],
+) -> SchemaChangeOp:
+    """The op for a column whose properties differ; a new value expression for a computed column is the only one
+    update_all() can apply."""
+    # a computed column becoming a data column, or vice versa, changes more than the value expression
+    is_new_value_expr = altered == ['value'] and spec.get('value') is not None and col_md['is_computed']
```
aaron-siegel, line 221: `spec.get('value') is not None` is clearer as `'value' in spec` IMO (they're equivalent since it can't be present as `None`)

I believe this implies `col_md['is_computed']` so the latter could be made into an assertion.

pixeltable/catalog/model/diff.py
```diff
@@ -207,6 +207,46 @@ def _add_index_change(idx: IndexDefinition, action: str = 'will be added') -> Sc
...
+            model={prop: getattr(model_props, prop) for prop in altered},
+            existing={prop: getattr(existing_props, prop) for prop in altered},
+            description=f'column {col_name!r} has altered properties: {", ".join(altered)}',
+        )
+
+    return SchemaChangeOp(
+        target='column',
+        name=col_name,
+        op='alter',
+        severity='additive',
```
aaron-siegel, line 237: it's destructive, no?

pixeltable/catalog/model/diff.py
```diff
@@ -207,6 +207,46 @@ def _add_index_change(idx: IndexDefinition, action: str = 'will be added') -> Sc
...
+            target='column',
+            name=col_name,
+            op='alter',
+            severity='unsupported',
+            model={prop: getattr(model_props, prop) for prop in altered},
+            existing={prop: getattr(existing_props, prop) for prop in altered},
+            description=f'column {col_name!r} has altered properties: {", ".join(altered)}',
+        )
+
+    return SchemaChangeOp(
```
aaron-siegel, line 233: What if there's a new_value_expr, whose result type differs from the existing one?

pixeltable/catalog/model/resolution.py
```diff
@@ -223,15 +225,23 @@ def prepare_model_updates(
     tvp: catalog.TableVersionPath,
     display_name: str,
     new_columns: dict[str, tuple[ColumnSpec, Literal['base_query', 'model_body']]],
+    altered_columns: dict[str, tuple[ColumnSpec, Literal['base_query', 'model_body']]],
     new_idxs: list[IndexDefinition],
-) -> tuple[list[catalog.Column], list[catalog.IndexSpec]]:
+) -> tuple[list[catalog.Column], list[catalog.IndexSpec], dict[str, exprs.Expr]]:
```
aaron-siegel, line 230: This return type is getting hard to parse; it'd benefit from being made into a NamedTuple.

tests/test_alter_column.py
aaron-siegel: Yes that's what I meant

aaron-siegel (changes requested):
I want to take a closer look at the catalog logic tomorrow; a lot has changed since the last time I looked at this PR. In the meantime, I'm posting these comments to keep things moving.

tests/test_alter_column.py
```diff
@@ -117,3 +118,226 @@ def test_alter_column_errors(self, db_root: DatabaseRoot) -> None:
...
+            t.alter_computed_column(c4=t.c5 + 1)
+
+        # a reference that makes the column depend on itself, directly or indirectly
+        t.add_computed_column(c6=t.c4 * 2)
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='circular dependency'):
+            t.alter_computed_column(c4=t.c4 + 1)
+        with pxt_raises(pxt.ErrorCode.UNSUPPORTED_OPERATION, match='circular dependency'):
+            t.alter_computed_column(c4=t.c3 + t.c6)
+
+        # a reference to a cell metadata property
```
aaron-siegel, line 143: i'd think this should be allowable. do we have a ticket for it?

pixeltable/catalog/model/diff.py
sergey-mkhitaryan: I guess I don't know how we define additive and destructive, but being that it's a reversible metadata-only change (no recompute), not sure how it's different from, say, adding an empty column. But if you think it's destructive, I can change that.

pixeltable/catalog/model/diff.py
sergey-mkhitaryan: The former: done. The latter: this can't be an assertion, it would trip if a data column is changed to a computed column.

pixeltable/catalog/model/diff.py
sergey-mkhitaryan: Then `altered` will be `['value', 'type']` and we won't reach this line. The outcome will be the diff is unsupported. `test_update_all_altered_column_unsupported` tests that (see `NewTypeTable`)

aaron-siegel (approved):

tests/test_alter_column.py
sergey-mkhitaryan: This is an existing limitation that alter computed column is inheriting from `add_computed_column()`. I haven't found a jira so I filed https://pixeltable.atlassian.net/browse/PXT-1420.

tests/test_table_model.py
```diff
@@ -1997,34 +2001,112 @@ class ExampleTableV3(TableModelV3, name='test_table'):
...
+
+        AlterModel.create_all(p(''))
+        pxt.get_table(p('alter_base')).insert([{'id': 1, 'extra': 7}])
+
+        # the altered column belongs to the table that drops 'extra'
+        SameTableModel = pxt.model_base()
+
+        class SameTableBase(SameTableModel, name='alter_base'):
+            id: pxt.Int
+            doubled = AlterBase.extra * 3
```
mkornacker, line 2030: I think this should be rejected because it's referencing a column from a different model.

tests/test_table_model.py
```diff
@@ -1997,34 +2001,112 @@ class ExampleTableV3(TableModelV3, name='test_table'):
...
+
+        # the altered column belongs to a view of the table that drops 'extra'
+        CrossTableModel = pxt.model_base()
+
+        class CrossTableBase(CrossTableModel, name='alter_base'):
+            id: pxt.Int
+            doubled = id * 2
+
+        class CrossTableView(CrossTableModel, name='alter_view', base=CrossTableBase):
+            vc = AlterBase.extra + 1
```
mkornacker, line 2049: Same here. This should be restricted to references within CrossTableBase.

tests/test_table_model.py
```diff
@@ -2802,3 +2934,270 @@ class M(TableModel, name='m'):
...
+
+        TableModel.create_all(root)
+        t = pxt.get_table(p('test_table'))
+        v = pxt.get_table(p('test_view'))
+        t.insert([{'id': 1, 'extra': 0}, {'id': 2, 'extra': 0}])
+        assert t.select(t.doubled).order_by(t.id).collect()['doubled'] == [2, 4]
+
+        AlteredModel = pxt.model_base()
+
+        class AlteredTable(AlteredModel, name='test_table'):
```
mkornacker, line 2962: Add comment: what changed.

tests/test_table_model.py
```diff
@@ -2802,3 +2934,270 @@ class M(TableModel, name='m'):
...
+        assert t.select(t.doubled).order_by(t.id).collect()['doubled'] == [100, 200]
+        v = pxt.get_table(p('test_view'))
+        assert v.select(v.vc1).order_by(v.id).collect()['vc1'] == [101, 201]
+
+        # one change set that adds a column an altered one depends on (bonus), and one that depends on an altered
+        # one (s). s is declared before bonus but reads it through unstored u, so Pixeltable needs to be smart about
+        # the order in which these columns are created and populated.
+        WidenedModel = pxt.model_base()
+
+        class WidenedTable(WidenedModel, name='test_table'):
```
mkornacker, line 2995: I'm losing track of what's changed/added.

tests/test_table_model.py
```diff
@@ -2802,3 +2934,270 @@ class M(TableModel, name='m'):
...
+        assert sorted(diff.ops[0].model.keys()) == ['type', 'value']
+
+        # referencing a cell metadata property
+        CellMdModel = pxt.model_base()
+
+        class CellMdTable(CellMdModel, name='test_table'):
+            id: pxt.Int
+            other: pxt.Int
+            derived = id * 2
+            derived2 = (id * 3 + (ExampleTable.derived.errortype != None).astype(pxt.Int)).astype(  # type: ignore[attr-defined]
```
mkornacker, line 3089: Does it need the astype()? That's pretty awful and worth a jira ticket.

tests/test_table_model.py
```diff
@@ -2802,3 +2934,270 @@ class M(TableModel, name='m'):
...
+        TableModel.create_all(root)
+        pxt.get_table(p('test_table')).insert([{'x': 1}])
+
+        # Two alter computed columns happen at once, such that neither of them alone would create a dependency cycle,
+        # but together they do. The cycle closes through a third column, c2, that doesn't change.
+        CycleModel = pxt.model_base()
+
+        class CycleTable(CycleModel, name='test_table'):
+            x: pxt.Int
+            c1 = ExampleTable.c3 + 1
```
mkornacker, line 3163: Here too: this shouldn't be allowed.

pixeltable_cli/client/commands/schema.py
```diff
@@ -503,6 +503,28 @@ def _update_output(plan: SchemaPlan, *, as_json: bool) -> None:
     for tbl in plan.tables:
         outcome = 'refused' if tbl.status == 'refused' else _RESOLUTIONS[tbl.resolution].applied
         print(f'{outcome:9s} {tbl.path}')
+    _print_recompute_notice(plan)
+
+
+def _print_recompute_notice(plan: SchemaPlan) -> None:
+    """Report the columns whose stored values the applied changes left stale.
```
mkornacker: Remove this docstring, it doesn't say anything useful.

pixeltable_cli/client/commands/schema.py
```diff
@@ -503,6 +503,28 @@ def _update_output(plan: SchemaPlan, *, as_json: bool) -> None:
...
+
+    This is the only signal a user gets: a value-expression change updates the definition without recomputing, and
```
mkornacker: This narrates a product decision, which has no place in a docstring.

pixeltable_cli/client/commands/schema.py
```diff
@@ -503,6 +503,28 @@ def _update_output(plan: SchemaPlan, *, as_json: bool) -> None:
...
+        for op in tbl.ops
+        if op.op == 'alter' and op.details.previous_value is not None and op.status == 'applied'
+    ]
+    if len(altered_computed_cols) == 0:
+        return
+    print()
+    print('the value expressions of these columns changed, but their stored values were not recomputed:')
+    for line in altered_computed_cols:
+        print(line)
+    print('run `pxt recompute` if you wish to recompute them.')
```
mkornacker, line 532: Print the exact commandline, so people/agents can copy/paste.

mkornacker (commented):
I'll comment on table_version tomorrow.

pixeltable/catalog/table_version.py
```diff
@@ -104,10 +147,9 @@ class TableVersion:
     # target for data operation propagation (only set for non-snapshots, and only records non-snapshot views)
     mutable_views: frozenset[TableVersionHandle]
 
-    # User and system columns that are live in this schema version, ordered by column id (asc). The order matters for
-    # iterator_columns().
+    # User and system columns that are live in this schema version, ordered by column id (asc)
```
mkornacker, line 150: Word more strongly: guaranteed to be ordered by column id (asc). Same below.

pixeltable/catalog/table_version.py
```diff
@@ -679,12 +707,8 @@ def add_columns(self, cols: list[Column], print_stats: bool, on_error: Literal['
...
         return status
 
-    def _add_columns(
-        self, cols: Iterable[Column], print_stats: bool, on_error: Literal['abort', 'ignore']
-    ) -> UpdateStatus:
-        """Add and populate columns within the current transaction"""
-        from pixeltable.plan import Planner
-
+    def _record_new_columns(self, cols: Iterable[Column]) -> None:
+        """Record columns in this version's metadata and create their store columns, leaving them unpopulated."""
```
mkornacker, line 711: This seems to diverge from _record_new_index() in that it creates the store columns. 

It's a bit hard to follow where column creation happens, both here and for index cols. Maybe something like a _materialize_store_columns() is a better abstraction instead of _populate..(), the former also doing the creation.

pixeltable/catalog/table_version.py
```diff
@@ -874,34 +940,43 @@ def apply_schema_change(
...
+            self._apply_column_changes_md(added_cols, altered_cols)
         for col, idx_name, idx in added_idxs:
             assert isinstance(col, Column)
-            status += self._add_index(col, idx_name, idx)
+            self._add_index_md(col, idx_name, idx)
 
+        get_runtime().catalog.record_column_dependencies(self)
+        self.path.clear_cached_md()
+
+    def complete_schema_change(self) -> UpdateStatus:
```
mkornacker, line 952: materialize_...() might be a more concrete name.

pixeltable/catalog/table_version.py
```diff
@@ -911,39 +986,70 @@ def _add_columns_in_version(
             col.id = self.next_col_id()
 
         # the ids exist now, so a placeholder can be resolved to the column it names
```
mkornacker, line 988: "to the actual column" ("column it names" is an llm/Claude tell).

pixeltable/catalog/table_version.py
```diff
@@ -911,39 +986,70 @@ def _add_columns_in_version(
...
 
-        status = self._add_columns(all_cols, print_stats=print_stats, on_error=on_error)
-        # create the indices and their md records only once the columns they index exist
-        for col, (idx, val_col, undo_col) in index_cols.items():
-            self._create_index(col, val_col, undo_col, idx_name=None, idx=idx)
+        self._record_new_columns(all_cols)
+        for col, idx, val_col, undo_col in default_idxs:
+            self._record_new_index(col, val_col, undo_col, idx_name=None, idx=idx)
+
+        # altered_cols are applied after cols are recorded, so that a new value expression can reference a column that
```
mkornacker, line 1011: Rephrase for clarity; 2 "so that".

pixeltable/catalog/table_version.py
```diff
@@ -911,39 +986,70 @@ def _add_columns_in_version(
...
+        for col, new_value_expr in resolved_alters:
+            self._alter_value_expr_in_version(col, new_value_expr)
+        if len(resolved_alters) > 0:
+            # the altered value expressions are part of the CVMD that create_add_column_plan() reads
+            self.path.clear_cached_md()
+
+    def _populate_new_columns(self, print_stats: bool, on_error: Literal['abort', 'ignore']) -> UpdateStatus:
+        """Compute the values of the columns added by the current schema version and build their store indices.
+
+        This does not recompute the columns whose expressions changed in the current schema version."""
```
mkornacker, line 1026: Rephrase for clarity: "does not recompute altered computed columns."

pixeltable/catalog/table_version.py
```diff
@@ -911,39 +986,70 @@ def _add_columns_in_version(
...
+        ]
+        if len(new_cols) == 0 and len(new_idx_ids) == 0:
+            return UpdateStatus()
+
+        status = self._populate_columns(self._population_order(new_cols), print_stats=print_stats, on_error=on_error)
+        for idx_id in new_idx_ids:
+            self.store_tbl.create_index(idx_id)
         return status
 
+    def _population_order(self, cols: list[Column]) -> list[Column]:
```
mkornacker, line 1043: topological_order() would be more intuitive.

mkornacker (approved):

tests/test_table_model.py
sergey-mkhitaryan: True. Filed https://pixeltable.atlassian.net/browse/PXT-1468 to fix it and updated the new tests not to take advantage of this bug.

tests/test_table_model.py
sergey-mkhitaryan: Rewrote it not to require astypes at all. The outer one was unnecessary to begin with, I assume that is what you had in mind for the jira.

pixeltable/catalog/table_version.py
sergey-mkhitaryan: `_record_new_columns()` now only updates md, does not create store columns. `_populate_new_columns()` is renamed to `_materialize_new_columns()` and store column creation now happens there. I like it better this way.

sergey-mkhitaryan:
grok review

pixeltable/catalog/local_table.py
```diff
@@ -755,6 +763,51 @@ def do_alter_column() -> None:
...
+                    f'Column {col.name!r}: the new value needs to be a Pixeltable expression, '
+                    f'but it is a `{type(spec)}`',
+                )
+            new_value_expr = new_value_expr.copy()
+            new_value_expr.bind_rel_paths()
+            self._verify_computed_col_value(col.name, new_value_expr)
+
+            tv = self._tbl_version.get()
+            if recompute:
+                assert tv.is_data_versioned, 'TODO: implement recompute for operational tables [PXT-1101]'
```
pierrebrunelle, line 803: Raise `RequestError(UNSUPPORTED_OPERATION)` here. `recompute=True` is the default and is a user-reachable path on an operational table (including `TableProxy` → `_alter_computed_column`). `assert` is for internals; `revert()` already uses `RequestError` for this mode. Keep PXT-1101 in a comment, not in the exception text. `test_alter_computed_column` returns early before this call.

pixeltable/catalog/model/diff.py
```diff
@@ -207,6 +207,46 @@ def _add_index_change(idx: IndexDefinition, action: str = 'will be added') -> Sc
...
+    spec: ColumnSpec,
+    model_props: _ColumnProperties,
+    existing_props: _ColumnProperties,
+    col_md: ColumnMetadata,
+    altered: list[str],
+) -> SchemaChangeOp:
+    """The op for a column whose properties differ. At this point, the only supported change is a new value expression
+    for a computed column."""
+    # a computed column becoming a data column, or vice versa, changes more than the value expression
+    is_new_value_expr = altered == ['value'] and 'value' in spec and col_md['is_computed']
```
pierrebrunelle, line 221: Value equality for this alter path still uses `display_str()` in `_ColumnProperties.from_spec`. That is only `fn_name(args)`. A `@pxt.query` body lives in `QueryTemplateFunction._as_dict()['df']` and is what the catalog stores. Changing only the body of `titles_after()` while keeping `matches = titles_after(cutoff)` compares equal and `update_all()` leaves the old template. Compare a bound serialization (or fingerprint that `df`) after `ColumnRefByName` substitution.

pixeltable/catalog/table_version.py
```diff
@@ -1003,6 +1109,150 @@ def alter_column(self, col: Column, type_: ts.ColumnType) -> None:
...
+        old_value_expr = col.value_expr
+        self._alter_value_expr_in_version(col, new_value_expr)
+        get_runtime().catalog.record_column_dependencies(self)
+        self.path.clear_cached_md()
+        _logger.info(
+            f'Altered value expression of column {col.name!r} in table {self.name} from {old_value_expr} to '
+            f'{new_value_expr}, new version: {self.version}'
+        )
+
+        if not recompute:
```
pierrebrunelle, line 1249: Call `set_version_update_status(UpdateStatus())` before `_write_md()` when `recompute` is false. `bump_version()` does not clear `_version_md.update_status`, so an insert of N rows followed by `alter_computed_column(..., recompute=False)` makes `get_versions()[0]` report those N inserts on the schema version. The model path already writes an empty status via `materialize_schema_change()`.

pixeltable_cli/client/commands/schema.py
```diff
@@ -513,6 +513,25 @@ def _update_output(plan: SchemaPlan, *, as_json: bool) -> None:
...
         outcome = 'refused' if tbl.status == 'refused' else _RESOLUTIONS[tbl.resolution].applied
         print(f'{outcome:9s} {tbl.path}')
+    _print_recompute_notice(plan)
+
+
+def _print_recompute_notice(plan: SchemaPlan) -> None:
+    altered_cols_by_tbl: dict[str, list[str]] = {}
+    for tbl in plan.tables:
+        for op in tbl.ops:
+            if op.op == 'alter' and op.details.previous_value is not None and op.status == 'applied':
```
pierrebrunelle: This selects every applied alter with `previous_value` set. That includes `stored=False` columns (no stored values; the new expr is already used on read) and operational tables, where `pxt recompute` still hits the PXT-1101 assert. Carry `stored` / `is_data_versioned` on the op details and print the command only for stored columns on data-versioned tables.

pixeltable/catalog/table.py
```diff
@@ -459,6 +459,43 @@ def alter_column(self, column: str | ColumnRef, *, type_: TypeForm) -> None:
             ... tbl.alter_column('col', type_=pxt.String | None)
         """
 
+    @abc.abstractmethod
+    def alter_computed_column(
```
pierrebrunelle, line 463: Add `pixeltable.Table.alter_computed_column` to `docs/public_api.opml` next to `add_computed_column` and `recompute_columns`. The Table section enumerates methods; without this entry the generated SDK reference omits the new API.

pixeltable/catalog/table_version.py
```diff
@@ -1003,6 +1109,150 @@ def alter_column(self, col: Column, type_: ts.ColumnType) -> None:
...
+            raise excs.RequestError(
+                excs.ErrorCode.UNSUPPORTED_OPERATION,
+                f'Column {col.name!r}: the new value expression has type `{new_value_expr.col_type}`, but the '
+                f'column has type `{col.col_type}`. Changing the type of a computed column is not supported.',
+            )
+
+    def validate_column_dependencies(self) -> None:
+        """Verify that this version's value expressions and predicate reference columns that exist.
+
+        This is a complete check of the table's metadata and its dependencies on other tables. It guarantees that
```
pierrebrunelle: Delete the “complete check” / “guarantees that the table can be loaded” sentences. This function walks `ColumnRef`s in value exprs, the predicate, and iterator args, then same-table cycles. Query-template refs are not in that graph (PXT-1343). A comment must not claim a guarantee the next lines do not implement.

pixeltable/catalog/table_version.py
```diff
@@ -1003,6 +1109,150 @@ def alter_column(self, col: Column, type_: ts.ColumnType) -> None:
...
+            col = self.lookup_column(ref.col_md.qcolid)
+            return col is not None and col.get_tbl().effective_version == ref.col_md.col_effective_version
+
+        def unresolved_ref_error(dependent: str, ref: exprs.ColumnRef) -> excs.RequestError:
+            qid = ref.col_md.qcolid
+            tbl: TableVersion | None = self
+            while tbl is not None and tbl.id != qid.tbl_id:
+                tbl = tbl.base.get() if tbl.base is not None else None
+            if tbl is None:
+                # the reference names a table that is neither this one nor one of its ancestors
```
pierrebrunelle: Delete this comment. “the reference names a table” is the banned `<x> names <y>` form, and the `if tbl is None` branch already states the condition.

pierrebrunelle (commented):
Needs changes before merge.

Adds `Table.alter_computed_column()` (type must stay the same; new deps allowed; default `recompute=True, cascade=True`) and model-driven alters via `update_all()` / `pxt schema update` (metadata only; CLI prints `pxt recompute …`).

Catalog path is sound: per-table `apply_schema_change_md()`, then holistic `validate_column_dependencies()` (including mutable views, predicates, iterator args), then `materialize_schema_change()`. One version + one `revert()` for alter+recompute. `_init_schema()` now loads columns in topological order.

Aaron and Marcel approved older SHAs. This head addresses Marcel’s naming/comments/tests. It does not fix the user-reachable operational `assert`, query-UDF detection via `display_str()`, or `get_versions()` inheriting the prior version’s row counts.

Raise `RequestError(UNSUPPORTED_OPERATION)` when `recompute=True` on an operational table. Default `recompute=True` is a user-reachable path (including `TableProxy`); `assert` is for internals.

`display_str()` is only `fn_name(args)`. A `@pxt.query` body lives in `QueryTemplateFunction._as_dict()['df']`. Changing only the body while keeping `matches = titles_after(cutoff)` compares equal and `update_all()` leaves the old template. Compare a bound serialization after `ColumnRefByName` substitution.

Call `set_version_update_status(UpdateStatus())` before `_write_md()` when `recompute` is false, so `get_versions()[0]` does not inherit prior insert counts.

Print the CLI recompute notice only for stored columns on data-versioned tables.

Add `pixeltable.Table.alter_computed_column` to `docs/public_api.opml` next to `add_computed_column` and `recompute_columns`.

pixeltable/catalog/local_table.py
pierrebrunelle, line 803: Raise `RequestError(UNSUPPORTED_OPERATION)` here. `recompute=True` is the default and is reachable on an operational table (including via `TableProxy`). `revert()` already uses `RequestError` for this mode. Keep PXT-1101 in a code comment, not in the exception text. `test_alter_computed_column` returns early before this call.

pixeltable/catalog/model/diff.py
```diff
@@ -207,6 +207,46 @@ def _add_index_change(idx: IndexDefinition, action: str = 'will be added') -> Sc
...
+            existing={prop: getattr(existing_props, prop) for prop in altered},
+            description=f'column {col_name!r} has altered properties: {", ".join(altered)}',
+        )
+
+    return SchemaChangeOp(
+        target='column',
+        name=col_name,
+        op='alter',
+        severity='additive',
+        model={'value': model_props.value},
```
pierrebrunelle, line 238: `model_props.value` / `existing_props.value` are `display_str()` output (`fn_name(args)`). A `@pxt.query` body lives in `QueryTemplateFunction._as_dict()['df']` and is what the catalog stores. Changing only the body of `titles_after()` while keeping `matches = titles_after(cutoff)` compares equal, so `update_all()` leaves the old template in metadata. Compare a bound serialization (or fingerprint that `df`) after `ColumnRefByName` substitution. `test_update_all_query_udf_over_model` only changes call arguments.

pixeltable/catalog/table_version.py
```diff
@@ -1003,6 +1109,150 @@ def alter_column(self, col: Column, type_: ts.ColumnType) -> None:
...
+        self._alter_value_expr_in_version(col, new_value_expr)
+        get_runtime().catalog.record_column_dependencies(self)
+        self.path.clear_cached_md()
+        _logger.info(
+            f'Altered value expression of column {col.name!r} in table {self.name} from {old_value_expr} to '
+            f'{new_value_expr}, new version: {self.version}'
+        )
+
+        if not recompute:
+            self._write_md(new_version=True, new_schema_version=True)
```
pierrebrunelle, line 1250: Call `set_version_update_status(UpdateStatus())` before `_write_md()` when `recompute` is false. `bump_version()` does not clear `_version_md.update_status`, so an insert of N rows followed by `alter_computed_column(..., recompute=False)` makes `get_versions()[0]` report those N inserts on the schema version. The model path already writes an empty status via `materialize_schema_change()`.

pierrebrunelle (commented):
## Needs changes before merge

Adds `Table.alter_computed_column()` (same type; new deps allowed; default `recompute=True, cascade=True`) and model-driven alters via `update_all()` / `pxt schema update` (metadata only; CLI prints `pxt recompute …`).

Catalog path looks right: per-table `apply_schema_change_md()`, holistic `validate_column_dependencies()` (mutable views, predicates, iterator args), then `materialize_schema_change()`. Alter + recompute share one version, so one `revert()` undoes both. `_init_schema()` loads columns in topological order.

Aaron approved `91fd2ed` and Marcel approved `f8b14cc`. This head (`1b62b807`) picks up Marcel’s naming/comments/tests. It still leaves a user-reachable operational `assert`, query-UDF body changes invisible to the model diff, and `get_versions()` inheriting the prior version’s row counts after a metadata-only alter.

### Blockers
1. Default `recompute=True` on an operational table raises `AssertionError` (inline).
2. Model diff can miss a `@pxt.query` body-only change (inline).

### Also fix
- Clear `update_status` on metadata-only alter (inline).
- Restrict CLI recompute notice to stored cols on data-versioned tables (inline).
- Add `pixeltable.Table.alter_computed_column` to `docs/public_api.opml` next to `add_computed_column` / `recompute_columns` (file not in this diff).

### Tests to add
- Operational table, default `alter_computed_column(...)`: expect `UNSUPPORTED_OPERATION`, not `AssertionError`.
- After insert then `alter_computed_column(..., recompute=False)`, `get_versions()[0]` insert/update counts are 0.
- Model: same `titles_after(cutoff)` call, different query body → additive update; recompute evaluates the new body.
- Model cycle: `c1 = c2 + 1` and `c2 = c1 + 1` in one `update_all()`.
- CLI: alter only `stored=False`; notice must not list it.

### Non-issues on this head
Drop+alter / dependency subset, iterator args, snapshot refs, cellmd props, populate-before-alter, view predicate drops, iterator-output rejection, library silence vs CLI notice, DFS RecursionError on thousand-deep chains, PXT-1343 query-UDF drop dependents, additive-vs-destructive (Aaron asked; later approval stands), Marcel’s post-approval renames/tests/exact `pxt recompute` command.

pixeltable_cli/client/commands/schema.py
pierrebrunelle: This selects every applied alter with `previous_value` set. That includes `stored=False` columns (no stored values; the new expr is already used on read) and operational tables, where `pxt recompute` still hits the PXT-1101 path. Carry `stored` / versioning mode on the op details and print the command only for stored columns on data-versioned tables.

pixeltable/catalog/table.py
pierrebrunelle, line 463: Add `pixeltable.Table.alter_computed_column` to `docs/public_api.opml` next to `add_computed_column` and `recompute_columns`. The Table section enumerates methods; without this entry the generated SDK reference omits the new API. `alter_column` is also missing there; do not use that as a reason to skip this one.

pixeltable/catalog/table_version.py
pierrebrunelle: Delete the “complete check” / “guarantees that the table can be loaded” sentences. This walks `ColumnRef`s in value exprs, the predicate, and iterator args, then same-table cycles. Query-template refs are not in that graph (PXT-1343). A comment must not claim a guarantee the next lines do not implement.

pixeltable/catalog/table_version.py
pierrebrunelle: Delete this comment. “the reference names a table” is the banned `<x> names <y>` form, and the `if tbl is None` branch already states the condition.

pixeltable/catalog/local_table.py
sergey-mkhitaryan: nope

pixeltable/catalog/model/diff.py
sergey-mkhitaryan: This is real (though not exactly a new bug that this pr introduces). Filed https://pixeltable.atlassian.net/browse/PXT-1470 to address it separately in the interest of time.

pixeltable/catalog/table_version.py
sergey-mkhitaryan: fixed

PR #1602, opened 2026-09-08
Drop db_name from the hosted database protocol

mkornacker:


tests/pixeltable_cli/test_schema.py
```diff
@@ -347,6 +347,9 @@ def test_query_udf_column(self, cli: PxtRunner, apps: Callable[[str], str], db_r
         ]
         assert_in_agreement(cli, apps('retrieval.py'), target)
 
+    @pytest.mark.db_roots(
+        'local', reason='TODO: run against a hosted database, once a pod carries an API key for home bucket access'
```
aaron-siegel: link to a JIRA ticket (i.e., PXT-nnnn instead of TODO)

same comment for the other TODOs below

tests/pixeltable_cli/test_service.py
```diff
@@ -13,7 +13,12 @@
 from ..utils import DatabaseRoot, get_audio_files, get_documents, get_video_files, skip_test_if_not_installed
 from .conftest import BackgroundPxt, PxtRunner
 
-pytestmark = pytest.mark.db_roots('local', reason='a local service serves the in-process catalog')
+# proxy is excluded because get_manager() hands any non-local path to ServiceManagerProxy, so a
+# 'pxt://local:db' target reaches the cloud management API, which knows no org named 'local'.
+pytestmark = [
```
aaron-siegel: Claude loves these file-level `pytestmark` declarations, but I find it a lot clearer to put decorators on the `class TestService`

aaron-siegel (approved):

pixeltable_cli/client/hosted.py
```diff
@@ -104,7 +104,7 @@ def _print_workers(workers: list[dict[str, Any]]) -> None:
 
 
 def print_db(db: dict[str, Any]) -> None:
-    name = db.get('db_name') or db.get('db', '')
+    name = db.get('db_slug') or db.get('db_name') or db.get('db', '')
```
christopherpestano: Shouldn't we remove db_name here?

tests/pixeltable_cli/test_schema.py
mkornacker: rephrased.

tests/pixeltable_cli/test_service.py
mkornacker: Done.

tests/pixeltable_cli/conftest.py
```diff
@@ -177,6 +184,83 @@ def _copy_app_corpus(session_project: pathlib.Path) -> pathlib.Path:
...
+    """
+    if db_root.id != 'cloud':
+        return
+    if session_project in _hosted_image_built_from:
+        return
+    _copy_app_corpus(session_project)
+    # the model as well as the package: spacy resolves 'en_core_web_sm' by import, not by download
+    write_requirements(session_project, pixeltable_wheel, 'spacy', _SPACY_MODEL)
+    base_uri = os.environ['PXTTEST_CLOUD_DB_URI']
+    (session_project / 'pixeltable.toml').write_text(
```
christopherpestano: This seems hacky, if there a case where service update relies on db update to reach a consistent state then shouldn't it call it automatically? Or does this require app update to be proper?

christopherpestano (approved):

pixeltable_cli/client/hosted.py
mkornacker: I guess let's just stick with the immutable name for now.

tests/pixeltable_cli/conftest.py
aaron-siegel: Oh, good call Chris, I missed this!

I recently took out a similar fixture that Claude put in (which took me some effort to strip out).

@mkornacker Why do we need this hosted_image fixture?

aaron-siegel (changes requested):

tests/pixeltable_cli/test_schema.py
```diff
@@ -347,6 +347,7 @@ def test_query_udf_column(self, cli: PxtRunner, apps: Callable[[str], str], db_r
         ]
         assert_in_agreement(cli, apps('retrieval.py'), target)
 
+    @pytest.mark.db_roots('local', reason='TODO: re-enable for hosted once cloud PR 199 is in')
```
aaron-siegel, line 350: No need for this exclusion if the cloud PR is in flight; the 'cloud' variants don't run in CI

aaron-siegel (approved):

PR #1604, opened 2026-09-08
[PXT-1370 + PXT-1349 + PXT-1323] Hosted media defaults to the database's home bucket

christopherpestano:
Media on a hosted database used to land in the pod's media directory, which is an emptyDir: inserted files, uploads through a service endpoint, and computed media were all lost on the next pod restart, and query results referred to them through a /media URL that only the proxy tunnel can serve.

Env now recognizes the hosted database it runs in (Env.hosted_db, from the PXTCLOUD_ORG and PXTCLOUD_DB variables the cloud sets on its pods) and defaults input_media_dest and output_media_dest to that database's home bucket, pxtfs://org:db/home. An explicit destination on a column still wins, and so does a user default set in config or through a pxt secret, so the bucket only applies when nothing else was asked for. Local Pixeltable and the local proxy daemon are unchanged. _prefetch_remote_parts() reads the same identity instead of the environment directly.

Services now return presigned HTTP URLs for media stored in an object store, so a response never carries a raw pxtfs or s3 URI. The URI resolution shared with the presigned_url() UDF moved into ObjectOps.presigned_url(). Because hosted media no longer comes from the pod, the service's own /media route is no longer on the path for cloud services, which is what made the URLs in PXT-1349 unusable.

The boto3 client caches for S3-compatible stores were held per thread through the API client registry, so a 16-thread media prefetch built 16 sessions and minted credentials 16 times, enough to exceed a hosted daemon's memory limit once media reads went through the bucket. They are now one process-wide cache on Env (Env.object_store_clients()), and the pxtfs entry build is serialized so concurrent readers of one bucket share one session.

Tests cover the default's precedence and the end-to-end flow through a fake bucket store (inserted file, in-memory image, computed output, read back, deleted on drop). The cloud docs page now describes the home bucket as the default and destination as the override.


pixeltable/utils/pxt_store.py
```diff
@@ -48,10 +49,7 @@ class _PxtStoreCacheEntry:
     no_space_warned: bool = False  # tracks whether warning has been issued for no space left in pixeltable store
 
 
-# pxt_store clients are thread-local (via Runtime._clients), consistent with r2/s3/b2/tigris.
-@env.register_client('pxt_store', credential_param=None)
-def _() -> S3CompatClientDict:
-    return S3CompatClientDict(profile=None, clients={})
+_pxt_store_entries_lock = threading.Lock()
```
mkornacker, line 54: Missing comment: what is this protecting.

pixeltable/env.py
```diff
@@ -140,6 +140,8 @@ class Env:
     _default_input_media_dest: str | None
     _default_output_media_dest: str | None
     _pxt_api_key: str | None
+    _object_store_clients: dict[str, S3CompatClientDict]  # by storage target ('r2', 'r2_resource', 's3', ...)
```
mkornacker: Do we have a defined set of storage targets? If so, create a Literal or something. Having everything be :str makes it really hard to understand the code or do any meaningful type checks.

pixeltable/env.py
```diff
@@ -227,6 +231,15 @@ def http_address(self) -> str:
     def is_proxy_daemon(self) -> bool:
         return os.environ.get('PIXELTABLE_PROXY_DAEMON') == '1'
 
+    @property
+    def hosted_db(self) -> tuple[str, str] | None:
+        """(org, db) of the hosted database this process serves; the cloud sets both variables on its pods."""
```
mkornacker: "this process serves": please remove vacuous filler commentary (what process if not this one?) left by Claude (here and everywhere else in this PR).

pixeltable/service/proxy_dispatch.py
```diff
@@ -154,13 +154,13 @@ def _prefetch_remote_parts(request: ProxyRequest) -> None:
             raise excs.RequestError(
                 excs.ErrorCode.INVALID_ARGUMENT, f'Invalid uploaded media object key: {remote_key!r}'
             )
-    org = os.environ.get('PXTCLOUD_ORG')
-    db = os.environ.get('PXTCLOUD_DB')
-    if not (org and db):
+    hosted_db = Env.get().hosted_db
```
mkornacker: Make it something like Env.hosted_db(*, required: bool = False) and then raise there if it's not set but required. That way, we have a uniform message and the hosted_db caller doesn't need to know the internals of hosted_db.

mkornacker (changes requested):
Missing: augmented test cases in test_fastapi.py that show that you're getting fetchable media urls back.

pixeltable/service/proxy_dispatch.py
christopherpestano: Sure replaced with Env.get().hosted_db(required=True)

pixeltable/env.py
christopherpestano: Yes we do, created an enum with the given storage targets and replaced references to instead be to the enum

pixeltable/env.py
christopherpestano: Removed this and some other excessive claude comments

pixeltable/utils/pxt_store.py
christopherpestano: Added comments as to where the read then write happens

tests/serving/test_fastapi.py
```diff
@@ -1445,10 +1447,116 @@ def thumb_by_id(img_id: int) -> pxt.Query:
         assert resp.headers['content-type'].startswith('image/')
         assert len(resp.content) > 0
 
+    @pytest.mark.db_roots('local', reason='patches the in-process object store registry')
+    def test_object_store_media_urls(self, uses_db: None, monkeypatch: pytest.MonkeyPatch) -> None:
+        """Media stored in an object store comes back from insert and query routes as a presigned HTTP URL that
```
mkornacker: We already have lots of tests here for inserts and queries, and a lot of these already deal with media data, why does this need to be special-cased? We'd get better coverage if you augment the existing test cases to check for correct urls.

mkornacker (approved):

tests/serving/test_fastapi.py
christopherpestano: I agree but most of the tests are already marked as not working against cloud per the top market `@pytest.mark.db_roots('local', 'proxy', reason='Numerous failures; re-run once other known issues are fixed')`, this scoped coverage at least assures us that this specific feature is working as expected even if the broader suite is not reliable.

PR #1607, opened 2026-09-09
Fail a request immediately (no retries) if the pod becomes unresponsive waiting for a response

aaron-siegel:


tests/test_proxy_daemon.py
```diff
@@ -349,3 +350,122 @@ def failing_handler(request: proxy_protocol.ProxyRequest) -> None:
...
+
+    def close(self) -> None:
+        self.closed = True
+        self.sock.close()
+        self._peer.close()
+
+    def close_peer(self) -> None:
+        """Make the connection read as closed by the server."""
+        self._peer.close()
+
```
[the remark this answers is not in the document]
aaron-siegel: Per Claude:

  Leaving this as is — the concern doesn't reproduce, and both halves of the suggested fix are
  already covered elsewhere.

  `close()` is idempotent today: `socket.close()` is a no-op on an already-closed socket, so there's
  nothing to add. And `socket.socket` already has the defensive `__del__` — that's the machinery that
  closes these. A second `__del__` on `_ScriptedConn` would duplicate it, and hand-written `__del__`
  methods bring their own problems (object resurrection, swallowed exceptions).

  Measured rather than assumed:

      $ python -W error::ResourceWarning -m pytest tests/test_proxy_daemon.py -k TestTunnelRetries
      5 passed

  No `ResourceWarning: unclosed socket`. I also ran 20 iterations of the same allocation pattern
  (three socketpairs per iteration, one left parked in the pool, all dropped at function exit) and
  counted `/dev/fd` before and after: delta 0. The conns are reachable only from the test function's
  locals, so refcounting reclaims them as soon as it returns — nothing accumulates across the suite.

  If we do want explicit teardown later, the right shape is a fixture that closes the pool, not
  `__del__`.

pixeltable/service/proxy_client.py
```diff
@@ -150,6 +151,22 @@ def connect(self) -> None:
         pass  # socket already set in __init__
 
 
+def _is_server_closed(conn: http.client.HTTPConnection) -> bool:
+    """Return `True` if the remote end of an idle connection has already closed it.
+
+    A socket that is readable before any request has been written to it carries the server's FIN (or bytes
```
mkornacker: What is "or bytes no request asked for"?

pixeltable/service/proxy_client.py
```diff
@@ -159,11 +176,24 @@ def __init__(self, connect: Callable[[], http.client.HTTPConnection], max_size:
         self._lock = threading.Lock()
         self._idle: list[http.client.HTTPConnection] = []
 
+    def _take_idle(self) -> http.client.HTTPConnection | None:
```
mkornacker: Does this have any other callers beside borrow()? If not, inline. The docstring isn't particularly comprehensible.

pixeltable/service/proxy_client.py
```diff
@@ -267,8 +301,18 @@ def _request(self, method: str, path: str, body: bytes | None = None, content_ty
         def _attempt() -> bytes:
             with self._pool.borrow() as conn:
                 conn.request(method, path, body=body, headers=headers)
-                response = conn.getresponse()
-                content = response.read()
+                try:
+                    response = conn.getresponse()
+                    content = response.read()
+                except _TUNNEL_TRANSIENT_EXC as exc:
```
mkornacker, line 307: Aren't we basically circumventing the retry mechanism here for exactly the same class of exceptions as in line 295?? Wouldn't it be simpler not to have the retry mechanism in there in the first place?

aaron-siegel: We still need the retry mechanism when sending the initial request - that's the whole point

pixeltable/service/proxy_client.py
```diff
@@ -150,6 +151,22 @@ def connect(self) -> None:
...
+    """Return `True` if the remote end of an idle connection has already closed it.
+
+    A socket that is readable before any request has been written to it carries the server's FIN (or bytes
+    no request asked for); either way it cannot carry the next request.
+    """
+    sock = conn.sock
+    if sock is None:
+        return True
+    try:
+        rlist, _, _ = select([sock], [], [], 0)
```
christopherpestano: Select has a limit of 1024 fds, whereas the os is happy (and often does) to spawn more than 1024 fds, need to use selectors.DefaultSelector instead otherwise clients with more than 1024 fds will have their connections closed wrongly.

pixeltable/service/proxy_client.py
aaron-siegel: Possibly I could refactor the logic a bit to make it clearer.

pixeltable/service/proxy_client.py
```diff
@@ -267,8 +301,18 @@ def _request(self, method: str, path: str, body: bytes | None = None, content_ty
         def _attempt() -> bytes:
             with self._pool.borrow() as conn:
                 conn.request(method, path, body=body, headers=headers)
```
christopherpestano, line 300: According to Claude we actually have a race condition here, in between us checking that the socket is closed and us then making a new request it's actually possible for the server to close the connection for idleness, which would result the next few lines failing and returning an error "as if" the server had crashed. Realistically in such cases we should retry at least once otherwise it'd be difficult to catch and explain where these odd .1% failures to customers.

aaron-siegel: Not sure I agree with this one - all the error message says is that the server is unresponsive, not that it necessarily crashed. And the cost of retrying a large query, even just once, is high (we'll kill twice as many pod workers that way).

pixeltable/service/proxy_client.py
aaron-siegel: Done

pixeltable/service/proxy_client.py
aaron-siegel: Done

pixeltable/service/proxy_client.py
aaron-siegel: Done

pixeltable/service/proxy_client.py
aaron-siegel: I added some code comments to clarify.

mkornacker (approved):

PR #1609, opened 2026-09-09
Upload `bytes` and `ndarray`s out of band during insertion [PXT-1314, PXT-1318, PXT-1319, PXT-1320]

aaron-siegel:
- Upload `bytes` and `ndarray`s out of band via the same pathway (home bucket intermediary) used for media objects. (Data of less than 512 bytes is still inlined.)
- Reduce the footprint of certain tests in `'cloud'` configuration
- Set `memory_mb = 2048` on the cloud DB configuration in `pyproject.toml`

### DIFF
--- a/pixeltable/service/db.py
+++ b/pixeltable/service/db.py
@@ -353,7 +353,9 @@ def _build_image(db_path: catalog.Path, archive_key: str, fingerprint: ProjectFi
     if current.last_build_state == 'FAILED':
         raise excs.ExternalServiceError(
             excs.ErrorCode.PROVIDER_ERROR,
-            f'The image build for {db_path.uri_str} failed: {current.last_build_error or "no reason was reported"}',
+            # covers the whole runtime update, not just the image build: the reported error also comes from
+            # the rollout onto the built image
+            f'The runtime update for {db_path.uri_str} failed: {current.last_build_error or "no reason was reported"}',
             provider='pixeltable_cloud',
         )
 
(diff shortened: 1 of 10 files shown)


pixeltable/service/db.py
```diff
@@ -353,7 +353,7 @@ def _build_image(db_path: catalog.Path, archive_key: str, fingerprint: ProjectFi
     if current.last_build_state == 'FAILED':
         raise excs.ExternalServiceError(
             excs.ErrorCode.PROVIDER_ERROR,
-            f'The image build for {db_path.uri_str} failed: {current.last_build_error or "no reason was reported"}',
+            f'The runtime update for {db_path.uri_str} failed: {current.last_build_error or "no reason was reported"}',
```
mkornacker: 'image build' is more accurate, 'runtime update' doesn't mean anything.

tests/test_proxy_daemon.py
```diff
@@ -97,6 +103,80 @@ def test_inline_sink_wire_format(self, tmp_path: pathlib.Path) -> None:
         assert sink.binary_parts[2] == b'abc'
         assert proxy_protocol.collect_remote_keys(wire) == []
 
+    def test_scalar_out_of_band_threshold(self, tmp_path: pathlib.Path) -> None:
```
mkornacker: Shouldn't we have end-to-end coverage of the new logic via test_inlined_objects?

Or is that disabled?

I'd rather not add tests for internal interfaces.

mkornacker (approved):

pixeltable/service/db.py
aaron-siegel: Fixed, I'm not even sure how that crept in. Possibly a bungled git merge (sorry)

tests/test_proxy_daemon.py
aaron-siegel: I agree with you on this one. I removed the test

PR #1610, opened 2026-09-10
Image/archive split

mkornacker:
- `pxt db` reconciles a hosted database from one DatabaseSpec: diff, update, restart,
  build-image, delete. Replaces the separate create/build-image operations.
- A project is fingerprinted in two parts: the image (manifests, python, system deps)
  and the archive (all files). Only what moved is rebuilt or re-uploaded.
- `pxt service restart` and `pxt db restart` cycle pods onto what they already run.
- Error carries an error_code and must be raised as a subclass; UserError is exported.


pixeltable/utils/project.py
```diff
@@ -152,12 +146,90 @@ def _archive_files(project_root: Path, config: DatabaseConfig | None) -> list[Pa
...
+    """An archive, and the content hash of every file written into it."""
+
+    path: Path
+
+    # path relative to the project root -> sha256 of the bytes written into the archive
+    files: dict[str, str]
+
+
+@dataclasses.dataclass
+class PackagedContext:
```
aaron-siegel, line 176: identical to PackagedArchive?

pixeltable/utils/project.py
```diff
@@ -192,36 +264,313 @@ def create_project_archive(
...
+                # uv takes a table, or a list of them where it picks one by marker or extra
+                entries = source if isinstance(source, list) else [source]
+                if not any(isinstance(e, dict) and ('path' in e or 'workspace' in e) for e in entries):
+                    continue
+                raise excs.RequestError(
+                    excs.ErrorCode.INVALID_CONFIGURATION,
+                    f'dependency {name!r} is declared as a local source in {f.name}, which cannot be '
+                    'installed in a hosted image; publish it to an index and depend on the published version',
+                )
+            for requirement in _declared_dependencies(parsed):
```
aaron-siegel, line 503: If I understand things right, this throws an exception if there is any unresolvable dependency in pyproject.toml.

This is very likely to result in Pixeltable rejecting perfectly valid configurations. For example, if the dev dependencies (or some optional depedency group) contains a reference to a local package, but the database's uv-options are given with --no-dev.

I think we are being FAR too clever in trying to anticipate the user's intentions viz. their pyproject.toml and requirements.txt. My preference would be to just trust the user's configuration and provide good error reporting if the image build fails, rather than try to anticipate so many different kinds of failures up front. We can always add fail-fast logic in later, when we have more time to be thoughtful about it.

aaron-siegel (changes requested):

pixeltable/utils/project.py
mkornacker: Consolidated.

pixeltable/utils/project.py
mkornacker: Changed.

tests/pixeltable_cli/requirements.txt
```diff
@@ -0,0 +1,5 @@
+# pixeltable is here for its dependencies; the image reinstalls the platform's build of it
```
aaron-siegel: This is no longer true, correct?

tests/conftest.py
```diff
@@ -352,8 +352,9 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
         else:
             params = ('local', 'proxy', 'cloud')  # Default is all three targets
 
-        if os.environ.get('PXTTEST_CLOUD_DB_URI') is None:
-            # If the cloud db URI is not set, skip generating any cloud tests. We short-circuit them here rather
+        # each package's cloud fixture reads its own variable; either one turns the axis on
+        if os.environ.get('PXTTEST_CLOUD_DB_URI') is None and os.environ.get('PXTTEST_CLI_DB_URI') is None:
```
aaron-siegel: I'd prefer to have two different axes (`'cloud'` and `'cloud-cli'`), if we're going to do this.

pixeltable/service/db.py
```diff
@@ -51,124 +43,77 @@
...
     if db_path.org is None or db_path.db is None:
         return None
-    state = _get_db_state(db_path)
-    return None if state is None else state.fingerprint
+    report = _get_db_report(db_path)
+    return None if report is None or report.current is None else report.current.resources.fingerprint
+
+
+def create_db_update_ops(target: DatabaseResources, current: DatabaseResources | None) -> list[DbChangeOp]:
+    """The operations needed to reconcile current with target."""
```
christopherpestano, line 66: It feels a bit odd to me to put all this effort towards moving towards an ops based approach to driving db update but still having the plan be constructed on the client side, rather shouldn't it be the lambda that creates this?

pixeltable/service/db.py
```diff
@@ -177,77 +122,121 @@ def db_update(db_uri: str, *, allow_destructive: bool = False) -> DbPlan:
...
+        update={
+            'cpu': running.cpu,
+            'memory_mb': running.memory_mb,
+            'disk_gb': running.disk_gb,
+            'workers': running.workers,
+        }
+    )
+    settled, stored = _update_db(db_path, config, target, force_image_build=True)
+    image_op = DbChangeOp.build_image()
+    image_op.status = 'applied' if settled.last_build_outcome == 'SUCCEEDED' else 'skipped'
```
christopherpestano, line 228: Hardcoded string rather than enum

pixeltable/service/management_protocol.py
```diff
@@ -86,20 +83,51 @@ class GetDbRequest(BaseModel):
     db: str
 
 
+class GetDbResponse(BaseModel):
+    report: DatabaseReport
+
+    # one entry per pod serving the database; DatabaseResources.workers is how many should run
+    worker_status: list[dict[str, Any]] = Field(default_factory=list)
```
christopherpestano, line 90: More informative type hint than Any possible?

pixeltable/service/management_protocol.py
```diff
@@ -86,20 +83,51 @@ class GetDbRequest(BaseModel):
...
     db: str
-    default_bucket: str | None = None
-    workers: int | None = None
-    cpu: float | None = None
-    memory_mb: int | None = None
-    disk_gb: int | None = None
+    target: DatabaseResources | None = None
+
+    # compute the plan without recording the spec, acting on it, or handing out an upload url
+    dry_run: bool = False
```
christopherpestano, line 107: What does dry_run do that diff doesn't already do? I've noticed claude likes to shoehorn it everywhere - I was never a fan.

pixeltable/service/proxy_daemon.py
```diff
@@ -389,6 +389,25 @@ def _cleanup(*_: Any) -> None:
     uvicorn.Server(uvicorn.Config(app, log_level=log_level, log_config=None)).run(sockets=[sock])
 
 
+# get_archive returns 404 both for a database with no project and for one whose release did not
+# resolve just now; retrying tells the two apart
+_ARCHIVE_FETCH_DELAYS = (0.0, 1.0, 2.0, 4.0)
+
+
+def _unpack_project(db_uri: str, project_dir: Path) -> bool:
```
christopherpestano, line 397: Handrolled exponential backoff, we already have tenacity as a req dep just use the retry decorator from there instead of having this custom logic.

pixeltable/service/service_md.py
```diff
@@ -0,0 +1,99 @@
...
+    # where the instance answers; derived on both sides, so a stored one is never read back
+    endpoint: str = ''
+
+    # the app file's module path, relative to the project root
+    app_module: str
+
+    spec: ServiceSpec
+
+    # whether the instance emits OpenTelemetry traces
+    otel: bool = False
```
christopherpestano, line 44: Why are we exposing otel right now? This shouldn't be enabled currently.

pixeltable/service/service_md.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class ServiceResources(pydantic.BaseModel):
+    """What one of a service instance's pods gets, and how many of them serve it."""
+
+    cpu: float = 0.5
+    memory_mb: int = 512
+    disk_gb: int = 10
+    workers_min: int = 1
+
```
christopherpestano, line 23: DatabaseResources takes None, whereas this hardcodes the defaults (i.e. cpu, mem, etc) we should probably go with none since the defaults rightfully belong in the control plane. Also I'm not too certain as to why we need a separate pod resources md class for service and database, the common part should be factored out at the very least.

aaron-siegel (approved):

pixeltable/serving/service.py
```diff
@@ -342,7 +360,7 @@ def _service_diff(
         resolution = 'blocked'
     elif running is None:
         resolution = 'create'
```
christopherpestano, line 362: Can we change this resolution to an enum

christopherpestano (approved):
My comments are non blocking, the overall shape is fine, lets fix these issues later

tests/pixeltable_cli/requirements.txt
mkornacker: Done.

tests/conftest.py
mkornacker: Done.

pixeltable/service/management_protocol.py
mkornacker: Diff is implemented with dry_run=True.

pixeltable/service/service_md.py
mkornacker: No harm in capturing it in the metadata now.

PR #1614, opened 2026-09-10
[PXT-1385] Honor on_error='ignore' for object-store put failures

pierrebrunelle:
## Summary
- ObjectStoreSaveNode asserted `new_file_url is not None` before handling put exceptions, so `on_error='ignore'` crashed with `AssertionError` instead of recording a cell error (PXT-1385).
- Move the assert to the success path at the use site, matching `CachePrefetchNode`.
- Add a regression test that injects put failure for abort vs ignore.

## Test plan
- [x] `pytest tests/test_destination.py::TestDestination::test_dest_put_failure_on_error_ignore`
- [ ] CI slimtest / destination tests

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pixeltable/exec/object_store_save_node.py
+++ b/pixeltable/exec/object_store_save_node.py
@@ -185,7 +185,6 @@ class ObjectStoreSaveNode(ExecNode):
             new_file_url, exc = f.result()
             if exc is not None and not ignore_errors:
                 raise exc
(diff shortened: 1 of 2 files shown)


pierrebrunelle:
Closing: head was based on a diverged local main. Replacing with a branch cut from current upstream/main.

PR #1622, opened 2026-09-11
PXT-1373: potential livelock in _finalize_pending_ops()

mkornacker:


pixeltable/catalog/catalog.py
```diff
@@ -961,11 +963,50 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
                     )
 
                     tbl_version = tbl_md.current_version if tbl_md.is_snapshot else None
-                    tv = (
-                        self._get_tbl_version(TableVersionKey(tbl_id, tbl_version), check_pending_ops=False)
-                        if op.needs_tv
-                        else None
-                    )
+                    tv: TableVersion | None = None
+                    try:
```
sergey-mkhitaryan, line 967: 2 issues with this:
1. bug: `raise` on line 1007 aborts the transaction, including the tbl status update to ROLLBACK from line 999
2. The nested `try` and `is_final` combo smells: this code is already a pyramid of doom, let's please try not to make it worse.

Here's the sketch from Claude on how the same fix can be done without a nested `try`. It needs a separate transaction, but I think that it's still better. The observation is that the only thing that the nested `try` "knows" that the outer one doesn't is whether or not the op was attempted. But the same knowledge can be communicated with just a bool:

```
while True:
    op_attempted = False          # reset per iteration
    try:
        with self.begin_xact(...) as conn, self._allow_tbl_md_read():
            ...
            tv = self._get_tbl_version(...) if op.needs_tv else None
            op_attempted = True   # past this point the op may be partially executed
            ...
    except AssertionError: raise                       # unchanged
    except excs.PixeltableWarning: raise               # unchanged
    except PendingTableOpsError as e: ...              # unchanged
    except sql_exc.DBAPIError as e: ...                # unchanged
    except Exception as e:
        if excs.is_table_not_found_error(e):
            return None

        if not op_attempted:
            # the op never ran, so it has nothing to undo
            if is_rollback or tbl_md is None or not tbl_md.pending_stmt.can_abort():
                raise
            exc = e
            with self.begin_xact(for_write=True, write_tbl_ids=[tbl_id], ...):
                <set tbl_state = ROLLBACK>
                if self._set_pending_op_status(tbl_id, op, OpStatus.ABORTED, is_final_op=op.op_sn == 0):
                    raise            # statement resolved: reaches the original caller
            continue

        if not is_rollback and tbl_md is not None and tbl_md.pending_stmt.can_abort():
            ...                                        # unchanged
```

(Note, the sketch does not fix the transaction abort on raise)

tests/test_catalog.py
```diff
@@ -153,6 +154,29 @@ def test_finalize_pending_ops_retriable_error(self, uses_db: None, fault_injecti
         fault.assert_count(1)
         _ = t.select(t.b).collect()
 
+    @pytest.mark.db_roots('local', reason='fault-injection/concurrency test against the in-process catalog internals')
```
sergey-mkhitaryan, line 157: unrelated to this pr obviously, but I'm wondering if `db_roots` can detect the `fault_injection` fixture and become local-only automatically.

pixeltable/catalog/catalog.py
```diff
@@ -884,11 +884,13 @@ def _finalize_pending_ops(self, tbl_id: UUID) -> Exception | None:
         - this process starts with the first pending op, because it could have been partially executed
         - when done, deletes all table ops and resets tbl_state to LIVE
 
-        If an exception occurred during finalization, that exception is returned. PendingOpsErrors encountered during
-        finalization are dealt with recursively.
+        If an exception occurred during finalization, that exception is returned, except where it left the
```
sergey-mkhitaryan, line 887: Why can't we always return an error (or raise an error, but not both)? Why special case some failure conditions?

sergey-mkhitaryan (changes requested):

PR #1623, opened 2026-09-11
PXT-1393: add_update_route and batch_update reject a computed uuid7

mkornacker:


pixeltable_cli/server/routes.py
```diff
@@ -635,19 +633,21 @@ def dashboard_table_export(req: Request) -> RawResponse:
 _COUNT_POOL_WORKERS = 16
 
 
-def _coerce_pk(s: str) -> Any:
-    """Numeric-looking PK strings become int or float; everything else stays a string.
+_PK_PARSERS: dict[str, Callable[[str], Any]] = {'Int': int, 'Float': float, 'String': str, 'UUID': uuid.UUID}
```
aaron-siegel, line 636: Are these all the allowable primary key types? I could envision DATE being useful as a pk, for example.

aaron-siegel (approved):

pixeltable_cli/server/routes.py
mkornacker: Sure, we can add more later if anyone asks.

PR #1624, opened 2026-09-11
[PXT-1407] Presign media that hosted compute and query routes make on the fly

christopherpestano:
Compute routes on a hosted database returned /media URLs of the service itself, which a client cannot reach. The same happened to query routes whose select list transforms media inline. Insert routes were fine because the save node moves their media into the home bucket and the service presigns the resulting URI. Compute and inline query media are never stored, so they stayed as local files on the pod.

A service in a hosted pod now stages such a file in the database's home bucket under the uploads prefix, which the bucket expires after a day, and returns a presigned URL with the same one hour expiry the stored media already gets. Local services are unchanged and keep serving the file from /media.

The media test app gains a query route that scales the poster in the query, and the service test now fetches the persisted poster, the query thumb, and the compute frame thumb the same way: presigned on the cloud axis, /media locally. On the cloud axis that run passed against a chrisdev database built from this branch.

### DIFF
--- a/pixeltable/serving/_fastapi.py
+++ b/pixeltable/serving/_fastapi.py
@@ -45,6 +45,7 @@ from pixeltable.config import Config
 from pixeltable.env import Env
 from pixeltable.exec.globals import INLINED_OBJECT_MD_KEY
 from pixeltable.runtime import close_threadpool_runtimes
+from pixeltable.service.proxy_protocol import PxtStorePartSink
 from pixeltable.serving import SqlExport
 from pixeltable.serving.globals import SqlExporter
 from pixeltable.utils import image as image_utils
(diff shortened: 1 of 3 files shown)


pixeltable/serving/_fastapi.py
```diff
@@ -372,12 +381,39 @@ def route_id(self) -> int:
     def route_type(self) -> Literal['insert', 'update', 'delete', 'compute', 'query']:
         return self.route.spec.route_type
 
-    def __call__(self, request: Request, **kwargs: Any) -> Any:
-        sample_url = str(request.url_for(_MEDIA_ROUTE_NAME, path='_'))
-        media_url_base = sample_url[:-1]
+    def _url_for_media(self, request: Request) -> tuple[Callable[[str], str], PxtStorePartSink | None]:
```
aaron-siegel, line 384: I find this pattern a little confusing. Rather than return a (Callable, PxtStorePartSink) pair, it'd be more logical to me to introduce a `MediaManager` abstraction (or some such - we don't have to call it that) with methods `url_for_media()` and `flush()`.

aaron-siegel (approved):

pixeltable/serving/_fastapi.py
christopherpestano: Encapsulated in class _ResponseMedia as given abstraction for handling the uploading the response media.

PR #1627, opened 2026-09-14
[PXT-1391] Per-database media destinations and OTLP endpoint

christopherpestano:
Adds db_input_media_dest, db_output_media_dest, db_exporter_otlp_endpoint and db_exporter_otlp_protocol to [[pixeltable.database]] entries. A process reads the entry for its own database: the hosted one on its pod, else the local one. A value the entry sets wins over the environment variable and a pxt.init() override; where the entry is silent, the environment and then the shared [pixeltable] / [otel] sections apply as before.

The docs describe the three places a default can be set and their order, and the deployment and observability pages show the database entry alongside the environment variables and config.toml. The OTLP headers stay in the environment or in pxt secret set.

Tests: test_config covers the lookup order, the hosted entry, and rejection of keys the entry does not accept; test_env covers the home bucket fallback; the otel package test covers the endpoint from the entry; the test_service cloud axis checks that inserted media lands where the entry points.


pierrebrunelle:
@codex review

pierrebrunelle:
@claude review

pierrebrunelle:
@claude review

mkornacker (changes requested):
Overarching comment:
Please revise/remove all Claude-generated junk prose, in comments and elsewhere. It's a clear tell of AI slop.

Example: "where media that names no destination goes"

pixeltable/config.py
```diff
@@ -84,6 +91,15 @@ def _check_python_version(cls, v: str | None) -> str | None:
 # the entry in [[pixeltable.database]] that configures the local database
 LOCAL_DATABASE = 'local'
 
+# the settings a [[pixeltable.database]] entry sets for its database, as the section and key of the lookup that
+# reads each one; the entry's field has the key's name
+_DATABASE_SETTINGS = {
```
mkornacker: Why are these singled out? There's a whole bunch of other settings.

pixeltable/config.py
```diff
@@ -84,6 +91,15 @@ def _check_python_version(cls, v: str | None) -> str | None:
 # the entry in [[pixeltable.database]] that configures the local database
 LOCAL_DATABASE = 'local'
 
+# the settings a [[pixeltable.database]] entry sets for its database, as the section and key of the lookup that
```
mkornacker: Completely incomprehensible.

pixeltable/config.py
```diff
@@ -354,6 +370,7 @@ def __init__(
         self.__reported_env_vars = set()
         self.__config_file = Path(self.lookup_env('pixeltable', 'config', str(self.__home / 'config.toml')))
         self.__project_config_file = self.__resolve_project_config_file()
+        self.__database_sources: dict[str, dict[str, Path | None]] = {}
```
mkornacker: Missing in decls.

mkornacker (changes requested):

docs/release/platform/configuration.mdx
```diff
@@ -52,15 +52,39 @@ rate_limit = 600  # Mistral uses a single rate limit for all models
...
 | PXT_PORT | | (int) Port the background process listens on; default is 22089. See [Dashboard](/platform/dashboard) |
 | PIXELTABLE_API_KEY | [pixeltable]<br/>api_key | (string) API key for Pixeltable Cloud. Prefer the exported env name; create the key under API Keys. |
-| PIXELTABLE_INPUT_MEDIA_DEST | [pixeltable]<br/>input_media_dest | (string) Default destination URI for media files that are inserted into tables |
-| PIXELTABLE_OUTPUT_MEDIA_DEST | [pixeltable]<br/>output_media_dest | (string) Default destination URI for media files that are generated by Pixeltable operations |
-| OTEL_EXPORTER_OTLP_ENDPOINT | [otel]<br/>exporter_otlp_endpoint | (string) OTLP endpoint for traces. Requires `pip install 'pixeltable[otel]'`. See [Observability](/platform/observability) |
-| OTEL_EXPORTER_OTLP_HEADERS | [otel]<br/>exporter_otlp_headers | (string) OTLP auth headers, e.g. `Authorization=Basic <token>` |
+| PIXELTABLE_INPUT_MEDIA_DEST | [[pixeltable.database]]<br/>input_media_dest | (string) Default destination URI for media files that are inserted into tables. See [Per-database settings](#per-database-settings) |
```
aaron-siegel: This is misleading; the environment variable would map to the `[pixeltable]` config section, not `[[pixeltable.database]]`

(Are we keeping the global defaults? I think we should.)

docs/release/platform/configuration.mdx
```diff
@@ -52,15 +52,39 @@ rate_limit = 600  # Mistral uses a single rate limit for all models
...
 | PIXELTABLE_TIGRIS_PROFILE | [pixeltable]<br/>tigris_profile | (string) Name of an S3-compatible profile for accessing Tigris. Defaults to the standard AWS credential chain if not set. |
 
+### Per-database settings
+
+Media destinations and the OTLP endpoint and protocol belong to one database, so each `[[pixeltable.database]]` entry sets its own. The entry with no `name` configures the local database; an entry named `pxt://org:db` configures that hosted database, whose pods read the entry from the project files `pxt db update` uploads.
+
+```toml
+# pixeltable.toml
+[[pixeltable.database]]
+input_media_dest = 's3://bucket/input/'
```
aaron-siegel: Suggestion: call these `db_input_media_dest`, etc., to clarify the scope and clearly distinguish them from global settings

We should also make the precedence order clear somewhere (assuming we're keeping the global settings?)

db setting > env var > global config setting

pixeltable/config.py
aaron-siegel: Why do we need this comment?

pixeltable/config.py
```diff
@@ -668,32 +692,53 @@ def get_database_config(self, db_uri: catalog.Path) -> DatabaseConfig | None:
...
+        entry = self.__config_dict.get('pixeltable', {}).get('database')
+        if entry is None:
+            return None
+        own = next((db for db in entry[0] if db.name == name), None)
+        if own is None:
+            return None
+        return own, self.__database_sources[name]
+
+    def __database_bindings(self, section: str) -> dict[str, tuple[str, Path | None]]:
+        """The vars or secrets of this process's database, each with the file that supplied it."""
```
aaron-siegel: I thought secrets were going away from db config.

pixeltable/env.py
```diff
@@ -237,12 +237,15 @@ def is_proxy_daemon(self) -> bool:
         return os.environ.get('PIXELTABLE_PROXY_DAEMON') == '1'
 
     @overload
-    def hosted_db(self, *, required: Literal[True]) -> tuple[str, str]: ...
+    @staticmethod
+    def hosted_db(*, required: Literal[True]) -> tuple[str, str]: ...
 
     @overload
-    def hosted_db(self, *, required: bool = False) -> tuple[str, str] | None: ...
+    @staticmethod
```
aaron-siegel: Standard python convention is to always use `@classmethod` unless there's a specific reason why `@staticmethod` is necessary.

aaron-siegel (changes requested):
I'm not thrilled with the overall architecture, which pushes the resolution complexity into the config module.

My own instinct would be to keep the config module simple - its role is to hybridize configuration files into a key/value store with validation, not to internalize scope resolution logic that is specific to particular settings.

I think it'd be cleaner to move that logic into Env - and the fact that we need to backport Env.hosted_db() into Config supports that view.

Happy to discuss.

docs/release/platform/configuration.mdx
christopherpestano: True, changed it back to [pixeltable], yes we are keeping the globaal defaults

docs/release/platform/configuration.mdx
christopherpestano: Sure that makes differentiating them in the pydantic model a bit easier too, the precedence order has been clarified now (in several places), though the way I'm implementing it is env var > db setting > global config setting since that way overrides like for pxt.init still are taken before db specific settings (which is what I would expect).

pixeltable/config.py
christopherpestano: Completely redid the comments in this pr

christopherpestano: Not really needed removed

pixeltable/config.py
christopherpestano: It's because this is a whitelist of what we "intended" for the user to have available as db specific overrides to the global [pixeltable] section. Look at the new `class DatabaseSetting(enum.StrEnum)` in this file which should be bit clearer in its intent. If wanted we can broaden to all that's fine too we'd just delete the enum and go off the prefix.

pixeltable/config.py
christopherpestano: Added

pixeltable/config.py
christopherpestano: my bad hadn't merged main in yet

pixeltable/env.py
christopherpestano: That's standard python convention? We aren't even using cls here? I guess a model level function would've also been cleaner, regardless changed to @classmethod for uniformity with repo convention.

pixeltable/config.py
```diff
@@ -56,6 +69,17 @@ class DatabaseConfig(pydantic.BaseModel):
     disk_gb: int | None = None
     workers: int | None = None
 
+    @pydantic.model_validator(mode='before')
+    @classmethod
+    def _collect_settings(cls, data: Any) -> Any:
+        # moves the db_<key> keys of a parsed entry into `settings`, so they validate against DatabaseSetting
```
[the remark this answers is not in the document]
christopherpestano: I felt like it clarifies it rather than paraphrases that code snippet was confusing for me initially

docs/release/integrations/cloud-storage.mdx
```diff
@@ -54,16 +54,29 @@ Do not put `PIXELTABLE_API_KEY` in `config.toml` as `api_key`. Use the env name.
 
 There are two ways to configure cloud storage destinations. These apply to **local** Pixeltable and to **bring-your-own** buckets. Hosted Cloud tables skip this section.
```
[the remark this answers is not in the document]
christopherpestano: fixed

tests/pixeltable_cli/test_service.py
```diff
@@ -56,6 +56,14 @@
 
 _REQUEST_TIMEOUT = 30.0
 
+# where the cloud axis's database entry sends inserted media, under its home bucket
+_INPUT_MEDIA_PREFIX = 'entry-input'
+
+_SPACY_MODEL = (
+    'en_core_web_sm @ https://github.com/explosion/spacy-models/releases/download/'
+    'en_core_web_sm-3.8.0/en_core_web_sm-3.8.0-py3-none-any.whl'
+)
```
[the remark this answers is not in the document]
christopherpestano: fixed

docs/release/howto/deployment/cloud.mdx
```diff
@@ -147,6 +147,8 @@ class Images(TableModel, name='images'):
     )
 ```
 
+Or set `db_input_media_dest` / `db_output_media_dest` on the database's `[[pixeltable.database]]` entry to change the default for inserted or computed media. Details: [Cloud storage](/integrations/cloud-storage#pixeltable-cloud-home-bucket).
```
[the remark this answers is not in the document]
christopherpestano: fixed

docs/release/integrations/cloud-storage.mdx
```diff
@@ -52,18 +52,31 @@ Do not put `PIXELTABLE_API_KEY` in `config.toml` as `api_key`. Use the env name.
...
 ## Configuration
 
-There are two ways to configure cloud storage destinations. These apply to **local** Pixeltable and to **bring-your-own** buckets. Hosted Cloud tables skip this section.
+Cloud storage destinations are set as a default for the database, or per column. **Local** Pixeltable and **bring-your-own** buckets need this. A hosted Cloud database defaults to its home bucket and only needs it to use a bucket of its own.
 
-### Global default destinations
+### Default destinations
 
-Set default destinations for all media columns with environment variables (see [Configuration](/platform/configuration)):
+A default destination applies to every media column that names none. It can be set in three places, listed here from highest to lowest precedence (see [Configuration](/platform/configuration#per-database-settings) for details).
```
aaron-siegel: "A default destination applies to every media column that names none." I had to read this several times to process it - it reads as though "none" is something one names. I don't think we need to "explain" what a default is?

Something like "You can set default destinations for all media columns in a variety of ways." would be shorter and clearer

docs/release/integrations/cloud-storage.mdx
```diff
@@ -52,18 +52,31 @@ Do not put `PIXELTABLE_API_KEY` in `config.toml` as `api_key`. Use the env name.
...
+
+**The database entry** in `pixeltable.toml`, for one database. The entry with no `name` configures the local database; an entry named `pxt://org:db` configures that hosted database:
```
aaron-siegel: "The entry with no `name`..." is confusing out of context. Better to link to a doc page that gives a top level understanding of the `[[pixeltable.database]]` config abstraction.

docs/release/integrations/cloud-storage.mdx
```diff
@@ -97,16 +110,18 @@ Then `pxt schema update app.py my_app`. In a notebook or a test, the same column
...
 
 Destinations are resolved in this order:
 1. **Explicit column destination**: highest priority (computed columns only)
-2. **Global default**: `PIXELTABLE_INPUT_MEDIA_DEST` / `PIXELTABLE_OUTPUT_MEDIA_DEST` (or the matching `config.toml` dest keys)
-3. **Hosted Cloud process**: `pxtfs://org:db/home` when the database is running in Cloud
-4. **Local process**: local disk
+2. **Database entry**: `db_input_media_dest` / `db_output_media_dest` in `[[pixeltable.database]]`
+3. **Environment variable**: `PIXELTABLE_INPUT_MEDIA_DEST` / `PIXELTABLE_OUTPUT_MEDIA_DEST`
+4. **Global config**: `input_media_dest` / `output_media_dest` under `[pixeltable]` in `config.toml`
+5. **Hosted Cloud process**: `pxtfs://org:db/home` when the database is running in Cloud
```
aaron-siegel: 5 and 6 aren't separate priorities (I realize this is a holdover from before - but might as well clean it up now): **Fallback destination**: `pxtfs://org:db/home` when the database is hosted in Cloud; local disk (usually `~/.pixeltable/media`) when it's running locally

docs/release/platform/configuration.mdx
```diff
@@ -55,12 +55,36 @@ rate_limit = 600  # Mistral uses a single rate limit for all models
...
+db_exporter_otlp_protocol = 'grpc'
+
+[[pixeltable.database]]
+name = 'pxt://myorg:mydb'
+db_input_media_dest = 's3://bucket/prod/input/'
+db_output_media_dest = 's3://bucket/prod/output/'
+db_exporter_otlp_endpoint = 'https://otlp.example.com'
+```
+
+Precedence, highest first: the database entry, then the environment variable from the table above or a `pxt.init()` override, then `[pixeltable]` / `[otel]` in `config.toml`, which every database on the installation shares. A hosted database with none of these uses its home bucket for media.
```
aaron-siegel: Maybe make this a bulleted list like the other one (I liked that format)

docs/release/integrations/cloud-storage.mdx
christopherpestano: Merged 5 & 6 into one item

docs/release/platform/configuration.mdx
christopherpestano: Changed to bulleted list

docs/release/integrations/cloud-storage.mdx
christopherpestano: Just removed that chunk and left it more simply "You can set default destinations for media columns in three places"

docs/release/integrations/cloud-storage.mdx
christopherpestano: Removed the llm verbage and linked to the "Cloud configuration reference"

pixeltable/config.py
```diff
@@ -64,6 +69,18 @@ class DatabaseConfig(pydantic.BaseModel):
     disk_gb: int | None = None
     workers: int | None = None
 
+    @pydantic.model_validator(mode='before')
+    @classmethod
+    def _collect_settings(cls, data: Any) -> Any:
```
mkornacker: Can we have a stricter type than Any here?

pixeltable/config.py
```diff
@@ -64,6 +69,18 @@ class DatabaseConfig(pydantic.BaseModel):
...
+        # moves the db_<key> keys of a parsed entry into `settings`, so they validate against DatabaseSetting
+        if not isinstance(data, dict):
+            return data
+        values = {setting.value for setting in DatabaseSetting}
+        settings = {key: data.pop(key) for key in list(data) if key in values}
+        if settings:
```
mkornacker: Use explicit comparisons, which we do everywhere else.

pixeltable/config.py
```diff
@@ -538,30 +549,30 @@ def __load_user_config(self) -> dict[str, dict[str, tuple[Any, Path]]]:
...
-    def __merge_databases(self, layers: list[tuple[list[DatabaseConfig], Path | None]]) -> list[DatabaseConfig]:
-        """Combine the database entries of the config files, entry by entry, each with the file it came from.
+    def __merge_databases(self) -> tuple[list[DatabaseConfig], Path | None] | None:
+        """Combine the database entries of the home and project configs, entry by entry, with the last file that
+        supplied any.
 
-        Entries are matched by name, and a field a later file sets wins, so a project adding a var keeps the
-        secrets the home config binds for the same database. The file that supplied each field, and each var
-        and secret binding, is recorded in __database_sources.
+        Entries are matched by name, and a field the project sets wins, so a project adding a var keeps the
```
mkornacker: This narrates the logic below. It's better to do that as comments in the body itself.

pixeltable/config.py
christopherpestano: Changed it to take and return dict[str, Any] and added a clearer error message for a list element that is not a table.

pixeltable/config.py
christopherpestano: Done.

pixeltable/config.py
christopherpestano: Moved the reasoning to inline comments.

pierrebrunelle:
grok review

docs/release/platform/cli.mdx
```diff
@@ -815,12 +815,20 @@ All cloud commands accept `--json` for machine-readable output.
...
 #### `[[pixeltable.database]]` -- what a project declares about one database
 
-Defines which project files the database gets, what goes into its image, and its compute settings.
+Defines which project files the database gets, what goes into its image, its compute settings, where its media goes, and where and how it exports telemetry.
 
 ```toml
 [[pixeltable.database]]
 name = 'pxt://myorg:prod'           # matches the URI passed to pxt db update
 
+# where media that names no destination goes (default: the database's home bucket)
```
pierrebrunelle, line 824: Delete this comment. It is the same "names none" phrasing already called out on this PR.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
```diff
@@ -56,7 +56,8 @@ def init(
 
     Call once, before the first Pixeltable operation. Each argument overrides the matching `[otel]`
     config setting and its standard `OTEL_*` environment variable; when an argument is left as None the
-    value resolves from that env var (highest priority) then the `[otel]` config section. Pass
+    value resolves from the database's `[[pixeltable.database]]` entry where it sets one (the endpoint and
```
pierrebrunelle, line 59: Name `db_exporter_otlp_endpoint` / `db_exporter_otlp_protocol` in the Args below. The summary already has that source and the Args do not.

docs/release/platform/observability.ipynb
```diff
@@ -69,7 +69,14 @@
     "export OTEL_EXPORTER_OTLP_HEADERS=\"Authorization=Basic $(echo -n '<instance-id>:<service-account-token>' | base64 | tr -d '\\n')\"\n",
     "```\n",
     "\n",
-    "**Option 2: `~/.pixeltable/config.toml`.** Generate the token first (`echo -n '<instance-id>:<service-account-token>' | base64 | tr -d '\\n'`), then add:\n",
+    "**Option 2: the database entry in `pixeltable.toml`.** Put the endpoint on the `[[pixeltable.database]]` entry of the database (the entry with no `name` is the local database; see [Configuration](/platform/configuration#per-database-settings)). The headers are a secret and stay out of the file: keep `OTEL_EXPORTER_OTLP_HEADERS` in the environment, and on a hosted database set it with `pxt secret set pxt://org:db OTEL_EXPORTER_OTLP_HEADERS=...`, then `pxt db restart` so the running database sees it.\n",
```
pierrebrunelle, line 72: Say headers stay out of the database entry. Option 3 still writes `exporter_otlp_headers` into `[otel]`.

docs/release/integrations/cloud-storage.mdx
```diff
@@ -52,18 +52,31 @@ Do not put `PIXELTABLE_API_KEY` in `config.toml` as `api_key`. Use the env name.
...
+
+**Environment variables**, for the current process:
 
 ```bash
 export PIXELTABLE_INPUT_MEDIA_DEST="s3://my-bucket/input/"
 export PIXELTABLE_OUTPUT_MEDIA_DEST="s3://my-bucket/output/"
 ```
 
-Or the same dest keys in `~/.pixeltable/config.toml` (not the API key):
+**The global config** in `~/.pixeltable/config.toml`, for every database on the installation (not the API key):
```
pierrebrunelle, line 79: Refer to the installation `config.toml`. Do not spell `~/.pixeltable/` in user text.

pixeltable/config.py
```diff
@@ -303,6 +326,9 @@ class Config:
     # the directory holding the project config file, or None when there is no project
     __project_root: Path | None
 
+    # database name -> field or 'vars.<name>' binding -> the file that supplied it
```
pierrebrunelle, line 329: Record `settings.<name>` here. `__merge_databases()` writes those keys, not only fields and `vars.<name>`.

packages/opentelemetry-instrumentation-pixeltable/tests/test_otel.py
```diff
@@ -190,6 +190,28 @@ def test_protocol_grpc(tmp_path: Path) -> None:
...
+        encoding='utf-8',
+    )
+    pxt_otel.init()
+    tp = trace.get_tracer_provider()
+    exporter = tp._active_span_processor._span_processors[0].span_exporter  # type: ignore[attr-defined]
+    assert exporter._endpoint == 'http://127.0.0.1:9/v1/traces', exporter._endpoint
+
+
+def test_endpoint_from_database_entry(tmp_path: Path) -> None:
+    # the entry for the process's database names the endpoint, in place of the [otel] section or the env var
```
pierrebrunelle, line 207: Replace `names the endpoint`. Say the entry sets it.

pierrebrunelle (commented):
Do not merge yet.

Delete `# where media that names no destination goes` from the CLI reference. That is the same "names none" phrasing already rejected on this PR.

`__resolve()` reads `db_<key>` on the process's `[[pixeltable.database]]` entry, then `pxt.init()` / env, then the shared `[pixeltable]` / `[otel]` sections. Hosted vs local selection and unknown-key rejection are tested.

`pxt.init()` in `pixeltable/globals.py` still says a process override beats the config file. That is false for these four keys.

Name `db_exporter_otlp_endpoint` / `db_exporter_otlp_protocol` in the `pxt_otel.init()` Args, not only in the summary paragraph.

Drop `~/.pixeltable/` from the user pages this PR edited, and stop telling Option 2 that OTLP headers stay out of every file while Option 3 writes them to `[otel]`.

docs/release/platform/cli.mdx
pierrebrunelle, line 824: Delete this comment. It is the slop phrase already called out on this PR.

docs/release/platform/observability.ipynb
```diff
@@ -69,7 +69,14 @@
...
+    "\n",
+    "```toml\n",
+    "[[pixeltable.database]]\n",
+    "db_exporter_otlp_endpoint = 'https://otlp-gateway-<zone>.grafana.net/otlp'\n",
+    "```\n",
+    "\n",
+    "**Option 3: `~/.pixeltable/config.toml`**, shared by every database on the installation. Generate the token first (`echo -n '<instance-id>:<service-account-token>' | base64 | tr -d '\\n'`), then add:\n",
```
pierrebrunelle, line 79: Refer to the installation `config.toml`. Do not spell `~/.pixeltable/`. Also drop `exporter_otlp_headers` from this Option 3 example; Option 2 already says headers stay out of config files.

docs/release/platform/observability.ipynb
```diff
@@ -49,8 +49,8 @@
     "## Turn on the OpenTelemetry bridge\n",
     "\n",
     "Telemetry is opt-in: call `pxt_otel.init()` once per process, before the first table operation. It\n",
-    "configures itself from environment variables or from the `[otel]` section of `~/.pixeltable/config.toml`\n",
-    "(env vars take precedence).\n",
+    "configures itself from the database's `[[pixeltable.database]]` entry in `pixeltable.toml`, from environment\n",
+    "variables, or from the `[otel]` section of `~/.pixeltable/config.toml`, in that order of precedence.\n",
```
pierrebrunelle, line 53: Same here: name the installation config file, not `~/.pixeltable/config.toml`.

docs/release/integrations/cloud-storage.mdx
pierrebrunelle, line 79: Refer to the installation `config.toml`. Do not spell `~/.pixeltable/` in user text.

docs/release/integrations/cloud-storage.mdx
```diff
@@ -305,14 +319,19 @@ https://www.pixeltable.com/dashboard/{org-slug}/{db-slug}/storage/home/browse
...
 ```toml
-[pixeltable]
-input_media_dest = "s3://my-app-bucket/uploads/"
-output_media_dest = "s3://my-app-bucket/generated/"
+[[pixeltable.database]]
+db_input_media_dest = "s3://my-app-bucket/uploads/"
+db_output_media_dest = "s3://my-app-bucket/generated/"
+```
+
+and, optionally, the AWS profile in `~/.pixeltable/config.toml` (default credentials are used if not set):
```
pierrebrunelle, line 330: Same: installation `config.toml`, not `~/.pixeltable/config.toml`.

docs/release/integrations/cloud-storage.mdx
```diff
@@ -97,16 +110,17 @@ Then `pxt schema update app.py my_app`. In a notebook or a test, the same column
...
 
 Destinations are resolved in this order:
 1. **Explicit column destination**: highest priority (computed columns only)
-2. **Global default**: `PIXELTABLE_INPUT_MEDIA_DEST` / `PIXELTABLE_OUTPUT_MEDIA_DEST` (or the matching `config.toml` dest keys)
-3. **Hosted Cloud process**: `pxtfs://org:db/home` when the database is running in Cloud
-4. **Local process**: local disk
+2. **Database entry**: `db_input_media_dest` / `db_output_media_dest` in `[[pixeltable.database]]`
+3. **Environment variable**: `PIXELTABLE_INPUT_MEDIA_DEST` / `PIXELTABLE_OUTPUT_MEDIA_DEST`
+4. **Global config**: `input_media_dest` / `output_media_dest` under `[pixeltable]` in `config.toml`
+5. **Fallback destination**: `pxtfs://org:db/home` when the database is hosted in Cloud; local disk (usually `~/.pixeltable/media`) when it is running locally
```
pierrebrunelle, line 123: Refer to the local media directory without spelling `~/.pixeltable/media`.

pixeltable/config.py
pierrebrunelle, line 329: Record `settings.<name>` here. `__merge_databases()` writes those keys, not only fields and `vars.<name>`.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
pierrebrunelle, line 59: Name `db_exporter_otlp_endpoint` / `db_exporter_otlp_protocol` in the Args below. The summary has that source and the Args do not.

packages/opentelemetry-instrumentation-pixeltable/tests/test_otel.py
pierrebrunelle, line 207: Say the entry sets the endpoint. Do not say it names it.

pierrebrunelle (commented):
Needs changes before merge.

Delete `# where media that names no destination goes` from the CLI reference. That is the slop phrase already rejected on this PR.

`__resolve()` reads `db_<key>` on the process's `[[pixeltable.database]]` entry, then `pxt.init()` / env, then the shared `[pixeltable]` / `[otel]` sections. Hosted vs local selection and unknown-key rejection are tested.

`pxt.init()` Args in `pixeltable/globals.py` still say a process override beats the config file. That is false for these four keys.

Drop `~/.pixeltable/` from the user pages this PR edited. Option 2 says OTLP headers stay out of config files while Option 3 still writes `exporter_otlp_headers` into `[otel]`.

Name `db_exporter_otlp_endpoint` / `db_exporter_otlp_protocol` in the `pxt_otel.init()` Args, and record `settings.<name>` in the `__database_sources` comment.

docs/release/integrations/cloud-storage.mdx
aaron-siegel: I usually address this by saying something like "The user config (typically `~/.pixeltable/config.toml`)" (since its location is configurable)

I don't know if "global config" is the right nomenclature, it's more of a user config (as opposed to project config)

docs/release/integrations/cloud-storage.mdx
```diff
@@ -97,16 +110,17 @@ Then `pxt schema update app.py my_app`. In a notebook or a test, the same column
 `t.add_computed_column(thumbnail=t.image.resize((128, 128)), destination=...)`.
 
 <Note>
-The `destination` parameter only applies to **stored computed columns**. For input columns, use the global `input_media_dest` configuration.
+The `destination` parameter only applies to **stored computed columns**. For input columns, set a default input destination as shown above.
```
aaron-siegel: There's a lot of random use of boldface in this text. Is it necessary? I find overuse of bold/italics highlighting to be alternately distracting and preachy. (I understand LLMs like it a lot)

docs/release/integrations/cloud-storage.mdx
aaron-siegel: How strongly do you feel about this? Here it does say "usually".

I find it helps to be concrete rather than indirect, and can clarify things especially for new users. The "usually" qualifier draws attention to the fact that it may be different on some systems

docs/release/platform/cli.mdx
```diff
@@ -815,12 +815,20 @@ All cloud commands accept `--json` for machine-readable output.
 
 #### `[[pixeltable.database]]` -- what a project declares about one database
 
-Defines which project files the database gets, what goes into its image, and its compute settings.
+Defines which project files the database gets, what goes into its image, its compute settings, where its media goes, and where and how it exports telemetry.
```
aaron-siegel: This is a mouthful. With so many items it'd be clearer as something like:

Defines the attributes of a specific Pixeltable database (which may be either the local database, or a cloud-hosted database), including:
- Which project files to include
- Which dependencies to include
- Compute settings
- Media destinations
- Telemetry configuration

packages/opentelemetry-instrumentation-pixeltable/tests/test_otel.py
```diff
@@ -190,6 +190,28 @@ def test_protocol_grpc(tmp_path: Path) -> None:
     )
 
 
+def check_endpoint_from_database_entry() -> None:
+    for k in [k for k in os.environ if k.startswith(('OTEL_EXPORTER_OTLP_ENDPOINT', 'OTEL_EXPORTER_OTLP_PROTOCOL'))]:
```
aaron-siegel: I always find this pattern really hard to read and I prefer the more traditional:
```
for k in os.environ:
    if k.startswith(...):
        ...
```

pixeltable/config.py
```diff
@@ -646,26 +689,45 @@ def get_database_config(self, db_uri: catalog.Path) -> DatabaseConfig | None:
...
-        ):
-            entry = config.get('pixeltable', {}).get('database')
-            if entry is None or not isinstance(entry[0], list):
-                continue
-            local = next((db for db in entry[0] if db.name == LOCAL_DATABASE), None)
-            if local is None:
-                continue
-            result.update({name: (value, source) for name, value in (local.vars or {}).items()})
-        return result
+        from pixeltable.env import Env  # env imports this module
```
aaron-siegel: I think I pointed this out in the previous review round. Importing `env` into config is an antipattern; the hosted_db() resolution logic should live in env. Config is really a glorified key/value store with typing hints and a defined resolution order.

pixeltable/config.py
```diff
@@ -692,18 +754,22 @@ def __lookup_config_entry(self, section: str, key: str) -> tuple[Any, Path | Non
...
-            if entry is None:
-                return None
-            value = entry[0]
-
-        if value is None:
-            return None  # Not specified
+    def __resolve(self, section: str, key: str) -> tuple[Any, Path | Literal['env'] | None] | None:
+        """The value of section.key and its source, in precedence order: the connected database's db_<key>, then a
+        pxt.init() override or the environment, then the config files."""
+        setting = self.__database_setting(section, key)
```
aaron-siegel, line 721: We're baking too much of this logic into Config itself IMO. It should live somewhere else (in Env or an auxiliary module that manages database configs.)

aaron-siegel (changes requested):

docs/release/integrations/cloud-storage.mdx
christopherpestano: I would disagree here, global config is the right nomenclature because it's relative to the project config which is per dir. Calling it "user config" makes it unclear about what it actually is (I mean isn't the user setting all the configs regardless?), it's the config that applies to all pixeltable projects thus "global"

christopherpestano: Also ~/.pixeltable keeps it obvious for users, better to be too explicit than leave users confused.

docs/release/platform/cli.mdx
christopherpestano: Honestly I feel like these are overblown, but changed to "without an" anyways

packages/opentelemetry-instrumentation-pixeltable/tests/test_otel.py
christopherpestano: Fixed

christopherpestano: Fixed

packages/opentelemetry-instrumentation-pixeltable/tests/test_otel.py
christopherpestano: Sure changed

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
christopherpestano: Huh no the args below are explicitly the non-db variant

docs/release/integrations/cloud-storage.mdx
christopherpestano: Removed

docs/release/integrations/cloud-storage.mdx
christopherpestano: I don't think it's wrong here to put ~/.pixeltable/media even for windows users powershell will redirect them to the appropriate folder were they to put that in terminal (just that for them the actual underlying file path is denoted differently).

docs/release/integrations/cloud-storage.mdx
christopherpestano: No let's be concrete rather than indirect

docs/release/platform/cli.mdx
christopherpestano: This feels like a nit, changed anyways

packages/opentelemetry-instrumentation-pixeltable/tests/test_otel.py
christopherpestano: Fixed

pixeltable/config.py
christopherpestano: Config does need to be aware of the current db for settings/var resolution, but I get your point. I changed it instead so that env gets these values from config on _set_up, and here instead of importing env we just directly do the get_string_value on self, ultimately the source of truth being in both cases being `get_string_value('db', section='pxtcloud')`.

pixeltable/config.py
christopherpestano: I'd have to disagree here, given that the authoritative source of where config files are combined and stored. It makes sense to me that if you ask it for the value of a given key that it gives you back the applicable value for the database that you are connected to. I did see some room for simplification to align the changes more with the original code, but the core idea that database resolution is outside of the scope of config seems off to me. Let's discuss this when we get a chance

christopherpestano:
grok review

pierrebrunelle (commented):
Do not merge yet.

`__resolve()` order, hosted vs local selection, and unknown-key rejection are tested. Config no longer imports env. The CLI "names none" phrasing is gone.

`pxt.init()` in `pixeltable/globals.py` still says a process override beats the config file. That is false when the process `[[pixeltable.database]]` entry sets `db_input_media_dest`, `db_output_media_dest`, `db_exporter_otlp_endpoint`, or `db_exporter_otlp_protocol`. (File not in this diff.)

Name `db_exporter_otlp_endpoint` and `db_exporter_otlp_protocol` in the `pxt_otel.init()` Args, not only in the summary. The omitted-kwarg path is `__resolve()`, database first.

Drop `exporter_otlp_headers` from the observability Option 3 `[otel]` example. Option 2 already says headers stay out of the file and live in the environment / `pxt secret set`.

The per-database intro in `configuration.mdx` still says the entry only overrides the shared section. It also beats env and `pxt.init()`. The otel package README still describes the old two-place order (file not in this diff).

Not repeated: Aaron's Config-vs-Env architecture threads; the `~/.pixeltable/` wording fight.

packages/opentelemetry-instrumentation-pixeltable/src/opentelemetry/instrumentation/pixeltable/_sdk.py
pierrebrunelle, line 59: Name `db_exporter_otlp_endpoint` and `db_exporter_otlp_protocol` in the Args below too. When those kwargs are None, `__resolve()` reads the database fields before env and `[otel]`.

docs/release/platform/observability.ipynb
pierrebrunelle: Remove `exporter_otlp_headers` from the Option 3 example below. Option 2 already puts headers in the environment or in `pxt secret set`.

docs/release/platform/configuration.mdx
```diff
@@ -55,12 +55,40 @@ rate_limit = 600  # Mistral uses a single rate limit for all models
...
 | PIXELTABLE_S3_PROFILE | [pixeltable]<br/>s3_profile | (string) Name of AWS config profile to use when accessing Amazon S3 resources. If not specified, default AWS credentials will be used. |
 | PIXELTABLE_B2_PROFILE | [pixeltable]<br/>b2_profile | (string) Name of an S3-compatible profile for accessing Backblaze B2. Defaults to the standard AWS credential chain if not set. |
 | PIXELTABLE_TIGRIS_PROFILE | [pixeltable]<br/>tigris_profile | (string) Name of an S3-compatible profile for accessing Tigris. Defaults to the standard AWS credential chain if not set. |
 
+### Per-database settings
+
+Media destinations and the OTLP endpoint and protocol can also be set for a single database. Each database has a `[[pixeltable.database]]` entry in `pixeltable.toml`. Add a `db_<key>` field to that entry, and it overrides `<key>` in the shared `[pixeltable]` or `[otel]` section for that database only. A `[[pixeltable.database]]` entry with no `name` corresponds to the local database. A `[[pixeltable.database]]` entry named `pxt://org:db` is a hosted database; its pods pick up the config when doing a `pxt db update`, and a running service moves onto it with `pxt service update`.
```
pierrebrunelle: Say the entry beats env and `pxt.init()` as well. The list below already has that order; this sentence still says it only overrides the shared section.

pixeltable/config.py
```diff
@@ -646,13 +658,16 @@ def get_database_config(self, db_uri: catalog.Path) -> DatabaseConfig | None:
...
 
-    def __database_bindings(self) -> dict[str, tuple[str, Path | None]]:
-        """Return the local database's vars, each with the file that supplied it.
+    def __own_database_values(self, field: Literal['vars', 'settings']) -> dict[str, tuple[str, Path | None]]:
+        """Return the `vars` or `settings` of the database we are connected to, each with the file that supplied it.
 
-        [[pixeltable.database]] is an array, which the section path of a var does not address; the path names
-        the entry for the local database, from which a process reads its vars. A binding the project supplies
-        wins over one of the same name in the home config.
+        On a hosted pod that is the entry named pxt://org:db. Anywhere else it is the local entry. Each file is read
```
pierrebrunelle: Rewrite this sentence. On a hosted pod the process reads the `pxt://org:db` entry; the pod is not that entry.

tests/pixeltable_cli/test_service.py
```diff
@@ -670,6 +679,14 @@ def test_media(
         body = resp.json()
         assert body['clip_id'] == 1, body
         assert pxt.get_table(f'{target}/frames').count() > 0
+        # the uploaded video is persisted where the database's entry sends inserted media for a hosted table, in
```
pierrebrunelle: Delete this comment. `expected_prefix` already says where the uploaded video is stored.

christopherpestano: Deleted

docs/release/platform/configuration.mdx
christopherpestano: Just removed the sentence since the below part already has the order

christopherpestano:
grok review

pierrebrunelle (commented):
Looks good on the last five asks. Head `70538fef`.

`pxt.init()` Args now say a `db_<key>` field on the process `[[pixeltable.database]]` entry beats `config_overrides`. `pxt_otel.init()` Args name `db_exporter_otlp_endpoint` and `db_exporter_otlp_protocol`; when those kwargs are omitted, `__resolve()` reads the entry first. Observability Option 3 no longer puts `exporter_otlp_headers` in `[otel]`. The configuration per-database list is database, then env / `pxt.init()`, then the shared sections. The otel README uses that same three-place order.

Against [PXT-1391](https://pixeltable.atlassian.net/browse/PXT-1391): dests and OTLP endpoint on the entry, headers as a secret / env (not on the entry). Matches the last review pass. Ticket Done-when still lists all four on the entry; leaving headers off the entry was the prior ask, so not re-opened here.

Still open from Aaron: Config-vs-Env on `__own_database_values` / `__resolve()` (CHANGES_REQUESTED). Not re-asked.

Sibling: open [#1417](https://github.com/pixeltable/pixeltable/pull/1417) rewrites the same `pxt_otel.init()` Args. Whoever merges second must keep database-first on the omitted-kwarg path and not drop `db_exporter_otlp_*`. [#1657](https://github.com/pixeltable/pixeltable/pull/1657) `Config.is_overridden()` docs still say `pxt.init()` outranks every config file; false for the four `db_*` keys after this lands.

Nits (non-blocking): `configuration.mdx` still says pods pick up config when doing `pxt db update` (the user runs that command; pods then read the uploaded project). Observability option numbers still list env before database against the intro’s database-first order.

aaron-siegel (approved):

christopherpestano:
grok review

docs/release/platform/configuration.mdx
```diff
@@ -55,12 +55,40 @@ rate_limit = 600  # Mistral uses a single rate limit for all models
...
 | PIXELTABLE_S3_PROFILE | [pixeltable]<br/>s3_profile | (string) Name of AWS config profile to use when accessing Amazon S3 resources. If not specified, default AWS credentials will be used. |
 | PIXELTABLE_B2_PROFILE | [pixeltable]<br/>b2_profile | (string) Name of an S3-compatible profile for accessing Backblaze B2. Defaults to the standard AWS credential chain if not set. |
 | PIXELTABLE_TIGRIS_PROFILE | [pixeltable]<br/>tigris_profile | (string) Name of an S3-compatible profile for accessing Tigris. Defaults to the standard AWS credential chain if not set. |
 
+### Per-database settings
+
+Media destinations and the OTLP endpoint and protocol can also be set for a single database. Each database has a `[[pixeltable.database]]` entry in `pixeltable.toml`. Add a `db_<key>` field to that entry to set `<key>` for that database only. A `[[pixeltable.database]]` entry with no `name` corresponds to the local database. A `[[pixeltable.database]]` entry named `pxt://org:db` is a hosted database; its pods pick up the config when doing a `pxt db update`, and a running service moves onto it with `pxt service update`.
```
pierrebrunelle: The user runs `pxt db update`. Pods then read the uploaded project. They do not run that command.

docs/release/platform/observability.ipynb
```diff
@@ -49,8 +49,8 @@
...
     "\n",
     "**Option 1: environment variables.** Set these **before starting the Jupyter kernel** (the kernel\n",
```
pierrebrunelle: The intro is database, then env, then `[otel]`. Number the options in that order.

pixeltable/config.py
```diff
@@ -752,17 +779,14 @@ def config_keys(self) -> list[ConfigKey]:
...
-        if entry is None:
+        resolved = self.__resolve(section, key)
+        if resolved is None or resolved[1] is None:
             return 'unset'
-        path = entry[1]
-        return path if path is not None else 'unset'
+        return resolved[1]
 
     def is_overridden(self, key: str, section: str = 'pixeltable') -> bool:
         """Whether pxt.init() supplied this setting, which outranks the environment and every config file.
```
pierrebrunelle: False for the four `db_*` keys. `__resolve()` reads the entry first.

pierrebrunelle (commented):
Looks good on this head (`cff96adf`, merge of main / #1657 onto `70538fef`). Prior five asks stay fixed. PXT-1391 dests and OTLP fields survived the merge. Config still does not import Env.

Must-fix
None.

Nits (non-blocking, same two as last pass, plus one docstring that arrived with #1657)

- `configuration.mdx`: the user runs `pxt db update`. Pods then read the uploaded project. They do not run that command.
- Observability options: the intro is database, then env, then `[otel]`. Number the options in that order.
- `Config.is_overridden()` (now in this file from #1657): the docstring still says `pxt.init()` outranks every config file. False for the four `db_*` keys.

Against [PXT-1391](https://pixeltable.atlassian.net/browse/PXT-1391): dests and OTLP endpoint on the entry, headers as a secret / env. Ticket Done-when still lists all four on the entry; not re-opened.

Not repeated: Aaron's Config-vs-Env threads; the `~/.pixeltable/` wording fight; headers on the entry.

Sibling: open [#1417](https://github.com/pixeltable/pixeltable/pull/1417) still documents the old env-then-`[otel]` omitted-kwarg path. Whoever merges second must keep database-first and not drop `db_exporter_otlp_*`. Open [#1662](https://github.com/pixeltable/pixeltable/pull/1662) folds `DatabaseConfig.name` to lowercase; this PR's `__own_database_values` compares `pxt://{org}:{db}` with `==`, so the second merge should fold the lookup name too.

docs/release/platform/configuration.mdx
christopherpestano: junk

docs/release/platform/observability.ipynb
christopherpestano: fixed

pixeltable/config.py
christopherpestano: fixed

PR #1639, opened 2026-09-17
pxt org create: make an organization from the CLI

amithadke:
```
pxt org create acme
pxt org create acme --name "Acme Inc" --location aws/us-east-1
```

One call makes the organization, its first database and its bucket. The control plane owns the whole
sequence — the WorkOS organization, the membership, the metadata, `main` — so the CLI needs no WorkOS
credentials and sends no org id.

Requires the cloud-side PR (pixeltable/pixeltable-cloud#226).

- Creating re-mints this machine's session scoped to the new organization. A token issued before it
  existed carries no claim for it, so without that `pxt db list` would fail on the very next line.
  If the re-mint fails the organization still exists, and it says exactly that rather than implying
  the create failed.
- Slug uniqueness is the control plane's to decide — only its database knows — so the CLI sends the
  name and reports what comes back.

Stacked on #1638.

### Note for reviewers

The control plane already has a `CreateOrgRequest` of its own, for the dashboard, and it
discriminates on the `operation_type` string. So there are deliberately two models for one
operation: this one omits `org_id`, and `'create_org'` is what joins them. Adding this model to the
cloud's request union would collide on that discriminator — there is a comment saying so.

### Verification

7 protocol tests; 83 in the stack. `make check` clean. Not yet exercised end to end against a live
environment — that needs the cloud side deployed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aaron-siegel:
Do we really want this? What's the user story for a Pixeltable end user creating an organization via the CLI?

PR #1640, opened 2026-09-17
Unpack archive before launching proxy daemon

aaron-siegel:
The image/archive split broke the deploy of Pixeltable core for testing purposes.

This fix (and its companion cloud PR#227) unpacks the archive *before* launching the proxy daemon. (In the present architecture, the proxy daemon drives the archive extraction; this fails for Pixeltable core, because the code to run the proxy daemon is inside the archive bundle.)


pixeltable/service/fetch_archive.py
```diff
@@ -0,0 +1,80 @@
+"""Unpack a hosted database's project archive, ahead of the pod that serves it.
+
+A pod cannot fetch its own archive. The image installs the project's dependencies but not the
+project itself (`uv sync --no-install-project`), because the image context carries only the
```
mkornacker: "carries", "named", etc.

mkornacker (approved):

PR #1641, opened 2026-09-17
pxt key: one command for API keys and scoped runtime keys

amithadke:
Stacked on #1639. Client side only — the control plane half is pixeltable/pixeltable-cloud#228.

One command, two shapes. `pxt key create ci` makes a key that acts as you. `pxt key create app --grant access:pxt://org:db/services/ingest` makes one that belongs to the organization and reaches only what it was granted.

A grant is a verb on a resource, never a bare resource: naming a service does not say whether the holder may call it or reconfigure it. `access` and `manage` are independent.

- `management_protocol.py` — `create_key`/`get_key`/`list_keys`/`update_key`/`delete_key`, `KeyRecord`; replaces the old `*ApiKey*` models
- `commands/key.py` — `pxt key list|create|update|delete`
- `server/routes.py`, `main.py` — wiring
- `docs/release/platform/cli.mdx` — the command, the grant table, the reference entry

`update` is a delta, so widening a key does not depend on resending what it already had. Revoking a grant the key never held is an error rather than a no-op — a typo there would otherwise read as a completed revocation.

No tests here: `pxt key` has none. Worth a contract test against the control plane's response shapes — that mismatch is what cost the time on this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pixeltable/service/management_protocol.py
```diff
@@ -425,3 +431,97 @@ class ListOrgsRequest(BaseModel):
 
 class ListOrgsResponse(BaseModel):
     orgs: list[OrgRecord]
+
+
+# API keys
```
mkornacker: This looks like it belongs in a docs section.

PR #1647, opened 2026-09-19
Docs: fix what executing every page's code found

pierrebrunelle:
Docs only. Third documentation pass; the first to **execute the code in the pages** rather than check that symbols exist.


pierrebrunelle:
Addressed in `52ceb1428`, one commit covering all six threads.

**spaCy model** (Devin, Copilot ×2): confirmed. `get_spacy_model()` calls `spacy.load()` and raises when `en_core_web_sm` is absent (`utils/spacy.py:31`); it never downloads. `iterators.mdx` and `doc-ingest-website.mdx` now give both steps.

**Notebook guideline** (Devin, Copilot): the previous commit fixed the introduction and left the checklist, example, and troubleshooting sections contradicting it, plus `AGENTS.md:378`. All now state one rule: one title source, a leading H1 or a raw frontmatter `title`, never both. The brittle "93 of 100" count is gone from the guideline and from `.github/instructions/docs.instructions.md`.

Also from checking the pinned doctools converter (`convert_notebooks.py:31-38`): it generates Kaggle, Colab, and **download** links, not a GitHub view link, and it hardcodes `blob/release/`. The guideline said "GitHub" and still told authors to hand-write links in frontmatter `description`; both corrected. The `release` tag it points at has not moved since 2026-09-02, two releases back. That is a doctools or release-process fix, not a notebook one, and is out of scope here.

**Iterator skeleton**: labeled as such, with the contract stated: after `seek(pos)`, the next `__next__()` returns the row at `pos`.

**skill.md** (Copilot): "uses its startup secret values", and the hosted service address is fully qualified.

pierrebrunelle:
`1739efffd`: `docs/release/skill.md` is now a pointer to the canonical skill in `pixeltable/pixeltable-skill` (install command plus a link to the raw `SKILL.md`), rather than a second copy that has to be kept in sync. Frontmatter stays valid so it is still recognized as a skill file. This supersedes #1649, which is closed. `.github/copilot-instructions.md:84` drops `skill.md` from the `app.py` co-change list; the other four copies still match `_EXAMPLE_APP`.

docs/release/skill.md
```diff
@@ -1,152 +1,22 @@
...
-```bash
-curl -X POST http://127.0.0.1:<port>/docs \
-  -H 'Content-Type: application/json' \
-  -d '{"title": "Hello", "body": "world"}'
-# {"id":"...","title_upper":"HELLO","summary":"Hello"}
-```
-
-`add_update_route` matches the row by its primary key, so the request body carries `id`
-even though `inputs` does not list it:
+The Pixeltable skill lives in one place, [pixeltable/pixeltable-skill](https://github.com/pixeltable/pixeltable-skill).
```
pierrebrunelle, line 12: `docs/release/resources/ai-skills.mdx` still says the hosted capability file is `docs.pixeltable.com/skill.md`. After this change that URL is a pointer (install command plus a link to `pixeltable/pixeltable-skill`), not the skill body. `2e25d531` and `c85a7932` already made that correction in `AGENTS.md` and `building-pixeltable-with-llms.mdx`. Please update `ai-skills.mdx` the same way before merge: `/skill.md` points at the canonical skill; install remains `npx skills add pixeltable/pixeltable-skill`.

AGENTS.md
```diff
@@ -375,8 +375,8 @@ order on every sentence added or edited:
 
 Documentation notebooks are in `docs/release/`. Follow `docs/_guidelines/GUIDELINES_FOR_NOTEBOOKS.md`:
 
-- Start with YAML frontmatter in a **Raw cell** (not Markdown)
-- No H1 headers in markdown (title comes from frontmatter)
+- Use one title source: a leading markdown H1, or a first **Raw cell** with YAML `title`
+- Do not include an H1 when using a raw frontmatter title
 - Use `##` for main sections, `###` for subsections
 - Clear outputs before committing unless output is instructive
```
pierrebrunelle, line 381: Nit: this PR updates the notebook title bullets here and in `.github/instructions/docs.instructions.md`. The instructions file now says never advise clearing all outputs (`tool/check_notebooks.py` requires outputs on at least 50% of code cells). This next bullet still says "Clear outputs before committing unless output is instructive," so an agent that follows it fails that check. Drop the bullet, or state the 50% rule.

pierrebrunelle (commented):
**Test review** (would be REQUEST_CHANGES if this weren't your own PR — GitHub blocks that).

Docs-only. The executed-example fixes (iterator/`unstored_cols`/`seek`, spaCy model download, `notes_app.py` imports, notebook title rule) match the code. The `skill.md` pointer is consistent with `AGENTS.md` and `building-pixeltable-with-llms.mdx`.

**Blocking before merge:** `docs/release/resources/ai-skills.mdx` still calls `/skill.md` the hosted capability file after this PR turned that file into a pointer. Same mismatch `2e25d531` and `c85a7932` already fixed elsewhere — update that sentence before merge.

Also a nit on the notebook outputs bullet in `AGENTS.md` (inline).

aaron-siegel (approved):

docs/release/skill.md
pierrebrunelle: Addressed in https://github.com/pixeltable/pixeltable/pull/1654.

AGENTS.md
pierrebrunelle: Addressed in https://github.com/pixeltable/pixeltable/pull/1654.

PR #1648, opened 2026-09-19
[PXT-1436] Name the install command in optional-package errors

pierrebrunelle:
## Summary

Optional-package errors (`require_package`, the spaCy model hint, and the
`[serve]`/`[otel]` extras hints) tell users to run `pip install`. For
installs through the Homebrew `pxt` formula, `pip` resolves outside the
formula's isolated virtualenv and cannot reach it, so the suggested
command fails (PEP 668) or installs into the wrong environment.

- Detect the formula venv via `sys.prefix` (`Cellar/pxt/`).
- In that venv, name `pxt-pip` (the shim the formula installs); kegs
  built before the shim fall back to `{sys.executable} -m pip`.
- Other environments keep `pip install`.
- `shlex.quote` keeps extras specs shell-safe (`'pixeltable[serve]'`).

#### Test plan

- [x] `TestInstallHints` covers all three branches plus the
      `require_package` error path
- [x] Verified end-to-end under the installed keg interpreter:
      `require_package("spacy")` emits `pxt-pip install -U spacy`

---

<a href="https://app.devin.ai/review/pixeltable/pixeltable/pull/1648" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


pierrebrunelle:
Closing: upstream prefers no Homebrew-specific handling in core error text. The formula-side fix (pxt-pip shim + caveats, pixeltable/homebrew-tap) already gives brew users a working install path; the residual here is that missing-package errors keep naming generic `pip install`, which cannot reach the formula venv. Can re-open if the team wants the env-aware hint later.

PR #1650, opened 2026-09-21
pxt login, pxt key, pxt whoami

mkornacker:


pierrebrunelle:
## Review (second pass, cross-checked with pixeltable/pixeltable-cloud#246)

**The WorkOS side is right and needs no change.** `pixeltable/service/auth.py` uses the native CLI Auth endpoints (`api.workos.com/user_management/authorize/device`, `/user_management/authenticate`), form-encoded, no client secret, RFC 8628 polling with `slow_down` backoff and the `expires_in` deadline, refresh-token rotation saved before the new access token is used. Verified today: the dev client id gets a device code from that endpoint with no WorkOS dashboard change and no second WorkOS application. Do not switch to `<authkit-domain>/oauth2/*` (that is WorkOS Connect; its tokens would fail the control plane's issuer check).

The "did not answer with JSON" error on dev is the control plane, not this code: dev does not serve `/.well-known/pixeltable-auth` until pixeltable-cloud#246 is published.

### Blockers

1. **Refresh never sends `organization_id`** (`pixeltable/service/auth.py:256-263`). After `pxt org create` the session stays org-less, and the control plane rejects every op except `create_org`. `organization_id` is read at `:198` and printed, but sent nowhere. Re-scope once after create (refresh with `organization_id=<new org>`) and save. `test_auth.py::test_renewal_rotated_token` only asserts `grant_type` and `refresh_token`.
2. **`pxt login` crashes on Windows.** `session_cache.py:66` calls `os.fchmod` (absent on win32, verified), and `:53-54` asserts `S_IMODE == 0o600`, which Windows reports as `0o666` even after `chmod`, so guarding `fchmod` alone is not enough. Windows CI is nightly only, so PR CI stays green. Guard both under `os.name == 'posix'` and replace the asserts (they also vanish under `-O`).
3. **`pxt logout` needs the network.** `server/routes.py:827-828` builds the browser logout URL (which fetches discovery) before `session_cache.clear`; any failure leaves the token on disk. Clear first, best-effort the URL.
4. **`split_pxt_uri` namespace change breaks the shared parser.** `pixeltable_cli/utils.py:79-90` + `:131-136`: executed, `resolve_dot_segments('pxt://acme:main/catalog/sub/..')` -> `'pxt://acme:main'` (namespace dropped), and `pxt://acme:main/catalog` yields `path=None`, so a directory literally named `catalog` is unaddressable. Parse grant URIs inside `commands/key.py` and leave the shared parser alone.

### Should-fix (follow-up acceptable)

- `auth.py:121-130`: error body parsed on 5xx too, so a WorkOS outage renders as "run `pxt login` again". Parse only on 4xx.
- `routes.py:814-820`: any `excs.Error` (including a 503) becomes "That credential was not accepted", exit 1. Distinguish provider auth errors.
- `management_client.py:36-46`: `LIST_KEYS` and `LIST_SECRETS` missing from `_READ_OPS` (no idle-connection retry). Two lines.
- `session_cache.py:91-107`: unsynchronized read-modify-write across processes; low frequency, `os.replace` keeps the file intact. Advisory lock later.
- `pxt org create` with an API key configured always fails; the fix is on the control plane (pixeltable-cloud#246 item 3), noted here because `docs/release/platform/cli.mdx` implies it works.
- `cli.mdx:747` "Creating one does not move you into it. Give the organization in the URI" describes a flow the control plane cannot serve: the org comes from the token claim, not the URI. Reword once item 1 lands.

### Copilot triage

Valid and covered above: safe cache validation, namespace preservation, OAuth errors on 4xx only, `LIST_KEYS` in `_READ_OPS`, clear credentials on logout, re-raise non-credential failures. Overstated/ignore: hide config path in messages (the path is the actionable part), `whoami` identity when the API key wins (the CLI already prints the precedence caveat), cross-process lock (disproportionate now), trailing-slash preservation.

### Merge order

Land pixeltable-cloud#246 and publish its Lambda first (discovery, `list_orgs` exemption, `update_db` guard), then this.

### Not verified

Tests were read, not run. Live WorkOS behavior for a multi-org user's device-grant token (org-scoped or not) was not exercised; if it is org-less there too, item 1 affects every multi-org user, not only new accounts.

pierrebrunelle:
Superseded by #1657, which continues from this branch with its history kept and fixes the review items. The paired control plane change is pixeltable/pixeltable-cloud#254. This core change merges first, but the PyPI release waits until the control plane is deployed and serves discovery. Thanks @mkornacker and @amithadke: you're both credited as co-authors in the squash.

PR #1652, opened 2026-09-21
[PXT-1386] Fix for cross-device move

aaron-siegel:


pixeltable/utils/local_store.py
```diff
@@ -152,7 +153,13 @@ def resolve_destination(
 
     def move_local_file(self, src_path: Path, dest: FileDestination) -> str | None:
         assert dest.local_path is not None
-        src_path.rename(dest.local_path)
+        try:
+            src_path.rename(dest.local_path)
+        except OSError as e:
+            if e.errno != errno.EXDEV:
+                raise
+            # report the move as unsupported, so that the caller falls back to copy-then-delete
```
pierrebrunelle: Delete this comment. Returning `None` is already `move_local_file()`'s "move not supported" signal, and `put_file_resolved()` already does the copy-then-delete fallback. The comment only restates that caller.

pierrebrunelle (commented):
**Written verdict: looks good to merge** (COMMENT only — not a formal approve).

`LocalStore.move_local_file()` was letting `OSError(EXDEV)` escape on cross-filesystem rename. Catching only `EXDEV` and returning `None` matches the existing `ObjectOps.put_file_resolved()` contract (copy then delete TempStore), while same-device stays an atomic rename and other `OSError`s still propagate. Catching only `EXDEV` is the right bound (narrower than `shutil.move()`, which swallows any `OSError` from `os.rename()`).

No public API change. The regression test drives the real fallback via monkeypatched `Path.rename`, checks dest bytes and source removal, is marked `db_roots('local')`, and belongs in default CI.

One inline nit on a restating comment. Residual (not blocking): fallback is not atomic (same as every non-local store); unrelated `Path.rename()` sites in `fetch_archive.py` / `transactional_directory.py` can still raise `EXDEV` and are out of scope for PXT-1386.

pixeltable/utils/local_store.py
```diff
@@ -152,7 +153,12 @@ def resolve_destination(
 
     def move_local_file(self, src_path: Path, dest: FileDestination) -> str | None:
         assert dest.local_path is not None
-        src_path.rename(dest.local_path)
+        try:
+            src_path.rename(dest.local_path)
+        except OSError as e:
+            if e.errno != errno.EXDEV:
```
sergey-mkhitaryan: i'd invert this, return None on EXDEV (and explain that this move is unsupported), otherwise raise

sergey-mkhitaryan (approved):

pixeltable/utils/local_store.py
aaron-siegel: Done.

PR #1653, opened 2026-09-21
[PXT-1385] Don't assert on failed object store save with on_error='ignore'

aaron-siegel:


pixeltable/exec/object_store_save_node.py
```diff
@@ -185,9 +185,9 @@ def __process_completions(self, done: set[futures.Future], ignore_errors: bool)
             new_file_url, exc = f.result()
             if exc is not None and not ignore_errors:
                 raise exc
-            assert new_file_url is not None
 
             if exc is None:
+                assert new_file_url is not None
```
pierrebrunelle, line 190: Assert move looks correct and matches the success-path idea in `CachePrefetchNode` (placement is slightly earlier than the use-site style in #1615, but fine).

Related gap once ignore works: single-destination TempStore sources still leak on this failure path. `put_file_resolved(..., relocate_or_delete=True)` only deletes after a successful copy/move, and `delete_destinations` only covers multi-dest paths (`unique_destinations[path] > 1`). After `set_exc` clears the row's file URL/path, nothing reclaims `work_designator.src_path` when `exc is not None`. Please delete the TempStore source once all work for that designator (or row) has finished — including failed work — or fold single-dest temps into the same post-completion cleanup as multi-dest. Without that, `on_error='ignore'` turns put failures into orphaned media under TempStore (the new tests hit this via `save_media_to_temp`).

pixeltable/exprs/row_builder.py
```diff
@@ -503,7 +503,9 @@ def set_exc(self, data_row: DataRow, slot_idx: int, exc: Exception) -> None:
         """Record an exception in data_row and propagate it to dependents"""
         data_row.set_exc(slot_idx, exc)
         for idx in self._exc_dependents[slot_idx]:
-            data_row.set_exc(idx, exc)
+            if not data_row.has_exc(idx):
```
pierrebrunelle, line 507: Good catch relative to #1614/#1615. `DataRow.set_exc` asserts `excs[slot_idx] is None`, so two failing parents that share a dependent (see `test_dest_put_failure_shared_dependent`) used to re-trip an AssertionError after the ObjectStoreSaveNode assert was fixed. First-wins is the right invariant here.

tests/test_destination.py
```diff
@@ -382,6 +382,56 @@ def test_dest_local_copy(self, uses_db: None) -> None:
...
+        status = t.insert(rows, on_error='ignore')
+        assert status.num_rows == 1
+        assert status.num_excs >= 1
+        assert 'test_dest_put_failure.img_rot' in status.cols_with_excs
+
+        res = t.select(msg=t.img_rot.errormsg, url=t.img_rot.fileurl).collect()
+        assert len(res) == 1
+        assert 'injected put failure' in res[0]['msg']
+        assert res[0]['url'] is None
+        assert ObjectOps.count(t._id, dest=dest_uri) == 0
```
pierrebrunelle, line 410: Dest object count is a good check. If you add TempStore cleanup for failed single-dest puts, please also assert the TempStore source is gone after the ignore path (file count / path absence) so this case cannot regress silently.

pierrebrunelle (commented):
## Verdict: needs changes before merge

Fixes the PXT-1385 assert crash under `on_error='ignore'`, and correctly adds the second assert surface that #1614/#1615 missed (`DataRow.set_exc` when two failing parents share a dependent). Tests are stronger than the earlier attempts. One cleanup gap should be closed before merge (or explicitly deferred with a follow-up that blocks shipping the ignore path alone).

### What changed
1. **`ObjectStoreSaveNode.__process_completions`** — move `assert new_file_url is not None` under `if exc is None` so a failed put can fall through to `row_builder.set_exc` instead of raising `AssertionError`.
2. **`RowBuilder.set_exc`** — skip dependents that already have an exception. Required because `DataRow.set_exc` asserts `excs[slot_idx] is None`; two failed media slots sharing a dependent used to re-trip that assert after the ObjectStoreSaveNode fix.
3. **Tests** — abort vs ignore regression plus shared-dependent collision coverage.

### Blocking follow-through
Failed **single-destination** TempStore persists under `on_error='ignore'` still leave orphaned media. `put_file_resolved(..., relocate_or_delete=True)` only deletes after a successful copy/move, and `delete_destinations` only lists multi-dest paths. After `set_exc` clears the row URL/path, nothing reclaims the TempStore `src_path` on the failure path. The new tests exercise exactly this via `save_media_to_temp`. Making ignore work without that cleanup turns put failures into durable TempStore orphans after a successful insert.

### Otherwise solid
- Assert placement differs slightly from `CachePrefetchNode` / #1615 (here at the top of the success block vs at the use site) but is correct.
- `RowBuilder` comment is justified (design choice not obvious from the guard alone).
- CI still pending at review time (`generate-matrix` green; slim/minimal/static-checks not settled).

Compared to closed #1614/#1615: same ObjectStoreSaveNode intent; the `row_builder` change + shared-dependent test are the material upgrades.

pixeltable/exprs/row_builder.py
```diff
@@ -503,7 +503,9 @@ def set_exc(self, data_row: DataRow, slot_idx: int, exc: Exception) -> None:
         """Record an exception in data_row and propagate it to dependents"""
         data_row.set_exc(slot_idx, exc)
         for idx in self._exc_dependents[slot_idx]:
-            data_row.set_exc(idx, exc)
+            # DataRow.set_exc() asserts the slot is unset, and several failing dependencies can share a dependent
+            if not data_row.has_exc(idx):
```
sergey-mkhitaryan, line 507: This only protects the dependencies, but the same problem exists with the primary slot -- see `data_row.set_exc(slot_idx, exc)` above.

Also notice that `ExprEvalNode.dispatch_exc()` implements its own error propagation, calling `DataRow.set_exc()` directly. This complicates things, and I'm sure that Claude can write a repro via that path.

Maybe the right approach is to move the first-write-wins down to `DataRow.set_exc()`, and drop the assert there. Not sure.

sergey-mkhitaryan (changes requested):

tests/test_destination.py
```diff
@@ -382,6 +382,56 @@ def test_dest_local_copy(self, uses_db: None) -> None:
         # Ensure that local file is copied to a specified destination
         assert ObjectOps.count(t._id, dest=dest1_uri) == len(r)
 
+    @pytest.mark.db_roots('local', reason='media destination/object-store internals')
+    def test_dest_put_failure(self, monkeypatch: pytest.MonkeyPatch, uses_db: None) -> None:
+        """A failed persist aborts under on_error='abort' and becomes a cell error under on_error='ignore'."""
+        dest_uri = f'{self.resolve_destination_uri(StorageTarget.LOCAL_STORE)}/bucket1'
+        t = pxt.create_table('test_dest_put_failure', schema={'img': pxt.Image | None})
+        t.add_computed_column(img_rot=t.img.rotate(90), destination=dest_uri)
```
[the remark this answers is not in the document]
aaron-siegel: This was a false instruction in `tests.instructions.md`.

sergey-mkhitaryan:
grok review

pixeltable/utils/object_stores.py
```diff
@@ -570,16 +570,19 @@ def put_file_resolved(
         """Worker-thread version of put_file: performs move-or-copy with no catalog access."""
         from pixeltable.utils.local_store import TempStore
 
-        if relocate_or_delete:
-            # File is temporary, used only once, so we can delete it after copy if it can't be moved
-            assert TempStore.contains_path(src_path)
+        if not relocate_or_delete:
+            return store.copy_local_file(src_path, dest)
+
+        # src_path is a single-use TempStore file
```
pierrebrunelle: Delete this comment. The assert on the next line already states that src_path is in TempStore.

pierrebrunelle (commented):
Looks good on 110922ae.

Prior asks are fixed. `put_file_resolved()` now deletes a single-use TempStore source in `finally` when the put fails. `DataRow.set_exc()` keeps the first exception, so the primary slot and `ExprEvalNode.dispatch_exc()` no longer assert. `test_dest_put_failure` asserts `TempStore.count(t._id) == 0`.

PXT-1385 holds: `on_error='ignore'` stores the row and records the upload error on `img_rot`; `on_error='abort'` still raises the put `OSError`. Copilot's `DatabaseRoot` flag was a bad instruction; the `tests.instructions.md` correction matches these `uses_db` monkeypatch tests.

Sergey's `CHANGES_REQUESTED` review can be re-checked. The thread is already resolved.

### Must-fix

None.

### Nits

Delete the comment above `assert TempStore.contains_path(src_path)` in `put_file_resolved()`. The assert already states that.

sergey-mkhitaryan (approved):

PR #1655, opened 2026-09-22
Docs: document stitch_tiles order-expression argument

### LINKED ISSUE #1595: `image.stitch_tiles` docstring omits its required leading order-expression argument
`pixeltable/functions/image.py` declares:

```python
@pxt.uda(requires_order_by=True)
class stitch_tiles(pxt.Aggregator):
```

`requires_order_by=True` means the ordering expression is the **first positional argument**, and passing `order_by=` raises (`func/aggregate_function.py` rejects it with "order_by invalid, this function requires the first argument to be the ordering expression").

The docstring's `Args:` section documents four parameters — `tile`, `tile_box`, `width`, `height` — while its own `Examples:` block passes five:

```python
stitch_tiles(tiles.pos, tiles.overlay, tiles.tile_box, tiles.image.width, tiles.image.height)
```

Nothing in the docstring explains the extra leading argument, so the signature and the example disagree with no way to reconcile them from the docs alone.

### Contrast with `make_video`

`pixeltable/functions/video/editing.py` has the identical decorator and states the convention outright:

> Call as `make_video(order_expr, frame, *, fps=25)`. `order_expr` is a frame index or timestamp. `order_by=` is invalid.

That sentence was added in 0.7.5. `stitch_tiles` has no equivalent, and it is the only other `requires_order_by=True` UDA in `pixeltable.functions` — so the two built-ins that share this calling convention document it inconsistently.

### Fix

Add the same "Call as ..." line to `stitch_tiles`, and name the order expression in its `Args:`.

Found while auditing UDA calling conventions for an agent-facing skill. The failure mode is that the natural guess — `order_by=` — raises at runtime rather than being caught statically.

pierrebrunelle:
## Summary

- `stitch_tiles` is `requires_order_by=True`, so the ordering expression is the first positional argument and `order_by=` raises at runtime. The docstring's example passed five arguments while `Args:` documented four.
- Adds the same "Call as ..." convention line `make_video` uses, and names `order_expr` in `Args:`.

Closes #1595

#### Test plan

- [x] `ruff format --check` and `ruff check` pass on `pixeltable/functions/image.py`
- [x] Docstring-only change, no behavior change

Generated with [Devin](https://devin.ai)

---

<a href="https://app.devin.ai/review/pixeltable/pixeltable/pull/1655" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


pierrebrunelle (commented):
Looks good to merge.

Docstring-only fix for #1595. `stitch_tiles` is `requires_order_by=True`, so the ordering expression is the first positional argument and `order_by=` raises; the example already passed five arguments while Args named four. The new Call-as line matches `make_video`, and naming `order_expr` in Args closes that gap. Runtime behavior and existing stitch tests are unchanged.

Note: `make_video` explains `order_expr` in the Call-as paragraph and keeps it out of Args. Putting it under Args here matches the issue ask and this docstring's existing habit of documenting the call-site arguments.

pixeltable/functions/image.py
aaron-siegel, line 604: I realize this isn't new to this PR, but this isn't at all what this example is doing.

aaron-siegel (approved):

pixeltable/functions/image.py
```diff
@@ -584,7 +584,11 @@ class stitch_tiles(pxt.Aggregator):
     The reconstruction is naive: tiles are pasted in the order they arrive, so where tiles overlap, later tiles
     overwrite earlier ones. It composites pixels and does not merge or deduplicate detections across tile seams.
 
+    Call as `stitch_tiles(order_expr, tile, tile_box, width, height)`. `order_by=` is invalid.
+
     Args:
+        order_expr: The expression that orders the tiles within a group, typically the tile position emitted
```
aaron-siegel: Not a good parameter name; there's no other place in the Pixeltable docs where we put `_expr` in a UDF/UDA parameter.

`order_key` would be a better parameter name.

aaron-siegel (changes requested):

pixeltable/functions/image.py
pierrebrunelle: Renamed to `order_key` in 5a86f2d67, here and in `make_video`, which was the one other docstring using `order_expr`, so the two stay consistent.

The Args entry now also says "Positional only". The aggregate takes `args[0]` and has no parameter by that name, so `stitch_tiles(order_key=tiles.pos, ...)` raises "requires an ordering expression as its first argument". An unqualified Args entry would suggest that keyword works.

pixeltable/functions/image.py
pierrebrunelle: Agreed, fixed in 5a86f2d67. The prose now describes only the stitch step the code performs, and links the cookbook for the full tile, segment, stitch pipeline.

Looking into this turned up a real bug: the cookbook called `stitch_tiles` without the ordering expression (4 args), which fails with `missing a required argument: 'height'`. Its saved output predated that. Added `tiles.pos`; the cell needs a real re-run to refresh the output, which I will do before merge.

PR #1656, opened 2026-09-22
exclude cloud tests from make slimtest

sergey-mkhitaryan:
I expect `make slimtest` to finish quickly, that is the main point of it. Cloud tests slow it down by a large factor.


pierrebrunelle (commented):
Looks good to merge.

`-k 'not cloud'` drops the `db_roots` cloud / cloud-cli parametrizations from the slim file list when a cloud env is configured. That matches what slimtest is for. No other files needed.

aaron-siegel (approved):

